PR #1112, opened 2020-10-22
Extension support page

### LINKED ISSUE #742: Add link to support page on bottom bar
We should add link to support page (life ring icon) to bottom/status bar. The location should be right corner and `Workspace` switcher should be moved to the left corner. Support page will contain links to open Github issue, link to community slack and option for commercial support.

There should be also `Support` system menu item under Help, that will also open the support page.

ixrock:
Add link to support page on bottom bar, close #742

**PR's most noticeable changes:**

- Support page extension: 
  - adds top menu item (`Help -> Support`)
  - adds status-bar support-ring icon (app's bottom-right corner)
  - adds global page accessible from `/support`
- switching to manual updates for `material-icons-font` from master repo (`npm` doesn't contain the latest icons pack, e.g. `support` icon)
- `Tooltip` component fixes: previously didn't work in some case (e.g. small icon in the screen corner with bigger tooltip)
- `PageLayout` new common component: used for `Support`, `Preferences` and `ClusterSettings` pages

**Screenshots**

Screenshot 2020-10-22 at 13 19 46
Screenshot 2020-10-22 at 13 20 18


src/extensions/extension-loader.ts
```diff
@@ -35,42 +34,43 @@ export class ExtensionLoader {
...
     })
   }
 
-  protected autoloadExtensions(getLensRuntimeEnv: () => LensExtensionRuntimeEnv, callback: (instance: LensExtension) => void) {
+  protected autoloadExtensions(callback: (instance: LensExtension) => void) {
     return reaction(() => this.extensions.toJS(), (installedExtensions) => {
-      for(const [id, ext] of installedExtensions) {
-        let instance = this.instances.get(ext.manifestPath)
+      for (const [id, ext] of installedExtensions) {
+        let instance = this.instances.get(ext.name)
```
ixrock, line 64: Without this change same extension might be enabled multiple times..
Possible reason: initial checking for `ext.manifestPath`, but actual saving as installed-extensions happened via `ext.id` key

src/extensions/lens-extension.ts
```diff
@@ -34,7 +33,6 @@ export class LensExtension implements ExtensionModel {
   @observable manifest: ExtensionManifest;
   @observable manifestPath: string;
   @observable isEnabled = false;
-  @observable.ref runtime: LensExtensionRuntimeEnv;
```
ixrock, line 37: Removed initially planned "runtime". Now everything could be served via webpack's externals into global Lens runtime.

extensions/telemetry/package.json
```diff
@@ -9,8 +9,8 @@
     "styles": []
   },
   "scripts": {
-    "build": "webpack --config webpack.config.js",
-    "dev": "npm run build --watch"
```
ixrock, line 13: For some reasons reusing command didn't work for me from WebStorm (maybe because need of `$@` at the end or different running context, dunno). Btw `webpack.config.js` is a default config name (even `.ts` is supported, when used in conjunction with `ts-node`).

src/extensions/core-api/index.ts
```diff
@@ -0,0 +1,22 @@
...
+import type { WindowManager } from "../../main/window-manager";
+
+// APIs
+import * as EventBus from "./event-bus"
+import * as Store from "./stores"
+import * as Util from "./utils"
+import * as Registry from "../registries"
+import * as CommonVars from "../../common/vars";
+
+export let windowManager: WindowManager;
```
ixrock, line 14: Exported in runtime in `src/main/index.ts`

src/renderer/i18n.ts
```diff
@@ -14,7 +14,7 @@ export interface ILanguage {
 
 export const _i18n = setupI18n({
   missing: (message, id) => {
-    console.warn('Missing localization:', message, id);
+    // console.warn('Missing localization:', message, id);
```
ixrock, line 17: Turned off to avoid noise for now until we have officially `i18n` supported.

extensions/support-page/main.ts
```diff
@@ -0,0 +1,23 @@
...
+export default class SupportPageMainExtension extends LensMainExtension {
+  async onActivate() {
+    console.log("support page extension activated")
+  }
+
+  async registerAppMenus(registry: Registry.MenuRegistry) {
+    this.disposers.push(
+      registry.add({
+        parentId: "help",
+        label: "Support",
```
jim-docker, line 13: this basically duplicates the "Community Slack" and "Report an Issue" Help submenu items that are already there...

src/extensions/extension-api.ts
```diff
@@ -1,2 +1,4 @@
-export * from "./core-extension-api"
-export * from "./renderer-extension-api"
+// Extension-api types generation bundle (used by rollup.js)
+
+export * from "./core-api"
```
jim-docker, line 3: should this be main-api for consistency?

extensions/example-extension/renderer.tsx
```diff
@@ -10,7 +10,7 @@ export default class ExampleExtension extends LensRendererExtension {
   registerPages(registry: Registry.PageRegistry) {
     this.disposers.push(
       registry.add({
-        type: Registry.DynamicPageType.CLUSTER,
+        type: Registry.PageRegistryType.CLUSTER,
```
jim-docker, line 13: I know it's out of scope for this PR but why can't `LensRendererExtension` have something like:
```
  registerPages(registry: Registry.PageRegistry) {
    for (let page of this.pages()) {
      this.disposers.push(registry.add(page))
  }
```
and then the subclasses just have to implement `pages()` which returns an array of the pages to add?

src/extensions/extension-loader.ts
```diff
@@ -35,42 +34,43 @@ export class ExtensionLoader {
...
+    this.autoloadExtensions((instance: LensMainExtension) => {
+      instance.registerAppMenus(menuRegistry);
     })
   }
 
-  loadOnMainRenderer(getLensRuntimeEnv: () => LensExtensionRuntimeEnv) {
-    logger.info('[EXTENSIONS-LOADER]: load on main renderer')
-    this.autoloadExtensions(getLensRuntimeEnv, (instance: LensRendererExtension) => {
+  loadOnClusterManagerRenderer() {
+    logger.info('[EXTENSIONS-LOADER]: load on main renderer (cluster manager)')
```
jim-docker, line 45: what does `main renderer` mean?

src/extensions/extension-loader.ts
```diff
@@ -35,42 +34,43 @@ export class ExtensionLoader {
...
+    this.autoloadExtensions((instance: LensRendererExtension) => {
       instance.registerPages(pageRegistry)
```
jim-docker, line 47: should this line be removed from here since it is added to `loadOnClusterRenderer()`?

src/renderer/components/layout/page-layout.tsx
```diff
@@ -0,0 +1,82 @@
...
+  back?: (evt: React.MouseEvent | KeyboardEvent) => void;
+}
+
+const defaultProps: Partial<PageLayoutProps> = {
+  provideBackButtonNavigation: true,
+  contentGaps: true,
+}
+
+@observer
+export class PageLayout extends React.Component<PageLayoutProps> {
```
jim-docker, line 26: nice

jim-docker (approved):

src/extensions/extension-loader.ts
ixrock: Nope, it's registering pages in different contexts (cluster-manager and dashboard / single cluster)

src/extensions/extension-loader.ts
ixrock: Main renderer (term added not by me) means it's a top renderer for main `BrowserWindow` when app starts and "not-main" renderer it's a renderer process(es) (child renderer process in iframe) where dashboard for single cluster is running.

extensions/example-extension/renderer.tsx
ixrock: This API definitely will be improved to git rid of everything redundant for end user of the extensions.

src/extensions/extension-api.ts
ixrock: It's not only about `main`. Currently this api also consumed in renderer process (whole file imported in `bootstrap.tsx`).
Anyway I would split both apis with strong separation what's and where included later for more future-proof concept.

extensions/support-page/main.ts
ixrock: So need to remove from that place? What is the question / propose?

jim-docker: Probably remove this "Support" menu item (although it's nice to demonstrate that this can be done). The existing menu items have the advantage of taking you directly to the link in one click, whereas with "Support" it's one click to get to the support page then another to get to the link.

ixrock:
@jakolehm can this be merged already?

src/extensions/extension-loader.ts
jakolehm: Yeah, I noticed this bug also in #1125.

src/extensions/registries/page-registry.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import type React from "react";
+import type { RouteProps } from "react-router";
+import type { IconProps } from "../../renderer/components/icon";
+import type { IClassName } from "../../renderer/utils";
+import type { TabRoute } from "../../renderer/components/layout/tab-layout";
+import { BaseRegistry } from "./base-registry";
+import { computed } from "mobx";
+
+export enum PageRegistryType {
```
jakolehm, line 11: It feels that we should have separate APIs for each type (just to be more futureproof). Ofc that should be done in a separate PR.

jakolehm (approved):

PR #1113, opened 2020-10-22
Add more telemetry data

nevalla:


src/extensions/core-api/stores.ts
```diff
@@ -1 +1,3 @@
 export { ExtensionStore } from "../extension-store"
+export { clusterStore, ClusterModel } from "../../common/cluster-store"
```
nevalla, line 2: What type we should export for `Cluster`?

jakolehm (approved):

src/extensions/core-api/stores.ts
nevalla: `Cluster` type is now also exported, so maybe use it here too

src/extensions/core-extension-api.ts
```diff
@@ -7,10 +7,12 @@ export * from "./lens-renderer-extension"
 import * as EventBus from "./core-api/event-bus"
 import * as Store from "./core-api/stores"
 import * as Util from "./core-api/utils"
+import * as App from "./core-api/app"
```
ixrock: BTW this file is actually moved to `extensions/core-api/index.ts`
But it seems it's still exists in base branch (removed in follow up PR: https://github.com/lensapp/lens/pull/1140)

nevalla: fixed and removed also here

Issue #430, opened 2020-06-10
Improved logs UI


jakolehm:
**What would you like to be added**:

- Get rid of modal -> logs would be visible as tabs in the dock.
- Selectable time-window
- Search logs

donbeave:
It will be nice to add ability to enable/disable timestamp in the logs line. For example, the default behavior for Java apps (especially Spring Boot) is print logs with date-time, in this case, Lens will have almost duplicated information (for each line):

<img width="1765" alt="image" src="https://user-images.githubusercontent.com/139017/85364878-b6635700-b556-11ea-84a1-ec6fa668712e.png">

Maybe it's better to switch off timing by default and enable it when users want it for particular deploys?

donbeave:
It will be nice to make logs as a separate window and also bigger with the ability to view logs for different pods in the same time. I mean multiple opened windows.

ppalucha:
Search would be nice...

devth:
Also: tail the logs for an entire namespace instead of only individual pods.

mih-kopylov:
@devth or for a deployment / service (when there's a bunch of pods inside)

Nuru:
@donbeave wrote:
> It will be nice to add ability to enable/disable timestamp in the logs line. 

That feature already exists. Click the clock icon ![Clock Icon](https://user-images.githubusercontent.com/2793313/93032810-5a095400-f5e8-11ea-93d6-cf24284bb26b.png) in the upper right of of the dialog to toggle the timestamps.

Nuru:
I suggest looking at [stern](https://github.com/wercker/stern) as a source of suggested features for combining logs, particularly combining all logs for a DaemonSet or ReplicaSet into one stream.

I also would very much appreciate the ability to search and filter the logs, similar to what is available with  `less` and `grep`. 

It would also be helpful to be able to open the logs in a separate window, so I can view the logs as I work on other resources.

aleksfront:
Closed in https://github.com/lensapp/lens/pull/1114
Selectable time-window, workload logs and other improvements are listed in their own issues.

kilpatty:
Not sure if I'm missing this, but is there a link to the time-window PR @aleksfront?

aleksfront:
[...]

PR #1114, opened 2020-10-22
Log search

aleksfront:
Part of #430 
- Adds search controls to the right of toolbar and allows to search through loaded log buffer.
- Replaced log output from raw text container to optimized virtual list component to keep fast rendering performance regardless of log amount.
- Highlights log row on hover
- Focuses search field when keypressed `CmdOrCtrl+F`

log scrolling
log searching


src/renderer/components/dock/pod-log-search.tsx
```diff
@@ -0,0 +1,70 @@
...
+        material="keyboard_arrow_down"
+        tooltip={_i18n._(t`Next`)}
+        onClick={onNextOverlay}
+        disabled={jumpDisabled}
+      />
+      <Icon
+        material="close"
+        tooltip={_i18n._(t`Clear`)}
+        onClick={() => setSearch("")}
+      />
```
jim-docker: I guess it doesn't work keeping it inside the SearchInput. It might fit better to the left of the up/down arrows if not in the SearchInput?

src/renderer/components/dock/search.store.ts
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  /**
+   * Does searching within text array, create a list of search keyword occurences.
+   * Each keyword "occurency" is saved as index of the the line where keyword founded
+   * @param text An array of any textual data (logs, for example)
+   * @param query Search query from input
+   * @returns {Array} Array of line indexes [0, 0, 14, 17, 17, 17, 20...]
+   */
+  findOccurencies(text: string[], query: string) {
```
jim-docker:
```suggestion
  findOccurences(text: string[], query: string) {
```

jim-docker (commented):
It appears to include the timestamps in the search even if the timestamps are not shown (I searched for '2'). This is confusing when clicking next as the orange highlight disappears.

src/renderer/components/dock/pod-log-search.tsx
aleksfront: I follow the usual rules for placing elements in the search, such as native search in the browser.

nevalla:
Looks nice!

Some improvement requests:
- would be nice if you could jump to next item immeditaly from search field by pressing enter key (like VS code does). Now it requires arrow down icon click and then you can use enter.
- Would be nice to have search bar activated by cmd + F (or ctrl + F), although potentially we could have multiple search bars visible the same time which might make things complicated

aleksfront:
> * Would be nice to have search bar activated by cmd + F (or ctrl + F), although potentially we could have multiple search bars visible the same time which might make things complicated

Can you describe it a bit more? What you mean by activating a search? And how many search inputs could be visible since we have one for each log tab? Also, almost every time we have table search bar visible on top of the page, but it didn't correlates with logs in any way.

nevalla:
> Can you describe it a bit more? What you mean by activating a search? And how many search inputs could be visible since we have one for each log tab? Also, almost every time we have table search bar visible on top of the page, but it didn't correlates with logs in any way.

Yes I was meaning search bar on top of the page. Ideally it should work like above (activate == move focus)

image

image

src/renderer/components/dock/pod-log-search.tsx
jim-docker: I see. I didn't make the connection because the input field is delineated by a rectangle whereas the native search in the browser delineates the whole search tool by a rectangle, and separates the input field from the "prev next clear" buttons with just a vertical line

jim-docker:
This looks really nice but there's something funky with the timestamps. If you click the timestamps icon, it toggles. If you then press Enter it toggles timestamps again. Is this expected?  Sometimes though the icon stays 'bright' regardless of the actual timestamp toggle state, so I think there is a bug there somewhere.

If Enter should only effect the 'search next' operation can you force the text cursor to stay in the search edit field?

aleksfront:
> This looks really nice but there's something funky with the timestamps. If you click the timestamps icon, it toggles. If you then press Enter it toggles timestamps again. Is this expected? Sometimes though the icon stays 'bright' regardless of the actual timestamp toggle state, so I think there is a bug there somewhere.

Yes, its a known issue caused by same styles applied to `:hover` and `:focus` elements so there's illusion that icons preserve their state after clicked. I will redefine `:focus` styling to avoid this problem. Maybe will create a blue outline which is quite common.

src/renderer/components/dock/pod-logs.tsx
```diff
@@ -86,60 +87,124 @@ export class PodLogs extends React.Component<Props> {
...
+    setTimeout(() => this.hideHorizontalScroll = false, 500);
+  }
+
+  /**
+   * A function is called by VirtualList for rendering each of the row
+   * @param rowIndex {Number} index of the log element in logs array
+   * @returns A react element with a row itself
+   */
+  getLogRow = (rowIndex: number) => {
+    const isSeparator = this.logs[rowIndex] === "---newlogs---"; // TODO: Use constant separator
```
Nokel81: Why is "---newlogs---" within the logs list itself? It seems very weird that if I search of "new" it would find this line. Is there a problem with keeping an index?

src/renderer/components/virtual-list/virtual-list.tsx
```diff
@@ -4,24 +4,26 @@ import "./virtual-list.scss";
...
 import { TableRowProps } from "../table/table-row";
 import { ItemObject } from "../../item.store";
 import throttle from "lodash/throttle";
 import debounce from "lodash/debounce";
 import isEqual from "lodash/isEqual";
 import ResizeSensor from "css-element-queries/src/ResizeSensor";
 
-interface Props {
-  items: ItemObject[];
+interface Props<T extends ItemObject = any> {
```
Nokel81, line 16: Does defaulting this to `any` gain us anything. I believe that this is a code smell.

src/common/__tests__/search-store.test.ts
```diff
@@ -0,0 +1,80 @@
...
+  beforeEach(async () => {
+    searchStore = new SearchStore();
+  })
+
+  it("does nothing with empty search query", () => {
+    searchStore.onSearch([], "");
+    expect(searchStore.occurrences).toEqual([]);
+  })
+
+  it("doesn't brake if no text provided", () => {
```
jim-docker: break?

jim-docker (commented):
doesn't scroll horizontally to make search text visible:

searchbug

src/renderer/components/dock/pod-logs.tsx
aleksfront: Removing separator stuff from this PR

src/renderer/components/virtual-list/virtual-list.tsx
aleksfront: It does, since `ItemObject` contains 2 specific methods `getId()` & `getName()` which is related to workspaces. But from now on we're using virtual list for any kind of lists, so this line helps to avoid typescript errors.

src/common/__tests__/search-store.test.ts
aleksfront: Of course.

aleksfront:
Added feature: horizontal scroll to founded keyword
horizontal scroll

nevalla (approved):
LGTM

jakolehm:
@aleksfront @nevalla integration tests are still failing.

nevalla:
> @aleksfront @nevalla integration tests are still failing.

Yes, we need to fix extensions loading in `master` to get those pass.

jim-docker (approved):

nevalla:
@aleksfront Please try to rebase with `master` to see if integration tests will pass now since #1233 is merged.

aleksfront:
@nevalla It's green now! 🚀

mih-kopylov:
Awesome, thanks, guys!

When it's gonna be released?

PR #1097, opened 2020-10-18
Force revision type to number for sort

outsideris:
Fix #1020 

I didn't fully understand Lens's internal. 
When I checked the [HelmRelease data here](https://github.com/lensapp/lens/blob/master/src/renderer/api/endpoints/helm-releases.api.ts#L74), `revision` field is already `string`. So, `getRevision` return `string` typed revision. 

I'm not sure why [`revision: number`](https://github.com/lensapp/lens/blob/master/src/renderer/api/endpoints/helm-releases.api.ts#L144) doesn't force to number type. Because I'm not TS expert.

So, I set `revision` type to string and then `parseInt` in `getRevision()`. 
It works now.

스크린샷 2020-10-18 오후 9 13 05
스크린샷 2020-10-18 오후 9 13 10

PR #1124, opened 2020-10-23
[BugFix] Releases sort by revision incorrect

pashevskii:


pashevskii:
Incorrect sorting is related to HelmRelease constructor. Object.assign has some specific behavior and ignores datatypes of target objects fields, a more detailed explanation is [here ](https://stackoverflow.com/questions/39906054/typescript-object-assign-confusion)

aleksfront:
Duplicate #1097

PR #1143, opened 2020-10-26
Allowing any item for StatusBarRegistration

### LINKED ISSUE #1127: StatusBarRegistration does not follow common structure
**Describe the bug**

Other extension registrations have `components` container, StatusBarRegistration does not. Maybe we need to introduce an interface for these?

### LINKED ISSUE #1129: StatusBarRegistration should allow also text (in addition to Icon)
Extension StatusBarRegistration should allow to show also text component (in addition to Icon). Basically extension should be able to show Icon, Text (or both). See VSCode for inspiration.

aleksfront:
* Replacing `icon` to `item` namings
* Adding Support text next to icon
* Changing icon to question mark
* Hover effect for status bar items (using general stylies since we're not activated inner extension styles yet, plus workspace selector could have similar styling)

Oct-26-2020 15-49-16

Fixes #1127 
Fixes #1129 

### DIFF
--- a/extensions/support-page/renderer.tsx
+++ b/extensions/support-page/renderer.tsx
@@ -21,15 +21,17 @@ export default class SupportPageRendererExtension extends LensRendererExtension 
     )
   }
 
-  registerStatusBarIcon(registry: Registry.StatusBarRegistry) {
+  registerStatusBarItem(registry: Registry.StatusBarRegistry) {
     this.disposers.push(
       registry.add({
-        icon: (
-          <Component.Icon
-            material="support"
-            tooltip="Support"
+        item: (
+          <div
+            className="flex align-center gaps hover-highlight"
             onClick={() => Navigation.navigate(supportPageURL())}
-          />
+          >
+            <Component.Icon material="help_outline" small />
+            <span>Support</span>
+          </div>
         )
       })
     )
--- a/src/extensions/lens-main-extension.ts
+++ b/src/extensions/lens-main-extension.ts
@@ -7,7 +7,7 @@ export class LensMainExtension extends LensExtension {
     //
   }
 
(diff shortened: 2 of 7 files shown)


src/extensions/lens-main-extension.ts
```diff
@@ -7,7 +7,7 @@ export class LensMainExtension extends LensExtension {
     //
   }
 
-  registerStatusBarIcon(registry: StatusBarRegistry) {
+  registerStatusBarItem(registry: StatusBarRegistry) {
```
ixrock: This was added by accident, there is no "status bar" in main process :) So plz remove.

extensions/support-page/renderer.tsx
```diff
@@ -21,15 +21,17 @@ export default class SupportPageRendererExtension extends LensRendererExtension
...
-          <Component.Icon
-            material="support"
-            tooltip="Support"
+        item: (
+          <div
+            className="flex align-center gaps hover-highlight"
             onClick={() => Navigation.navigate(supportPageURL())}
-          />
+          >
+            <Component.Icon material="help_outline" small />
```
ixrock, line 32: Why changed to `help_outline`?

ixrock (approved):

src/extensions/lens-main-extension.ts
jakolehm: Yes, this should be removed.

jakolehm (changes requested):

extensions/support-page/renderer.tsx
aleksfront: Just thinking its a better fit since there might be some quesitions without situations that require some kind of a rescue 😄  But yes, its a subjective thing. It also has a less outline than previous one.

PR #1147, opened 2020-10-26
Load extensions also from local folder

nevalla:

### DIFF
--- a/src/extensions/extension-manager.ts
+++ b/src/extensions/extension-manager.ts
@@ -79,36 +79,57 @@ export class ExtensionManager {
   }
 
   async loadExtensions() {
-    const extensions = await this.loadFromFolder([this.inTreeFolderPath, this.localFolderPath])
+    const bundledExtensions = await this.loadBundledExtensions()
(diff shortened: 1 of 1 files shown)


src/extensions/extension-manager.ts
```diff
@@ -74,31 +79,36 @@ export class ExtensionManager {
   }
 
   async loadExtensions() {
-    const extensions = await this.loadFromFolder(this.folderPath);
+    const extensions = await this.loadFromFolder([this.inTreeFolderPath, this.localFolderPath])
```
jakolehm: We need to separate in-tree & local-folder loading because they behave a bit differently (local-folder extensions are enabled by default and they can be enabled/disabled when we have UI available).

jakolehm (changes requested):

src/extensions/extension-manager.ts
nevalla: Changed. PTAL.

jakolehm (approved):
LGTM

PR #1149, opened 2020-10-26
refactor registering extensions

jim-docker:
extension implementors do not need to be aware of disposer and registry internals

This is a very simple modification where the existing functionality remains the same but the important thing is it's less complicated for the extension implementor.


jakolehm:
Not sure why this was closed 🤔

PR #1151, opened 2020-10-26
got extensions to load when checking multiple folders

jim-docker:

I noticed even the extensions from ./extensions folder were not loading. This seems to work... You might want to make it more async again


nevalla (approved):
Thanks!

PR #1155, opened 2020-10-27
Update chart details when selecting another chart

### LINKED ISSUE #1156: Chart detailed information was not updated
**To Reproduce**
1. Go to Apps - Charts
2. Select any chart
**** Information about the chart is correct ****
3. Select another chart 

**Expected behavior**
Information about the chart was updated

**Current behavior**
Information about the chart was not updated, only repo/name was changed (still shows chart details from 2 point)

pashevskii:
Fixed #1156

### DIFF
--- a/src/renderer/components/+apps-helm-charts/helm-chart-details.tsx
+++ b/src/renderer/components/+apps-helm-charts/helm-chart-details.tsx
@@ -3,7 +3,7 @@ import "./helm-chart-details.scss";
 import React, { Component } from "react";
 import { HelmChart, helmChartsApi } from "../../api/endpoints/helm-charts.api";
 import { t, Trans } from "@lingui/macro";
-import { observable, toJS } from "mobx";
+import { observable, autorun } from "mobx";
 import { observer } from "mobx-react";
 import { Drawer, DrawerItem } from "../drawer";
 import { autobind, stopPropagation } from "../../utils";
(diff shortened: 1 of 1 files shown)


pashevskii:
Related Issue https://github.com/lensapp/lens/issues/1156

src/renderer/components/+apps-helm-charts/helm-chart-details.tsx
```diff
@@ -47,6 +47,21 @@ export class HelmChartDetails extends Component<Props> {
...
+        this.readme = result.readme
+        this.chartVersions = result.versions
+        this.selectedChart = result.versions[0]
+      },
+      error => {
+        this.error = error;
+      })
+    }
+  }
+
```
aleksfront: Can you please revactor your changes a bit in a way that get request won't be duplicated in `componentDidMount` and `componentDidUpdate`?

You could use mobx `autorun` or `reaction` defined inside `componentDidMount` so it will trigger the fetch for both described methods (when `props.chart` will be changed).

You can reference usage of reaction inside `src/renderer/components/+user-management-roles-bindings/role-binding-details.tsx` for example.

aleksfront (changes requested):

src/renderer/components/+apps-helm-charts/helm-chart-details.tsx
pashevskii: Hi, thanks for your comment! 
Changed to autorun because "reaction" isn't triggered on component init

aleksfront: @pashevskii Did you push your changes with autorun? I still see an old code in PR.

pashevskii: Yeah,  forget to add in the commit before pushing, fixed.

aleksfront (approved):
LGTM!

PR #1157, opened 2020-10-27
Add extensions/api.d.ts to .eslintignore

nevalla:

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+src/extensions/npm/extensions/api.d.ts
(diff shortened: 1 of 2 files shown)


ixrock (approved):
Can we use existing `.eslintrc.js` for that purpose?

nevalla:
> Can we use existing `.eslintrc.js` for that purpose?

Yes, `ignorePatterns` seems to do the same (https://eslint.org/docs/user-guide/configuring#ignoring-files-and-directories). Will update the PR using that.

PR #1163, opened 2020-10-28
Fix bundled extension install on Linux/AppImage

### LINKED ISSUE #1162: Lens fails to install bundled extensions on Linux/AppImage
Because "npm install" does not support read-only filesystems.

```
npm ERR! code EROFS
npm ERR! syscall access
npm ERR! path /tmp/.mount_Lens-4n78npS/resources/extensions/metrics-cluster-feature
npm ERR! errno -30
npm ERR! rofs EROFS: read-only file system, access '/tmp/.mount_Lens-4n78npS/resources/extensions/metrics-cluster-feature'
npm ERR! rofs Often virtualized file systems, or other file systems
npm ERR! rofs that don't support symlinks, give this error.
```

jakolehm:
Fixes #1162

### DIFF
--- a/src/extensions/extension-manager.ts
+++ b/src/extensions/extension-manager.ts
@@ -25,8 +25,16 @@ export class ExtensionManager {
     return extensionPackagesRoot()
   }
 
+  get inTreeTargetPath() {
+    return path.join(this.extensionPackagesRoot, "extensions")
+  }
+
   get inTreeFolderPath(): string {
     return path.resolve(__static, "../extensions");
+  }
+
+  get nodeModulesPath(): string {
+    return path.join(this.extensionPackagesRoot, "node_modules")
   }
 
   get localFolderPath(): string {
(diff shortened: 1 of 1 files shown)


nevalla (approved):
LGTM

src/extensions/extension-manager.ts
```diff
@@ -39,7 +47,13 @@ export class ExtensionManager {
 
   async load() {
     logger.info("[EXTENSION-MANAGER] loading extensions from " + this.extensionPackagesRoot)
-    await fs.ensureDir(path.join(this.extensionPackagesRoot, "node_modules"))
+    if (this.inTreeFolderPath !== this.inTreeTargetPath) {
```
nevalla, line 50: when this condition is false? (ie `this.inTreeFolderPath === this.inTreeTargetPath`)

jakolehm: I'm not actually sure (I think it depends on an environment). This is just to safeguard our in-tree extensions (if these match we would remove sources).

jakolehm: Other option is to check if `this.inTreeFolderPath` is read-only.

PR #1167, opened 2020-10-28
Fix npm package missing types

jakolehm:


aleksfront (approved):

ixrock:
What it does? I guess the real issue was that file `api.d.ts` just not included while publishing to npm?

PR #1170, opened 2020-10-28
Fix: path checks in custom extension loader

aleksfront:
Accepting only directories and checking for `package.json` availability while loading an extension from home folder.

It allows to avoid parsing `.DS_Store` file which can be created automatically in any folder by macOS. And skips directories without manifest `package.json`.

### DIFF
--- a/src/extensions/extension-manager.ts
+++ b/src/extensions/extension-manager.ts
@@ -61,6 +61,7 @@ export class ExtensionManager {
   async getExtensionByManifest(manifestPath: string): Promise<InstalledExtension> {
     let manifestJson: ExtensionManifest;
     try {
+      fs.accessSync(manifestPath, fs.constants.F_OK); // check manifest file for existence
       manifestJson = __non_webpack_require__(manifestPath)
       this.packagesJson.dependencies[manifestJson.name] = path.dirname(manifestPath)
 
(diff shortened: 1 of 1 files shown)


src/extensions/extension-manager.ts
```diff
@@ -131,6 +131,9 @@ export class ExtensionManager {
         continue
       }
       const absPath = path.resolve(folderPath, fileName);
+      if (!fs.existsSync(absPath) || !fs.lstatSync(absPath).isDirectory()) {
+        continue;
+      }
       const manifestPath = path.resolve(absPath, "package.json");
       await fs.access(manifestPath, fs.constants.F_OK)
```
jim-docker: I think this is where @stevejr encountered an issue. We're assuming any directory here will contain an extension. For Steve there's a directory (.git) that doesn't have package.json in it. Should we catch here, log a warning, and continue?

jim-docker: Also, there should probably be a similar change for lines 111-113

aleksfront: Thanks for pointing to this problem. Resolved by moving checking for `package.json` into `getExtensionByManifest()`

nevalla (approved):
LGTM

PR #1171, opened 2020-10-29
Optimize ExtensionManager#load

jakolehm:
- removes duplicate call of `this.installPackages()`
- only copies in-tree extensions if folder is read-only (npm install symlinks won't work on read-only fs)
- optimizes `npm install` calls


nevalla (approved):
LGTM

PR #1172, opened 2020-10-29
Fix regression caused by #1171

jakolehm:
- point package.json dependencies to correct path
- don't use package-lock.json, it causes problems

### DIFF
--- a/src/extensions/extension-manager.ts
+++ b/src/extensions/extension-manager.ts
@@ -17,6 +17,8 @@ type PackageJson = {
 
 export class ExtensionManager {
 
+  protected bundledFolderPath: string
+
   protected packagesJson: PackageJson = {
     dependencies: {}
   }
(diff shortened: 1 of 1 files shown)


src/extensions/extension-manager.ts
```diff
@@ -45,15 +47,24 @@ export class ExtensionManager {
...
   }
 
+  get packageJsonPath() {
+    return path.join(this.extensionPackagesRoot, "package.json")
+  }
+
   async load() {
     logger.info("[EXTENSION-MANAGER] loading extensions from " + this.extensionPackagesRoot)
+    if (fs.existsSync(path.join(this.extensionPackagesRoot, "package-lock.json"))) {
+      await fs.remove(path.join(this.extensionPackagesRoot, "package-lock.json"))
```
ixrock, line 57: Why do we need this with `--no-package-lock`?

ixrock (approved):
LGTM

src/extensions/extension-manager.ts
jakolehm: Current version has generated that file and it needs to be removed.. We can probably remove this cleanup at some point.

PR #1173, opened 2020-10-29
Using external styles for extensions

### LINKED ISSUE #1146: Allow to use external styles in extensions

ixrock:
Since we use `webpack` it feels natural to bundle all possible assets within it's executable `dist/renderer.js`
This will simplify process of loading styles and it's better for performance in general.

close #1146


jim-docker (approved):
LGTM

PR #1174, opened 2020-10-29
License extension 

### LINKED ISSUE #1141: License menu item should come from an extension
Reason: visible license/eula only applies to built binary distribution.

ixrock:
close #1141


extensions/license/main.ts
```diff
@@ -0,0 +1,14 @@
+import { LensMainExtension } from "@k8slens/extensions";
+import { shell } from "electron";
```
ixrock: This needs to double check. 
It's compiled to `require("electron")` in `main.js` but how it's eventually consumed inside Lens runtime with such call I'm not sure (works well in `development`-mode though).
Currently after build with `yarn build:mac` and running `Lens.app` extensions declared in `package.json ` not loaded at all..

ixrock: Seems like this is broken in `master` as well.

**Steps to reproduce**:
1. `make build-extensions`
1. `yarn build:mac`
1. Run `dist/mac/Lens.app`

image

jim-docker: > This needs to double check.
> It's compiled to `require("electron")` in `main.js` but how it's eventually consumed inside Lens runtime with such call I'm not sure (works well in `development`-mode though).
> Currently after build with `yarn build:mac` and running `Lens.app` extensions declared in `package.json ` not loaded at all..

Could wrap and expose `shell. openExternal()` through the extensions api?

jim-docker: > Seems like this is broken in `master` as well.
> 
> **Steps to reproduce**:
> 
> 1. `make build-extensions`
> 2. `yarn build:mac`
> 3. Run `dist/mac/Lens.app`
> 
seems to be fixed...

jim-docker (approved):

extensions/license/main.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export default class LicenseLensMainExtension extends LensMainExtension {
+  appMenus = [
+    {
+      parentId: "help",
+      label: "License",
+      async click() {
+        shell.openExternal("https://k8slens.dev/licenses/eula.md");
```
jakolehm: Would be safer if we expose `openExternal` (or whole `shell`) via extension-api (maybe via `Util` namespace?).

jakolehm (changes requested):

extensions/license/main.ts
ixrock: Don't know what I'm doing wrong or is it some cached files.. but this is still happening on my machine (even after `master` merge)

ixrock:
@jakolehm PTAL, added helper `Util.openExternal` to extensions-api

jakolehm (changes requested):
We need to keep LICENSE file in the root folder.

ixrock:
@jakolehm could be merged already?

jakolehm (approved):

PR #1175, opened 2020-10-29
Restart deployment

### LINKED ISSUE #603: Restart deployment
A way to restart all pods of a deployment.
This can be done by add/replace an annotation lensapp.kubernetes.io/restartedAt to spec.template.metadata.annotations like kubectl rollout restart. It can be used to restart the rollout of a Deployment, StatefulSet or DaemonSet.

Idea from here: https://github.com/kubenav/kubenav/issues/80

**Why is this needed**:
To restart all pods of a deployment

- Kubernetes distribution: [AKS]
- Desktop OS: Win10

pashevskii:
Fixed #603

### DIFF
--- a/src/renderer/api/endpoints/deployment.api.ts
+++ b/src/renderer/api/endpoints/deployment.api.ts
@@ -1,3 +1,5 @@
+import moment from "moment";
+
 import { IAffinity, WorkloadKubeObject } from "../workload-kube-object";
 import { autobind } from "../../utils";
 import { KubeApi } from "../kube-api";
@@ -23,6 +25,25 @@ export class DeploymentApi extends KubeApi<Deployment> {
       }
     })
   }
+
+  restart(params: { namespace: string; name: string }) {
+    return this.request.patch(this.getUrl(params), {
+      data: {
+        spec: {
+          template: {
+            metadata: {
+              annotations: {"kubectl.kubernetes.io/restartedAt" : moment.utc().format()}
+            }
+          }
+        }
+      }
+    },
+    {
+      headers: {
+        'content-type': 'application/strategic-merge-patch+json'
+      }
+    })
+  }
 }
 
 @autobind()
@@ -38,6 +59,7 @@ export class Deployment extends WorkloadKubeObject {
       metadata: {
         creationTimestamp?: string;
         labels: { [app: string]: string };
+        annotations?: { [app: string]: string };
       };
       spec: {
         containers: {
--- a/src/renderer/api/json-api.ts
+++ b/src/renderer/api/json-api.ts
@@ -64,7 +64,7 @@ export class JsonApi<D = JsonApiData, P extends JsonApiParams = JsonApiParams> {
   }
 
   patch<T = D>(path: string, params?: P, reqInit: RequestInit = {}) {
-    return this.request<T>(path, params, { ...reqInit, method: "patch" });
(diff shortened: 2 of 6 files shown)


src/renderer/api/json-api.ts
```diff
@@ -64,7 +64,7 @@ export class JsonApi<D = JsonApiData, P extends JsonApiParams = JsonApiParams> {
   }
 
   patch<T = D>(path: string, params?: P, reqInit: RequestInit = {}) {
-    return this.request<T>(path, params, { ...reqInit, method: "patch" });
+    return this.request<T>(path, params, { ...reqInit, method: "PATCH" });
```
pashevskii, line 67: JFYI: https://stackoverflow.com/questions/34666680/fetch-patch-request-is-not-allowed
Method PATCH is case sensitive

src/renderer/api/endpoints/deployment.api.ts
```diff
@@ -23,6 +25,25 @@ export class DeploymentApi extends KubeApi<Deployment> {
       }
     })
   }
+
+  restart(params: { namespace: string; name: string }) {
```
pashevskii, line 29: JFYI:
https://github.com/kubernetes/kubectl/blob/release-1.16/pkg/polymorphichelpers/objectrestarter.go#L32 
The request represents this kubectl method

src/renderer/api/endpoints/deployment.api.ts
```diff
@@ -23,6 +25,25 @@ export class DeploymentApi extends KubeApi<Deployment> {
...
+    return this.request.patch(this.getUrl(params), {
+      data: {
+        spec: {
+          template: {
+            metadata: {
+              annotations: {"kubectl.kubrnetes.io/restartedAt" : moment.utc().format()}
```
aleksfront: Typo in annotation? `kubectl.kubrnetes.io/restartedAt` -> `kubectl.kubernetes.io/restartedAt`

src/renderer/components/+workloads-deployments/deployments.tsx
```diff
@@ -86,14 +89,14 @@ export class Deployments extends React.Component<Props> {
...
-          return <DeploymentMenu object={item}/>
+          return <DeploymentRestartMenu object={item}/>
         }}
       />
     )
   }
 }
 
-export function DeploymentMenu(props: KubeObjectMenuProps<Deployment>) {
+export function DeploymentScaleMenu(props: KubeObjectMenuProps<Deployment>) {
```
aleksfront: Its still should be called `DeploymentMenu` since it must include different kind of menu items. So there must be only 1 menu function with restart and scale options in it. 

Can you please leave `DeploymentMenu` but expand it and include both restart and scale menu items inside of it? This way we will keep only 1 `kubeObjectMenuRegistry.add()` call and bring all items in one place.

src/renderer/api/endpoints/deployment.api.ts
pashevskii: oh, year, fixed

src/renderer/components/+workloads-deployments/deployments.tsx
pashevskii: Fixed!

aleksfront (approved):
LGTM

PR #1176, opened 2020-10-29
Improve how extensions can manage cluster/workspace stores

jakolehm:
- extension can claim workspace/cluster to be managed by them
- lens only shows workspaces/clusters that are explicitly set as activated (they are not activated automatically, except locally managed objects)
- workspace/cluster store api cleanup (unify few method calls)

### DIFF
--- a/src/common/cluster-store.ts
+++ b/src/common/cluster-store.ts
@@ -73,6 +73,10 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
     return filePath;
   }
 
+  @observable activeCluster: ClusterId;
+  @observable removedClusters = observable.map<ClusterId, Cluster>();
+  @observable clusters = observable.map<ClusterId, Cluster>();
+
   private constructor() {
     super({
       configName: "lens-cluster-store",
@@ -80,16 +84,17 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
       migrations: migrations,
     });
 
+    this.pushStateToViewsPeriodically()
+  }
+
+  protected pushStateToViewsPeriodically() {
     if (!ipcRenderer) {
+      // This is a bit of a hack, we need to do this because we might loose messages that are sent before a view is ready
       setInterval(() => {
         this.pushState()
       }, 5000)
     }
   }
-
-  @observable activeCluster: ClusterId;
-  @observable removedClusters = observable.map<ClusterId, Cluster>();
-  @observable clusters = observable.map<ClusterId, Cluster>();
 
   registerIpcListener() {
     logger.info(`[CLUSTER-STORE] start to listen (${webFrame.routingId})`)
--- a/src/common/cluster-store.ts
+++ b/src/common/cluster-store.ts
@@ -170,14 +170,13 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
   }
 
   @action
-  addCluster(model: ClusterModel): Cluster {
+  addCluster(model: ClusterModel | Cluster ): Cluster {
     appEventBus.emit({name: "cluster", action: "add"})
-    let cluster: Cluster = null
-    if (model instanceof Cluster === false) {
+    let cluster = model as Cluster;
+    if (!(model instanceof Cluster)) {
       cluster = new Cluster(model)
     }
     this.clusters.set(model.id, cluster);
-
     return cluster
   }
 
--- a/src/common/cluster-store.ts
+++ b/src/common/cluster-store.ts
@@ -87,7 +87,7 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
     }
   }
 
-  @observable activeClusterId: ClusterId;
+  @observable activeCluster: ClusterId;
   @observable removedClusters = observable.map<ClusterId, Cluster>();
   @observable clusters = observable.map<ClusterId, Cluster>();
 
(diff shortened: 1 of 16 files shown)


src/common/cluster-store.ts
```diff
@@ -78,6 +79,12 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
       accessPropertiesByDotNotation: false, // To make dots safe in cluster context names
       migrations: migrations,
     });
+
+    if (!ipcRenderer) {
```
ixrock, line 91: I would put this to own method with descriptive name what it does.

src/common/cluster-store.ts
```diff
@@ -145,12 +160,29 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
...
+  addCluster(model: ClusterModel): Cluster {
+    appEventBus.emit({name: "cluster", action: "add"})
+    let cluster: Cluster = null
+    if (model instanceof Cluster === false) {
+      cluster = new Cluster(model)
+    }
+    this.clusters.set(model.id, cluster);
+
+    return cluster
+  }
```
ixrock:
```suggestion
  addCluster(model: ClusterModel | Cluster): Cluster {
    appEventBus.emit({name: "cluster", action: "add"})
    let cluster = model as Cluster;
    if (!(model instanceof Cluster)) {
      cluster = new Cluster(model)
    }
    this.clusters.set(model.id, cluster);
    return cluster
  }
```

src/common/ipc.ts
```diff
@@ -61,6 +61,9 @@ export interface IpcBroadcastParams<A extends any[] = any> {
 }
 
 export function broadcastIpc({ channel, frameId, frameOnly, webContentId, filter, args = [] }: IpcBroadcastParams) {
+  if (!webContents) {
+    return
```
ixrock: Why this is needed? This function is intended to work only in `main` process and calls from renderer must not happen. Otherwise suggesting to apply this check at places of usage.

src/common/workspace-store.ts
```diff
@@ -23,15 +81,34 @@ export class WorkspaceStore extends BaseStore<WorkspaceStoreModel> {
...
+      setInterval(() => {
+        this.pushState()
+      }, 5000)
+    }
+  }
+
+  registerIpcListener() {
+    logger.info("[WORKSPACE-STORE] starting to listen state events")
+    ipcRenderer.on("workspace:state", (event, workspaceId: string, state: WorkspaceState) => {
+      console.log(workspaceId, state)
```
ixrock: Forgotten `console.log`?

src/common/workspace-store.ts
```diff
@@ -103,7 +195,11 @@ export class WorkspaceStore extends BaseStore<WorkspaceStoreModel> {
     }
     if (workspaces.length) {
       this.workspaces.clear();
-      workspaces.forEach(workspace => {
+      workspaces.forEach(ws => {
+        const workspace = new Workspace(ws)
+        if (!workspace.ownerRef) {
```
ixrock:
```suggestion
        if (!workspace.isManaged) {
```

src/renderer/components/+workspaces/workspaces.tsx
```diff
@@ -107,11 +111,12 @@ export class Workspaces extends React.Component {
           <Trans>Workspaces</Trans>
         </h2>
         <div className="items flex column gaps">
-          {this.workspaces.map(({ id: workspaceId, name, description }) => {
+          {this.workspaces.map(({ id: workspaceId, name, description, ownerRef }) => {
```
ixrock, line 114:
```suggestion
          {this.workspaces.map(({ id: workspaceId, name, description, isManaged }) => {
```

src/common/workspace-store.ts
```diff
@@ -1,19 +1,77 @@
...
+    Object.assign(this, data)
+
+    if (!ipcRenderer) {
+      reaction(() => this.getState(), () => {
+        this.pushState()
+      })
+    }
+  }
+
+  isManaged(): boolean {
```
ixrock:
```suggestion
  get isManaged(): boolean {
```

ixrock (approved):
LGTM

ixrock:
Found a bug, active-cluster icon for some reasons don't get updated, please check.
image

src/common/ipc.ts
jakolehm: I can take it away but function name does not really tell that this is only from main->renderer (given that this locates under common).

jakolehm (commented):
> Found a bug, active-cluster icon for some reasons don't get updated, please check.

Noticed the same.. not sure why this does not update (renderer does get update via ipc) 🤔

src/common/ipc.ts
ixrock: Maybe rename to `broadcastToViews` / `broadcastToWebContents` / `broadcastFromMain`? 
Or maybe `webContents = webContents || remote.webContents` if this make it work from renderer?

ixrock: Anyway, silently ignore a call (even unexpected) in renderer doesn't sound right IMHO..

jakolehm:
> Found a bug, active-cluster icon for some reasons don't get updated, please check.

Bug seems to be in `ClustersMenu`, it does not properly react to store changes. Will fix asap.

src/common/ipc.ts
jakolehm: I'll revert this change, out-of-scope for this PR.

jakolehm:
@nevalla @ixrock PTAL

nevalla (approved):
LGTM

PR #1180, opened 2020-10-30
Telemetry improvements

nevalla:


extensions/telemetry/main.ts
```diff
@@ -7,6 +7,7 @@ export default class TelemetryMainExtension extends LensMainExtension {
   async onActivate() {
     console.log("telemetry main extension activated")
     tracker.start()
+    tracker.reportPeriodically()
```
nevalla, line 10: previously reporting was done in both main and renderer processes, sending it from main processes only now.

src/main/cluster-detectors/nodes-count-detector.ts
```diff
@@ -5,13 +5,12 @@ export class NodesCountDetector extends BaseClusterDetector {
   key = ClusterMetadataKey.NODES_COUNT
 
   public async detect() {
+    if (!this.cluster.accessible) return null;
```
nevalla, line 8: prevent overriding data with null if cluster is not accessible

ixrock (approved):
LGTM

PR #1188, opened 2020-10-30
Implement KubeObjectStatusText extension API

### LINKED ISSUE #1118: Allow to register text to status column of resource list
Extensions should be able to render text (optionally with color) to status column of resource list.

nevalla:
Fixes #1118 

**Usage** 
```typescript
// renderer.ts
export default class PodStatusTextExtension extends LensRendererExtension {
 KubeObjectStatusTexts = [
    {
      kind: "Pod",
      apiVersions: ["v1"],
      resolve: (pod: K8sApi.Pod) => resolve(pod)
    }
  ]
}

// pod-status-resolver.tsx
import { K8sApi } from "@k8slens/extensions";

export function resolve(pod: K8sApi.Pod): K8sApi.KubeObjectStatus {
  if (pod.getName().startsWith("nginx-deployment") {
    return {
      level:  K8sApi.KubeObjectStatusLevel.INFO,
      text: "Image scanned",
    }
  }

  return null
}
```

image


jakolehm:
why not just:
```typescript
export interface ResourceStatusRegistration {
  kind: string;
  apiVersions: string[];
  resolver: (object: KubeObject) => ResourceStatus | null;
}
```

where `ResourceStatus` is an interface with text & color?

nevalla:
> why not just:
> 
> ```ts
> export interface ResourceStatusRegistration {
>   kind: string;
>   apiVersions: string[];
>   resolver: (object: KubeObject) => ResourceStatus | null;
> }
> ```
> 
> where `ResourceStatus` is an interface with text & color?

Much better...will change it.

nevalla:
Now all resource lists are using new `KubeResourceStatusIcon` component instead of `KubeEventIcon`. New component will display multiple status messages (not only from events) in a tooltip (see the image in the description).

src/extensions/registries/index.ts
```diff
@@ -7,3 +7,4 @@ export * from "./status-bar-registry"
 export * from "./kube-object-detail-registry";
 export * from "./kube-object-menu-registry";
 export * from "./cluster-feature-registry"
+export * from "./resource-status-registry"
```
jakolehm: This should be named similar to other `kube-object-*` registries, maybe `kube-object-status-registry`?

src/renderer/components/kube-resource-status-icon/kube-resource-status-icon.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { KubeObject } from "../../api/kube-object";
+import { cssNames, formatDuration } from "../../utils";
+import { ResourceStatusRegistration, ResourceStatus, ResourceStatusLevel, resourceStatusRegistry } from "../../../extensions/registries/resource-status-registry"
+import { computed } from "mobx";
+
+interface Props {
+  object: KubeObject;
+}
+
+export class KubeResourceStatusIcon extends React.Component<Props> {
```
jakolehm: `KubeObjectStatusIcon`

jakolehm (changes requested):

extensions/event-resource-status/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "version": "0.1.0",
+  "description": "Adds resource status from events",
+  "renderer": "dist/renderer.js",
+  "lens": {
+    "metadata": {},
+    "styles": []
+  },
+  "scripts": {
+    "build": "webpack --config webpack.config.js",
+    "dev": "npm run build --watch"
```
Nokel81: To comply with #1201 please add a `"test"` script.

extensions/event-resource-status/Makefile
```diff
@@ -0,0 +1,5 @@
+install-deps:
+	npm install
+
+build: install-deps
+	npm run build
```
Nokel81: To comply with #1201 please add a `"test"` target.

src/extensions/registries/index.ts
nevalla: Renamed

src/renderer/components/kube-resource-status-icon/kube-resource-status-icon.tsx
nevalla: Done

src/extensions/renderer-api/index.ts
```diff
@@ -4,9 +4,11 @@
 import * as Component from "./components"
 import * as K8sApi from "./k8s-api"
 import * as Navigation from "./navigation"
+import * as KubeObjectStatus from "./kube-object-status"
 
 export {
   Component,
   K8sApi,
   Navigation,
+  KubeObjectStatus
```
jakolehm: Do we really need separate namespace for this? 🤔

src/extensions/renderer-api/k8s-api.ts
```diff
@@ -2,6 +2,7 @@ export { isAllowedResource } from "../../common/rbac"
 export { apiManager } from "../../renderer/api/api-manager";
 export { KubeObjectStore } from "../../renderer/kube-object.store"
 export { KubeApi, forCluster, IKubeApiCluster } from "../../renderer/api/kube-api";
+export type { EventStore } from "../../renderer/components/+events/event.store"
```
jakolehm, line 5: I think we need a separate issue about missing store types?

src/extensions/renderer-api/index.ts
nevalla: I would rather not to add new namespace, but it does not fit very well in to any existing namespaces. But I'm open for suggestions.

extensions/event-resource-status/package.json
nevalla: Done

extensions/event-resource-status/Makefile
nevalla: Done

src/extensions/renderer-api/index.ts
jakolehm: Everything else related to "KubeObject" is under `K8sApi`, would that make sense here?

nevalla: Yes. Moved it under `K8sApi`.

src/extensions/renderer-api/kube-object-status.ts
```diff
@@ -0,0 +1 @@
+export { KubeObjectStatus as Status, KubeObjectStatusLevel as Level } from "../registries/kube-object-status-registry"
```
jakolehm: Does not look right? We should not use generic `Status`/`Level` if these are exposed within `K8sApi` namespace.

src/extensions/registries/index.ts
```diff
@@ -7,3 +7,4 @@ export * from "./status-bar-registry"
 export * from "./kube-object-detail-registry";
 export * from "./kube-object-menu-registry";
 export * from "./cluster-feature-registry"
+export * from "./kube-object-status-registry"
```
jakolehm, line 10: Isn't this exporting all `KubeObjectStatus` stuff? Is that intentional?

src/extensions/renderer-api/kube-object-status.ts
nevalla: Yes, this is still wrong here. Will fix

src/extensions/registries/index.ts
nevalla: I moved `KubeObjectStatus` types to `K8sApi`. Feels more natural to have them there.

src/extensions/renderer-api/kube-object-status.ts
nevalla: Done

extensions/kube-object-event-status/renderer.tsx
```diff
@@ -0,0 +1,42 @@
...
+  kubeObjectStatusTexts = [
+    {
+      kind: "Pod",
+      apiVersions: ["v1"],
+      resolve: (object: K8sApi.Pod) => resolveStatusForPods(object)
+    },
+    {
+      kind: "ReplicaSet",
+      apiVersions: ["v1"],
+      resolve: (object: K8sApi.KubeObject) => resolveStatus(object)
```
jakolehm: Why some of these use specific types and other use generic `KubeObject`?

src/extensions/renderer-api/k8s-api.ts
```diff
@@ -29,3 +30,15 @@ export { Role, roleApi } from "../../renderer/api/endpoints";
 export { RoleBinding, roleBindingApi } from "../../renderer/api/endpoints";
 export { ClusterRole, clusterRoleApi } from "../../renderer/api/endpoints";
 export { ClusterRoleBinding, clusterRoleBindingApi } from "../../renderer/api/endpoints";
+
+export type KubeObjectStatus = {
```
jakolehm: Does not feel right to have them here, everything else is exported from somewhere outside.

extensions/kube-object-event-status/renderer.tsx
nevalla: `resolveStatusFromPods` and `resolveStatusFromCronJobs` requires specific type. I've changed this now to use always specific type to be more consistent.

src/extensions/renderer-api/k8s-api.ts
nevalla: Moved these types to separate file where those are exported from now

nevalla:
Tests are now again green. @jakolehm PTAL

jakolehm (approved):
LGTM

jakolehm:
Just realised this PR was missing docs.

PR #1192, opened 2020-11-02
Fix UI staleness after network issues

### LINKED ISSUE #1110: Stale resources
**Describe the bug**
Since a couple of releases ago (my guess would be since 3.6.0, but I didn't really track this), Lens frequently shows outdated resource states.

Just today I wanted to check some pods to see if they were updated. Lens was already running and connected to the cluster, but in a different view. I changed over to the Pods view and filtered by namespace to see the pods in question.

In the overview table, they showed up as running 54w. Clicking on each of the pods gave me the same info, as well as a running image version of 10.x.x. The reality was however that the pods were recreated 11h ago and ran with a new image version 11.x.x. When I realized that through kubectl and switched back to Lens, the state in the still open Pods view had adjusted to the correct values.

I've seen similar issues with all kinds of resources - services, pvcs, various crds (such as traefik ingressroutes, elasticsearch, ...).

- Lens Version: 3.6.6
- OS: Windows 10
- Installation method (e.g. snap or AppImage in Linux): Chocolatey

jakolehm:
- adds network online/offline events
- connected cluster is automatically set to offline when network goes offline.. Lens still tries to check if cluster is accessible because minikube and other local clusters are connectable even if network is offline
- forces recheck when network comes back
- fixes a bug in watch api where error during STREAM_END did stop retries

Might fix #1110

### DIFF
--- a/src/main/cluster-manager.ts
+++ b/src/main/cluster-manager.ts
@@ -1,5 +1,6 @@
 import "../common/cluster-ipc";
 import type http from "http"
+import { ipcMain } from "electron"
 import { autorun } from "mobx";
 import { clusterStore, getClusterIdFromHost } from "../common/cluster-store"
 import { Cluster } from "./cluster"
@@ -30,6 +31,29 @@ export class ClusterManager {
     }, {
       delay: 250
     });
+
+    ipcMain.on("network:offline", () => { this.onNetworkOffline() })
+    ipcMain.on("network:online", () => { this.onNetworkOnline() })
+  }
+
+  protected onNetworkOffline() {
+    logger.info("[CLUSTER-MANAGER]: network is offline")
+    clusterStore.enabledClustersList.forEach((cluster) => {
+      if (!cluster.disconnected) {
+        cluster.online = false
+        cluster.accessible = false
+        cluster.refreshConnectionStatus().catch((e) => e)
+      }
+    })
+  }
+
+  protected onNetworkOnline() {
+    logger.info("[CLUSTER-MANAGER]: network is online")
+    clusterStore.enabledClustersList.forEach((cluster) => {
+      if (!cluster.disconnected) {
+        cluster.refreshConnectionStatus().catch((e) => e)
+      }
+    })
   }
 
   stop() {
--- a/src/main/cluster.ts
+++ b/src/main/cluster.ts
@@ -67,12 +67,12 @@ export class Cluster implements ClusterModel, ClusterState {
   @observable kubeConfigPath: string;
   @observable apiUrl: string; // cluster server url
   @observable kubeProxyUrl: string; // lens-proxy to kube-api url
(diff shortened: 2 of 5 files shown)


src/main/cluster-manager.ts
```diff
@@ -30,6 +31,29 @@ export class ClusterManager {
     }, {
       delay: 250
     });
+
+    ipcMain.on("network:offline", () => { this.onNetworkOffline() })
+    ipcMain.on("network:online", () => { this.onNetworkOnline() })
```
ixrock, line 36:
```suggestion
    ipcMain.on("network:offline", () => this.onNetworkOffline())
    ipcMain.on("network:online", () => this.onNetworkOnline())
```

src/main/cluster-manager.ts
```diff
@@ -30,6 +31,29 @@ export class ClusterManager {
...
+  }
+
+  protected onNetworkOffline() {
+    logger.info("[CLUSTER-MANAGER]: network is offline")
+    clusterStore.enabledClustersList.forEach((cluster) => {
+      if (!cluster.disconnected) {
+        cluster.online = false
+        cluster.accessible = false
+        cluster.refreshConnectionStatus().catch((e) => e)
```
ixrock, line 45: So, this still marks local clusters (e.g. minikube) with `online == true`, right?

src/main/cluster.ts
```diff
@@ -127,19 +127,20 @@ export class Cluster implements ClusterModel, ClusterState {
...
-    if (ipcMain) {
-      this.eventDisposers.push(
-        reaction(() => this.getState(), () => this.pushState()),
-        () => {
-          clearInterval(refreshTimer);
-          clearInterval(refreshMetadataTimer);
-        },
-      );
+    if (!ipcMain) {
+      return
```
ixrock: Why this is needed? `cluster.activate() -> cluster.bindEvents()` called only at `main` AFAIK?

ixrock (approved):
Looking good!

src/main/cluster.ts
jakolehm: We cannot be sure where this is called if/when `Cluster` is exposed to extensions.

src/main/cluster-manager.ts
jakolehm: Yes, that's right.

src/main/cluster.ts
ixrock: Well, maybe then it should be inside `cluster.activate()`?

ixrock: And `bindEvents()` actually is protected method which is not intended to call outside..

jakolehm: I'll revert this change (not needed for this PR).. I think we have to refactor `Cluster` class quite a bit to make it safe/usable in both main & renderer contexts.

PR #1208, opened 2020-11-03
UI for enabling/disabling extensions

### LINKED ISSUE #1134: UI for enabling/disabling extensions
Lens needs an UI where user can enable/disable extensions. Note: extensions mentioned in Lens app package.json are always enabled.

UI should be accessible via "Lens -> Extensions" menu item (it should be after "Preferences").

ixrock:
_Changes:_

- enable/disable user extensions + extensions-page available from top menu (or `Cmd+Shift+E`)
- fixed bug when on initial app opening an active cluster don't apply all extensions (dev-only bug, but still annoying)
- fixed `tooltip.tsx` positioning bug
- now every extension has own store instance extended from `ExtensionStore` for keeping state
(could be also overridden inside extension by `extension.init(store: ExtensionStore)`

close #1134

image


src/extensions/core-api/cluster-feature.ts
```diff
@@ -1 +1,2 @@
-export { ClusterFeature as Feature, ClusterFeatureStatus as FeatureStatus } from "../cluster-feature"
+export { ClusterFeature as Feature } from "../cluster-feature"
+export type { ClusterFeatureStatus as FeatureStatus } from "../cluster-feature"
```
ixrock, line 2: Removed warnings in `webpack:main` compilation.

src/renderer/components/tooltip/tooltip.tsx
```diff
@@ -167,7 +167,6 @@ export class Tooltip extends React.Component<TooltipProps> {
       top = topCenter;
       break;
     case "top_right":
-    default:
```
ixrock, line 170: Forgotten debugging.

src/extensions/extension-loader.ts
```diff
@@ -1,90 +1,121 @@
...
 
 // lazy load so that we get correct userData
 export function extensionPackagesRoot() {
   return path.join((app || remote.app).getPath("userData"))
 }
 
-export class ExtensionLoader {
-  @observable extensions = observable.map<ExtensionId, InstalledExtension>([], { deep: false });
-  @observable instances = observable.map<ExtensionId, LensExtension>([], { deep: false })
+export class ExtensionLoader extends BaseStore<ExtensionLoaderStoreModel> {
```
aleksfront: Maybe reconsider naming since `ExtensionLoader` now uses `BaseStore`. Probably `ExtensionStore`? I know that we have (`src/extensions/extension-store.ts`) also.

aleksfront (approved):

src/extensions/extension-loader.ts
```diff
@@ -1,90 +1,121 @@
...
-export interface InstalledExtension extends ExtensionModel {
+export interface ExtensionLoaderStoreModel {
+  extensions: LensExtensionStoreModel[]
+}
+
+export interface InstalledExtension {
+  manifest: LensExtensionManifest;
   manifestPath: string;
-  manifest: ExtensionManifest;
+  isBundled?: boolean; // defined in package.json
```
jakolehm: Probably should be readonly? (same applies to manifest/manifestPath)

src/extensions/extension-loader.ts
jakolehm: IMHO we should have separate `ExtensionStore` from `ExtensionLoader`. Now `ExtensionLoader` has multiple responsibilities which makes it complex.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { _i18n } from "../../i18n";
+import { Button } from "../button";
+import { WizardLayout } from "../layout/wizard-layout";
+import { Input } from "../input";
+import { Icon } from "../icon";
+import { extensionLoader } from "../../../extensions/extension-loader";
+import { extensionManager } from "../../../extensions/extension-manager";
+
+@observer
+export class Extensions extends React.Component {
```
jakolehm, line 17: Shouldn't this be a "global" page (like Preferences)?

ixrock: What do you mean? It's kinda "global" page, see `cluster-manager.tsx`:
```
            <Route component={Preferences} {...preferencesRoute} />
            <Route component={Extensions} {...extensionsRoute} />
```

aleksfront: It can be, just thinking of a reason to hide clusters and status bar

jakolehm: Why don't we hide clusters/status bar here if we do that for Preferences?

ixrock: fixed

Nokel81:
This PR has broken the example extension @ixrock

ixrock:
@Nokel81 yes, I know :) You got fixed it in your PR, not critical.

PR #1217, opened 2020-11-03
add @types/node to devDeps for npm extension

### LINKED ISSUE #1169: Extension npm package should have @types/node as dev dependency

Nokel81:

fixes #1169


src/extensions/npm/extensions/yarn.lock
```diff
@@ -0,0 +1,8 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
jakolehm: `yarn.lock` is unnecessary, maybe gitignore it?

jakolehm (changes requested):
We need to remove `@types/node` from in-tree extensions to see if this actually works.

jakolehm (approved):

PR #1225, opened 2020-11-04
Refactor ipc module

### LINKED ISSUE #1222: Extension instances are not synced properly to cluster iframes
**Describe the bug**

User needs to hit refresh before extension is visible in a cluster menu if cluster is automatically opened on app start.

jakolehm:
- `broadcastMessage` : broadcasts a message to listeners in any/all contexts (main, renderer, iframe)
- `subscribeToBroadcast`: subscribes a listener to a channel
- `requestMain`: sends a message to main process and waits a response (async)
- `handleRequest`: handles requests done via `requestMain`

Fixes #1222


Nokel81:
Is this PR still in draft form?

jakolehm:
> Is this PR still in draft form?

Not anymore.

jakolehm:
I think we should expose these refactored functions to extension-api.

src/common/cluster-ipc.ts
```diff
@@ -1,59 +1,43 @@
...
-    channel: "cluster:activate",
-    handle: (clusterId: ClusterId, force = false) => {
-      const cluster = clusterStore.getById(clusterId);
-      if (cluster) {
-        return cluster.activate(force);
-      }
-    },
-  }),
+if (ipcMain) {
+  handleRequest("cluster:activate", (event, clusterId: ClusterId, force = false) => {
```
ixrock: I'd like to avoid copy-pasting strings like `cluster:activate` over whole project.. (easier to navigate in project / less bugs-prone imho). Is it possible to wrap / encapsulate event-name like it was before?

src/common/base-store.ts
```diff
@@ -36,8 +36,12 @@ export class BaseStore<T = any> extends Singleton {
     return path.basename(this.storeConfig.path);
   }
 
-  get syncChannel() {
-    return `store-sync:${this.name}`
+  get syncRendererChannel() {
```
ixrock: Why we need this separation in event-sync for main/renderer?

src/common/ipc.ts
```diff
@@ -1,79 +1,70 @@
...
+}
 
-export interface IpcChannelOptions {
-  channel: IpcChannel; // main <-> renderer communication channel name
-  handle?: (...args: any[]) => Promise<any> | any; // message handler
-  autoBind?: boolean; // auto-bind message handler in main-process, default: true
-  timeout?: number; // timeout for waiting response from the sender
-  once?: boolean; // one-time event
+export async function requestMain(channel: string, ...args: any[]) {
+  return await ipcRenderer.invoke(channel, ...args)
```
ixrock: Probably no need to `await`

src/renderer/components/layout/sidebar.tsx
```diff
@@ -72,9 +72,15 @@ export class Sidebar extends React.Component<Props> {
     });
   }
 
+  @computed
```
ixrock: There is nothing to `compute` :) It's better to use when deriving new state from `observable`-s.

src/common/base-store.ts
jakolehm: `broadcastMessage` would send same message in a loop if we use same channel for both directions (because broadcast message will get delivered also to the same process).

src/common/cluster-ipc.ts
jakolehm: Agreed, let's see if I can refactor whole `cluster-ipc` module away.

jakolehm: Strings are gone now.

src/common/base-store.ts
```diff
@@ -37,12 +37,16 @@ export class BaseStore<T = any> extends Singleton {
...
   }
 
-  get path() {
-    return this.storeConfig.path;
+  get syncRendererChannel() {
+    return `store-sync-renderer:${this.name}`
+  }
+
+  get syncMainChannel() {
+    return `store-sync-main:${this.name}`
```
ixrock: This event-name should be more unique to not overlap from different extensions which use `ExtensionStore < BaseStore`.

```suggestion
    return `store-sync-main:${this.path}`
```

src/common/cluster-ipc.ts
```diff
@@ -1,59 +1,49 @@
...
-  }),
+if (ipcMain) {
+  handleRequest(clusterActivateHandler, (event, clusterId: ClusterId, force = false) => {
+    const cluster = clusterStore.getById(clusterId);
+    if (cluster) {
+      return cluster.activate(force);
+    }
+  })
+
+  handleRequest(clusterSetFrameIdHandler, (event, clusterId: ClusterId, frameId?: number) => {
```
ixrock: `frameId` probably should be non-optional?

src/common/ipc.ts
```diff
@@ -1,79 +1,72 @@
...
+async function getSubFrames(): Promise<number[]> {
+  const subFrames: number[] = [];
+  const { clusterStore } = await import("./cluster-store");
+  clusterStore.clustersList.forEach(cluster => {
+    if (cluster.frameId) {
+      subFrames.push(cluster.frameId)
+    }
+  });
+  return subFrames;
+}
```
ixrock: Instead of managing `frameId` in cluster/cluster-store, maybe better to handle a `Map<clusterId, frameId` here and re-use in broadcast? (single place where it used currently)

```ts
export const clusterFrameMap = observable.map<ClusterId, string>();

if(ipcMain){
  handleRequest(clusterSetFrameIdHandler, (clusterId: ClusterId, frameId: number) => {
    clusterFrameId.set(clusterId, frameId)
  })
}
```

ixrock (approved):
LGTM

src/common/ipc.ts
ixrock: Btw now `cluster-ipc.ts` looks like `ipc-cluster-server` :D

src/common/base-store.ts
jakolehm: Good catch!

src/common/ipc.ts
jakolehm: > Instead of managing frameId in cluster/cluster-store, maybe better to handle a Map<clusterId, frameId here and re-use in broadcast? 

That's a good idea, refactored and now we also got rid of `Cluster.frameId`.

nevalla (approved):
LGTM

PR #1230, opened 2020-11-05
Documentation (MKDOCS)

msa0311:
Initial creation of documentation, structure, content and styles using mkdocs.


.github/workflows/main.yml
```diff
@@ -0,0 +1,49 @@
+name: Publish docs via GitHub Pages
+on:
+  push:
+    branches:
+      - extensions-docs
```
jakolehm: Should this be `master`?

src/main/menu.ts
```diff
@@ -193,6 +193,21 @@ export function buildMenu(windowManager: WindowManager) {
           navigate(whatsNewURL())
         },
       },
+<<<<<<< HEAD
```
jakolehm: This conflict needs to be resolved.

jakolehm (changes requested):

src/main/menu.ts
msa0311: done

.github/workflows/main.yml
msa0311: yes, correct, just updated the pull request!

jakolehm (approved):

PR #1233, opened 2020-11-05
Extensions api fixes

ixrock:
_Changes:_

- fix: listing extensions itself instead of it's instances in UI
- fix: create extension instance only when enabled
- extensions active state now handled in common `extensions-store.ts`

### DIFF
--- a/src/extensions/extension-loader.ts
+++ b/src/extensions/extension-loader.ts
@@ -97,22 +97,24 @@ export class ExtensionLoader {
       for (const [extId, ext] of installedExtensions) {
         let instance = this.instances.get(extId);
         if (ext.enabled && !instance) {
-          const extensionModule = this.requireExtension(ext)
-          if (!extensionModule) {
-            continue;
-          }
           try {
+            const extensionModule = this.requireExtension(ext)
+            if (!extensionModule) continue;
             const LensExtensionClass: LensExtensionConstructor = extensionModule.default;
             instance = new LensExtensionClass(ext);
             instance.whenEnabled(() => register(instance));
             instance.enable();
             this.instances.set(extId, instance);
           } catch (err) {
-            logger.error(`[EXTENSIONS-LOADER]: init extension instance error`, { ext, err })
+            logger.error(`[EXTENSION-LOADER]: activation extension error`, { ext, err })
           }
         } else if (!ext.enabled && instance) {
-          instance.disable();
-          this.instances.delete(extId);
+          try {
+            instance.disable();
+            this.instances.delete(extId);
+          } catch (err) {
+            logger.error(`[EXTENSION-LOADER]: deactivation extension error`, { ext, err })
+          }
         }
       }
     }, {
--- a/src/extensions/extension-loader.ts
+++ b/src/extensions/extension-loader.ts
@@ -16,46 +16,34 @@ export function extensionPackagesRoot() {
 }
 
 export class ExtensionLoader {
-  @observable isLoaded = false;
   protected extensions = observable.map<LensExtensionId, InstalledExtension>();
-  protected instances = observable.map<LensExtensionId, LensExtension>()
+  protected instances = observable.map<LensExtensionId, LensExtension>();
+
+  @observable isLoaded = false;
+  whenLoaded = when(() => this.isLoaded);
 
   constructor() {
     if (ipcRenderer) {
-      ipcRenderer.on("extensions:loaded", (event, extensions: InstalledExtension[]) => {
+      ipcRenderer.on("extensions:loaded", (event, extensions: [LensExtensionId, InstalledExtension][]) => {
         this.isLoaded = true;
-        extensions.forEach((ext) => {
-          if (!this.extensions.has(ext.manifestPath)) {
-            this.extensions.set(ext.manifestPath, ext)
+        extensions.forEach(([extId, ext]) => {
+          if (!this.extensions.has(extId)) {
+            this.extensions.set(extId, ext)
           }
         })
       });
     }
-    this.manageExtensionsState();
+    extensionsStore.manageState(this);
   }
 
-  @computed get userExtensions(): InstalledExtension[] {
-    return Array.from(this.toJSON().values()).filter(ext => !ext.isBundled)
-  }
-
-  protected async manageExtensionsState() {
-    await extensionsStore.whenLoaded;
-    await when(() => this.isLoaded);
-
-    // apply initial state
-    this.extensions.forEach((ext, extId) => {
-      ext.enabled = ext.isBundled || extensionsStore.isEnabled(extId);
+  @computed get userExtensions(): Map<LensExtensionId, InstalledExtension> {
+    const extensions = this.extensions.toJS();
+    extensions.forEach((ext, extId) => {
+      if (ext.isBundled) {
+        extensions.delete(extId);
+      }
     })
-
-    // handle updated state from store
-    reaction(() => extensionsStore.extensions.toJS(), extensionsState => {
-      extensionsState.forEach((state, extId) => {
-        const ext = this.extensions.get(extId);
-        if (ext && !ext.isBundled && ext.enabled !== state.enabled) {
-          ext.enabled = state.enabled;
-        }
-      })
-    });
+    return extensions;
   }
 
   @action
@@ -93,14 +81,13 @@ export class ExtensionLoader {
   }
 
   protected autoInitExtensions(register: (ext: LensExtension) => Function[]) {
-    return reaction(() => this.toJSON(), (installedExtensions) => {
+    return reaction(() => this.toJSON(), installedExtensions => {
       for (const [extId, ext] of installedExtensions) {
         let instance = this.instances.get(extId);
-        if (ext.enabled && !instance) {
+        if (ext.isEnabled && !instance) {
           try {
-            const extensionModule = this.requireExtension(ext)
-            if (!extensionModule) continue;
-            const LensExtensionClass: LensExtensionConstructor = extensionModule.default;
+            const LensExtensionClass: LensExtensionConstructor = this.requireExtension(ext)
+            if (!LensExtensionClass) continue;
             instance = new LensExtensionClass(ext);
             instance.whenEnabled(() => register(instance));
             instance.enable();
@@ -108,7 +95,7 @@ export class ExtensionLoader {
           } catch (err) {
             logger.error(`[EXTENSION-LOADER]: activation extension error`, { ext, err })
           }
-        } else if (!ext.enabled && instance) {
+        } else if (!ext.isEnabled && instance) {
           try {
             instance.disable();
             this.instances.delete(extId);
@@ -131,7 +118,7 @@ export class ExtensionLoader {
         extEntrypoint = path.resolve(path.join(path.dirname(extension.manifestPath), extension.manifest.main))
       }
       if (extEntrypoint !== "") {
-        return __non_webpack_require__(extEntrypoint)
+        return __non_webpack_require__(extEntrypoint).default;
       }
     } catch (err) {
       console.error(`[EXTENSION-LOADER]: can't load extension main at ${extEntrypoint}: ${err}`, { extension });
@@ -139,7 +126,11 @@ export class ExtensionLoader {
     }
   }
 
-  toJSON() {
+  getExtension(extId: LensExtensionId): InstalledExtension {
+    return this.extensions.get(extId);
+  }
+
+  toJSON(): Map<LensExtensionId, InstalledExtension> {
     return toJS(this.extensions, {
       exportMapsAsObjects: false,
       recurseEverything: true,
@@ -153,7 +144,7 @@ export class ExtensionLoader {
       frameId: frameId,
       frameOnly: !!frameId,
       args: [
-        Array.from(this.toJSON().values()),
+        Array.from(this.toJSON()),
       ],
     })
   }
--- a/src/extensions/extension-loader.ts
+++ b/src/extensions/extension-loader.ts
@@ -4,10 +4,11 @@ import type { LensRendererExtension } from "./lens-renderer-extension"
 import type { InstalledExtension } from "./extension-manager";
 import path from "path"
 import { broadcastIpc } from "../common/ipc"
-import { computed, observable, reaction, when } from "mobx"
+import { action, computed, observable, reaction, toJS, when } from "mobx"
 import logger from "../main/logger"
 import { app, ipcRenderer, remote } from "electron"
 import * as registries from "./registries";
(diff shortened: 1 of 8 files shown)


src/extensions/lens-extension.ts
```diff
@@ -14,35 +13,17 @@ export interface LensExtensionManifest {
...
-}
-
-export class LensExtension<S extends ExtensionStore<LensExtensionStoreModel> = any> {
-  protected store: S;
+export class LensExtension {
   readonly manifest: LensExtensionManifest;
   readonly manifestPath: string;
   readonly isBundled: boolean;
 
+  @observable isEnabled = false;
```
jakolehm: Should this be `private`? Does it need to be observable?

ixrock: Yes, it needs to be observable (used in `extension.whenEnabled`)

ixrock: Added `private` modifier.

nevalla (approved):
LGTM. @jakolehm PTAL again.

src/extensions/extension-loader.ts
```diff
@@ -71,22 +93,26 @@ export class ExtensionLoader {
...
             instance = new LensExtensionClass(ext);
             instance.whenEnabled(() => register(instance));
-            this.instances.set(ext.manifestPath, instance);
+            instance.enable();
+            this.instances.set(extId, instance);
           } catch (err) {
             logger.error(`[EXTENSIONS-LOADER]: init extension instance error`, { ext, err })
           }
+        } else if (!ext.enabled && instance) {
+          instance.disable();
```
jim-docker: Should we try/catch here since `disable()` calls the extension developer implemented `onDeactivate()`?

src/extensions/lens-extension.ts
```diff
@@ -64,15 +45,15 @@ export class LensExtension<S extends ExtensionStore<LensExtensionStoreModel> = a
   @action
   async enable() {
     if (this.isEnabled) return;
-    this.store.data.isEnabled = true;
+    this.isEnabled = true;
     this.onActivate();
```
jim-docker, line 49: or maybe we should try/catch here (and in `disable()`) in case the extension developer implemented `onActivate()`/`onDeactivate()` throw?

src/extensions/extension-loader.ts
ixrock: Yeah, good idea.

src/extensions/extension-loader.ts
```diff
@@ -30,18 +31,39 @@ export class ExtensionLoader {
         })
       });
     }
+    this.manageExtensionsState();
   }
 
-  @computed get userExtensions(): LensExtension[] {
-    return [...this.instances.values()].filter(ext => !ext.isBundled)
+  @computed get userExtensions(): InstalledExtension[] {
+    return Array.from(this.toJSON().values()).filter(ext => !ext.isBundled)
```
Nokel81: Is there a performance difference between the array spread format and the array from call?

src/extensions/extension-loader.ts
```diff
@@ -30,18 +31,39 @@ export class ExtensionLoader {
...
-    const installedExtensions = await extensionManager.load();
-    this.extensions.replace(installedExtensions);
+  protected async manageExtensionsState() {
+    await extensionsStore.whenLoaded;
+    await when(() => this.isLoaded);
+
+    // apply initial state
+    this.extensions.forEach((ext, extId) => {
+      ext.enabled = ext.isBundled || extensionsStore.isEnabled(extId);
+    })
```
Nokel81: Shouldn't this be done in the `InstalledExtension` constructor?

src/extensions/extension-loader.ts
```diff
@@ -71,21 +93,27 @@ export class ExtensionLoader {
...
-          }
+    return reaction(() => this.toJSON(), (installedExtensions) => {
+      for (const [extId, ext] of installedExtensions) {
+        let instance = this.instances.get(extId);
+        if (ext.enabled && !instance) {
           try {
+            const extensionModule = this.requireExtension(ext)
+            if (!extensionModule) continue;
             const LensExtensionClass: LensExtensionConstructor = extensionModule.default;
             instance = new LensExtensionClass(ext);
```
Nokel81, line 91: Yes it is technically the same type, but line 98 above really should be `this.instances.has(extId)` because that is what it cares about.

src/extensions/extension-loader.ts
```diff
@@ -71,21 +93,27 @@ export class ExtensionLoader {
...
-            continue
-          }
+    return reaction(() => this.toJSON(), (installedExtensions) => {
+      for (const [extId, ext] of installedExtensions) {
+        let instance = this.instances.get(extId);
+        if (ext.enabled && !instance) {
           try {
+            const extensionModule = this.requireExtension(ext)
+            if (!extensionModule) continue;
             const LensExtensionClass: LensExtensionConstructor = extensionModule.default;
```
Nokel81: `requireExtension`'s return value should be `{ default: LensExtensionConstructor }` to avoid this.

src/extensions/extension-loader.ts
```diff
@@ -30,18 +31,39 @@ export class ExtensionLoader {
...
-    const { extensionManager } = await import("./extension-manager");
-    const installedExtensions = await extensionManager.load();
-    this.extensions.replace(installedExtensions);
+  protected async manageExtensionsState() {
+    await extensionsStore.whenLoaded;
+    await when(() => this.isLoaded);
+
+    // apply initial state
+    this.extensions.forEach((ext, extId) => {
+      ext.enabled = ext.isBundled || extensionsStore.isEnabled(extId);
```
Nokel81: nvm, this should be done purely in the `extensionsStore`. I don't see any reason why it shouldn't know that entire truth. I also see a problem here where there is a disconnect between `this.extensions` and `extensionStore` for the enabled flag. There should be one source of truth.

src/extensions/extension-manager.ts
```diff
@@ -8,9 +8,10 @@ import { extensionPackagesRoot } from "./extension-loader"
 import { getBundledExtensions } from "../common/utils/app-version"
 
 export interface InstalledExtension {
```
Nokel81, line 10: shouldn't this be a class? That way there isn't an abstraction boundary being broken by the loader.

src/extensions/extension-manager.ts
```diff
@@ -8,9 +8,10 @@ import { extensionPackagesRoot } from "./extension-loader"
...
-  manifest: LensExtensionManifest;
-  manifestPath: string;
-  isBundled?: boolean; // defined in package.json
+  readonly manifest: LensExtensionManifest;
+  readonly manifestPath: string;
+  readonly isBundled?: boolean; // defined in package.json
```
Nokel81: Even if you leave this as an interface, this should not be optional. Also, which `package.json`. I know it means `Lens`'s but it could mean an extension's. Please update the comment.

src/extensions/extensions-store.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  @observable extensions = observable.map<LensExtensionId, LensExtensionState>();
+
+  @action
+  setEnabled(extId: LensExtensionId, enabled: boolean) {
+    const state = this.extensions.get(extId);
+    this.extensions.set(extId, {
+      ...(state || {}),
+      enabled: enabled,
+    })
```
Nokel81: We are talking about observable objects here, I don't believe that all this code is necessary. 
```suggestion
    this.extensions.get(extId).enabled = enabled
```

src/extensions/extensions-store.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  isEnabled(extensionId: LensExtensionId) {
+    const state = this.extensions.get(extensionId);
+    return !state /* enabled by default */ || state.enabled;
```
Nokel81, line 60: I think that is would be clearer if the following was used. Because the null-coelessing operator is exactly what we are trying to model.

```suggestion
    return state?.enabled ?? true
```

src/extensions/lens-extension.ts
```diff
@@ -14,35 +13,17 @@ export interface LensExtensionManifest {
...
-}
-
-export class LensExtension<S extends ExtensionStore<LensExtensionStoreModel> = any> {
-  protected store: S;
+export class LensExtension {
   readonly manifest: LensExtensionManifest;
   readonly manifestPath: string;
   readonly isBundled: boolean;
 
+  @observable private isEnabled = false;
```
Nokel81, line 21: With this being private, I don't think it implements the `InstalledExtension` interface anymore, is that desirable?

src/main/index.ts
```diff
@@ -76,7 +79,7 @@ app.on("ready", async () => {
   }
 
   LensExtensionsApi.windowManager = windowManager = new WindowManager(proxyPort);
-  extensionLoader.init(); // call after windowManager to see splash earlier
+  extensionLoader.init(await extensionManager.load()); // call after windowManager to see splash earlier
```
Nokel81, line 82: why are we `load`-ing the extrensionManager twice in the same function?

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -79,11 +82,11 @@ export class Extensions extends React.Component {
...
           </div>
-          {!isEnabled && (
-            <Button plain active onClick={() => ext.enable()}>Enable</Button>
+          {!enabled && (
+            <Button plain active onClick={() => extensionsStore.setEnabled(extId, true)}>Enable</Button>
           )}
-          {isEnabled && (
-            <Button accent onClick={() => ext.disable()}>Disable</Button>
+          {enabled && (
+            <Button accent onClick={() => extensionsStore.setEnabled(extId, false)}>Disable</Button>
```
Nokel81: I really think that this should either be a ternary or a separate function (I lean towards separate function) since that is what you really want (either or).

Nokel81 (changes requested):

src/extensions/extension-loader.ts
ixrock: Nope, it need to wait loading extensions. Only for `bundled` is possible to define on earlier stages.

src/extensions/extensions-store.ts
ixrock: State is empty for all extensions by default (they are enabled)

Nokel81: Ah I see

src/extensions/extension-loader.ts
ixrock: Nope, just for readability.

src/main/index.ts
ixrock: wat? 🥱

Nokel81: oh oops sorry, I misread `extensionStore` as `extensionManager`. My bad.

src/extensions/lens-extension.ts
ixrock: Why you think it should implement? Using `private` field was suggested by @jakolehm

Nokel81: Maybe I don't have a grasp of how these types interact, I would have assumed that the `extension-manager.ts` file keeps a record to instances of this type. Does it not?

ixrock: Nope.. `extension-loader.ts` keeps both `InstalledExtension` list and enabled instances of those.

Nokel81: IMO that seems very strange, because we are keeping two lists of information that should always be the same. Why not just one list?

ixrock: Extension-manager doesn't keep anything.. Its purpose for install & load extensions.

Nokel81: Isn't `extension-loader` for installing and loading extensions? I would have assumed that `extension-manager` is the manager of truth and then uses `loader` to load stuff and `store` to persist data between restarts...

src/extensions/extension-manager.ts
```diff
@@ -8,9 +8,10 @@ import { extensionPackagesRoot } from "./extension-loader"
...
+  enabled?: boolean;
 }
```
Nokel81, line 16: I want to comment on Line 25 and maybe this should be done in a follow up PR (make that a certainly as to not hold up this PR too much) but I am very confused about `ExtensionManager` since it doesn't actually hold on to a list of extensions that are loaded.

ixrock:
@jakolehm @Nokel81 PTAL again

src/extensions/extensions-store.ts
ixrock: I personally don't like `??` operator and `||` looks more logically as `OR`.
So if checking something that expected to be an object I prefer `||`.

src/extensions/extension-manager.ts
```diff
@@ -8,9 +8,10 @@ import { extensionPackagesRoot } from "./extension-loader"
 import { getBundledExtensions } from "../common/utils/app-version"
 
 export interface InstalledExtension {
-  manifest: LensExtensionManifest;
-  manifestPath: string;
-  isBundled?: boolean; // defined in package.json
+  readonly manifest: LensExtensionManifest;
+  readonly manifestPath: string;
+  readonly isBundled?: boolean; // defined in project root's package.json
```
Nokel81: nit: still don't think this should be optional

src/extensions/extensions-store.ts
Nokel81: Is it just how the `??` operator looks? Because from a meaning standpoint `??` is closer to what you want than `||`.

Nokel81 (approved):
I'll approve this PR so that we can fix master with it. But I am still dissatisfied with this design of this part of the system. I think it is too interconnected and doesn't really match how the rest of our systems are designed.

src/extensions/extension-manager.ts
ixrock: yep, forgot to remove `?`

ixrock:
Yes, maybe, but afaik it was desire of @jakolehm to keep these parts separated (manager, loader & store).

src/extensions/lens-extension.ts
jakolehm: About private: without private extension instance itself cold toggle this which does not make any sense to me and is not really desirable. Why would we allow extension to do this?

`ExtensionLoader`: it's only purpose is to load (read: execute) extension instances. List of extensions are given to this class, it does not care how that list has generated.

`ExtensionManager`: it's only purpose is to gather a list of extensions and then _install_ those to central `node_modules`  folder (including all dependencies). Note: this does not _load_ (or create) extension instances.

Naming things is one of the most difficult things in computer science so we probably could have better names for these. Still I strongly believe that separating complex things to smaller units is better than having one giant class that tries to do everything.

PR #1234, opened 2020-11-05
Keeping Icon focus state only for keyboard

### LINKED ISSUE #1228: Weird icons active/hover state
There is equal styles applied for `:active` and `:hover` icons which causes frozen icon effect - it keeps hovered state even after clicked. Also, there's some strange hovering issues in cluster add button and drawer toolbar icons.

add cluster btn
toolbar hovering
sticked state

aleksfront:
Removing `:focus` styles for `<Icon />` component when using mouse. This fixes issue when Icon seems frozen in hover state even if cursor went away.

Determining mouse/keyboard usage by toggling `.mouse-intent` class name on `body` element.

Added outline styling for Icon's `:focus` state.

keyboard focusing
mouse focusing

Close #1228

### DIFF
--- a/src/renderer/components/layout/main-layout.tsx
+++ b/src/renderer/components/layout/main-layout.tsx
@@ -38,17 +38,6 @@ export class MainLayout extends React.Component<MainLayoutProps> {
     (sidebarWidth) => this.storage.merge({ sidebarWidth })
   );
 
-  componentDidMount() {
-    // Toggling .mouse-intent class if mouse or keyboard using in the dashboard
-    // This allows to remove focus styling to elements when mouse is used
-    window.addEventListener("click", (evt) => {
-      if (!evt.detail) return; // No mouse used (e.g. hitted spacebar on button)
-      document.body.classList.add("mouse-intent");
-    })
-    window.addEventListener("keydown", (evt) => {
-      document.body.classList.remove("mouse-intent");
-    })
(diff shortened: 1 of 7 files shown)


src/renderer/components/layout/main-layout.tsx
```diff
@@ -38,6 +38,18 @@ export class MainLayout extends React.Component<MainLayoutProps> {
     (sidebarWidth) => this.storage.merge({ sidebarWidth })
   );
 
+  componentDidMount() {
+    // Toggling .mouse-intent class if mouse or keyboard using in the dashboard
+    // This allows to remove focus styling to elements when mouse is used
+    window.addEventListener("click", (evt) => {
```
ixrock: Why this is in `MainLayout`? 
In addition there is no unbind events in `componentWillUnmount`..

ixrock: Plus it's better to keep it close to `icon` component if it's related only to this.

ixrock (changes requested):

aleksfront:
@ixrock Please do re-review. Replaced custom solution with `what-input` dependency which could track used input device
https://www.npmjs.com/package/what-input
https://bundlephobia.com/result?p=what-input@5.2.10

ixrock (approved):

PR #1087, opened 2020-10-14
Fix rendering of boolean values in CRDs

Nokel81:
- add optional special casing for boolean values in DrawerItems and
  TableRows since React (imo annoying fashion) does not render boolean
  values by default.
- add a spinner on the sidebar for when the CRD menu is expeanded but
  the entries have not been loaded yet.
- Add ability to double click a Badge to expand, also make it so that
  Badges highligh all text on first click.

Signed-off-by: Sebastian Malton <sebastian@malton.name>

fixes #1071

PR #1236, opened 2020-11-05
add button to expand and view whole badge contents

Nokel81:

This PR is takes work previously in #1087 since it is a separate item.

`Badge`'s currently cap the size of their rendering to a single line of text (to max displayable width) and use an ellipsis to show that more content is available. It was previously not possible to expand a `Badge` to view the whole of its content.

This PR adds a button that is displayed only if there is more content to view to expand the `Badge`.

This PR also moves to a fork of material-icons that is maintained and even improved. The old version (the original by google) has not been updated in over 4 years and is missing some icons (compared to what the website says it has).


src/renderer/components/badge/badge.tsx
```diff
@@ -3,21 +3,59 @@ import "./badge.scss"
...
+import { Icon } from "../icon";
 
 interface Props extends React.HTMLAttributes<any>, TooltipDecoratorProps {
   small?: boolean;
   label?: React.ReactNode;
 }
 
 @withTooltip
+@observer
 export class Badge extends React.Component<Props> {
```
ixrock: Please move all new logic to new component based on `Badge` and put to same folder. 
It's simple component which is not intended to display everything, but mostly plain-text labels.

package.json
```diff
@@ -230,6 +230,7 @@
     "lodash": "^4.17.15",
     "mac-ca": "^1.0.4",
     "marked": "^1.1.0",
+    "material-design-icons-iconfont": "^6.1.0",
```
ixrock: Why this is needed? We have embedded (downloaded) font already for material-icons (from latest master).
See `renderer/fonts/MaterialIcons-Regular.ttf`

ixrock (dismissed):

package.json
Nokel81: See my comment on the main description. That "latest master" is years old and out of date.

ixrock: Updated 2 months ago..
https://github.com/google/material-design-icons/tree/master/font

ixrock: Which icons are missing there ^?

Nokel81: Well none the less, even that font is missing "open_in_full" and "close_fullscreen". So it might have been updated but it is still incomplete.

ixrock: They both available currently in master:
```
<Icon material={"open_in_full"}/>
<Icon material={"close_fullscreen"}/>
```
image

src/renderer/components/badge/badge.tsx
Nokel81: I agree that is is designed to display plain-text labels. But I disagree that this logic should be moved out. The ubiquity of `Badge` being used for even things that are most likely large (like annotations) means that this code should be in `Badge`. Plus I do not see any place where `Badge` is used where a user might not want to be able to temporarily see the whole thing.

package.json
Nokel81: Okay I see what has happened. I started work on this 22 days ago, but you only updated to the latest version 12 days ago, so when I was looking into this we were behind.

src/renderer/components/icon/icon.tsx
```diff
@@ -87,7 +87,7 @@ export class Icon extends React.PureComponent<IconProps> {
 
     // render as inline svg-icon
     if (svg) {
-      const svgIconText = require(`!!raw-loader!./${svg}.svg`).default;
+      const svgIconText = require("!!raw-loader!./" + svg + ".svg").default;
```
ixrock: Is there any difference? :)

Nokel81: No, but I think I had already reverted this by the time... lol.

I just, in general, prefer the "`" syntax

nevalla:
Are we going to use current `Badge` component as is somewhere else? If not, do we need to have that kind of abstract component that we want to extend? IMHO if we want to use this feature on all badges that we currently have, it makes sense to implement it to directly to current component. If we need to have separate badge components in future , let's do the refactoring then.

nevalla:
IMHO the icon is way too big atm. Maybe even size of `10` might be enough, wdyt?

Nokel81:
@nevalla Sure sounds good

Nokel81:
@nevalla Can you PTL again. I experimented around with different sizes and found that `10px` was a tad small so I went with `12px`.

jakolehm:
> @nevalla Can you PTL again. I experimented around with different sizes and found that `10px` was a tad small so I went with `12px`.

@Nokel81 could you attach a screenshot?

Nokel81:
sure, @jakolehm 

12px:
Screen Shot 2020-11-16 at 9 48 31 AM
--
10px:
Screen Shot 2020-11-16 at 9 50 05 AM
--

miskun:
Can we do it like this:
1. the "label" will always contain the entire value.
2. If the value does not fit in the label "display area", we'll show portion of it and "..." to hint there is more data hidden at the end (or in beginning or around the display area). 
3. if value is not displayed 100%, we can have several options how to peek the data that is hidden;
3.1. auto-scroll; gently scroll the contents of label; e.g. we can trigger this on hover; AND/OR
3.2. user select; if user clicks on display area, make it possible to copy & paste portion of value. if selection is expanded left/right, scroll accordingly. CMD+A should select entire value.

I hope you get the idea from above description

Nokel81:
@miskun 

1. do you mean the hover text on `Badge`'s? If so I believe that this is already the case.
2. We already show a `...` if there is more information in the badge then what we can currently display
3. So I take it from this suggestion that you would prefer an auto scroll design instead of a toggleable design. I could definitely do that instead. We already support copying a portion of the text (the part which is visible).

Personally, I find slow auto-scrolling text rather annoying to read. Especially when it is very long (and some of these secrets are many (4+) lines long. Something that an auto-scrolling alternative would not be able to support is view the whole of an item or even the whole of multiple items.

During the development of this improvement I experimented with not having a button but instead making the whole `Badge` clickable. This was not very good UX because when you went to highlight it would randomly close instead. There is the experimental `onselectionchange` event which I could have used but that was not an avenue that I went down.

miskun:
Think about label as text input field with read only attribute. You can
click it, cursor will display; scroll around, select all or part...

ixrock:
Can we just toggle label by clicking it (short/full-version)?

Nokel81:
@ixrock Do you mean instead of having that icon button? We could, I did try that but it would close while highlighting text. I find that a small button is a much better nonce then changing the cursor on hover.

ixrock:
@Nokel81 why it would close while you highlighting something? It should close only when it clicked again I assume.
If we just move same `onClick` from icon to `Badge` label/top-element?

Nokel81:
@ixrock Because while highlighting text on `onClick` even does fire. I think I know of a possible solution (either using the WIP `onselectionchange` event or by doing some manual checking on selection change in the `onClick` handler) though neither of those seem very pretty.

miskun:
I think the bottom line is this: let's not add a button/icon to interact
(whatever reason). Clicking the label / text should be enough.

Nokel81:
Closing in favour of #1400

PR #1245, opened 2020-11-05
Add designated folder for extensions

### LINKED ISSUE #1179: Provide easy way for extensions to store files
Each extension should have their own "sandbox" where files are actually stored.

Nokel81:
This PR adds a method onto `LensExtension` for getting a unique path to a folder that the extension can use for its own purposes. 

The folder is based on the extension ID but is not passed in so it cannot be faked. The folder name is salted and then hashed to obfuscate it and the mapping is stored in a lens store which is, by design, not otherwise exposed to extension writers. This is done so that the folder's name is not able to be determined without calling this function.

No security is applied to these folders.

fixes #1179


ixrock:
This should be merged first https://github.com/lensapp/lens/pull/1233

Nokel81:
okay, ping me when that is merged and I will rebase again

package.json
```diff
@@ -191,7 +191,8 @@
       "node-menu",
       "metrics-cluster-feature",
       "license-menu-item",
-      "support-page"
+      "support-page",
+      "example-extension"
```
nevalla: no need to enable `example-extension` for all.

Nokel81: I'll try find another way to force it's build failure to block PRs. Since I actually suspect that no extension build failure blocks PRs at the moment. 

However, I do think that we should (at least for testing) always have it activated because we should be constantly testing it to make sure that it doesn't start throwing errors.

jakolehm: If we want to have it tested automatically then we need to come up with a way to inject it during tests.

Nokel81: okay sounds good

extensions/example-extension/main.ts
```diff
@@ -2,11 +2,12 @@ import { LensMainExtension } from "@k8slens/extensions";
 
 export default class ExampleExtensionMain extends LensMainExtension {
   async onActivate() {
-    console.log('EXAMPLE EXTENSION MAIN: ACTIVATED', this.id, this.name, this.version);
+    console.log('EXAMPLE EXTENSION MAIN: ACTIVATED', this.id, this.name, this.version)
     console.log("EXAMPLE EXTENSION MAIN: has access to special folder", await this.getExtensionFileFolder())
+    throw 5
```
ixrock: What is it? Forgot to remove?

Nokel81: Yes

src/main/menu.ts
```diff
@@ -26,7 +26,7 @@ export function showAbout(browserWindow: BrowserWindow) {
     `Copyright 2020 Mirantis, Inc.`,
   ]
   dialog.showMessageBoxSync(browserWindow, {
-    title: `${isWindows ? " ".repeat(2) : ""}${appName}`,
+    title: `${" ".repeat(isWindows as any * 2)}${appName}`,
```
ixrock: This is totally not obvious: `isWindows as any * 2`, please revert.
Are you trying to reduce some symbols or improve readability? :)

Nokel81: oops, I should try to have better commit control. I was seeing if this would work is JS since it is a very python-esk type of operation.

integration/__tests__/app.tests.ts
```diff
@@ -71,27 +74,56 @@ describe("Lens integration tests", () => {
       await app.client.keys(['Meta', 'Q'])
       await app.client.keys('Meta')
     })
+
+    it.skip("should register all in tree extensions", async () => {
+      // TODO: skipping since it doesn't seem like "index.ts" is being run.
+
+      await when(() => extensionLoader.isLoaded)
```
ixrock: `await extensionLoader.whenLoader`

src/extensions/extension-loader.ts
```diff
@@ -86,20 +86,25 @@ export class ExtensionLoader {
...
+        if (ext.isEnabled && !alreadyInit) {
           try {
-            const LensExtensionClass: LensExtensionConstructor = this.requireExtension(ext)
-            if (!LensExtensionClass) continue;
-            instance = new LensExtensionClass(ext);
+            const LensExtensionClass = this.requireExtension(ext)
+            if (!LensExtensionClass) {
+              continue;
+            }
+
```
ixrock, line 135: Redundant space? Or is it intentional?

src/main/extension-filesystem.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    const dirPath = this.registeredExtensions.get(extensionName)
+    await ensureDir(dirPath)
+    return dirPath
+  }
+
+  @action
+  protected fromStore({ extensions }: FSProvisionModel = { extensions: {} }): void {
+    for (const extension in extensions) {
```
ixrock: Normally it's a bad practice to use `for-in` for objects.. 
Define `registeredExtensions` as `registeredExtensions = observable.map<LensExtensionId, string>()` and you will get merge-api: `this.registeredExtensions.merge(extensions)`

src/main/extension-filesystem.ts
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * This function retrieves the saved path to the folder which the extension
+   * can saves files to. If the folder is not present then it is created.
+   * @param extensionName the name of the extension requesting the path
+   * @returns path to the folder that the extension can safely write files to.
+   */
+  async requestDirectory(extensionName: string): Promise<string> {
+    if (!this.registeredExtensions.has(extensionName)) {
+      const salt = randomBytes(32).toString("hex")
+      const hashedName = SHA256(`${extensionName}/${salt}`).toString()
```
ixrock: Looks overcomplicated. Why all this hashes needed? For security? Why not expose folders as actual extension names?

src/main/extension-filesystem.ts
```diff
@@ -0,0 +1,58 @@
...
+  @action
+  protected fromStore({ extensions }: FSProvisionModel = { extensions: {} }): void {
+    for (const extension in extensions) {
+      this.registeredExtensions.set(extension, extensions[extension])
+    }
+  }
+
+  toJSON(): FSProvisionModel {
+    return toJS({
+      extensions: Object.fromEntries(this.registeredExtensions.entries()),
```
ixrock:
```suggestion
      extensions: this.registeredExtensions.toJSON(),
```

src/main/index.ts
```diff
@@ -23,6 +23,12 @@ import { appEventBus } from "../common/event-bus"
 import { extensionLoader } from "../extensions/extension-loader";
 import { extensionManager } from "../extensions/extension-manager";
 import { extensionsStore } from "../extensions/extensions-store";
+import { filesystemProvisionerStore } from "./extension-filesystem";
+import packageInfo from "../../package.json"
+
+if (isTestEnv) {
+  packageInfo.lens.extensions.push("example-extension")
```
ixrock: Isn't it could be included inside actual test instead?

src/extensions/extension-loader.ts
Nokel81: Intentional, I like spaces after blocks as it is good eye candy for reading code. Helps with grouping similar lines, sort of like spaces between paragraphs.

src/main/extension-filesystem.ts
Nokel81: Good point

src/main/extension-filesystem.ts
Nokel81: I mentioned this in my original post. Maybe I should copy some of that here.... Namely, I want to force extension writers to use the methods provided in this PR and I don't want the name of the folder to be pre-determinable. That way it is harder (but not impossible since we are doing no actual sand-boxing) for a malicious extension to find the folder of a specific other extension.

src/main/index.ts
Nokel81: I actually want it loaded why in development mode as well because we should be testing to make sure **all** in-tree extensions are being built and run during normal development so that none of them lag behind (as the example has been)

src/extensions/extension-loader.ts
ixrock: Alright, but imho there is enough space since you've added `\n}` after `continue;`

Nokel81: The block counts as well... We don't need to save on whitespace here... Is there something that you don't like about spreading out each step within a function?

src/main/index.ts
ixrock: Maybe we should remove this example-extension at all since it's actually not participating in app work and there is separated repo for examples https://github.com/lensapp/lens-extension-samples ?

src/main/extension-filesystem.ts
```diff
@@ -1,17 +1,19 @@
 import { randomBytes } from "crypto"
 import { SHA256 } from "crypto-js"
 import { app } from "electron"
-import { ensureDir, mkdir, pathExists } from "fs-extra"
+import fse from "fs-extra"
 import { action, observable, toJS } from "mobx"
 import path from "path"
 import { BaseStore } from "../common/base-store"
 
+type LensExtensionId = string;
```
ixrock: This should be exported from `lens-extension.ts` as `type` (already defined there)

src/main/index.ts
Nokel81: Maybe... how are we making sure that that repo stays up to date?

src/main/extension-filesystem.ts
Nokel81: Ah fair, thanks

src/main/index.ts
ixrock: By updating used `@k8slens/extensions`? Anyway imho since we have that repo this is the place for samples.

src/main/extension-filesystem.ts
ixrock: Okay, sounds reasonable.

src/main/index.ts
Nokel81: I don't understand "... that repo this is the place for samples". Do you mean that the https://github.com/lensapp/lens-extension-samples repo should be for our samples and this in-tree should just be removed? (If that is the case then I am for it).

src/extensions/extension-loader.ts
ixrock: Didn't get what you mean about "spreading out each step"

src/main/index.ts
ixrock: yes

extensions/telemetry/src/telemetry-preferences-store.ts
```diff
@@ -23,6 +23,10 @@ export class TelemetryPreferencesStore extends Store.ExtensionStore<TelemetryPre
     this.data.enabled = v
   }
 
+  protected fromStore(data: TelemetryPreferencesModel): void {
+    this.data = data
```
jakolehm: Why is this needed?

Nokel81: I changed `BaseStore` to be an `abstract class` with `fromStore` and `toJSON` as its two `abstract` methods. This was to force people writing new stores to remember that those are the two methods that **need** to be implemented. This was the only class that didn't, on the whole (now that you brought my mind back to it) this class should really be closer in design to the other stores and not keep `data` around at all, just a field `enabled`.

integration/__tests__/app.tests.ts
```diff
@@ -71,27 +74,56 @@ describe("Lens integration tests", () => {
       await app.client.keys(['Meta', 'Q'])
       await app.client.keys('Meta')
     })
+
+    it.skip("should register all in tree extensions", async () => {
```
jakolehm: Do we really need this if it's not working at all?

package.json
```diff
@@ -185,17 +185,6 @@
       "confinement": "classic"
     }
   },
-  "lens": {
```
jakolehm: Why are these removed?

src/common/base-store.ts
```diff
@@ -172,16 +175,19 @@ export class BaseStore<T = any> extends Singleton {
     return subFrames;
   }
 
-  @action
-  protected fromStore(data: T) {
-    if (!data) return;
-    this.data = data;
```
jakolehm: Are we really happy with this change.. does this mean that each store needs to decide where it sets data?

src/common/utils/app-version.ts
```diff
@@ -7,7 +7,3 @@ export function getAppVersion(): string {
 export function getBundledKubectlVersion(): string {
   return packageInfo.config.bundledKubectlVersion;
 }
-
-export function getBundledExtensions(): string[] {
```
jakolehm: Why?

jakolehm (changes requested):
`filesystemProvisionerStore` itself looks good but this PR introduces bunch of other changes that are not related to PR title at all. We should keep PRs small and focused for easier review process. Could you split non-related changes to separate PRs with descriptive title/description why those changes are done.

Nokel81:
@jakolehm You are correct about the sidecar changes. I have moved them to a separate branch.  Are you able to review again?

src/main/index.ts
```diff
@@ -55,6 +56,7 @@ app.on("ready", async () => {
     clusterStore.load(),
     workspaceStore.load(),
     extensionsStore.load(),
+    filesystemProvisionerStore.load(),
```
jakolehm, line 63: Should `filesystemProvisionerStore` load also in renderer?

Nokel81: Yes probably.

Nokel81: Done

jakolehm (approved):
LGTM

PR #1259, opened 2020-11-06
Docs: extension common capabilities

jakolehm:
image

### DIFF
--- a/docs/extensions/capabilities/common-capabilities.md
+++ b/docs/extensions/capabilities/common-capabilities.md
@@ -58,17 +58,33 @@ export default class ExampleMainExtension extends LensMainExtension {
 
 ## Renderer Extension
 
-A renderer extension runs in a browser context and it's visible directly via Lens main window. If you want to see logs from this extension you need to check them via Developer Tools -> console.
+A renderer extension runs in a browser context and it's visible directly via Lens main window. If you want to see logs from this extension you need to check them via View -> Toggle Developer Tools -> Console.
 
+### Activate
+
+An extension can register a custom callback that is executed when an extension is activated (started).
+
+``` javascript
+import { LensRendererExtension } from "@k8slens/extensions"
+
+export default class ExampleExtension extends LensRendererExtension {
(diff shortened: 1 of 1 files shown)


docs/extensions/capabilities/common-capabilities.md
```diff
@@ -0,0 +1,258 @@
+# Common Capabilities
+
+Common Capabilities are important building blocks for your extensions. Almost all extensions use some of these functionalities. Here is how you can take advantage of them.
+
+## Main Extension
```
jim-docker, line 5: I feel there's going to be some confusion about what we call an extension. An extension (as in a package we get from the extension marketplace) consists of one or both of the extension classes/objects "Main Extension" and "Renderer Extension". What about "Main Extension Object", or "Main Extender", or something else to provide distinction?

docs/extensions/capabilities/common-capabilities.md
```diff
@@ -0,0 +1,258 @@
...
+
+A main extension runs in the background and it's not visible directly via Lens main window. If you want to see logs from this extension you need to start Lens from the command line.
```
jim-docker: Do you mean a main extension object does not add content directly in the UI? Maybe something like:
A main extension runs in the background and, apart from app menu items, does not add content to the Lens UI.

docs/extensions/capabilities/common-capabilities.md
jakolehm: I think we should clarify these in extension anatomy? @nevalla wdyt?

docs/extensions/capabilities/common-capabilities.md
jakolehm: Yes, thanks!

nevalla (approved):
LGTM

docs/extensions/capabilities/common-capabilities.md
nevalla: Yes, there are room for improvements there.

Issue #1251, opened 2020-11-06
Isolate global styles from extensions


aleksfront:
Currently there are global styling affecting extensions, for example `cluster-manager.scss`
```
.ClusterManager main>* {
    position: absolute;
    left: 0;
    top: 0;
    right: 0;
    bottom: 0;
    display: flex;
    background-color: var(--mainBackground);
}
.ClusterManager main>*>* {
    flex: 1;
}
```
It provides unwanted side-effects for extension developers so they have no choice to overide some of them with a bunch of unnecessary properties with `!important` rule.

We need to find a way to isolate all global styling like layout, fonts and others from extension or provide necessary minimum amount of them.

ixrock:
Maybe we should isolate extensions inside shadow-dom somehow?
https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM

stefcameron:
> Maybe we should isolate extensions inside shadow-dom somehow?
> https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM

That's an interesting idea, however, while the Shadow DOM encapsulates styles defined within it so they don't leak out, it doesn't prevent external styles from affecting its internal elements. Here's a [sandbox](https://codesandbox.io/s/admiring-lalande-p4p2o) that implements a custom element that uses a Shadow DOM. The `styles.css` defines the body font as bold, and this affects the internal Shadow DOM elements.

ixrock:
@stefcameron probably we must reset styles to prevent leaking from outside like this:

```scss
// top-root element
.MyExtension { 
  all: initial;  
  // all other styles below
 }
```
And don't forget to isolate styles for extension itself with `:host { inner extension's style } <-- this is loaded inside shadow-dom`

aleksfront:
As @stefcameron mentioned in https://github.com/lensapp/lens/pull/1331#pullrequestreview-528334560 we should be responsible of isolating new styles to maximum level and get rid of selectors like `* >` etc. In that case only few basic ones described in `app.scss` would affect extensions and can easily be overwritten in rare cases.
@ixrock And there's no point of using shadow DOM for now at least, since it only encapsulates styles inside but not preventing global ones to affect inner contents.

PR #1267, opened 2020-11-06
Docs: extension styling and theming

aleksfront:
styling docs


docs/extensions/capabilities/styling.md
```diff
@@ -0,0 +1,104 @@
+# Styling
+Lens provides a set of global styles and UI components that can be used by any extension to preserve look and feel of the application. However, it’s always up to the developer whether to use them or provide completely different styles for the extension.
```
stefcameron: I think you can only make this claim
> However, it’s always up to the developer whether to use them or provide completely different styles for the extension.
if you have fixed #1251

docs/extensions/capabilities/styling.md
```diff
@@ -0,0 +1,104 @@
...
+When Lens gets loaded it transforms selected theme `json` file into list of [CSS Custom Properties (CSS Variables)](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) which then gets injected into `:root` element so any of the down-level components can use them.
+
+[css vars listed in devtools]
+
+When user changes a theme, the process is repeated, new css variables appear instead of previous ones.
+
+If you want to follow a selected theme to keep the 'native' Lens look and feel, respecting the light/dark appearance of your extension, you can use provided variables and build-in lens components such as buttons, dropdowns, checkboxes etc.
+
+## Injected styles
+Every extention is affected by list of default global styles defined in `src/renderer/components/app.scss`. These are basic browser resets like setting `box-sizing` property for every element, default text and background colors, default font size, basic headings visualisation etc.
```
stefcameron, line 34: This is the stuff I think extensions should be isolated from (including `flex.box` as mentioned earlier) as part of #1251. Overriding these to reset them would be quite a challenge. I realize that's separate from this issue.

For now, this paragraph is very useful, since it states the current behavior, and points users to the Lens code that defines all the styles.

docs/extensions/capabilities/styling.md
```diff
@@ -0,0 +1,104 @@
...
+$font-main: 'Roboto', 'Helvetica', 'Arial', sans-serif !default;
+$font-monospace: Lucida Console, Monaco, Consolas, monospace;
+$font-size-small: floor(1.5 * $unit);
+$font-size: floor(1.75 * $unit);
+$font-size-big: floor(2 * $unit);
+$font-weight-thin: 300;
+$font-weight-normal: 400;
+$font-weight-bold: 500;
+...
+```
```
stefcameron: This is good to know. #1251 should move all these into CSS variables to make them more accessible, like the colors are, regardless of using SCSS or not.

I'm also not sure how an extension would reference this SCSS file. It's not published in the `@k8slens/extensions` package...

Having the file reference is useful, but without being able to reference it in the extension's SCSS build, not very useful at all. If they were CSS variables, that would be a different story...

docs/extensions/capabilities/styling.md
```diff
@@ -0,0 +1,104 @@
...
+You can use them to set consistent paddings, define font-sizes in your SCSS files e.g.
+```
+.status {
+  padding-left: $padding * 2;
+  font-size: $font-size-small;
+}
+```
+
+### theme-vars.scss
+In `src/renderer/themes/theme-vars.scss` there are list of theme-defined colors. Most of their values are different for light and dark themes. You can use them to preserve consitent view of extension with respect of selected theme.
```
stefcameron: Same here: How would I reference this in an extension's build? These all reference CSS vars, so that's much better (more accessible to a wider range of extension implementations; my extension doesn't use SCSS, BTW, it uses Emotion and Lens components).

stefcameron: If you want to encourage extension developers to use SCSS, then you need:
1. to provide these SCSS files via some NPM package import;
2. document what all these variables are and where they should be used;
3. make sure that all your extension examples come with SCSS processor "built-in" to the build (none of them do, right now).

stefcameron (commented):
This is a great start at documentation for styles! It'll be very helpful, I'm sure. I learned about `flex.box` thanks to this; I knew there was some kind of library being used for flexbox in Lens, but didn't know what it was. Now at least I know what some styles available to me are.

My feedback in comments...

docs/extensions/capabilities/styling.md
aleksfront: I guess there's no urge for now to encourage developers to use exactly SCSS since the only benefit currently is to use few basic variables. I've transfered them into CSS vars and removed encouragement message.

docs/extensions/capabilities/styling.md
aleksfront: Converted all of these into CSS vars.

docs/extensions/capabilities/styling.md
aleksfront: Removed for now

aleksfront:
@stefcameron Thanks for your detailed feedback! I've made changes and added `Color reference` page where each theme color is documented.

docs/extensions/capabilities/styling.md
```diff
@@ -0,0 +1,99 @@
...
+## Themes
+Lens using 2 built-in themes located in `src/renderer/themes` folder each for light and dark color schemes. Active theme can be changed in the `Preferences` page.
+![Color Theme](images/theme-selector.png)
+
+When Lens gets loaded it transforms selected theme `json` file into list of [CSS Custom Properties (CSS Variables)](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) which then gets injected into `:root` element so any of the down-level components can use them.
+![CSS vars listed in devtools](images/css-vars-in-devtools.png)
+
+When user changes a theme, the process is repeated, new css variables appear instead of previous ones.
+
+If you want to follow a selected theme to keep the 'native' Lens look and feel, respecting the light/dark appearance of your extension, you can use provided variables and build-in lens components such as buttons, dropdowns, checkboxes etc.
```
ixrock, line 31:
```suggestion
If you want to follow a selected theme to keep the 'native' Lens look and feel, respecting the light/dark appearance of your extension, you can use provided variables and built-in Lens components such as `Button`, `Select`, `Table`, etc.
```

docs/extensions/capabilities/styling.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Injected styles
+Every extention is affected by list of default global styles defined in `src/renderer/components/app.scss`. These are basic browser resets like setting `box-sizing` property for every element, default text and background colors, default font size, basic headings visualisation etc.
```
ixrock, line 34:
```suggestion
Every extension is affected by list of default global styles defined in `src/renderer/components/app.scss`. These are basic browser resets like setting `box-sizing` property for every element, default text and background colors, default font size, basic headings visualisation etc.
```

src/renderer/components/vars.scss
```diff
@@ -5,17 +5,17 @@
...
-$font-weight-normal: 400;
-$font-weight-bold: 500;
+$font-main: var(--font-main);
+$font-monospace: var(--font-monospace);
+$font-size-small: var(--font-size-small);
+$font-size: var(--font-size);
+$font-size-big: var(--font-size-big);
+$font-weight-thin: var(--font-weight-thin);
+$font-weight-normal: var(--font-weight-normal);
+$font-weight-bold: var(--font-weight-bold);
```
ixrock, line 18: Do we still need all these `scss-var -> css-var` aliases?

ixrock (approved):
What if we make theming files as pure css-files instead of `json`?
Like this:
```
/**
  @file theme/lens-some-theme.css
  @name Lens Dark
  @type dark
  @description Original Lens dark theme
  @author Mirantis
*/

:root {
  --mainBackground: green;
  // ...
}
```

_Pros:_
- easier to navigate in project + refactor when needed
- easier to apply theme (turn on/off `style` element)
- all colors preview are shown in editor when file is opened (e.g. vscode, webstorm)
- comments inside `.css` are allowed

_Cons:_
- to show theme's metadata we would need to parse jsdoc (or similar) top comment

docs/extensions/capabilities/color-reference.md
```diff
@@ -0,0 +1,130 @@
+# Theme color reference
+You can use CSS variables generated from theme `.json` files to style an extension with respect of active theme.
+
+## Base colors
+- `blue`: blue color.
+- `magenta`: magenta color.
+- `golden`: gold/yellow color.
```
stefcameron, line 7: This reference is great! I would just suggest that you add the `--` prefix to all the names so it's more obvious that they are the CSS variable names, like `--blue` instead of just `blue`. And since you mention the "theme .json", why not link to the file in the code base? Otherwise, I would rephrase the top paragraph to be this:

"You can use theme-based CSS Variables to style an extension according to the active theme."

(i.e. leave-out the detail about where the variable names come from, since that's an implementation detail that an extension shouldn't have to worry about).

docs/extensions/capabilities/styling.md
```diff
@@ -1,8 +1,8 @@
-# Styling
-Lens provides a set of global styles and UI components that can be used by any extension to preserve look and feel of the application. However, it’s always up to the developer whether to use them or provide completely different styles for the extension.
+# Styling an extension
+Lens provides a set of global styles and UI components that can be used by any extension to preserve look and feel of the application.
 
 ## Styling approach
-Lens heavily uses SCSS preprocessor and it's advised to stick with same approach for extension developers in order to use some of the predefined variables and mixins described below.
+Lens heavily uses SCSS preprocessor with a set of predefined variables and mixins.
```
stefcameron, line 5: Since you've now moved everything pertinent to extensions into CSS variables, I would recommend removing this line. Extensions don't need to care about how Lens does things internally. They should only care about how they're supposed to do things, and what Lens does that will affect them (namely, the CSS variables available to them, as well as anything that would affect layout like `flex.box`, which you mention below).

src/renderer/components/vars.scss
stefcameron: I think so, since the values were moved into CSS Vars -- unless you SCSS code references the CSS Vars directly, but if that were the case, then I would imagine there would be many more files updated in this PR to move away from using the SCSS variable names.

docs/extensions/capabilities/styling.md
```diff
@@ -38,67 +36,64 @@ Every extention is affected by list of default global styles defined in `src/ren
...
+  "textColorSecondary": "#51575d",
+  "textColorAccent": "#333333",
+  "borderColor": "#c9cfd3",
+  "borderFaintColor": "#dfdfdf",
+  "mainBackground": "#f1f1f1",
+  "contentColor": "#ffffff",
+  "layoutBackground": "#e8e8e8",
+  "layoutTabsBackground": "#f8f8f8",
+  "layoutTabsActiveColor": "#333333",
+  "layoutTabsLineColor": "#87909c80"
```
stefcameron, line 82: I think we should add the `--` CSS var prefix to each one.

docs/extensions/capabilities/styling.md
```diff
@@ -1,8 +1,8 @@
-# Styling
-Lens provides a set of global styles and UI components that can be used by any extension to preserve look and feel of the application. However, it’s always up to the developer whether to use them or provide completely different styles for the extension.
+# Styling an extension
+Lens provides a set of global styles and UI components that can be used by any extension to preserve look and feel of the application.
 
 ## Styling approach
```
stefcameron, line 4: I would rename this section to "Layout" and move the new "Basic Styling" (which is mostly about layout) section into this "Layout" section. I would leave-out the fonts, though, since that's more theme than layout.

So maybe like this:
```markdown
## Layout

For layout tasks, Lens is using a combination of CSS Variables, and flex.box...
...
...However, feel free to use any styling technique or framework like [Emotion](https://github.com/emotion-js/emotion) or just plain CSS if you prefer.

### Layout Variables

(The content from "Basic Styling" excluding fonts.)
```

Then, we could have the Theme section, for colors and fonts (even though you're currently not changing fonts with the themes, but that's an implementation detail, IMO), where "Themable Colors" becomes "Theme Variables":

```markdown
## Layout

...(as above)...

## Themes

Lens is using 2 built-in themes located in `src/renderer/themes` folder each for light and dark color schemes. Active theme can be changed in the `Preferences` page...
...

### Theme Variables

After theme file gets parsed it provides list of theme-defined colors. Most of their values are different for light and dark themes. You can use them to preserve consistent view of extension with respect of selected theme...
...
Colors values are located inside `src/renderer/themes/lens-dark.json` and `src/renderer/themes/lens-light.json` files.
```

So the document outline would look like this:
```markdown
# Styling an extension

## Layout

### Layout Variables

## Themes

### Theme Variables

## Injected Styles

## CSS-in-JS
```

stefcameron (commented):
Nice improvements! Having that color reference, as well as the layout and font variable references, will be a __huge__ help to developers!

stefcameron:
@ixrock 
> What if we make theming files as pure css-files instead of json?

This is something to consider. It would add one less step of indirection between styles and their definition. But note that there are JSON parsers out there that support comments, so you could still add comments in your JSON file, albeit that may be difficult depending on what is parsing that JSON file (if you can customize the parser it uses).

JSON or CSS, however, I think the best thing here would be to colocate the definition of the variables with the variables themselves. IMO, the closer the docs are to the code, the more likely they are to remain accurate. So if we could take all those color variable explanations in `color-reference.md` and move them into JSON or CSS comments next to each variable they apply to, and then generate `color-reference.md` from this code, that would be a real win for everyone.

On a related note, since Lens is using CSS variables, this article might be interesting, particularly how Kent implements theme switching with `<body data-theme="light | dark">` and a matching CSS selector that defines the same set of variables, but with different colors. https://epicreact.dev/css-variables/

jakolehm:
Note to reviewers: let's improve things in follow-up PRs 🙂

PR #1278, opened 2020-11-09
ApiManager.registerStore should be simplified

### LINKED ISSUE #1275: ApiManager.registerStore should be simplified
It seems that we actually need just store as parameter (store already knows used api).

ixrock:
fix #1275


jakolehm (approved):
LGTM

PR #1303, opened 2020-11-10
Removing !important statements from theme variables

### LINKED ISSUE #1292: Global theme CSS Variable values should not be "!important"
image

The `!important` keyword in CSS can cause some real pain when trying to style things. In general, it should be avoided. If you have a layout that requires the use of "!important", save a few exceptions, you should rethink your layout to see why you have styles that are colliding with one another. It's best to work _with_ the cascading nature, and specificity of CSS rather than work against it.

If extensions are going to use CSS Variables for styling according to the Lens theme, the "!important" directive/keyword should be removed from ALL variable definitions. Otherwise, they will cause unintended/unexpected side effects when developers try to use them in their layouts.

aleksfront:
Also, fixing stale theme loading error (`kontena-dark` gets replaced by `lens-dark`)

Fixes #1292 

css variables

### DIFF
--- a/src/common/__tests__/user-store.test.ts
+++ b/src/common/__tests__/user-store.test.ts
@@ -57,11 +57,12 @@ describe("user store tests", () => {
       expect(us.preferences.colorTheme).toBe('light')
     })
 
-    it("correctly resets theme to default value", () => {
+    it("correctly resets theme to default value", async () => {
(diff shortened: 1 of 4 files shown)


nevalla (approved):
LGTM

stefcameron:
Thank you! 😄

PR #1307, opened 2020-11-10
Generate extension api reference docs using library mode

jakolehm:
image


package.json
```diff
@@ -38,8 +38,8 @@
     "download:helm": "yarn run ts-node build/download_helm.ts",
     "build:tray-icons": "yarn run ts-node build/build_tray_icon.ts",
     "lint": "eslint $@ --ext js,ts,tsx --max-warnings=0  src/",
-    "mkdocs-serve-local": "yarn typedocs-extensions-api && docker build -t mkdocs-serve-local:latest mkdocs/ && docker run --rm -it -p 8000:8000 -v ${PWD}:/docs mkdocs-serve-local:latest",
-    "typedocs-extensions-api": "npx typedoc --plugin typedoc-plugin-markdown --readme none --name @k8slens/extensions --ignoreCompilerErrors --out docs/extensions/api --mode modules --excludePrivate --includes src/ src/extensions/extension-api.ts"
+    "mkdocs-serve-local": "docker build -t mkdocs-serve-local:latest mkdocs/ && docker run --rm -it -p 8000:8000 -v ${PWD}:/docs mkdocs-serve-local:latest",
+    "typedocs-extensions-api": "yarn typedoc --ignoreCompilerErrors --readme docs/extensions/typedoc-readme.md.tpl --name @k8slens/extensions --out docs/extensions/api --mode library --excludePrivate --hideBreadcrumbs --includes src/ src/extensions/extension-api.ts"
```
Nokel81, line 42: Do we really want to `ignoreCompilerErrors`?

package.json
```diff
@@ -381,7 +381,7 @@
     "ts-loader": "^7.0.5",
     "ts-node": "^8.10.2",
     "type-fest": "^0.18.0",
-    "typedoc": "^0.19.2",
+    "typedoc": "0.17.0-3",
```
Nokel81, line 384: Why the downgrade?

package.json
jakolehm: This PR does not change `ignoreCompilerErrors` behaviour.

package.json
jakolehm: Library mode is only supported in this specific version. I'm still investigating if we can get similar results with newer versions.

package.json
Nokel81: I see, still odd though.

package.json
Nokel81: Fair (weird) but fair.

Nokel81 (approved):

yarn.lock
```diff
@@ -3412,6 +3412,13 @@ babel-runtime@^6.26.0:
     core-js "^2.4.0"
     regenerator-runtime "^0.11.0"
 
+backbone@^1.4.0:
```
nevalla, line 3415: ❤️

nevalla (approved):
LGTM

package.json
jakolehm: For some reason it cannot resolve symbols like `__static` -> causing errors.

jakolehm:
Had to downgrade typedoc markdown plugin (latest version did not produce good results with older typedoc).

@nevalla PTAL again

nevalla (approved):

PR #1312, opened 2020-11-10
Remove Makefile from all intree extensions

### LINKED ISSUE #1271: Remove unnecessary Makefiles from in-tree extensions

Nokel81:
- Makefiles are not really common (rightly so) in JS projects and script
  entries in the package.json are good enough and what they're there
  for.

- Move to using yarn for installing dependencies and running scripts as
  that is what we mostly use for the rest of the project.

- Move from using make syntax to a dedicated script for building and
  testing extensions.

fixes #1271


extensions/example-extension/package.json
```diff
@@ -11,7 +11,7 @@
   "scripts": {
     "build": "webpack --config webpack.config.js",
     "dev": "npm run build --watch",
-    "test": "echo NO TESTS"
+    "test": "jest --passWithNoTests --env=jsdom src $@"
```
jim-docker, line 14: add jest to devDependencies?

Makefile
```diff
@@ -56,10 +56,10 @@ else
 endif
 
 build-extensions:
-	$(foreach dir, $(wildcard $(EXTENSIONS_DIR)/*), $(MAKE) -C $(dir) build;)
+	yarn intree-extensions $(EXTENSIONS_DIR) build
```
jim-docker: how does yarn find intree-extensions?

scripts/intree-extensions.ts
```diff
@@ -0,0 +1,94 @@
...
+  console.log(`Testing: ${extensionPath}`)
+
+  const test = execa("yarn", ["test"], { cwd: extensionPath })
+  test.stdout.pipe(makeSink(prefix))
+  await test
+}
+
+async function main() {
+  const cwd = process.cwd()
+  const [extensionsFolder, action] = args
```
jim-docker: should you test if these are empty? (user could run `yarn intree-extensions` manually?)

scripts/intree-extensions.ts
```diff
@@ -0,0 +1,94 @@
...
+  const prefixes = names.map(name => `${name}:${" ".repeat(prefixWidth - name.length)}`)
+
+  switch (action.toLowerCase()) {
+    case "build":
+      await Promise.all(folders.map((folder, i) => buildInTreeExtension(folder, prefixes[i])))
+      break
+    case "test":
+      await Promise.all(folders.map((folder, i) => testInTreeExtension(folder, prefixes[i])))
+      break
+  }
```
jim-docker: `default:` ?

extensions/example-extension/package.json
Nokel81: Fair point

Makefile
Nokel81: `yarn <verb>` means roughly the same as `yarn run <verb>` if `<verb>` is not a predefined action in `yarn`. So it looks at the "scripts" object in the `package.json` file at the root of this repo.

scripts/intree-extensions.ts
Nokel81: Probably, good catch.

scripts/intree-extensions.ts
Nokel81: How does log an error and exit sound?

Makefile
jim-docker: ah, I missed your change to package.json

jakolehm: My initial idea was just to replace this foreach with a call to npm ... is that a bad idea / why we did implement our own logic (which kinda reinvents a wheel)?

I mean something like:
```
$(foreach dir, $(wildcard $(EXTENSIONS_DIR)/*), cd $(dir) &&  npm install && npm build;)
```

scripts/intree-extensions.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+  })
+}
+
+const args = process.argv.slice(2) // remove node call and file name call
+
+async function buildInTreeExtension(extensionPath: string, prefix: string) {
+  console.log(`Building: ${extensionPath}`)
+
+  const install = execa("yarn", { cwd: extensionPath })
```
jakolehm: We should not use yarn because Lens installs extensions using `npm install`, right?

Makefile
Nokel81: There is nothing wrong with that. Reverted.

extensions/example-extension/main.ts
```diff
@@ -2,10 +2,10 @@ import { LensMainExtension } from "@k8slens/extensions";
 
 export default class ExampleExtensionMain extends LensMainExtension {
   onActivate() {
-    console.log('EXAMPLE EXTENSION MAIN: ACTIVATED', this.getMeta());
+    console.log('EXAMPLE EXTENSION MAIN: ACTIVATED', this.name, this.id);
```
jim-docker, line 5: this means the example-extension is broken in master(?) Since #1208 I think, where `getMeta()` was removed

Makefile
```diff
@@ -56,10 +56,10 @@ else
 endif
 
 build-extensions:
-	$(foreach dir, $(wildcard $(EXTENSIONS_DIR)/*), $(MAKE) -C $(dir) build;)
+	$(foreach dir, $(wildcard $(EXTENSIONS_DIR)/*), (cd $(dir) && npm install && npm run build || exit);)
```
jim-docker: that's a lot less work :)

jim-docker (approved):

package.json
```diff
@@ -340,6 +340,7 @@
     "electron-builder": "^22.7.0",
     "electron-notarize": "^0.3.0",
     "eslint": "^7.7.0",
+    "execa": "^4.1.0",
```
jakolehm: Not needed anymore?

extensions/example-extension/main.ts
Nokel81: Yup, this is why we should build all intree extensions on every PR (this PR adds this) and why I think (at least while developing) that the example extension should also be loaded...

package.json
Nokel81: No, good catch (though it might be useful in the future since it makes working with `child_process` a lot easier

PR #1322, opened 2020-11-11
Expose registration and related component interfaces to extensions

nevalla:
This PR will expose registration and related component interfaces to extensions. This is mainly needed for generated api reference.


src/extensions/interfaces/registrations.ts
```diff
@@ -0,0 +1,7 @@
+export type { AppPreferenceRegistration, AppPreferenceComponents } from "../registries/app-preference-registry"
```
nevalla, line 1: not sure do we need `type` here since those are only interfaces?

jakolehm (approved):
LGTM

PR #1323, opened 2020-11-11
Remove windowManager from extension api

### LINKED ISSUE #1320: WindowManager should not be exposed via extension-api

jakolehm:
Fixes #1320


ixrock (approved):
LGTM

src/main/tray.ts
```diff
@@ -16,7 +16,7 @@ import { isDevelopment } from "../common/vars";
 export let tray: Tray;
 
 // refresh icon when MacOS dark/light theme has changed
-nativeTheme.on("updated", () => tray?.setImage(getTrayIcon()));
+nativeTheme?.on("updated", () => tray?.setImage(getTrayIcon()));
```
ixrock, line 19: `nativeTheme` not available in some OS-es?

jakolehm: Now that windowManager is exposed to both main/renderer this needs to be protected, `nativeTheme` is undefined in renderer :/

ixrock: Then probably better to move this binding to `initTray(windowManager: WindowManager)`

extensions/support-page/main.ts
```diff
@@ -7,7 +7,7 @@ export default class SupportPageMainExtension extends LensMainExtension {
       parentId: "help",
       label: "Support",
       click() {
-        windowManager.navigate(supportPageURL());
+        Navigation.navigate(supportPageURL());
```
jim-docker: I'm wondering if it's possible to wrap this as a method in the `LensExtension` class so this line could be
```suggestion
        this.navigate(supportPageURL());
```
and no need to import `Navigation` (and maybe even don't need to expose it in the extensions api?)

jim-docker: We could even have a `LensExtension` `navigate()` method that accepts a page id (which I guess is essentially the `PageRegistration` `path` field) I think it would be very nice to eliminate the need for setting up a route (i.e. `supportPageURL from "./src/support.route"). Though I think it is actually optional now, right?

src/renderer/navigation.ts
```diff
@@ -3,13 +3,24 @@
...
 import logger from "../main/logger";
 import { clusterViewRoute, IClusterViewRouteParams } from "./components/cluster-manager/cluster-view.route";
 
-export const history = typeof window !== "undefined" ? createBrowserHistory() : createMemoryHistory();
-export const navigation = createObservableHistory(history);
+let history: History
+let navigation: IObservableHistory
+
+if (ipcRenderer) {
+  history = typeof window !== "undefined" ? createBrowserHistory() : createMemoryHistory();
```
ixrock: Btw probably we don't need check for `window` anymore? Is is used in `main` somehow?

extensions/support-page/main.ts
jakolehm: Yes, eventually I had to do like you suggested .. otherwise prod build fails.

src/renderer/navigation.ts
ixrock: and we can remove `createMemoryHistory()`?

jakolehm: Right, I'll cleanup more.

jakolehm: Done.

src/extensions/lens-main-extension.ts
```diff
@@ -1,7 +1,12 @@
 import type { MenuRegistration } from "./registries/menu-registry";
 import { observable } from "mobx";
 import { LensExtension } from "./lens-extension"
+import { WindowManager } from "../main/window-manager";
 
 export class LensMainExtension extends LensExtension {
   @observable.shallow appMenus: MenuRegistration[] = []
+
+  async navigate(location: string, frameId?: number) {
+    await WindowManager.getInstance<WindowManager>().navigate(location, frameId)
```
jim-docker, line 10: whew, nice

src/extensions/lens-main-extension.ts
```diff
@@ -1,7 +1,12 @@
 import type { MenuRegistration } from "./registries/menu-registry";
 import { observable } from "mobx";
 import { LensExtension } from "./lens-extension"
+import { WindowManager } from "../main/window-manager";
 
 export class LensMainExtension extends LensExtension {
   @observable.shallow appMenus: MenuRegistration[] = []
+
+  async navigate(location: string, frameId?: number) {
```
ixrock, line 9: Should we add similar method to `lens-renderer-extension.ts`?

jim-docker (approved):

src/extensions/lens-main-extension.ts
jakolehm: Yes, done. I still left `Navigation.navigate` because that's handy in React components.

PR #1327, opened 2020-11-11
Fix: allow to fetch kube-api objects with label & field selectors

ixrock:
_Example:_

```ts
  const pods = await podsApi.list({}, {
    labelSelector: [
      "k8s-app=kube-dns",
      "pod-template-hash=66bff467f8"
    ]
  });
```


jakolehm (approved):
LGTM

PR #1328, opened 2020-11-11
Refactoring: removed kube versioned api

ixrock:


jakolehm:
@nevalla PTAL

src/renderer/api/kube-api.ts
```diff
@@ -86,15 +93,36 @@ export class KubeApi<T extends KubeObject = any> {
...
 
+  get apiVersionWithGroup() {
+    return [this.apiGroup, this.apiVersionPreferred ?? this.apiVersion]
+      .filter(Boolean)
+      .join("/")
+  }
+
+  protected async checkPreferredVersion() {
+    if (!this.options.checkPreferredVersion || this.apiVersionPreferred === undefined) {
+      return;
```
nevalla: isn't this always returning since we don't update `apiVersionPreferred` elsewhere?

ixrock: Oh, right. Have to tweak `if` condition..

jakolehm (approved):
LGTM

src/renderer/api/kube-api.ts
```diff
@@ -86,15 +93,35 @@ export class KubeApi<T extends KubeObject = any> {
...
+      .filter(Boolean)
+      .join("/")
+  }
+
+  protected async checkPreferredVersion() {
+    if (this.options.checkPreferredVersion && this.apiVersionPreferred === undefined) {
+      const res = await this.request.get<IKubePreferredVersion>(`${this.apiPrefix}/${this.apiGroup}`);
+      Object.defineProperty(this, "apiVersionPreferred", {
+        value: res?.preferredVersion?.version ?? null,
+      });
```
Nokel81, line 116: Just looking through PRs to know what has changed.

May I ask why this is required instead of `this.apiVersionPreferred = res?.preferredVersion?.version ?? null;`?

PR #1330, opened 2020-11-11
Allow extensions to embed extra kube object details

### LINKED ISSUE #1329: Allow extensions to register additional details for kubernetes objects
Currently extensions only allow to register whole details content area (for custom resources). Sometimes it would be nice to display additional information directly in kubernetes object details (information would be in right context).

jakolehm:
Fixes #1329


jakolehm:
It seems that we don't really need this.

PR #1331, opened 2020-11-11
Removing parental element styles that affect extensions

### LINKED ISSUE #1251: Isolate global styles from extensions

aleksfront:
Removing annoying parent styles that affect `Cluster Page` and `Bottom Bar` extensions.

Outer styles for another registires kept untouched since for some places like menu items outer styling is preferred way to keep things simple for extension developers. For example, if some extension adding a menu item to pod actions menu, new item visualisation most probably would look alike other items. So there's no need for developer to style whole look of a menu item.

Also, moved styling of `Support Page` extension to [emotion](https://emotion.sh/docs/introduction) styled components considered it as a prefered way to isolate custom styles from ones provided by Lens.

Fixes #1251


extensions/support-page/renderer.tsx
```diff
@@ -18,12 +26,14 @@ export default class SupportPageRendererExtension extends LensRendererExtension
   statusBarItems = [
     {
       item: (
-        <div
-          className="flex align-center gaps hover-highlight"
-          onClick={() => Navigation.navigate(supportPageURL())}
-        >
-          <Component.Icon material="help" smallest />
-        </div>
+        <Icon>
```
ixrock: Looks ugly. Let's find webpack plugin to convert all class-names to some hashes to avoid collisions (instead of `styled` components)?

aleksfront: It's not ugly, it's just a matter of habit imho.

aleksfront: Hovewer, reverted to scss solution since it's inner plugin.

ixrock: I mean this is ugly `Icon > Component.Icon`

stefcameron (approved):
@aleksfront Thank you for working on this issue! I'm sure it'll improve the DX for extensions.

IMO, it's a little strong to state that this PR "removes global styles that affect extensions", however, because https://github.com/lensapp/lens/blob/master/src/renderer/components/app.scss still contains a lot of styles that affect the entire app, including all types of extensions, because of all the element-based selectors. It does fix #1251 , though. 🎉 

The redeeming factor, however, is that they are element-based selectors, which means they get the lowest CSS specificity values and are easily overridden. So if extension authors don't like the default, they don't have to automatically resort to `!important` on everything to break out of them.

As long as new changes continue to make use of CSS variables, and they get documented, and refrain from adding global styles unless it really makes sense, the DX of implementing extensions should remain very good in terms of styling.

aleksfront:
@stefcameron Totally agree. The previous statement was wrong, I've changed name of the PR.

ixrock (approved):

PR #1334, opened 2020-11-11
PageRegistration & BaseRegistry refactoring

### LINKED ISSUE #1258: Move navigation/menus registration from PageRegistration to own registry
Now it's embedded into a global page registration. Would be clearer if a menu item would be a separated api.

ixrock:
_Changes:_

- Registering menu items (ClusterManager/Cluster) moved away to separated registry from `PageRegistry`
- Refactored `BaseRegistry` so now all extension pages registered under `/extension/:name` route prefix
- Refactored `TabLayout`

close #1258 

_Fix for hello-world example:_ https://github.com/lensapp/hello-world-extension

```ts
import { Component, Interface, LensRendererExtension } from "@k8slens/extensions";
import { ExampleIcon, ExamplePage } from "./page"
import React from "react"

  clusterPages = [
    {
      id: Symbol("example-extension"),
      components: {
      }
    }
  ]

  clusterPageMenus = [
    {
      id: this.clusterPages[0].id,
      title: "Hello World",
      components: {
        Icon: ExampleIcon
      }
    }
  ]
}
```


package.json
```diff
@@ -241,6 +241,7 @@
     "openid-client": "^3.15.2",
     "path-to-regexp": "^6.1.0",
     "proper-lockfile": "^4.1.1",
+    "react-router": "^5.2.0",
```
ixrock: This is moved to normal dependencies since required in runtime (`matchPath` is used in `page-registry.ts` and `page-menu-registry.ts`)

extensions/support-page/main.ts
```diff
@@ -1,13 +1,13 @@
...
 
 export default class SupportPageMainExtension extends LensMainExtension {
   appMenus = [
     {
       parentId: "help",
       label: "Support",
-      click() {
-        windowManager.navigate(supportPageURL());
+      click: () => {
+        windowManager.navigate(this.getPageUrl(pageUrl)); // todo: simplify
```
jim-docker: Did you see my comment [here](https://github.com/lensapp/lens/pull/1323#discussion_r521352669)? Maybe this is similar to what you are already planning...

ixrock: 1. Page registration doesn't have ID
1. Registered page exists only in `LensRendererExtension` but access to URL might need from `LensMainExtension` as well (like from top-menu support item)
1. If even this possible should be done in another PR

package.json
ixrock: `react` apparently also required internally in `react-router` when used in runtime

extensions/support-page/renderer.tsx
```diff
@@ -20,9 +18,9 @@ export default class SupportPageRendererExtension extends LensRendererExtension
       item: (
         <div
           className="flex align-center gaps hover-highlight"
-          onClick={() => Navigation.navigate(supportPageURL())}
+          onClick={() => Navigation.navigate(this.getPageUrl(pageUrl))}
```
jim-docker: If you instead have a `navigate()` method on `LensRendererExtension` (or `LensExtension`?) then this could be 
```suggestion
          onClick={() => this.navigate(pageUrl)}
```
(the `navigate()` method would call `this.getPageUrl()`, no need to expose it. This aligns with what @jakolehm did in https://github.com/lensapp/lens/pull/1323

docs/extensions/capabilities/common-capabilities.md
```diff
@@ -100,13 +100,22 @@ import { ExamplePage } from "./src/example-page"
 export default class ExampleRendererExtension extends LensRendererExtension {
   globalPages = [
     {
-      path: "/example-route",
-      hideInMenu: true,
+      routePath: "/items/:id?",
```
jim-docker: Seems cumbersome if the extension developer has to specify this. Can it default to this?

docs/extensions/capabilities/common-capabilities.md
```diff
@@ -146,11 +155,20 @@ import { ExampleIcon, ExamplePage } from "./src/page"
 export default class ExampleExtension extends LensRendererExtension {
   clusterPages = [
     {
-      path: "/extension-example",
-      title: "Example Extension",
+      routePath: "/extension-example",
+      exact: true,
```
jim-docker: can default be `true`? So developer doesn't have to set it unless they want it `false` (which assumes they know what they are doing)

extensions/support-page/renderer.tsx
jim-docker: guess that didn't work out. Would be nice if it were possible ...

src/extensions/lens-main-extension.ts
```diff
@@ -7,6 +7,8 @@ export class LensMainExtension extends LensExtension {
   @observable.shallow appMenus: MenuRegistration[] = []
 
   async navigate(location: string, frameId?: number) {
-    await WindowManager.getInstance<WindowManager>().navigate(location, frameId)
+    const windowManager = WindowManager.getInstance<WindowManager>();
+    const url = this.getPageUrl(location);
+    await windowManager.navigate(url, frameId)
   }
```
jim-docker: I like this a lot but @jakolehm couldn't get it to work?

src/renderer/components/layout/sidebar.tsx
```diff
@@ -98,108 +97,100 @@ export class Sidebar extends React.Component<Props> {
           </div>
           <div className="sidebar-nav flex column box grow-fixed">
             <SidebarNavItem
-              id="cluster"
```
jim-docker, line 101: I think the integration tests rely on these ids

extensions/support-page/renderer.tsx
jakolehm: See #1345

src/extensions/lens-main-extension.ts
jakolehm: > I like this a lot but @jakolehm couldn't get it to work?

It works on main extension.

src/renderer/components/layout/sidebar.tsx
jakolehm: @ixrock pls revert removal of ids.

jakolehm (changes requested):

src/extensions/registries/base-registry.ts
```diff
@@ -1,23 +1,47 @@
...
   }
 
   @action
-  remove(...items: T[]) {
-    items.forEach(item => {
-      this.items.remove(item); // works because of {deep: false};
-    })
+  remove(items: T[], key: LensExtension = null) {
+    const storedItems = this.items.get(key);
+    if (storedItems) {
```
nevalla: could be also be

```
if (!storedItems) return;
```

extensions/support-page/renderer.tsx
ixrock: Worked in dev/local build :/

docs/extensions/capabilities/common-capabilities.md
ixrock: It's just an example. Now could be optional since we have route-prefix for all extensions.

docs/extensions/capabilities/common-capabilities.md
ixrock: This is optional. Could be default whatever we want :)

src/renderer/components/layout/sidebar.tsx
ixrock: This is bad.

nevalla: yes it's bad, but let's fix that in separate PR

src/extensions/registries/page-menu-registry.ts
```diff
@@ -0,0 +1,34 @@
...
+import type React from "react";
+import type { IconProps } from "../../renderer/components/icon";
+import { BaseRegistry, BaseRegistryItem, BaseRegistryItemId } from "./base-registry";
+
+export interface PageMenuRegistration extends BaseRegistryItem {
+  id: BaseRegistryItemId; // required id from page-registry item to match with
+  url?: string; // when not provided initial extension's path used, e.g. "/extension/lens-extension-name"
+  title: React.ReactNode;
+  components: PageMenuComponents;
+  subMenus?: PageSubMenuRegistration[];
```
jakolehm, line 12: How does this work with global pages?

src/extensions/registries/page-registry.ts
```diff
@@ -1,31 +1,33 @@
...
 
-export interface PageRegistration extends RouteProps {
-  className?: IClassName;
-  url?: string; // initial url to be used for building menus and tabs, otherwise "path" applied by default
-  title?: React.ReactNode; // used in sidebar's & tabs-layout if provided
-  hideInMenu?: boolean; // hide element within app's navigation menu
-  subPages?: (PageRegistration & TabRoute)[];
+export interface PageRegistration extends BaseRegistryItem {
+  routePath?: string; // additional (suffix) route path to base extension's route: "/extension/:name"
+  exact?: boolean; // route matching flag, see: https://reactrouter.com/web/api/NavLink/exact-bool
```
jakolehm, line 8: This is a bit mystery flag ... when does it needs to be set?

src/extensions/registries/page-menu-registry.ts
ixrock: it doesn't :D

src/extensions/registries/page-registry.ts
ixrock: I was hoping provided url from comment will explain this.. but.. it's about strict matching route:
e.g. if you have route like `/page` with `exact: true` it will match only this, but not `/page/sub-page`

src/extensions/registries/page-menu-registry.ts
ixrock: fixed

nevalla:
Something is off with the support icon/extension. Cannot click it.
image

ixrock:
@nevalla you should recompile support-page extension (also got this after master-merge)
This is comes from https://github.com/lensapp/lens/pull/1331

jakolehm (approved):

PR #1335, opened 2020-11-11
Allow to specify KubeObjectDetailRegistration priority

jakolehm:
Allows extensions (or core) to specify placement of object details. For example:

```ts
  kubeObjectDetailItems = [
    {
      kind: "Pod",
      apiVersions: ["v1"],
      priority: 10,
      components: {
        Details: (props: any) => <CustomPodDetails pod={props.object} />
      }
    }
  ]
```
Would be shown above Pod events.

### DIFF
--- a/src/renderer/components/+network-ingresses/ingress-details.tsx
+++ b/src/renderer/components/+network-ingresses/ingress-details.tsx
@@ -127,8 +127,6 @@ export class IngressDetails extends React.Component<Props> {
 
         <DrawerTitle title={<Trans>Load-Balancer Ingress Points</Trans>}/>
         {this.renderIngressPoints(ingressPoints)}
-
-        <KubeEventDetails object={ingress}/>
       </div>
     )
   }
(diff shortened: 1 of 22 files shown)


src/renderer/components/+network-ingresses/ingress-details.tsx
```diff
@@ -141,3 +139,11 @@ kubeObjectDetailRegistry.add({
     Details: (props) => <IngressDetails {...props} />
   }
 })
+kubeObjectDetailRegistry.add({
+  kind: "Ingress",
+  apiVersions: ["extensions/v1beta1"],
```
nevalla, line 144: should we register also `networking.k8s.io/v1`?

jakolehm: Isn't that out-of-scope of this PR?

nevalla: probably since we are using `extensions/v1beta1` in api too.

jakolehm: `networking.k8s.io/v1` should be done in a separate PR, it most likely will affect also how UI should be rendered.

nevalla (approved):
LGTM

src/renderer/components/+network-ingresses/ingress-details.tsx
jakolehm: Created an issue: https://github.com/lensapp/lens/issues/1347

PR #1355, opened 2020-11-12
Documentation Update: Properly create tags and aliases, add workflow …

msa0311:
Documentation Update: 
- Properly create tags and aliases
  - create new tag and update alias "latest" on release publish (for release and pre-release until Lens 4.0.0 is GA, see #1408)
  - master contains always the docs generated from last master branch commit
- add workflow to delete old unwanted versions from gh-pages branch using mkdocs delete

Documentation delete Version GH-Action

Signed-off-by: Mario Sarcher <msarcher@mirantis.com>

### DIFF
--- a/.github/workflows/mkdocs-delete-version.yml
+++ b/.github/workflows/mkdocs-delete-version.yml
@@ -1,0 +1,38 @@
+name: Delete Documentation Version
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version string to be deleted (e.g."v0.0.1")'
+        required: true
+jobs:
+  build:
+    name: Delete docs Version
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install git+https://${{ secrets.GH_TOKEN }}@github.com/lensapp/mkdocs-material-insiders.git
+          pip install mike
+          
+      - name: Checkout Release from lens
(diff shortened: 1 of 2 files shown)


.github/workflows/mkdocs-delete-version.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install git+https://${{ secrets.GH_TOKEN }}@github.com/lensapp/mkdocs-material-insiders.git
+          pip install mike
+          
+      - name: Checkout Release from lens
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
stevejr, line 27: @msa0311 - shouldn't we check out the tag here rather than master?

msa0311: As far as I understood how actions/checkout@v2 works it uses github.ref as default which should always be the correct tag / branch. 

https://github.com/actions/checkout

```
   # The branch, tag or SHA to checkout. When checking out the repository that
    # triggered a workflow, this defaults to the reference or SHA for that event.
    # Otherwise, uses the default branch.
    ref: ''
```

stevejr: Yes but here the branch is not triggering the action as its manual - the user is is supplying the tag and as we do not specify anything in the check I assume it will default to master/main?

jakolehm (approved):

.github/workflows/mkdocs-delete-version.yml
stevejr: Can review later if needed

stevejr (approved):
LGTM

PR #1361, opened 2020-11-12
Fix: navigating to extension page

ixrock:
Proper handling extension page base route.

_Fixes:_
- extension page route matches to actual extension name, e.g. `/extension/lens-hello-world`
- extension name updated in accordance to be supported in URL (`@` removed, `/` => `-`)

### DIFF
--- a/src/extensions/lens-extension.ts
+++ b/src/extensions/lens-extension.ts
@@ -44,13 +44,9 @@ export class LensExtension {
     return this.manifest.description
   }
 
-  getPageUrl(baseUrl = "") {
+  getPageRoute(suffixPathOrUrl = "") {
     const validUrlName = this.name.replace("@", "").replace("/", "-");
-    return compile(this.routePrefix)({ name: validUrlName }) + baseUrl;
-  }
-
-  getPageRoute(baseRoute = "") {
-    return this.getPageUrl() + baseRoute;
+    return compile(this.routePrefix)({ name: validUrlName }) + suffixPathOrUrl;
   }
 
   @action
--- a/src/extensions/lens-extension.ts
+++ b/src/extensions/lens-extension.ts
@@ -49,7 +49,7 @@ export class LensExtension {
   }
 
   getPageRoute(baseRoute = "") {
-    return this.routePrefix + baseRoute;
+    return this.getPageUrl() + baseRoute;
   }
 
   @action
--- a/src/extensions/lens-extension.ts
+++ b/src/extensions/lens-extension.ts
@@ -45,7 +45,8 @@ export class LensExtension {
   }
 
   getPageUrl(baseUrl = "") {
-    return compile(this.routePrefix)({ name: this.name }) + baseUrl;
+    const validUrlName = this.name.replace("@", "").replace("/", "-");
+    return compile(this.routePrefix)({ name: validUrlName }) + baseUrl;
   }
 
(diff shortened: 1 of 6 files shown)


src/extensions/lens-extension.ts
```diff
@@ -45,11 +45,12 @@ export class LensExtension {
   }
 
   getPageUrl(baseUrl = "") {
-    return compile(this.routePrefix)({ name: this.name }) + baseUrl;
+    const validUrlName = this.name.replace("@", "").replace("/", "-");
```
jim-docker, line 48: Is there a uniqueness requirement for the name? This would have to be part of that because `@foo` would clash with one already named `foo`

ixrock: It won't clash since when package name with `@` the format is `@org/package-name`, not just `@some-name`

stefcameron: I think we should be OK because I don't think `@fred` is a valid package name on NPM (and hopefully not on other registries like GH). You'd have to have a package name too, like `@fred/fred`, which would become `fred-fred` here, and wouldn't conflict with a package published as just `fred` (without a scope).

jim-docker: I can see it is unlikely but it's theoretically possible to have mirantis-foo and @mirantis/foo

jim-docker (approved):
tried it out with lens-extension-cc repo, LGTM

stefcameron:
> tried it out with lens-extension-cc repo, LGTM

Thanks for doing that, @jim-docker ! Glad to know it will work. 😄

src/extensions/lens-extension.ts
jakolehm: What's the difference now between `getPageUrl` and `getPageRoute`? To me they look identical 🤔

ixrock: @jakolehm you've asked already similar question just some days ago (page-registry `path` vs `url`) 😂
`getPageRoute` contain route placeholders like `:name`, etc.

jakolehm: @ixrock `getPageRoute` calls `getPageUrl` and just appends `baseRoute` to it. Isn't that exactly what happens if `getPageUrl(baseRoute)` is called??

ixrock: They are the same when no additional (suffix) route provided for the page:
```ts
export class PageRegistry<T extends PageRegistration> extends BaseRegistry<T> {
  getItems() {
    return super.getItems().map(item => {
      item.routePath = item.extension.getPageRoute(item.routePath)
      return item
    });
  }
}
```

ixrock: This is `baseRoute` of page-item, not `baseRoute` of extension.

nevalla: to me it looks like `this.getPageUrl('foo')` returns the same as `this.getPageRoute('foo')`

ixrock: Oh, now I see what you mean.. Both functions are similar but for different purposes..
Will refactor..

ixrock: @jakolehm PTAL, also fixed proper sub-pages route-paths and sub-menus urls.

ixrock:
Closing in favour https://github.com/lensapp/lens/pull/1364

PR #1368, opened 2020-11-13
Add posibility to add custom helm repository

pashevskii:
The commit adds the possibility to add a custom helm repository instead of selecting prepared ones
The commit adds new button in preferences - helm section
image

Warning! Implementation is ugly, need your comments on how to improve it.

### DIFF
--- a/src/main/helm/helm-repo-manager.ts
+++ b/src/main/helm/helm-repo-manager.ts
@@ -22,7 +22,7 @@ export interface HelmRepo {
   cacheFilePath?: string
   caFile?: string,
   certFile?: string,
-  insecure_skip_tls_verify?: boolean,
+  insecureSkipTlsVerify?: boolean,
   keyFile?: string,
   username?: string,
   password?: string,
@@ -125,6 +125,25 @@ export class HelmRepoManager extends Singleton {
     logger.info(`[HELM]: adding repo "${name}" from ${url}`);
     const helm = await helmCli.binaryPath();
     const { stdout } = await promiseExec(`"${helm}" repo add ${name} ${url}`).catch((error) => {
+      throw(error.stderr);
+    });
+
+    return stdout;
+  }
+
+  public async addСustomRepo(repoAttributes : HelmRepo) {
+    logger.info(`[HELM]: adding repo "${repoAttributes.name}" from ${repoAttributes.url}`);
+    const helm = await helmCli.binaryPath();
+
+    const insecureSkipTlsVerify = repoAttributes.insecureSkipTlsVerify ? " --insecure-skip-tls-verify" : "";
+    const username = repoAttributes.username ? ` --username "${repoAttributes.username}"` : "";
+    const password = repoAttributes.password ? ` --password "${repoAttributes.password}"` : "";
+    const caFile = repoAttributes.caFile ? ` --ca-file "${repoAttributes.caFile}"` : "";
+    const keyFile = repoAttributes.keyFile ? ` --key-file "${repoAttributes.keyFile}"` : "";
+    const certFile = repoAttributes.certFile ? ` --cert-file "${repoAttributes.certFile}"` : "";
+
+    const addRepoCommand = `"${helm}" repo add ${repoAttributes.name} ${repoAttributes.url}${insecureSkipTlsVerify}${username}${password}${caFile}${keyFile}${certFile}`;
+    const { stdout } = await promiseExec(addRepoCommand).catch((error) => {
       throw(error.stderr);
     });
 
--- a/src/renderer/components/+preferences/add-helm-repo-dialog.scss
+++ b/src/renderer/components/+preferences/add-helm-repo-dialog.scss
@@ -1,0 +1,22 @@
+.AddHelmRepoDialog {
+  --flex-gap: #{$margin * 1.5};
+
+  .accordion {
+    font-weight: bold;
+    align-self: flex-start;
+    padding: 0;
+    &:focus:not(:active) {
+      box-shadow: none;
+    }
+  }
+  .Icon {
+    --color-active: black;
+  }
+
+  .fields-title {
+    display: grid;
+    grid-template-columns: min-content auto;
+    align-items: center;
+    gap: $margin / 2;
+  }
+}
--- a/src/renderer/components/+preferences/add-helm-repo-dialog.tsx
+++ b/src/renderer/components/+preferences/add-helm-repo-dialog.tsx
@@ -1,0 +1,178 @@
+import "./add-helm-repo-dialog.scss";
+
+import React from "react";
+import { remote, FileFilter } from "electron";
+import { observable } from "mobx";
+import { observer } from "mobx-react";
+import { t, Trans } from "@lingui/macro";
+import { _i18n } from "../../i18n";
+import { Dialog, DialogProps } from "../dialog";
+import { Wizard, WizardStep } from "../wizard";
+import { Input } from "../input";
+import { Checkbox } from "../checkbox";
+import { Button } from "../button";
+import { systemName, isUrl, isPath } from "../input/input_validators";
+import { SubTitle } from "../layout/sub-title";
+import { Icon } from "../icon";
+import { Notifications } from "../notifications";
+import { HelmRepo, repoManager } from "../../../main/helm/helm-repo-manager";
+
+interface Props extends Partial<DialogProps> {
+  onAddRepo: Function
+}
+
+enum FileType {
+  CaFile = "caFile",
+  KeyFile = "keyFile",
+  CertFile = "certFile",
+}
+
+@observer
+export class AddHelmRepoDialog extends React.Component<Props> {
+  private emptyRepo = {name: "", url: "", username: "", password: "", insecureSkipTlsVerify: false, caFile:"", keyFile: "", certFile: ""};
+
+  private static keyExtensions = ["key", "keystore", "jks", "p12", "pfx", "pem"];
+  private static certExtensions = ["crt", "cer", "ca-bundle", "p7b", "p7c" , "p7s", "p12", "pfx", "pem"];
+
+  @observable static isOpen = false;
+
+  static open() {
+    AddHelmRepoDialog.isOpen = true;
+  }
+
+  static close() {
+    AddHelmRepoDialog.isOpen = false;
+  }
+
+  @observable helmRepo : HelmRepo = this.emptyRepo;
+  @observable showOptions = false;
+
+  close = () => {
+    AddHelmRepoDialog.close();
+    this.helmRepo = this.emptyRepo;
+    this.showOptions = false;
+  };
+
+  setFilepath(type: FileType, value: string) {
+    this.helmRepo[type] = value;
+  }
+
+  getFilePath(type: FileType) : string {
+    return this.helmRepo[type];
+  }
+
+  async selectFileDialog(type: FileType, fileFilter: FileFilter) {
+    const { dialog, BrowserWindow } = remote;
+    const { canceled, filePaths } = await dialog.showOpenDialog(BrowserWindow.getFocusedWindow(), {
+      defaultPath: this.getFilePath(type),
+      properties: ["openFile", "showHiddenFiles"],
+      message: _i18n._(t`Select file`),
+      buttonLabel: _i18n._(t`Use file`),
+      filters: [
+        fileFilter,
+        { name: "Any", extensions: ["*"]}
+      ]
+    });
+
+    if (!canceled && filePaths.length) {
+      this.setFilepath(type, filePaths[0]);
+    }
+  }
+
+  async addCustomRepo() {
+    try {
+      await repoManager.addСustomRepo(this.helmRepo);
+      Notifications.ok(<Trans>Helm repository <b>{this.helmRepo.name}</b> has added</Trans>);
+      this.props.onAddRepo();
+      this.close();
+    } catch (err) {
+      Notifications.error(<Trans>Adding helm branch <b>{this.helmRepo.name}</b> has failed: {String(err)}</Trans>);
+    }
+  }
+
+  renderFileInput(placeholder:string, fileType:FileType ,fileExtensions:string[]){
+    return(
+      <div className="flex gaps align-center">
+        <Input
+          placeholder={placeholder}
+          validators = {isPath}
+          className="box grow"
+          value={this.getFilePath(fileType)}
+          onChange={v => this.setFilepath(fileType, v)}
+        />
+        <Icon
+          material="folder"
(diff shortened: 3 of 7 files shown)


src/renderer/components/+preferences/add-helm-repo-dialog.tsx
```diff
@@ -0,0 +1,188 @@
...
+  static close() {
+    AddHelmRepoDialog.isOpen = false;
+  }
+
+  @observable helmRepo : HelmRepo = {name: "", url: "", username: "", password: "", insecureSkipTlsVerify: false, caFile:"", keyFile: "", certFile: ""}
+  @observable showOptions = false
+
+  close = () => {
+    AddHelmRepoDialog.close();
+    this.helmRepo = {name: "", url: "", username: "", password: "", insecureSkipTlsVerify: false, caFile:"", keyFile: "", certFile: ""}
```
aleksfront: We can move this initial object into private variable to avoid code duplication.

src/renderer/components/+preferences/add-helm-repo-dialog.tsx
```diff
@@ -0,0 +1,188 @@
...
+    this.showOptions = false
+  }
+
+  setFilepath(type: FileType, value: string) {
```
aleksfront, line 56: If we would use serialized string `FileType` enum, we can replace all switch statement to this
```js
this.helmRepo[type] = value;
```

src/renderer/components/+preferences/add-helm-repo-dialog.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+interface Props extends Partial<DialogProps> {
+  onAddRepo: Function
+}
+
+enum FileType{
+  CaFile,
+  KeyFile,
+  CertFlie,
+}
```
aleksfront: We can try to use [String enum](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) here to remove long switch statements in code below.
```
enum FileType {
  CaFile = "caFile",
  KeyFile = "keyFile",
  CertFile = "certFile",
}

File: src/renderer/components/+preferences/add-helm-repo-dialog.tsx [hunk 4]
```diff
@@ -0,0 +1,188 @@
...
+    return (
+      <Dialog
+        {...dialogProps}
+        className="AddHelmRepoDialog"
+        isOpen={AddHelmRepoDialog.isOpen}
+        close={this.close}
+      >
+        <Wizard header={header} done={this.close}>
+          <WizardStep contentClass="flow column" nextLabel={<Trans>Add</Trans>} next={()=>{this.addCustomRepo()}}>
+            <div className="secret-name">
```
aleksfront: Replace className to `flex column gaps`.

aleksfront (changes requested):
Implementation is good, just need to add couple of UI fixes.
1. Please put `Add Custom Helm Repo` dialog to the right of repo selector. You can do this by moving select and button into parent div element like this:
```
<h2><Trans>Helm</Trans></h2>
<div className="flex gaps">
  <Select id="HelmRepoSelect"
    placeholder={<Trans>Repositories</Trans>}
    isLoading={this.helmLoading}
    isDisabled={this.helmLoading}
    options={this.helmOptions}
    onChange={this.onRepoSelect}
    formatOptionLabel={this.formatHelmOptionLabel}
    controlShouldRenderValue={false}
    className="box grow"
  />
  <Button
    primary
    label={<Trans>Add Custom Helm Repo</Trans>}
    onClick={() => {AddHelmRepoDialog.open()}}
  />
</div>
```
helm repos

2. Fix focus/tab states in newly created dialog. When you open it, focus appears not on the first input field, but on the second one. Same happens when you click "More" button - focus activated on last field in list.

3. We can remove first title in dialog so the whole view would be like on the following screenshot
add custom helm repo

Other visual fixes will be done in separate PRs since whole dialog view should be redefined.

aleksfront:
@pashevskii Also, please merge current master changes into your branch and resolve conflict in `preferences.tsx` and keep in mind that our codebase now require stricter lint rules, so you need to set semicolons everywhere at least.

Check out these commands `yarn lint` and `yarn lint:fix`.

pashevskii:
> @pashevskii Also, please merge current master changes into your branch and resolve conflict in `preferences.tsx` and keep in mind that our codebase now require stricter lint rules, so you need to set semicolons everywhere at least.
> 
> Check out these commands `yarn lint` and `yarn lint:fix`.

Ok, thank you for your comments, I'll try to fix it today. 
Need your opinion about "More + " improvised accordion element, it looks not convenient enough, because you have to click on a small clickable area. Probably I should cover this one with some invisible area which has the same onClick function or replace the element with a button, what do you think about it?

aleksfront:
> > @pashevskii Also, please merge current master changes into your branch and resolve conflict in `preferences.tsx` and keep in mind that our codebase now require stricter lint rules, so you need to set semicolons everywhere at least.
> > Check out these commands `yarn lint` and `yarn lint:fix`.
> 
> Ok, thank you for your comments, I'll try to fix it today.
> Need your opinion about "More + " improvised accordion element, it looks not convenient enough, because you have to click on a small clickable area. Probably I should cover this one with some invisible area which has the same onClick function or replace the element with a button, what do you think about it?

Yes, you can wrap text and plus icon within `<Button />` component, but not to forget to set `plain` prop on it.

pashevskii:
Fixed, I hope I didn't forget anything

src/main/helm/helm-repo-manager.ts
```diff
@@ -131,6 +131,25 @@ export class HelmRepoManager extends Singleton {
     return stdout;
   }
 
+  public async addСustomRepo( repoAttributes : HelmRepo) {
```
aleksfront: Forgotten space after "("

src/renderer/components/+preferences/add-helm-repo-dialog.tsx
```diff
@@ -0,0 +1,173 @@
...
+}
+
+enum FileType {
+  CaFile = "caFile",
+  KeyFile = "keyFile",
+  CertFile = "certFile",
+}
+
+
+
```
aleksfront: Please remove those empty lines

pashevskii:
> Yes, you can wrap text and plus icon within `<Button />` component, but not to forget to set `plain` prop on it.

I've added the plain button instead of the label and hidden all side effects in CSS (focus on the button). 
image

src/renderer/components/+preferences/add-helm-repo-dialog.scss
```diff
@@ -0,0 +1,24 @@
...
+  .accordion {
+    font-weight: bold;
+    text-align: left;
+    padding: 0;
+    color: inherit;
+    &:focus:not(:active) {
+      color: inherit;
+      box-shadow: 0 0 0 0px inset;
+    }
+  }
```
aleksfront: Replace `.accordion` class with following statements
```
.accordion {
    font-weight: bold;
    align-self: flex-start;
    padding: 0;

    &:focus:not(:active) {
      box-shadow: none;
    }
  }
```
We want to preserve focus state on element.

src/renderer/components/+preferences/add-helm-repo-dialog.tsx
```diff
@@ -0,0 +1,171 @@
...
+    this.helmRepo[type] = value;
+  }
+
+  getFilePath(type: FileType) : string {
+    return this.helmRepo[type];
+  }
+
+  async selectFileDialog(type: FileType) {
+    const { dialog, BrowserWindow } = remote;
+    const { canceled, filePaths } = await dialog.showOpenDialog(BrowserWindow.getFocusedWindow(), {
```
aleksfront, line 66: Does our dialog need to filter extensions for key files? If so, please provide a list of them.
See https://github.com/lensapp/lens/blob/master/src/renderer/components/+extensions/extensions.tsx#L149 for reference.

aleksfront:
@pashevskii Left 3 more comments, ptal. Otherwise looks good.

src/renderer/components/+preferences/add-helm-repo-dialog.tsx
```diff
@@ -0,0 +1,171 @@
...
+        className="AddHelmRepoDialog"
+        isOpen={AddHelmRepoDialog.isOpen}
+        close={this.close}
+      >
+        <Wizard header={header} done={this.close}>
+          <WizardStep contentClass="flow column" nextLabel={<Trans>Add</Trans>} next={()=>{this.addCustomRepo();}}>
+            <div className="flex column gaps">
+              <Input
+                autoFocus required
+                placeholder={_i18n._(t`Chart Name`)}
```
aleksfront: Please change a placeholder here to `Helm repo name`.

aleksfront (dismissed):

src/renderer/components/+preferences/add-helm-repo-dialog.tsx
pashevskii: Hmm, checked helm code and didn't find any file extension checks. Added some default lists of extensions:
image
Is it ok? 
Also, I've added "Any" filter in every file dialog

src/renderer/components/+preferences/add-helm-repo-dialog.scss
pashevskii: Fixed. As for me, looks a little bit strange :)
image

aleksfront: Oh, sorry, probably we need to restore `color: inherit` into parent `accordion` class. The intention is to keep visible focused state, so button should be blue with focus, but gray without focus.

src/renderer/components/+preferences/add-helm-repo-dialog.tsx
aleksfront: Looks good and very detailed 😄

src/renderer/components/+preferences/add-helm-repo-dialog.scss
pashevskii: It works how you said, just I would like to leave it grey

src/renderer/components/+preferences/add-helm-repo-dialog.tsx
pashevskii: Just the laziest way to set extensions :) Otherwise, I'll need  to convert FileType to a separate class, and I'm not sure in profit of this

src/renderer/components/+preferences/add-helm-repo-dialog.scss
pashevskii: Hm, I was not understanding what is strange in this accordion, now I got it, configuration-secrets-create form contains similar things, but they are not "focusable" (I stole the idea from that form)

src/renderer/components/+preferences/add-helm-repo-dialog.tsx
pashevskii: Just took from [this](https://doc-archives.microstrategy.com/producthelp/10.10/SystemAdmin/WebHelp/Lang_1033/Content/certificate_file_overview.htm) :)

src/main/helm/helm-repo-manager.ts
```diff
@@ -131,6 +131,25 @@ export class HelmRepoManager extends Singleton {
...
+    const caFile = repoAttributes.caFile ? ` --ca-file "${repoAttributes.caFile}"` : "";
+    const keyFile = repoAttributes.keyFile ? ` --key-file "${repoAttributes.keyFile}"` : "";
+    const certFile = repoAttributes.certFile ? ` --cert-file "${repoAttributes.certFile}"` : "";
+
+    const addRepoCommand = `"${helm}" repo add ${repoAttributes.name} ${repoAttributes.url}${insecureSkipTlsVerify}${username}${password}${caFile}${keyFile}${certFile}`;
+    const { stdout } = await promiseExec(addRepoCommand).catch((error) => {
+      throw(error.stderr);
+    });
+
+    return stdout;
```
Nokel81, line 150:
```suggestion
    try {
      const { stdout } = await promiseExec(addRepoCommand);
      
      return stdout;
    }  catch (error) {
      logger.error("[HELM]: adding repo failed", error);
      
      return null;
    }
```

src/renderer/components/+preferences/add-helm-repo-dialog.tsx
```diff
@@ -0,0 +1,175 @@
...
+      <>
+        <SubTitle title={<Trans>Security settings</Trans>} />
+        <Checkbox
+          label={_i18n._(t`Skip TLS certificate checks for the repository`)}
+          value={this.helmRepo.insecureSkipTlsVerify}
+          onChange={v => this.helmRepo.insecureSkipTlsVerify = v}
+        />
+        {this.renderFileInput(_i18n._(t`Key file`), FileType.KeyFile, ["key", "keystore", "jks", "p12", "pfx", "pem"] )}
+        {this.renderFileInput(_i18n._(t`Ca file`), FileType.CaFile, ["crt", "cer", "ca-bundle", "p7b", "p7c" , "p7s", "p12", "pfx", "pem"] )}
+        {this.renderFileInput(_i18n._(t`Cerificate file`), FileType.CertFile, ["crt", "cer", "ca-bundle", "p7b", "p7c" , "p7s", "p12", "pfx", "pem"] )}
```
Nokel81: I would like to see these extracted to `private static`'s as there is some duplication here.

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -134,16 +136,25 @@ export class Preferences extends React.Component {
...
+            options={this.helmOptions}
+            onChange={this.onRepoSelect}
+            formatOptionLabel={this.formatHelmOptionLabel}
+            controlShouldRenderValue={false}
+            className="box grow"
+          />
+          <Button
+            primary
+            label={<Trans>Add Custom Helm Repo</Trans>}
+            onClick={() => {AddHelmRepoDialog.open();}}
```
Nokel81:
```suggestion
            onClick={AddHelmRepoDialog.open}
```

src/main/helm/helm-repo-manager.ts
pashevskii: @Nokel81 It seems the suggestion blocks throwing error on [UI](https://github.com/lensapp/lens/pull/1368/commits/612109f555b83821993ea5c70dd057fa84908490#diff-605267243c2bec0a8b0e324535d62a4c8ec44b8eed1e59a99d4782bc1c1d3212R86), am I right?

Nokel81: You are, good catch.

Nokel81 (approved):
Looks good to me.

aleksfront (approved):
LGTM

PR #1250, opened 2020-11-06
Add webpack hot-reload/cache plugins to improve Lens DX

chenhunghan:
for better DX..

- Add `webpack-dev-server`/[react-refresh](https://www.npmjs.com/package/react-refresh)/[react-refresh-webpack-plugin](https://github.com/pmmmwh/react-refresh-webpack-plugin) for hot-reload support, and add condition (see [here](https://github.com/lensapp/lens/pull/1250/commits/79abfe49d9bad0c11add8a74758aa765d1aec7fb))to redirect static file requests (`/LensDev.js`/`LensDev.html`) to  `webpack-dev-server`. React-refresh is experimental, but seems to work fine. 
- Add `type-fest` for support of typescript for [react-refresh-webpack-plugin](https://github.com/pmmmwh/react-refresh-webpack-plugin) according to its README.
- Add [hard-source-webpack-plugin](https://github.com/mzgoddard/hard-source-webpack-plugin) for caching both main and renderer, the caching will improve webpack re-pack speed (like the old webpack DLL plugin). Webpack 5 support this out-of--the-box so we should remove this when upgrade to Webpack 5.
- I tried to upgrade to Webpack 5 but failed, too many error regarding the plugins.

PR #1373, opened 2020-11-13
add retry cap to handleStaticFile to prevent an infitite loop

Nokel81:

This is a fix followup to #1250 which sometimes can cause an infinite loop during loading if `LensDev.html` was not generated recently.


chenhunghan (approved):
Thanks, looks good for me

PR #1374, opened 2020-11-13
Default PageRegistration.routePath to path based on id

jakolehm:


src/extensions/registries/page-registry.ts
```diff
@@ -35,9 +32,12 @@ export class PageRegistry<T extends PageRegistration> extends BaseRegistry<T> {
...
   @action
   add(items: T[], ext?: LensExtension) {
-    const normalizedItems = items.map((i) => {
-      i.routePath = getPageUrl(ext, i.routePath)
-      return i
+    const normalizedItems = items.map((page) => {
+      if (!page.routePath) {
+        page.routePath = `/${page.id}`
+      }
+      page.routePath = getPageUrl(ext, page.routePath)
```
Nokel81, line 39: This call in dangerous because `ext` could be undefined here but `getPagUrl` assume that it isn't

Nokel81 (changes requested):
Please add another test when `ext` is undefined as well.

jim-docker (approved):

src/extensions/registries/page-registry.ts
jakolehm: It was like that... but I agree with you. This add is called only from a single place where ext is always given. Ext is only optional because otherwise method signature is not compatible with parent class.

Nokel81: okay sounds good

Nokel81 (approved):

PR #1376, opened 2020-11-13
Remove unused build/notarize.js

Nokel81:
- File is not referenced anywhere in the repo

- Allows us to set "allowJs" to `false` in the `tsconfig.json` which
  reduces the random errors when opening that file in vscode

- Also remove the electron-notarize package as no longer used anywhere


jakolehm (changes requested):
It's used, see: https://github.com/lensapp/lens/blob/master/package.json#L90

Nokel81:
Sorry, I did a search but that didn't come up for some reason.

PR #1377, opened 2020-11-13
set "allowJs" to false in tsconfig

Nokel81:
- None of the .ts files include any of the .js files

- All .js files are either passed to electron-builder or used as config
  files for other programs.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -18,7 +18,7 @@
     "experimentalDecorators": true,
     "emitDecoratorMetadata": true,
     "skipLibCheck": true,
-    "allowJs": true,
+    "allowJs": false,
     "esModuleInterop": true,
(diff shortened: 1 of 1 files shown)


jim-docker (approved):
hopefully there's not a gotcha somewhere...

tsconfig.json
```diff
@@ -18,7 +18,7 @@
     "experimentalDecorators": true,
     "emitDecoratorMetadata": true,
     "skipLibCheck": true,
-    "allowJs": true,
+    "allowJs": false,
```
ixrock, line 21: Why this is useful? Faster compilation?

Nokel81: Just in general a preference to being more tight in what we accept (and getting software to verify it for us).

PR #1380, opened 2020-11-14
Fix PageRegistry.getByPageMenuTarget issue when package name has special chars

jakolehm:
Regression in 4.0.0-beta.2.


nevalla (approved):
LGTM

PR #1381, opened 2020-11-14
Fix support extension menu route

jakolehm:
Regression in 4.0.0-beta.2.

### DIFF
--- a/extensions/support-page/main.ts
+++ b/extensions/support-page/main.ts
@@ -6,7 +6,7 @@ export default class SupportPageMainExtension extends LensMainExtension {
       parentId: "help",
       label: "Support",
       click: () => {
-        this.navigate();
+        this.navigate("/support");
       }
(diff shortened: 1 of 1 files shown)


extensions/support-page/main.ts
```diff
@@ -6,7 +6,7 @@ export default class SupportPageMainExtension extends LensMainExtension {
       parentId: "help",
       label: "Support",
       click: () => {
-        this.navigate();
+        this.navigate("/support");
```
ixrock, line 9: Probably we don't even need to specify route? Default extension page url should be enough.

ixrock (approved):

extensions/support-page/main.ts
ixrock: No? IMHO redundant and would be good enough to `/extension/lens-support-page`

PR #1386, opened 2020-11-15
PageRegistration / PageMenuRegistration fixes & more simplification

### LINKED ISSUE #1383: PageMenuRegistration.target seems is required
_Current interface:_

```ts
export interface PageMenuRegistration {
  target?: PageMenuTarget;
  title: React.ReactNode;
  components: PageMenuComponents;
}
```

_How to reproduce:_

```ts
  clusterPages: Interface.PageRegistration[] = [
    {
      id: "example-extension-id",
      components: {
      }
    }
  ]

  clusterPageMenus: Interface.PageMenuRegistration[] = [
    {
      title: "Hello World",
      components: {
        Icon: ExampleIcon
      }
    }
  ]
}
```

_Works only with:_
```
      target: {
        pageId: "example-extension-id"
      },
```

ixrock:
_Most noticeable changes / fixes:_

- deprecated `page.routePath` from page-registration in favour `page.id` since they declare same thing for extension route-path at registration step (but still `routePath` remains available after registration with `pageRegistry.getItems()`)
- reliable matching page to page-menu in `pageRegistry.getByPageMenuTarget(target)` with participating custom url-params in final url compilation (if provided via `page.id` and corresponding `pageMenu.target`)
- possibility to provide params to `extension.navigate(pageId: string?, params?: P)` helpers
- make `page.id` and `pageMenu.target` optional (close #1383)

**Example 1**: 
_Minimum required page and page-menu configuration to match together:__

```ts
  clusterPages: Interface.PageRegistration[] = [
    {
      components: {
      }
    },
  ]

  clusterPageMenus: Interface.PageMenuRegistration[] = [
    // menu item matches with the page above and provide final link as "/extension/lens-hello-world"
    {
      title: "Hello World",
      components: {
        Icon: ExampleIcon
      }
    }
  ]
}
```

**Example 2**
_Extension has single registered page with param :exampleId and 2 menu targets to match with different id's_

```ts
  clusterPages: Interface.PageRegistration[] = [
    {
      id: "/example/:exampleId?",
      components: {
      }
    },
  ]

  clusterPageMenus: Interface.PageMenuRegistration[] = [
    {
      target: {
        pageId: "//example///:exampleId", // matching as well since final path optimized via `path.join`
        params: {
          exampleId: "123" // page-menu url: "/extension/lens-hello-world/example/123"
        }
      },
      title: "Hello World",
      components: {
        Icon: ExampleIcon
      }
    },
    {
      target: {
        pageId: "example/:exampleId",
        params: {
          exampleId: "456" // page-menu url: "/extension/lens-hello-world/example/456"
        }
      },
      title: "Example Extension",
      components: {
        Icon: ExampleIcon
      }
    }
  ]
}
```

Screenshot 2020-11-16 at 14 47 37
Screenshot 2020-11-16 at 14 47 51

### DIFF
--- a/extensions/support-page/main.ts
+++ b/extensions/support-page/main.ts
@@ -6,7 +6,7 @@ export default class SupportPageMainExtension extends LensMainExtension {
       parentId: "help",
       label: "Support",
       click: () => {
-        this.navigate("/support");
+        this.navigate();
       }
     }
   ]
--- a/src/extensions/lens-main-extension.ts
+++ b/src/extensions/lens-main-extension.ts
@@ -2,14 +2,18 @@ import type { MenuRegistration } from "./registries/menu-registry";
 import { observable } from "mobx";
 import { LensExtension } from "./lens-extension"
 import { WindowManager } from "../main/window-manager";
-import { getPageUrl } from "./registries/page-registry"
+import { getExtensionPageUrl } from "./registries/page-registry"
 
 export class LensMainExtension extends LensExtension {
   @observable.shallow appMenus: MenuRegistration[] = []
 
-  async navigate(location?: string, frameId?: number) {
+  async navigate<P extends object>(pageId?: string, params?: P, frameId?: number) {
     const windowManager = WindowManager.getInstance<WindowManager>();
-    const url = getPageUrl(this, location); // get full path to extension's page
-    await windowManager.navigate(url, frameId);
+    const pageUrl = getExtensionPageUrl({
+      extensionId: this.name,
+      pageId: pageId,
+      params: params ?? {}, // compile to url with params
+    });
+    await windowManager.navigate(pageUrl, frameId);
   }
 }
--- a/src/extensions/registries/__tests__/page-registry.test.ts
+++ b/src/extensions/registries/__tests__/page-registry.test.ts
@@ -1,4 +1,4 @@
-import { getPageUrl, globalPageRegistry } from "../page-registry"
+import { getExtensionPageUrl, globalPageRegistry, PageRegistration } from "../page-registry"
 import { LensExtension } from "../../lens-extension"
 import React from "react";
 
@@ -18,20 +18,19 @@ describe("getPageUrl", () => {
   })
 
   it("returns a page url for extension", () => {
-    expect(getPageUrl(ext)).toBe("/extension/foo-bar")
+    expect(getExtensionPageUrl({ extensionId: ext.name })).toBe("/extension/foo-bar")
   })
 
   it("allows to pass base url as parameter", () => {
-    expect(getPageUrl(ext, "/test")).toBe("/extension/foo-bar/test")
+    expect(getExtensionPageUrl({ extensionId: ext.name, pageId: "/test" })).toBe("/extension/foo-bar/test")
   })
 
   it("removes @", () => {
-    ext.manifest.name = "@foo/bar"
-    expect(getPageUrl(ext)).toBe("/extension/foo-bar")
+    expect(getExtensionPageUrl({ extensionId: "@foo/bar" })).toBe("/extension/foo-bar")
   })
 
   it("adds / prefix", () => {
-    expect(getPageUrl(ext, "test")).toBe("/extension/foo-bar/test")
+    expect(getExtensionPageUrl({ extensionId: ext.name, pageId: "test" })).toBe("/extension/foo-bar/test")
   })
 })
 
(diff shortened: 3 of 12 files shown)


extensions/support-page/main.ts
```diff
@@ -6,7 +6,7 @@ export default class SupportPageMainExtension extends LensMainExtension {
       parentId: "help",
       label: "Support",
       click: () => {
-        this.navigate("/support");
+        this.navigate();
```
jim-docker, line 9: Interesting. This couldn't be more simple :) Makes sense if your extension only has one page but if there are more than one page you would probably give them all ids (rather than giving one the `""` id)? Or maybe one of the pages could be considered the "top" page. Is this the idea? Guess it's up to the extension developer...

src/extensions/lens-main-extension.ts
```diff
@@ -2,14 +2,18 @@ import type { MenuRegistration } from "./registries/menu-registry";
...
 import { LensExtension } from "./lens-extension"
 import { WindowManager } from "../main/window-manager";
-import { getPageUrl } from "./registries/page-registry"
+import { getExtensionPageUrl } from "./registries/page-registry"
 
 export class LensMainExtension extends LensExtension {
   @observable.shallow appMenus: MenuRegistration[] = []
 
-  async navigate(location?: string, frameId?: number) {
+  async navigate<P extends object>(pageId?: string, params?: P, frameId?: number) {
```
jim-docker, line 10: nice. Would we introduce another function for navigating to a page in another extension? (adding another optional param here for the extension id would probably be too unwieldy) Like
```
async navigateExtension<P extends object>(extensionId: extId, pageId?: string, params?: P, frameId?: number)
```

src/extensions/registries/__tests__/page-registry.test.ts
```diff
@@ -57,12 +56,32 @@ describe("globalPageRegistry", () => {
...
+      expect(page.routePath).toEqual(getExtensionPageUrl({ extensionId: ext.name }));
+    });
+
+    it("returns matching page", () => {
+      const page = globalPageRegistry.getByPageMenuTarget({
+        pageId: "test-page",
+        extensionId: ext.name
+      })
+      expect(page.id).toEqual("test-page")
+    })
```
jim-docker: duplicate test of the one below?

jim-docker (approved):

src/extensions/lens-main-extension.ts
jakolehm: This breaks the API compatibility, right? Would it be too bad if we kept `navigation` as is and introduce this as a new method (with additional `extensionId?`)?

src/extensions/registries/page-registry.ts
```diff
@@ -1,59 +1,91 @@
...
 import { compile } from "path-to-regexp";
 import { BaseRegistry } from "./base-registry";
-import { LensExtension } from "../lens-extension"
-import type { PageMenuTarget } from "./page-menu-registry";
+import { LensExtension } from "../lens-extension";
+import logger from "../../main/logger";
 
 export interface PageRegistration {
-  id: string; // will be automatically prefixed with extension name
-  routePath?: string; // additional (suffix) route path to base extension's route: "/extension/:name"
```
jakolehm, line 12: This also breaks API compatibility.

extensions/support-page/main.ts
ixrock: Yes, I guess it's up to developer to manage proper ID-s with multiple pages. We could introduce though at least `page.id` validation on pages registration to check uniqueness of ID-s within same extension.

src/extensions/lens-main-extension.ts
ixrock: It breaks API only in second argument, right? It's done to be more aligned with `ext.navigate(location)` in renderer.
And using this `frameId` mostly for internal usage, at least for now?
Otherwise we can introduce new method like `ext.navigateWith(opts: PageMenuTarget & {frameId?: number})`. 
Wdyt?

src/extensions/registries/__tests__/page-registry.test.ts
ixrock: Yup, will clean up

src/extensions/registries/page-registry.ts
ixrock: Added deprecation flag for now, but since we're still in beta and this was `optional` I would suggest still to break it now.

src/extensions/lens-main-extension.ts
jakolehm: @ixrock you mean that `this.navigate("/my-page")` still works?

src/extensions/registries/page-registry.ts
jakolehm: Deprecation is better, we probably should remove it completely before GA.

src/extensions/lens-main-extension.ts
ixrock: @jakolehm Yes, it should

jakolehm (approved):

PR #1387, opened 2020-11-15
Generate extension-api .d.ts and .js files using tsc

### LINKED ISSUE #1262: Export `@k8slens/extensions` for writing extension unit tests

Export modules of `@k8slens/extensions` (currently has only `.d.ts`) for development/testing purposes only.

**Why is this needed**:

To be able to write unit tests for extensions.

Currently test throws

```
Cannot find module '@k8slens/extensions' from 'page.tsx'

Require stack:
 page.tsx
 page.test.tsx

> 1 | import { LensRendererExtension, Component } from "@k8slens/extensions";
```

jakolehm:
- generate `.d.ts` & `.js` files using `tsc`
- remove rollup (not used anymore)
- fixes #1262


src/common/utils/debouncePromise.ts
```diff
@@ -4,6 +4,6 @@ export function debouncePromise<T, F extends any[]>(func: (...args: F) => T | Pr
   let timer: NodeJS.Timeout;
   return (...params: any[]) => new Promise((resolve, reject) => {
     clearTimeout(timer);
-    timer = setTimeout(() => resolve(func.apply(this, params)), timeout);
+    timer = global.setTimeout(() => resolve(func.apply(this, params)), timeout);
```
ixrock, line 7: Why `global.setTimeout`? Any difference?

ixrock (approved):
Cool! And how it actually does it work? Is it because we started to re-export everything to  main/renderer api bundles? Haven't checked it in work yet though. Usually `tsc` compiler could generate such single file declarations only for AMD-modules or something.

ixrock:
Helm test seems to be failed?
✕ ensures helm repos (60001 ms)

src/common/utils/debouncePromise.ts
jakolehm: Typescript complained about this:
> src/common/utils/debouncePromise.ts:7:5 - error TS2322: Type 'number' is not assignable to type 'Timeout'.

Basically typescript thinks it runs in a browser (`lib.DOM` is enabled) unless we explicitly say to use nodejs function.

jakolehm:
> Cool! And how it actually does it work? Is it because we started to re-export everything to main/renderer api bundles? Haven't checked it in work yet though. Usually `tsc` compiler could generate such single file declarations only for AMD-modules or something.

This does _not_ generate single file declaration but does that really matter? Package.json is configured like this:
```
  "main": "dist/src/extensions/extension-api.js",
  "types": "dist/src/extensions/extension-api.d.ts",
```
which essentially means we are only exposing extension-api through `import "@k8slens/extensions`. Non-obfuscated d.ts/js is much more easy to explore from extension developers point of view.

jakolehm:
Let's not merge this yet, it seems that generated js files need to be tweaked.

jakolehm:
@ixrock @chenhunghan PTAL (again)

chenhunghan:
LGTM

chenhunghan (approved):

tsconfig.extensions.json
```diff
@@ -0,0 +1,11 @@
...
+  "extends": "./tsconfig.json",
+  "files": [
+    "src/extensions/extension-api.ts",
+  ],
+  "compilerOptions": {
+    "module": "CommonJS",
+    "sourceMap": false,
+    "declaration": true
+  }
+}
```
ixrock, line 11:
```suggestion
{
  "extends": "./tsconfig.json",
  "include": [
    "types",
    "src/extensions/extension-api.ts"
  ],
  "compilerOptions": {
    "module": "CommonJS",
    "sourceMap": false,
    "declaration": true,
    "declarationDir": "src/extensions/npm/extensions/dist",
    "emitDeclarationOnly": true
  }
}
```

ixrock (approved):
Awesome! Works much better for DX and probably faster compilation than `rollup`.
Added `emitDeclarationOnly": true` to `tsconfig.extensions.json` since we don't need to publish *.js files to NPM (all served from runtime).

tsconfig.extensions.json
chenhunghan: I think @jakolehm aims to fix this issue https://github.com/lensapp/lens/issues/1262 so these `.js` are needed (for the purpose of writing tests only)

jakolehm: Yes, we need `.js` files for test runtime.

ixrock: Alright, got it! 👍

PR #1397, opened 2020-11-16
Badge: option to show short/full versions of contents

ixrock:
Simplified version without icon to toggle Badge contents. Selecting text from Badge doesn't trigger state changing.
For version with small icon clicking area much smaller => harder to click => worse UI/UX.

close  #1236
close  #1400 

Screenshot 2020-11-16 at 20 07 33

Screenshot 2020-11-16 at 20 07 50


src/renderer/components/badge/badge.scss
```diff
@@ -6,10 +6,25 @@
...
   white-space: nowrap;
   max-width: 100%;
-  overflow: hidden;
-  text-overflow: ellipsis;
+  position: relative;
+  cursor: pointer;
+
+  &:hover {
+    background-color: $mainBackground;
+  }
```
Nokel81: This is confusing for items that don't have extra text to display.

src/renderer/components/badge/badge.tsx
```diff
@@ -3,21 +3,43 @@ import "./badge.scss"
...
 
 @withTooltip
+@observer
 export class Badge extends React.Component<BadgeProps> {
+  @observable isExpanded = false;
+
+  @autobind()
+  onMouseRelease() {
+    const textSelected = document.getSelection().toString().trim() !== "";
+    if (textSelected) return;
```
Nokel81: If you highlight and then stop highlighting text and then release, it still closes. This is pretty weird.

Nokel81 (changes requested):

src/renderer/components/badge/badge.tsx
```diff
@@ -3,21 +3,43 @@ import "./badge.scss"
...
 import { cssNames } from "../../utils/cssNames";
 import { TooltipDecoratorProps, withTooltip } from "../tooltip";
+import { observer } from "mobx-react";
+import { autobind } from "../../utils";
+import { observable } from "mobx";
 
 export interface BadgeProps extends React.HTMLAttributes<any>, TooltipDecoratorProps {
   small?: boolean;
   label?: React.ReactNode;
+  isExpanded?: boolean; // by default: false (minimized)
```
Nokel81: Why is this a prop, what has the parent go to do with expanding the badge?

src/renderer/components/badge/badge.tsx
ixrock: Nope, can't reproduce. Note: it's bound to `onMouseUp`

src/renderer/components/badge/badge.tsx
ixrock: `isExpanded={true}` will allow to show all contents from the Badge always (forced expanded state)

src/renderer/components/badge/badge.scss
ixrock: No it's not. Actually it improves contrast/readability of text-label. Just try it in live :)

src/renderer/components/badge/badge.tsx
Nokel81: Okay sounds good

src/renderer/components/badge/badge.tsx
Nokel81: Screen Recording 2020-11-16 at 1 29 56 PM

src/renderer/components/badge/badge.scss
Nokel81: I did, and that is where I noticed it. It is also bad UI to have the mouse change to `"pointer"` when there is no action to be taken.

src/renderer/components/badge/badge.tsx
ixrock: Yes, in your example actual selection won't happen, so why not? Very rare use-case imho (to return mouse to the start and just like "oh, i have changed by mind"?)
Could be caught easily with `document.onselectionchange` if _really needed_.

src/renderer/components/badge/badge.scss
ixrock: Having `cursor: pointer` for some of them which is not actually interactive at all doesn't sound too bad for me.

src/renderer/components/badge/badge.tsx
Nokel81: Maybe, but I am talking about intent. The user probably would find this jarring since he didn't intend to close the badge.

src/renderer/components/badge/badge.scss
Nokel81: The point of `cursor: pointer` is as an affordance and should be used as one.

miskun (dismissed):
I'm not reviewing the code but I have reviewed the implementation (concept) which is good! Some other people should review actual code.

src/renderer/components/badge/badge.scss
ixrock: Ok, removed when the badge considered as "non-interactive"

src/renderer/components/badge/badge.tsx
```diff
@@ -13,6 +13,15 @@ export interface BadgeProps extends React.HTMLAttributes<any>, TooltipDecoratorP
...
 }
 
+const badgeMeta = observable({
+  hasTextSelected: false,
+});
+
+// Common handler for all Badge instances
+document.addEventListener("selectionchange", () => {
+  badgeMeta.hasTextSelected = window.getSelection().toString().trim().length > 0
+});
```
Nokel81: This is no different than checking it in the handler because clearing the selection will clear this field too. That is why I was adding the `||=`

aleksfront (dismissed):

aleksfront:
Moved `<Badge/>` styling to CSS Modules, checked intended behavior.


https://user-images.githubusercontent.com/9607060/125440038-521fb43e-8517-492c-89a1-903edfdae91f.mov

src/renderer/components/badge/badge.tsx
```diff
@@ -19,29 +19,81 @@
...
 }
 
+const badgeMeta = observable({
+  hasTextSelected: false,
+});
+
+// Common handler for all Badge instances
+document.addEventListener("selectionchange", () => {
+  badgeMeta.hasTextSelected = window.getSelection().toString().trim().length > 0;
+});
```
Nokel81: @aleksfront the bug I mentioned is here this should be `||=` and then cleared in `onMouseUp`.

src/renderer/components/badge/badge.tsx
```diff
@@ -19,29 +19,81 @@
...
   flat?: boolean;
   label?: React.ReactNode;
+  expandable?: boolean;
+  disabled?: boolean;
 }
 
+const badgeMeta = observable({
+  hasTextSelected: false,
+});
+
```
Nokel81: nit: I think this should be a private static

```ts
class Badge
  private static meta = observable({
    hasTextSelected: false,
  });
```

Nokel81 (changes requested):

src/renderer/components/badge/badge.tsx
aleksfront: Fixed the bug with your suggestions:

https://user-images.githubusercontent.com/9607060/125458490-bf1a1899-941b-4c09-ab3a-2dcb61e94d1f.mov

Nokel81 (approved):

PR #1398, opened 2020-11-16
Electron 9.3.4

jakolehm:
https://www.electronjs.org/releases/stable?version=9#9.3.4

### DIFF
--- a/integration/__tests__/app.tests.ts
+++ b/integration/__tests__/app.tests.ts
@@ -29,9 +29,25 @@ describe("Lens integration tests", () => {
     await app.client.waitUntilWindowLoaded()
   }
 
+  const click = async (selector: string) => {
+    return (await app.client.$(selector)).click()
+  }
+
+  const waitForVisible = async (selector: string) => {
+    return (await app.client.$(selector)).waitForDisplayed()
+  }
+
+  const waitForEnabled = async (selector: string) => {
+    return (await app.client.$(selector)).waitForEnabled()
+  }
+
+  const waitForExist = async (selector: string) => {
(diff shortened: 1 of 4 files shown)


Nokel81 (approved):

jakolehm:
Something really strange going on with integration tests. We might need to postpone this update to later date/release.

integration/__tests__/app.tests.ts
```diff
@@ -97,21 +113,21 @@ describe("Lens integration tests", () => {
...
+    await click("div.Select__control") // hide the context drop-down list (it might be obscuring the Add cluster(s) button)
+    await click("button.primary") // add minikube cluster
   }
 
   const waitForMinikubeDashboard = async (app: Application) => {
     await app.client.waitUntilTextExists("pre.kube-auth-out", "Authentication proxy started")
-    await app.client.waitForExist(`iframe[name="minikube"]`)
-    await app.client.frame("minikube")
+    await (await app.client.$(`iframe[name="minikube"]`)).waitForExist()
+    await app.client.switchToFrame("minikube")
```
jim-docker, line 130: Build job shows
```
    Malformed type for "id" parameter of command switchToFrame
    Expected: (number|object|null)
    Actual: string
```
When I hover in VS Code it thinks it should be `id: string | number | object` so deps may still be off?

According to docs it wants the element's id or you can pass the object (gotten from $). I tried this but it just times out:
```suggestion
    const iframeElement = await app.client.$(`iframe[name="minikube"]`)
    await iframeElement.waitForExist()
    await app.client.switchToFrame(iframeElement)
```

jakolehm:
Replaced by #1562

PR #1400, opened 2020-11-16
Add ability to click a Badge to expand to view its whole contents

Nokel81:
Description of changes:
-
Make `<Badge>` dynamically expandable based on if it is overflowing its current view. A badge is signified as exapandable by both the `...` from `text-overflow: ellipsis` and from the change of background colour during `onHover`.

Reason for complications:
-
The code surrounding the "hasHighlightedText" is needed to prevent the badge from transitioning between extension states during a click event that includes a highlight event.

closes #1397

---

Video of this branch:

https://user-images.githubusercontent.com/8225332/119888366-42b69080-bf03-11eb-99ac-2004c57a8302.mov


src/renderer/components/badge/badge.tsx
```diff
@@ -16,26 +16,70 @@ export interface BadgeProps extends React.HTMLAttributes<any>, TooltipDecoratorP
 @withTooltip
 @observer
 export class Badge extends React.Component<BadgeProps> {
-  @observable isExpanded = false;
+  @observable _isExpanded = false;
+  @observable hasHighlightedText = false;
+  @observable interval?: NodeJS.Timeout;
```
ixrock: Still think it's too overcomplicated and I'm against using timers where they really not necessary.
Otherwise to detect element size changes we could use `css-element-queries` (already in package.json) or https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver

Nokel81: I think you are misunderstanding what that interval is doing. During the period of time when the mouse has been clicked (and only then) this interval periodically checks too see if any text has been selected. This has nothing to do with resizing the object so I don't see how `ResizeObserver` relates.

ixrock: But it should do something in ideal world with `ResizeObserver` since it detects when element size changes (badges might be with dynamic width / `%`).
About periodically checking via interval I can't get why we need this as a function? Use-case for me still obscure.. 🧐 Otherwise for dealing with `document.getSelection()` I would recommend to use `document.addEventListener("selectionchange")` instead of timers.

Nokel81: Interesting point, I tried directly setting `document.onselectionchange` since that is what the MDN docs seemed to imply but I was getting weird inconsistencies between the functions not seeing the same values. I would indeed much prefer an "interrupt" based system versus polling.

Let me do some more investigation.

ixrock: And still, why it's not enough just to check a selected text at `onMouseUp` before doing something (show/hide)?

Nokel81: Well two fold, once I implemented this, it did fix the close on "decided not to highlight any text" situation as well as it doesn't close it when the click would normally clear the current selection. Both of these situations, imo, make the badge act a **lot** more consistently with how other systems work.

Nokel81: Done, thanks for the suggestion.

src/renderer/components/badge/badge.tsx
```diff
@@ -16,36 +16,67 @@ export interface BadgeProps extends React.HTMLAttributes<any>, TooltipDecoratorP
...
+    const labelClass = cssNames("label-content", {
+      isExpanded: this.isExpanded,
     })
+
     return (
-      <div {...elemProps} className={classNames} onMouseUp={this.onMouseUp} ref={this.bindRef}>
-        {label}
-        {children}
+      <div {...elemProps} className={classNames} onMouseUp={this.onMouseUp} onMouseDown={this.onMouseDown} >
+        <div className={labelClass} ref={this.setRef}>
```
ixrock: This wrapper redundant and removed in my PR.

Nokel81: Good point, I think I previously added it because of how `position: absolute` works. Should we also remove `position: relative` from the `css`?

src/renderer/components/badge/badge.tsx
```diff
@@ -16,36 +16,67 @@ export interface BadgeProps extends React.HTMLAttributes<any>, TooltipDecoratorP
...
+
+  @autobind()
+  onMouseUp() {
+    document.removeEventListener("selectionchange", this.onSelectionChange)
+
+    if (!this.hasHighlightedText) {
+      this._isExpanded = !this._isExpanded
+    }
+
+    this.hasHighlightedText = false
```
ixrock: Is it same logic what's we get in result? https://github.com/lensapp/lens/pull/1397/commits/e4e4d7efe1d9eac987897d6f7a37f8d2bd919ec8

Nokel81: Not quite, see my comment on your PR

PR #1407, opened 2020-11-17
Cache make build-extensions

jakolehm:
After first run this is now no-op. Each run took ~13 mins on Windows which crippled our pipeline.


nevalla (commented):
LGTM

chenhunghan (approved):
LG to @nevalla

nevalla (approved):

PR #1410, opened 2020-11-17
Add React Developers Tools

panuhorsmalahti:
Add React Developers Tools to the Electron app for debugging purposes.


src/main/developer-tools.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Installs Electron developer tools in the development build.
+ * The dependency is not bundled to the production build.
+ */
+export const installDeveloperTools = async () => {
+  if (process.env.NODE_ENV === 'development') {
```
nevalla, line 6: `isDevelopment` variable is available in `common/vars.ts`

chenhunghan:
I tested the branch locally, the React developer tools is working fine, not sure why the integration failed https://dev.azure.com/lensapp/lensapp/_build/results?buildId=3499&view=logs&j=93056758-5bfb-5750-f113-e720ddefdb4c&t=39774cca-89df-5118-1dfb-4fca56fe47ed&l=994

src/main/developer-tools.ts
panuhorsmalahti: If we use the environment variable through a variable assignment, then webpack can't optimize the build. In other words if `isDevelopment` is used, the code inside the if branch is included in the production build. However it _appears_ the dependency itself (`electron-devtools-installer`) is excluded in both cases, but I'd still keep using process.env directly to make sure future changes will exclude the code too.

nevalla (approved):
LGTM

PR #1414, opened 2020-11-17
Track service start and stop events

### LINKED ISSUE #1333: ts alert after close
Lens: 4.0.0-alpha.5
Electron: 9.1.2
Chrome: 83.0.4103.122

close some window (don't remember which one) and then:

2020-11-11 at 16 52

### LINKED ISSUE #1356: 4.0.0-beta.1 - Crash on exit
**Describe the bug**
Close the application

**To Reproduce**
1. Open app..
2. Close app..

**Expected behavior**
Exists cleanly

**Screenshots**
image

- Lens Version: 4.0.0-beta.1
- OS: OSX

nevalla:
This PR will send tracking events for `service:start` and `service:stop` events. Service start event is send when electron app receives ready status (`app:ready`). App start is now sent when main application window is opened.

Service stop event is send when user exits the application either from system menu or tray menu. When application exists, there's a timeout before exiting. During that timeout the application has time to close cluster connections properly and send the final `service:stop` event. ~~User is transferred to dedicated `ClosingPage` when application starts to exit~~

This PR probably fixes #1333 and fixes #1356, but requires verification


src/main/index.ts
```diff
@@ -82,7 +82,7 @@ app.on("ready", async () => {
   extensionLoader.init(await extensionManager.load()); // call after windowManager to see splash earlier
 
   setTimeout(() => {
-    appEventBus.emit({ name: "app", action: "start" })
+    appEventBus.emit({ name: "service", action: "start" })
   }, 1000)
```
ixrock, line 86: Why it's done here via timeout btw?

src/main/exit-app.ts
```diff
@@ -0,0 +1,18 @@
...
+  const windowManager = WindowManager.getInstance<WindowManager>()
+  const clusterManager = ClusterManager.getInstance<ClusterManager>()
+  appEventBus.emit({ name: "service", action: "close" })
+  windowManager?.hide();
+  clusterManager?.stop();
+  logger.info('SERVICE:QUIT');
+  setTimeout(() => {
+    app.exit()
+  }, 1000)
+}
```
ixrock:
```suggestion
export function exitApp(timeoutMs = 1000) {
  const windowManager = WindowManager.getInstance<WindowManager>()
  const clusterManager = ClusterManager.getInstance<ClusterManager>()
  appEventBus.emit({ name: "service", action: "close" })
  windowManager?.hide();
  clusterManager?.stop();
  logger.info('SERVICE:QUIT');
  setTimeout(() => {
    app.exit()
  }, timeoutMs)
}
```

ixrock (approved):
LGTM

src/main/index.ts
nevalla: Don't really know, perhaps @jakolehm remembers

src/main/exit-app.ts
```diff
@@ -0,0 +1,18 @@
...
+import { ClusterManager } from "./cluster-manager";
+import logger from "./logger";
+
+
+export function exitApp() {
+  const windowManager = WindowManager.getInstance<WindowManager>()
+  const clusterManager = ClusterManager.getInstance<ClusterManager>()
+  appEventBus.emit({ name: "service", action: "close" })
+  windowManager?.hide();
+  clusterManager?.stop();
```
Nokel81: These are guaranteed to exist, the `getInstance` function creates if not present.

nevalla: Good point.

src/main/exit-app.ts
nevalla: I'm hesitating to change this as suggested, since the caller requires to know some internal logic of this method to determine proper timeout.

Nokel81: Agreed, especially since this is (and should be) only called after everything else

src/main/index.ts
jakolehm: It was hacky solution for a problem where we didn't control extension loading properly (basically nobody was listening this event right away). Timeout can be removed if `extensionLoader.init()` blocks until extensions are loaded.

jakolehm (approved):

PR #1415, opened 2020-11-17
Fix Ctrl+Q bug

pashevskii:
Hi, I found some issues related to not-working Ctrl+Q/File->Quit flows:
Lens just closes the window but still works, however the same button in the tray fully closes the app

So, IMHO the deal in the wrong event, where the system prevents app.exit() call on windows closing.

The fix has tested on Ubuntu, in theory, it should not affect Mac users because of app.exit() calling on Ctrl+Q

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -93,10 +93,9 @@ app.on("activate", (event, hasVisibleWindows) => {
   }
 });
 
-// Quit app on Cmd+Q (MacOS)
-app.on("will-quit", (event) => {
-  logger.info('APP:QUIT');
-  event.preventDefault(); // prevent app's default shutdown (e.g. required for telemetry, etc.)
+// Prevent app.exit() on closing all windows
+app.on("window-all-closed", () => {
+  logger.info('APP:CLOSE');
   clusterManager?.stop(); // close cluster connections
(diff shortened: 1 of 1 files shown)


ixrock (changes requested):
This is done to be able to handle async events when app is closing:
```
event.preventDefault(); // prevent app's default shutdown (e.g. required for telemetry, etc.)
```
Probably could be skipped ^ if no pending async actions on exit (e.g. telemetry disabled, @jakolehm  @nevalla wdyt?)
In `Windows` menu there is a bug in `File -> Close` item (should be re-used from `macAppMenu`, see: `menu.ts`)

pashevskii:
Hi @ixrock, thank you for your answer, sorry, it seems there is some misunderstanding from my side...
I suppose that in this case event.preventDefault() function fully prevents termination process, isn't it? 
I think we should terminate the app after  Ctrl+Q instead of just closing the window
As an option we can call app.exit() after clusterManager?.stop() line, it will stop clusters and terminate app.

nevalla:
I did some renaming of File menu items in #1420. They should now reflect better the role of those items.

pashevskii:
Looked through  @nevalla  PRs, It seems he has fixed this problem. I'll close the PR

Issue #191, opened 2020-03-31
uncaught exception ERR_INVALID_ARG_TYPE


matti:
when quitting lens

![Screenshot_1_4_2020__0_20](https://user-images.githubusercontent.com/3893/78076203-a6e8cd80-73ae-11ea-9651-267c4cded835.png)

jakolehm:
Does this happen every time you quit the app?

jakolehm:
Probably already fixed.

PR #1416, opened 2020-11-17
Catch errors when responding from proxy error

### LINKED ISSUE #1356: 4.0.0-beta.1 - Crash on exit

nevalla:
This PR will catch errors when responding from proxy error. Most of the proxy errors are happening when closing the app and it's not possible to respond anymore from proxy server.

Fixes #1356 . Also before this fix saw this error #191, but not anymore after the fix.

### DIFF
--- a/src/main/lens-proxy.ts
+++ b/src/main/lens-proxy.ts
@@ -131,7 +131,11 @@ export class LensProxy {
           }
         }
       }
-      res.writeHead(500).end("Oops, something went wrong.")
+      try {
+        res.writeHead(500).end("Oops, something went wrong.")
+      } catch (e) {
+        // Can't send headers. Probably app is closing.
+      }
     })
 
     return proxy;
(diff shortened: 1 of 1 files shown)


src/main/lens-proxy.ts
```diff
@@ -131,7 +131,11 @@ export class LensProxy {
           }
         }
       }
-      res.writeHead(500).end("Oops, something went wrong.")
+      try {
+        res.writeHead(500).end("Oops, something went wrong.")
+      } catch (e) {
+        // Can't send headers. Probably app is closing.
```
jakolehm: Should we log an error?

nevalla: I have a feeling that logger is going south also at this point, but does not hurt if that does not cause any further issues

ixrock: Are we writing logs to `~/Library/Logs/Electron/lens.log`, right?

nevalla: yes. Tested logging the error and at least on Windows it did not cause any further issues.

ixrock (approved):

Issue #1375, opened 2020-11-13
Support context-sensitive cluster page menu item extensions


stefcameron:
The Extensions API should support context-sensitive cluster page menu items.

For example, only showing (and/or enabling) a menu item based on a condition, which could be anything interesting, including:

- the state of the cluster,
- the type of cluster,
- the origin (i.e. vendor) of the cluster,
- etc.

jakolehm:
I think this is already possible, see related documentation PR https://github.com/lensapp/lens/pull/1422.

Nokel81:
This seems to be now considered a bug (see #1422 )

stefcameron:
@jakolehm 
> I think this is already possible, see related documentation PR #1422.

I don't really see that as _obvious_ in #1422. That PR is showing me how to react to an observable property, but there's also the other piece of whether or not Lens will remember extra metadata I might add (can I even add extra metadata?) to a cluster object I add into the store -- between quit and restart too -- such that after a restart, when my extension loads, and a cluster gets selected, that metadata is still there.

If this is also already handled by Lens, it would be really important to point out. Otherwise, I feel like the only way I could add context-sensitive cluster menu items is if I had clusters anew every time I load Lens, because that's the only way my extension can come to know that a cluster with a specific ID is indeed a "special" cluster.

Or my extension would have to store all that state itself.

Either way, I think this is an important piece of the puzzle that many will be wondering about, and should be clearly documented somewhere.

Nokel81:
So this is not possible, and needs to have some design work attached to it. See my comment (https://github.com/lensapp/lens/issues/1432#issuecomment-733039997)

stefcameron:
@Nokel81 Thanks for the update. `LensRendererExtension.isEnabledForCluster()` from #1489 is a step in the right direction, however neither does it, nor does #1432, address the metadata part of the problem. Now I can have my extension decide if it's cluster pages should be enabled for a given cluster, however, unless my extension is the one that added that cluster, and therefore knows about it -- and has also used its own custom "way" of storing metadata for that cluster -- my extension _still_ cannot reliably decide whether it should be enabled for that cluster or not. I suppose it can be left up to an extension to store its cluster metadata somewhere on disk, but for light-weight metadata associated with the primary object type in Lens (the cluster), it seems a bit heavy-handed.

[...]

PR #1422, opened 2020-11-17
Add starting point for mobx related documentaion

Nokel81:
add a brief overview to our use of mobx and mark all observable fields as observable in our docs

### DIFF
--- a/docs/extensions/guides/README.md
+++ b/docs/extensions/guides/README.md
@@ -19,6 +19,7 @@ Each guide or sample will include:
 | [Stores](stores.md) | |
 | [Components](components.md) | |
 | [KubeObjectListLayout](kube-object-list-layout.md) | |
+| [Working with mobx](working-with-mobx.md) | |
 
 ## Samples
 
--- a/docs/extensions/guides/working-with-mobx.md
+++ b/docs/extensions/guides/working-with-mobx.md
@@ -24,7 +24,7 @@ Further reading is available from `mobx`'s [website](https://mobx.js.org/the-gis
 
 ## Basic usage of mobx:
 
-When using `Lens`'s extension's API, some of the provided types are marked as `observable` (or others) and are documented as such.
+When using `Lens`'s extension's API, some of the provided types are marked as `observable` and are documented as such.
 These can be used as normal types and the combination of `mobx` and `react` work to determine when a rerender should occur.
 
 ---
@@ -37,12 +37,15 @@ That could be achieved using roughly the following code:
 ```typescript
 import { LensMainExtension, MenuRegistration } from "@k8slens/extensions";
 import observables from "./observables" // a collection of observable data
+import { IReactionDisposer } from "mobx";
 
 interface MenuRegistrationWithId extends MenuRegistration {
   id?: string;
 }
 
 export default class ExtensionMain extends LensMainExtension {
+  menuReactionDispose?: IReactionDisposer;
+
   appMenus: MenuRegistrationWithId[] = [
     {
       parentId: "file",
@@ -53,10 +56,8 @@ export default class ExtensionMain extends LensMainExtension {
     },
   ];
 
-  constructor() {
-    super()
-
-    reaction(
+  onActivate() {
+    this.menuReactionDispose = reaction(
       () => observables.clusterIsInState,
       clusterIsInState => {
         if (clusterIsInState) {
@@ -73,6 +74,11 @@ export default class ExtensionMain extends LensMainExtension {
         }
       }
     )
+  }
+
+  onDeactivate() {
+    this.menuReactionDispose?.() // always clenaup mobx disposers
+    this.menuReactionDispose = undefined
   }
 }
 ```
--- a/docs/extensions/guides/working-with-mobx.md
+++ b/docs/extensions/guides/working-with-mobx.md
@@ -7,11 +7,11 @@ This helps with having a more declarative style of managing state, as opposed to
 You should already have a basic understanding of how `React` handles state ([read here](https://reactjs.org/docs/faq-state.html) for more information).
 However, if you do not, here is a quick overview.
 
-A `React.Component` is generic over both `Props` and `State` (with default empty object types).
-`Props` should be considered read-only from the point of view of the component and is the mechanism for passing in "arguments" to a component.
-`State` is a component's internal state and can be read by accessing the parent field `state`.
-`State` **must** be updated using the `setState` parent method which merges the new data with the old state.
-`React` does do some optimizations around re-rendering components after quick successions of `setState` calls.
+- A `React.Component` is generic over both `Props` and `State` (with default empty object types).
+- `Props` should be considered read-only from the point of view of the component and is the mechanism for passing in "arguments" to a component.
+- `State` is a component's internal state and can be read by accessing the parent field `state`.
+- `State` **must** be updated using the `setState` parent method which merges the new data with the old state.
+- `React` does do some optimizations around re-rendering components after quick successions of `setState` calls.
 
 ## How mobx works:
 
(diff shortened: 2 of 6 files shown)


docs/extensions/guides/working-with-mobx.md
```diff
@@ -0,0 +1,78 @@
...
+`mobx` is a package that provides an abstraction over `React`'s state management. The three main concepts are:
+- `observable`: data stored in the component's `state`
+- `action`: a function that modifies any `observable` data
+- `computed`: data that is derived from `observable` data but is not actually stored. Think of this as computing `isEmpty` vs an `observable` field called `count`.
+
+Further reading is available from `mobx`'s [website](https://mobx.js.org/the-gist-of-mobx.html).
+
+## Basic usage of mobx:
+
+When using `Lens`'s extension's API, some of the provided types are marked as `observable` (or others) and are documented as such.
```
jim-docker: what do you mean by "(or others)"?

docs/extensions/guides/working-with-mobx.md
```diff
@@ -0,0 +1,78 @@
...
+  ];
+
+  constructor() {
+    super()
+
+    reaction(
+      () => observables.clusterIsInState,
+      clusterIsInState => {
+        if (clusterIsInState) {
+          this.appMenus.push({
```
jim-docker: What happens here? Does this trigger ExtensionLoader.init() by virtue of appMenus being an observable?

docs/extensions/guides/working-with-mobx.md
```diff
@@ -0,0 +1,78 @@
...
+    {
+      parentId: "file",
+      label: strings.extension.appMenu["label"](),
+      click() {
+        console.log("you clicked the label menu")
+      },
+    },
+  ];
+
+  constructor() {
```
jakolehm: Maybe `onActivate()` is better?

chenhunghan:
I am just dropping a comment here not reviewing, thanks @Nokel81 I think this help (Mobx noob like) me a lot.

docs/extensions/guides/working-with-mobx.md
Nokel81: I meant `observable.shallow` and friends. Which I guess are basically the same. I could just remove that remark.

docs/extensions/guides/working-with-mobx.md
Nokel81: That is what I gathered from looking around. Yes, that is what the observable implies.

docs/extensions/guides/working-with-mobx.md
jakolehm: Let's remove `(or others)`, it might confuse readers.

Nokel81: Sounds good

Nokel81:
fixes  #1375 (as per Jari's comment)

docs/extensions/guides/working-with-mobx.md
ixrock: 1. `ExtensionLoader.init()` is called only once when app is started

1. `this.appMenus.push()` probably won't trigger anything since observable only a link to array itself?
```ts
    this.autoInitExtensions((ext: LensMainExtension) => [
      registries.menuRegistry.add(ext.appMenus)
    ]);
```
Otherwise if it even works (since `menuRegistry.add` iterate over the array) and extension is enabled and already added some menus it should de-register items first before modification? Otherwise possible some duplications?

Nokel81: So this example doesn't work anyway?

src/extensions/extension-loader.ts
```diff
@@ -19,6 +19,9 @@ export class ExtensionLoader {
   protected extensions = observable.map<LensExtensionId, InstalledExtension>();
   protected instances = observable.map<LensExtensionId, LensExtension>();
 
+  /**
+   * @observable
```
ixrock: What does this `@jsdoc`?

docs/extensions/guides/working-with-mobx.md
ixrock: Probably.. (haven't time to check)

Nokel81: So maybe it is disingenuous to mark `appMenus` as `observable` then. Since we never actually "observe" the actions.

ixrock: Or we should add such support and make them truly observable? :)

Nokel81: I think we should, and it was asked for in #1375. If we do want to work on it, then this PR shouldn't be merged until after that work is done (or it doesn't matter if both are finished by 4.0.0)

src/extensions/extension-loader.ts
Nokel81: I don't understand your question. I added these comments so that these variable's observable status would be added to the docs.

Nokel81:
So this doesn't actually fix #1375 , but I have filled a bug #1432 (since the current behaviour is actually a bug)

Nokel81:
blocked by #1432

src/extensions/extension-loader.ts
ixrock: Alright, it's not clear where this code / comment actually participate :)

Nokel81: Would you prefer something more along the lines of `This field is mobx observable`?

ixrock: Since PR should be targeted on docs only? Otherwise why only in some `.ts` files here I see these comments?

Nokel81: Because comments are part of documentation

docs/extensions/guides/working-with-mobx.md
```diff
@@ -0,0 +1,84 @@
...
+
+When using `Lens`'s extension's API, some of the provided types are marked as `observable` and are documented as such.
+These can be used as normal types and the combination of `mobx` and `react` work to determine when a rerender should occur.
+
+---
+
+## Example:
+
+Imagine that your extension wants to conditionally add an app menu if some other observable value is ever `true` (and remove it when it is `false`).
+That could be achieved using roughly the following code:
```
jim-docker: Since this is an intro to using mobx in Lens, if you stick with this example, it might be a good idea to explain more of what exactly happens (once you figure it out :)). Like pushing a menu to appMenus causes what reaction exactly?
Otherwise try a more basic example?

Nokel81: That is probably a good idea. I think having an example like this (once the bug is fixed) would be very good, but to start with having a simpler example probably would be better.

docs/extensions/guides/working-with-mobx.md
```diff
@@ -0,0 +1,84 @@
...
+Lens uses `mobx` as its state manager on top of React's state management system.
+This helps with having a more declarative style of managing state, as opposed to `React`'s native `setState` mechanism.
+You should already have a basic understanding of how `React` handles state ([read here](https://reactjs.org/docs/faq-state.html) for more information).
+However, if you do not, here is a quick overview.
+
+A `React.Component` is generic over both `Props` and `State` (with default empty object types).
+`Props` should be considered read-only from the point of view of the component and is the mechanism for passing in "arguments" to a component.
+`State` is a component's internal state and can be read by accessing the parent field `state`.
+`State` **must** be updated using the `setState` parent method which merges the new data with the old state.
+`React` does do some optimizations around re-rendering components after quick successions of `setState` calls.
```
stefcameron: Seems like this should be a bulleted list?

stefcameron: Maybe not, just seemed like it. 🙂

docs/extensions/guides/working-with-mobx.md
```diff
@@ -0,0 +1,84 @@
...
+          this.appMenus.push({
+            parentId: "file",
+            label: "Remove from current state",
+            click() {
+              console.log("currently does nothing, removing...")
+            },
+            id: "clusterIsInState",
+          })
+        } else {
+          this.appMenus = this.appMenus.filter(m => x.id !== "clusterIsInState")
```
stefcameron: Should it be `m.id` instead of `x.id`?

stefcameron (commented):
@Nokel81 Thank you for this write-up! I think this will be really helpful to get started, and someone from the community has already commented as such, too!

One general observation about the docs (not just this, but docs in general) so far is that they have a 100% bias toward class-based components (from what I've seen so far). Personally, I haven't used class-based components for at least 1.5 years now. I only use functional components with hooks. So it would be nice to have some sample code sometimes that uses those instead of classes since the mapping is not straightforward, and I'm sure lots of folks coming to Lens will be coming from the same place. When I read the docs, I still have to go dig around to figure out, "ok, now how do I do this with hooks...?" 😄

docs/extensions/guides/working-with-mobx.md
Nokel81: Yup thanks for the check.

Nokel81:
@stefcameron Thanks for the quick review. Yes our docs are heavily tilted towards class components. That is probably because a vast super majority of our components are classes. 

Side note: this is my first React project so all I know are class components.

stefcameron:
> Side note: this is my first React project so all I know are class components.

Nice! Way to go jumping in! 🎉  Now let me broaden your horizons: https://reactjs.org/docs/hooks-intro.html 😄  I believe this is the direction the framework is taking in the long run. You can create your own hooks too, which is really useful. I created a few in my Lens extension.

Nokel81:
Would you all be able to take another look at this PR. I removed the example because it doesn't work and won't work for a decent while yet. We can add it back later.

docs/extensions/guides/README.md
```diff
@@ -19,10 +19,11 @@ Each guide or sample will include:
 | [Stores](stores.md) | |
 | [Components](components.md) | |
 | [KubeObjectListLayout](kube-object-list-layout.md) | |
+| [Working with mobx](working-with-mobx.md) | |
```
jakolehm, line 22: This probably should be included in menu/navigation?

Nokel81: I don't understand. I thought that is what this was doing.

jakolehm: I mean this: https://github.com/lensapp/lens/blob/master/mkdocs.yml#L31-L35

Nokel81: Then yes, will do.

Nokel81:
@jakolehm Added to navigation.

jakolehm (approved):

PR #1425, opened 2020-11-17
make targets running npm .PHONY

Nokel81:

I have found that locally `make dev` was no longer working for me (due to missing packages). I noticed that `make` wasn't running some of the dependencies on the `dev` make target and I saw that those targets should rightly be marked as `.PHONY` since they don't updated their names in the filesystem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,9 @@ endif
 binaries/client:
 	yarn download-bins
 
+.PHONY: node_modules
 node_modules:
-	yarn install --frozen-lockfile --verbose
+	yarn install --frozen-lockfile
 	yarn check --verify-tree --integrity
 
 static/build/LensDev.html:
@@ -63,9 +64,11 @@ else
 	yarn dist
 endif
 
+.PHONY: $(extension_node_modules)
 $(extension_node_modules):
 	cd $(@:/node_modules=) && npm install --no-audit --no-fund
 
+.PHONY: $(extension_dists)
 $(extension_dists): src/extensions/npm/extensions/dist
 	cd $(@:/dist=) && npm run build
 
--- a/Makefile
+++ b/Makefile
@@ -66,9 +66,8 @@ endif
 
 .PHONY: $(extension_node_modules)
 $(extension_node_modules):
-	cd $(@:/node_modules=) && npm install --no-audit --no-fund
+	cd $(@:/node_modules=) && yarn import && yarn install && rm yarn.lock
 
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -63,9 +64,11 @@ else
 	yarn dist
 endif
 
+.PHONY: $(extension_node_modules)
 $(extension_node_modules):
 	cd $(@:/node_modules=) && npm install --no-audit --no-fund
 
+.PHONY: $(extension_dists)
```
jakolehm: This makes no sense to me, phony will trigger `npm run build` always -> multiple times in CI. See #1407 .

Nokel81: Maybe I was wrong about that one. I know that the other two `.PHONY` rules are required but I am now less sure about this one. 

How about I remove this line and keep the other two?

jakolehm: `$(extension_node_modules):` is also super slow (because of npm).. IMHO phony is not good fit there. Top level phony is probably ok because `yarn install` is fast (not sure about `yarn check`).

Nokel81: Then we should use `yarn install` (and then `.gitignore` the yarn lock file). `.PHONY` is exactly what this is for because `make`'s rule about when to rebuild a dependency is entirely based on filesystem information.

Makefile
```diff
@@ -63,8 +64,9 @@ else
 	yarn dist
 endif
 
+.PHONY: $(extension_node_modules)
 $(extension_node_modules):
-	cd $(@:/node_modules=) && npm install --no-audit --no-fund
+	cd $(@:/node_modules=) && yarn import && yarn install && rm yarn.lock
```
jakolehm, line 69: `yarn import / install` feels hacky, maybe we can just add `--prefer-offline` to `npm install` and use `make -jX` to run these in parallel (it should not take too many seconds).

Nokel81: According to the docs there is no `--prefer-offline` for `npm install` (see https://docs.npmjs.com/cli/v6/commands/npm-install)

Nokel81:
@nevalla I still think that the caching is buggy, this PR has failed several times because it cannot find `webpack` in the `node_modules` of a specific extension. However, I have doesn't checked and all the extensions do have it as a listed `devDependency`.

Nokel81:
closing due to locally we can just run `make -B dev`. Though I still think that we should move away from `make` all together since it isn't really a good fit for JS/TS projects.

PR #1431, opened 2020-11-18
Make BaseStore abstract

Nokel81:
I believe that this should be the type signature for `BaseStore`, it is rather confusing that these methods are almost always overloaded (and in the single case where it wasn't it really should have been). 

I believe that this is a better structure for our code. This also raises the question as to if `LensExtension` should be `abstract` as well, so that developers know that they should and need to use/implement `onActivate` and `onDeactivate`. That would be a very good afforance in the design.


panuhorsmalahti (approved):
LGTM

jakolehm (approved):

PR #1438, opened 2020-11-19
Store prometheus cluster metadata based on metrics request responses

nevalla:
This PR will store prometheus related data to the cluster metadata:
```
"prometheus": {
  "provider": "lens"|"operator"|"helm"|"stacklight",
  "autoDetected": true|false,
  "success": true|false
},
```

Previously if no provider detected, `lens` provider was returned. Now `null` is returned and it's handled respectively. Also, prometheus information is now re-initialised when cluster preferences are changed.


src/main/routes/metrics-route.ts
```diff
@@ -22,11 +23,8 @@ async function loadMetrics(promQueries: string[], cluster: Cluster, prometheusPa
         try {
           return await cluster.getMetrics(prometheusPath, { query, ...queryParams })
         } catch (error) {
-          if (lastAttempt || error?.statusCode === 404) {
-            return {
-              status: error.toString(),
-              data: { result: [] },
-            }
+          if (lastAttempt || (error?.statusCode >= 400 && error?.statusCode < 500)) {
```
nevalla, line 27: Fail fast if getting any error response

src/main/cluster.ts
```diff
@@ -138,6 +139,7 @@ export class Cluster implements ClusterModel, ClusterState {
     if (ipcMain) {
       this.eventDisposers.push(
         reaction(() => this.getState(), () => this.pushState()),
+        reaction(() => this.preferences, () => this.contextHandler.setupPrometheus(this.preferences)),
```
ixrock: Are you planning to react to any changes of `this.preference.*` internals? (or only when whole preferences object overwritten?) Otherwise probably `toJS(this.preferences, {recurseEverything: true})`?

nevalla: basically I'm interested only in `prometheus` and `prometheusProvider`. Any easy way react only changes in those two properties?

src/main/routes/metrics-route.ts
```diff
@@ -36,20 +34,25 @@ async function loadMetrics(promQueries: string[], cluster: Cluster, prometheusPa
...
     try {
       const [prometheusPath, prometheusProvider] = await Promise.all([
         cluster.contextHandler.getPrometheusPath(),
         cluster.contextHandler.getPrometheusProvider()
       ])
-
+      prometheusMetadata["provider"] = prometheusProvider?.id
+      prometheusMetadata["autoDetected"] = !cluster.preferences.prometheusProvider?.type
+      if (!prometheusPath) {
+        prometheusMetadata["success"] = false
```
ixrock: Why access  to `prometheusMetadata` via `["fieldName"]`? Some intention?

nevalla: just leftovers when I did not have proper type for prometheusMetadata. Now it's there so I will change those. Thanks for noticing!

src/main/cluster.ts
ixrock: Maybe something like this?

_cluster-store.ts_
```ts
export interface ClusterPreferences extends ClusterPrometheusPreferences {
  terminalCWD?: string;
  clusterName?: string;
  iconOrder?: number;
  icon?: string;
  httpsProxy?: string;
}

export interface ClusterPrometheusPreferences {
  prometheus?: {
    namespace: string;
    service: string;
    port: number;
    prefix: string;
  };
  prometheusProvider?: {
    type: string;
  };
}
```

_cluster.ts_
```ts
  @computed get prometheusPreferences(): ClusterPrometheusPreferences {
    const { prometheus, prometheusProvider } = this.preferences;
    return toJS({ prometheus, prometheusProvider }, {
      recurseEverything: true,
    })
  }
// and just react to computed prop?
reaction(() => this.prometheusPreferences, prefs => console.log(prefs))
```

nevalla: Thanks got it working with small modifications!

src/main/context-handler.ts
```diff
@@ -52,13 +57,7 @@ export class ContextHandler {
...
     });
     const resolvedPrometheusServices = await Promise.all(prometheusPromises);
-    const service = resolvedPrometheusServices.filter(n => n)[0];
-    return service || {
-      id: "lens",
-      namespace: "lens-metrics",
-      service: "prometheus",
-      port: 80
-    };
+    return resolvedPrometheusServices.filter(n => n)[0];
```
Nokel81, line 60: Do we want to return `undefined` if there are no services?

src/main/routes/metrics-route.ts
```diff
@@ -1,7 +1,9 @@
+
```
Nokel81:
```suggestion
```

src/main/routes/metrics-route.ts
```diff
@@ -22,11 +24,8 @@ async function loadMetrics(promQueries: string[], cluster: Cluster, prometheusPa
         try {
           return await cluster.getMetrics(prometheusPath, { query, ...queryParams });
         } catch (error) {
-          if (lastAttempt || error?.statusCode === 404) {
-            return {
-              status: error.toString(),
-              data: { result: [] },
-            };
+          if (lastAttempt || (error?.statusCode >= 400 && error?.statusCode < 500)) {
+            throw new Error("Metrics not available");
```
Nokel81, line 29:
```suggestion
            logger.error("[Metrics]: metrics not available", { err });
            throw new Error("Metrics not available");
```

src/main/context-handler.ts
nevalla: yes

Nokel81 (approved):
looks good to me

Issue #1347, opened 2020-11-12
Ingress: support networking.k8s.io/v1


jakolehm:
Add support for stable ingress api, `networking.k8s.io/v1`.

panuhorsmalahti:
Fixed in #1439

PR #1439, opened 2020-11-19
Support networking.k8s.io/v1 for Ingress

panuhorsmalahti:
- Tested on Kubernetes 1.19, 1.18 and 1.17 with both the old and new version of ingress resource
- Adds supports for specifying a fallback API base for the API in case the latest/default is not supported in the Kubernetes version of the cluster

### DIFF
--- a/src/renderer/api/kube-api.ts
+++ b/src/renderer/api/kube-api.ts
@@ -162,7 +162,7 @@ export class KubeApi<T extends KubeObject = any> {
       return {
         apiPrefix: this.apiPrefix,
         apiGroup: this.apiGroup
-      }
+      };
     }
 
     throw new Error(`Can't find working API for the Kubernetes resource ${this.apiResource}`);
--- a/src/renderer/api/kube-api.ts
+++ b/src/renderer/api/kube-api.ts
@@ -8,10 +8,22 @@ import { apiKube } from "./index";
 import { kubeWatchApi } from "./kube-watch-api";
 import { apiManager } from "./api-manager";
 import { createKubeApiURL, parseKubeApi } from "./kube-api-parse";
-import { apiKubePrefix, isDevelopment } from "../../common/vars";
+import { apiKubePrefix, isDevelopment, isTestEnv } from "../../common/vars";
 
 export interface IKubeApiOptions<T extends KubeObject> {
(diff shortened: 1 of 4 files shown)


src/renderer/api/kube-api.ts
```diff
@@ -8,10 +8,17 @@ import { apiKube } from "./index";
 import { kubeWatchApi } from "./kube-watch-api";
 import { apiManager } from "./api-manager";
 import { createKubeApiURL, parseKubeApi } from "./kube-api-parse";
-import { apiKubePrefix, isDevelopment } from "../../common/vars";
+import { apiKubePrefix, isDevelopment, isTestEnv } from "../../common/vars";
 
 export interface IKubeApiOptions<T extends KubeObject> {
   apiBase?: string; // base api-path for listing all resources, e.g. "/api/v1/pods"
+
+  // If the API uses a different API endpoint (e.g. apiBase) depending on the cluster version,
```
jakolehm: Maybe use `tsdoc` format here (would appear in generated extension api docs)?

panuhorsmalahti: Fixed, although I'm not confident it will end up in the API Docs (it seems the interface is not listed there, just referenced).

I also didn't find clear information if documenting interfaces is even supported in TSDoc at the moment.

panuhorsmalahti:
Fixes https://github.com/lensapp/lens/issues/1347

jakolehm:
@panuhorsmalahti please fix conflicts.

src/renderer/components/+network-ingresses/ingress-details.tsx
```diff
@@ -100,6 +103,10 @@ export class IngressDetails extends React.Component<Props> {
       <Trans>Network</Trans>,
       <Trans>Duration</Trans>,
     ];
+
+    const serviceName = spec?.defaultBackend?.service.name ?? spec?.backend?.serviceName;
```
jakolehm: Should `Ingress` class have helper methods for `serviceName` and `servicePort`?

panuhorsmalahti: Good idea, added

jakolehm (approved):
LGTM 👍

PR #1054, opened 2020-10-08
Open last active cluster view when switching between workspaces

stevejr:
This enhancement will store the last active cluster used in a workspace. If you then switch to a different
workspace and switch back the cluster view will be restored.

If a cluster is disconnected then it will not become active after switching back to the workspace - only an active cluster in the workspace will be "re-activated" when switching back to the workspace.

Signed-off-by: Steve Richards <srichards@mirantis.com>

fixes #827

PR #1444, opened 2020-11-19
Open last active cluster after switching workspaces

### LINKED ISSUE #827: Dashboard doesn't refresh after changing workspace
**Describe the bug**
When changing from one workspace to another the dashboard is "blank" (there is no current cluster shown) 

**To Reproduce**
1. With two workspaces, click on workspace widget (bottom left) and change to the other workspace
2. Observe that the workspace is shown, with no active cluster(even if there was an active cluster when the workspace was last selected)
3. Click on a cluster icon to cause the refresh

**Expected behavior**
When toggling between workspaces the dashboard should show previous snapshot of the workspace
**Screenshots**

wsRefresh

- Lens Version: 3.6.0-rc.1
- OS: OSX
- Installation method (e.g. snap or AppImage in Linux): dmg

aleksfront:
Based on https://github.com/lensapp/lens/pull/1054/files

Saves last active cluster for each workspace and opens it when workspaces switched.

switching workspaces

Fixes #827


src/common/workspace-store.ts
```diff
@@ -184,6 +187,13 @@ export class WorkspaceStore extends BaseStore<WorkspaceStoreModel> {
     clusterStore.removeByWorkspaceId(id)
   }
 
+  @action
+  setLastActiveClusterId(clusterId: ClusterId, workspaceId = this.currentWorkspaceId) {
+    if (clusterId != null) {
```
Nokel81: This possibility should be marked in the signature

src/renderer/components/+workspaces/workspace-menu.tsx
```diff
@@ -17,6 +20,16 @@ interface Props extends Partial<MenuProps> {
 export class WorkspaceMenu extends React.Component<Props> {
   @observable menuVisible = false;
 
+  activateWorkspace = (id: WorkspaceId) => {
+    const clusterId = workspaceStore.getById(id).lastActiveClusterId;
```
Nokel81, line 24: What happens when there is no `lastActiveClusterId`?

src/renderer/components/+workspaces/workspace-menu.tsx
```diff
@@ -17,6 +20,16 @@ interface Props extends Partial<MenuProps> {
...
+    workspaceStore.setActive(id);
+    if (clusterId) {
+      navigate(clusterViewURL({ params: { clusterId } }));
+    } else {
+      navigate(landingURL());
+    }
```
Nokel81, line 30: Why is this code different from the code in `workspaces.tsx`?

src/renderer/components/cluster-manager/clusters-menu.tsx
```diff
@@ -57,7 +57,11 @@ export class ClustersMenu extends React.Component<Props> {
       menu.append(new MenuItem({
         label: _i18n._(t`Disconnect`),
         click: async () => {
+          const lastActiveClusterId = workspaceStore.currentWorkspace.lastActiveClusterId;
           if (clusterStore.isActive(cluster.id)) {
+            if (lastActiveClusterId === cluster.id) {
+              workspaceStore.setLastActiveClusterId("");
```
Nokel81: This is very weird, `""` is not a valid `ClusterId` (so some function might throw if this gets out). There should be a better way to represent this (`undefined` maybe given that `lastActiveClusterId` is optional.

src/renderer/components/+workspaces/workspace-menu.tsx
aleksfront: If `lastActiveCluster` id became `undefined` then user will be redirected to Landing page following code below.

src/renderer/components/+workspaces/workspace-menu.tsx
aleksfront: When you hit on workspace link from Workspaces page, there's no redirect to dashboard - user will stay on same page and see that `current` indicator will jump to clicked workspace.

src/renderer/components/cluster-manager/clusters-menu.tsx
aleksfront: Good catch! Refactored this part. A lot of code removed!

Nokel81:
Rerequest my review when you are ready for it

aleksfront:
> Rerequest my review when you are ready for it

It's ready, except for failed tests, need to fix them. Please take a second look.

src/common/workspace-store.ts
```diff
@@ -7,10 +7,12 @@ import { broadcastMessage } from "../common/ipc";
 import logger from "../main/logger";
 
 export type WorkspaceId = string;
+export type ClusterId = string;
 
 export interface WorkspaceStoreModel {
+  workspaces: WorkspaceModel[];
   currentWorkspace?: WorkspaceId;
-  workspaces: WorkspaceModel[]
+  lastActiveClusterId?: ClusterId;
```
Nokel81: Wait, does this mean that we only track the last active cluster once for the whole application (and per workspace)?

src/common/workspace-store.ts
```diff
@@ -7,10 +7,12 @@ import { broadcastMessage } from "../common/ipc";
 import logger from "../main/logger";
 
 export type WorkspaceId = string;
+export type ClusterId = string;
```
Nokel81: This should be imported from some place else

src/common/workspace-store.ts
Nokel81: Or are we only saving the last active cluster globally (still seems odd)?

aleksfront: We're tracking last active cluster for each workspace. This field belongs to `Workspace` class.

src/common/workspace-store.ts
aleksfront: Fixed

src/common/workspace-store.ts
Nokel81: See my other (sniped) comment

aleksfront: You mean this one? https://github.com/lensapp/lens/pull/1444#discussion_r526988672
Isn't it about same question about clarification of where `lastActiveClusterId` field is located?

Nokel81: It is, I noticed that I had asked about the one located on `WorkspaceStoreModel` which seems strange to me. I would have though that it should be been on `WorskaceModel`

aleksfront: Oh, that's true. Moved into `WorkspaceModel`.

Nokel81 (approved):
Looks good!

src/common/workspace-store.ts
ixrock: It's already defined in `cluster-store.ts` as:
```
export type ClusterId = string;
```
So just `import type {ClusterId} from "../cluster-store.ts"`

src/common/cluster-store.ts
```diff
@@ -11,9 +11,10 @@ import { appEventBus } from "./event-bus"
 import { dumpConfigYaml } from "./kube-helpers";
 import { saveToAppFiles } from "./utils/saveToAppFiles";
 import { KubeConfig } from "@kubernetes/client-node";
+import { subscribeToBroadcast, unsubscribeAllFromBroadcast } from "./ipc";
 import _ from "lodash";
 import move from "array-move";
-import { subscribeToBroadcast, unsubscribeAllFromBroadcast } from "./ipc";
+import type { WorkspaceId } from "./workspace-store";
```
ixrock, line 17: You already importing it as normal module on the top => so no need to for extra import?

src/renderer/components/+workspaces/workspace-menu.tsx
ixrock: Maybe it's better to select first cluster in workspace and show it instead landing page? If no clusters -> landing?

src/common/cluster-store.ts
aleksfront: If not imported separately as a `type`, webpack warnings will appear in compilation time.

PR #1452, opened 2020-11-19
Allow extension cluster menus to have a parent

### LINKED ISSUE #1442: Possibility to register cluster sub-menus
Currently extension-api allows only to register top-level menu items to cluster menu. Extension-api should allow only registration of sub-items.

jakolehm:
- introduces `id` and `parentId` properties (both optional)
- TabLayout is created automatically if cluster menu item has sub-items

Example:

```ts
    clusterPageMenus = [
    {
      id: "example",
      title: "Example Extension",
      components: {
        Icon: ExampleIcon,
      }
    },
    {
      parentId: "example",
      target: { pageId: "example1", params: {} },
      title: "Sub menu 1",
      components: {
        Icon: ExampleIcon,
      }
    },
    {
      parentId: "example",
      target: { pageId: "example2", params: {} },
      title: "Sub menu 2",
      components: {
        Icon: ExampleIcon,
      }
    }
  ]
```

produces:

image

Fixes #1442


src/renderer/components/app.tsx
```diff
@@ -72,9 +75,58 @@ export class App extends React.Component {
...
+    });
+    if (routes.length > 0) {
+      routes.unshift({
+        routePath: page.routePath,
+        url: getExtensionPageUrl({ extensionId: page.extensionId, pageId: page.id, params: menuItem.target.params }),
+        title: <Trans>Overview</Trans>,
+        component: page.components.Page,
+        exact: page.exact
+      })
+    }
```
Nokel81: Why do we force an `Overview` page, that seems weird.

jakolehm: See the attached screenshot. Extension basically does not know if something registers as a sub-menu... it's better to create consistent UI with "Overview" than duplicate top-level menu name in a first sub-menu item.

Nokel81: Wouldn't the API be better then if it does know if it registers as a sub menu? I see two main problems: one, is that there is no way to change that title text; two, there doesn't seem to be a way to remove this, if the extension developer doesn't want it.

jakolehm: Example:
- extension A registers a menu "Foo"
- extension B registers a sub-menu "Bar" to "Foo"

Nokel81: Okay, so then have a tab layout of size 1.

Nokel81: Also, I don't see any handling of the following case:

- extension A registers a menu "Foo"
- extension B registers a sub-menu "Bar" to "Fooo" (by accident). 

should an error be reported (somewhere)?

jakolehm: > Okay, so then have a tab layout of size 1.

Not sure if I understand your point. Our menu does not really work like that?

jim-docker: But other (builtin) submenus don't have an overview page

What about something like:
```
  clusterPageMenus = [
    {
      target: { submenu: "example" },
      title: "Example Extension",
      components: {
        Icon: ExampleIcon,
      }
    },
    {
      parentId: "example",
      target: { pageId: "example1", params: {} },
      title: "Overview",
      components: {
        Icon: ExampleIcon,
      }
    },
    {
      parentId: "example",
      target: { pageId: "example2", params: {} },
      title: "Sub menu 1",
      components: {
        Icon: ExampleIcon,
      }
    }
  ]
```

Nokel81: I agree with Jim, this shouldn't be "automagically" added. And with regards to the tab layout not working like that. It seems very strange that the tab layout should care how many entries it has at all.

jakolehm: Ok, based on discussions (on Slack) I'll limit scope of this PR to work only within an extension. We can improve (and introduce more possibilities) on follow up PRs.

Nokel81: Sounds good

jakolehm:
@Nokel81 @jim-docker PTAL (example in description is updated to match current logic)

src/renderer/components/app.tsx
```diff
@@ -72,9 +75,48 @@ export class App extends React.Component {
     return workloadsURL();
   }
 
+  getTabLayoutRoutes(menuItem: ClusterPageMenuRegistration) {
+    const routes: TabLayoutRoute[] = [];
+    if (!menuItem.id) {
+      return routes;
+    }
+    clusterPageMenuRegistry.getSubItems(menuItem).forEach((item) => {
+      const page = clusterPageRegistry.getByPageMenuTarget(item.target);
```
Nokel81, line 84: There doesn't seem to be an restrictions here about registering an item to a different extension.

src/renderer/components/layout/sidebar.tsx
```diff
@@ -76,6 +76,29 @@ export class Sidebar extends React.Component<Props> {
...
+      return [];
+    }
+    const routes: TabLayoutRoute[] = [];
+
+    clusterPageMenuRegistry.getSubItems(menu).forEach((subItem) => {
+      const subPage = clusterPageRegistry.getByPageMenuTarget(subItem.target);
+      console.log(subItem, subPage)
+      if (subPage) {
+        routes.push({
+          routePath: subPage.id,
```
jim-docker: should this be `subPage.routePath`?

src/renderer/components/layout/sidebar.tsx
```diff
@@ -76,6 +76,29 @@ export class Sidebar extends React.Component<Props> {
     });
   }
 
+  getTabLayoutRoutes(menu: ClusterPageMenuRegistration, page: RegisteredPage): TabLayoutRoute[] {
```
jim-docker: `page` is not used

src/renderer/components/app.tsx
jakolehm: Fixed.

src/renderer/components/layout/sidebar.tsx
jakolehm: Yes, good catch!

src/renderer/components/layout/sidebar.tsx
jakolehm: Fixed.

src/renderer/components/layout/sidebar.tsx
```diff
@@ -76,6 +76,29 @@ export class Sidebar extends React.Component<Props> {
...
+  getTabLayoutRoutes(menu: ClusterPageMenuRegistration): TabLayoutRoute[] {
+    console.log("Menu", menu)
+    if (!menu.id) {
+      return [];
+    }
+    const routes: TabLayoutRoute[] = [];
+
+    clusterPageMenuRegistry.getSubItems(menu).forEach((subItem) => {
+      const subPage = clusterPageRegistry.getByPageMenuTarget(subItem.target);
+      console.log(subItem, subPage)
```
Nokel81: please remove
```suggestion
```

src/renderer/components/layout/sidebar.tsx
```diff
@@ -76,6 +76,29 @@ export class Sidebar extends React.Component<Props> {
     });
   }
 
+  getTabLayoutRoutes(menu: ClusterPageMenuRegistration): TabLayoutRoute[] {
+    console.log("Menu", menu)
```
Nokel81:
```suggestion
```

src/renderer/components/layout/sidebar.tsx
```diff
@@ -191,17 +212,25 @@ export class Sidebar extends React.Component<Props> {
...
               return (
                 <SidebarNavItem
                   key={pageUrl} url={pageUrl}
-                  text={title} icon={<Icon/>}
+                  text={menuItem.title} icon={<menuItem.components.Icon/>}
                   isActive={isActive}
+                  subMenus={tabRoutes}
                 />
               );
             })}
```
Nokel81: Last thing, probably minor. But this is getting rather long so (imo) it should be in its own render function. Maybe `renderRegisteredMenus`?

src/renderer/components/layout/sidebar.tsx
```diff
@@ -191,17 +212,25 @@ export class Sidebar extends React.Component<Props> {
             >
               {this.renderCustomResources()}
```
Nokel81, line 238: I just noticed that there is no place here that renders sub menus registered to Lens native menus. Is that still on the table for this PR?

jakolehm: Out-of-scope, it needs more refactoring of Lens internals.

Nokel81: Makes sense

src/renderer/components/layout/sidebar.tsx
Nokel81: If not, I will approve

jakolehm: Fixed.

Nokel81 (approved):

PR #1078, opened 2020-10-13
Pre-Validate kubeconfig before making available in Lens

stevejr:
Fixes: #843

Signed-off-by: Steve Richards <srichards@mirantis.com>

PR #1453, opened 2020-11-19
mark uncreatable Cluster objects in cluster-store as dead

### LINKED ISSUE #1316: kube.yml exception handling, file has been removed or is not found
**Describe the bug**
If the kube.yml is deleted or is no longer reachable Lens is not aware.  It appears to try to connect to an endpoint that has no configuration.

**To Reproduce**
1. Add cluster using `kube.yml`
2. Click on newly added cluster to verify connectivity
3. Close Lens and it should gracefully quit
4. While Lens is not running, delete the `kube.yml`
5. Re-launch cluster and click on the cluster recently added
6. Observe the permanent connecting graphic, it should never get past this screen
7. Quit Lens it will not quit
8. Try adding a new cluster File->Add Cluster (see screenshot for java error)
9. Force quit Lens
10. Re-launch Lens
11. Remove cluster associated with missing kube.yml
12. Quit->Force Quit
13. Re-launch Lens and all is well again

**Expected behavior**
Expect that Lens will have a defense for situations where the kube.yml is not found/reachable and alert the user of the condition. This condition should not impact the usage for the app or other clusters.

**Screenshots**

connecting

java_error_add_cluster

- Lens Version: 3.8.6
- OS: [e.g. OSX]  MacOS 10.15.7 (19H15)
- Installation method (e.g. snap or AppImage in Linux): Image
 
**Logs:**
```
Uncaught Exception:
TypeError: Object has been destroyed
    at De.navigate (/Applications/Lens.app/Contents/Resources/app.asar/static/build/main.js:1:110727)
    at s (/Applications/Lens.app/Contents/Resources/app.asar/static/build/main.js:1:107355)
    at click (/Applications/Lens.app/Contents/Resources/app.asar/static/build/main.js:1:108106)
    at MenuItem.click (electron/js2c/browser_init.js:65:1624)
    at Function.executeCommand (electron/js2c/browser_init.js:73:1102)
```
**Kubeconfig:**
```
N/A
```

**Additional context**
In my case I upgraded to Catalina OS and the security file access permisions changes were most likely the cause IMHO.

### LINKED ISSUE #1561: Error in context will prevent Lens from opening
**Describe the bug**
When there's an incomplete context in .kube/config file, Lens won't start up (no splash screen)

**To Reproduce**
Create a new context with kubectl, do not specify the whole info

example: kubectl config set-context aksdemo

Then open Lens

**Expected behavior**
To provide a message to the user or to ignore the error, but not to freeze without any output, any user may go crazy

**Screenshots**
No screen shot but this is what happens when you open in debug mode

~ $ bash
bash-5.0$ DEBUG=true
bash-5.0$ /Applications/Lens.app/Contents/MacOS/Lens
info: [IPC]: setup channel "cluster:activate"
info: [IPC]: setup channel "cluster:set-frame-id"
info: [IPC]: setup channel "cluster:refresh"
info: [IPC]: setup channel "cluster:disconnect"
info: [IPC]: setup channel "cluster:install-feature"
info: [IPC]: setup channel "cluster:uninstall-feature"
info: [IPC]: setup channel "cluster:upgrade-feature"
info: 🚀 Starting Lens from "/Users/gabriel/Library/Application Support/Lens"
info: Checking for update
info: [STORE]: LOADED from /Users/gabriel/Library/Application Support/Lens/config.json
info: [STORE]: LOADED from /Users/gabriel/Library/Application Support/Lens/lens-cluster-store.json
info: [STORE]: LOADED from /Users/gabriel/Library/Application Support/Lens/lens-workspace-store.json
(node:6931) UnhandledPromiseRejectionWarning: Error: contexts[3].context.cluster is missing
    at /Applications/Lens.app/Contents/Resources/app.asar/node_modules/@kubernetes/client-node/dist/config_types.js:111:19
    at Function.map (/Applications/Lens.app/Contents/Resources/app.asar/node_modules/@kubernetes/client-node/node_modules/underscore/underscore.js:207:24)
    at Object.newContexts (/Applications/Lens.app/Contents/Resources/app.asar/node_modules/@kubernetes/client-node/dist/config_types.js:92:14)
    at KubeConfig.loadFromString (/Applications/Lens.app/Contents/Resources/app.asar/node_modules/@kubernetes/client-node/dist/config.js:106:40)
    at KubeConfig.loadFromFile (/Applications/Lens.app/Contents/Resources/app.asar/node_modules/@kubernetes/client-node/dist/config.js:75:14)
    at m (/Applications/Lens.app/Contents/Resources/app.asar/static/build/main.js:1:40656)
    at V.getKubeconfig (/Applications/Lens.app/Contents/Resources/app.asar/static/build/main.js:1:17503)
    at new V (/Applications/Lens.app/Contents/Resources/app.asar/static/build/main.js:1:14732)
    at de.fromStore (/Applications/Lens.app/Contents/Resources/app.asar/static/build/main.js:1:30480)
    at executeAction (/Applications/Lens.app/Contents/Resources/app.asar/node_modules/mobx/lib/mobx.js:927:19)
(node:6931) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:6931) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
error: contexts[3].context.cluster is missing
info: Update for version 3.6.9 is not available (latest version: 3.6.9, downgrade is disallowed).
^Cbash-5.0$

Only after control C you regain control, no window is open.

- Lens Version: 3.6.9
- OS: [e.g. OSX] OSX Big Sur 11.0.1
- Installation method (e.g. snap or AppImage in Linux): brew install lens --cask
 
**Logs:**
```
~ $ bash
bash-5.0$ DEBUG=true
bash-5.0$ /Applications/Lens.app/Contents/MacOS/Lens
info: [IPC]: setup channel "cluster:activate"
info: [IPC]: setup channel "cluster:set-frame-id"
info: [IPC]: setup channel "cluster:refresh"
info: [IPC]: setup channel "cluster:disconnect"
info: [IPC]: setup channel "cluster:install-feature"
info: [IPC]: setup channel "cluster:uninstall-feature"
info: [IPC]: setup channel "cluster:upgrade-feature"
info: 🚀 Starting Lens from "/Users/gabriel/Library/Application Support/Lens"
info: Checking for update
info: [STORE]: LOADED from /Users/gabriel/Library/Application Support/Lens/config.json
info: [STORE]: LOADED from /Users/gabriel/Library/Application Support/Lens/lens-cluster-store.json
info: [STORE]: LOADED from /Users/gabriel/Library/Application Support/Lens/lens-workspace-store.json
(node:6931) UnhandledPromiseRejectionWarning: Error: contexts[3].context.cluster is missing
    at /Applications/Lens.app/Contents/Resources/app.asar/node_modules/@kubernetes/client-node/dist/config_types.js:111:19
    at Function.map (/Applications/Lens.app/Contents/Resources/app.asar/node_modules/@kubernetes/client-node/node_modules/underscore/underscore.js:207:24)
    at Object.newContexts (/Applications/Lens.app/Contents/Resources/app.asar/node_modules/@kubernetes/client-node/dist/config_types.js:92:14)
    at KubeConfig.loadFromString (/Applications/Lens.app/Contents/Resources/app.asar/node_modules/@kubernetes/client-node/dist/config.js:106:40)
    at KubeConfig.loadFromFile (/Applications/Lens.app/Contents/Resources/app.asar/node_modules/@kubernetes/client-node/dist/config.js:75:14)
    at m (/Applications/Lens.app/Contents/Resources/app.asar/static/build/main.js:1:40656)
    at V.getKubeconfig (/Applications/Lens.app/Contents/Resources/app.asar/static/build/main.js:1:17503)
    at new V (/Applications/Lens.app/Contents/Resources/app.asar/static/build/main.js:1:14732)
    at de.fromStore (/Applications/Lens.app/Contents/Resources/app.asar/static/build/main.js:1:30480)
    at executeAction (/Applications/Lens.app/Contents/Resources/app.asar/node_modules/mobx/lib/mobx.js:927:19)
(node:6931) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:6931) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
error: contexts[3].context.cluster is missing
info: Update for version 3.6.9 is not available (latest version: 3.6.9, downgrade is disallowed).
^Cbash-5.0$
```

**Kubeconfig:**
```
a context exists in kubeconfig, but its context.cluster is missing
```
That was the problem, context existing with no details. Don't expect Lens to freeze without of that, just output an error.

**Additional context**

Nokel81:
- Display a special error for each of the currently 2 error types from
  loading the kubeconfig

- Special handling of dead clusters in proxy handler so that there are
  no unhandled promise rejections

- add a rule to mark switch case statements as should be indented one level.

fixes #1316

### DIFF
--- a/extensions/telemetry/src/tracker.ts
+++ b/extensions/telemetry/src/tracker.ts
@@ -9,7 +9,7 @@ import { comparer } from "mobx";
 export class Tracker extends Util.Singleton {
   static readonly GA_ID = "UA-159377374-1";
   static readonly SEGMENT_KEY = "YENwswyhlOgz8P7EFKUtIZ2MfON7Yxqb";
-  protected eventHandlers: Array<(ev: EventBus.AppEvent ) => void> = [];
+  protected eventHandlers: Array<(ev: EventBus.AppEvent) => void> = [];
   protected started = false;
   protected visitor: ua.Visitor;
   protected analytics: Analytics;
@@ -124,30 +124,28 @@ export class Tracker extends Util.Singleton {
   }
 
   protected resolveOS() {
-    let os = "";
     if (App.isMac) {
-      os = "MacOS";
-    } else if(App.isWindows) {
-      os = "Windows";
-    } else if (App.isLinux) {
-      os = "Linux";
-      if (App.isSnap) {
-        os += "; Snap";
-      } else {
-        os += "; AppImage";
-      }
-    } else {
-      os = "Unknown";
+      return "MacOS";
     }
-    return os;
+
+    if (App.isWindows) {
+      return "Windows";
+    }
+
+    if (App.isLinux) {
+      return `Linux; ${App.isSnap ? "Snap" : "AppImage"}`;
+    }
+
+    return "Unknown";
   }
 
   protected async event(eventCategory: string, eventAction: string, otherParams = {}) {
+    const allowed = await this.isTelemetryAllowed();
+    if (!allowed) {
+      return;
+    }
+
     try {
-      const allowed = await this.isTelemetryAllowed();
-      if (!allowed) {
-        return;
-      }
       this.visitor.event({
         ec: eventCategory,
         ea: eventAction,
--- a/extensions/telemetry/src/tracker.ts
+++ b/extensions/telemetry/src/tracker.ts
@@ -9,7 +9,7 @@ import { comparer } from "mobx";
 export class Tracker extends Util.Singleton {
   static readonly GA_ID = "UA-159377374-1";
   static readonly SEGMENT_KEY = "YENwswyhlOgz8P7EFKUtIZ2MfON7Yxqb";
-  protected eventHandlers: Array<(ev: EventBus.AppEvent) => void> = [];
+  protected eventHandlers: Array<(ev: EventBus.AppEvent ) => void> = [];
   protected started = false;
   protected visitor: ua.Visitor;
   protected analytics: Analytics;
@@ -124,28 +124,30 @@ export class Tracker extends Util.Singleton {
   }
 
   protected resolveOS() {
+    let os = "";
     if (App.isMac) {
-      return "MacOS";
+      os = "MacOS";
+    } else if(App.isWindows) {
+      os = "Windows";
+    } else if (App.isLinux) {
+      os = "Linux";
+      if (App.isSnap) {
+        os += "; Snap";
+      } else {
+        os += "; AppImage";
+      }
+    } else {
+      os = "Unknown";
     }
-
-    if (App.isWindows) {
-      return "Windows";
-    }
-
-    if (App.isLinux) {
-      return `Linux; ${App.isSnap ? "Snap" : "AppImage"}`;
-    }
-
-    return "Unknown";
+    return os;
   }
 
   protected async event(eventCategory: string, eventAction: string, otherParams = {}) {
-    const allowed = await this.isTelemetryAllowed();
-    if (!allowed) {
-      return;
-    }
-
     try {
+      const allowed = await this.isTelemetryAllowed();
+      if (!allowed) {
+        return;
+      }
       this.visitor.event({
         ec: eventCategory,
         ea: eventAction,
--- a/src/common/cluster-store.ts
+++ b/src/common/cluster-store.ts
@@ -202,7 +202,7 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
     return this.deadClusters.get(id);
   }
 
-  getByWorkspaceId(workspaceId: string): ClusterRenderInfo[] {
+  getRendererInfoByWorkspace(workspaceId: string): ClusterRenderInfo[] {
     const aliveClusters: ClusterRenderInfo[] = this.clustersList.filter(c => c.workspace === workspaceId);
     const deadClusters: ClusterRenderInfo[] = this.deadClustersList
       .filter(([c]) => c.workspace === workspaceId)
@@ -217,6 +217,11 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
 
 
     return _.sortBy([...aliveClusters, ...deadClusters], c => c.preferences?.iconOrder);
+  }
+
+  getByWorkspaceId(workspaceId: string): Cluster[] {
+    const clusters = this.clustersList.filter(c => c.workspace === workspaceId);
+    return _.sortBy(clusters, c => c.preferences?.iconOrder);
   }
 
   getCountInWorkspace(workspaceId: string): number {
--- a/src/common/cluster-store.ts
+++ b/src/common/cluster-store.ts
@@ -82,13 +82,13 @@ export interface ClusterRenderInfo extends ClusterModel {
   online: boolean;
 }
 
-export class LensClusterStoreImpl extends BaseStore<ClusterStoreModel> {
+export class ClusterStore extends BaseStore<ClusterStoreModel> {
   static getCustomKubeConfigPath(clusterId: ClusterId): string {
     return path.resolve((app || remote.app).getPath("userData"), "kubeconfigs", clusterId);
   }
 
   static embedCustomKubeConfig(clusterId: ClusterId, kubeConfig: KubeConfig | string): string {
-    const filePath = LensClusterStoreImpl.getCustomKubeConfigPath(clusterId);
+    const filePath = ClusterStore.getCustomKubeConfigPath(clusterId);
     const fileContents = typeof kubeConfig == "string" ? kubeConfig : dumpConfigYaml(kubeConfig);
     saveToAppFiles(filePath, fileContents, { mode: 0o600 });
     return filePath;
@@ -200,21 +200,6 @@ export class LensClusterStoreImpl extends BaseStore<ClusterStoreModel> {
     return this.deadClusters.get(id);
   }
 
-  protected getRendererInfoByWorkspace(workspaceId: string): ClusterRenderInfo[] {
-    const aliveClusters: ClusterRenderInfo[] = this.clustersList.filter(c => c.workspace === workspaceId);
-    const deadClusters: ClusterRenderInfo[] = this.deadClustersList
-      .filter(([c]) => c.workspace === workspaceId)
-      .map(([cluster, error]) => ({
-        DeadError: error,
-        name: cluster.contextName,
-        online: false,
-        ...cluster,
-      }));
-
-
-    return _.sortBy([...aliveClusters, ...deadClusters], c => c.preferences?.iconOrder);
-  }
-
   getByWorkspaceId(workspaceId: string): Cluster[] {
     const clusters = this.clustersList.filter(c => c.workspace === workspaceId);
     return _.sortBy(clusters, c => c.preferences?.iconOrder);
@@ -262,7 +247,7 @@ export class LensClusterStoreImpl extends BaseStore<ClusterStoreModel> {
         this.setActive(null);
       }
       // remove only custom kubeconfigs (pasted as text)
-      if (cluster.kubeConfigPath == LensClusterStoreImpl.getCustomKubeConfigPath(clusterId)) {
+      if (cluster.kubeConfigPath == ClusterStore.getCustomKubeConfigPath(clusterId)) {
         unlink(cluster.kubeConfigPath).catch(() => null);
       }
     }
@@ -348,20 +333,22 @@ export class LensClusterStoreImpl extends BaseStore<ClusterStoreModel> {
   }
 }
 
-export class LensClusterStore extends LensClusterStoreImpl {
-  public getRendererInfoByWorkspace(workspaceId: string): ClusterRenderInfo[] {
-    return super.getRendererInfoByWorkspace(workspaceId);
-  }
-}
-
-export class ClusterStore extends LensClusterStoreImpl {
-  private constructor() {
-    super();
-  }
-}
-
 export const clusterStore = ClusterStore.getInstance<ClusterStore>();
-export const lensClusterStore = clusterStore as LensClusterStore;
+
+export function getRendererInfoByWorkspace(workspaceId: string): ClusterRenderInfo[] {
+  const aliveClusters: ClusterRenderInfo[] = clusterStore.clustersList.filter(c => c.workspace === workspaceId);
+  const deadClusters: ClusterRenderInfo[] = clusterStore.deadClustersList
+    .filter(([c]) => c.workspace === workspaceId)
+    .map(([cluster, error]) => ({
+      DeadError: error,
+      name: cluster.contextName,
+      online: false,
+      ...cluster,
+    }));
+
+
+  return _.sortBy([...aliveClusters, ...deadClusters], c => c.preferences?.iconOrder);
+}
 
 export function getClusterIdFromHost(hostname: string): ClusterId {
   const subDomains = hostname.split(":")[0].split(".");
--- a/src/common/cluster-store.ts
+++ b/src/common/cluster-store.ts
@@ -49,9 +49,6 @@ export interface ClusterModel {
   metadata?: ClusterMetadata;
   ownerRef?: string;
   accessibleNamespaces?: string[];
-
-  /** @deprecated */
-  kubeConfig?: string; // yaml
 }
 
 export interface ClusterPreferences extends ClusterPrometheusPreferences{
--- a/src/common/cluster-store.ts
+++ b/src/common/cluster-store.ts
@@ -8,13 +8,14 @@ import { Cluster, ClusterState } from "../main/cluster";
 import migrations from "../migrations/cluster-store";
 import logger from "../main/logger";
 import { appEventBus } from "./event-bus";
-import { dumpConfigYaml } from "./kube-helpers";
+import { dumpConfigYaml, LoadKubeError } from "./kube-helpers";
 import { saveToAppFiles } from "./utils/saveToAppFiles";
 import { KubeConfig } from "@kubernetes/client-node";
 import { subscribeToBroadcast, unsubscribeAllFromBroadcast } from "./ipc";
 import _ from "lodash";
 import move from "array-move";
 import type { WorkspaceId } from "./workspace-store";
+import { notificationsStore, NotificationStatus } from "./notifications.store";
 
 export interface ClusterIconUpload {
   clusterId: string;
@@ -73,6 +74,14 @@ export interface ClusterPrometheusPreferences {
   };
 }
 
+export interface ClusterRenderInfo extends ClusterModel {
+  DeadError?: LoadKubeError; // this != undefined => dead
+  isAdmin: boolean;
+  name: string;
+  eventCount: number;
+  online: boolean;
+}
+
 export class ClusterStore extends BaseStore<ClusterStoreModel> {
   static getCustomKubeConfigPath(clusterId: ClusterId): string {
     return path.resolve((app || remote.app).getPath("userData"), "kubeconfigs", clusterId);
@@ -88,6 +97,7 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
   @observable activeCluster: ClusterId;
   @observable removedClusters = observable.map<ClusterId, Cluster>();
   @observable clusters = observable.map<ClusterId, Cluster>();
+  @observable deadClusters = observable.map<ClusterId, [ClusterModel, LoadKubeError]>();
 
   private constructor() {
     super({
@@ -137,6 +147,10 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
     return Array.from(this.clusters.values());
   }
 
+  @computed get deadClustersList(): [ClusterModel, LoadKubeError][] {
+    return Array.from(this.deadClusters.values());
+  }
+
   @computed get enabledClustersList(): Cluster[] {
     return this.clustersList.filter((c) => c.enabled);
   }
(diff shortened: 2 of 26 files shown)


nevalla:
Please could you add screenshots to get an idea how these changes look.

Nokel81:
Changes to the tray rendering:
image

Cluster that has an error on it (menu view):
Screen Shot 2020-11-20 at 9 07 39 AM

View for when you click the cluster with a loading error:
Screen Shot 2020-11-20 at 9 08 13 AM

Notes:
- Clicking on a cluster that has a loading error does not currently change the "current cluster"
- Clicking on a cluster that has an error doesn't do a request to `main` (even though their is `main`changes to prevent errors)

extensions/telemetry/src/tracker.ts
```diff
@@ -123,30 +123,28 @@ export class Tracker extends Util.Singleton {
...
-    return os;
+
+    return "Unknown";
   }
 
   protected async event(eventCategory: string, eventAction: string, otherParams = {}) {
+    const allowed = await this.isTelemetryAllowed();
+    if (!allowed) {
+      return;
+    }
```
jim-docker: Is there any harm leaving this inside the try block? What if one day `isTelemetryAllowed()` is modified to throw an error?

src/common/kube-helpers.ts
```diff
@@ -16,13 +16,55 @@ function resolveTilde(filePath: string) {
...
+
+export class AccessError extends LoadKubeError {
+  type = "AccessError";
+
+  constructor(public pathname: string) {
+    super();
+  }
+
+  toString() {
+    return `Failed to loading the kube config. Permission denied.\n\n${this.pathname}`;
```
jim-docker: extraneous "to"? (Failed to loading...)

src/common/kube-helpers.ts
```diff
@@ -16,13 +16,55 @@ function resolveTilde(filePath: string) {
...
+
+export class ExistError extends LoadKubeError {
+  type = "ExistError";
+
+  constructor(public pathname: string) {
+    super();
+  }
+
+  toString() {
+    return `Failed to loading the kube config. No such file.\n\n${this.pathname}`;
```
jim-docker: same here

src/common/kube-helpers.ts
```diff
@@ -16,13 +16,55 @@ function resolveTilde(filePath: string) {
...
+  }
+}
+
+export function loadConfig(pathOrContent: string): KubeConfig {
   const kc = new KubeConfig();
+  try {
+    kc.loadFromFile(pathOrContent);
```
jim-docker: don't need `resolveTilde()` anymore?

src/common/kube-helpers.ts
```diff
@@ -16,13 +16,55 @@ function resolveTilde(filePath: string) {
...
 
-  if (fse.pathExistsSync(pathOrContent)) {
-    kc.loadFromFile(path.resolve(resolveTilde(pathOrContent)));
-  } else {
-    kc.loadFromString(pathOrContent);
+        kc.loadFromString(pathOrContent);
+        break;
+      }
+      case "EACCES":
+        throw new AccessError(pathOrContent);
```
jim-docker, line 65: Do we know in this case that `loadFromString()` should not be tried? (I think it was being tried previously)

jim-docker (commented):
There's a side-effect where the dead cluster notifications are shown again whenever the active cluster is changed. Do we want this? I think it would be better to show them when you try to activate one of them (or just the one time when they are loaded when Lens starts)

deadclusteroffice

Nokel81:
@jim-docker So, thanks for pointing out that UI effect. Yes I know about it. However, there isn't really a good way to tell if the notification is there previously because the renderer re-syncs the list from the `.json` file on cluster switch and because we are using `iframe` each cluster has its own instance of the store.

What I want to do is something like #1472 because that would help mitigate some of this problem (namely, only main would add the notifications).

I figured that because this is rather annoying it would nudge the user into fixing the problem.

extensions/telemetry/src/tracker.ts
Nokel81: No there isn't any harm in leaving it inside the try block. However, I think that the `console.error(..)` would be misleading if that function was the one to throw.

src/common/kube-helpers.ts
Nokel81: Thanks for catching this.

src/common/kube-helpers.ts
Nokel81: Good catch

src/common/kube-helpers.ts
Nokel81: Yes, this case means that the file that `pathOrContent` "points" to exists but is inaccessible (due to permissions). I guess in theory you could have a valid kube-config file a valid path but I think that is rare enough that we don't need to handle it. 

I guess I could try. The thing is that `loadFromString` doesn't do a lot of validation (namely if the yaml file is missing fields it is fine) so it is hard to figure out if it "failed". In other words, a string of the form `/a/path/to/a/file.yml` when passed to `loadFromString()` doesn't throw, it doesn't modify the `kc` at all.

src/common/kube-helpers.ts
```diff
@@ -16,13 +16,56 @@ function resolveTilde(filePath: string) {
...
+  }
+
+  toString() {
+    return `Failed loading the kube config. No such file.\n\n${this.pathname}`;
+  }
+}
+
+export function loadConfig(pathOrContent: string): KubeConfig {
   const kc = new KubeConfig();
+  const pathVersion = resolveTilde(pathOrContent);
```
jim-docker, line 49: probably unlikely but doing this here risks wrongly modifying the config for the case this should be loaded from string. Also, for when it is a file, the `path.resolve()` call should be done too, since it does some normalization. Though maybe this should be done in `loadFromFile()`...

Nokel81: in JS (and TS) `strings` are immutable. All "methods" on strings that return a string as a result return a new string. We cannot do it in "loadFromFile" because that is a method on `KubeConfig` which is a dependency provided type.

jim-docker: I was thinking `loadFromFile()` would normalize it for its own purposes (before it opened it). But you're right, the string should match inside and outside the `loadFromFile()` call.

src/common/kube-helpers.ts
jim-docker: `loadFromString()` does more in #1078 , wonder if it should be merged

Nokel81: I think this is still necessary before that is merged.

jim-docker (approved):

src/main/lens-proxy.ts
```diff
@@ -10,6 +10,8 @@ import { Router } from "./router";
 import { ClusterManager } from "./cluster-manager";
 import { ContextHandler } from "./context-handler";
 import logger from "./logger";
+import _ from "lodash";
+import { clusterStore } from "../common/cluster-store";
```
nevalla: Is this redundant import?

nevalla (commented):
Just thinking would it make things simpler if cluster object knows is it dead or not instead of having separate lists of clusters and dead clusters.

Nokel81:
@nevalla I disagree that it would be simpler. 

1. It is currently the constructor that throws. That means making sure that after we catch it is in a valid "enough" state to know that it is "dead" and keep track of that going forward. I think that this is completely the wrong direction as it a large maintenance burden going forward.
2. Having a separate list might have more fields in the store but it means that as a developer I can clearly see what sort of data I have for each type of cluster (since they are really two different forms of cluster information)

nevalla:
> @nevalla I disagree that it would be simpler.
> 
> 1. It is currently the constructor that throws. That means making sure that after we catch it is in a valid "enough" state to know that it is "dead" and keep track of that going forward. I think that this is completely the wrong direction as it a large maintenance burden going forward.

By quick looking, it requires that we catch the error in the constructor and set cluster as dead and store the error in cluster object too and finally return the cluster. Then elsewhere we could just rely on this dead state.

extensions/telemetry/src/tracker.ts
```diff
@@ -123,30 +123,28 @@ export class Tracker extends Util.Singleton {
...
-      } else {
-        os += "; AppImage";
-      }
-    } else {
-      os = "Unknown";
+      return "MacOS";
     }
-    return os;
+
+    if (App.isWindows) {
```
jakolehm: Why does this PR touch telemetry extension?

Nokel81:
I strongly disagree that we should do it that way. Doing a quick look at `Cluster`, every single method on it would need a check added to it to make sure it wasn't in the "dead" state. And either throw or return some other sentinel value. 

A dead cluster cannot do the same sort of things as a cluster where we have access to the kube config. I do not see any reason to conflate them.

src/common/cluster-store.ts
```diff
@@ -229,14 +267,37 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
...
 
     // update new clusters
     for (const clusterModel of clusters) {
       let cluster = currentClusters.get(clusterModel.id);
       if (cluster) {
         cluster.updateModel(clusterModel);
       } else {
-        cluster = new Cluster(clusterModel);
+        try {
+          cluster = new Cluster(clusterModel);
```
jakolehm, line 277: It's a bit surprising that a constructor throws an error. Should we actually refactor `Cluster` constructor?

Nokel81:
@nevalla If you are okay with my explanation. I will merge this.

nevalla:
> I strongly disagree that we should do it that way. Doing a quick look at `Cluster`, every single method on it would need a check added to it to make sure it wasn't in the "dead" state. And either throw or return some other sentinel value.
> 
> A dead cluster cannot do the same sort of things as a cluster where we have access to the kube config. I do not see any reason to conflate them.

Fair.

src/common/cluster-store.ts
```diff
@@ -229,14 +267,37 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
...
+            notificationsStore.add({
+              message: `Cluster ${clusterModel.contextName} is reporting an error: ${err.toString()}`,
+              id: `${clusterModel.id}_IS_DEAD_NOTIFICATION`, // this **should** prevent double errors because we are loading the stores twice
+              status: NotificationStatus.ERROR,
+              timeout: 10000,
+            });
+            continue;
+          }
+
+          throw err;
```
nevalla, line 296: Are we catching this properly?

src/renderer/api/json-api.ts
```diff
@@ -92,28 +92,27 @@ export class JsonApi<D = JsonApiData, P extends JsonApiParams = JsonApiParams> {
     });
   }
 
-  protected parseResponse<D>(res: Response, log: JsonApiLog): Promise<D> {
+  protected async parseResponse<D>(res: Response, log: JsonApiLog): Promise<D> {
```
nevalla: Is this fixing/changing something related to this PR?

src/renderer/kube-object.store.ts
```diff
@@ -175,14 +175,15 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
 
       switch (type) {
         case "ADDED":
-        case "MODIFIED":
+        case "MODIFIED": {
```
nevalla, line 178: just curious, why `{...}` is needed here?

nevalla (commented):
Added couple of more comments. @jakolehm PTAL too.

src/renderer/kube-object.store.ts
Nokel81: Because of historial reasons, variables in seperate `case` statements are in the same scope as all other variables in the same `switch` statement. 

Namely, the following will fail to compile due to name collisions

```ts
switch (foo) {
    case 0:
        const x = 1;
        break;
    case 1:
        const x = 2;
        break;
}
```

The `{...}` just creates a new block scope to prevent this.

src/common/cluster-store.ts
Nokel81: Only if we were previously (which is to say, probably not). I catch the main error that happens now above.

src/renderer/api/json-api.ts
Nokel81: No... I probably should just remove this file from this PR then. Though this code is clearer and uses the `async` style, so maybe it is fine.

src/common/cluster-store.ts
```diff
@@ -43,8 +44,10 @@ export interface ClusterModel {
   ownerRef?: string;
   accessibleNamespaces?: string[];
 
-  /** @deprecated */
-  kubeConfig?: string; // yaml
+  /**
+   * @deprecated
+   */
+  kubeConfig?: string;
```
jakolehm, line 56: Not related to this PR but should we actually remove this now that we are doing major version bump.

src/main/lens-proxy.ts
```diff
@@ -64,50 +65,55 @@ export class LensProxy {
...
-      socket.on('data', function (chunk) {
-        proxySocket.write(chunk);
-      });
-      socket.on('end', function () {
-        proxySocket.end();
-      });
-      socket.on('error', function () {
-        proxySocket.end();
-      });
+    if (!cluster) {
```
jakolehm: This causes a quite big diff .. is it necessary for this PR?

src/renderer/api/json-api.ts
jakolehm: PR should have only changes related to it's scope, let's remove this from here. (this applies also to other non-related changes).

src/common/cluster-store.ts
```diff
@@ -64,6 +67,14 @@ export interface ClusterPreferences {
   httpsProxy?: string;
 }
 
+export interface ClusterRenderInfo extends ClusterModel {
```
jakolehm, line 79: This feels a bit weird. Now we have `ClusterModel`, `ClusterState`, `Cluster`  and `ClusterRenderInfo` ... how is this different from `ClusterState`?

src/common/cluster-store.ts
Nokel81: I tried. But it messes with the migrations and I didn't want to try and figure it out. I think we should try to at least.

src/common/cluster-store.ts
Nokel81: `DeadError` and `name` mostly. I guess it is sort of weird but I don't see it as a big problem.

src/common/cluster-store.ts
Nokel81: I don't think it is entirely weird that a constructor throws an error. Especially if it is creating dependencies itself. The other option would be for the constructor to take those values in as parameters. We would then have to move all this checking code to all the places that this constructor is called currently.

src/common/cluster-store.ts
jakolehm: I was also referring to `getByWorkspaceId` which now returns `ClusterRenderInfo[]` instead of `Cluster[]`. Would it make sense to include these fields in `ClusterState` and return `Cluster[]` as it was previously?

src/common/cluster-store.ts
jakolehm: Right, probably out of scope for this PR.

src/common/cluster-store.ts
Nokel81: I don't see why it would make sense for `DeadError` to be part of `ClusterState` let alone `Cluster` (see my responses to Lauri above). I guess `name` could be, but it doesn't really feel like a part of its state.

I can understand the desire for `getByWorkspaceId` to return `Cluster[]`. I will see what I can do with method overloading. I figured it would be fine if it didn't because the only two uses of the function were for rendering stuff, which is where I got the requirements for `ClusterRenderInfo` from.

Nokel81: I didn't manage to get the overloading to work (maybe it is still possible and I am not skilled at typescript enough). So I just made a new function and kept `getByWorkspaceId` returning `Cluster[]`.

Nokel81:
@jakolehm or @nevalla If there are no more concerns, shall I merge this?

src/common/cluster-store.ts
jakolehm: I'm not really comfortable with this.. `ClusterStore` is also exposed to extension-api and `ClusterRenderInfo` is also leaked there... could we move `getRendererInfoByWorkspace` (and `ClusterRenderInfo`) out of this store, it's highly related to a single UI component?

Nokel81: And the tray. `getWorkspaceById` (except for swapping icon order above) is only used in those two locations too.

I'll see if I can do that, maybe "protected" and a child class?

src/common/cluster-store.ts
```diff
@@ -182,10 +196,20 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
...
+
   getByWorkspaceId(workspaceId: string): Cluster[] {
-    const clusters = Array.from(this.clusters.values())
-      .filter(cluster => cluster.workspace === workspaceId);
-    return _.sortBy(clusters, cluster => cluster.preferences.iconOrder);
+    const clusters = this.clustersList.filter(c => c.workspace === workspaceId);
+    return _.sortBy(clusters, c => c.preferences?.iconOrder);
+  }
+
+  getCountInWorkspace(workspaceId: string): number {
```
jakolehm, line 208: It's not obvious that this counts also dead and disabled clusters? Maybe should not be part of ClusterStore API?

If we really want to keep it here then name should be `getVisibleCountInWorkspace` and it should not count disabled clusters.

src/common/cluster-store.ts
```diff
@@ -88,8 +97,9 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
   @observable activeCluster: ClusterId;
   @observable removedClusters = observable.map<ClusterId, Cluster>();
   @observable clusters = observable.map<ClusterId, Cluster>();
+  @observable deadClusters = observable.map<ClusterId, [ClusterModel, LoadKubeError]>();
```
jakolehm, line 100: Introduces new extension API.

src/common/cluster-store.ts
```diff
@@ -137,6 +147,10 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
     return Array.from(this.clusters.values());
   }
 
+  @computed get deadClustersList(): [ClusterModel, LoadKubeError][] {
```
jakolehm, line 150: Introduces new extension API.

src/common/cluster-store.ts
```diff
@@ -182,10 +196,20 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
     return this.clusters.get(id);
   }
 
+  getDeadById(id: ClusterId): [ClusterModel, LoadKubeError] {
```
jakolehm, line 199: Introduces new extension API.

src/common/cluster-store.ts
```diff
@@ -88,8 +97,9 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
   @observable activeCluster: ClusterId;
   @observable removedClusters = observable.map<ClusterId, Cluster>();
   @observable clusters = observable.map<ClusterId, Cluster>();
+  @observable deadClusters = observable.map<ClusterId, [ClusterModel, LoadKubeError]>();
 
-  private constructor() {
+  protected constructor() {
```
jakolehm: Private?

Nokel81:
Closing this as #1956 also fixes this and this PR is both very old and also tries to do the notification (in a bad way too). I still think that these clusters should be in a separate list but once I have time to do that I'll make a new PR.

PR #1455, opened 2020-11-19
make LensExtension abstract as an affordence to extension devs

Nokel81:
I believe that LensExtension should be `abstract` as the two methods (`onActivate` and `onDeactivate`) both really should be implemented (even for simple extensions).

This would also be a very good affordance to extension developers to remind them that these two methods exist and initializing items that are used only for the duration of the extension's lifetime should be created and deleted within these (instead of in the constructor).

### DIFF
--- a/src/extensions/lens-extension.ts
+++ b/src/extensions/lens-extension.ts
@@ -3,7 +3,7 @@ import { action, observable, reaction } from "mobx";
 import logger from "../main/logger";
 
 export type LensExtensionId = string; // path to manifest (package.json)
-export type LensExtensionConstructor = new (...args: ConstructorParameters<typeof LensExtension>) => LensExtension;
+export type LensExtensionConstructor = new (ext: InstalledExtension) => LensExtension;
 
 export interface LensExtensionManifest {
   name: string;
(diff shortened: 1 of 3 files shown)


panuhorsmalahti (approved):

jakolehm:
> I believe that LensExtension should be abstract as the two methods (onActivate and onDeactivate) both really should be implemented (even for simple extensions).

I'm not convinced, this changes whole lifecycle of an extension. Currently all registrations happen outside of `onActivate`, if we force developers to define `onActivate` then do we also force all registration happen there? That's like huuuge change how extensions work.

jakolehm (changes requested):
Blocking this so that we don't accidentally merge.

Nokel81:
Okay, I understand that I wasn't that clear with my reasoning behind this change. The main point of this was merely to try and push developers to implementing those functions and to not have their own constructor at all.

Nokel81:
Also @jakolehm Abstract classes in javascript can have fields and can have those fields have initializes so this wouldn't change that. However, it does change some other things (related to constructors). So I will close this as it is more work than what we probably want to do.

PR #1456, opened 2020-11-19
Allow to install packed extensions from URL or local file

### LINKED ISSUE #1227: Option to install an extension from filesystem/url
Extension page should have a mechanism for installing an extension from local filesystem package or from an URL (both tar.gz npm packages)

ixrock:
__Done_:_
- [x] UI part of extensions-page (unpack, validate, load .tgz)
- [x] new react-component `Clipboard` for copying html-element full text-content or partial into clipboard
- [x] new react-component `DropFileInput` to handle drag-n-dropped files from external app, e.g. Finder / other file explorer (logic adopted originally from `add-cluster-page`) 

_Todo:_
- [x] fix: downloading file by URL
- [x] unpack requested extensions to `~/.k8slens/extensions`

requires #1461 #1482 (otherwise just restart the app after installation)
close #1227 

_Screenshots:_


src/renderer/components/+add-cluster/add-cluster.tsx
```diff
@@ -119,6 +118,11 @@ export class AddCluster extends React.Component {
     }
   };
 
+  onDropKubeConfig = (files: File[]) => {
```
panuhorsmalahti, line 123: We could show error if the user tried to drag & drop multiple files. I suppose this could be done in DropFileInput too.

ixrock: D&D for multiple files is intentionally supported, why not?

panuhorsmalahti: In this function it seems only one dropped file is actually used?

ixrock: Yes, right, for adding clusters expected only single kube-config file in D&D, so first one is used by default I would say. There is no error to handle imho since UI expects single file path to local kube-config.

package.json
```diff
@@ -47,7 +47,7 @@
     "bundledHelmVersion": "3.3.4"
   },
   "engines": {
-    "node": ">=12.0 <13.0"
+    "node": ">=12 <=14"
```
ixrock: Current latest stable version`14.15.1`

src/renderer/components/notifications/notifications.store.ts
```diff
@@ -23,44 +22,50 @@ export interface Notification {
...
+      this.notifications.splice(index, 1, notification); // update existing with same id
+    } else {
+      this.notifications.push(notification); // add new
     }
-    const index = this.notifications.findIndex(item => item.id === notification.id);
-    if (index > -1) this.notifications.splice(index, 1, notification);
-    else this.notifications.push(notification);
-    this.addAutoHideTimer(notification);
+    this.addAutoHideTimer(id);
+    return () => this.remove(id);
```
ixrock, line 62: Line caused refactoring of notifications 😅

ixrock:
@jakolehm PTAL / test

src/extensions/extension-manager.ts
```diff
@@ -97,6 +97,15 @@ export class ExtensionManager {
     }
   }
 
+  getNpmPackageTarballUrl(packageName: string) {
+    try {
+      const command = [this.npmPath, "view", packageName, "dist.tarball", "--silent"];
+      return child_process.execSync(command.join(" "), { encoding: "utf8" }).trim();
```
ixrock: Currently doesn't work in production build.. Any ideas?

panuhorsmalahti: Not sure why, but I don't think we want to block the application while running this command? Why not use .exec?

src/common/utils/downloadFile.ts
```diff
@@ -0,0 +1,36 @@
...
+    });
+    req.on("complete", () => {
+      resolve(Buffer.concat(fileChunks));
+    });
+    req.on("error", err => {
+      reject({ url, err });
+    });
+  });
+  return {
+    url: url,
```
panuhorsmalahti: [nit] `url,` is enough - I think we should enable object-shorthand in eslint to make this automatic

src/common/utils/downloadFile.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export function downloadFile(opts: DownloadFileOptions): DownloadFileTicket {
+  const { url, gzip = true } = opts;
+  const fileChunks: Buffer[] = [];
+  const req = request(url, { gzip });
+  const promise: Promise<Buffer> = new Promise((resolve, reject) => {
+    req.on("data", (chunk: Buffer) => {
+      fileChunks.push(chunk);
+    });
+    req.on("complete", () => {
```
panuhorsmalahti: complete and error events should be reacted with ".once"

src/common/utils/tar.ts
```diff
@@ -0,0 +1,49 @@
...
+    } = opts;
+
+    await tar.list({
+      file: tarFilePath,
+      filter: fileMatcher,
+      onentry(entry: FileStat) {
+        entry.on("data", chunk => {
+          fileChunks.push(chunk);
+        });
+        entry.on("error", err => {
```
panuhorsmalahti: Can use .once here too (if available in FileStat).

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +54,238 @@ export class Extensions extends React.Component {
...
+    logger.info('Install from D&D');
+    return this.requestInstall(
+      files.map(file => ({
+        fileName: path.basename(file.path),
+        filePath: file.path,
+      }))
+    );
+  };
+
+  async requestInstall(installRequests: InstallRequest[]) {
```
nevalla: Maybe split this method into separate methods if possible, since now it's quite a long.

package.json
jakolehm: This is not valid, node version needs to match used Electron nodejs version.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +54,238 @@ export class Extensions extends React.Component {
...
+      this.downloadUrl = "";
+    } else {
+      this.installFromSelectFileDialog();
+    }
+  };
+
+  installFromNpmOrUrl = async (url = this.downloadUrl) => {
+    if (!InputValidators.isUrl.validate(url)) {
+      const npmPackageName = url;
+      url = extensionManager.getNpmPackageTarballUrl(npmPackageName);
```
panuhorsmalahti: I'd avoid reassigning to function parameters. We should also configure this with eslint (no-param-reassign).

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +54,238 @@ export class Extensions extends React.Component {
...
+        return;
+      }
+    }
+    try {
+      const { promise: filePromise } = downloadFile({ url });
+      this.requestInstall([{
+        fileName: path.basename(url),
+        data: await filePromise,
+      }]);
+    } catch (err) {
```
panuhorsmalahti, line 107: [nit] unnecessary abbreviation, e.g. "error" can be used instead

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +54,238 @@ export class Extensions extends React.Component {
...
+        filePath: file.path,
+      }))
+    );
+  };
+
+  async requestInstall(installRequests: InstallRequest[]) {
+    const pendingFiles: Promise<any>[] = [];
+
+    // read extensions with provided system path if any
+    installRequests.forEach(ext => {
```
panuhorsmalahti: This could be cleaner by first using `.filter` to skip ext which have `ext.data`, then using `.map`. You can then await the value.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +54,238 @@ export class Extensions extends React.Component {
...
+      if (ext.data) return;
+      const promise = fse.readFile(ext.filePath)
+        .then(data => ext.data = data)
+        .catch(err => {
+          Notifications.error(`Error while reading "${ext.filePath}": ${String(err)}`);
+        });
+      pendingFiles.push(promise);
+    });
+    await Promise.all(pendingFiles);
+    installRequests = installRequests.filter(item => item.data); // remove items with reading errors
```
panuhorsmalahti: Avoid mutation of function parameters, create a new variable instead

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +54,238 @@ export class Extensions extends React.Component {
...
+
+    // prepare temp folder
+    const tempFolder = path.join(os.tmpdir(), "lens-extensions");
+    await fse.ensureDir(tempFolder);
+
+    // copy files to temp, get extension info from package.json and do basic validation
```
panuhorsmalahti: This function is quite long and does many things, I'd split it into 2 or 3 functions.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +54,238 @@ export class Extensions extends React.Component {
...
+            </Clipboard>
+          </p>
+          <p className="hint">
+            <Trans><b>Pro-Tip 2</b>: you can drag & drop extension's tarball here to request installation</Trans>
+          </p>
+        </div>
+        <div className="more-info flex inline gaps align-center">
+          <Icon material="local_fire_department"/>
+          <p>
+            Check out documentation to <a href="https://docs.k8slens.dev/" target="_blank">learn more</a>
```
panuhorsmalahti: This url is referenced in a few places in our codebase. We should add it as a constant.

src/renderer/components/+extensions/extensions.tsx
jakolehm: I agree with @nevalla , this should be split (maybe separate validation from install).

jakolehm (changes requested):

src/renderer/components/+extensions/extensions.tsx
jakolehm: We should not fallback to npm, URL is the only supported method for remote packages.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +54,238 @@ export class Extensions extends React.Component {
...
+      const { fileName, manifest } = install;
+      const { name, version, description } = manifest;
+      const extensionFolder = this.getExtensionDestFolder(name);
+      const folderExists = fse.existsSync(extensionFolder);
+      const removeNotification = Notifications.info(
+        <div className="InstallingExtensionNotification flex gaps align-center">
+          <div className="flex column gaps">
+            <p>Install extension <b title={fileName}>{name}@{version}</b>?</p>
+            <p>Description: <em>{description}</em></p>
+            {folderExists && (
```
jakolehm: We should probably show this notification only if `folderExists`. Extension should be also automatically enabled after install.

ixrock: `foldersExists &&` does what you say, extension is enabled by default (if was not disabled previously)

src/renderer/components/+extensions/extensions.tsx
ixrock: Alright, let's keep it simple ;)

src/renderer/components/+extensions/extensions.tsx
ixrock: Reason?

src/renderer/components/+extensions/extensions.tsx
ixrock: Ok, I'll check

src/renderer/components/+extensions/extensions.tsx
jakolehm: > `foldersExists &&` does what you say, extension is enabled by default (if was not disabled previously)

This notification still always prompts user to hit "Install" button, that's unnecessary unless folder exists.

src/renderer/components/+extensions/extensions.tsx
ixrock: Sure

src/renderer/components/+extensions/extensions.tsx
ixrock: Why? Do you have clear explanation why it's bad? :)

src/renderer/components/+extensions/extensions.tsx
ixrock: How otherwise you can re-install/update? 
Oh, or do you mean install automatically if "all good and not installed yet"? And show notification when exists?

jakolehm: > Oh, or do you mean install automatically if "all good and not installed yet"? And show notification when exists?

Yes 🙂

src/renderer/components/+extensions/extensions.tsx
panuhorsmalahti: Just a common coding practice to make the code easier to read. The eslint rule has some arguments for it: https://eslint.org/docs/rules/no-param-reassign

Often you can also give the new variable a slightly different meaning, e.g. `successfulRequests` here.

src/renderer/components/+extensions/extensions.tsx
panuhorsmalahti: Answered in the other comment.

package.json
ixrock: Node version in Electron: `12.14.1`, how it's related if I haven't changed minimum version?

jakolehm: Only 12.x versions work, why would we allow any other version to be used? That also applies to changes in build pipeline, they must use valid node version.

ixrock: So, maybe we should upgrade `electron` (latest 11.0.3.)? I believe `14.x` has some fixed security issues:
https://nodejs.org/en/blog/vulnerability/november-2020-security-releases/

jakolehm: It's not trivial upgrade (tests fail even if we upgrade to 9.3.x, see: https://github.com/lensapp/lens/pull/1398). Anyway this is totally out of scope of this PR.

src/extensions/extension-manager.ts
ixrock: Doesn't matter anymore, I'll remove auto-detection from NPM as we support only direct URLs.

src/common/utils/downloadFile.ts
ixrock: What's the difference in that case?

src/renderer/components/+extensions/extensions.tsx
ixrock: Yes

src/common/utils/downloadFile.ts
panuhorsmalahti: Promise should be resolved or rejected once, not multiple times. The events might only be emitted once in reality, but it's better/safer to only react to the first event since we want the callback to only react the first time.

package.json
ixrock: Okay

src/common/utils/downloadFile.ts
ixrock: Ok, but can't get why it's safer when:
```
const test = new Promise(resolve => {
  resolve(1);
  setTimeout(() => {
    resolve(2);
    console.log("resolving again won't trigger any errors");
  }, 1000)
});
```

src/renderer/components/+extensions/extensions.tsx
ixrock: What? :D I agree if it would be just `e`.
Maybe `evt` also unnecessary abbreviation for event-handlers all around and we should use `event` instead?

src/common/utils/downloadFile.ts
panuhorsmalahti: Resolving promise twice is an "error" and can lead to difficult to understand/debug bugs, even if there's no error logged for it.

src/renderer/components/+extensions/extensions.tsx
panuhorsmalahti: I'd use `event` also. This is suggested in e.g. book "Code Complete", which has some empirical studies for it. Not saying this is super-important, thus marked as "nit". (For me "nit" means that it doesn't need to be fixed necessarily, but would be nice).

src/common/utils/downloadFile.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export interface DownloadFileOptions {
+  url: string;
+  gzip?: boolean;
+}
+
+export interface DownloadFileTicket {
+  url: string;
+  promise: Promise<Buffer>;
+  cancel(): void;
```
Nokel81, line 11: Wouldn't it be better for this to be a `CancalablePromise<Buffer>`?

src/common/utils/index.ts
```diff
@@ -15,3 +15,5 @@ export * from "./saveToAppFiles";
 export * from "./singleton";
 export * from "./openExternal";
 export * from "./rectify-array";
+export * from "./downloadFile";
+export * from "./escapeRegExp";
```
Nokel81, line 19: shouldn't this include `tar.ts`?

src/common/utils/tar.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    await tar.list({
+      file: tarPath,
+      filter: path => path === filePath,
+      onentry(entry: FileStat) {
+        entry.on("data", chunk => {
+          fileChunks.push(chunk);
+        });
+        entry.once("error", err => {
+          reject(`Reading ${entry.path} error: ${err}`);
```
Nokel81: Promises shouldn't be rejected with strings. Please use `new Error`

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -1,21 +1,49 @@
...
+}
+
+interface InstallRequestValidated extends InstallRequestPreloaded {
+  manifest: LensExtensionManifest;
+  tmpFile: string; // temp file for unpacking
+}
 
 @observer
 export class Extensions extends React.Component {
+  private supportedFormats = [".tar", ".tgz"];
```
Nokel81, line 43: should this be static?

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +60,264 @@ export class Extensions extends React.Component {
     return extensionDiscovery.localFolderPath;
   }
 
+  getExtensionPackageTemp(fileName = "") {
+    return path.join(os.tmpdir(), "lens-extensions", fileName);
+  }
+
+  getExtensionDestFolder(name: string) {
+    return path.join(this.extensionsPath, sanitizeExtensionName(name));
```
Nokel81, line 67: This sanitation is not as complete as you think. So far I don't think we put any requirements on extension names and (at least on Windows) we do not sanitize out `\` or even `.`s. 

Plus I don't see why `@` cannot be part of the dest folder path.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +60,264 @@ export class Extensions extends React.Component {
...
+    }
+  };
+
+  addExtensions = () => {
+    const { downloadUrl } = this;
+    if (downloadUrl && InputValidators.isUrl.validate(downloadUrl)) {
+      this.installFromUrl(downloadUrl);
+    } else {
+      this.installFromSelectFileDialog();
+    }
```
Nokel81, line 97: What if the downloadURL is set and not valid. That shouldn't open up the file select dialog.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +60,264 @@ export class Extensions extends React.Component {
...
+  installFromUrl = async (url: string) => {
+    try {
+      const { promise: filePromise } = downloadFile({ url });
+      this.requestInstall([{
+        fileName: path.basename(url),
+        data: await filePromise,
+      }]);
+    } catch (err) {
+      Notifications.error(
+        <p>Installation via URL has failed: <b>{String(err)}</b></p>
```
Nokel81, line 109: It seems weird that the error is `bold`. I cannot think of other notifications that do that.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +60,264 @@ export class Extensions extends React.Component {
...
+            req.data = data;
+            preloadedRequests.push(req);
+          }).catch(err => {
+            if (showError) {
+              Notifications.error(`Error while reading "${req.filePath}": ${String(err)}`);
+            }
+          });
+        })
+    );
+    return preloadedRequests as InstallRequestPreloaded[];
```
Nokel81, line 140: No need to filter, items get promise wrapped when awaited.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +60,264 @@ export class Extensions extends React.Component {
...
+  }
+
+  async validatePackage(filePath: string): Promise<LensExtensionManifest> {
+    const tarFiles = await listTarEntries(filePath);
+
+    // tarball from npm contains single root folder "package/*"
```
Nokel81, line 146: by `package/*` do you mean `<package>/*`? (namely, is the root folder called "package" or is it the package's name.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +60,264 @@ export class Extensions extends React.Component {
...
+    const hasSingleRootFolder = tarFiles.every(entry => entry.startsWith(tarFiles[0]));
+    const manifestLocation = hasSingleRootFolder ? path.join(tarFiles[0], manifestFilename) : manifestFilename;
+
+    if(!tarFiles.includes(manifestLocation)) {
+      throw `invalid package, ${manifestFilename} not found`;
```
Nokel81: Please do not throw strings.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +60,264 @@ export class Extensions extends React.Component {
...
+    }
+    const manifest = await readFileFromTar<LensExtensionManifest>({
+      tarPath: filePath,
+      filePath: manifestLocation,
+      parseJson: true,
+    });
+    if (!manifest.lens && !manifest.renderer) {
+      throw `${manifestFilename} must specify "main" and/or "renderer" fields`;
```
Nokel81: same here

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +60,264 @@ export class Extensions extends React.Component {
...
+    const manifestLocation = hasSingleRootFolder ? path.join(tarFiles[0], manifestFilename) : manifestFilename;
+
+    if(!tarFiles.includes(manifestLocation)) {
+      throw `invalid package, ${manifestFilename} not found`;
+    }
+    const manifest = await readFileFromTar<LensExtensionManifest>({
+      tarPath: filePath,
+      filePath: manifestLocation,
+      parseJson: true,
+    });
```
Nokel81, line 158: I think that it would be better if `readFileFromTar` took as an optional parameter `validateJson`. That way we can be assured that the types coming in are valid.

Something like the following:

```typescript
interface ParseJson<T> {
    parseJson: true;
    validateJson?: (from: any) => T;
}

interface NoParseJson {
    parseJson?: false;
}

type ReadFileFromTarOpts<T> = (NoParseJson | ParseJson<T>) & {
  tarPath: string;
  filePath: string;
}

function readFileFromTar<R>({tarPath, filePath, ...parsing}: ReadFileFromTarOpts<R>) {
  if (parsing.parseJson === true) {
    const options: ParseJson<R> = parsing;
    options.validateJson?.(...)
  }
}
```

src/common/utils/tar.ts
```diff
@@ -0,0 +1,55 @@
...
+        });
+        entry.once("end", () => {
+          const data = Buffer.concat(fileChunks);
+          const result = parseJson ? JSON.parse(data.toString("utf8")) : data;
+          resolve(result);
```
Nokel81, line 29: Also this is wrong. If `R` is unchanged but `parseJson: true` then the types are wrong. I am surprised that typescript doesn't catch this.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +60,264 @@ export class Extensions extends React.Component {
...
+  }
+
+  async createTempFilesAndValidate(requests: InstallRequestPreloaded[], { showErrors = true } = {}) {
+    const validatedRequests: InstallRequestValidated[] = [];
+
+    // copy files to temp
+    await fse.ensureDir(this.getExtensionPackageTemp());
+    requests.forEach(req => {
+      const tempFile = this.getExtensionPackageTemp(req.fileName);
+      fse.writeFileSync(tempFile, req.data);
```
Nokel81, line 172: Why are we writing files synchronously in an `async` function?

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +60,264 @@ export class Extensions extends React.Component {
...
+    // validate packages
+    await Promise.all(
+      requests.map(async req => {
+        const tempFile = this.getExtensionPackageTemp(req.fileName);
+        try {
+          const manifest = await this.validatePackage(tempFile);
+          validatedRequests.push({
+            ...req,
+            manifest,
+            tmpFile: tempFile,
```
Nokel81: I think this shows that either `InstallRequestValidated` should use `tempFile` or `tempFile` should be renamed to `tmpFile`. I think the former is better.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +60,264 @@ export class Extensions extends React.Component {
...
+              </div>
+            </div>
+            <Button autoFocus label="Install" onClick={() => {
+              removeNotification();
+              this.unpackExtension(install);
+            }}/>
+          </div>
+        );
+      }
+    });
```
Nokel81, line 231: I don't see why these cannot be don't asynchronously.

src/renderer/components/clipboard/clipboard.tsx
```diff
@@ -0,0 +1,62 @@
...
+  render() {
+    try {
+      const rootElem = this.rootReactElem;
+      return React.cloneElement(rootElem, {
+        className: cssNames(Clipboard.displayName, rootElem.props.className),
+        onClick: this.onClick,
+      });
+    } catch (err) {
+      logger.error(`Invalid usage components/CopyToClick usage. Children must contain root html element.`, { err: String(err) });
+      return this.rootReactElem;
```
Nokel81, line 59: I think you need to justify doing this instead of doing a normal react component. Is there a good reason why this couldn't just be something like the following:

```tsx
render() {
    return (
        <div className="Clipboard" onClick={this.onClick}>
            {this.props.children}
        </div>
    );
}
```

src/renderer/components/input/drop-file-input.tsx
```diff
@@ -0,0 +1,76 @@
...
+  @autobind()
+  onDragLeave() {
+    this.dropAreaActive = false;
+  }
+
+  @autobind()
+  onDragOver(evt: React.DragEvent<T>) {
+    if (this.props.onDragOver) {
+      this.props.onDragOver(evt);
+    }
```
Nokel81, line 36: [nit] (namely, something that I would do but not necessary)

```suggestion
    this.props.onDragOver?.(evt);
```

src/renderer/components/notifications/notifications.store.ts
```diff
@@ -23,44 +22,50 @@ export interface Notification {
...
+  getById(id: NotificationId): Notification | null {
+    return this.notifications.find(item => item.id === id) ?? null;
+  }
+
+  addAutoHideTimer(id: NotificationId) {
+    const notification = this.getById(id);
+    if (!notification) return;
+    this.removeAutoHideTimer(id);
+    if (notification.timeout) {
+      const timer = window.setTimeout(() => this.remove(id), notification.timeout);
```
Nokel81:
```suggestion
    this.removeAutoHideTimer(id);
    const notification = this.getById(id);
    if (notification?.timeout) {
      const timer = window.setTimeout(() => this.remove(id), notification.timeout);
```

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +60,264 @@ export class Extensions extends React.Component {
...
-          Check out documentation to <a href="https://docs.k8slens.dev/" target="_blank">learn more</a>
+        <div className="install-extension flex column gaps">
+          <SubTitle title="Install extensions:"/>
+          <Input
+            showErrorsAsTooltip={true}
+            className="box grow"
+            theme="round-black"
+            iconLeft="link"
+            placeholder={`URL to packed extension (${this.supportedFormats.join(", ")})`}
+            validators={InputValidators.isUrl}
```
Nokel81, line 292: This validator is wrong, the string "123.com" is reported as invalid. Also it seems that even if the validator returns an error we can still click the "Add extensions" button

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -32,27 +60,264 @@ export class Extensions extends React.Component {
...
+            value={this.downloadUrl}
+            onChange={v => this.downloadUrl = v}
+            onSubmit={this.addExtensions}
+          />
+          <Button
+            primary
+            label="Add extensions"
```
Nokel81, line 299: This should be i18n-ed

Nokel81 (dismissed):

src/common/utils/downloadFile.ts
ixrock: Nope, it wouldn't

Nokel81:
Also, I tested this with `https://registry.npmjs.org/@mirantis/lens-extension-cc/-/lens-extension-cc-1.0.2.tgz` which is the value I got from running `npm view @mirantis/lens-extension-cc dist.tarball`. And this PR reported a failure because "invalid package, package.json not found".

However, if I download that package and unpack it. I get the following folder structure:
Screen Shot 2020-11-24 at 2 10 45 PM

I think that this PR should handle correctly how npm bundles its code into tarballs.

src/common/utils/index.ts
ixrock: Yes, don't see any obstacles :)

src/common/utils/tar.ts
ixrock: Why not?

src/renderer/components/+extensions/extensions.tsx
ixrock: Not critical and `static` is useful mostly when required from outside of component.

src/renderer/components/+extensions/extensions.tsx
ixrock: Ok, so you suggesting to use it without sanitizing here?

ixrock: Anyway we have to sanitize it since name might contain `/` and currently loading extensions supported only from single level from `~./k8slens/extensions/<folder>`

src/common/utils/tar.ts
Nokel81: For a number of reasons, you loose the locality of the error (where it came from). It is no longer possible to get the source error (because it has been stringified). We should be able to inspect errors. What I really would like to see is something like the following:

```ts
export class ReadTarError {
    constructor(public path: string, public source: Error) {}

    toString() {
        return `Reading ${this.path} failed. ${this.source}`;
    }
}
```

src/renderer/components/+extensions/extensions.tsx
Nokel81: True, but I am thinking more about security. If someone crafted the name to be `@foo\..\..\/blarr` on windows, this would be "sanitized" to `foo\..\..\-blarr` and if we were ever to say "overWrite: true" (which I think I saw in this PR somewhere) then we would be writing to some unknown place.

src/renderer/components/+extensions/extensions.tsx
ixrock: Doesn't really matter in this case (and it's more readable imho)

Nokel81: How so? You could easily make a function for each of these, and then do an `await Promise.all(validatedRequests.map(this.confirmInstall))`

src/renderer/components/clipboard/clipboard.tsx
ixrock: Yes, of course: reason that this component should not put any extra-layout since it might break styles, etc.
Otherwise event-handler probably _might be_ implemented in some place above in DOM-tree and this `children` needs to be matched with`event.target`..

Nokel81: event-handler?

src/renderer/components/+extensions/extensions.tsx
ixrock: Dunno, this is out of scope of this PR. This validator also matching to empty string which is also invalid.

src/renderer/components/+extensions/extensions.tsx
ixrock: Until we officially support i18n it is NOT (and I recommend to everyone don't spend time on it)

src/renderer/components/+extensions/extensions.tsx
Nokel81: The validator being wrong is yes, but not being able to click "add extensions" if it returns an error

src/common/utils/downloadFile.ts
Nokel81: okay.

ixrock:
> I think that this PR should handle correctly how npm bundles its code into tarballs.

Yes it should and I expected it to work, so you just found one more 🐞 Congrats!  🎉 (FIXED)

src/renderer/components/clipboard/clipboard.tsx
ixrock: Yes, this component modifies `onClick`

Nokel81: I can sort of see your need, but I think we should just design around that. See the following: https://github.com/yannickcr/eslint-plugin-react/issues/678

React wants to eventually remove that anyway.

src/renderer/components/+extensions/extensions.tsx
ixrock: It's not valid name for NPM-package (`@foo\..\..\/blarr`) so no need to worry about it. 
@jakolehm your word?

src/renderer/components/clipboard/clipboard.tsx
ixrock: Yes, we will figure this out once it became at least deprecated officially from `react`-team.. 
Or when we have more free time :)

ixrock: > event-handler?

I was talking about event-delegation pattern..

src/renderer/components/+extensions/extensions.tsx
ixrock: So yeah, `123.com` is invalid URL, try `new URL("123.com")` :D
Just prepend with http(s) and all good, lol.

ixrock: BTW you might want to fix this meanwhile in separated PR! :)

ixrock:
@jakolehm ready to merge? any issues so far? ofc after #1510

jakolehm (approved):
LGTM. I spotted few usability issues but they should be fixed in separate PRs (I'll open issues about those).

src/renderer/components/+extensions/extensions.tsx
jakolehm: @Nokel81 has a point here.. I think it's possible for an attacker to handcraft a package with invalid name. Of course if user installs a malicious package then it can do even worse things via runtime ... so maybe not super critical to fix here.

Let's create a separate issue about this.

panuhorsmalahti (approved):

PR #1469, opened 2020-11-20
Replace support-page extension with a link to the documentation support page

jakolehm:


jim-docker (commented):
Do we want it on the status bar too? Personally, I don't think it needs to be there but it was there in the extension.

jakolehm:
> Do we want it on the status bar too? Personally, I don't think it needs to be there but it was there in the extension.

Nope, not needed.

jakolehm:
Related PR #1473

miskun (approved):
LGTM

nevalla (approved):
LGTM

PR #1474, opened 2020-11-20
change BaseRegistry to only have one type parameter

Nokel81:

I noticed recently that there was an `as I[]` call in the `BaseRegistry`'s `add` method, which seemed odd to me. After some more investigation, it seemed to have been added as a way for the registry to capture its child's type requirements for the shadowing of the `add` method.

Therefore I propose this simplification. Since child types can just use whatever signature they want. I personally don't see a downside to having this as the system.

### DIFF
--- a/src/extensions/registries/kube-object-detail-registry.ts
+++ b/src/extensions/registries/kube-object-detail-registry.ts
@@ -14,15 +14,10 @@ export interface KubeObjectDetailRegistration {
 
 export class KubeObjectDetailRegistry extends BaseRegistry<KubeObjectDetailRegistration> {
   getItemsForKind(kind: string, apiVersion: string) {
-    const items = this.getItems().filter((item) => {
-      return item.kind === kind && item.apiVersions.includes(apiVersion);
-    }).map((item) => {
-      if (item.priority === null) {
-        item.priority = 50;
-      }
-      return item;
-    });
-    return items.sort((a, b) => b.priority - a.priority);
+    return this.getItems()
+      .filter(item => item.kind === kind && item.apiVersions.includes(apiVersion))
+      .map(item => (item.priority ??= 50, item))
+      .sort((a, b) => b.priority - a.priority);
   }
 }
 
--- a/src/extensions/registries/kube-object-detail-registry.ts
+++ b/src/extensions/registries/kube-object-detail-registry.ts
@@ -14,10 +14,15 @@ export interface KubeObjectDetailRegistration {
 
 export class KubeObjectDetailRegistry extends BaseRegistry<KubeObjectDetailRegistration> {
   getItemsForKind(kind: string, apiVersion: string) {
-    return this.getItems()
-      .filter(item => item.kind === kind && item.apiVersions.includes(apiVersion))
-      .map(item => (item.priority ??= 50, item))
-      .sort((a, b) => b.priority - a.priority);
+    const items = this.getItems().filter((item) => {
+      return item.kind === kind && item.apiVersions.includes(apiVersion);
+    }).map((item) => {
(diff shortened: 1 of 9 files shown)


src/extensions/registries/page-registry.ts
```diff
@@ -59,24 +60,27 @@ export function getExtensionPageUrl<P extends object>({ extensionId, pageId = ""
...
     } catch (err) {
       logger.error(`[EXTENSION]: page-registration failed`, {
         items,
         extension: ext,
         error: String(err),
       });
+
+      return noop;
     }
-    return super.add(registeredPages);
```
jim-docker: can't you keep calling this and let BaseRegistry handle the noop situation? Seems cleaner the way it was. Or am I missing the point?

Nokel81: This was mostly done so that I could remove the above `let` and make it a `const`

src/extensions/registries/kube-object-detail-registry.ts
```diff
@@ -14,15 +14,10 @@ export interface KubeObjectDetailRegistration {
...
-    const items = this.getItems().filter((item) => {
-      return item.kind === kind && item.apiVersions.includes(apiVersion);
-    }).map((item) => {
-      if (item.priority === null) {
-        item.priority = 50;
-      }
-      return item;
-    });
-    return items.sort((a, b) => b.priority - a.priority);
+    return this.getItems()
```
jakolehm: Not related to PR title?

src/extensions/registries/kube-object-menu-registry.ts
```diff
@@ -13,9 +13,8 @@ export interface KubeObjectMenuRegistration {
 
 export class KubeObjectMenuRegistry extends BaseRegistry<KubeObjectMenuRegistration> {
   getItemsForKind(kind: string, apiVersion: string) {
-    return this.getItems().filter((item) => {
-      return item.kind === kind && item.apiVersions.includes(apiVersion);
-    });
+    return this.getItems()
```
jakolehm: Not related to PR title?

src/extensions/registries/page-menu-registry.ts
```diff
@@ -30,28 +30,16 @@ export interface PageMenuComponents {
...
   add(items: PageMenuRegistration[], ext: LensExtension) {
-    const normalizedItems = items.map(menuItem => {
-      menuItem.target = {
-        extensionId: ext.name,
-        ...(menuItem.target || {}),
-      };
-      return menuItem;
-    });
-    return super.add(normalizedItems);
+    const norm = items.map(item => ((item.target ??= {}).extensionId = ext.name, item));
```
jakolehm: Not directly related to PR title, this just changes syntax and I'm not 💯 convinced that this is more readable.

src/extensions/registries/page-registry.ts
jakolehm: I agree with @jim-docker , is `const` really that important here?

src/extensions/registries/page-menu-registry.ts
```diff
@@ -30,28 +30,16 @@ export interface PageMenuComponents {
...
```
jakolehm: Not directly related to PR title, this just changes syntax and I'm not 💯 convinced that this is more readable.

panuhorsmalahti: Isin't this also (slightly) semantically different? In the new code extensionId is always overwritten by ext.name, but in the old code extensionId would only be added if missing from target. (No idea if that has any real effect).

jakolehm (approved):

PR #1481, opened 2020-11-23
Track more events from adding a cluster

nevalla:
This PR will add more telemetry events when user adds a new cluster. These will give us better understanding if some phase is constantly failing and we can improve the flow.

### DIFF
--- a/src/renderer/components/+add-cluster/add-cluster.tsx
+++ b/src/renderer/components/+add-cluster/add-cluster.tsx
@@ -24,6 +24,7 @@ import { cssNames } from "../../utils";
 import { Notifications } from "../notifications";
 import { Tab, Tabs } from "../tabs";
 import { ExecValidationNotFoundError } from "../../../common/custom-errors";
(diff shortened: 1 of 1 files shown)


src/renderer/components/+add-cluster/add-cluster.tsx
```diff
@@ -48,6 +49,7 @@ export class AddCluster extends React.Component {
   componentDidMount() {
     clusterStore.setActive(null);
     this.setKubeConfig(userStore.kubeConfigPath);
+    appEventBus.emit({ name: "cluster-add", action: "start" });
```
nevalla, line 52: do we already have some syntax recommendation `cluster-add` vs `clusterAdd`?

aleksfront: Not sure about recommendations but I see that we already has some kebab-case notations, for example in `src/main/node-shell-session.ts`

aleksfront (approved):
LGTM

jakolehm (approved):

PR #1482, opened 2020-11-23
Watch for added/removed local extensions

### LINKED ISSUE #1461: Lens should notice added/removed extension folders
Currently Lens (ExtensionManager) scans extension folder only during boot/init. We should improve this so that Lens detects added/removed folders and notifies ExtensionLoader.

Relates to #1456 and #1199

panuhorsmalahti:
- If extension is copied to the extensions folder, it's dependencies are installed and it's loaded to the application. It's not enabled by default, however
- If the extensions is added from the UI, we need to later implement automatic enabling for it
- Extensions removal is also detected, but not completely implemented.
- Extension classes were slightly refactored to have less responsibilities
- Using chokidar to watch as it should be better than fs.watch
- Tested on macOS only for now
- Fixes #1461


src/main/index.ts
```diff
@@ -79,8 +80,22 @@ app.on("ready", async () => {
   }
 
   extensionLoader.init();
+
+  extensionDiscovery.init();
   windowManager = WindowManager.getInstance<WindowManager>(proxyPort);
-  extensionLoader.initExtensions(await extensionManager.load()); // call after windowManager to see splash earlier
+
+  // call after windowManager to see splash earlier
```
jakolehm, line 87: This reminded me that maybe we should keep splash visible until all extensions are loaded. Story for another PR.

jakolehm (approved):
LGTM

PR #1485, opened 2020-11-23
Resolve extension enabled status when loading it

### LINKED ISSUE #1471: Extensions are silently disabled on Lens restart
After quitting Lens and restarting, extensions that were enabled are no longer enabled. Happening in master and 4.0.0-beta.4

nevalla:
Fixes #1471 


src/extensions/extensions-store.ts
```diff
@@ -34,11 +34,6 @@ export class ExtensionsStore extends BaseStore<LensExtensionsStoreModel> {
     await extensionLoader.whenLoaded;
     await this.whenLoaded;
 
-    // activate user-extensions when state is ready
```
nevalla, line 50: extensionManager has not loaded extensions at this point so we cannot determine the status here.

Nokel81 (approved):

nevalla:
This is conflicted with https://github.com/lensapp/lens/pull/1482, so let's re-implement this after it is merged.

src/extensions/extension-discovery.ts
```diff
@@ -213,24 +214,31 @@ export class ExtensionDiscovery {
...
 
     try {
       // check manifest file for existence
       fs.accessSync(manifestPath, fs.constants.F_OK);
 
       manifestJson = __non_webpack_require__(manifestPath);
+      const installedManifestPath = path.join(this.nodeModulesPath, manifestJson.name, "package.json");
       this.packagesJson.dependencies[manifestJson.name] = path.dirname(manifestPath);
 
+      if (!isBundled) {
```
panuhorsmalahti: [nit] mutation could be avoided: `const isEnabled = isBundled || extensionsStore.isEnabled(installedManifestPath);`

src/extensions/extensions-store.ts
ixrock: Should we add above something like `await extensionManager.whenReady`?

panuhorsmalahti (approved):

src/extensions/extension-discovery.ts
nevalla: 👍 will change

PR #1491, opened 2020-11-23
Fix azure pipeline yarn cache

jakolehm:


panuhorsmalahti:
Once this is verified to fix the issue, can you add explanation of the issue?

jakolehm:
> Once this is verified to fix the issue, can you add explanation of the issue?

I already squashed commits but I did test this with changed yarn.lock file (plus revert). My theory is that without these changes it's possible to get cache from a different OS. If Linux gets cache from Windows then things will go south ...

panuhorsmalahti (approved):

PR #1497, opened 2020-11-24
Use isolatedModules with Babel

jakolehm:
See https://www.typescriptlang.org/tsconfig#isolatedModules


jakolehm:
Babel is gone, no need for this.

PR #1502, opened 2020-11-24
Add LensRendererExtension.isEnabledForCluster

### LINKED ISSUE #1489: Add a function that defines if extension should be enabled for a cluster
Something like this within `LensExtension`:

```ts
    return true;
  }
```

which Lens would use to detect if extension instance needs to be wired to a cluster frame.

jakolehm:
Fixes #1489


Nokel81 (approved):

nevalla (approved):

jakolehm:
MacOS failure seems to be an issue with Azure. Going to hit merge.

PR #1505, opened 2020-11-24
[BREAKING]: remove deprecated routePath before GA

### LINKED ISSUE #1493: Lens Extension API should not be released with some APIs depricated.
**Describe the bug**
Currently `PageRegistration` has the field `routePath` which is marked as `@deprecated`. This API should be removed before GA so that we don't need to support it going forward.

Nokel81:

fixes #1493

### DIFF
--- a/docs/extensions/capabilities/common-capabilities.md
+++ b/docs/extensions/capabilities/common-capabilities.md
@@ -101,7 +101,6 @@ export default class ExampleRendererExtension extends LensRendererExtension {
   globalPages = [
     {
       id: "example",
-      routePath: "/example",
       components: {
         Page: ExamplePage,
       }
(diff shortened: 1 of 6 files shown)


jakolehm (approved):

ixrock:
@jakolehm just curios: what has changed from now and then, when I suggested to remove this field, instead of keeping deprecated as you insisted? _(during page-registry refactoring)_

jakolehm:
> @jakolehm just curios: what has changed from now and then, when I suggested to remove this field, instead of keeping deprecated as you insisted? _(during page-registry refactoring)_

We had too many breaking changes while we refactored this, that's the only reason. As #1493 describes we should remove all deprecated stuff before we cut RC/GA.

PR #1510, opened 2020-11-25
Tweak install extension ux

jakolehm:
Tweaks #1456

image


ixrock:
> image

Is it looks better now? Or there are was some useless info? (right column) 🙈

ixrock (approved):

jakolehm:
> Is it looks better now? Or there are was some useless info? (right column) 🙈

Rewrote texts and took unnecessary information away. This UI is for Lens users, not Lens extension developers.. less is more 🙂 .

ixrock:
`Write less, do more` © jQuery :D

>> This UI is for Lens users, not Lens extension developers..

Some users already requested to clean up UI? Or how do you know what they need?

Sebastian for example already has used  the tip for getting tarball url from NPM...

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -269,55 +268,31 @@ export class Extensions extends React.Component {
...
             validators={InputValidators.isUrl}
             value={this.downloadUrl}
             onChange={v => this.downloadUrl = v}
             onSubmit={this.addExtensions}
           />
           <Button
             primary
-            label="Add extensions"
+            label="Install"
             onClick={this.addExtensions}
```
ixrock, line 294: [nit]™ for sake of consistency ^_^
```suggestion
            label="Install"
            onClick={this.installExtensions}
```

jakolehm:
> > > This UI is for Lens users, not Lens extension developers..
> 
> Some users already requested to clean up UI? Or how do you know what they need?
> 
> Sebastian for example already has used tip for getting tarball url from NPM...

I don't really get your point, this UI is targeted for Lens users so internals of extension API are not exactly what they need to see (or know). That said this PR is just my idea how UI would be cleaner and more user friendly. There are still pieces that needs to be tweaked, for example url input should follow with an icon to browse extension package (see how kubeconfig browse works). But we should improve these in follow-up PRs so that we can get #1456 approved & merged sooner. 

Let's wait for other maintainers opinions (especially @miskun & @aleksfront).

aleksfront:
1. Install button doesn't actually install anything, but shows Open dialog, it can be confusing. For example, when you paste a valid url and hit install, the behavior isn't appears to be right.
2. Error object needs to be parsed into string and be more clear. Now we have `[object Object]` notation and some unclear message.
ext error 1
ext error 2

jakolehm:
> 1. Install button doesn't actually install anything, but shows Open dialog, it can be confusing. For example, when you paste a valid url and hit install, the behavior isn't appears to be right.

Yes I agree, it's out of scope but we definitely should improve that (I already mentioned that we should streamline input/button to look like kubeconfig input).

> 2. Error object needs to be parsed into string and be more clear. Now we have `[object Object]` notation and some unclear message.
> 

Also out of scope, lets open separate issues about these.

ixrock:
> * Install button doesn't actually install anything, but shows Open dialog, it can be confusing. For example, when you paste a valid url and hit install, the behavior isn't appears to be right.

Install button designed as single "gateway" to install:
- when URL is filled it acts as "install-from-url"
- when there is no URL it opens select-files-dialog (but now with simplified UX it's unclear indeed)
- also there is no mentioning about D&D support to install files with new layout

> * Error object needs to be parsed into string and be more clear. Now we have `[object Object]` notation and some unclear message.

- Yeah, this bug is happening when URL is pointing not-a-tar file or non-existing but valid URL, therefore some object/non-stringifiable error is returned and we have `[object Object]` and other irrelevant error messages.

jakolehm:
> > * Install button doesn't actually install anything, but shows Open dialog, it can be confusing. For example, when you paste a valid url and hit install, the behavior isn't appears to be right.
> 
> Install button designed as single "gateway" to install:
> 
> * when URL is filled it acts as "install-from-url"
> * when there is no URL it opens select-files-dialog (but now with simplified UX it's unclear indeed)
> * also there is no mentioning about D&D support to install files with new layout

To me the biggest usability issue here is this single "gateway" button. It's really hard to understand if input field is mandatory or not. That's why I'm suggesting something that we already have in "browse kubeconfig" UI (Add Cluster page). My UI skills are not enough for this kind of change so let's fix these in follow-up PRs 🙂 

Drag & drop: maybe we need to put it back.. it's not something that usually is tipped in app UI's but I just realised that we have similar pro-tip in "Add Cluster".

jakolehm:
@ixrock d&d pro-tip is back 🎉

PR #1516, opened 2020-11-25
Pod logs refactoring

aleksfront:
Refactoring log components by spreading `PodLogs` monolith to logical and presentational ones.

Also fixing founded bugs including:
* Unclear appearing of jump to bottom button
* Loading more and more logs when switching tabs
* Log controls blinking

funbox logs


src/renderer/components/dock/pod-log-list.tsx
```diff
@@ -0,0 +1,224 @@
...
+      return;
+    }
+    if (scrolledToBeginning && newLogsLoaded) {
+      this.virtualListDiv.current.scrollTop = (logs.length - prevProps.logs.length) * this.lineHeight;
+    }
+    if (fewLogsLoaded) {
+      this.isJumpButtonVisible = false;
+    }
+    if (!logs.length) {
+      this.isLastLineVisible = false;
```
Nokel81, line 63: This field really sounds like it is computed. Shouldn't it be marked as such?

src/renderer/components/dock/pod-log-list.tsx
```diff
@@ -0,0 +1,224 @@
...
+      pieces.forEach((piece, index) => {
+        const active = isActiveOverlay(rowIndex, index);
+        const lastItem = index === pieces.length - 1;
+        const overlayValue = matches.next().value;
+        const overlay = !lastItem ?
+          <span
+            className={cssNames("overlay", { active })}
+            dangerouslySetInnerHTML={{ __html: ansiToHtml(overlayValue) }}
+          /> :
+          null;
```
Nokel81: This format is rather confusing to read.
```suggestion
        const overlay = !lastItem
          ? <span
            className={cssNames("overlay", { active })}
            dangerouslySetInnerHTML={{ __html: ansiToHtml(overlayValue) }}
          />
          : null;
```

src/renderer/components/dock/pod-logs.tsx
```diff
@@ -1,67 +1,34 @@
...
-  private colorConverter = new AnsiUp();
+  private logListElement = React.createRef<PodLogList>(); // A reference for VirtualList component
 
   componentDidMount() {
     disposeOnUnmount(this, [
       reaction(() => this.props.tab.id, async () => {
-        await this.load();
-        this.scrollToBottom();
+        await this.reload();
       }, { fireImmediately: true }),
```
Nokel81: Couldn't we just pass `this.reload` to the reaction directly?

Nokel81 (approved):

src/renderer/components/dock/pod-logs.tsx
aleksfront: Can you elaborate this a bit please?

Nokel81: Sure. It seems like we have the following snippet. `async () => { await this.reload(); }`. I wonder if we could instead just have, `this.reload` if it has been marked with `@autobind()`.

src/renderer/components/dock/pod-log-list.tsx
aleksfront: I don't think so. This is an observable flag not really different from other ones like `isJumpButtonVisible`. A `computed` is used to calculate some value from other observables. `isLastLineVisible` just holds bool value and doesn't need to be "computed".

src/renderer/components/dock/pod-logs.tsx
aleksfront: Yes, it works as expected. Thanks for review!

PR #1518, opened 2020-11-25
Remove extension when folder is removed during runtime

### LINKED ISSUE #1461: Lens should notice added/removed extension folders

panuhorsmalahti:
- prettier is added only to support inline snapshots
- If user has extension page open and the extension folder is deleted, "Page not found" 'error' will be displayed. We could redirect to a working page. I'll add issue for this.

Fixes #1461

### DIFF
--- a/src/extensions/extension-loader.ts
+++ b/src/extensions/extension-loader.ts
@@ -1,4 +1,5 @@
 import { app, ipcRenderer, remote } from "electron";
+import { EventEmitter } from "events";
 import { action, computed, observable, reaction, toJS, when } from "mobx";
 import path from "path";
 import { getHostedCluster } from "../common/cluster-store";
@@ -26,26 +27,32 @@ export class ExtensionLoader {
   protected instances = observable.map<LensExtensionId, LensExtension>();
   protected readonly requestExtensionsChannel = "extensions:loaded";
 
+  // emits event "remove" of type LensExtension when the extension is removed
+  private events = new EventEmitter();
+
   @observable isLoaded = false;
   whenLoaded = when(() => this.isLoaded);
 
   @computed get userExtensions(): Map<LensExtensionId, InstalledExtension> {
     const extensions = this.extensions.toJS();
+
     extensions.forEach((ext, extId) => {
       if (ext.isBundled) {
         extensions.delete(extId);
       }
     });
+
     return extensions;
   }
 
   @action
   async init() {
     if (ipcRenderer) {
-      this.initRenderer();
+      await this.initRenderer();
     } else {
-      this.initMain();
+      await this.initMain();
     }
+
     extensionsStore.manageState(this);
   }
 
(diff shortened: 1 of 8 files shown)


src/extensions/extension-loader.ts
```diff
@@ -57,11 +64,28 @@ export class ExtensionLoader {
...
+    if (instance) {
+      try {
+        instance.disable();
+        this.events.emit("remove", instance);
+        this.instances.delete(lensExtensionId);
+      } catch (error) {
+        logger.error(`${logModule}: deactivation extension error`, { lensExtensionId, error });
+      }
+    }
+  }
```
Nokel81, line 80: What do you think about the `if (!instance) { return; }` style instead?

src/extensions/extension-loader.ts
```diff
@@ -57,11 +64,28 @@ export class ExtensionLoader {
...
+
   removeExtension(lensExtensionId: LensExtensionId) {
-    // TODO: Remove the extension properly (from menus etc.)
+    this.removeInstance(lensExtensionId);
+
     if (!this.extensions.delete(lensExtensionId)) {
       throw new Error(`Can't remove extension ${lensExtensionId}, doesn't exist.`);
     }
+
```
Nokel81, line 88: New line at the end of a function?

yarn.lock
```diff
@@ -11480,6 +11480,11 @@ prepend-http@^2.0.0:
   resolved "https://registry.yarnpkg.com/prepend-http/-/prepend-http-2.0.0.tgz#e92434bfa5ea8c19f41cdfd401d741a3c819d897"
   integrity sha1-6SQ0v6XqjBn0HN/UAddBo8gZ2Jc=
 
+prettier@^2.2.0:
+  version "2.2.0"
+  resolved "https://registry.yarnpkg.com/prettier/-/prettier-2.2.0.tgz#8a03c7777883b29b37fb2c4348c66a78e980418b"
+  integrity sha512-yYerpkvseM4iKD/BXLYUkQV5aKt4tQPqaGW6EsZjzyu0r7sVZZNPJW4Y8MyKmicp6t42XUPcBVA+H6sB3gqndw==
+
```
Nokel81, line 11487: I don't see any references to it in any of this diff. Are you sure you are actually using it for your intended purpose?

panuhorsmalahti: That was mentioned in the issue description, jest requires this dependency for inline snapshots: https://github.com/facebook/jest/issues/8467

Nokel81: Ah I saw that comment, but I assumed that it would have to be explicitly "used" somewhere.

jakolehm (approved):

src/extensions/extension-loader.ts
panuhorsmalahti: I don't see that being necessarily better, but fixed

src/extensions/extension-loader.ts
panuhorsmalahti: Fixed

PR #1521, opened 2020-11-25
Replace cluster warning event polling with watches

jakolehm:
Warning event polling is legacy from pre-3.6 era. It seems that 3.6 refactoring also (accidentally?) changed how badge is shown. It should not be visible for an active cluster.

### DIFF
--- a/src/renderer/components/app.tsx
+++ b/src/renderer/components/app.tsx
@@ -73,11 +73,17 @@ export class App extends React.Component {
   }
 
   async componentDidMount() {
-    console.log("hello");
     const cluster = getHostedCluster();
-    await eventStore.loadAll();
-    await nodesStore.loadAll();
-    await podsStore.loadAll();
+    const promises: Promise<void>[] = [];
+    if (isAllowedResource("events") && isAllowedResource("pods")) {
+      promises.push(eventStore.loadAll());
+      promises.push(podsStore.loadAll());
+    }
+    if (isAllowedResource("nodes")) {
+      promises.push(nodesStore.loadAll());
+    }
+    await Promise.all(promises);
+
     eventStore.subscribe();
     nodesStore.subscribe();
     podsStore.subscribe();
--- a/src/renderer/components/app.tsx
+++ b/src/renderer/components/app.tsx
@@ -47,6 +47,7 @@ import { eventStore } from "./+events/event.store";
 import { reaction, computed } from "mobx";
 import { nodesStore } from "./+nodes/nodes.store";
 import { podsStore } from "./+workloads-pods/pods.store";
+import { sum } from "lodash";
 
 @observer
 export class App extends React.Component {
@@ -83,10 +84,13 @@ export class App extends React.Component {
       promises.push(nodesStore.loadAll());
     }
     await Promise.all(promises);
-
-    eventStore.subscribe();
-    nodesStore.subscribe();
-    podsStore.subscribe();
+    if (eventStore.isLoaded && podsStore.isLoaded) {
+      eventStore.subscribe();
+      podsStore.subscribe();
+    }
+    if (nodesStore.isLoaded) {
+      nodesStore.subscribe();
+    }
 
     reaction(() => this.warningsCount, (count) => {
       broadcastMessage(`cluster-warning-event-count:${cluster.id}`, count);
@@ -95,10 +99,8 @@ export class App extends React.Component {
 
   @computed
   get warningsCount() {
-    let warnings = 0;
-    nodesStore.items.forEach(node => {
-      warnings = warnings + node.getWarningConditions().length;
-    });
+    let warnings = sum(nodesStore.items
+      .map(node => node.getWarningConditions().length));
     warnings = warnings + eventStore.getWarnings().length;
 
     return warnings;
--- a/src/renderer/components/app.tsx
+++ b/src/renderer/components/app.tsx
@@ -37,13 +37,16 @@ import { webFrame } from "electron";
 import { clusterPageRegistry, getExtensionPageUrl, PageRegistration, RegisteredPage } from "../../extensions/registries/page-registry";
 import { extensionLoader } from "../../extensions/extension-loader";
 import { appEventBus } from "../../common/event-bus";
-import { requestMain } from "../../common/ipc";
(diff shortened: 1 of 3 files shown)


src/renderer/components/app.tsx
```diff
@@ -69,6 +72,32 @@ export class App extends React.Component {
     whatInput.ask(); // Start to monitor user input device
   }
 
+  async componentDidMount() {
+    console.log("hello");
```
Nokel81:
```suggestion
```

src/renderer/components/app.tsx
```diff
@@ -69,6 +72,32 @@ export class App extends React.Component {
...
+    const cluster = getHostedCluster();
+    await eventStore.loadAll();
+    await nodesStore.loadAll();
+    await podsStore.loadAll();
```
Nokel81: If these are not dependent of each other then should be done concurrently.

```suggestion
    await Promise.all([
      eventStore.loadAll(),
      nodesStore.loadAll(),
      podsStore.loadAll(),
    ]);
```

src/renderer/components/app.tsx
```diff
@@ -69,6 +72,32 @@ export class App extends React.Component {
...
+
+  @computed
+  get warningsCount() {
+    let warnings = 0;
+    nodesStore.items.forEach(node => {
+      warnings = warnings + node.getWarningConditions().length;
+    });
+    warnings = warnings + eventStore.getWarnings().length;
+
+    return warnings;
```
Nokel81:
```suggestion
    const nodeWarnings = _.sum(nodesStore.items
      .map(node => node.getWarningConditions().length));
    const eventWarnings = eventStore.getWarnings().length; 

    return nodeWarnings + eventWarnings;
```

src/renderer/components/cluster-icon/cluster-icon.tsx
```diff
@@ -29,12 +33,31 @@ const defaultProps: Partial<Props> = {
...
+
+  get eventCountBroadcast() {
+    return `cluster-warning-event-count:${this.props.cluster.id}`;
+  }
+
+  componentDidMount() {
+    const subscriber = subscribeToBroadcast(this.eventCountBroadcast, (ev, eventCount) => {
+      if (eventCount != this.eventCount) {
+        this.eventCount = eventCount;
+      }
```
Nokel81: The outcome seems to be the same as:
```suggestion
      this.eventCount = eventCount;
```
so why the check?

src/renderer/components/app.tsx
panuhorsmalahti: Remove console.log

src/renderer/components/cluster-icon/cluster-icon.tsx
```diff
@@ -48,7 +71,7 @@ export class ClusterIcon extends React.Component<Props> {
         )}
         {icon && <img src={icon} alt={name}/>}
         {!icon && <Hashicon value={clusterId} options={options}/>}
-        {showErrors && isAdmin && eventCount > 0 && (
+        {showErrors && eventCount > 0 && !isActive && (
```
Nokel81, line 72: So non-admins now are allowed to see events?

src/renderer/components/app.tsx
```diff
@@ -69,6 +72,32 @@ export class App extends React.Component {
...
+
+    reaction(() => this.warningsCount, (count) => {
+      broadcastMessage(`cluster-warning-event-count:${cluster.id}`, count);
+    });
+  }
+
+  @computed
+  get warningsCount() {
+    let warnings = 0;
+    nodesStore.items.forEach(node => {
```
panuhorsmalahti: Array.prototype.reduce is better for reducing an array to a single value

src/renderer/components/app.tsx
jakolehm: Oops 🤦

src/renderer/components/cluster-icon/cluster-icon.tsx
jakolehm: Yes, idk why we checked `isAdmin` here.. previously `eventCount` calculation (in main) was only done for admins.

src/renderer/components/app.tsx
Nokel81: Or even [lodash's sum](https://lodash.com/docs/4.17.15#sum)

src/renderer/components/app.tsx
jakolehm: They are not, I was mainly worried about hitting kubernetes api too hard (if these collections are huge). But maybe it's better to do these concurrently.

jakolehm: Actually, I see a problem here.. we need to check if user has access to these resources before trying to load them.

jakolehm: Fixed.

src/renderer/components/cluster-icon/cluster-icon.tsx
jakolehm: Fixed.

src/renderer/components/app.tsx
jakolehm: Fixed.

src/renderer/components/app.tsx
```diff
@@ -73,11 +73,17 @@ export class App extends React.Component {
...
+      promises.push(podsStore.loadAll());
+    }
+    if (isAllowedResource("nodes")) {
+      promises.push(nodesStore.loadAll());
+    }
+    await Promise.all(promises);
+
     eventStore.subscribe();
     nodesStore.subscribe();
     podsStore.subscribe();
```
Nokel81: Shouldn't we only subscribe if we are allowed to load? Or does it not matter?

src/renderer/components/cluster-icon/cluster-icon.tsx
Nokel81: Okay sounds good.

src/renderer/components/app.tsx
jakolehm: Yes we should only subscribe then.. this is already fixed 🙂

Nokel81: Ah race condition

Nokel81 (approved):

nevalla (approved):

PR #1522, opened 2020-11-25
Installing extensions UI improvements

### LINKED ISSUE #1513: Extension UI improvements
* Install input should accept both filesystem path and URL
* input should follow with "browse" button
*  install should be disabled until input has valid path/url
* remove 2 column layout, follow similar layout as we have on preferences page

Example of input/browse/install buttons:
image

ixrock:
_Features_
- redesigned & simplified UI/UX
- allow to install extension from system path besides URL

_Bugfixes_
- `InputValudators.isUrl` and `InputValudators.isPath` should not be valid for empty string
- `<SearchInput>` clear-text-icon was broken without actually required `onClear` prop
- `<DropFileInput>` was skipping `className` from modifying target element

close #1513

_Screenshots_

Screenshot 2020-11-25 at 15 25 31
Screenshot 2020-11-25 at 15 25 47

### DIFF
--- a/package.json
+++ b/package.json
@@ -225,7 +225,6 @@
     "electron-devtools-installer": "^3.1.1",
     "electron-updater": "^4.3.1",
     "electron-window-state": "^5.0.3",
-    "file-type": "^14.7.1",
     "filenamify": "^4.1.0",
     "fs-extra": "^9.0.1",
     "handlebars": "^4.7.6",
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -225,7 +225,6 @@
     "electron-devtools-installer": "^3.1.1",
     "electron-updater": "^4.3.1",
     "electron-window-state": "^5.0.3",
-    "file-type": "^14.7.1",
```
ixrock, line 228: Unused package. Tried to validate for mime-type, but seems doesn't work for `.tgz`

jakolehm (approved):
LGTM. Big improvement 👍

PR #1528, opened 2020-11-25
add step to verify docs on area/documentation PRs

### LINKED ISSUE #1268: Verify that docs markdown/config is correct on each PR
We should validate each PR with mkdocs. Maybe these could be done only if PR has `area/documentation` label.

Nokel81:

fixes #1268


.azure-pipelines.yml
```diff
@@ -162,3 +162,27 @@ jobs:
         condition: "and(succeeded(), startsWith(variables['Build.SourceBranch'], 'refs/tags/'))"
         env:
           NPM_TOKEN: $(NPM_TOKEN)
+---
+variables:
```
jakolehm: I think it would make sense to have this test on GH actions side where documentation is generated. See https://github.com/lensapp/lens/blob/master/.github/workflows/main.yml . Or if kept here then run only if docs are changed.

chenhunghan (approved):
Looks good, but why not just use Github Actions? maybe in the future...

jakolehm (dismissed):
Yaml might not be valid because I don't see any checks (except DCO) here?

.azure-pipelines.yml
Nokel81: Makes sense, will try using GH actions.

Nokel81:
@jakolehm GH has been set up, and it passes after adding the WIP pages.

PR #1532, opened 2020-11-26
Column filters

pashevskii:
the PR adds a new menu where a user can change the visibility of columns
image

Now this feature is available to Pods table only (just for example) , we can add the column filters adding 2 props in ItemListLayout (tableId = "some uniq name" and isConfigurable = true)
All changed filtering information is stored in .config/LensDev/config.json
 as "hiddenTableColumns": {"tableId": \<array of hidden columns\>} 

Also, the commit adds a new showWithCollumn property to TableCellProps, this property is for some columns, which have no title and its behaviour should be the same as specified header

### DIFF
--- a/src/common/user-store.ts
+++ b/src/common/user-store.ts
@@ -28,6 +28,7 @@ export interface UserPreferences {
   downloadBinariesPath?: string;
   kubectlBinariesPath?: string;
   openAtLogin?: boolean;
+  hiddenTableColumns?: Record<string, string[]>
 }
 
 export class UserStore extends BaseStore<UserStoreModel> {
@@ -54,6 +55,7 @@ export class UserStore extends BaseStore<UserStoreModel> {
     downloadMirror: "default",
     downloadKubectlBinaries: true,  // Download kubectl binaries matching cluster version
     openAtLogin: false,
+    hiddenTableColumns: {},
   };
 
   protected async handleOnLoad() {
--- a/src/renderer/components/+workloads-pods/pods.tsx
+++ b/src/renderer/components/+workloads-pods/pods.tsx
@@ -74,6 +74,8 @@ export class Pods extends React.Component<Props> {
       <KubeObjectListLayout
         className="Pods" store={podsStore}
         dependentStores={[volumeClaimStore, eventStore]}
+        tableId = "workloads_pods"
+        isConfigurable
         sortingCallbacks={{
           [sortBy.name]: (pod: Pod) => pod.getName(),
           [sortBy.namespace]: (pod: Pod) => pod.getNs(),
--- a/src/renderer/components/+workloads-pods/pods.tsx
+++ b/src/renderer/components/+workloads-pods/pods.tsx
@@ -96,7 +96,7 @@ export class Pods extends React.Component<Props> {
         renderHeaderTitle="Pods"
         renderTableHeader={[
           { title: "Name", className: "name", sortBy: sortBy.name },
-          { className: "warning" },
+          { className: "warning", showWithColumn: "name" },
           { title: "Namespace", className: "namespace", sortBy: sortBy.namespace },
           { title: "Containers", className: "containers", sortBy: sortBy.containers },
           { title: "Restarts", className: "restarts", sortBy: sortBy.restarts },
--- a/src/renderer/components/item-object-list/item-list-layout.tsx
+++ b/src/renderer/components/item-object-list/item-list-layout.tsx
@@ -18,6 +18,11 @@ import { PageFiltersList } from "./page-filters-list";
 import { PageFiltersSelect } from "./page-filters-select";
 import { NamespaceSelectFilter } from "../+namespaces/namespace-select";
 import { themeStore } from "../../theme.store";
+import { MenuActions} from "../menu/menu-actions";
+import { MenuItem } from "../menu";
+import { Checkbox } from "../checkbox";
+import { userStore } from "../../../common/user-store";
+import logger from "../../../main/logger";
 
 // todo: refactor, split to small re-usable components
 
@@ -32,6 +37,7 @@ interface IHeaderPlaceholders {
 }
 
 export interface ItemListLayoutProps<T extends ItemObject = ItemObject> {
+  tableId?: string;
   className: IClassName;
   store: ItemStore<T>;
   dependentStores?: ItemStore[];
@@ -50,6 +56,7 @@ export interface ItemListLayoutProps<T extends ItemObject = ItemObject> {
   isReady?: boolean; // show loading indicator while not ready
   isSelectable?: boolean; // show checkbox in rows for selecting items
   isSearchable?: boolean; // apply search-filter & add search-input
+  isConfigurable?: boolean;
   copyClassNameFromHeadCells?: boolean;
   sortingCallbacks?: { [sortBy: string]: TableSortCallback };
   tableProps?: Partial<TableProps>; // low-level table configuration
@@ -74,6 +81,7 @@ const defaultProps: Partial<ItemListLayoutProps> = {
   showHeader: true,
   isSearchable: true,
   isSelectable: true,
+  isConfigurable: false,
   copyClassNameFromHeadCells: true,
   dependentStores: [],
   filterItems: [],
@@ -89,7 +97,7 @@ interface ItemListLayoutUserSettings {
 @observer
 export class ItemListLayout extends React.Component<ItemListLayoutProps> {
   static defaultProps = defaultProps as object;
-
+  @observable hiddenColumnNames = new Set<string>();
   @observable isUnmounting = false;
 
   // default user settings (ui show-hide tweaks mostly)
@@ -111,7 +119,9 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
   }
 
   async componentDidMount() {
-    const { store, dependentStores, isClusterScoped } = this.props;
+    const { store, dependentStores, isClusterScoped, isConfigurable, tableId, renderTableHeader } = this.props;
+    if (this.canBeConfigured) this.hiddenColumnNames = new Set(userStore.preferences?.hiddenTableColumns?.[tableId]);
+
     const stores = [store, ...dependentStores];
 
     if (!isClusterScoped) stores.push(namespaceStore);
@@ -129,9 +139,11 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
 
   componentWillUnmount() {
     this.isUnmounting = true;
-    const { store, isSelectable } = this.props;
-
+    const { store, isSelectable, isConfigurable, tableId } = this.props;
     if (isSelectable) store.resetSelection();
+    if (this.canBeConfigured) {
+      userStore.preferences.hiddenTableColumns[tableId] = Array.from(this.hiddenColumnNames);
+    }
   }
 
   private filterCallbacks: { [type: string]: ItemsFilter } = {
@@ -215,6 +227,35 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
     return this.applyFilters(filterItems, allItems);
   }
 
+  updateColumnFilter(checkboxValue: boolean, columnName: string) {
+    if (checkboxValue){
+      this.hiddenColumnNames.delete(columnName);
+    } else {
+      this.hiddenColumnNames.add(columnName);
+    }
+  }
+
+  showColumn(index: number): boolean {
+    const {isConfigurable, tableId , renderTableHeader} = this.props;
+    if (!this.canBeConfigured) return true;
+    return !this.hiddenColumnNames.has(renderTableHeader[index].className);
+  }
+
+  @computed get canBeConfigured(): boolean {
+    const { isConfigurable, tableId, renderTableHeader } = this.props;
+
+    if (!isConfigurable || !tableId) {
+      return false;
+    }
+
+    if (!renderTableHeader?.every(({ className }) => className)) {
+      logger.warning("[ItemObjectList]: cannot configure an object list without all headers being identifiable");
+      return false;
+    }
+
+    return true;
+  }
+
   @autobind()
   getRow(uid: string) {
     const {
@@ -257,8 +298,7 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
                   cellProps.className = cssNames(cellProps.className, headCell.className);
                 }
               }
-
-              return <TableCell key={index} {...cellProps} />;
+              return this.showColumn(index) ? <TableCell key={index} {...cellProps} /> : null;
             })
         }
         {renderItemMenu && (
@@ -395,8 +435,9 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
 
   renderList() {
     const {
-      isSelectable, tableProps = {}, renderTableHeader, renderItemMenu,
-      store, hasDetailsView, addRemoveButtons = {}, virtual, sortingCallbacks, detailsItem
+      isSelectable, tableProps = {}, renderTableHeader,
+      store, hasDetailsView, addRemoveButtons = {}, virtual, sortingCallbacks, detailsItem,
+      isConfigurable, tableId
     } = this.props;
     const { isReady, removeItemsDialog, items } = this;
     const { selectedItems } = store;
@@ -430,14 +471,19 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
                     onClick={prevDefault(() => store.toggleSelectionAll(items))}
                   />
                 )}
-                {renderTableHeader.map((cellProps, index) => <TableCell key={index} {...cellProps} />)}
-                {renderItemMenu && <TableCell className="menu" />}
+                {renderTableHeader.map((cellProps, index) => this.showColumn(index) ? <TableCell key={index} {...cellProps} /> : null)}
+                {
+                  <TableCell className="menu" >
+                    {this.canBeConfigured && this.renderColumnMenu()}
+                  </TableCell>
+                }
               </TableHead>
             )}
             {
               !virtual && items.map(item => this.getRow(item.getId()))
             }
           </Table>
+
         )}
         <AddRemoveButtons
           onRemove={selectedItems.length ? removeItemsDialog : null}
@@ -445,6 +491,25 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
           {...addRemoveButtons}
         />
       </div>
+    );
+  }
+
+  renderColumnMenu() {
+    const { renderTableHeader} = this.props;
+    return(
+      <MenuActions
+        toolbar = {false}
+        autoCloseOnSelect = {false}
+        className={cssNames("KubeObjectMenu")}
+      >
+        {renderTableHeader.map((cellProps, index) => (
+          <MenuItem key={index}>
+            <Checkbox label = {cellProps.title ? cellProps.title : `<${cellProps.className}>`}
+              value ={!this.hiddenColumnNames.has(cellProps.className)}
+              onChange = {v =>this.updateColumnFilter(v, cellProps.className)} />
+          </MenuItem>))
+        }
+      </MenuActions>
     );
   }
 
--- a/src/renderer/components/item-object-list/item-list-layout.tsx
+++ b/src/renderer/components/item-object-list/item-list-layout.tsx
@@ -1,4 +1,5 @@
 import "./item-list-layout.scss";
+import "./table-menu.scss";
 import groupBy from "lodash/groupBy";
 
 import React, { ReactNode } from "react";
@@ -119,7 +120,8 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
   }
 
   async componentDidMount() {
-    const { store, dependentStores, isClusterScoped, isConfigurable, tableId, renderTableHeader } = this.props;
+    const { store, dependentStores, isClusterScoped, tableId } = this.props;
+
     if (this.canBeConfigured) this.hiddenColumnNames = new Set(userStore.preferences?.hiddenTableColumns?.[tableId]);
 
     const stores = [store, ...dependentStores];
@@ -139,11 +141,9 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
 
   componentWillUnmount() {
     this.isUnmounting = true;
-    const { store, isSelectable, isConfigurable, tableId } = this.props;
+    const { store, isSelectable } = this.props;
+
     if (isSelectable) store.resetSelection();
-    if (this.canBeConfigured) {
-      userStore.preferences.hiddenTableColumns[tableId] = Array.from(this.hiddenColumnNames);
-    }
   }
 
   private filterCallbacks: { [type: string]: ItemsFilter } = {
@@ -233,15 +233,21 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
     } else {
       this.hiddenColumnNames.add(columnName);
     }
+
+    if (this.canBeConfigured) {
+      userStore.preferences.hiddenTableColumns[this.props.tableId] = Array.from(this.hiddenColumnNames);
+    }
   }
 
-  showColumn(index: number): boolean {
-    const {isConfigurable, tableId , renderTableHeader} = this.props;
+  columnIsVisible(index: number): boolean {
+    const {renderTableHeader} = this.props;
+
     if (!this.canBeConfigured) return true;
-    return !this.hiddenColumnNames.has(renderTableHeader[index].className);
+
+    return !this.hiddenColumnNames.has(renderTableHeader[index].showWithColumn ?? renderTableHeader[index].className);
   }
 
-  @computed get canBeConfigured(): boolean {
+  get canBeConfigured(): boolean {
     const { isConfigurable, tableId, renderTableHeader } = this.props;
 
     if (!isConfigurable || !tableId) {
@@ -250,6 +256,7 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
 
     if (!renderTableHeader?.every(({ className }) => className)) {
       logger.warning("[ItemObjectList]: cannot configure an object list without all headers being identifiable");
+
       return false;
     }
 
@@ -298,7 +305,8 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
                   cellProps.className = cssNames(cellProps.className, headCell.className);
                 }
               }
-              return this.showColumn(index) ? <TableCell key={index} {...cellProps} /> : null;
+
+              return this.columnIsVisible(index) ? <TableCell key={index} {...cellProps} /> : null;
             })
         }
         {renderItemMenu && (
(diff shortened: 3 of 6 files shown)


src/common/user-store.ts
```diff
@@ -28,6 +28,7 @@ export interface UserPreferences {
   downloadBinariesPath?: string;
   kubectlBinariesPath?: string;
   openAtLogin?: boolean;
+  tablesConfiguration?: Map<string, number>
```
Nokel81: This type needs to be `JSON.strignify`-able. Please change this to `Record<K, V>`.

src/common/user-store.ts
```diff
@@ -54,6 +55,7 @@ export class UserStore extends BaseStore<UserStoreModel> {
     downloadMirror: "default",
     downloadKubectlBinaries: true,  // Download kubectl binaries matching cluster version
     openAtLogin: true,
+    tablesConfiguration: new Map<string, number>(),
```
Nokel81: This should be `observable.Map`

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -111,7 +118,12 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
   }
 
   async componentDidMount() {
-    const { store, dependentStores, isClusterScoped } = this.props;
+    const { store, dependentStores, isClusterScoped, isConfigurable, tableId } = this.props;
+    if (isConfigurable && tableId) {
+      const filters = get(userStore.preferences.tablesConfiguration, tableId);
```
Nokel81: We don't seem to use `get` and `set` functions from `mobx` in this repo. Please change to plain accesses.
```suggestion
      const filters = userStore.preferences.tablesConfiguration.get(tableId);
```

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -111,7 +118,12 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
+      if (filters != undefined) this.columnFilter = filters;
+      else this.columnFilter = ( 2 ** this.props.renderTableHeader.length ) - 1;
```
Nokel81: this.columnFilter = filters ?? (( 2 ** this.props.renderTableHeader.length ) - 1);

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -126,8 +138,11 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
 
   componentWillUnmount() {
     this.isUnmounting = true;
-    const { store, isSelectable } = this.props;
+    const { store, isSelectable, isConfigurable, tableId } = this.props;
     if (isSelectable) store.resetSelection();
+    if (isConfigurable && tableId) {
+      set(userStore.preferences.tablesConfiguration ,tableId, this.columnFilter);
```
Nokel81:
```suggestion
      userStore.preferences.tablesConfiguration.set(tableId, this.columnFilter);
```

src/renderer/components/menu/menu-actions.tsx
```diff
@@ -94,7 +95,7 @@ export class MenuActions extends React.Component<MenuActionsProps> {
           className={menuClassName}
           usePortal={autoClose}
           closeOnScroll={autoClose}
-          closeOnClickItem={autoClose}
+          closeOnClickItem={autoCloseOnSelect == undefined ? autoClose : autoCloseOnSelect}
```
Nokel81:
```suggestion
          closeOnClickItem={autoCloseOnSelect ?? autoClose}
```

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -197,6 +212,20 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
     return this.applyFilters(filterItems, allItems);
   }
 
+  updateColumnFilter(checkboxValue: boolean, index: number) {
+    if (checkboxValue){
+      this.columnFilter = this.columnFilter | ( 1 << index );
+    } else {
+      this.columnFilter = this.columnFilter ^ ( 1 << index );
+    }
+  }
```
Nokel81: Why all this complications about numbers? Since the goal of the feature is to only show specific columns. Wouldn't just a `Set<ColumnId>` be sufficiant (and then stored in the store as `ColumnId[]`)?

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -197,6 +212,20 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
+
+  showColumn(index: number): boolean {
+    const {isConfigurable, tableId } = this.props;
+    if (!isConfigurable || !tableId) return true;
+    return ( this.columnFilter & 1 << index ) > 0;
+  }
```
Nokel81: This current form is very susceptible to breaking (or at least odd behaviour) if we have `> 52` columns or if we ever rearrange the columns. The former is most likely a non-issue but the later is very much on the table. Since we plan on supporting re-sizable and rearrange-able columns in tables.

I am of the opinion that this feature should be designed with those in mind.

Nokel81 (changes requested):

src/renderer/components/item-object-list/item-list-layout.tsx
pashevskii: @Nokel81 Thanks for your comments, how do you feel about storing all columns settings as Record<String, Object> on user-store.ts side and work with the object here as obj{name: {visible: true}, ..., age {visible: false}}  where name/age are just column names and represent a container of all related properties (visibility for now, wight, order, etc. for the future)? 
I think using column names instead of ids is safer and more user-friendly

Nokel81: I definitely agree that column names are better than ids. I wonder why the store shouldn't just be `string[]` and then in the code it be a `Set<string>`? This way it seems simpler than doing an object with a `visible` field.

pashevskii: Hmm, I was sure that we should show New columns, which will be added in master code but not hidden by the user manually

pashevskii: @Nokel81 Fixed. One more question, how about ignoring table config saving on 
`this.columnProps = = this.props.renderTableHeader.map(cellProps => { return cellProps.className; })`
I mean the case when the user didn't change anything, I think we can skip columns saving and load those from renderTableHeader on the next component mounting

Also, I would like to do something with unlabeled columns (like 'warning'), my idea is about linking these columns behavior with the nearest labeled columns (for instance pods.name + pods.warning). Does it make sense?

Nokel81: > Hmm, I was sure that we should show New columns, which will be added in master code but not hidden by the user manually

I completely agree, I guess I would then store "hidden" columns.

src/common/user-store.ts
```diff
@@ -28,6 +28,7 @@ export interface UserPreferences {
   downloadBinariesPath?: string;
   kubectlBinariesPath?: string;
   openAtLogin?: boolean;
+  tablesConfiguration?: Record<string, string[]>
```
Nokel81:
```suggestion
  hiddenTableColumns?: Record<string, string[]>
```

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -90,7 +97,7 @@ interface ItemListLayoutUserSettings {
 @observer
 export class ItemListLayout extends React.Component<ItemListLayoutProps> {
   static defaultProps = defaultProps as object;
-
+  @observable columnProps = new Set<string>();
```
Nokel81: I think a better name is in order here, since you aren't actually using this data as `props` anywhere.
```suggestion
  @observable hiddenColumnNames = new Set<string>();
```

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -111,7 +118,11 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
   }
 
   async componentDidMount() {
-    const { store, dependentStores, isClusterScoped } = this.props;
+    const { store, dependentStores, isClusterScoped, isConfigurable, tableId, renderTableHeader } = this.props;
+    if (isConfigurable && tableId) {
+      this.columnProps = new Set(userStore.preferences.tablesConfiguration[tableId]
+        ?? renderTableHeader.map(cellProps => { return cellProps.className; }));
```
Nokel81: With the above this is simpler since an empty set implies "show all columns".
```suggestion
      this.hiddenColumnNames = new Set(userStore.preferences?.tablesConfiguration?.[tableId]);
```

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -197,6 +211,20 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
     return this.applyFilters(filterItems, allItems);
   }
 
+  updateColumnFilter(checkboxValue: boolean, className: string) {
```
Nokel81:
```suggestion
  updateColumnFilter(checkboxValue: boolean, columnName: string) {
```

src/renderer/components/+workloads-pods/pods.tsx
```diff
@@ -73,6 +73,8 @@ export class Pods extends React.Component<Props> {
       <KubeObjectListLayout
         className="Pods" store={podsStore}
         dependentStores={[volumeClaimStore, eventStore]}
+        tableId = "workloads_pods"
+        isConfigurable ={true}
```
Nokel81:
```suggestion
        isConfigurable
```

src/renderer/components/item-object-list/item-list-layout.tsx
pashevskii: Ok, will fix it tomorrow

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -197,6 +211,20 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
+      this.columnProps.add(className);
+    } else {
+      this.columnProps.delete(className);
+    }
+  }
+
+  showColumn(index: number): boolean {
+    const {isConfigurable, tableId , renderTableHeader} = this.props;
+    if (!isConfigurable || !tableId) return true;
+    return this.columnProps.has(renderTableHeader[index].className);
```
Nokel81: Are we assuming that `isConfigurable && tableId` ==> `renderTableHeader`? I think it would probably be a good idea if you added the following function.

```typescript
get canBeConfigured(): boolean {
  const { isConfigurable, tableId, renderTableHeader } = this.props;

  if (!isConfigurable || !tableId) {
    return false;
  }

  if (!renderTableHeader?.all(({ className }) => className)) {
    logger.warning("[ItemObjectList]: cannot configure an object list without all headers being identifiable");
    return false;
  }

  return true;
}
```

Nokel81 (commented):
I don't quite see what you mean about "hidden" columns. Are they the ones with no `className` in the `renderTableHeader`? If so, I would ignore the warning I put in the `get`-er above and just have them as not configurable. The down the line they can be made configurable rather easily.

Thanks for doing this work though. It is a very good addition.

pashevskii:
> I don't quite see what you mean about "hidden" columns. Are they the ones with no `className` in the `renderTableHeader`? If so, I would ignore the warning I put in the `get`-er above and just have them as not configurable. The down the line they can be made configurable rather easily.
> 
> Thanks for doing this work though. It is a very good addition.

Not exactly, there are some column headers which have no title, for instance [one](https://github.com/lensapp/lens/blob/master/src/renderer/components/%2Bworkloads-pods/pods.tsx#L95), [two](https://github.com/lensapp/lens/blob/master/src/renderer/components/%2Bworkloads-replicasets/replicasets.tsx#L61) ..
Mostly, these columns are for showing status icon near the name of an artifact. I render these columns as <column.className > but not sure that it is a good idea. I think we should work with them as a part of the previous (or next) column and show/hide both columns by clicking by the leader column (Name for Pods), do not show it as a separate checkbox
image

As an option, we could add some property for those columns like `partOf="<masterColumn.className>"` in TableCellProps and manage the visibility of the cells checking the masterColumn visibility state

Nokel81:
I would say a combination of `showWithOther` and `showWithId` (or some other names, I don't particularly like these) would work well. That way the columns don't have to be in a specific order. For instance in [helm-charts](https://github.com/lensapp/lens/blob/master/src/renderer/components/%2Bapps-helm-charts/helm-charts.tsx#L79) the nameless column is first.

I do think that if the above option is not present then we just default to always showing it.

pashevskii:
> I would say a combination of `showWithOther` and `showWithId` (or some other names, I don't particularly like these) would work well. That way the columns don't have to be in a specific order. For instance in [helm-charts](https://github.com/lensapp/lens/blob/master/src/renderer/components/%2Bapps-helm-charts/helm-charts.tsx#L79) the nameless column is first.
> 
> I do think that if the above option is not present then we just default to always showing it.

Added showWithCollumn property in [second commit](https://github.com/lensapp/lens/pull/1532/commits/455d2d19e33529e48516435b59e40f09cd097da6)

src/renderer/components/table/table-cell.tsx
```diff
@@ -15,6 +15,7 @@ export interface TableCellProps extends React.DOMAttributes<HTMLDivElement> {
   isChecked?: boolean; // mark checkbox as checked or not
   renderBoolean?: boolean; // show "true" or "false" for all of the children elements are "typeof boolean"
   sortBy?: TableSortBy; // column name, must be same as key in sortable object <Table sortable={}/>
+  showWithColumn?: string //visibility state of the column depends on another one state
```
Nokel81: This should have a comment saying that it should be the same as the `className` of the other column.

Nokel81 (dismissed):
Looks to me

src/renderer/components/table/table-cell.tsx
pashevskii: Extended this comment, resolved conflicts

Nokel81:
@aleksfront If you have time would you be able to take a look at this as well?

pashevskii:
@Nokel81 could you please to approve again (all changes are in the second commit, just removed unused vars)

Nokel81 (dismissed):

pashevskii:
@Nokel81 Probably I should apply the feature for all tables within the PR, what do you think about it?

Nokel81:
I would suggest to make a follow up PR

pashevskii:
> I would suggest to make a follow up PR

Ok, thanks

src/renderer/components/table/table-cell.tsx
```diff
@@ -15,6 +15,7 @@ export interface TableCellProps extends React.DOMAttributes<HTMLDivElement> {
   isChecked?: boolean; // mark checkbox as checked or not
   renderBoolean?: boolean; // show "true" or "false" for all of the children elements are "typeof boolean"
   sortBy?: TableSortBy; // column name, must be same as key in sortable object <Table sortable={}/>
+  showWithColumn?: string // className of another column, if it is not empty the current column is not shown in the filter menu, visibility of this one is the same as a specified column, applicable to headers only
```
aleksfront, line 18: I am agaisnt the decision of adding such a property. It adds complexity and overall show/hide TableCell behavior becomes not so straightforward for devs and consumers.

There might be columns which shouldn't be hidden, like actions menu and the one with checkboxes. So, instead we must provide a property to indicate such cols and ignore them within `showColumn()` function.

Those "untouchable" columns can be displayed like disabled ones in Column list without any `pointer-events` allowed. Or maybe it's better just to hide them completely.

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -216,6 +231,38 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
+
+  showColumn(index: number): boolean {
+    const {renderTableHeader} = this.props;
+
+    if (!this.canBeConfigured) return true;
+
+    return !this.hiddenColumnNames.has(renderTableHeader[index].showWithColumn ?? renderTableHeader[index].className);
+  }
+
+  @computed get canBeConfigured(): boolean {
```
aleksfront: It shouldn't be marked as `@computed` since getter doesn't use any observable, but usual props.

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -216,6 +231,38 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
+    const { isConfigurable, tableId, renderTableHeader } = this.props;
+
+    if (!isConfigurable || !tableId) {
+      return false;
+    }
+
+    if (!renderTableHeader?.every(({ className }) => className)) {
```
aleksfront, line 257: We shouldn't rely on class names because we might have a lack of some and we don't have any specific naming convention for developers. So why not introduce required `name` prop for `TableCell` and rely on it?

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -462,6 +514,27 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
+    const { renderTableHeader} = this.props;
+
+    return (
+      <MenuActions
+        toolbar = {false}
+        autoCloseOnSelect = {false}
+        className={cssNames("KubeObjectMenu")}
+      >
+        {renderTableHeader.map((cellProps, index) => (
+          !cellProps.showWithColumn &&
```
aleksfront, line 514: Please look if we can avoid using `showWithColumn` prop.

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -112,7 +120,10 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
   }
 
   async componentDidMount() {
-    const { store, dependentStores, isClusterScoped } = this.props;
+    const { store, dependentStores, isClusterScoped, tableId } = this.props;
+
+    if (this.canBeConfigured) this.hiddenColumnNames = new Set(userStore.preferences?.hiddenTableColumns?.[tableId]);
```
aleksfront, line 125: Is there any special reason for using intermediate observable `this.hiddenColumnNames`? Why not use store values directly (since they're observables too)? See how we did this in `Preferences` https://github.com/lensapp/lens/blob/master/src/renderer/components/+preferences/preferences.tsx#L181 for example/

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -130,9 +141,13 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
 
   componentWillUnmount() {
     this.isUnmounting = true;
-    const { store, isSelectable } = this.props;
+    const { store, isSelectable, tableId } = this.props;
 
     if (isSelectable) store.resetSelection();
+
+    if (this.canBeConfigured) {
+      userStore.preferences.hiddenTableColumns[tableId] = Array.from(this.hiddenColumnNames);
```
aleksfront: 1. We should save new values on every checkbox click, not only in `componentWillUnmount()`. User can reload page, or move away without triggering `componentWillUnmount()` causing old `hiddenTableColumns` values stays untouched.

2. Please check if it's possible to get rid of intermediate property `hiddenColumnNames` and use store directly.

src/renderer/components/+workloads-pods/pods.tsx
```diff
@@ -92,7 +94,7 @@ export class Pods extends React.Component<Props> {
         renderHeaderTitle={<Trans>Pods</Trans>}
         renderTableHeader={[
           { title: <Trans>Name</Trans>, className: "name", sortBy: sortBy.name },
-          { className: "warning" },
+          { className: "warning", showWithColumn: "name" },
```
aleksfront, line 99: Why it must be shown only with name? I am agaisnt the decision of adding such a property. It adds complexity and overall show/hide TableCell behavior becomes not so straightforward for devs and consumers. Imho, all columns must be showed/hidden by hitting according checkbox only.

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -216,6 +231,38 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
     return this.applyFilters(filterItems, allItems);
   }
 
+  updateColumnFilter(checkboxValue: boolean, columnName: string) {
```
aleksfront, line 230: Ideally, it should save new values to store directly.

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -216,6 +231,38 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
+    if (checkboxValue){
+      this.hiddenColumnNames.delete(columnName);
+    } else {
+      this.hiddenColumnNames.add(columnName);
+    }
+  }
+
+  showColumn(index: number): boolean {
```
aleksfront: Nothing critical and maybe it's just me, but maybe `showColumn` method could have another name since it feels like it will render something without returning any value. @Nokel81 wdyt?

aleksfront (changes requested):

src/renderer/components/table/table-cell.tsx
Nokel81: This is an optional property so I don't see how "some columns shouldn't be hidden" applies. 

I would say that this actually makes the action menu simpler for the users as we don't have to come up with some name if we want to hide a title-less column.

src/renderer/components/item-object-list/item-list-layout.tsx
Nokel81: It cannot be `required` since that would require changing every table. I think that a new prop would probably be better but I didn't think I could justify the extra effort on the part of the developers.

src/renderer/components/item-object-list/item-list-layout.tsx
Nokel81: Line 181 is not a good comparison. We want set like behaviour but sets are not JSON serialization by default and I don't think this PR should have that as well. We can do that work later.

src/renderer/components/item-object-list/item-list-layout.tsx
Nokel81: `columnIsVisible`?

pashevskii: fixed

src/renderer/components/item-object-list/item-list-layout.tsx
pashevskii: fixed

src/renderer/components/table/table-cell.tsx
pashevskii: @aleksfront Partially agreed with you, but how about the "icon" header in Apps->Charts, IMHO the column is just a visualization of the "Name" column and it should be shown as part of the name or the user should have the ability to manage this.  The same situation is for deamonsets, "warning" title-less column looks a part of the "Pods" column
Also, sometimes the rendering table looks glitchy (for instance: when there is the first column is title-less and has no data):
image
(I agree with the fact that this set of columns has no sense, just to highlight the issue)
Probably we could hide the column (or set width = 0) in cases when this one has no shown objects and do not check it in showColumn(), from my pov it will be much better instead of showing empty column without possibility to hide it

src/renderer/components/item-object-list/item-list-layout.tsx
pashevskii: fixed

src/renderer/components/item-object-list/item-list-layout.tsx
pashevskii: I thought about it but I've decided to avoid global changes in root objects. The current number of tables is not too huge and I can implement new prop for all tables and set the prop of every column.

aleksfront:
@pashevskii There's 4 options we should do before merge this PR:
1. Update with latest `master` changes and check if column filters still works (we've added tooltips for pod table cells)
2. Save column filters right after checkbox click. Currently, if you're refresh the page - you will lost your changes since storage is updated in `componentWillUnmount()`
3. Get rid of `Warning: React does not recognize the `showWithColumn` prop on a DOM element`. Somehow this new prop is passed to naked DOM element causing the error.
Screenshot 2020-12-23 at 12 31 31
4. Expand menu clickable area to the whole item (probably put `onClick` event on `MenuItem` itself instead of `Checkbox`).
Dec-23-2020 12-34-00

Nokel81:
I personally think that number 4 is the best option.

pashevskii:
@aleksfront , @Nokel81 
1. Have done, it looks working Ok
2. Fixed
3. Should be fixed
4. Partially fixed. How about extending the checkbox area? (I've added it in the last commit) menuItem contains a checkbox and it inherits onClick evens, as a result, we will get "double-click" on clicking on the checkbox, the first one is onClick of MenuItem, the second one of Checkbox, which was inherited from MenuItem. More info is [here](https://stackoverflow.com/questions/50819162/why-is-my-function-being-called-twice-in-react)

aleksfront (dismissed):
LGTM, clicking behavior we could improve in upcoming PRs.

aleksfront:
@Nokel81 Please take another look 👀

pashevskii:
@aleksfront I rebased  on the master and resolved conflicts after "Remove lingui" commit, please approve it again

src/renderer/components/+workloads-pods/pods.tsx
Nokel81: This this still an issue for you Aleks? I think this also ties into not relying on the class names.

src/renderer/components/item-object-list/item-list-layout.tsx
Nokel81: You could reuse `title` and add a new optional property `hideTitle?: boolean`. That way columns similar to `className: "warning"` could still have it not rendered in the table (but do render it in the menu)

src/renderer/components/table/table-cell.tsx
Nokel81: Regarding the checkbox as the title, is it possible to force it being empty by setting its title to `" "`? Or maybe even `{ title: "Selected", className: "hidden", ... }`?

aleksfront:
@Nokel81 @ixrock Please take a look. I think this PR is ready to be merged.

aleksfront (approved):

Nokel81:
@aleksfront I left some review comments last Thursday (mostly as responses to your comments). By approving this PR are you saying that those are no longer issues?

src/renderer/components/+workloads-pods/pods.tsx
aleksfront: Not an issue anymore. Let's stick with current behavior.

Nokel81 (approved):

aleksfront:
> @aleksfront I left some review comments last Thursday (mostly as responses to your comments). By approving this PR are you saying that those are no longer issues?

Yes, I agreed to keep `showWithColumns` feature and rely on class names for now.

PR #1537, opened 2020-11-26
VirtualList component refactoring

aleksfront:
This PR cleans up `VirtualList` component, it's types and child `Row` component. Also provides more simple implementation of `getTableRow` method across different places.

Now `item` parameter passing from VirtualList directly to `getTableRow` which allows to avoid unnecessary `find` or `findIndex` searching for every table row.


panuhorsmalahti (approved):

src/renderer/components/virtual-list/virtual-list.tsx
```diff
@@ -126,8 +126,7 @@ interface RowProps extends ListChildComponentProps {
 const Row = observer((props: RowProps) => {
   const { index, style, data } = props;
   const { items, getRow } = data;
-  const item = items[index];
-  const uid = typeof item == "string" ? index : items[index].getId();
+  const uid = items[index].getId ? items[index].getId() : index;
```
ixrock: How you can be sure you're that you are checking `getId` from object when you defined items as `any: []`?
```suggestion
  const uid = items[index]?.getId() ?? index;
```

src/renderer/components/virtual-list/virtual-list.tsx
```diff
@@ -13,8 +13,8 @@ import debounce from "lodash/debounce";
 import isEqual from "lodash/isEqual";
 import ResizeSensor from "css-element-queries/src/ResizeSensor";
 
-interface Props<T extends ItemObject = any> {
-  items: T[];
+interface Props {
+  items: any[];
```
ixrock, line 14: Maybe this also works? Or even without `| any`

```suggestion
interface Props<T = ItemObject | any> {
  items: T[];
```

ixrock (approved):

src/renderer/components/virtual-list/virtual-list.tsx
Nokel81: If `getId` is optionally present, shouldn't it be the following?
```suggestion
  const uid = items[index]?.getId?.() ?? index;
```

panuhorsmalahti: The original code is (also) correct. getId will return undefined if it's not a property of the object, which is falsy, so index would be used in that case. I don't see how null coalescing operator is more readable than the conditional operator. (I think they're equivalent readability wise).

jakolehm (approved):

aleksfront:
@panuhorsmalahti @ixrock Made more generic changes to `Table` and `VirtualList` components, please re-review this time.

jakolehm (changes requested):
I think we should rethink this. It causes a breaking change to `Table` component but actually it might be better to create a new component called `VirtualTable` that is designed for just virtual use cases and mark `virtual` property in `Table` as deprecated.

aleksfront:
Closing this in favor of separating `Table` and `VirtualTable` components in other PR.

PR #1539, opened 2020-11-26
Disable Uninstall and Enable/Disable buttons while uninstalling

panuhorsmalahti:
* Disable Uninstall and Enable/Disable buttons while uninstalling
* Add "spinner" (e.g. waiting animation) to Uninstall button while install is in progress
* Add Notification for uninstall

### DIFF
--- a/src/renderer/components/+extensions/extensions.tsx
+++ b/src/renderer/components/+extensions/extensions.tsx
@@ -1,8 +1,9 @@
 import { t, Trans } from "@lingui/macro";
 import { remote, shell } from "electron";
 import fse from "fs-extra";
-import { computed, observable } from "mobx";
-import { observer } from "mobx-react";
+import { map, omit } from "lodash";
+import { computed, observable, ObservableMap, reaction } from "mobx";
+import { disposeOnUnmount, observer } from "mobx-react";
 import os from "os";
 import path from "path";
 import React from "react";
@@ -38,6 +39,12 @@ interface InstallRequestValidated extends InstallRequestPreloaded {
   tempFile: string; // temp system path to packed extension for unpacking
 }
 
+interface ExtensionState {
+  displayName: string;
+  // Possible states the extension can be
+  state: "uninstalling";
+}
+
 @observer
 export class Extensions extends React.Component {
   private supportedFormats = [".tar", ".tgz"];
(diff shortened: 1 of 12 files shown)


src/renderer/components/+extensions/extensions.tsx
```diff
@@ -38,8 +39,28 @@ interface InstallRequestValidated extends InstallRequestPreloaded {
...
+  };
+}
+
+/**
+ * Extensions that were removed from extensions but are still in "uninstalling" state
+ */
+const getRemovedUninstalling = (extensions: InstalledExtension[], extensionState: State["extensionState"]) => 
+  Object.entries(extensionState).filter(([id, extensionState]) => 
+    extensionState.state === "uninstalling" && !extensions.find(extension => extension.id === id)
+  ).map(([id, extension]) => ({ ...extension, id }));
```
Nokel81: top level items should be functions preferably (not bound anonymous functions).

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -278,14 +322,30 @@ export class Extensions extends React.Component {
...
+      this.setState(state => ({
+        ...state,
+        extensionState: {
+          ...state.extensionState,
+          [extension.id]: {
+            state: "uninstalling",
+            displayName
+          }
+        }
+      }));
```
Nokel81: We use `mobx` instead of `setState`. I think that this easily could be an `observable.Map<string, {...}>` instead.

Nokel81 (changes requested):

src/renderer/components/+extensions/extensions.tsx
panuhorsmalahti: I think local state (e.g. setState) is fine for component local state that should not be accessed outside. Our application is already using local state in various components.

Nokel81: mobx manages local state. Those are mostly older components.

src/renderer/components/+extensions/extensions.tsx
panuhorsmalahti: Changed

src/renderer/components/+extensions/extensions.tsx
panuhorsmalahti: Well, MobX _can_ manage local state, but that's a design choice. I'll convert it to MobX then.

panuhorsmalahti: Converted

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -38,6 +39,21 @@ interface InstallRequestValidated extends InstallRequestPreloaded {
...
+}
+
+/**
+ * Extensions that were removed from extensions but are still in "uninstalling" state
+ */
+function getRemovedUninstalling(extensions: InstalledExtension[], extensionState: ObservableMap<string, ExtensionState>) {
+  return Array.from(extensionState.entries()).filter(([id, extension]) => 
+    extension.state === "uninstalling" && !extensions.find(extension => extension.id === id)
+  ).map(([id, extension]) => ({ ...extension, id }));
+}
```
Nokel81: I think this would be better as a `@computed get removedUninstalling() { ... }` on the `Extensions` class bellow.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -49,17 +65,38 @@ export class Extensions extends React.Component {
...
   @computed get extensions() {
     const searchText = this.search.toLowerCase();
     return Array.from(extensionLoader.userExtensions.values()).filter(ext => {
       const { name, description } = ext.manifest;
       return [
         name.toLowerCase().includes(searchText),
-        description.toLowerCase().includes(searchText),
-      ].some(v => v);
+        description?.toLowerCase().includes(searchText),
+      ].some(value => value);
```
Nokel81, line 99: Would using `_.filter` be more readable?

Nokel81 (approved):
Except for a few minor things. This looks good.

src/renderer/components/+extensions/extensions.tsx
panuhorsmalahti: Maybe, but not going to touch unrelated code much here, just wanted to fix one char variable naming..

src/renderer/components/+extensions/extensions.tsx
panuhorsmalahti: Converted

jakolehm (approved):

PR #1546, opened 2020-11-27
Add extension-api facades to cluster & workspace stores

jakolehm:
Both Cluster and Workspace stores currently leak internals to extension-api. This PR adds a facade store for both of them and exposes only usable apis. 

This also allows us to work more freely on store internals (we only expose apis explicitly). Good example of accidental leaks is #1453 .

### DIFF
--- a/src/extensions/stores/cluster-store.ts
+++ b/src/extensions/stores/cluster-store.ts
@@ -2,6 +2,7 @@ import { clusterStore as internalClusterStore, ClusterId } from "../../common/cl
 import type { ClusterModel } from "../../common/cluster-store";
 import { Cluster } from "../../main/cluster";
 import { Singleton } from "../core-api/utils";
+import { ObservableMap } from "mobx";
 
 export { Cluster } from "../../main/cluster";
 export type { ClusterModel, ClusterId } from "../../common/cluster-store";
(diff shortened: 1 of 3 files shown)


nevalla:
I like the idea. With this we can refactor internal store without breaking the Extensions API and we don't need to expose all internal methods to extensions.

src/extensions/stores/cluster-store.ts
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Store for all added clusters
+ */
+export class ClusterStore extends Singleton {
+
+  /**
+   * Active cluster id
+   */
+  activeClusterId = internalClusterStore.activeCluster;
```
Nokel81: have you tested that this works? When I was thinking about it yesterday I wasn't convinced that this would work and that we would have to do something like the following:

```suggestion
  get activeClusterId() {
    return internalClusterStore.activeCluster;
  }
  
  set activeClusterId(newId: string) {
      internalClusterStore.activeCluster = newId;
  }
```

src/extensions/stores/cluster-store.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  /**
+   * Map of all clusters
+   */
+  clusters = internalClusterStore.clusters;
```
Nokel81: I am pretty sure that this would definitely break if either side were to for instance do a `= []` instead of `.length = 0`.

jakolehm: Right, maybe needs to be `get`?

src/extensions/stores/cluster-store.ts
jakolehm: get/set is better.

src/extensions/stores/workspace-store.ts
```diff
@@ -0,0 +1,109 @@
...
+   * Default workspace id, this workspace is always present
+   */
+  static readonly defaultId: WorkspaceId = InternalWorkspaceStore.defaultId;
+
+  /**
+   * Currently active workspace id
+   */
+  get currentWorkspaceId(): string {
+    return internalWorkspaceStore.currentWorkspaceId;
+  }
```
Nokel81, line 22: Do we not support extensions setting this?

Nokel81 (approved):
Looks good, definitely a better way of making sure that we don't leak APIs to extensions.

src/extensions/stores/workspace-store.ts
jakolehm: Good catch, added.

PR #1551, opened 2020-11-27
Disable extension install button while installing

panuhorsmalahti:
- Disable Install button and Install input field while installing
- Fix install notification to be made when the install is actually complete

Screenshot 2020-11-27 at 14 26 49


jakolehm (approved):
LGTM. Much better!

PR #1572, opened 2020-11-30
Enable installed extensions by default

### LINKED ISSUE #1553: Enable installed extension automatically

panuhorsmalahti:
Fixes #1553

### DIFF
--- a/src/extensions/extension-loader.ts
+++ b/src/extensions/extension-loader.ts
@@ -1,5 +1,6 @@
 import { app, ipcRenderer, remote } from "electron";
 import { EventEmitter } from "events";
+import { isEqual } from "lodash";
 import { action, computed, observable, reaction, toJS, when } from "mobx";
 import path from "path";
 import { getHostedCluster } from "../common/cluster-store";
@@ -25,7 +26,12 @@ const logModule = "[EXTENSIONS-LOADER]";
 export class ExtensionLoader {
   protected extensions = observable.map<LensExtensionId, InstalledExtension>();
   protected instances = observable.map<LensExtensionId, LensExtension>();
-  protected readonly requestExtensionsChannel = "extensions:loaded";
+
+  // IPC channel to broadcast changes to extensions from main
+  protected static readonly extensionsMainChannel = "extensions:main";
+
+  // IPC channel to broadcast changes to extensions from renderer
+  protected static readonly extensionsRendererChannel = "extensions:renderer";
 
   // emits event "remove" of type LensExtension when the extension is removed
   private events = new EventEmitter();
@@ -95,28 +101,27 @@ export class ExtensionLoader {
     this.loadOnMain();
     this.broadcastExtensions();
 
-    reaction(() => this.extensions.toJS(), () => {
+    reaction(() => this.toJSON(), () => {
       this.broadcastExtensions();
     });
 
-    handleRequest(this.requestExtensionsChannel, () => {
+    handleRequest(ExtensionLoader.extensionsMainChannel, () => {
       return Array.from(this.toJSON());
+    });
+
+    subscribeToBroadcast(ExtensionLoader.extensionsRendererChannel, (_event, extensions: [LensExtensionId, InstalledExtension][]) => {
+      this.syncExtensions(extensions);
     });
   }
 
   protected async initRenderer() {
     const extensionListHandler = (extensions: [LensExtensionId, InstalledExtension][]) => {
       this.isLoaded = true;
+      this.syncExtensions(extensions);
+
       const receivedExtensionIds = extensions.map(([lensExtensionId]) => lensExtensionId);
-
-      // Add new extensions
-      extensions.forEach(([extId, ext]) => {
-        if (!this.extensions.has(extId)) {
-          this.extensions.set(extId, ext);
-        }
-      });
-
-      // Remove deleted extensions
+          
+      // Remove deleted extensions in renderer side only
       this.extensions.forEach((_, lensExtensionId) => {
         if (!receivedExtensionIds.includes(lensExtensionId)) {
           this.removeExtension(lensExtensionId);
@@ -124,14 +129,26 @@ export class ExtensionLoader {
       });
     };
 
-    requestMain(this.requestExtensionsChannel).then(extensionListHandler);
-    subscribeToBroadcast(this.requestExtensionsChannel, (event, extensions: [LensExtensionId, InstalledExtension][]) => {
+    reaction(() => this.toJSON(), () => {
+      this.broadcastExtensions(false);
+    });
+
+    requestMain(ExtensionLoader.extensionsMainChannel).then(extensionListHandler);
+    subscribeToBroadcast(ExtensionLoader.extensionsMainChannel, (_event, extensions: [LensExtensionId, InstalledExtension][]) => {
       extensionListHandler(extensions);
     });
   }
 
+  syncExtensions(extensions: [LensExtensionId, InstalledExtension][]) {
+    extensions.forEach(([lensExtensionId, extension]) => {
+      if (!isEqual(this.extensions.get(lensExtensionId), extension)) {
+        this.extensions.set(lensExtensionId, extension);
+      }
+    });
+  }
+
   loadOnMain() {
-    logger.info(`${logModule}: load on main`);
+    logger.debug(`${logModule}: load on main`);
     this.autoInitExtensions(async (extension: LensMainExtension) => {
       // Each .add returns a function to remove the item
       const removeItems = [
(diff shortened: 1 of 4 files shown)


src/renderer/components/+extensions/extensions.tsx
```diff
@@ -91,11 +91,23 @@ export class Extensions extends React.Component {
...
+
           Notifications.ok(
             <p>Extension <b>{displayName}</b> successfully installed!</p>
           );
           this.extensionState.delete(id);
           this.installPath = "";
+
+          // Enable installed extensions by default.
+          // Timeout is required since we're modifying the observable inside reaction.
+          setTimeout(() => {
```
ixrock: > // Timeout is required since we're modifying the observable inside reaction.

Sounds weird.. Observables could/should be modified from anywhere.. 
Maybe you need `runInAction` or mark a method with `@action` when updating something?

panuhorsmalahti: It's definitely weird, and it's a bit difficult to track what's going on with all the various reactions and computed extensions and store interaction.

I tried `runInAction` and  `@action` but those didn't work.

panuhorsmalahti: Digging deeper into this one, it seems that the extensions-store will disable it if done sync in the reaction. I'm trying to find a better solution.

panuhorsmalahti: I removed the timeout but also had to remove the reaction to update state from store updates in extensions-store. I didn't find a reason for that reaction, and enabled state seems to work fine with it removed.

src/extensions/extensions-store.ts
```diff
@@ -43,16 +43,6 @@ export class ExtensionsStore extends BaseStore<LensExtensionsStoreModel> {
     await extensionLoader.whenLoaded;
     await this.whenLoaded;
 
-    // apply state on changes from store
```
nevalla, line 48: Why are these lines removed? Removing this reaction seems to break enabling/disabling extensions from UI

panuhorsmalahti: Removing these lines makes it possible to set isEnabled = true in the reaction. I didn't see any bugs from this change at least in my testing..

It would be nice for someone to explain the logic of why we need these lines.

panuhorsmalahti: Okay, this does break the disabling to an extent. I'll change the PR yet again.

src/renderer/components/+extensions/extensions.tsx
panuhorsmalahti: This should be fixed now.

src/extensions/extensions-store.ts
panuhorsmalahti: PR fixed, enable and disable should work now.

src/extensions/extension-loader.ts
```diff
@@ -25,7 +26,12 @@ const logModule = "[EXTENSIONS-LOADER]";
 export class ExtensionLoader {
   protected extensions = observable.map<LensExtensionId, InstalledExtension>();
   protected instances = observable.map<LensExtensionId, LensExtension>();
-  protected readonly requestExtensionsChannel = "extensions:loaded";
+
+  // IPC channel to broadcast changes to extensions from main
+  protected readonly extensionsMainChannel = "extensions:main";
```
Nokel81:
```suggestion
  protected static readonly extensionsMainChannel = "extensions:main";
```

src/extensions/extension-loader.ts
```diff
@@ -25,7 +26,12 @@ const logModule = "[EXTENSIONS-LOADER]";
...
+
+  // IPC channel to broadcast changes to extensions from renderer
+  protected readonly extensionsRendererChannel = "extensions:renderer";
```
Nokel81:
```suggestion
  protected static readonly extensionsRendererChannel = "extensions:renderer";
```

src/extensions/extension-loader.ts
```diff
@@ -95,43 +101,53 @@ export class ExtensionLoader {
     this.loadOnMain();
     this.broadcastExtensions();
 
-    reaction(() => this.extensions.toJS(), () => {
+    reaction(() => this.toJSON(), () => {
       this.broadcastExtensions();
     });
 
-    handleRequest(this.requestExtensionsChannel, () => {
+    handleRequest(this.extensionsMainChannel, () => {
```
Nokel81:
```suggestion
    handleRequest(ExtensionLoader.extensionsMainChannel, () => {
```

src/extensions/extension-loader.ts
```diff
@@ -95,43 +101,53 @@ export class ExtensionLoader {
...
       return Array.from(this.toJSON());
     });
+
+    subscribeToBroadcast(this.extensionsRendererChannel, (_event, extensions: [LensExtensionId, InstalledExtension][]) => {
```
Nokel81:
```suggestion
    subscribeToBroadcast(ExtensionLoader.extensionsRendererChannel, (_event, extensions: [LensExtensionId, InstalledExtension][]) => {
```

src/extensions/extension-loader.ts
```diff
@@ -95,43 +101,53 @@ export class ExtensionLoader {
...
       });
     };
 
-    requestMain(this.requestExtensionsChannel).then(extensionListHandler);
-    subscribeToBroadcast(this.requestExtensionsChannel, (event, extensions: [LensExtensionId, InstalledExtension][]) => {
+    reaction(() => this.toJSON(), () => {
+      this.broadcastExtensions(false);
+    });
+
+    requestMain(this.extensionsMainChannel).then(extensionListHandler);
```
Nokel81:
```suggestion
    requestMain(ExtensionLoader.extensionsMainChannel).then(extensionListHandler);
```

src/extensions/extension-loader.ts
```diff
@@ -95,43 +101,53 @@ export class ExtensionLoader {
...
+    subscribeToBroadcast(this.extensionsMainChannel, (_event, extensions: [LensExtensionId, InstalledExtension][]) => {
```
Nokel81:
```suggestion
    subscribeToBroadcast(ExtensionLoader.extensionsMainChannel, (_event, extensions: [LensExtensionId, InstalledExtension][]) => {
```

src/extensions/extension-loader.ts
```diff
@@ -151,7 +167,6 @@ export class ExtensionLoader {
   }
 
   loadOnClusterManagerRenderer() {
-    logger.info(`${logModule}: load on main renderer (cluster manager)`);
```
Nokel81, line 154: Please don't remove these, I think that it is valuable logging information. (If you must, then change it to debug)

src/extensions/extension-loader.ts
```diff
@@ -257,8 +271,8 @@ export class ExtensionLoader {
     });
   }
 
-  broadcastExtensions() {
-    broadcastMessage(this.requestExtensionsChannel, Array.from(this.toJSON()));
+  broadcastExtensions(main = true) {
+    broadcastMessage(main ? this.extensionsMainChannel : this.extensionsRendererChannel, Array.from(this.toJSON()));
```
Nokel81:
```suggestion
    broadcastMessage(main ? ExtensionLoader.extensionsMainChannel : ExtensionLoader.extensionsRendererChannel, Array.from(this.toJSON()));
```

src/extensions/extensions-store.ts
```diff
@@ -61,7 +51,10 @@ export class ExtensionsStore extends BaseStore<LensExtensionsStoreModel> {
 
   isEnabled(extId: LensExtensionId) {
     const state = this.state.get(extId);
-    return state && state.enabled; // by default false
+
+    // By default false, so that copied extensions are disabled by default.
+    // If user installs the extension from the UI, the Extensions component will specifically enable it.
+    return Boolean(state && state.enabled);
```
Nokel81:
```suggestion
    return !!state?.enabled;
```

Nokel81 (approved):

src/extensions/extensions-store.ts
panuhorsmalahti: I think `Boolean` is more readable than using `!!`. I think `?.` / `&&` are equivalent in readability.

src/extensions/extension-loader.ts
panuhorsmalahti: Added them back as .debug

src/extensions/extension-loader.ts
panuhorsmalahti: Changed

src/extensions/extension-loader.ts
panuhorsmalahti: Changed

src/extensions/extensions-store.ts
Nokel81: I a sort of moot on `Boolean` since it doesn't seem to be used much. I quite prefer `?.` over `&&` as the later can get very long quite quickly.

panuhorsmalahti: I converted to `?.` but kept `Boolean` for now. It's true some don't know it, but I think it's pretty standard JavaScript, and every project is an opportunity to make people familiar with it. Quick look at Google seems to favor `Boolean` over `!!`.

jakolehm (approved):
LGTM

PR #1573, opened 2020-11-30
Hide disabled workspaces/clusters

### LINKED ISSUE #1556: Workspace added via extension api not handled properly on workspaces page
image

The lens-minikube extension creates the minikube workspace. When the extension is disabled (assuming https://github.com/lensapp/lens-extension-samples/pull/12 has been merged), the workspace is disabled. From the Lens UI, as shown above, the disabled minikube workspace does not appear in the Workspaces popup list in the bottom left corner (expected), but it is visible in the list on the Workspaces page (unexpected).

Also, even when the minikube workspace is enabled, when viewed in the Workspaces page list, there are no edit and delete icons. If this is intended then there is no way to delete the workspace unless the extension does it (when the extension is disabled perhaps? Or when uninstalled?)

jakolehm:
- fix: workspace/cluster was not enabled automatically on save
- fix: workspace listing listed also disabled workspaces
- fix: workspace-store setInterval leaked during tests, refactored it to use `requestMain()` in `load()`
- fix: cluster enabled not being synced from main to renderer
- fix: cluster with ownerRef removal via context menu

Fixes #1556

### DIFF
--- a/src/common/__tests__/cluster-store.test.ts
+++ b/src/common/__tests__/cluster-store.test.ts
@@ -48,6 +48,7 @@ describe("empty config", () => {
       expect(storedCluster.id).toBe("foo");
       expect(storedCluster.preferences.terminalCWD).toBe("/tmp");
       expect(storedCluster.preferences.icon).toBe("data:image/jpeg;base64, iVBORw0KGgoAAAANSUhEUgAAA1wAAAKoCAYAAABjkf5");
+      expect(storedCluster.enabled).toBe(true);
     });
 
     it("adds cluster to default workspace", () => {
@@ -170,7 +171,8 @@ describe("config with existing clusters", () => {
               kubeConfig: "foo",
               contextName: "foo",
               preferences: { terminalCWD: "/foo" },
-              workspace: "foo"
+              workspace: "foo",
+              ownerRef: "foo"
             },
           ]
         })
@@ -207,6 +209,12 @@ describe("config with existing clusters", () => {
     expect(storedClusters[1].id).toBe("cluster2");
     expect(storedClusters[1].preferences.terminalCWD).toBe("/foo2");
     expect(storedClusters[2].id).toBe("cluster3");
+  });
+
+  it("marks owned cluster disabled by default", () => {
+    const storedClusters = clusterStore.clustersList;
+    expect(storedClusters[0].enabled).toBe(true);
+    expect(storedClusters[2].enabled).toBe(false);
   });
 });
 
--- a/src/common/__tests__/cluster-store.test.ts
+++ b/src/common/__tests__/cluster-store.test.ts
@@ -7,6 +7,20 @@ import { workspaceStore } from "../workspace-store";
 
 const testDataIcon = fs.readFileSync("test-data/cluster-store-migration-icon.png");
 
+jest.mock("electron", () => {
+  return {
+    app: {
+      getVersion: () => "99.99.99",
+      getPath: () => "tmp",
+      getLocale: () => "en"
+    },
+    ipcMain: {
+      handle: jest.fn(),
+      on: jest.fn()
+    }
+  };
+});
+
 let clusterStore: ClusterStore;
 
 describe("empty config", () => {
--- a/src/common/cluster-store.ts
+++ b/src/common/cluster-store.ts
@@ -205,7 +205,9 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
     if (!(model instanceof Cluster)) {
       cluster = new Cluster(model);
     }
-    cluster.enabled = true;
+    if (!cluster.isManaged) {
+      cluster.enabled = true;
+    }
     this.clusters.set(model.id, cluster);
     return cluster;
   }
--- a/src/common/cluster-store.ts
+++ b/src/common/cluster-store.ts
@@ -40,13 +40,30 @@ export interface ClusterStoreModel {
 export type ClusterId = string;
 
 export interface ClusterModel {
+  /** Unique id for a cluster */
   id: ClusterId;
+
+  /** Path to cluster kubeconfig */
   kubeConfigPath: string;
+
+  /** Workspace id */
   workspace?: WorkspaceId;
+
+  /** User context in kubeconfig  */
   contextName?: string;
+
+  /** Preferences */
   preferences?: ClusterPreferences;
+
+  /** Metadata */
   metadata?: ClusterMetadata;
+
+  /**
+   * If extension sets ownerRef it has to explicitly mark a cluster as enabled during onActive (or when cluster is saved)
+   */
   ownerRef?: string;
+
+  /** List of accessible namespaces */
   accessibleNamespaces?: string[];
 
   /** @deprecated */
@@ -89,7 +106,7 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
   @observable removedClusters = observable.map<ClusterId, Cluster>();
   @observable clusters = observable.map<ClusterId, Cluster>();
 
-  private stateRequestChannel = "cluster:states";
+  private static stateRequestChannel = "cluster:states";
 
   private constructor() {
     super({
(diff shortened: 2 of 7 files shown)


Nokel81:
Currently reviewing... tried to load the `minikube` sample extension but cannot seem to get it to load. Will look into it.

src/common/__tests__/cluster-store.test.ts
```diff
@@ -208,6 +210,12 @@ describe("config with existing clusters", () => {
     expect(storedClusters[1].preferences.terminalCWD).toBe("/foo2");
     expect(storedClusters[2].id).toBe("cluster3");
   });
+
+  it("marks owned cluster disabled by default", () => {
```
jim-docker, line 228: enabled?

src/common/workspace-store.ts
```diff
@@ -77,16 +77,41 @@ export class Workspace implements WorkspaceModel, WorkspaceState {
...
+  async load() {
+    await super.load();
+    type workspaceStateSync = {
+      id: string;
+      state: WorkspaceState;
+    };
+    if (ipcRenderer) {
+      logger.info("[WORKSPACE-STORE] requesting initial state sync");
+      const workspaceStates: workspaceStateSync[] = await requestMain(this.stateRequestChannel);
+      console.log(workspaceStates);
```
aleksfront: Please remove this line.

src/common/workspace-store.ts
```diff
@@ -77,16 +77,41 @@ export class Workspace implements WorkspaceModel, WorkspaceState {
...
     });
+  }
 
-    if (!ipcRenderer) {
-      setInterval(() => {
-        this.pushState();
-      }, 5000);
+  async load() {
+    await super.load();
+    type workspaceStateSync = {
```
ixrock, line 124: Type definition could be non-top level?

src/renderer/components/cluster-manager/clusters-menu.tsx
```diff
@@ -106,7 +106,7 @@ export class ClustersMenu extends React.Component<Props> {
     const { className } = this.props;
     const { newContexts } = userStore;
     const workspace = workspaceStore.getById(workspaceStore.currentWorkspaceId);
-    const clusters = clusterStore.getByWorkspaceId(workspace.id);
+    const clusters = clusterStore.getByWorkspaceId(workspace.id).filter(cluster => cluster.enabled === true);
```
ixrock:
```suggestion
    const clusters = clusterStore.getByWorkspaceId(workspace.id).filter(cluster => cluster.enabled);
```

Nokel81 (dismissed):
This PR does not fix the linked issue. I have tested it with the `mini-kube` and when switching to the created `minikube` workspace there is no icon for the cluster. The issue also mentions being able to remove workspaces added by extensions which is not mentioned in this PR.

I also don't know if the correct solution is to auto-enable clusters and workspaces. That seems to be forcing the extensions' hand. I would say that fixing the rendering would be better (have it greyed out or crossed out would be better). And once the workspace is enabled it should be possible to view the cluster. I was able to use the tray icon to access the cluster but not in the window.

jakolehm:
> This PR does not fix the linked issue. I have tested it with the `mini-kube` and when switching to the created `minikube` workspace there is no icon for the cluster. The issue also mentions being able to remove workspaces added by extensions which is not mentioned in this PR.
> 

Did you test this with https://github.com/lensapp/lens-extension-samples/pull/12 ? 

 If an extension marks a cluster/workspace as owned then those cannot be removed by a user (that's by design). If extension does not mark those as owned then they can be controlled by a user.


> I also don't know if the correct solution is to auto-enable clusters and workspaces. That seems to be forcing the extensions' hand. I would say that fixing the rendering would be better (have it greyed out or crossed out would be better). And once the workspace is enabled it should be possible to view the cluster. I was able to use the tray icon to access the cluster but not in the window.

What is the use case for adding a cluster/workspace and not enabling it? I tried to follow principle-of-least-surprise here... but we can also take this away and force extension to mark a cluster/workspace as enabled after it has been saved.

Nokel81:
I had not tested with that PR, but I have now. It now opens that cluster when I activate the workspace. Though I still think it is strange for the cluster not to show up (even if it were to be grey-ed out and "disabled") that would be better.

jakolehm:
> Though I still think it is strange for the cluster not to show up (even if it were to be grey-ed out and "disabled") that would be better.

The thing is that it _should_ show up because minikube extension enables it. For some reason that change is not applied to store items.. investigating.

jakolehm:
@Nokel81 cluster not showing up was a bug that was revealed by fixing this visibility thing... Enabled state was never transferred to renderer because it was not part of `ClusterState`. Should work now.

Nokel81:
Will test, but I really don't like hiding things that are not "enabled" they should still be shown, but in a "disabled" state (such as at 50% opacity with no clicking or hovering)

src/common/__tests__/cluster-store.test.ts
jakolehm: No, disabled 🙂 If store is loaded from disk clusters with ownerRef will be disabled by default -> extension needs to mark them as enabled.

Nokel81:
After testing, I do now see that the minikube cluster is now visible. However, I really don't think that we should be hiding the non-enabled clusters.

jakolehm:
> Will test, but I really don't like hiding things that are not "enabled" they should still be shown, but in a "disabled" state (such as at 50% opacity with no clicking or hovering)

"I don't like" is out-of-scope of this PR 🙂  Disabled state is hidden for a reason. 

Let me take an example:

An extension wants to sync workspaces (with clusters) from remote source (think a cloud service). To achieve this extension marks workspace (and clusters) as owned (via ownerRef) so that it can take total responsibility of lifecycle of those objects (user cannot add or remove these manually). If extension is down (or disabled) user should not be able to see any of these objects because they might be stale (and out of users control).

If an extension does not want to take total responsibility of these objects then it can just leave ownerRef blank, they will be enabled by default even if extension is not loaded.

jim-docker:
> After testing, I do now see that the minikube cluster is now visible. However, I really don't think that we should be hiding the non-enabled clusters.

If it is owned by an extension then it should be up to the extension when it is enabled/visible? Does enabled here really just mean visible?
I'm concerned about orphaned workspaces and clusters. What should happen when an extension is uninstalled? If the owner is gone they should be enabled/removable?

src/common/cluster-store.ts
```diff
@@ -89,6 +89,8 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
   @observable removedClusters = observable.map<ClusterId, Cluster>();
   @observable clusters = observable.map<ClusterId, Cluster>();
 
+  private stateRequestChannel = "cluster:states";
```
Nokel81:
```suggestion
  private static stateRequestChannel = "cluster:states";
```

src/common/cluster-store.ts
```diff
@@ -102,8 +104,40 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
...
 
+  async load() {
+    await super.load();
+    type clusterStateSync = {
+      id: string;
+      state: ClusterState;
+    };
+    if (ipcRenderer) {
+      logger.info("[CLUSTER-STORE] requesting initial state sync");
+      const clusterStates: clusterStateSync[] = await requestMain(this.stateRequestChannel);
```
Nokel81:
```suggestion
      const clusterStates: clusterStateSync[] = await requestMain(ClusterStore.stateRequestChannel);
```

src/common/cluster-store.ts
```diff
@@ -102,8 +104,40 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
...
+      clusterStates.forEach((clusterState) => {
+        const cluster = this.getById(clusterState.id);
+        if (cluster) {
+          cluster.setState(clusterState.state);
+        }
+      });
+    } else {
+      handleRequest(this.stateRequestChannel, (): clusterStateSync[] => {
```
Nokel81:
```suggestion
      handleRequest(ClusterStore.stateRequestChannel, (): clusterStateSync[] => {
```

jakolehm:
@nevalla also PTAL.

Nokel81 (approved):

PR #1578, opened 2020-11-30
Fix extension status-bar item default styles

jakolehm:
- set default icon/text color (white)
- add default hover

### DIFF
--- a/src/renderer/components/cluster-manager/bottom-bar.scss
+++ b/src/renderer/components/cluster-manager/bottom-bar.scss
@@ -17,4 +17,15 @@
       cursor: pointer;
     }
   }
+
+  .extensions {
+    font-size: var(--font-size-small);
+    color: white;
+    .item {
+      &:hover {
+        background-color: #ffffff33;
+        cursor: pointer;
+      }
+    }
+  }
 }
--- a/src/renderer/components/cluster-manager/bottom-bar.scss
+++ b/src/renderer/components/cluster-manager/bottom-bar.scss
@@ -22,6 +22,7 @@
     font-size: var(--font-size-small);
     color: white;
     .item {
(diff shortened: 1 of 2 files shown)


src/renderer/components/cluster-manager/bottom-bar.scss
```diff
@@ -17,4 +17,15 @@
       cursor: pointer;
     }
   }
+
+  .extensions {
+    font-size: var(--font-size-small);
+    color: white;
+    .item {
+      &:hover {
+        background-color: #ffffff33;
```
Nokel81, line 27: This should be part of the colour scheme file instead of being hard coded.

aleksfront (commented):
We should set such explicit styles only if we sure that most of the extensions will use it. I'm wondering about some status message ones that doesn't involve the use of a mouse events. We're trying to isolate parent styling to prevent extension developers to fight against them. So the question is should we encourage developers to style their own bottom line items?

aleksfront:
I've removed similar styles with one of recent PRs 😄 . PTAL https://github.com/lensapp/lens/pull/1331/files

jakolehm:
@aleksfront maybe the hover can be taken away.. but I don't see any reason not to default to `white`, it should be Lens responsibility to keep status-bar look&feel consistent. Currently it defaults to some gray-ish color that looks really weird -> every extension developer needs to set the same default over and over again.

aleksfront (approved):

src/renderer/components/cluster-manager/bottom-bar.scss
jakolehm: This colour was already hardcoded here. I think the problem is that bottom-bar does not follow dark/light theme -> we cannot use those variables here. /cc @aleksfront

PR #1579, opened 2020-11-30
fix symlinking extensions into .k8slens/extensions folder

Nokel81:

### DIFF
--- a/src/extensions/extension-discovery.ts
+++ b/src/extensions/extension-discovery.ts
@@ -114,7 +114,7 @@ export class ExtensionDiscovery {
     // chokidar works better than fs.watch
     chokidar.watch(this.localFolderPath, {
       // Dont watch recursively into subdirectories
-      depth: 0,
+      depth: 1,
       // Try to wait until the file has been completely copied.
       // The OS might emit an event for added file even it's not completely written to the filesysten.
       awaitWriteFinish: {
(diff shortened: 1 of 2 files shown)


src/extensions/extension-discovery.ts
```diff
@@ -114,7 +114,7 @@ export class ExtensionDiscovery {
     // chokidar works better than fs.watch
     chokidar.watch(this.localFolderPath, {
       // Dont watch recursively into subdirectories
-      depth: 0,
+      depth: 1,
```
chenhunghan, line 117: Could you explain a little bit more (in the PR description) why this fix symlinking? There seems to be some discussion in Slack, but maybe some team member (like me) didn't fully follow.

panuhorsmalahti: Could add comment to the code explaining that depth: 1 is required for symlinking to work.

panuhorsmalahti (approved):

src/extensions/extension-discovery.ts
panuhorsmalahti: @nevalla also noticed that depth: 1 is required in Windows

Nokel81: Should I add that to the comment as well?

panuhorsmalahti: Not necessarily I guess, just FYI.

PR #1435, opened 2020-11-19
Split internal functionality of Cluster to ManagedCluster

jakolehm:
- splits functionality that can only work in main process to a facade class called `ManagedCluster`
- simplifies Cluster api that is exposed via extension-api

Requires #1225

PR #1589, opened 2020-12-01
Tag cluster & workspace as beta features in extension api

jakolehm:
Both of those APIs still include internal details that we really have to cleanup. Tagging these as `beta` will set better expectation level and also allows us to tweak these APIs within 4.x series (until we take the beta tag away).


jakolehm:
Cluster cleanup has been already started but it didn't make it into 4.0: https://github.com/lensapp/lens/pull/1435

src/common/workspace-store.ts
```diff
@@ -26,11 +26,38 @@ export interface WorkspaceState {
...
+/**
+ * Workspace
+ *
+ * @beta
+ */
 export class Workspace implements WorkspaceModel, WorkspaceState {
+  /**
+   * Unique id
+   * @observable
+   */
```
Nokel81: Won't this whole PR conflict with #1573 ?

src/common/workspace-store.ts
```diff
@@ -44,23 +71,39 @@ export class Workspace implements WorkspaceModel, WorkspaceState {
...
   getState(): WorkspaceState {
     return {
       enabled: this.enabled
     };
   }
 
+  /**
+   * Push state
+   *
+   * @interal
```
Nokel81, line 106: I really think that any API marked as `@internal` should be not exported in a similar style to #1546.

src/main/cluster.ts
```diff
@@ -48,10 +48,31 @@ export interface ClusterState {
   allowedResources: string[]
 }
 
+/**
+ * Cluster
+ *
+ * @beta
```
Nokel81, line 55: This PR should probably have some doc describing what we mean by these markers and what guarantees (or not) they mean.

src/common/workspace-store.ts
jakolehm: I also think the same but we ran out of time. See #1435.

Nokel81: Sounds good

Nokel81 (approved):

nevalla (approved):
LGTM

PR #1591, opened 2020-12-01
Disable extension install button immediately

### LINKED ISSUE #1569: Disable Extension install button immediately
At the moment the Install button is disabled after the url has been downloaded, not immediately after the user clicks Install. We should get rid of this delay.

panuhorsmalahti:
* Disable extension install button immediately after clicked, not when the unpacking starts
* Convert a few sync blocking calls to async
* Fixes #1569

### DIFF
--- a/src/renderer/components/+extensions/extensions.tsx
+++ b/src/renderer/components/+extensions/extensions.tsx
@@ -62,6 +62,9 @@ export class Extensions extends React.Component {
   @observable search = "";
   @observable installPath = "";
 
+  // True if the preliminary install steps have started, but unpackExtension has not started yet
+  @observable startingInstall = false;
+
   /**
    * Extensions that were removed from extensions but are still in "uninstalling" state
    */
@@ -152,6 +155,8 @@ export class Extensions extends React.Component {
     const { installPath } = this;
 
     if (!installPath) return;
+
+    this.startingInstall = true;
     const fileName = path.basename(installPath);
 
     try {
@@ -161,13 +166,14 @@ export class Extensions extends React.Component {
         const { promise: filePromise } = downloadFile({ url: installPath, timeout: 60000 /*1m*/ });
         const data = await filePromise;
 
-        this.requestInstall({ fileName, data });
+        await this.requestInstall({ fileName, data });
       }
       // otherwise installing from system path
       else if (InputValidators.isPath.validate(installPath)) {
-        this.requestInstall({ fileName, filePath: installPath });
+        await this.requestInstall({ fileName, filePath: installPath });
       }
     } catch (error) {
+      this.startingInstall = false;
       Notifications.error(
         <p>Installation has failed: <b>{String(error)}</b></p>
       );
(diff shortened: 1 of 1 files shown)


src/renderer/components/+extensions/extensions.tsx
```diff
@@ -158,13 +164,14 @@ export class Extensions extends React.Component {
...
-        this.requestInstall({ fileName, data });
+        await this.requestInstall({ fileName, data });
       }
       // otherwise installing from system path
       else if (InputValidators.isPath.validate(installPath)) {
-        this.requestInstall({ fileName, filePath: installPath });
+        await this.requestInstall({ fileName, filePath: installPath });
       }
     } catch (error) {
+      this.startingInstall = false;
```
ixrock, line 176: Probably should be within `finally`-block.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -232,10 +240,10 @@ export class Extensions extends React.Component {
     // copy files to temp
     await fse.ensureDir(this.getExtensionPackageTemp());
 
-    requests.forEach(req => {
-      const tempFile = this.getExtensionPackageTemp(req.fileName);
-      fse.writeFileSync(tempFile, req.data);
-    });
+    for (const request of requests) {
```
ixrock, line 256: Replacement to `for-of` because `fse.writeFile` should break whole method & throw errors if any?

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -62,6 +62,9 @@ export class Extensions extends React.Component {
   @observable search = "";
   @observable installPath = "";
 
+  // True if the preliminary install steps have started, but unpackExtension has not started yet
+  @observable startingInstall = false;
```
ixrock, line 66:
```suggestion
  @observable installing = false;
```

ixrock (commented):
Btw, why do we have to disable button while installing extension? What if I want (as end user) start another installation while first one not finished?

panuhorsmalahti:
> Btw, why do we have to disable button while installing extension? What if I want (as end user) start another installation while first one not finished?

This was a design decision by @jakolehm. I think we want to make the extension system simple for now.

src/renderer/components/+extensions/extensions.tsx
panuhorsmalahti: "installing" would be misleading, as it's only true in the initial phase of the install. There is a computed `isInstalling` which returns e.g. "installing".

src/renderer/components/+extensions/extensions.tsx
panuhorsmalahti: Yeah, if I used forEach with async callback then we could lose errors

src/renderer/components/+extensions/extensions.tsx
panuhorsmalahti: finally wouldn't work, because we could reach that before the pre-install phase is completed.

aleksfront (approved):
LGTM

PR #1596, opened 2020-12-01
Fix extensions installation on Windows

### LINKED ISSUE #1564: Extensions install is broken on Windows
Lens fails to unpack npm package during the installation.

nevalla:
Fixes #1564 

### DIFF
--- a/src/renderer/components/+extensions/extensions.tsx
+++ b/src/renderer/components/+extensions/extensions.tsx
@@ -189,7 +189,7 @@ export class Extensions extends React.Component {
         .filter(req => !req.data && req.filePath)
         .map(async request => {
           try {
-            const data = fse.readFileSync(request.filePath);
+            const data = await fse.readFile(request.filePath);
             request.data = data;
             preloadedRequests.push(request);
             return request;
@@ -197,7 +197,7 @@ export class Extensions extends React.Component {
             if (showError) {
               Notifications.error(`Error while reading "${request.filePath}": ${String(error)}`);
             }
-          };
+          }
         })
     );
 
--- a/src/renderer/components/+extensions/extensions.tsx
+++ b/src/renderer/components/+extensions/extensions.tsx
@@ -47,7 +47,7 @@ interface ExtensionState {
 
 @observer
 export class Extensions extends React.Component {
-  private supportedFormats = [".tar", ".tgz"];
+  private supportedFormats = ["tar", "tgz"];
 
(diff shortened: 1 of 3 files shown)


src/renderer/components/+extensions/extensions.tsx
```diff
@@ -187,15 +187,17 @@ export class Extensions extends React.Component {
     await Promise.all(
       requests
         .filter(req => !req.data && req.filePath)
-        .map(request => {
-          return fse.readFile(request.filePath).then(data => {
+        .map(async request => {
+          try {
+            const data = fse.readFileSync(request.filePath);
```
nevalla: I don't know why, but using `readFileSync` is notably faster on Windows

panuhorsmalahti: I don't think sync methods should be used in the UI, since they lock the UI. (No idea why it would be faster).

panuhorsmalahti (approved):

src/renderer/components/+extensions/extensions.tsx
Nokel81: I, personally, take that a step further, I don't think file IO at all should be done in the renderer.

panuhorsmalahti: Agree, I think we should start to move "non-UI" code to "services".

nevalla: Got it working with `readFile`.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -187,15 +187,17 @@ export class Extensions extends React.Component {
     await Promise.all(
       requests
         .filter(req => !req.data && req.filePath)
-        .map(request => {
-          return fse.readFile(request.filePath).then(data => {
+        .map(async request => {
+          try {
+            const data = await fse.readFile(request.filePath);
```
jakolehm, line 192: Does this change anything else than syntax?

jakolehm (approved):

src/renderer/components/+extensions/extensions.tsx
nevalla: No. Initially I noticed that previous implementation was slow on Windows and `readFileSync` was significantly faster. Then I just converted it to use `readFile` with await and did not see any effect. Perhaps the original issue with slowness was something else.

PR #1598, opened 2020-12-01
Convert StatusBarRegistration to use components field

### LINKED ISSUE #1571: Extension API statusBarItems doesn't register status bar items
**Describe the bug**

I have an extension using `statusBarItems` api <https://docs.k8slens.dev/master/extensions/capabilities/common-capabilities/#status-bar-items>. but no icons are shown on the status bar after I launch Lens.

Can't see the status bar item in DOM tree as well, so I guess is not a CSS issue but related to extension manage/loader?

**To Reproduce**
1. Open Lens
2. No status bar item

**Expected behavior**

An icon on status bar.

**Screenshots**

Screenshot 2020-11-30 at 16 45 54

- Lens Version: 4.0.0-rc.1
- OS: Big Sur
- Installation method (e.g. snap or AppImage in Linux): .dmg
 
**Logs:**

No error logs in developer tools console.

**Additional context**

The source code of my extension
```tsx
import { LensRendererExtension, Interface, Component } from "@k8slens/extensions";
import React from "react"

const { Icon } = Component;

const StatusBarItemIcon = ({ navigate }: { navigate?: () => void }): JSX.Element => (
  <Icon
    material="link"
    interactive
    style={{
      color: "rgb(255, 255, 0)"
    }}
    onClick={() => navigate()}
  />
)

export default class RendererExtension extends LensRendererExtension {
  statusBarItems: Interface.StatusBarRegistration[] = [
    {
      item: (): JSX.Element => <StatusBarItemIcon navigate={this.navigate.bind(this)} />,
    }
  ]
}
```

Nokel81:
- More similar to all other *Registration types for extensions

- Simpler fix for using the components.Icon type, now accepts functions
  that return component instance like all other *Registration types

- Fixed not linting the whole repo, including fixing a type decl

fixes #1571

### DIFF
--- a/docs/extensions/capabilities/common-capabilities.md
+++ b/docs/extensions/capabilities/common-capabilities.md
@@ -40,7 +40,7 @@ This extension can register custom app menus that will be displayed on OS native
 
 Example:
 
-``` typescript
+```typescript
 import { LensMainExtension, windowManager } from "@k8slens/extensions"
 
 export default class ExampleMainExtension extends LensMainExtension {
@@ -92,7 +92,7 @@ export default class ExampleMainExtension extends LensRendererExtension {
 
 This extension can register custom global pages (views) to Lens's main window. The global page is a full-screen page that hides all other content from a window.
 
-``` typescript
+```typescript
 import React from "react"
 import { Component, LensRendererExtension } from "@k8slens/extensions"
 import { ExamplePage } from "./src/example-page"
@@ -123,7 +123,7 @@ export default class ExampleRendererExtension extends LensRendererExtension {
 
 This extension can register custom app preferences. It is responsible for storing a state for custom preferences.
 
-``` typescript
+```typescript
 import React from "react"
 import { LensRendererExtension } from "@k8slens/extensions"
 import { myCustomPreferencesStore } from "./src/my-custom-preferences-store"
@@ -147,7 +147,7 @@ export default class ExampleRendererExtension extends LensRendererExtension {
 
 This extension can register custom cluster pages. These pages are visible in a cluster menu when a cluster is opened.
 
-``` typescript
+```typescript
 import React from "react"
 import { LensRendererExtension } from "@k8slens/extensions";
 import { ExampleIcon, ExamplePage } from "./src/page"
@@ -180,7 +180,7 @@ export default class ExampleExtension extends LensRendererExtension {
 
 This extension can register installable features for a cluster. These features are visible in the "Cluster Settings" page.
 
-``` typescript
+```typescript
 import React from "react"
 import { LensRendererExtension } from "@k8slens/extensions"
 import { MyCustomFeature } from "./src/my-custom-feature"
@@ -209,18 +209,20 @@ export default class ExampleExtension extends LensRendererExtension {
 
 This extension can register custom icons and text to a status bar area.
 
-``` typescript
+```typescript
 import React from "react";
 import { Component, LensRendererExtension, Navigation } from "@k8slens/extensions";
 
 export default class ExampleExtension extends LensRendererExtension {
   statusBarItems = [
     {
-      item: (
-        <div className="flex align-center gaps hover-highlight" onClick={() => this.navigate("/example-page")} >
-          <Component.Icon material="favorite" />
-        </div>
-      )
+      components: {
+        Item: (
+          <div className="flex align-center gaps hover-highlight" onClick={() => this.navigate("/example-page")} >
+            <Component.Icon material="favorite" />
+          </div>
+        )
+      }
     }
   ]
 }
@@ -231,7 +233,7 @@ export default class ExampleExtension extends LensRendererExtension {
 
 This extension can register custom menu items (actions) for specified Kubernetes kinds/apiVersions.
 
-``` typescript
+```typescript
 import React from "react"
 import { LensRendererExtension } from "@k8slens/extensions";
 import { CustomMenuItem, CustomMenuItemProps } from "./src/custom-menu-item"
@@ -254,7 +256,7 @@ export default class ExampleExtension extends LensRendererExtension {
 
 This extension can register custom details (content) for specified Kubernetes kinds/apiVersions.
 
-``` typescript
+```typescript
 import React from "react"
 import { LensRendererExtension } from "@k8slens/extensions";
 import { CustomKindDetails, CustomKindDetailsProps } from "./src/custom-kind-details"
--- a/docs/extensions/guides/renderer-extension.md
+++ b/docs/extensions/guides/renderer-extension.md
@@ -629,7 +629,7 @@ export default class HelpExtension extends LensRendererExtension {
 
 The properties of the `statusBarItems` array objects are defined as follows:
 
-* `item` specifies the `React.Component` that will be shown on the status bar. By default, items are added starting from the right side of the status bar. Due to limited space in the status bar, `item` will typically specify only an icon or a short string of text. The example above reuses the `HelpIcon` from the [`globalPageMenus` guide](#globalpagemenus).
+* `Item` specifies the `React.Component` that will be shown on the status bar. By default, items are added starting from the right side of the status bar. Due to limited space in the status bar, `Item` will typically specify only an icon or a short string of text. The example above reuses the `HelpIcon` from the [`globalPageMenus` guide](#globalpagemenus).
 * `onClick` determines what the `statusBarItem` does when it is clicked. In the example, `onClick` is set to a function that calls the `LensRendererExtension` `navigate()` method. `navigate` takes the `id` of the associated global page as a parameter. Thus, clicking the status bar item activates the associated global pages.
 
 ### `kubeObjectMenuItems`
--- a/docs/extensions/guides/renderer-extension.md
+++ b/docs/extensions/guides/renderer-extension.md
@@ -22,7 +22,7 @@ All UI elements are based on React components.
 
 To create a renderer extension, extend the `LensRendererExtension` class:
 
-``` typescript
+```typescript
 import { LensRendererExtension } from "@k8slens/extensions";
 
 export default class ExampleExtensionMain extends LensRendererExtension {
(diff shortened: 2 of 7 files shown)


.eslintrc.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
   ignorePatterns: [
     "**/node_modules/**/*",
     "**/dist/**/*",
+    "**/static/**/*",
```
jim-docker: is this required to fulfill this PR's title? Otherwise could be done in a separate PR where it doesn't muddy things?

src/extensions/registries/status-bar-registry.ts
```diff
@@ -3,8 +3,12 @@
 import React from "react";
 import { BaseRegistry } from "./base-registry";
 
+export interface StatusBarComponents {
+  Icon?: React.ComponentType<any>;
```
jim-docker: Why `Icon`? Why not `Item`?

Nokel81: I guess I was caught up with the specific example I was working with. Good catch.

.eslintrc.js
Nokel81: Sounds good.

jim-docker:
Doesn't look like it impacts aqua's extension but it does impact @stefcameron's lens-extension-cc

docs/extensions/capabilities/common-capabilities.md
```diff
@@ -216,11 +216,13 @@ import { Component, LensRendererExtension, Navigation } from "@k8slens/extension
 export default class ExampleExtension extends LensRendererExtension {
   statusBarItems = [
     {
-      item: (
-        <div className="flex align-center gaps hover-highlight" onClick={() => this.navigate("/example-page")} >
-          <Component.Icon material="favorite" />
-        </div>
-      )
+      components: {
+        Icon: (
```
jim-docker: `Item`

docs/extensions/guides/renderer-extension.md
```diff
@@ -477,15 +477,17 @@ export default class HelpExtension extends LensRendererExtension {
...
-        <div
-          className="flex align-center gaps"
-          onClick={() => this.navigate("help")}
-        >
-          <HelpIcon />
-          My Status Bar Item
-        </div>
-      ),
+      components: {
+        Icon: (
```
jim-docker: `Item`

jim-docker (commented):
LGTM but @jakolehm should decide when this is merged

src/extensions/registries/status-bar-registry.ts
```diff
@@ -3,8 +3,12 @@
 import React from "react";
 import { BaseRegistry } from "./base-registry";
 
+export interface StatusBarComponents {
+  Item?: React.ComponentType<any>;
+}
+
 export interface StatusBarRegistration {
-  item?: React.ReactNode;
+  components: StatusBarComponents;
```
stefcameron, line 11: It feels strange to have this name be plural. Even based on the type def, we're still talking about a _singular_ component. Whether that "component" is a fragment that's technically an array of components (but still, I believe, seen as a single component within the realms of React), or is a classical or functional single component, it's still just one component, right...?

stefcameron (commented):
And yes, this does break my extension, thank you for the head's up, @jim-docker , but thankfully the fix will be very easy when this ships. I just have to remember to check this PR to know what version it ships in.

src/extensions/registries/status-bar-registry.ts
jim-docker: It's consistent across all the registration types. Each of the XXXRegistration interfaces now has a `components` field, which map to interfaces specifying one or more components:

```
export interface PageComponents {
  Page: React.ComponentType<any>;
}
```

```
export interface AppPreferenceComponents {
  Hint: React.ComponentType<any>;
  Input: React.ComponentType<any>;
}
```

better to use `components` for all rather than `component` for some and `components` for others. Or do you suggest changing all of them to `component`?

stefcameron: Oh, I see. Consistency is key, for sure. What does `React.ComponentType<any>` represent, though. An array? If so, then I support the plural name. If not, why the "s"?

In any case, I'm probably fighting a losing battle since there is a lot of precedent if all other types are already using the plural form, and updating all of them would break a lot more stuff than just my extension.

jakolehm:
> LGTM but @jakolehm should decide when this is merged

Code looks good but I'm not happy about the fact that we are willing to break the API here just before GA. #1606 contains minimal fix for the issue and that probably should be included in 4.0. This PR can be targeted for 4.1 if we come up a way to support multiple interfaces/versions (should be doable, see https://gist.github.com/jakolehm/7c79031856bab06e245d19b2eaa0ff38 for ideas).

jim-docker:
This wouldn't have to breaking if it supported both:
```
export interface StatusBarRegistration {
  item?: React.ComponentType<any>;    // deprecated, use components field instead
  components?: {
    Item: React.ComponentType<any>;
  }
}
```
but then we'd have to tolerate releasing an api with a deprecated field...

Nokel81:
That is probably what we would should do.

stefcameron:
Jim's proposal is what I was going to propose.

Nokel81:
I'll change my PR to that^^

roderickmarj (approved):

roderickmarj (approved):

roderickmarj (approved):

roderickmarj (approved):

roderickmarj (approved):

jakolehm (dismissed):

docs/extensions/capabilities/common-capabilities.md
```diff
@@ -40,7 +40,7 @@ This extension can register custom app menus that will be displayed on OS native
 
 Example:
 
-``` typescript
+```tsx
```
jim-docker: mkdocs doesn't support tsx here, right?

docs/extensions/guides/renderer-extension.md
```diff
@@ -570,15 +570,17 @@ export default class HelpExtension extends LensRendererExtension {
...
-      item: (
-        <div
-          className="flex align-center gaps"
-          onClick={() => this.navigate("help")}
-        >
-          <HelpIcon />
-          My Status Bar Item
-        </div>
-      ),
+      components: {
```
jim-docker, line 614: @jakolehm, are the public docs for 4.0 pulled from master or a release branch? I think @pauljwil is making changes to docs (which would be good for 4.0.x and above) in master, but the change in this PR is strictly 4.1

If #1789 is meant for 4.1 only then there's no issue. If we want #1789 to be reflected in 4.0.x docs then we have to be careful here

docs/extensions/guides/renderer-extension.md
```diff
@@ -570,15 +570,17 @@ export default class HelpExtension extends LensRendererExtension {
...
+        Item: (
+          <div
+            className="flex align-center gaps"
+            onClick={() => this.navigate("help")}
+          >
+            <HelpIcon />
+            My Status Bar Item
+          </div>
+        )
+      },
```
jim-docker, line 624: The descriptive text below would have to change as well.

src/extensions/registries/status-bar-registry.ts
```diff
@@ -3,7 +3,18 @@
 import React from "react";
 import { BaseRegistry } from "./base-registry";
 
-export interface StatusBarRegistration {
+export interface StatusBarComponents {
+  Item?: React.ComponentType;
+}
+
+export interface StatusBarRegistrationV2 {
```
jim-docker: Does this _have_ to be exported? When it comes time to actually remove the `item` field, won't we just eliminate this interface and redefine `StatusBarRegistration` accordingly? If so it would be better if we didn't export something we'll eventually remove.

src/renderer/components/cluster-manager/bottom-bar.tsx
```diff
@@ -5,14 +5,47 @@ import { observer } from "mobx-react";
 import { Icon } from "../icon";
 import { WorkspaceMenu } from "../+workspaces/workspace-menu";
 import { workspaceStore } from "../../../common/workspace-store";
-import { statusBarRegistry } from "../../../extensions/registries";
+import { StatusBarRegistration, statusBarRegistry } from "../../../extensions/registries";
+import _ from "lodash";
 
 @observer
 export class BottomBar extends React.Component {
+  renderRegistrated(registration: StatusBarRegistration) {
```
jim-docker: registrated? There are some obscure definitions for it. What about `renderRegisteredItem`?

docs/extensions/guides/renderer-extension.md
jakolehm: Docs are generated from a release.

jim-docker (dismissed):

jakolehm (dismissed):

ixrock (dismissed):

Nokel81:
@jakolehm, @jim-docker, @ixrock could you PTL again, I needed to rebase to clear conflicts

docs/extensions/guides/renderer-extension.md
```diff
@@ -597,23 +611,25 @@ export default class HelpExtension extends LensRendererExtension {
...
     },
   ];
 }
 ```
 
 The properties of the `statusBarItems` array objects are defined as follows:
 
-* `item` specifies the `React.Component` that will be shown on the status bar. By default, items are added starting from the right side of the status bar. Due to limited space in the status bar, `item` will typically specify only an icon or a short string of text. The example above reuses the `HelpIcon` from the [`globalPageMenus` guide](#globalpagemenus). 
+* `item` specifies the `React.Component` that will be shown on the status bar. By default, items are added starting from the right side of the status bar. Due to limited space in the status bar, `item` will typically specify only an icon or a short string of text. The example above reuses the `HelpIcon` from the [`globalPageMenus` guide](#globalpagemenus).
```
jim-docker: Should be `Item` not `item`

jim-docker (dismissed):

docs/extensions/capabilities/common-capabilities.md
```diff
@@ -40,7 +40,7 @@ This extension can register custom app menus that will be displayed on OS native
 
 Example:
 
-``` typescript
+```typescript
```
jim-docker, line 43: just curious, having a space is not OK? Seems to work fine

Nokel81: It is fine, I was just trying to unify them.

jim-docker (approved):

jakolehm (approved):

PR #1599, opened 2020-12-01
gitignore .vscode

Nokel81:


panuhorsmalahti (approved):
Alternatively we could commit default .vscode settings for the repo, if we want to specially support this IDE (which just might be the most popular IDE for this project).

jim-docker (approved):

PR #1602, opened 2020-12-01
persist extension installation state for the duration of the window's life

### LINKED ISSUE #1590: Extensions page state reset when page reopened
If the extensions page is closed and opened while extension install is being installed, the state is lost, e.g. the installing animation on the disabled button is not displayed.

Nokel81:
fixes #1590

known limitation: I didn't verify if an installation continues when the lens window is closed but I don't believe so since it seems to be done completely within the UI. If this isn't the case then this probably needs a longer lasting solution.

### DIFF
--- a/src/renderer/components/+extensions/extensions.tsx
+++ b/src/renderer/components/+extensions/extensions.tsx
@@ -22,6 +22,7 @@ import { PageLayout } from "../layout/page-layout";
 import { SubTitle } from "../layout/sub-title";
 import { Notifications } from "../notifications";
 import { TooltipPosition } from "../tooltip";
+import { ExtensionStateStore } from "./extension-install.store";
 import "./extensions.scss";
 
 interface InstallRequest {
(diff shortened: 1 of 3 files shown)


panuhorsmalahti (approved):

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -39,25 +40,20 @@ interface InstallRequestValidated extends InstallRequestPreloaded {
...
-  private installPathValidator: InputValidator = {
+  private static installPathValidator: InputValidator = {
     message: <Trans>Invalid URL or absolute path</Trans>,
     validate(value: string) {
       return InputValidators.isUrl.validate(value) || InputValidators.isPath.validate(value);
     }
   };
 
-  @observable
-  extensionState = observable.map<string, ExtensionState>();
```
panuhorsmalahti, line 60: I wonder if we could alternatively make this property `static` and thus singleton?

aleksfront (approved):
Looking good. Just minor thing: probably need to change store filename from `extension-install.store.ts`  to `extension-state.store.ts`.

src/renderer/components/+extensions/extensions.tsx
Nokel81: Probably, that is basically the same thing. I just tried to follow the same style as before.

Nokel81:
@panuhorsmalahti Can you give some guidance about the test that is failing? I cannot see how this test was ever passing.

PR #1603, opened 2020-12-01
fix spelling in documentation

Nokel81:

I noticed that there was some spelling mistakes in our docs. This PR fixes that.

- Chose `en_GB` as the dictionary as most of the team is in Europe. Though this is a team decision and can be easily changed before the PR goes through.

- We should probably set up some sort of CI to spell check our docs for us (after generating the API docs so that we spell check those too).

### DIFF
--- a/docs/clusters/settings.md
+++ b/docs/clusters/settings.md
@@ -1,8 +1,8 @@
 # Settings
 
-It is easy to configure Lens Clusters to your liking through its various settings. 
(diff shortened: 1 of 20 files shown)


panuhorsmalahti:
I'd use en_US to be honest. I don't think _team_ should determine what language(/variation) the docs should use. en_US is likely more common with technical documentation than en_GB.

jim-docker:
We need something to check spelling in PR titles too :)

PR #1605, opened 2020-12-01
store more than largest kube api request amount in the event store

### LINKED ISSUE #1592: Event buffer is too small for big clusters
**Describe the bug**

Lens currently uses `limit=1000` in `EventStore` to protect kubernetes api server. The same `1000` is also used for internal event buffer size which does not make that much of sense. Buffer will go full immediately after event query is done and first event comes from watch API -> Lens starts to loose events.

Nokel81:

fixes #1592

### DIFF
--- a/src/renderer/kube-object.store.ts
+++ b/src/renderer/kube-object.store.ts
@@ -11,7 +11,8 @@ import { getHostedCluster } from "../common/cluster-store";
 @autobind()
 export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemStore<T> {
   abstract api: KubeApi<T>;
-  public limit: number;
+  public readonly abstract limit: number;
+  public readonly abstract saveLimit: number;
 
   constructor() {
     super();
@@ -179,9 +180,9 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
       return;
     }
     // create latest non-observable copy of items to apply updates in one action (==single render)
-    let items = this.items.toJS();
+    const items = this.items.toJS();
 
-    this.eventsBuffer.clear().forEach(({ type, object }) => {
+    for (const {type, object} of this.eventsBuffer.clear()) {
       const { uid, selfLink } = object.metadata;
       const index = items.findIndex(item => item.getId() === uid);
       const item = items[index];
@@ -204,14 +205,9 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
           }
           break;
       }
-    });
-
-    // slice to max allowed items
-    if (this.limit && items.length > this.limit) {
-      items = items.slice(-this.limit);
     }
 
     // update items
-    this.items.replace(this.sortItems(items));
+    this.items.replace(this.sortItems(items.slice(-this.saveLimit)));
   }
 }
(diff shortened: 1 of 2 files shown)


src/renderer/kube-object.store.ts
```diff
@@ -11,7 +11,8 @@ import { getHostedCluster } from "../common/cluster-store";
 @autobind()
 export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemStore<T> {
   abstract api: KubeApi<T>;
-  public limit: number;
+  public readonly abstract limit: number;
```
panuhorsmalahti: We could call limit `requestLimit`

panuhorsmalahti (approved):

src/renderer/kube-object.store.ts
jakolehm: Changing name would be a breaking change.. even `abstract` is breaking change (CI shows failures).

src/renderer/kube-object.store.ts
```diff
@@ -11,7 +11,8 @@ import { getHostedCluster } from "../common/cluster-store";
...
+  public readonly abstract saveLimit: number;
```
jakolehm: Same here, probably cannot be abstract. Also `bufferSize` might be a better name?

jakolehm (changes requested):

src/renderer/kube-object.store.ts
Nokel81: @jakolehm Removed `abstract` can you review again, please?

Nokel81:
@jakolehm Would you be able to take a look at this again, thanks!

src/renderer/kube-object.store.ts
```diff
@@ -11,14 +11,25 @@ import { getHostedCluster } from "../common/cluster-store";
...
   constructor() {
     super();
     this.bindWatchEventsUpdater();
     kubeWatchApi.addListener(this, this.onWatchApiEvent);
   }
 
+  get query(): IKubeApiQueryParams {
+    const { limit } = this;
+
+    if (!limit) {
```
ixrock, line 26: So `{limit: 0}` doesn't make any sense?

Nokel81: This was just a direct move from the old line 66 below. I don't know... does kube treat `limit: 0` as the same as not passing `limit` at all? (given than kube is written in Go, that is probably the case sadly).

src/renderer/kube-object.store.ts
```diff
@@ -11,14 +11,25 @@ import { getHostedCluster } from "../common/cluster-store";
 @autobind()
 export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemStore<T> {
   abstract api: KubeApi<T>;
-  public limit: number;
+  public readonly limit?: number;
+  public readonly bufferSize: number = 50000;
```
ixrock, line 15:
```suggestion
  public readonly maxLimit: number = 50000;
```

ixrock (approved):
LGTM

src/renderer/kube-object.store.ts
ixrock: or `saveLimit` maybe?

Nokel81: I had `saveLimit` initially, Jari suggested `bufferSize`

jakolehm (approved):

PR #1609, opened 2020-12-02
Fix getExtensionPageUrl on Windows

### LINKED ISSUE #1607: Extension cluster pages does not open on Windows
Link to cluster pages coming from extensions is not right on Windows and pages do not open.

image

image

nevalla:
Fixes #1607 

### DIFF
--- a/src/extensions/registries/page-registry.ts
+++ b/src/extensions/registries/page-registry.ts
@@ -44,7 +44,7 @@ export function getExtensionPageUrl<P extends object>({ extensionId, pageId = ""
   const extensionBaseUrl = compile(`/extension/:name`)({
     name: sanitizeExtensionName(extensionId), // compile only with extension-id first and define base path
   });
-  const extPageRoutePath = path.join(extensionBaseUrl, pageId); // page-id might contain route :param-s, so don't compile yet
+  const extPageRoutePath = path.join(extensionBaseUrl, pageId).replace(/\\/g, "/"); // path.join returns \ as separator on Windows and needs to be replaced with /
 
   if (params) {
     return compile(extPageRoutePath)(params); // might throw error when required params not passed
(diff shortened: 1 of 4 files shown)


src/extensions/registries/page-registry.ts
```diff
@@ -44,7 +44,7 @@ export function getExtensionPageUrl<P extends object>({ extensionId, pageId = ""
   const extensionBaseUrl = compile(`/extension/:name`)({
     name: sanitizeExtensionName(extensionId), // compile only with extension-id first and define base path
   });
-  const extPageRoutePath = path.join(extensionBaseUrl, pageId); // page-id might contain route :param-s, so don't compile yet
+  const extPageRoutePath = path.join(extensionBaseUrl, pageId).replace(/\\/g, "/"); // path.join returns \ as separator on Windows and needs to be replaced with /
```
jakolehm: A test for this would be nice.

nevalla: test is already there and actually it was failing. We would have seen that if tests were running on windows too. Will be fixed in #1610

panuhorsmalahti: It doesn't seem correct to use path.join (intended for file paths) for joining the url. See https://stackoverflow.com/a/16301947

nevalla: it seems that we cannot achieve the same with `new URL()` since we are combining the path part of the url. However `path.posix.join` might be something that we want to use.

jakolehm (approved):

src/extensions/registries/page-registry.ts
panuhorsmalahti: path.posix.join also doesn't sound like optimal here, even if `new URL()` is insufficient. Surely there's some npm dependency to join url parts properly.

Issue #1615, opened 2020-12-02
Cluster detector fails to detect popular distros


jakolehm:
Like: 

- [x] kind
- [x] rke
- [x] k3s
- k0s
- [x] Mirantis Kubernetes Engine
- [x] Docker Desktop Kubernetes
- [x] Openshift
- [x] Microk8s

PR #1617, opened 2020-12-02
Detect rancher distros

nevalla:
This PR will detect Rancher distros `rke` and `k3s` instead of previously detected `rancher`.

Related #1615 


Nokel81 (approved):
Looks good, I wonder if we should somehow do all the checks and pick the highest accuracy one.

jakolehm (approved):

PR #1619, opened 2020-12-02
Detect kind

jakolehm:
This PR will detect [kind](https://kind.sigs.k8s.io/) clusters.

Related #1615

### DIFF
--- a/src/main/cluster-detectors/distribution-detector.ts
+++ b/src/main/cluster-detectors/distribution-detector.ts
@@ -40,6 +40,10 @@ export class DistributionDetector extends BaseClusterDetector {
       return { value: "microk8s", accuracy: 80};
     }
 
+    if (this.isKind()) {
+      return { value: "kind", accuracy: 70};
+    }
+
     if (this.isCustom()) {
       return { value: "custom", accuracy: 10};
     }
@@ -81,6 +85,10 @@ export class DistributionDetector extends BaseClusterDetector {
 
   protected isMicrok8s() {
     return this.cluster.contextName.startsWith("microk8s");
+  }
+
+  protected isKind() {
+    return this.cluster.contextName.startsWith("kind-");
   }
 
   protected isCustom() {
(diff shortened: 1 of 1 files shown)


miskun (approved):

src/main/cluster-detectors/distribution-detector.ts
```diff
@@ -83,6 +87,10 @@ export class DistributionDetector extends BaseClusterDetector {
     return this.cluster.contextName.startsWith("microk8s");
   }
 
+  protected isKind() {
+    return this.cluster.contextName.startsWith("kind-");
```
jim-docker: Why don't we use the cluster name that is associated with the context? Users can easily make multiple contexts [(call one dev, one prod)](https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/) but the cluster name is less likely to be changed(?)

Probably not that likely with test/toy distributions like kind/minikube...

jakolehm: I thought kind prefixes context names always with `kind-`, even if user gives a custom name for a cluster.

jakolehm: Will verify this asap ^

jakolehm: Tested, it starts always with `kubernetes-admin@kind-` ... so there was a bug (fixed now).

nevalla (approved):

src/main/cluster-detectors/distribution-detector.ts
jim-docker: > I thought kind prefixes context names always with kind-, even if user gives a custom name for a cluster.

It does. But I can add my own contexts that use the kind cluster:
```
kubectl config set-context prod-frontend --cluster=kind-cluster --user=kind-cluster
```

Or is this not the same context? Am I out of context?

PR #1620, opened 2020-12-02
Refactor ExtensionsStore

panuhorsmalahti:
- Move functionality from ExtensionsStore to ExtensionLoader to simplify the store's logic

### DIFF
--- a/src/extensions/extension-loader.ts
+++ b/src/extensions/extension-loader.ts
@@ -51,6 +51,17 @@ export class ExtensionLoader {
     return extensions;
   }
 
+  // Transform userExtensions to a state object for storing into ExtensionsStore
+  @computed get storeState() {
+    return Object.fromEntries(
+      Array.from(this.userExtensions)
+        .map(([extId, extension]) => [extId, {
+          enabled: extension.isEnabled,
+          name: extension.manifest.name,
+        }])
+    );
+  }
+
   @action
   async init() {
     if (ipcRenderer) {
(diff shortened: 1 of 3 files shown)


src/extensions/extension-loader.ts
```diff
@@ -51,6 +51,18 @@ export class ExtensionLoader {
...
+  // Transform userExtensions to a state object for storing into ExtensionsStore
+  @computed get storeState() {
+    return Array.from(this.userExtensions).reduce((state, [extId, ext]) => {
+      state[extId] = {
+        enabled: ext.isEnabled,
+        name: ext.manifest.name,
+      };
+
+      return state;
+    }, {} as Record<LensExtensionId, LensExtensionState>);
```
Nokel81:
```suggestion
    return Object.fromEntries(
      Array.from(this.userExtensions)
        .map([extId, ext] => [extId, {
          enabled: ext.isEnabled,
          name: ext.manifest.name,
        }])
    );
```

Nokel81 (approved):

src/extensions/extension-loader.ts
```diff
@@ -51,6 +51,18 @@ export class ExtensionLoader {
...
+  }
+
   @action
   async init() {
```
Nokel81, line 66: I guess in the future, this really shouldn't be a class and instead some functions since we really only should "init" once.

src/extensions/extension-loader.ts
panuhorsmalahti: Thanks, changed

PR #1624, opened 2020-12-02
Eslint no-trailing-spaces, quote-props, space-infix-ops and arrow-parens

panuhorsmalahti:
Add eslint no-trailing-spaces, quote-props, space-infix-ops and arrow-parens rules.


Nokel81 (approved):

chenhunghan:
I like `no-trailing-spaces`, `quote-props`, `space-infix-ops`, I think these bring consistent code reading experience :)

but, I had some bad memory about `arrow-parens`, it does make the code cleaner, but I found it annoys me when I had to add parameters to functions.

for example, this follow `arrow-parens: as-needed`
```js
clustersList.forEach(cluster => {})
```
but I need to add `()` if I want to have something like
```js
clustersList.forEach((cluster, index) => {})
```
which is an extra step.

I would recommend change it to `"as-needed", { "requireForBlockBody": true }` 
see <https://eslint.org/docs/rules/arrow-parens#requireforblockbody>

panuhorsmalahti:
> I would recommend change it to `"as-needed", { "requireForBlockBody": true }`
> see https://eslint.org/docs/rules/arrow-parens#requireforblockbody

Thanks, changed

chenhunghan (approved):

PR #1625, opened 2020-12-02
Detect Openshift

jakolehm:
This PR adds detection for Openshift (4.x).

Related #1615

### DIFF
--- a/src/main/cluster-detectors/distribution-detector.ts
+++ b/src/main/cluster-detectors/distribution-detector.ts
@@ -60,10 +60,6 @@ export class DistributionDetector extends BaseClusterDetector {
       return { value: "custom", accuracy: 10};
     }
 
-    if (this.isCustom()) {
-      return { value: "custom", accuracy: 10};
-    }
-
     if (await this.isOpenshift()) {
       return { value: "openshift", accuracy: 90};
     }
--- a/src/main/cluster-detectors/distribution-detector.ts
+++ b/src/main/cluster-detectors/distribution-detector.ts
@@ -135,7 +135,7 @@ export class DistributionDetector extends BaseClusterDetector {
     try {
       const response = await this.k8sRequest("");
 
-      return response.paths?.find((api: string) => api === "/apis/project.openshift.io") !== undefined;
+      return response.paths?.includes("/apis/project.openshift.io");
     } catch (e) {
       return false;
     }
--- a/src/main/cluster-detectors/distribution-detector.ts
+++ b/src/main/cluster-detectors/distribution-detector.ts
@@ -60,6 +60,14 @@ export class DistributionDetector extends BaseClusterDetector {
       return { value: "custom", accuracy: 10};
     }
 
+    if (this.isCustom()) {
+      return { value: "custom", accuracy: 10};
+    }
+
+    if (await this.isOpenshift()) {
+      return { value: "openshift", accuracy: 90};
+    }
+
     return { value: "unknown", accuracy: 10};
   }
 
@@ -121,5 +129,15 @@ export class DistributionDetector extends BaseClusterDetector {
 
   protected isK3s() {
     return this.version.includes("+k3s");
+  }
+
+  protected async isOpenshift() {
+    try {
+      const response = await this.k8sRequest("");
+
+      return response.paths?.find((api: string) => api === "/apis/project.openshift.io") !== undefined;
+    } catch (e) {
+      return false;
+    }
   }
 }


src/main/cluster-detectors/distribution-detector.ts
```diff
@@ -4,6 +4,7 @@ import { ClusterMetadataKey } from "../cluster";
 export class DistributionDetector extends BaseClusterDetector {
   key = ClusterMetadataKey.DISTRIBUTION;
   version: string;
+  private cache = new Map<string, any>();
```
jakolehm: Hmm.. it seems that each detect run creates a new instance from detector class so this cache does not help at all 🤔 /cc @nevalla

Nokel81: It could if you make it static. But do we want to have a cache for detection?

jakolehm: > It could if you make it static. But do we want to have a cache for detection?

Detector is per cluster which makes static a bit complex. We can have a cache per cluster because distribution does not really change ever.

jim-docker: Is this to avoid making unnecessary api calls? Don't you have the cluster object and can check if the distribution is already set? Or we don't persist the distribution? Or maybe I need to study the code more deeply...

jakolehm: @jim-docker yes, we don't want to hit k8s api unless really necessary. Idk how did I miss that this class gets `Cluster` object, I'll refactor 👍

jakolehm: Oh no, using `Cluster` here is not enough, if end up with `unknown` we need to keep detecting -> we still need a separate cache. I'll refactor even more ...

jakolehm: Decided to remove caching altogether after discussing with @nevalla . Detectors are triggered every 15 minutes which is not too often. We can improve this logic later.

jakolehm:
@jim-docker @Nokel81 ptal again.

src/main/cluster-detectors/distribution-detector.ts
```diff
@@ -60,6 +60,18 @@ export class DistributionDetector extends BaseClusterDetector {
...
+    }
+
+    if (this.isCustom()) {
+      return { value: "custom", accuracy: 10};
+    }
+
+    if (await this.isOpenshift()) {
+      return { value: "openshift", accuracy: 90};
+    }
+
```
Nokel81: This is very weird. Why are we detecting twice?

src/main/cluster-detectors/distribution-detector.ts
```diff
@@ -122,4 +134,14 @@ export class DistributionDetector extends BaseClusterDetector {
   protected isK3s() {
     return this.version.includes("+k3s");
   }
+
+  protected async isOpenshift() {
+    try {
+      const response = await this.k8sRequest("");
+
+      return response.paths?.find((api: string) => api === "/apis/project.openshift.io") !== undefined;
```
Nokel81:
```suggestion
      return response.paths?.includes("/apis/project.openshift.io");
```

src/main/cluster.ts
```diff
@@ -425,17 +425,8 @@ export class Cluster implements ClusterModel, ClusterState {
 
     try {
       const namespaceList = await api.listNamespace();
-      const nsAccessStatuses = await Promise.all(
-        namespaceList.body.items.map(ns => this.canI({
-          namespace: ns.metadata.name,
-          resource: "pods",
-          verb: "list",
-        }))
-      );
```
Nokel81: How is this marked as "remove" when it already doesn't exist on master...

src/main/cluster.ts
```diff
@@ -425,17 +425,8 @@ export class Cluster implements ClusterModel, ClusterState {
...
 
-      return namespaceList.body.items
-        .filter((ns, i) => nsAccessStatuses[i])
-        .map(ns => ns.metadata.name);
+      return namespaceList.body.items.map(ns => ns.metadata.name);
```
Nokel81:
```suggestion
      return _.filter(namespaceList.body.items.map(ns => ns.metadata?.name));
```

metadata is an optional property and so is name.

src/main/exit-app.ts
```diff
@@ -10,10 +10,10 @@ export function exitApp() {
   const clusterManager = ClusterManager.getInstance<ClusterManager>();
 
   appEventBus.emit({ name: "service", action: "close" });
-  windowManager.hide();
-  clusterManager.stop();
+  windowManager?.hide();
+  clusterManager?.stop();
```
Nokel81: Oh I see what has happened, please rebase with master.

src/main/cluster-detectors/distribution-detector.ts
jakolehm: Weird indeed, it probably came from rebase. Will fix.

nevalla (dismissed):

src/main/cluster-detectors/distribution-detector.ts
```diff
@@ -60,6 +60,14 @@ export class DistributionDetector extends BaseClusterDetector {
       return { value: "custom", accuracy: 10};
     }
 
+    if (this.isCustom()) {
+      return { value: "custom", accuracy: 10};
+    }
+
```
Nokel81: I am still confused as to why this PR is adding this? Was it removed in another PR?

jakolehm: I'm also confused.. sorry about this (rebase is hard, sometimes). Should be fixed now.

Nokel81 (approved):

nevalla (approved):

PR #1631, opened 2020-12-03
Add timeout to app start and cluster add tracking events

nevalla:
When Lens application is opened it opens last active dashboard. At that time telemetry extension might not be active yet and it will miss tracking events. This PR will add timeout when `app start` and `cluster open` events are emitted.


jakolehm (approved):
We might want to refactor this post 4.0.0 so that we don't need timeouts.

Nokel81 (approved):
I agree with Jari that this is most definitely a hack. Having tried to implement a "remembrance" queue before it is not an easy problem to solve. 

I think that there should be an issue opened to track the design work on this. Before this PR is merged.

PR #1635, opened 2020-12-03
Visualize extension loading

### LINKED ISSUE #1608: Visualize extensions loading
If the extensions page is opened while the extensions are still being loaded after the app startup, we should visualize with e.g. spinner or text so that the user is not confused why the extensions are missing.

panuhorsmalahti:
Add spinner to the extensions page which is displayed if the extensions are still loading as the application is starting.

Screenshot 2020-12-03 at 11 28 09

Fixes #1608

### DIFF
--- a/src/renderer/bootstrap.tsx
+++ b/src/renderer/bootstrap.tsx
@@ -1,21 +1,22 @@
-import "./components/app.scss";
-
-import React from "react";
 import * as Mobx from "mobx";
 import * as MobxReact from "mobx-react";
-import * as LensExtensions from "../extensions/extension-api";
-import { App } from "./components/app";
-import { LensApp } from "./lens-app";
+import React from "react";
 import { render, unmountComponentAtNode } from "react-dom";
-import { isMac } from "../common/vars";
-import { userStore } from "../common/user-store";
-import { workspaceStore } from "../common/workspace-store";
 import { clusterStore } from "../common/cluster-store";
-import { i18nStore } from "./i18n";
-import { themeStore } from "./theme.store";
-import { extensionsStore } from "../extensions/extensions-store";
+import { userStore } from "../common/user-store";
+import { isMac } from "../common/vars";
+import { workspaceStore } from "../common/workspace-store";
+import * as LensExtensions from "../extensions/extension-api";
+import { extensionDiscovery } from "../extensions/extension-discovery";
 import { extensionLoader } from "../extensions/extension-loader";
+import { extensionsStore } from "../extensions/extensions-store";
 import { filesystemProvisionerStore } from "../main/extension-filesystem";
+import { App } from "./components/app";
+import "./components/app.scss";
+import { i18nStore } from "./i18n";
+import { LensApp } from "./lens-app";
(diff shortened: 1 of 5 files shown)


src/renderer/bootstrap.tsx
```diff
@@ -1,21 +1,22 @@
...
+import * as LensExtensions from "../extensions/extension-api";
+import { extensionDiscovery } from "../extensions/extension-discovery";
 import { extensionLoader } from "../extensions/extension-loader";
+import { extensionsStore } from "../extensions/extensions-store";
 import { filesystemProvisionerStore } from "../main/extension-filesystem";
+import { App } from "./components/app";
+import "./components/app.scss";
+import { i18nStore } from "./i18n";
+import { LensApp } from "./lens-app";
+import { themeStore } from "./theme.store";
```
Nokel81: Did you sort these? Because I think that one big block of imports doesn't make it easier to read. If we do sort them then I would prefer to keep the `scss` file at the top (since that is the style we seem to have in this repo, which is nice imo). and then have all the external deps in one block and then have all the other deps in another block.

src/renderer/bootstrap.tsx
```diff
@@ -34,6 +35,7 @@ export async function bootstrap(App: AppComponent) {
   rootElem.classList.toggle("is-mac", isMac);
 
   extensionLoader.init();
+  extensionDiscovery.init();
```
Nokel81, line 38: Maybe I am just weird, but why are we asking if this might be done in `main` since this in under `src/renderer`. Why not just call `initRenderer()` directly?

src/renderer/bootstrap.tsx
panuhorsmalahti: I always use TypeScript feature "Organize imports". We could enable eslint rule for consistency though.

src/renderer/bootstrap.tsx
panuhorsmalahti: Well, calling initRenderer could work, but that's an implementation detail I guess. Now the caller doesn't need to know that the class has different init method for different "environments", making it easier to call the initialization.

Nokel81: Fair I guess. I do know that I have always been slightly annoyed that we use the "same" classes for both `main` and `renderer` but I guess this is not the PR to do that work in.

src/renderer/bootstrap.tsx
Nokel81: I think an eslint rule would be good, but in a different PR.

aleksfront: I'm also upvoting for using organizing imports utilities but keeping scss and `React` imports at the top.

aleksfront (dismissed):
Works great!

src/renderer/bootstrap.tsx
panuhorsmalahti: Moved scss / React to the top.

Nokel81 (approved):

chenhunghan (approved):

PR #1639, opened 2020-12-03
Respect static list of allowed namespaces

### LINKED ISSUE #1634: Cluster accessible namespaces does not have any effect
**Describe the bug**

If user specifies "Accessible Namespaces" via cluster settings it does not have any effect what's displayed in Cluster dashboard.

**Expected behavior**

Cluster should show only specified namespaces.

jakolehm:
Fixes #1634

### DIFF
--- a/src/renderer/components/+namespaces/namespace.store.ts
+++ b/src/renderer/components/+namespaces/namespace.store.ts
@@ -5,6 +5,7 @@ import { Namespace, namespacesApi } from "../../api/endpoints";
 import { IQueryParams, navigation, setQueryParams } from "../../navigation";
 import { apiManager } from "../../api/api-manager";
 import { isAllowedResource } from "../../../common/rbac";
+import { getHostedCluster } from "../../../common/cluster-store";
 
 @autobind()
 export class NamespaceStore extends KubeObjectStore<Namespace> {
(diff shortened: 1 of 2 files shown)


nevalla (approved):

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -41,6 +42,17 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
     };
   }
 
+  subscribe(apis = [this.api]) {
+    const { allowedNamespaces } = getHostedCluster();
+
+    // if user has given static list of namespaces let's not start watches because watch adds stuff that's not wanted
+    if (allowedNamespaces.length > 0) {
+      return () => { return; };
```
panuhorsmalahti, line 50: Why "return;"?

Nokel81: We have a lint against empty functions.

Nokel81, line 50:
```suggestion
      return noop;
```

Nokel81 (approved):
Looks good, though I think we should mark that empty function explicitly as a desired noop

PR #1640, opened 2020-12-03
Performance fix when cluster has lots of namespaces

jakolehm:
Cluster object triggers unnecessarily one "can-i" request per namespace concurrently every 30 seconds. This behaviour is not needed anymore because Lens actually uses `allowedResources` for checking if user has access to different resources.

If a cluster has hundreds of namespaces this will cause huge load on Kubernetes API server.


panuhorsmalahti (dismissed):

nevalla (dismissed):

jakolehm:
Had to tweak unit tests. @nevalla @panuhorsmalahti ptal again.

panuhorsmalahti (approved):

nevalla (approved):

PR #1645, opened 2020-12-03
switch to using public mkdocs for check docs step

Nokel81:

### DIFF
--- a/.github/workflows/check-docs.yml
+++ b/.github/workflows/check-docs.yml
@@ -6,7 +6,7 @@ jobs:
   build:
     name: Check Docs
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/check-docs.yml
```diff
@@ -18,10 +18,9 @@ jobs:
       - name: Install dependencies
```
stevejr: @Nokel81 - we can remove this step as we don't need to install any dependencies

jakolehm (dismissed):

stevejr (approved):
LGTM

.github/workflows/check-docs.yml
Nokel81: I think you looked at a very old version.

jakolehm (approved):

PR #1646, opened 2020-12-03
Split workflow into verify and build jobs

stevejr:
Signed-off-by: Steve Richards <srichards@mirantis.com>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,12 +7,44 @@ on:
     types:
       - published
 jobs:
+  verify-docs:
+    name: Verify docs
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [12.x]
+    steps:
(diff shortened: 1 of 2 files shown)


jakolehm (dismissed):

.github/workflows/main.yml
```diff
@@ -7,12 +7,44 @@ on:
...
+    name: Verify docs
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [12.x]
+    steps:
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
```
Nokel81, line 20: Do we need this (given that I was able to remove it in the other workflow?

stevejr: Yes as we actually do run mike locally in this build which builds and deploys the docs using the insiders mkdocs version

Issue #1457, opened 2020-11-19
Cluster submenu items referencing same page are both highlighted


jim-docker:
```
  clusterPageMenus = [
    {
      id: "example",
      title: "Greetings",
      components: {
        Icon: ExampleIcon,
      }
    },
    {
      parentId: "example",
      target: { pageId: "hello" },
      title: "Hello World",
      components: {
        Icon: ExampleIcon,
      }
    },
    {
      parentId: "example",
      target: { pageId: "hello" },
      title: "Bonjour le monde",
      components: {
        Icon: ExempleIcon,
      }
    }
  ]
```
![image](https://user-images.githubusercontent.com/40840436/99737131-ac0da080-2a95-11eb-89bd-e40dc1e6a51d.png)

It's an odd thing to do so maybe this is ok? Although, theoretically one could specify the same page with different params to pass it thus yielding a different route.

jakolehm:
Currently this is "by design". Sidebar does "isActive" matching based on routes so if multiple menu items points to a same route then they will be highlighted. ~Should not be an issue until extension api allows to do cross-referencing.~

Nokel81:
If this is by design shall we close this issue then?

jim-docker:
> If this is by design shall we close this issue then?

Just tested on 4.1.4, and it is appears to be fixed anyway (must be now including params in the route determination?)

Nokel81:
Cool, thanks for the update.

jim-docker:
Actually, it may not be fixed (my test extension does it slightly differently) but there is an easy workaround, which is a more correct way to do it anyway.

PR #1651, opened 2020-12-03
Navigation refactoring, handling extension page params

### LINKED ISSUE #1399: Add params field to page registration (extension api)
As of 4.0.0-beta.3 if an extension developer wants to register a page that accepts parameters they must be added via route-path notation in the `id` field. Since navigating to a page with parameters is done by passing a TS object, like
```typescript
  extension.navigate(pageId: string?, params?: P)
```
or via the `target` field in a page menu registration
```typescript
      target: {
        pageId: "example/:exampleId",
        params: {
          exampleId: "456" // page-menu url: "/extension/lens-hello-world/example/456"
        }
      },
```
then it would be more natural to define a `params` object in the page registration, instead of via a route-path encoded in the `id` field.

Being able to specify the page and routing like this would free the developer from needing to know the route-path syntax (I presume most developers will only care about navigating to pages, optionally passing/receiving parameters, and don't care about the specific route path):
```typescript
  clusterPages: Interface.PageRegistration[] = [
    {
      id: "example",
      params: {
        exampleId: "0",
      }
      components: {
      }
    },
  ]
  clusterPageMenus: Interface.PageMenuRegistration[] = [
    {
      target: {
        pageId: "example",
        params: {
          exampleId: "456" // page-menu url: "/extension/lens-hello-world/example/456"
        }
      },
      title: "Example Extension",
      components: {
        Icon: ExampleIcon
      }
    }
  ]
}
```
`id` and `pageId` are simply ids when used in this context

This requires 
- adding an optional `params` field to the `PageRegistration` interface, through which each param can be defined and given a default value (perhaps an 'optional' flag?)
- adding logic under the hood to check for a `params` field and if one is specified then treat the `id` field as just an id and build the route-path onto it from the `params` field. If the `params` field is not present then treat the `id` field as (possibly) a route-path (current behaviour), for advanced users who need to control the route path specification.
- adding a method to `LensRendererExtension` to parse the url to retrieve the parameters so a page object can call it to easily get them (without needing to understand the route-path syntax) Maybe this already exists?

ixrock:
- Custom page params
- Detached bunch of unrelated url-params from `renderer/navigation.ts` with new helper `Navigation.createPageParam({name: string, defaultValue?: any, parse?(), stringify?(), ...})`
- Simplified page-registry, so it doesn't need to care about dynamic params part of final page URL
- [BREAKING]: `pageInit.id` stops supporting dynamic param placeholders, e.g. `/users/:user`
- [BREAKING]: `pageInit.exact` option removed

close #1399 

image

_Example from `extensions/example-extension/renderer.tsx`:_

```ts
  clusterPages: Interface.PageRegistration[] = [
    {
      components: {
        Page: (props: Interface.PageComponentProps<ExamplePageParams>) => {
          return <ExamplePage {...props} extension={this}/>;
        },
      },
      params: {
        // setup basic param "exampleId" with default value "demo"
        exampleId: "demo",

        // setup advanced multi-values param "selectedNamespaces" with custom parsing/stringification
        selectedNamespaces: {
          defaultValueStringified: ["default", "kube-system"],
          multiValues: true,
          parse(values: string[]) { // from URL
            return values.map(name => namespaceStore.getByName(name)).filter(Boolean);
          },
          stringify(values: K8sApi.Namespace[]) { // to URL
            return values.map(namespace => namespace.getName());
          },
        }
      }
    }
  ];

  clusterPageMenus: Interface.PageMenuRegistration[] = [
    {
      title: "Example extension",
      components: {
        Icon: ExampleIcon,
      },
    },
  ];
}
```


src/extensions/cluster-feature.ts
```diff
@@ -44,23 +44,23 @@ export abstract class ClusterFeature {
...
-  abstract async install(cluster: Cluster): Promise<void>;
+  abstract install(cluster: Cluster): Promise<void>;
 
   /**
    * to be implemented in the derived class, this method is typically called by Lens when a user has indicated that this feature is to be upgraded. The implementation
    * of this method should upgrade the kubernetes resources already installed, if relevant to the feature
    * 
    * @param cluster the cluster that the feature is to be upgraded on
    */
-  abstract async upgrade(cluster: Cluster): Promise<void>;
```
ixrock, line 55: IDE / `tsc` complains.. or maybe some tsc versions messed up, I've tried to update to latest one at some point.

ixrock: image

src/extensions/registries/page-registry.ts
```diff
@@ -1,93 +1,95 @@
...
-export function getExtensionPageUrl<P extends object>({ extensionId, pageId = "", params }: PageMenuTarget<P>): string {
-  const extensionBaseUrl = compile(`/extension/:name`)({
-    name: sanitizeExtensionName(extensionId), // compile only with extension-id first and define base path
-  });
-  const extPageRoutePath = path.posix.join(extensionBaseUrl, pageId);
+export interface PageTarget<P = {}> {
+  extensionId?: string;
+  pageId?: string;
+  params?: Record<string, any | any[]> & P; // default target page params
+}
```
Nokel81: what is the point of having `any | any[]`?

Also, what is the point of having `& P` where P has no requirements to be an object. 

I think that the following would be better:
```suggestion
export interface PageTarget<P extends Record<string, any> = {}> {
  extensionId?: string;
  pageId?: string;
  params?: P;
}
```

src/extensions/registries/page-menu-registry.ts
```diff
@@ -27,9 +21,9 @@ export interface PageMenuComponents {
   Icon: React.ComponentType<IconProps>;
 }
 
-export class GlobalPageMenuRegistry extends BaseRegistry<PageMenuRegistration> {
+export class PageMenuRegistry<T extends PageMenuRegistration = any> extends BaseRegistry<T> {
```
Nokel81: I don't think we should set this to `any` because that would mean that `globalPageMenuRegistry` has type `PageMenuRegistry<any>` which is not what we want. I recommend either leaving it blank or making it default to `PageMenuRegistration`

src/extensions/registries/page-menu-registry.ts
```diff
@@ -43,33 +37,23 @@ export class GlobalPageMenuRegistry extends BaseRegistry<PageMenuRegistration> {
...
+export class ClusterPageMenuRegistry extends PageMenuRegistry<ClusterPageMenuRegistration> {
   getRootItems() {
     return this.getItems().filter((item) => !item.parentId);
   }
 
   getSubItems(parent: ClusterPageMenuRegistration) {
-    return this.getItems().filter((item) => item.parentId === parent.id && item.target.extensionId === parent.target.extensionId);
+    return this.getItems().filter((item) => {
+      return item.parentId === parent.id && item.target.extensionId === parent.target.extensionId;
+    });
```
Nokel81: Why the change? If you want to have them on separate lines then I would recommend the following instead:
```suggestion
    return this.getItems().filter((item) => (
      item.parentId === parent.id 
      && item.target.extensionId === parent.target.extensionId
    ));
```

src/extensions/registries/page-registry.ts
```diff
@@ -1,93 +1,95 @@
...
 import { BaseRegistry } from "./base-registry";
 import { LensExtension, sanitizeExtensionName } from "../lens-extension";
 import logger from "../../main/logger";
-import { rectify } from "../../common/utils";
 
 export interface PageRegistration {
   /**
-   * Page ID or additional route path to indicate uniqueness within current extension registered pages
-   * Might contain special url placeholders, e.g. "/users/:userId?" (? - marks as optional param)
+   * Page-id, part of of extension's page url, must be unique within same extension
```
Nokel81: The dash makes it look like it is one concept, which is slightly confusing when documenting something called "id".

```suggestion
   * Page's `id`, part of of extension's page url, must be unique within same extension
```

src/extensions/registries/__tests__/page-registry.test.ts
```diff
@@ -72,15 +72,15 @@ describe("globalPageRegistry", () => {
 
   describe("getByPageMenuTarget", () => {
```
Nokel81: The function name has changed
```suggestion
  describe("getByPageTarget", () => {
```

src/extensions/registries/__tests__/page-registry.test.ts
```diff
@@ -72,15 +72,15 @@ describe("globalPageRegistry", () => {
...
     it("matching to first registered page without id", () => {
-      const page = globalPageRegistry.getByPageMenuTarget({ extensionId: ext.name });
+      const page = globalPageRegistry.getByPageTarget({ extensionId: ext.name });
 
       expect(page.id).toEqual(undefined);
       expect(page.extensionId).toEqual(ext.name);
-      expect(page.routePath).toEqual(getExtensionPageUrl({ extensionId: ext.name }));
+      expect(page.url).toEqual(getExtensionPageUrl({ extensionId: ext.name }));
```
Nokel81, line 107: Is isn't really testing the implementation black-box style enough. Please write out the expected value (or better yes, test that function with a bunch of cases).

src/extensions/registries/page-registry.ts
```diff
@@ -1,93 +1,95 @@
...
+        withPrefix: false,
+      });
+    });
 
-  if (params) {
-    return compile(extPageRoutePath)(params); // might throw error when required params not passed
+    if (searchParams.length > 0) {
+      stringifiedParams = `?${searchParams.join("&")}`;
+    }
   }
```
Nokel81: suggestion:

```ts
const searchParams = _.map(page?.params, urlParam => (
  urlParam.toSearchString({
    value: targetParams[urlParam.name] ?? urlParam.getDefaultValue(),
    mergeGlobals: false,
    withPrefix: false,
  })
).join("&");
const stringifiedParams = searchParams ? `?${searchParams}` : "";
```

src/extensions/registries/page-registry.ts
```diff
@@ -1,93 +1,95 @@
...
+      const items = [pages].flat().map(page => this.registerPage(page, extension));
 
-    return super.add(registeredPages);
-  }
-
-  getUrl<P extends object>({ extensionId, id: pageId }: RegisteredPage, params?: P) {
-    return getExtensionPageUrl({ extensionId, pageId, params });
+      return super.add(items);
+    } catch (error) {
+      return Function; // no-op
```
Nokel81: why not `noop`? Then there wouldn't be need for the comment.
```suggestion
      return noop;
```

src/renderer/navigation/url-param.ts
```diff
@@ -0,0 +1,115 @@
...
+  public urlName: string;
+
+  constructor(private init: UrlParamInit<V>, private history: IObservableHistory) {
+    const { isSystem, name, skipEmpty = true } = init;
+
+    this.name = name;
+    this.init.skipEmpty = skipEmpty;
+
+    // prefixing to avoid collisions with extensions
+    this.urlName = `${isSystem ? UrlParam.SYSTEM_PREFIX : ""}${name}`;
```
Nokel81: these should be getters

src/renderer/navigation/url-param.ts
```diff
@@ -0,0 +1,115 @@
...
+  static SYSTEM_PREFIX = "lens-";
+
+  public name: string;
+  public urlName: string;
+
+  constructor(private init: UrlParamInit<V>, private history: IObservableHistory) {
+    const { isSystem, name, skipEmpty = true } = init;
+
+    this.name = name;
+    this.init.skipEmpty = skipEmpty;
```
Nokel81:
```suggestion
    this.init.skipEmpty ??= true;
```

src/renderer/components/table/table.tsx
```diff
@@ -1,19 +1,17 @@
 import "./table.scss";
 
 import React from "react";
+import lodash from "lodash";
```
Nokel81: Unless you have changed all the imports of this in the project, I think that we should try and be consistent.

```suggestion
import _ from "lodash";
```

src/renderer/components/select/select.tsx
```diff
@@ -72,7 +72,7 @@ export class Select extends React.Component<SelectProps> {
     const { autoConvertOptions, options } = this.props;
 
     if (autoConvertOptions && Array.isArray(options)) {
-      return options.map(opt => {
+      return (options as any[]).map(opt => {
```
Nokel81: why the need for this case, if the function above is `T | any` then anything should be accepted, right?

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -1,47 +1,46 @@
...
+      fireImmediately: true,
+      equals: comparer.identity,
     });
   }
 
-  getContextParams(): Partial<IQueryParams> {
-    return {
-      namespaces: this.contextNs
-    };
-  }
```
Nokel81: So this is a breaking change. We do export the `NamespaceStore`. This either has to be kept (since we are doing this release after 4.0.0) or this store has to be marked as a beta version.

Nokel81 (changes requested):
I have reviews some of these changes. I will continue the review tomorrow. But there are some things already that need to be changed.

src/extensions/registries/page-registry.ts
ixrock: Just for info that the value of param could be an array.

src/renderer/navigation/url-param.ts
ixrock: Doesn't really matter, probably better to mark with `readonly`

src/renderer/components/select/select.tsx
ixrock: It was `tsc` some complaining and not related to this PR (just dirty fix)

src/renderer/components/+namespaces/namespace.store.ts
ixrock: It's surprising to me that this part publishly exposed to extensions-api (which should not and if needed better to add some facade layer).

extensions/example-extension/page.tsx
```diff
@@ -15,16 +22,25 @@ export class ExamplePage extends React.Component<{ extension: LensRendererExtens
...
+
         <p>Hello from Example extension!</p>
         <p>File: <i>{__filename}</i></p>
-        <Component.Button accent label="Deactivate" onClick={this.deactivate}/>
+        <p>Location: <i>{location.href}</i></p>
+
+        <p className="url-params-demo flex column gaps">
+          <a onClick={() => exampleNameUrlParam.set("secret")}>Show secret button</a>
+          {exampleName === "secret" && (
+            <Component.Button accent label="Deactivate" onClick={this.deactivate}/>
```
jim-docker: Can't this be done without relying on page params? I thought a good example of using page params in an extension would be two cluster page menus with the same target page but setting different param values?

extensions/example-extension/renderer.tsx
```diff
@@ -9,13 +9,16 @@ export default class ExampleExtension extends LensRendererExtension {
       title: "Example Extension",
       components: {
         Page: () => <ExamplePage extension={this}/>,
-      }
+      },
+      params: [
+        exampleNameUrlParam,
+      ]
```
jim-docker: Isn't this specifying something the page already knows?

extensions/example-extension/renderer.tsx
```diff
@@ -9,13 +9,16 @@ export default class ExampleExtension extends LensRendererExtension {
...
     }
   ];
 
   clusterPageMenus = [
     {
-      target: { pageId: "example", params: {} },
+      target: { pageId: "example" },
```
jim-docker: And I thought this _was_ a good place to set specific param values. I'm confused.

src/extensions/cluster-feature.ts
panuhorsmalahti: It seems TypeScript has recently forbidden it. The implementation can still use async if wanted.

src/extensions/registries/page-menu-registry.ts
```diff
@@ -43,33 +37,23 @@ export class GlobalPageMenuRegistry extends BaseRegistry<PageMenuRegistration> {
...
   }
 
-  getByPage(page: RegisteredPage) {
-    return this.getItems().find((item) => item.target?.pageId == page.id && item.target?.extensionId === page.extensionId);
+  getByPage({ id: pageId, extensionId }: RegisteredPage) {
+    return this.getItems().find((item) => {
+      return item.target.pageId == pageId && item.target.extensionId === extensionId;
```
panuhorsmalahti: I don't think we need the return here either.

src/extensions/registries/page-registry.ts
```diff
@@ -1,93 +1,95 @@
...
-    name: sanitizeExtensionName(extensionId), // compile only with extension-id first and define base path
-  });
-  const extPageRoutePath = path.posix.join(extensionBaseUrl, pageId);
+export interface PageTarget<P = {}> {
+  extensionId?: string;
+  pageId?: string;
+  params?: Record<string, any | any[]> & P; // default target page params
+}
+
+export interface RegisteredPage extends PageRegistration {
```
panuhorsmalahti: Should this be named `ExtensionPage`?

src/extensions/registries/page-registry.ts
```diff
@@ -1,93 +1,95 @@
...
 
-  if (params) {
-    return compile(extPageRoutePath)(params); // might throw error when required params not passed
+    if (searchParams.length > 0) {
+      stringifiedParams = `?${searchParams.join("&")}`;
+    }
   }
 
-  return extPageRoutePath;
+  return path.posix.join("/extension", sanitizeExtensionName(extensionId), pageId, stringifiedParams);
```
panuhorsmalahti: Doesn't seem correct to use filepath join for urls.

src/renderer/components/table/table.tsx
panuhorsmalahti: Offtopic for this PR, but I think we should import individual functions from lodash (or even lodash sub packages, not sure if tree shaking works better with that).

src/extensions/cluster-feature.ts
```diff
@@ -44,23 +44,23 @@ export abstract class ClusterFeature {
    * 
    * @param cluster the cluster that the feature is to be installed on
    */
-  abstract async install(cluster: Cluster): Promise<void>;
+  abstract install(cluster: Cluster): Promise<void>;
```
jakolehm: Not related to PR topic, please remove.

src/extensions/cluster-feature.ts
jakolehm: Change not related to PR topic, please remove.

src/extensions/cluster-feature.ts
```diff
@@ -44,23 +44,23 @@ export abstract class ClusterFeature {
...
+  abstract upgrade(cluster: Cluster): Promise<void>;
 
   /**
    * to be implemented in the derived class, this method is typically called by Lens when a user has indicated that this feature is to be uninstalled. The implementation
    * of this method should uninstall kubernetes resources using the kubernetes api (K8sApi)
    * 
    * @param cluster the cluster that the feature is to be uninstalled from
    */
-  abstract async uninstall(cluster: Cluster): Promise<void>;
+  abstract uninstall(cluster: Cluster): Promise<void>;
```
jakolehm: Change not related to PR topic, please remove.

src/extensions/cluster-feature.ts
```diff
@@ -72,7 +72,7 @@ export abstract class ClusterFeature {
    * 
    * @return a promise, resolved with the updated ClusterFeatureStatus
    */
-  abstract async updateStatus(cluster: Cluster): Promise<ClusterFeatureStatus>;
+  abstract updateStatus(cluster: Cluster): Promise<ClusterFeatureStatus>;
```
jakolehm: Change not related to PR topic, please remove.

src/extensions/registries/base-registry.ts
```diff
@@ -13,7 +12,7 @@ export class BaseRegistry<T> {
   add(items: T | T[], ext?: LensExtension): () => void; // allow method overloading with required "ext"
   @action
   add(items: T | T[]) {
-    const itemArray = rectify(items);
+    const itemArray = [items].flat() as T[];
```
jakolehm, line 14: Why is this changed?

src/extensions/registries/page-registry.ts
```diff
@@ -1,93 +1,95 @@
...
-   * Without {exact:true} second page never matches since first page-id/route already includes partial route.
-   * @example const pages = [
-   *  {id: "/users", exact: true},
-   *  {id: "/users/:userId?"}
-   * ]
-   * Pro-tip: registering pages in opposite order will make same effect without "exact".
+   * Registered page params.
+   * Used to generate page url when provided in getExtensionPageUrl()-helper.
    */
-  exact?: boolean;
```
jakolehm, line 30: Isn't this a breaking change?

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -1,47 +1,46 @@
...
+  multiValues: true,
+  get defaultValue() {
+    return storage.get();
+  }
+});
+
 @autobind()
 export class NamespaceStore extends KubeObjectStore<Namespace> {
   api = namespacesApi;
-  contextNs = observable.array<string>();
```
jakolehm, line 13: Breaking change.

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -1,47 +1,46 @@
...
-
-  protected storage = createStorage<string[]>("context_ns", this.contextNs);
-
-  get initNamespaces() {
```
jakolehm, line 17: Breaking change.

src/extensions/cluster-feature.ts
panuhorsmalahti: I wonder why @ixrock IDE complains, shouldn't the IDE use the TypeScript compiler specified in package.json? Or did you override in your setting?

extensions/example-extension/page.tsx
ixrock: What? :) Didn't get you.. it's just an example.. What you're asking definitely possible!

extensions/example-extension/renderer.tsx
ixrock: No, it doesn't. It's locally defined url-param and it's passed here only to use defined default-values for this page (from menus, otherwise not required)

extensions/example-extension/renderer.tsx
ixrock: Yes, it's possible to define target params.
Side-note: 
- `page-menu-item` specify param _values_ in form `{[paramName]: any}`
- `page-registry` consume values with registered params

src/extensions/registries/page-menu-registry.ts
ixrock: Are we still talking about code-formatting? o_O

src/extensions/registries/page-registry.ts
ixrock: What is not correct? Yes if we provide some url-invalid `pageId` it will break url.. but it's not expected :)

src/extensions/registries/base-registry.ts
ixrock: Duplication in general for: `lodash#flatten`, `Array.prototype.flat`
Removed: `common/utils/rectify.ts`

src/extensions/registries/page-registry.ts
ixrock: Maybe, but not critical imho.

src/renderer/components/+namespaces/namespace.store.ts
ixrock: What? Why? o_O If it was exported to public API than it's bad..
Could be reverted to old names otherwise.. And exporting api needs some refactoring..

jakolehm: See: https://github.com/lensapp/lens/blob/master/src/extensions/renderer-api/k8s-api.ts#L58

ixrock:
Besides all technical details and breaking changes..  what do you think about the concept?

src/renderer/components/+namespaces/namespace.store.ts
ixrock: Namespace store should not be exported to public API.

jakolehm: Could you clarify why? It's already exported and extensions are going to need shared stores with Lens so that Lens does not end up doing multiple queries/watches to Kubernetes cluster.

src/extensions/registries/page-registry.ts
panuhorsmalahti: It's not safe for urls, see discussion: https://stackoverflow.com/questions/16301503/can-i-use-requirepath-join-to-safely-concatenate-urls

src/extensions/registries/page-menu-registry.ts
panuhorsmalahti: Code-formatting is valid to discuss in PRs (even if not the most important thing), but I'm mostly commenting since the formatting was changed for no apparent reason and I don't think it's an improvement.

But I do agree that eslint should handle these. Until we have more full eslint rules it's up to discussion in my opinion.

src/extensions/registries/page-registry.ts
ixrock: Ok, will be fixed (but this PR already contained actually `path.join` for that reason)

src/renderer/components/+namespaces/namespace.store.ts
ixrock: Same reason as you suggested to add facade for cluster-store and workloads-store..
and for avoiding introducing such breaking changes..

jakolehm: If this class must be changed so that it breaks the API then this PR needs to add a facade that keeps it compatible, right?

ixrock: I think we should now very carefully export everything to public API and if export required then provided some `facade`. Otherwise we're locking ourself more and more from internal apis changes.. That's a good start for legacy codebase :)

ixrock: > If this class must be changed so that it breaks the API then this PR needs to add a facade that keeps it compatible, right?

No, facade should be exported to public, so changing internal apis doesn't affect to public APIs.
This PR touches only Lens _internals_, not public API (only adds) (that's why it's surprising to me).

src/extensions/registries/page-registry.ts
nevalla: to be noted that when using `URL`, you need to provide base url in valid url format (ie http://...). Also it's not possible to join multiple paths, so in this case `path.posix.join` is IMHO valid.

src/renderer/components/+namespaces/namespace.store.ts
jakolehm: I meant: because there is no facade for this class then a PR that is going to break the API needs to add it. While doing so we probably will see if that's even possible to achieve. Anyway my original comment here was meant to indicate that this will change the public api, not only internals.

src/extensions/registries/page-registry.ts
panuhorsmalahti: Just because `URL` doesn't work, it doesn't mean `path.posix.join` is correct. It probably works for our case so far, but could cause problems later.

There seems to be popular package `url-join` to do this safely.

We can change all path.posix.join's in one go in another PR if that's wanted though.

src/renderer/components/table/table.tsx
ixrock: FYI: tree-shaking doesn't work with `webpack` as far as i know..

src/renderer/components/select/select.tsx
ixrock: this unrelated changes removed from this PR

src/extensions/cluster-feature.ts
ixrock: done

src/extensions/registries/__tests__/page-registry.test.ts
ixrock: Dunno. I haven't add any new tests to this PR.. In general ofc it need more coverage.

src/extensions/registries/page-menu-registry.ts
ixrock: Minor thing.. used in single place..

src/renderer/components/+namespaces/namespace.store.ts
ixrock: Ok, if we gonna continue export `namespace-store.ts` as-is it should have some facade.. or probably I will leave these interfaces as-is.. (just provided better names in general, no logic changed)

src/extensions/registries/page-registry.ts
Nokel81: Yeah but `any` implies that too. If you really want that then it could be `extends Record<string, any | any[]`

ixrock: Alright, not super important :D

src/extensions/registries/page-registry.ts
ixrock: Anyway.. This option could be supported if needed.

extensions/example-extension/page.tsx
jim-docker: Ok. I think we need an example that demonstrates a real use case

extensions/example-extension/renderer.tsx
jim-docker: Isn't this the same as not setting the `params` field at all?

ixrock: Yes, if they are don't have default values.

ixrock: Like in here:

```ts
export const exampleNameUrlParam = Navigation.createUrlParam<string>({
  name: "name",
  defaultValue: "demo",
});
```

extensions/example-extension/page.tsx
ixrock: > Can't this be done without relying on page params

btw maybe we're talking about different things? what needs to be done here?

ixrock: @jim-docker PTAL: 
- `extensions/example-extension/renderer.tsx` updated cluster-page-menus definitions
- `extensions/example-extension/page.tsx` - how registered url-param for page used in other cases

**To see example in action, add:**
`"example-extension"` to `lens.extensions` list in project root's `package.json`

**Screenshots:** 2 cluster menus targeting same page with different params. Menu items in both cases considered as `active` (bug or feature? @jakolehm wdyt?):
Screenshot 2020-12-04 at 21 56 48
Screenshot 2020-12-04 at 21 56 55

extensions/example-extension/renderer.tsx
ixrock: **UPD:** simplified default params definition for page-registration: 
```ts
params: {
  [name: string]: UrlParam | string /*default value*/ ;
}
``` 
So, if param is simple plain string then it's not required to use `createUrlParam` at all.

jim-docker: Can you call them `createPageParam` and `PageParam` (Page instead of Url)? Lens knows internally that the param is part of the Url but the extension developer doesn't (or doesn't need) to know it. Internally we can keep Url but whatever is exposed could be more generic.

extensions/example-extension/page.tsx
jim-docker: https://github.com/lensapp/lens/issues/1457

ixrock: @jim-docker can't get the point of #1457
In your example the menu produces 2 links with different texts but same URL-s. What's wrong there? :)
And as @jakolehm said it's "by design"? Or how it should behave actually?

extensions/example-extension/renderer.tsx
ixrock: Good idea, will update 👍

src/renderer/components/+namespaces/namespace.store.ts
ixrock: public apis reverted for the store

extensions/example-extension/page.tsx
ixrock: @jakolehm this is also related how we will decide to support `exact` option (see PR's `TODO`-list)

src/extensions/registries/page-registry.ts
```diff
@@ -1,93 +1,105 @@
...
+  extensionId: string;
+  url: string; // registered extension's page URL (without page params)
+}
+
+export function getExtensionPageUrl(target: PageTarget): string {
+  const { extensionId, pageId = "", params: targetParams = {} } = target;
+
+  const pagePath = ["/extension", sanitizeExtensionName(extensionId), pageId]
+    .filter(Boolean)
+    .join("/").replace(/\/+/g, "/").replace(/\/$/, ""); // normalize multi-slashes (e.g. coming from page.id)
```
Nokel81, line 56:
```suggestion
    .join("/")
    .replace(/\/+/g, "/")
    .replace(/\/$/, ""); // normalize multi-slashes (e.g. coming from page.id)
```

Nokel81 (dismissed):

src/extensions/registries/__tests__/page-registry.test.ts
```diff
@@ -17,6 +17,16 @@ describe("getPageUrl", () => {
...
       isEnabled: true
     });
+    globalPageRegistry.add({
+      id: "page-with-params",
+      components: {
+        Page: () => React.createElement("Page with params")
+      },
+      params: {
+        test1: "test1-default",
+        test2: "" // no default value, just declaration
```
Nokel81, line 27: I don't think that this should be required. If it isn't all good.

extensions/example-extension/page.tsx
jim-docker: I think it is the same root cause? The key is that to the user they are clearly different so why would they both be highlighted? You could use params to try to distinguish but #1457 shows that that may still not be enough...

extensions/example-extension/page.tsx
```diff
@@ -1,12 +1,19 @@
...
+import { Component, LensRendererExtension, Navigation } from "@k8slens/extensions";
 import React from "react";
+import path from "path";
+import { observer } from "mobx-react";
+import { CoffeeDoodle } from "react-open-doodles";
+
+export const exampleId = Navigation.createPageParam({
+  name: "exampleId",
+  defaultValue: "demo",
+});
```
stefcameron: I think we need a strong page-to-param association to guard against wires getting crossed as people use different params for different pages. Also, I think the name is counter-intuitive because it sounds like we're declaring parameters, but we're actually getting a way to access them:
```javascript
Navigation.getPageParam({
  pageId: 'example',
  name: 'exampleId',
  defaultValue: 'demo',
});
```
And then in `renderer.tsx`, you declare the params the page will use by just adding the `params` object to the page declaration, like:
```javascript
  globalPages = [
    {
      id: "example",
      components: {
      },
      params: {
        exampleId,
      }
    }
  ];
```

extensions/example-extension/renderer.tsx
```diff
@@ -9,17 +9,39 @@ export default class ExampleExtension extends LensRendererExtension {
       title: "Example Extension",
       components: {
         Page: () => <ExamplePage extension={this}/>,
+      },
+      params: {
+        // setup param "exampleId" with default value "demo"
+        // could be also {[paramName: string]: UrlParam} for advanced use-cases (custom parse/stringify)
+        exampleId: "demo"
       }
```
stefcameron: I find there are conflicting intentions here vs `Navigation.createPageParam(...)` in order to get the param because both places allow you to specify a default value.

So maybe it's cleaner to remove setting the default in `Navigation.getPageParam()` (what I think we should rename it to) and let the default be specified here, and then the way to _get_ the param's value becomes:
`const param = Navigation.getPageParam({ pageId: 'example', param: 'exampleId' });` and then the component can do `param.get()` when it renders to get the current value.

extensions/example-extension/renderer.tsx
```diff
@@ -9,17 +9,39 @@ export default class ExampleExtension extends LensRendererExtension {
       title: "Example Extension",
       components: {
         Page: () => <ExamplePage extension={this}/>,
```
stefcameron: I also think it'll be more intuitive to (also) provide param values here as a `props` object that can be passed into the component that gets rendered when the page is activated:
```javascript
components: {
  Page: (props) => <ExamplePage extension={this} exampleId={props.params.exampleId} />,
}
// or just
components: {
  Page: ({ params }) => <ExamplePage extension={this} params={params} />,
}
// or whatever the developer wants to do
```
This can be in addition to `Navigation.getPageParam()` if we want to keep both ways. IMO, this props way is more intuitive than `Navigation.getPageParam()`.

stefcameron (changes requested):
This is a good change, but I think the interface could be improved a bit.

extensions/example-extension/page.tsx
ixrock: #1457 example don't use any `params` , only `pageId` which leads both to the same page == both highlighted

src/extensions/registries/__tests__/page-registry.test.ts
ixrock: It should be required since it's declaration as well. Read as "what kind of params page is willing to accept and how to process them" (e.g. in menus with default values).

extensions/example-extension/renderer.tsx
ixrock: Nope, all good here. You can already define param like (so only passing those to `props.params` need to fix):
```
params: {
  exampleId: Navigation.createPageParam(...)
}
```
But renaming might be a good idea.

extensions/example-extension/page.tsx
ixrock: Probably it was just a wrong example for this usage.

extensions/example-extension/renderer.tsx
ixrock: right, and no reason to add `@observer` in that case

extensions/example-extension/renderer.tsx
stefcameron: 😮  Too many ways to use the same thing for different purposes... If you can do this, then we should probably keep it (and let it accept a defaultValue), but still also provide the reader, `Navigation.getPageParam(pageId, paramId)` to get the same object out of the thingamagig (another "store"?) that stores these parameters when you want to get the param instead of declare the param.

And this is optional, right? If I don't care about observability, I can just use a static default value?
```javascript
params: {
  exampleId: 'foo'
}
```

ixrock: This part of code from where? ^ What is optional? 
Declaring params is non-optional for page registration (if page need them it should know how to process them).

ixrock: ...and to avoid passing all unknown params to menu links (remember: extension's cross-linking in menu is possible with custom params)

ixrock: > And this is optional, right? If I don't care about observability, I can just use a static default value?

and if you don't care about param type value except `string` while you do `param.get()` (otherwise you can setup custom parsing/stringification logic, so could work with some real objects) or just use plain `string` for default values as in your code above.

stefcameron: I'm sorry I wasn't clear enough. What I meant to ask was if there are 3 supported ways of declaring params for a page:
```javascript
params: {
  foo: 'foo-default',
  bar: Navigator.createPageParam({ pageId: 'example', paramId: 'bar', defaultValue: 'bar-default' }),
  baz: ... // this other parsing/stringification thing you've mentioned, which I don't know about
}
```

(It feels a bit weird to use `Navigator.createPageParam()` here, almost too heavy-handed given that even if I don't use it, as in the 'foo' param, I can then still use `Navigator.getPageParam()` to get the same type of observable object for it. So I wouldn't be sad to see support for this dropped. I don't think @jim-docker would be sad either, based on the simplicity of his proposal in #1399)

And then to consume/receive those params, there are 2 ways:

First, in `renderer.tsx` as props from the page declaration (doesn't work yet, but you will add):
```javascript
//// renderer.tsx

...
clusterPages = [
    {
      id: "example",
      title: "Example Extension",
      components: {
        Page: ({ params }) => <ExamplePage extension={this} {...params} />,
      },
      params: {...} // the object from above
    }
]
```

Second, in `page.tsx` using the new `Navigation.getPageParam(pageId, paramId)` (doesn't exist yet, but you will hopefully add).
```javascript
//// page.tsx

const fooParam = Navigation.getPageParam('example', 'foo');

export function MyPage() {
  return <p>{fooParam.get()}</p>;
}

// or I could receive as props per my declaration example:
export function MyPage({ foo, bar, baz }) {
  return <p>{foo} {bar.get()} {baz???}</p>;
}
```

ixrock: Currently only 2 ways of declaring params for the page:

```js
params: {
 // basic: setup param-name and default value, converted anyway to `PageParam` object when provided to Page component in `props.params` == aka "normalized params"
  foo: 'foo-default',

 // advanced (via direct `PageParam` helper call): supports custom stringify/parse/multi-values/etc.
  bar: Navigator.createPageParam(...), 
}
```

stefcameron: Got it, so `bar` and `baz` are actually the same thing in my examples. I thought there was a third case. Thanks for clarifying.

package.json
```diff
@@ -235,12 +221,16 @@
...
+    "mobx-react": "^6.2.2",
     "mock-fs": "^4.12.0",
     "node-pty": "^0.9.0",
     "npm": "^6.14.8",
     "openid-client": "^3.15.2",
     "path-to-regexp": "^6.1.0",
     "proper-lockfile": "^4.1.1",
+    "react": "^17.0.1",
+    "react-dom": "^17.0.1",
+    "react-router": "^5.2.0",
```
ixrock, line 233: Moved to `dependencies` since used in both env-s (`main`/`rendrerer`, see: `page-registry.ts`)
NB: `react-router` required implicitly in runtime by `react` or `react-dom`

extensions/example-extension/renderer.tsx
```diff
@@ -1,25 +1,45 @@
...
+          defaultValueStringified: ["default", "kube-system"],
+          multiValues: true,
+          parse(values: string[]) { // from URL
+            return values.map(name => namespaceStore.getByName(name)).filter(Boolean);
+          },
+          stringify(values: K8sApi.Namespace[]) { // to URL
+            return values.map(namespace => namespace.getName());
+          },
+        }
       }
```
jim-docker, line 29: `params` is an optional field, which is good when the page doesn't support params,  but when I tried not specifying it, I got (at runtime):
```
TypeError
Cannot destructure property 'exampleId' of 'this.props.params' as it is undefined.
```

can there be some default behaviour when params is not specified such that `this.props.params` is not undefined when `ExamplePage` methods try to use it?

ixrock: Well, I guess this is intentional behaviour. If the extension during page registration didn't specify `params` why ~on earth~ it tries to get access to `params`? Otherwise could be provided with empty `{}`, but it won't help since it's not enough to avoid errors. Otherwise everything could be wrapped in `try-catch` :D

park-cheif5 (commented):
Let's change the world!! Make it a great day 😜

extensions/example-extension/renderer.tsx
jim-docker: Heh, yeah, I am experimenting. It's an extension dev error but it can reasonably happen because there is still some ambiguity as to what should get defined where. Wondering if we can catch such a situation in the implementation. Also trying to see if the example can be coded up to make things clearer/more robust...

ixrock: >  because there is still some ambiguity as to what should get defined where

for me everything pretty clear:
1. define `params` at page registration when needs
2. gets injected `props.params` to registered `Page` component
3. PROFIT!

All what needs to know is interface for `props.params`. Basically it's the same defined `params` object with same keys, but instead of config values you get instances of `PageParam` with that configs. It's getter/setter helper for everything what needs to "CRUD" that param.

jim-docker: Yes, I think you're right, the ambiguity is mostly in my head :) Because `PageParam` isn't imported directly it feels like it comes out of nowhere but that's the nature of interfaces. We'll have to carefully document `PageComponentProps` and `PageParams` for someone to understand how to use them to the full extent. 
Just need to try a few more examples...

extensions/example-extension/renderer.tsx
```diff
@@ -30,7 +30,7 @@ export default class ExampleExtension extends LensRendererExtension {
     }
   ];
 
-  clusterPageMenus: Interface.PageMenuRegistration[] = [
+  clusterPageMenus: Interface.ClusterPageMenuRegistration[] = [
```
jim-docker, line 33: this should be identified as a breaking change?

Nokel81: Probably

src/renderer/navigation/page-param.ts
```diff
@@ -0,0 +1,132 @@
+// Manage observable URL-param from document.location.search
+import { IObservableHistory } from "mobx-observable-history";
+
+export interface PageParamInit<V = any> {
+  name: string;
+  isSystem?: boolean;
```
jim-docker: This shouldn't be exposed in the extensions api

jim-docker (commented):
I think this PR breaks the foldout menu feature (using `id`/`parentId` fields of the `ClusterPageMenuRegistration` interface). I have examples that work in master but not in this branch.

extensions/example-extension/renderer.tsx
ixrock: Defining interfaces was totally optional, so 50/50

src/renderer/navigation/page-param.ts
ixrock: Why not? this interface is used in exporting helper `Navigation.createPageParam()` (more high-level util to work with url(page) params in other words).
Or do you mean we just should disallow developers to manage this flag?

extensions/example-extension/renderer.tsx
Nokel81: It is a breaking change since a name was changed, the easy solution is to do the following:

```typescript
/**
 * @deprecated use ClusterPageMenuRegistration
 */
export type PageMenuRegistration = ClusterPageMenuRegistration;
```

ixrock: Wait, it's not removed, `PageMenuRegistration` exists.. They are both exported:

```ts
export interface PageMenuRegistration {
  target?: PageTarget;
  title: React.ReactNode;
  components: PageMenuComponents;
}

export interface ClusterPageMenuRegistration extends PageMenuRegistration {
  id?: string;
  parentId?: string;
}
```

Nokel81: Ah I see, then it certainly isn't a breaking change. My bad.

src/renderer/navigation/page-param.ts
jim-docker: Yes, third-party devs shouldn't see this flag (just referring to the `isSystem` field)

ixrock:
> I think this PR breaks the foldout menu feature (using id/parentId fields of the ClusterPageMenuRegistration interface). I have examples that work in master but not in this branch.

@jim-docker could you please check this one more time? I've merged `master` and reverted most of the part for `sidebar` menu logic, so would possibly work now.

src/renderer/navigation/page-param.ts
ixrock: Fixed

extensions/example-extension/renderer.tsx
jim-docker: just by virtue of the fact that `id` and `parentId` are optional, right?

Nokel81: Yes basically

jim-docker:
> > I think this PR breaks the foldout menu feature (using id/parentId fields of the ClusterPageMenuRegistration interface). I have examples that work in master but not in this branch.
> 
> @jim-docker could you please check this one more time? I've merged `master` and reverted most of the part for `sidebar` menu logic, so would possibly work now.

Yes, works again 👍

jim-docker (dismissed):

src/renderer/navigation/index.ts
```diff
@@ -0,0 +1,8 @@
+// Navigation (renderer)
+
+import { bindEvents } from "./events";
+
+export * from "./history";
+export * from "./helpers";
+
+bindEvents();
```
jakolehm, line 8: Do we have to run this automatically on import?

jakolehm (dismissed):

src/renderer/navigation/index.ts
ixrock: No, it's called only once here.

jakolehm (approved):

aleksfront (approved):

PR #1677, opened 2020-12-07
Add check to extension file watch

panuhorsmalahti:
This can protect the file watch & extension installation starting in error under e.g. node_modules of a dependency. This avoids depending on chokidar "depth" parameter to guarantee it (and it doesn't seem like we can rely on that).

Somebody should test this on Windows.

### DIFF
--- a/src/extensions/extension-discovery.ts
+++ b/src/extensions/extension-discovery.ts
@@ -1,4 +1,4 @@
-import chokidar from "chokidar";
+import { watch } from "chokidar";
 import { ipcRenderer } from "electron";
 import { EventEmitter } from "events";
 import fs from "fs-extra";
@@ -138,7 +138,7 @@ export class ExtensionDiscovery {
     await this.whenLoaded;
 
     // chokidar works better than fs.watch
-    chokidar.watch(this.localFolderPath, {
+    watch(this.localFolderPath, {
       // For adding and removing symlinks to work, the depth has to be 1.
       depth: 1,
       // Try to wait until the file has been completely copied.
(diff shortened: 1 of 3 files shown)


src/extensions/extension-discovery.ts
```diff
@@ -156,9 +156,18 @@ export class ExtensionDiscovery {
...
 
   handleWatchFileAdd =  async (filePath: string) => {
-    if (path.basename(filePath) === manifestFilename) {
+    // e.g. "foo/package.json"
+    const relativePath = path.relative(this.localFolderPath, filePath);
+
+    // Converts "foo/package.json" to ["foo", "package.json"], where length of 2 implies
+    // that the added file is in a folder under local folder path.
+    // This safeguards against a file watch being triggered under a sub-directory which is not an extension.
+    const isUnderLocalFolderPath = relativePath.split(path.sep).length === 2;
```
Nokel81, line 165: are you positive that this is correct. Windows does accept `/` as a path separator even though the common one is `\`

panuhorsmalahti: I suspect that the paths from chokidar are using `\`, at least it's that way in the Windows CI pipeline. It would seem unusual to get `/` from that in some cases. I think `/` would be encountered only on some user input? But we're receiving file add events here.

Nokel81: Fair, that is probably good enough.

Nokel81 (approved):

jakolehm (approved):

PR #1688, opened 2020-12-08
Register cluster page component properly to a route

jakolehm:
~With observer `App.render()` is called multiple times which causes it to recreate extension cluster pages again and again. I'm not sure why `@observer` is needed here, everything seems to work better without it.~

Previously every observed reaction created a new function which triggered a new page component instance.

### DIFF
--- a/src/renderer/components/app.tsx
+++ b/src/renderer/components/app.tsx
@@ -1,4 +1,5 @@
 import React from "react";
+import { observer } from "mobx-react";
 import { Redirect, Route, Router, Switch } from "react-router";
 import { I18nProvider } from "@lingui/react";
 import { _i18n } from "../i18n";
@@ -48,6 +49,7 @@ import { nodesStore } from "./+nodes/nodes.store";
 import { podsStore } from "./+workloads-pods/pods.store";
 import { sum } from "lodash";
 
+@observer
 export class App extends React.Component {
   static async init() {
     const frameId = webFrame.routingId;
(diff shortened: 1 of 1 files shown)


src/renderer/components/app.tsx
```diff
@@ -49,7 +48,6 @@ import { nodesStore } from "./+nodes/nodes.store";
 import { podsStore } from "./+workloads-pods/pods.store";
 import { sum } from "lodash";
 
-@observer
```
ixrock: Removing this will break reactivity from `renderExtensionTabLayoutRoutes()` and `renderExtensionRoutes()`

ixrock: The rule of `@observer`: if some observables used in direct `render()` of component then it needs to used.

jakolehm: How to verify this?

jakolehm: Nevermind, I do see that routes won't get properly registered if an extension is enabled after app has started.

ixrock: Look at the component ~~as I did~~ to see if it's using _any_ observables :)

Maybe this is root of issues:
```
          const pageComponent = () => <page.components.Page/>;

          return <Route key={`extension-tab-layout-route-${index}`} path={page.url} component={pageComponent}/>;
```

With `const pageComponent = () => <page.components.Page/>;` in `render()` the component (arrow-function) re-created every time.

Should be like `<Route path={page.url} component={page.components.Page}/>`

jakolehm: Indeed that fixes the issue 🏆 Will push update asap.

jakolehm: Fixed.

panuhorsmalahti (approved):

nevalla (approved):

PR #1692, opened 2020-12-08
Query all objects using single api call if admin and namespace list is not overridden

jakolehm:
Fixes performance regression caused by #1634 .


nevalla (approved):
LGTM

Nokel81 (approved):

Issue #1684, opened 2020-12-07
Extension appears to be enabled but can't be found


jim-docker:
When installing an extension from tarball through the UI it appears to install successfully, and can be enabled but does not actually get loaded.

Root cause is actually the symlink points to a non-existent folder.

From the logs:
```
info: [EXTENSION-DISCOVERY] Added extension @mirantis/lens-extension-cc
info: [STORE]: SAVING /Users/jimehrismann/Library/Application Support/Lens/lens-extensions.json
[EXTENSIONS-LOADER]: can't load extension main at /Users/jimehrismann/Library/Application Support/Lens/node_modules/@mirantis/lens-extension-cc/dist/main.js: Error: Cannot find module '/Users/jimehrismann/Library/Application Support/Lens/node_modules/@mirantis/lens-extension-cc/dist/main.js'
```

`~/Library/Application Support/Lens/node_modules/@mirantis/lens-extension-cc` is a symlink pointing to `~/.k8slens/extensions/lens-extension-cc` but the extension is actually installed in `~/.k8slens/extensions/mirantis--lens-extension-cc`

Apart from this, it seems odd that the link is created in `~/Library/Application Support/Lens/node_modules/` and points to the files installed in `~/.k8slens/extensions/`. Shouldn't it be the other way around? The extension should actually be installed under `~/Library/Application Support/Lens/node_modules/` and a link to it be created in `~/.k8slens/extensions/`?

affects 4.0.0

panuhorsmalahti:
Can't reproduce just yet. In any case it's correct that the symlink is from `~/Library/Application Support/Lens/node_modules/` to `~/.k8slens/extensions/`.

panuhorsmalahti:
Reproduces on production build, not dev.

panuhorsmalahti:
This is what probably happened:
1. User copied extension to `/Users/<username>/.k8slens/extensions/lens-extension-cc`
2. User uninstalled extension from the UI (or just deleting the folder manually).
3. User installed extension from the Lens UI.

At this point `/Users/<username>/Library/Application Support/Lens/node_modules/@mirantis/lens-extension-cc` still points to the old `/Users/<username>/.k8slens/extensions/lens-extension-cc`.

I think the fix would be actually run npm uninstall (or something) to remove it from node_modules. It's already removed from package.json and `/Users/<username>/.k8slens/extensions`

panuhorsmalahti:
[...]

PR #1695, opened 2020-12-08
Remove broken symlink from node_modules on uninstall

panuhorsmalahti:
Partial fix for #1684.

We could improve this by also removing the symlink on install if the symlink exists. This PR will fix at least some of the cases, however. See issue for more details.

### DIFF
--- a/src/extensions/extension-discovery.ts
+++ b/src/extensions/extension-discovery.ts
@@ -197,7 +197,7 @@ export class ExtensionDiscovery {
 
         // The path to the manifest file is the lens extension id
         // Note that we need to use the symlinked path
-        const lensExtensionId = path.join(this.nodeModulesPath, extensionName, "package.json");
+        const lensExtensionId = path.join(this.nodeModulesPath, extensionName, manifestFilename);
 
         logger.info(`${logModule} removed extension ${extensionName}`);
         this.events.emit("remove", lensExtensionId as LensExtensionId);
@@ -208,12 +208,17 @@ export class ExtensionDiscovery {
   };
 
   /**
-   * Uninstalls extension by path.
+   * Uninstalls extension.
    * The application will detect the folder unlink and remove the extension from the UI automatically.
-   * @param absolutePath Path to the non-symlinked folder of the extension
+   * @param extension Extension to unistall.
    */
-  async uninstallExtension(absolutePath: string) {
-    logger.info(`${logModule} Uninstalling ${absolutePath}`);
+  async uninstallExtension({ absolutePath, manifest }: InstalledExtension) {
+    logger.info(`${logModule} Uninstalling ${manifest.name}`);
+
+    // remove the symlink under node_modules.
+    // If we don't remove the symlink, the uninstall would leave a non-working symlink,
+    // which wouldn't be fixed if the extension was reinstalled, causing the extension not to work.
+    await fs.remove(this.getInstalledPath(manifest.name));
 
     const exists = await fs.pathExists(absolutePath);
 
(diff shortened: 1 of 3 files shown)


src/extensions/extension-discovery.ts
```diff
@@ -208,12 +208,17 @@ export class ExtensionDiscovery {
...
    */
-  async uninstallExtension(absolutePath: string) {
-    logger.info(`${logModule} Uninstalling ${absolutePath}`);
+  async uninstallExtension({ absolutePath, manifest }: InstalledExtension) {
+    logger.info(`${logModule} Uninstalling ${manifest.name}`);
+
+    // remove the symlink under node_modules.
+    // If we don't remove the symlink, the uninstall would leave a non-working symlink,
+    // which wouldn't be fixed if the extension was reinstalled, causing the extension not to work.
+    await fs.remove(this.getInstalledPath(manifest.name));
```
jakolehm, line 221: Is there any possibility that symlink does not exist? Should we handle that case?

panuhorsmalahti: .remove does nothing if the path doesn't exist so this should be safe.

src/extensions/extension-discovery.ts
```diff
@@ -260,6 +265,22 @@ export class ExtensionDiscovery {
...
+    return path.join(this.nodeModulesPath, name);
+  }
+
+  /**
+   * Returns the symlinked path to the package.json,
+   * e.g. "/Users/<username>/Library/Application Support/Lens/node_modules/@publisher/extension/package.json"
+   */
+  protected getInstalledManifestPath(name: string) {
+    return path.join(this.getInstalledPath(name), manifestFilename);
+  }
```
Nokel81, line 282: It is true that I don't understand the whole of our extensions install process. However, I would like to know why we have to care about symlinks at all. Why not just "require" the package from `~/.k8slens/extensions` without needing symlinks?

panuhorsmalahti: Probably for making it easy to install the dependencies. Making a big change to the extensions npm stuff will be a separate task in any case.

Nokel81: of course

Nokel81 (approved):

jim-docker (commented):
LGTM but I'm going to give it some testing...

jim-docker:
This may be an artificial test situation but:
If I install the extension, verify it works, delete the installed files via 
```
rm -Rf ~/.k8slens/extensions/mirantis--lens-extension-cc
```
then go to Lens extension page and uninstall the extension, I observe 
1) a notification pops up saying "Error uninstalling..."
2) that the no-longer-valid link has been removed, as expected.
3) extension is removed from the list, as expected

However, if instead of clicking "Uninstall" after doing the `rm -Rf ...` I click "Disable", I observe 
1) Lens disables the extension, 
2) after a few seconds the extension disappears from the list (automatically uninstalled by watcher?)
3) but the link has **not** been removed.

jakolehm (approved):

panuhorsmalahti:
> This may be an artificial test situation but:
> If I install the extension, verify it works, delete the installed files via
> 
> ```
> rm -Rf ~/.k8slens/extensions/mirantis--lens-extension-cc
> ```
> 
> then go to Lens extension page and uninstall the extension, I observe
> 
> 1. a notification pops up saying "Error uninstalling..."
> 2. that the no-longer-valid link has been removed, as expected.
> 3. extension is removed from the list, as expected
> 
> However, if instead of clicking "Uninstall" after doing the `rm -Rf ...` I click "Disable", I observe
> 
> 1. Lens disables the extension,
> 2. after a few seconds the extension disappears from the list (automatically uninstalled by watcher?)
> 3. but the link has **not** been removed.

I believe that is working as intended. I can go through what is happening in more detail tomorrow. I think we can improve various error sitations with other PRs (I have few in mind).

PR #1698, opened 2020-12-08
Fix status brick alignment in pod-menu extension

pashevskii:
Just wrapped container name with span
Before the change:
image

After:
image


aleksfront (approved):
LGTM

Nokel81 (approved):

PR #1704, opened 2020-12-08
Full support for ReplicaSets

### LINKED ISSUE #373: Full support for ReplicaSets

List ReplicaSets under Workloads, generally provide all the features for ReplicaSets you do for StatefulSets

**Why is this needed**:

ReplicaSets are common intermediaries between Deployments and Pods.

3.4.0

- Kubernetes distribution: 1.15.11
- Desktop OS: macOS Catalina

vshakirova:
Hi guys. I moved ReplicaSet under the workload menu and added a ReplicaSet to the overview page. But I don't know how to arrange all those elements on the Overview page better.  
Also, I'd like to suggest the idea use `Pod["spec"]` in all resources (`template.spec` field) such as Deployments, StatefulSets, JobSpec, and e.t.c (please take a look at 3 screenshot).  It may help avoid code duplication. Of course, I'm not sure whether it is a good idea or not, please let me know. (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#podtemplatespec-v1-core)
And I updated some fields in `Pod.spec` according to this doc, if you don't mind.(https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go)  if I use `Pod[spec]` in replicaSet API it contains a lot of mandatory fields which are actually not required and it's uncomfortable for test cases.  

Screen Shot 2020-12-07 at 8 10 14 PM

Screen Shot 2020-12-07 at 8 09 58 PM

Screen Shot 2020-12-07 at 7 49 44 PM

Fixes https://github.com/lensapp/lens/issues/373

### DIFF
--- a/integration/__tests__/app.tests.ts
+++ b/integration/__tests__/app.tests.ts
@@ -274,6 +274,12 @@ describe("Lens integration tests", () => {
           expectedText: "Stateful Sets"
         },
         {
+          name: "ReplicaSets",
+          href: "replicasets",
+          expectedSelector: "h5.title",
+          expectedText: "Replica Sets"
+        },
+        {
           name: "Jobs",
           href: "jobs",
           expectedSelector: "h5.title",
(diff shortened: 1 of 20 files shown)


chenhunghan (approved):
Looks perfect for me, tested locally, maybe @aleksfront can take a look at the styling.

aleksfront (approved):
LGTM

aleksfront:
We should probably unify those scale dialogs (Deployment, ReplicaSet, StatefulSet) into one, but that's out of current scope.

PR #1718, opened 2020-12-09
Remove symlink on extension install and manual runtime uninstall

panuhorsmalahti:
* When extension is installed whether by copying or from the UI, any existing broken symlink is removed
* When extension is uninstalled manually by deleting the folder while app is running, broken symlink is removed
* I also removed the error if the extension doesn't exist on uninstall, as that should be a harmless error, and slightly improves https://github.com/lensapp/lens/pull/1695#issuecomment-740746986
* It would still be nice to handle following case:
    * User deletes extension manually from `/Users/<username>/.k8slens/extensions` when application is closed


Nokel81 (approved):

chenhunghan (approved):
LGTM

Issue #1690, opened 2020-12-08
Where is pod logs and terminal button


pydr:
**Describe the bug**
Dear lens develop team:
I updated my lens app to ver 4.0.0.
but I can't find pod logs and terminal button.

**Screenshots**
![download _1_.png](https://i.loli.net/2020/12/08/WpRiex9ubMvYUNC.png)

**Environment (please complete the following information):**
- Lens Version:
- OS: Ubuntu 18.04
- Installation method (e.g. snap or AppImage in Linux): snap

aabdullinov:
Hi, 
I have the same problem about pod logs on windows 10 desktop.
detailed screenshots in attachment.
![lens4 0 1](https://user-images.githubusercontent.com/73327504/101460333-a522d100-3963-11eb-8ccd-6227ddc8d865.PNG)
![lens4 0 2](https://user-images.githubusercontent.com/73327504/101460338-a653fe00-3963-11eb-931a-b1d5d33f7714.PNG)
![lens4 0 3](https://user-images.githubusercontent.com/73327504/101460339-a6ec9480-3963-11eb-9e8c-1373dc8564ed.PNG)
![lens4 0 4](https://user-images.githubusercontent.com/73327504/101460341-a7852b00-3963-11eb-8c30-9c07eeca9050.PNG)
![lens4 0 5](https://user-images.githubusercontent.com/73327504/101460344-a7852b00-3963-11eb-8234-34a6874bd9d5.PNG)

nevalla:
Logs and terminal buttons are now run as extensions and it seems that those are not loaded properly on your environment. If you open Developer tools (View / Toggle Developer Tools), do you see this line in the console:
```
info: [EXTENSION]: enabled lens-pod-menu@0.1.0
```

Or do you see any errors there?

pydr:
Thanks for your reply.
yes, I saw an error in the console:

![download _2_.png](https://i.loli.net/2020/12/08/Nwv9Vba74DJUxGe.png)

aabdullinov:
@nevalla
Thanks for your reply!
Yes, there are a lot of errors, and some of them is about "lens-pod-menu":

```
[...]

PR #1752, opened 2020-12-11
Display error dialog if extensions couldn't be loaded

panuhorsmalahti:
* Previously errors in npm install were ignored. This change adds error dialog on startup. The startup of the application will continue in any case. This change will hopefully indicate to users of #1690 failure state.
* Also capture npm install failure if the process exit code is not 0.
* Remove "--silent" parameter so we can read the error message.
* Change sync exists check to async

### DIFF
--- a/src/extensions/extension-discovery.ts
+++ b/src/extensions/extension-discovery.ts
@@ -266,18 +266,30 @@ export class ExtensionDiscovery {
 
     logger.info(`${logModule} loading extensions from ${extensionInstaller.extensionPackagesRoot}`);
 
-    if (fs.existsSync(path.join(extensionInstaller.extensionPackagesRoot, "package-lock.json"))) {
+    if (await fs.pathExists(path.join(extensionInstaller.extensionPackagesRoot, "package-lock.json"))) {
       await fs.remove(path.join(extensionInstaller.extensionPackagesRoot, "package-lock.json"));
     }
 
     try {
+      // Verify write access to static/extensions, which is needed for symlinking
       await fs.access(this.inTreeFolderPath, fs.constants.W_OK);
+
+      // Set bundled folder path to static/extensions
       this.bundledFolderPath = this.inTreeFolderPath;
     } catch {
-      // we need to copy in-tree extensions so that we can symlink them properly on "npm install"
+      // If there is error accessing static/extensions, we need to copy in-tree extensions so that we can symlink them properly on "npm install".
+      // The error can happen if there is read-only rights to static/extensions, which would fail symlinking.
+
+      // Remove e.g. /Users/<username>/Library/Application Support/LensDev/extensions
       await fs.remove(this.inTreeTargetPath);
+
+      // Create folder e.g. /Users/<username>/Library/Application Support/LensDev/extensions
       await fs.ensureDir(this.inTreeTargetPath);
+
+      // Copy static/extensions to e.g. /Users/<username>/Library/Application Support/LensDev/extensions
       await fs.copy(this.inTreeFolderPath, this.inTreeTargetPath);
+
+      // Set bundled folder path to e.g. /Users/<username>/Library/Application Support/LensDev/extensions
       this.bundledFolderPath = this.inTreeTargetPath;
     }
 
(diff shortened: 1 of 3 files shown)


panuhorsmalahti:
I'll also need to add process exit code handling to this PR. "error" event from the child process doesn't cover the process exiting with an error code.

chenhunghan (dismissed):

src/extensions/extension-discovery.ts
```diff
@@ -266,18 +266,30 @@ export class ExtensionDiscovery {
 
     logger.info(`${logModule} loading extensions from ${extensionInstaller.extensionPackagesRoot}`);
 
-    if (fs.existsSync(path.join(extensionInstaller.extensionPackagesRoot, "package-lock.json"))) {
+    if (await fs.pathExists(path.join(extensionInstaller.extensionPackagesRoot, "package-lock.json"))) {
       await fs.remove(path.join(extensionInstaller.extensionPackagesRoot, "package-lock.json"));
     }
```
Nokel81: Isn't this technically a race condition? Shouldn't it be something like the following:
```suggestion
    try {
      await fs.remove(path.join(extensionInstaller.extensionPackagesRoot, "package-lock.json"));
    } catch (error) {
      if (error.code !== "ENOENT") { // ignore file not found errors.
        throw error;
      }
    }
```

src/extensions/extension-discovery.ts
```diff
@@ -266,18 +266,30 @@ export class ExtensionDiscovery {
...
 
     try {
+      // Verify write access to static/extensions, which is needed for symlinking
       await fs.access(this.inTreeFolderPath, fs.constants.W_OK);
```
Nokel81, line 275: The docs for this specifically call out using `access` for cases like this as race conditions.

src/main/index.ts
```diff
@@ -94,17 +94,21 @@ app.on("ready", async () => {
...
+      extensionLoader.addExtension(extension);
+    });
+    extensionDiscovery.events.on("remove", (lensExtensionId: LensExtensionId) => {
+      extensionLoader.removeExtension(lensExtensionId);
+    });
 
-  extensionLoader.initExtensions(extensions);
+    extensionLoader.initExtensions(extensions);
+  } catch (error) {
+    dialog.showErrorBox("Lens Error", `Could not load extensions: ${error?.message || ""}`);
```
Nokel81: I don't think we should leave this blank if we don't have a message. The dangling `:` would make it look to me like something wasn't being handled correctly.

src/extensions/extension-discovery.ts
panuhorsmalahti: fs.remove doesn't actually throw if the path if missing, so we could just call that without the if. I didn't want to change the code too much, but I guess it's fine to change it that way.

src/extensions/extension-discovery.ts
panuhorsmalahti: I think our usage is a bit far from the case the docs explain. In our case if the app would _not_ have write access, but then 1ms later it magically would, we could _still_ use the alternative code path, which works but I suppose is slower. If you mean the case that we have write access now but then lose it, you're correct.

However, I think neither of these cases of the folder rights changing during runtime are happening in practice.

I guess the theoretical way to handle this is to run npm install, catch the error of "symlinking failed because of no write access" and then do the second path. No idea how possible that's in practice though.

Nokel81: The theoretical way would be the way I would want to do it, but yes. I don't know how to catch that case either.

src/main/index.ts
panuhorsmalahti: Well, we know (at this time) that the error we're receiving is a proper error with message. Other usage in this file will actually throw if message is missing.. But yeah I guess I'll change this if we want to be that defensive.

Nokel81 (dismissed):

Nokel81 (approved):

elizgarcia5 (approved):

elizgarcia5:
Ty

jakolehm (approved):

PR #1755, opened 2020-12-11
Catch HTTP Errors in case pod metrics resources not exist / access forbidden

msa0311:
If pod metrics cannot be loaded (not found or access forbidden) this results in a JS error. Its valid in this cases that API returns that error as metrics could just be not there or user does not have access.

Looks like its even causing display errors in the icon bar of the deployment. (Not tested with compiled binary)

Bildschirmfoto 2020-12-11 um 14 25 34

Signed-off-by: Mario Sarcher <msarcher@mirantis.com>

### DIFF
--- a/src/renderer/components/+workloads-pods/pods.store.ts
+++ b/src/renderer/components/+workloads-pods/pods.store.ts
@@ -23,9 +23,12 @@ export class PodsStore extends KubeObjectStore<Pod> {
   }
 
   async loadKubeMetrics(namespace?: string) {
-    const metrics = await podMetricsApi.list({ namespace });
-
-    this.kubeMetrics.replace(metrics);
+    try {
+      const metrics = await podMetricsApi.list({ namespace });
+      this.kubeMetrics.replace(metrics);
+    } catch (error) {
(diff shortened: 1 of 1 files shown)


src/renderer/components/+workloads-pods/pods.store.ts
```diff
@@ -23,9 +23,12 @@ export class PodsStore extends KubeObjectStore<Pod> {
...
 
   async loadKubeMetrics(namespace?: string) {
-    const metrics = await podMetricsApi.list({ namespace });
-
-    this.kubeMetrics.replace(metrics);
+    try {
+      const metrics = await podMetricsApi.list({ namespace });
+      this.kubeMetrics.replace(metrics);
+    } catch (error) {
+      console.log("catched", error);
```
Nokel81:
```suggestion
      console.error("loadKubeMetrics failed", error);
```

msa0311:
@Nokel81 thanks, added your change and squashed the commits

Nokel81 (approved):
LGTM, however you need to fix the lint

jim-docker (approved):

PR #1776, opened 2020-12-15
Fix getExtensionFileFolder() to use extension name

Nokel81:
An extension's ID is not a stable field between installs. I think that this should fix that.

### DIFF
--- a/src/extensions/lens-extension.ts
+++ b/src/extensions/lens-extension.ts
@@ -46,7 +46,7 @@ export class LensExtension {
    * folder name.
    */
   async getExtensionFileFolder(): Promise<string> {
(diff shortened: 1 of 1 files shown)


panuhorsmalahti:
Why is it not stable?

Nokel81:
I guess I figured that the install location of an extension is not guaranteed at all, so between reinstalls it might change.

jakolehm:
> I guess I figured that the install location of an extension is not guaranteed at all, so between reinstalls it might change.

Why does id change? Shouldn't it be stable?

Nokel81:
I guess it should be stable, unless the developer of an extension switches between symlinking it in and installing from NPM. But I think we shouldn't have to support this case.

PR #1742, opened 2020-12-10
Ensure only one app instance

jakolehm:
Otherwise Lens might have multiple app instances running (including multiple tray instances). See https://github.com/electron/electron/blob/master/docs/api/app.md#apprequestsingleinstancelock

PR #1727, opened 2020-12-10
Reworked Extension Guides

pauljwil:
Edited and reworked content in the Extension Guides Overview and the Generator and Main Extension guides.

Signed-off-by: Paul Williams <pawilliams@mirantis.com>

PR #1778, opened 2020-12-15
Release v4.0.3

jakolehm:
## Changes since v4.0.2

- Install in-tree extensions before others (#1782)
- Metrics-cluster-feature: bundle semver package (#1781)
- Reworked Extension Guides (#1727)
- Display error dialog if extensions couldn't be loaded (#1752)
- Ensure only one app instance (#1742)


panuhorsmalahti:
We should verify #1782 behavior wrt removing dependencies before release.

jakolehm:
> We should verify #1782 behavior wrt removing dependencies before release.

Not sure what do you mean by this.

nevalla (approved):

panuhorsmalahti (approved):
Extension loading could take a bit more time, but we can improve this later.

PR #1785, opened 2020-12-15
fix extension-filesystem being called in LensRendererExtension

Nokel81:

Without this the application throws an error if this function is called in the renderer thread.


panuhorsmalahti (approved):

nevalla (approved):
LGTM

stefcameron:
🎉

stefcameron:
@Nokel81 Thanks for fixing this! Will the generated `/Lens/extension_data/{uuid}` directory survive the extension's removal, or does it also get deleted when the extension is removed?

Nokel81:
Yes, it does not get removed.

PR #1792, opened 2020-12-17
Fix openshift distribution detector

jakolehm:
I think this got broken in some merge conflict resolving.. Openshift detect needs to happen before "custom" (because Openshift itself would be reported as custom).

### DIFF
--- a/src/main/cluster-detectors/distribution-detector.ts
+++ b/src/main/cluster-detectors/distribution-detector.ts
@@ -56,12 +56,12 @@ export class DistributionDetector extends BaseClusterDetector {
       return { value: "docker-desktop", accuracy: 80};
     }
 
-    if (this.isCustom()) {
-      return { value: "custom", accuracy: 10};
+    if (this.isCustom() && await this.isOpenshift()) {
+      return { value: "openshift", accuracy: 90};
     }
 
-    if (await this.isOpenshift()) {
-      return { value: "openshift", accuracy: 90};
+    if (this.isCustom()) {
+      return { value: "custom", accuracy: 10};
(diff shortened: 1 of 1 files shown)


nevalla (approved):

src/main/cluster-detectors/distribution-detector.ts
```diff
@@ -56,12 +56,12 @@ export class DistributionDetector extends BaseClusterDetector {
       return { value: "docker-desktop", accuracy: 80};
     }
 
-    if (this.isCustom()) {
-      return { value: "custom", accuracy: 10};
+    if (this.isCustom() && await this.isOpenshift()) {
```
Nokel81, line 59: Why checking both here?

jakolehm: It's a performance optimisation, we don't want to check Openshift (which is an API call) unless we know it's worth it.

Nokel81: Ah fair, would it then be better to do the following?

```typescript
if (this.isCustom()) {
  if (await this.isOpenshift()) {
    return { value: "openshift", accuracy: 90 };
  }

  return { value: "custom", accuracy: 10 };
}
```

jakolehm: I'm not sure if that's more readable.. involves more nesting.

Nokel81: okay, and the custom check is pretty easy anyway.

Nokel81 (approved):

PR #1796, opened 2020-12-17
Add support for LimitRange

### LINKED ISSUE #329: [Feature Request] LimitRanges and Describe Limits
I would like a section for the `LimitRange` kind. [View the docs here.](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/) As these are namespaced and all of the limits together apply to the namespace, it would be very nice to see the output of `kubectl describe limits -n default` on the namespace view. I imagine the best place for `LimitRange` kind would be directly below the `ResourceQuota` kind in the dashboard. 

**Why is this needed**:
No way to see the `LimitRange` kind or any of the information about them. 

**Environment you're Lens application is on:** 3.3.1

- Desktop OS: MacOSX

vshakirova:
Hi, lens team. I tried to create an appearance of this feature usable, but it seems that it is not the most effective way to do it because I placed `DrawItem` into `DrawItem`, not sure it is legal 🙂 
Need your advice. Thank you

Doc: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#limitrange-v1-core
This doc a bit outdated https://github.com/kubernetes/community/blob/master/contributors/design-proposals/resource-management/admission_control_limit_range.md
and I referred to https://github.com/kubernetes/kubernetes/blob/cea1d4e20b4a7886d8ff65f34c6d4f95efcb4742/plugin/pkg/admission/limitranger/admission_test.go

Screen Shot 2020-12-17 at 3 50 35 PM

Fixes https://github.com/lensapp/lens/issues/329

### DIFF
--- a/src/extensions/renderer-api/k8s-api.ts
+++ b/src/extensions/renderer-api/k8s-api.ts
@@ -14,6 +14,7 @@ export { ConfigMap, configMapApi } from "../../renderer/api/endpoints";
 export { Secret, secretsApi, ISecretRef } from "../../renderer/api/endpoints";
 export { ReplicaSet, replicaSetApi } from "../../renderer/api/endpoints";
 export { ResourceQuota, resourceQuotaApi } from "../../renderer/api/endpoints";
+export { LimitRange, limitRangeApi } from "../../renderer/api/endpoints";
 export { HorizontalPodAutoscaler, hpaApi } from "../../renderer/api/endpoints";
 export { PodDisruptionBudget, pdbApi } from "../../renderer/api/endpoints";
 export { Service, serviceApi } from "../../renderer/api/endpoints";
@@ -46,6 +47,7 @@ export type { ConfigMapsStore } from "../../renderer/components/+config-maps/con
 export type { SecretsStore } from "../../renderer/components/+config-secrets/secrets.store";
 export type { ReplicaSetStore } from "../../renderer/components/+workloads-replicasets/replicasets.store";
 export type { ResourceQuotasStore } from "../../renderer/components/+config-resource-quotas/resource-quotas.store";
+export type { LimitRangesStore } from "../../renderer/components/+config-limit-ranges/limit-ranges.store";
 export type { HPAStore } from "../../renderer/components/+config-autoscalers/hpa.store";
 export type { PodDisruptionBudgetsStore } from "../../renderer/components/+config-pod-disruption-budgets/pod-disruption-budgets.store";
 export type { ServiceStore } from "../../renderer/components/+network-services/services.store";
--- a/src/renderer/api/endpoints/limit-range.api.ts
+++ b/src/renderer/api/endpoints/limit-range.api.ts
@@ -1,0 +1,57 @@
+import { KubeObject } from "../kube-object";
+import { KubeApi } from "../kube-api";
+import { autobind } from "../../utils";
+
+export enum LimitType {
+  CONTAINER = "Container",
+  POD = "Pod",
+  PVC = "PersistentVolumeClaim",
+}
+
+export enum Resource {
+  MEMORY = "memory",
+  CPU = "cpu",
+  STORAGE = "storage",
+  EPHEMERAL_STORAGE = "ephemeral-storage",
+}
+
+export enum LimitPart {
+  MAX = "max",
+  MIN = "min",
+  DEFAULT = "default",
+  DEFAULT_REQUEST = "defaultRequest",
+  MAX_LIMIT_REQUEST_RATIO = "maxLimitRequestRatio",
+}
+
+type LimitRangeParts = Partial<Record<LimitPart, Record<string, string>>>;
+
+export interface LimitRangeItem extends LimitRangeParts {
+  type: string
+}
+
+@autobind()
+export class LimitRange extends KubeObject {
+  static kind = "LimitRange";
+  static namespaced = true;
+  static apiBase = "/api/v1/limitranges";
+
+  spec: {
+    limits: LimitRangeItem[];
+  };
+
+  getContainerLimits() {
+    return this.spec.limits.filter(limit => limit.type === LimitType.CONTAINER);
+  }
+
+  getPodLimits() {
+    return this.spec.limits.filter(limit => limit.type === LimitType.POD);
+  }
+
+  getPVCLimits() {
+    return this.spec.limits.filter(limit => limit.type === LimitType.PVC);
+  }
+}
+
+export const limitRangeApi = new KubeApi({
+  objectConstructor: LimitRange,
+});
--- a/src/renderer/components/+config-limit-ranges/limit-range-details.scss
+++ b/src/renderer/components/+config-limit-ranges/limit-range-details.scss
@@ -1,0 +1,12 @@
+.LimitRangeDetails {
+
+  .DrawerItem {
+    > .name {
+      font-weight: $font-weight-normal;
+      padding-left: 4px;
+    }
+    .DrawerItem {
+      padding-top: 4px;
+    }
+  }
+}
--- a/src/renderer/components/+config-limit-ranges/limit-range-details.tsx
+++ b/src/renderer/components/+config-limit-ranges/limit-range-details.tsx
@@ -1,0 +1,97 @@
+import "./limit-range-details.scss";
+
+import React from "react";
+import { observer } from "mobx-react";
+import { KubeObjectDetailsProps } from "../kube-object";
+import { LimitPart, LimitRange, LimitRangeItem, Resource } from "../../api/endpoints/limit-range.api";
+import { kubeObjectDetailRegistry } from "../../api/kube-object-detail-registry";
+import { KubeObjectMeta } from "../kube-object/kube-object-meta";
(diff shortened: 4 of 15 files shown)


src/extensions/renderer-api/k8s-api.ts
```diff
@@ -1,3 +1,4 @@
+
```
Nokel81:
```suggestion
```

src/renderer/components/+config-limit-ranges/limit-range-details.tsx
```diff
@@ -0,0 +1,134 @@
...
+import { ILimitRangeItem, LimitRange, Resource } from "../../api/endpoints/limit-range.api";
+import { kubeObjectDetailRegistry } from "../../api/kube-object-detail-registry";
+import { KubeObjectMeta } from "../kube-object/kube-object-meta";
+import { DrawerItem } from "../drawer/drawer-item";
+import { Badge } from "../badge";
+
+interface Props extends KubeObjectDetailsProps<LimitRange> {
+}
+
+function renderLimits(limit: ILimitRangeItem, resource: string): JSX.Element {
```
Nokel81:
```suggestion
function renderLimits(limit: ILimitRangeItem, resource: Resource): JSX.Element {
```

src/renderer/api/endpoints/limit-range.api.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export interface ILimitRangeItem {
+  type: string;
+  max?: { [key: string]: string };
+  min?: { [key: string]: string };
+  default?: { [key: string]: string };
+  defaultRequest?: { [key: string]: string };
+  maxLimitRequestRatio?: { [key: string]: string };
+}
```
Nokel81: Having thee limit names would be useful. Furthermore, interfaces don't need to be prefixed with `I`.

```suggestion
export enum LimitPart {
  MAX = "max",
  MIN = "min",
  DEFAULT = "default",
  DEFAULT_REQUEST = "defaultRequest",
  MAX_LIMIT_RATIO = "maxLimitRequestRatio",
}

type LimitRangeParts = Partial<Record<LimitPart, Record<string, string>>>;

export interface LimitRangeItem extends LimitRangeParts {
  type: string;
}
```

src/renderer/components/+config-limit-ranges/limit-range-details.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  return (
+    <div>
+      {limit.defaultRequest &&
+      Object.keys(limit.defaultRequest).filter(key => key === resource).map(resourceName => (
+        <Badge key={resourceName} label={`defaultRequest: ${limit.defaultRequest[resourceName]}`}/>
+      ))
+      }
```
Nokel81: It would be (imo) easier to read this if these were split out. It is also strange to see a filter select only one item and then map over it. Furthermore there is high amount of overlap between each of these limits. So factoring out the common items would be even better.

Something like the following:

```tsx
function renderLimit(limit: LimitRangeItem, part: LimitPart, resource: Resource) {
  const resourceLimit = limit[part]?.[resource];

  if (!resourceLimit) {
    return null;
  }

  return <Badge key={resource} label={`${part}: ${resourceLimit}`}/>;
}
```

And then this return could very easily be:
```
return (
  <>
    {renderLimit(limit, LimitPart.DEFAULT_REQUEST, resource)}
    {renderLimit(limit, LimitPart.DEFAULT, resource)}
    {renderLimit(limit, LimitPart.MAX_LIMIT_RATIO, resource)}
    {renderLimit(limit, LimitPart.MAX, resource)}
    {renderLimit(limit, LimitPart.MIN, resource)}
  </>
);
```

src/renderer/components/+config-limit-ranges/limit-range-details.scss
```diff
@@ -0,0 +1,20 @@
...
+      div:empty:after {
+        content: '—';
+        margin: 6px;
+      }
+    }
+  }
+
+  .firstChild {
+    padding-top: 4px;
+  }
```
Nokel81: This doesn't need to be a classname, there is the `:first-child` pseudoclass

src/renderer/components/+config-limit-ranges/limit-range-details.tsx
```diff
@@ -0,0 +1,134 @@
...
+          </DrawerItem>
+          <DrawerItem name={`Ephemeral Storage`}>
+            {
+              containerLimits.map(limit =>
+                renderLimits(limit, Resource.EPHEMERAL_STORAGE)
+              )
+            }
+          </DrawerItem>
+        </DrawerItem>
+        }
```
Nokel81: The same sort of factoring out could be done here too. A pointer:

1. The sub drawer items can be produced with from an array of `Resource`'s being passed in. Plus remember to add the `"key"` prop.

src/renderer/components/+config-limit-ranges/limit-ranges.route.ts
```diff
@@ -0,0 +1,12 @@
+import type { RouteProps } from "react-router";
+import { buildURL } from "../../../common/utils/buildUrl";
+
+
+export const limitRangesRoute: RouteProps = {
+  path: "/limitranges"
+};
+
+export interface ILimitRangeRouteParams {
```
Nokel81:
```suggestion
export interface LimitRangeRouteParams {
```

src/renderer/components/+config-limit-ranges/limit-ranges.tsx
```diff
@@ -0,0 +1,54 @@
...
+interface Props extends RouteComponentProps<ILimitRangeRouteParams> {
+}
+
+@observer
+export class LimitRanges extends React.Component<Props> {
+  render() {
+    return (
+      <>
+        <KubeObjectListLayout
+          className="LimitRanges" store={limitRangeStore}
```
Nokel81:
```suggestion
          className="LimitRanges"
          store={limitRangeStore}
```

src/renderer/components/+config-limit-ranges/limit-ranges.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+interface Props extends RouteComponentProps<ILimitRangeRouteParams> {
+}
+
+@observer
+export class LimitRanges extends React.Component<Props> {
+  render() {
+    return (
+      <>
```
Nokel81: No need for this since it just has a single child.

src/renderer/components/+config-limit-ranges/limit-range-details.tsx
vshakirova: Hi @Nokel81, thank you for your review. Now, the code looks more readable and cleaner.

src/renderer/components/+config-limit-ranges/limit-range-details.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { Badge } from "../badge";
+
+interface Props extends KubeObjectDetailsProps<LimitRange> {
+}
+
+function renderLimit(limit: LimitRangeItem, part: LimitPart, resource: Resource) {
+
+  const resourceLimit = limit[part]?.[resource];
+
+  if(!resourceLimit) {
```
nevalla:
```suggestion
  if (!resourceLimit) {
```

src/renderer/components/+config-limit-ranges/limit-ranges.route.ts
```diff
@@ -0,0 +1,12 @@
+import type { RouteProps } from "react-router";
+import { buildURL } from "../../../common/utils/buildUrl";
+
```
nevalla, line 3: Perhaps a redundant empty line?

src/renderer/components/+config-limit-ranges/limit-ranges.tsx
```diff
@@ -0,0 +1,53 @@
...
+        renderHeaderTitle={"Limit Ranges"}
+        renderTableHeader={[
+          {title: "Name", className: "name", sortBy: sortBy.name},
+          {className: "warning" },
+          {title: "Namespace", className: "namespace", sortBy: sortBy.namespace},
+          {title: "Age", className: "age", sortBy: sortBy.age},
+        ]}
+        renderTableContents={(limitRange: LimitRange) => [
+          limitRange.getName(),
+          <KubeObjectStatusIcon key="icon"  object={limitRange}/>,
```
nevalla:
```suggestion
          <KubeObjectStatusIcon key="icon" object={limitRange}/>,
```

nevalla (changes requested):
Generally looking good! Added couple minor suggestion about coding style.

src/renderer/components/+config-limit-ranges/limit-ranges.route.ts
vshakirova: thank you, @nevalla 
My bad, I've forgotten to format the code

nevalla (approved):

Nokel81 (approved):

aleksfront (approved):

PR #1799, opened 2020-12-17
Electron 9.4.0

jakolehm:
https://www.electronjs.org/releases/stable?version=9#9.4.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -335,7 +335,7 @@
     "css-element-queries": "^1.2.3",
     "css-loader": "^3.5.3",
     "dompurify": "^2.0.11",
-    "electron": "^9.3.5",
+    "electron": "^9.4.0",
     "electron-builder": "^22.7.0",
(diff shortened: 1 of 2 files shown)


Nokel81 (approved):

nevalla (approved):

jakolehm:
It seems that iframe does not load correctly always with this version. Maybe this version just reveals some race condition in Lens? 🤔

jakolehm:
> It seems that iframe does not load correctly always with this version. Maybe this version just reveals some race condition in Lens? 🤔

Could not reproduce anymore.

jakolehm:
> > It seems that iframe does not load correctly always with this version. Maybe this version just reveals some race condition in Lens? 🤔
> 
> Could not reproduce anymore.

And of course now I can after merge, a cluster dashboard does not open if it's automatically opened after the app starts. If I reload window then it starts to work.

Issue #1767, opened 2020-12-14
Kubernetes 1.20 has stopped propagating `selfLink`


nevalla:
**Describe the bug**

Kubernetes api server has stopped propagating `selfLink` (https://github.com/kubernetes/kubernetes/pull/94397) so Lens needs to use some other mechanism to detect/parse correct api for the resource. 

Related to #1711

nevalla:
To reproduce `minikube start --kubernetes-version=1.20.0`

tece81:
any idea how to solve this issue?
very interested!

tece81:
as a work-around for us end-users found this:
https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner/issues/25#issuecomment-742616668

ixrock:
@tece81 fix is coming! https://github.com/lensapp/lens/pull/1804

jakolehm:
Fixed in #1804

PR #1804, opened 2020-12-18
Generate metadata.selfLink if response does not have it

jakolehm:
Minimal fix for #1767 .

### DIFF
--- a/src/renderer/api/kube-api.ts
+++ b/src/renderer/api/kube-api.ts
@@ -267,8 +267,7 @@ export class KubeApi<T extends KubeObject = any> {
   }
 
   protected ensureObjectSelfLink(kubeObject: T): void {
-    if (!kubeObject.metadata?.selfLink) {
-      kubeObject.metadata ?? {};
+    if (!kubeObject.metadata.selfLink) {
       kubeObject.metadata.selfLink = this.generateSelfLink(kubeObject);
     }
   }
--- a/src/renderer/api/kube-api.ts
+++ b/src/renderer/api/kube-api.ts
@@ -79,6 +79,18 @@ export function forCluster<T extends KubeObject>(cluster: IKubeApiCluster, kubeC
   });
 }
 
+export function ensureObjectSelfLink(api: KubeApi, object: KubeJsonApiData) {
+  if (!object.metadata.selfLink) {
+    object.metadata.selfLink = createKubeApiURL({
+      apiPrefix: api.apiPrefix,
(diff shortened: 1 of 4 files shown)


src/renderer/api/kube-api.ts
```diff
@@ -256,11 +256,32 @@ export class KubeApi<T extends KubeObject = any> {
...
+      apiVersion: this.apiVersionWithGroup,
+      resource: this.apiResource,
+      namespace: this.isNamespaced ? kubeObject.getNs() : undefined,
+      name: kubeObject.getName(),
+    });
+  }
+
+  protected ensureObjectSelfLink(kubeObject: T): void {
+    if (!kubeObject.metadata?.selfLink) {
+      kubeObject.metadata ?? {};
```
ixrock: What does this line?

jakolehm: Oops, it's a bug.. fix coming.

src/renderer/kube-object.store.ts
```diff
@@ -195,10 +195,9 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
     const items = this.items.toJS();
 
     for (const {type, object} of this.eventsBuffer.clear()) {
-      const { uid, selfLink } = object.metadata;
-      const index = items.findIndex(item => item.getId() === uid);
+      const index = items.findIndex(item => item.getId() === object.metadata?.uid);
       const item = items[index];
-      const api = apiManager.getApi(selfLink);
+      const api = apiManager.getApiByKind(object.kind, object.apiVersion);
```
ixrock, line 200: Probably this required for new kube-objects coming from watch-apis? (since `watch-api` based on different transport the receiving kube objects coming raw/not fixed, right?)

```ts
const newItem = new api.objectConstructor(object);
api.ensureObjectSelfLink(newItem); // apply fix?
```

ixrock (dismissed):
LGTM

src/renderer/kube-object.store.ts
jakolehm: I think you are right, integration tests fail when an object comes from watch-api.

jakolehm:
@nevalla @ixrock PTAL

nevalla (approved):

src/renderer/kube-object.store.ts
jakolehm: Fixed.

ixrock (approved):

PR #1815, opened 2020-12-20
Fix extension loader race conditions

jakolehm:
Fixes following issues:

- discovery emitted "add" events on app boot -> caused multiple extension install runs
- watch started before bundled extensions were ready
- user installed extensions were re-installed on boot (now install happens only during add)
- discovery used single package.json which was mutated on the fly, now it uses package.json only for bundled extensions

### DIFF
--- a/src/extensions/extension-discovery.ts
+++ b/src/extensions/extension-discovery.ts
@@ -170,7 +170,7 @@ export class ExtensionDiscovery {
           await this.removeSymlinkByManifestPath(manifestPath);
 
           // Install dependencies for the new extension
-          await extensionInstaller.installPackage(extension.absolutePath);
+          await this.installPackage(extension.absolutePath);
 
           this.extensions.set(extension.id, extension);
           logger.info(`${logModule} Added extension ${extension.manifest.name}`);
@@ -367,6 +367,10 @@ export class ExtensionDiscovery {
     });
 
     return await extensionInstaller.installPackages(packageJsonPath, packagesJson);
+  }
+
+  async installPackage(name: string) {
+    return extensionInstaller.installPackage(name);
   }
 
   async loadBundledExtensions() {
--- a/src/extensions/extension-discovery.ts
+++ b/src/extensions/extension-discovery.ts
@@ -192,7 +192,7 @@ export class ExtensionDiscovery {
     const extensionFolderName = path.basename(filePath);
 
     if (path.relative(this.localFolderPath, filePath) === extensionFolderName) {
-      const extension = Array.from(this.extensions.values()).find((extension) => path.dirname(extension.absolutePath) === path.dirname(filePath));
+      const extension = Array.from(this.extensions.values()).find((extension) => extension.absolutePath === filePath);
 
       if (extension) {
         const extensionName = extension.manifest.name;
--- a/src/extensions/extension-discovery.ts
+++ b/src/extensions/extension-discovery.ts
@@ -346,9 +346,15 @@ export class ExtensionDiscovery {
   async ensureExtensions(): Promise<Map<LensExtensionId, InstalledExtension>> {
     const bundledExtensions = await this.loadBundledExtensions();
 
-    await this.installPackages(this.packageJsonPath, bundledExtensions);
+    await this.installBundledPackages(this.packageJsonPath, bundledExtensions);
 
     const userExtensions = await this.loadFromFolder(this.localFolderPath);
+
+    for (const extension of userExtensions) {
+      if (await fs.pathExists(extension.manifestPath) === false) {
+        await this.installPackage(extension.absolutePath);
+      }
+    }
     const extensions = bundledExtensions.concat(userExtensions);
 
     return this.extensions = new Map(extensions.map(extension => [extension.id, extension]));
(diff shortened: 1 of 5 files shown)


src/extensions/extension-discovery.ts
```diff
@@ -347,29 +337,40 @@ export class ExtensionDiscovery {
...
     }
   }
 
-  async loadExtensions(): Promise<Map<LensExtensionId, InstalledExtension>> {
+  async ensureExtensions(): Promise<Map<LensExtensionId, InstalledExtension>> {
     const bundledExtensions = await this.loadBundledExtensions();
 
-    await this.installPackages(); // install in-tree as a separate step
-    const localExtensions = await this.loadFromFolder(this.localFolderPath);
+    await this.installPackages(this.packageJsonPath, bundledExtensions);
```
panuhorsmalahti: We install only bundled extensions here, but when do we install user extensions?

src/extensions/extension-discovery.ts
```diff
@@ -347,29 +337,40 @@ export class ExtensionDiscovery {
...
-    return new Map(extensions.map(extension => [extension.id, extension]));
+    return this.extensions = new Map(extensions.map(extension => [extension.id, extension]));
   }
 
   /**
    * Write package.json to file system and install dependencies.
    */
-  installPackages() {
-    return extensionInstaller.installPackages(this.packageJsonPath, this.packagesJson);
+  async installPackages(packageJsonPath: string, extensions: InstalledExtension[]) {
```
panuhorsmalahti: This could cause problems, if we first install bundled extensions then user extensions, the second install will remove the bundled extensions (as we have only 1 shared package.json and node_modules at the moment). The caller should ensure that the installed extensions are additive.

src/main/index.ts
```diff
@@ -103,14 +103,16 @@ app.on("ready", async () => {
...
-
   extensionDiscovery.init();
   windowManager = WindowManager.getInstance<WindowManager>(proxyPort);
 
   // call after windowManager to see splash earlier
   try {
     const extensions = await extensionDiscovery.load();
 
+    // Start watching after bundled extensions are loaded
+    extensionDiscovery.watchExtensions();
```
panuhorsmalahti, line 114: I don't see how this is broken, watchExtensions contains `await this.whenLoaded;` which is set at the end of `load`

src/extensions/extension-discovery.ts
```diff
@@ -197,23 +192,19 @@ export class ExtensionDiscovery {
...
 
-        delete this.packagesJson.dependencies[extensionName];
-
-        // Reinstall dependencies to remove the extension from package.json
-        await this.installPackages();
-
         // The path to the manifest file is the lens extension id
         // Note that we need to use the symlinked path
-        const lensExtensionId = path.join(this.nodeModulesPath, extensionName, manifestFilename);
+        const lensExtensionId = extension.manifestPath;
```
panuhorsmalahti, line 205: Can we actually remove this variable and use `extension.id`?

src/extensions/extension-discovery.ts
jakolehm: We install them only when they are added.

src/extensions/extension-discovery.ts
jakolehm: We don't use this for installing user extensions.

src/extensions/extension-discovery.ts
jakolehm: Yes if they are the same thing.. I just tried to keep this compatible with current logic.

src/main/index.ts
jakolehm: We can probably revert this, I did separate these just to understand the current flow better.

src/extensions/extension-discovery.ts
panuhorsmalahti: They should be the same. In the old implementation the id was created manually because we didn't have the extension object itself.

MinKhent88 (approved):
#48320

MinKhent88 (dismissed):

nevalla (dismissed):

src/extensions/extension-discovery.ts
Nokel81: Maybe a change of function name then?

src/extensions/extension-discovery.ts
```diff
@@ -347,29 +337,40 @@ export class ExtensionDiscovery {
...
+  async ensureExtensions(): Promise<Map<LensExtensionId, InstalledExtension>> {
     const bundledExtensions = await this.loadBundledExtensions();
 
-    await this.installPackages(); // install in-tree as a separate step
-    const localExtensions = await this.loadFromFolder(this.localFolderPath);
+    await this.installPackages(this.packageJsonPath, bundledExtensions);
 
-    await this.installPackages();
-    const extensions = bundledExtensions.concat(localExtensions);
+    const userExtensions = await this.loadFromFolder(this.localFolderPath);
```
Nokel81, line 351: Would it be good, in a future PR, to remove the dependence on this folder so that we can no longer use explicit symlinks and instead have a list of "dev" extensions to just `require`?

jakolehm: Yes, that would be good improvement (but cannot be done in a patch release).

Nokel81: okay sounds good. Maybe 4.1?

jakolehm: 4.1 is already quite loaded but we can try.

panuhorsmalahti (approved):

nevalla (approved):

Issue #601, opened 2020-07-21
Allow to expand/resize column headers


spyro2000:
**What would you like to be added**:

Would love to expand individual columns to see the full pod/service names

**Why is this needed**:

Currently it's not possiblel to expand the column to see the full name of services/pods...

![image](https://user-images.githubusercontent.com/8479925/88028791-67024d80-cb39-11ea-8286-3d559b43ce73.png)

**Environment you are Lens application on:**

- Kubernetes distribution: (shouldn't matter)
- Desktop OS: Win10x64Prof

Thank you!

nevalla:
Also reported in #350

paul-theorem:
I'd like to see this as well. The relevant bits of my pods in argo jobs are cut off - can't see which pod is which.

dsgreene713:
any updates on when this feature might be available?

greedybro:
I'd like to see this implemented as well, since some columns are more relevant than others. Plus it would be amazing for readability on any size of screen.

jamiejackson:
This is the most-requested Lens feature, by a fair margin. https://github.com/lensapp/lens/issues?q=is%3Aissue+is%3Aopen+org%3Agithub+sort%3Areactions-%2B1

gilbrown123:
I would love to have it as well .

RSLukin:
same here, please add this

JackSinclairT:
Please please please please add this. The names of our nodes are quite big so being able to make 1 specific column bigger would help so much!!

johnfoxcloudreach:
yes, please

Be very much appreciated

Punkoivan:
2 years here, moving to atlassian-style issues, if you know what I mean...
I understand that this is open-source project, but as I can see it's already been 2 related PR (declined)
Along with https://github.com/lensapp/lens/issues/5444 I would say open source Lens is dead.

laurensramandt:
It's useless to put in any effort into ANY issue as long as #5444 isn't fixed.

twoTimesAgnew:
Would love to see this feature added! 
Currently, I need to hover over a pod name to see the important bit, the hash at the end.

zerog2k:
This is also an issue when trying to connect to eks clusters, as the default kubecontext naming for them is quite long (arns like `arn:aws:eks:us-west-2:ACCOUNTID:cluster/eks-cluster-foo-bar-1`, and they truncate the names, and do not have any hover tool tip.

darrenfurr:
[...]

PR #1823, opened 2020-12-21
Resizable columns

pashevskii:
Peek 2020-12-23 14-33

Warning! It is a WIP commit and implementation looks like a crutch. 

The commit adds the ability to resize columns in the table without broking flex layout. Related issue https://github.com/lensapp/lens/issues/601

What is working: 
 - flex layout on the initial state
 - resizing columns (excluding checkboxes)
 
 Problems:
 - <del>performance</del>
 - <del>columns state is not saved<del> [Fixed]
 - <del>sorting is broken<del> [Fixed]
 -<del> small UI issues like double-scrollbars </del>


pashevskii:
@aleksfront @Nokel81
Could you please take a look?
Unfortunately, I'm not sure that I followed a good way by adding resizing elements to the existing implementation (it is the simplest way, but it has a lot of restrictions and cons)
Probably we should replace table implementation with something like react-table, what do you think about?

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -109,6 +113,22 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
     disposeOnUnmount(this, [
       reaction(() => toJS(this.userSettings), settings => storage.set(settings)),
     ]);
+
+    if (props.isResizable && props.tableId) {
```
Nokel81: Like in your other table PR, I think this should be a `get canResize(): boolean`.

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -109,6 +113,22 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
+      const columnConfig = Object.entries(userStore.preferences.storeTableConfig).find((element)=> {return element[0] == props.tableId;})[1];
```
Nokel81:
```suggestion
      const columnConfig = userStore.preferences?.storeTableConfig?.[props.tableId] ?? {};
```

src/common/user-store.ts
```diff
@@ -28,6 +28,7 @@ export interface UserPreferences {
   downloadBinariesPath?: string;
   kubectlBinariesPath?: string;
   openAtLogin?: boolean;
+  storeTableConfig?: Object;
```
Nokel81: Using `Object` is a bad idea, and doesn't help with typing in TS.

```suggestion
  storeTableConfig?: Record<string, Record<string, number>>;
```

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -90,8 +94,8 @@ interface ItemListLayoutUserSettings {
 @observer
 export class ItemListLayout extends React.Component<ItemListLayoutProps> {
   static defaultProps = defaultProps as object;
-
   @observable isUnmounting = false;
+  @observable columnswidth = new Array<number>(this.props.renderTableHeader.length);
```
Nokel81: This being an array is brittle, especially since in the future we want to be able to support moving these columns around. This should be a map between the column IDs (`className` should work I think) and widths.

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -109,6 +113,22 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
+          const outerValue = value;
+          const columnIndex = props.renderTableHeader.findIndex((value) => {
+            return value.className == outerValue[0];});
+
+          if (columnIndex >= 0 && outerValue[1]) {
+            this.columnswidth[columnIndex] = Number(outerValue[1]);
+          }
+        });
+      }
+    }
```
Nokel81: If we switch to using an map (or object) then there is no need to do all this checking, we can just do `this.columnWidth[col.className]` or something similar.

```suggestion
      this.columnsWidth = columnConfig;
```

src/renderer/components/table/table-cell.tsx
```diff
@@ -15,6 +16,10 @@ export interface TableCellProps extends React.DOMAttributes<HTMLDivElement> {
   isChecked?: boolean; // mark checkbox as checked or not
   renderBoolean?: boolean; // show "true" or "false" for all of the children elements are "typeof boolean"
   sortBy?: TableSortBy; // column name, must be same as key in sortable object <Table sortable={}/>
+  style?: any;
```
Nokel81: This should not be overridden, `React.DOMAttributes<T>` has a perfectly good typing for this field.

src/renderer/components/table/table-cell.tsx
```diff
@@ -63,14 +68,30 @@ export class TableCell extends React.Component<TableCellProps> {
...
+    if (isResizable){
+      return (
+        <Resizable
+          as="div"
+          onClick={this.onClick}
+          className = {classNames}
+          enable={{top:false, right:true, bottom:false, left:false, topRight:false, bottomRight:false, bottomLeft:false, topLeft:false }}
+          onResize = {onResize}
+          onResizeStop = {onResizeStop}
+          {...cellProps}>
```
Nokel81:
```suggestion
          {...cellProps}
        >
```

src/renderer/components/table/table-cell.tsx
```diff
@@ -5,6 +5,7 @@ import React, { ReactNode } from "react";
 import { autobind, cssNames, displayBooleans } from "../../utils";
 import { Icon } from "../icon";
 import { Checkbox } from "../checkbox";
+import { Resizable, ResizeCallback } from "re-resizable";
```
Nokel81: We already have `ResizeAnchor` so I think we should use that.

src/renderer/components/table/table-cell.tsx
```diff
@@ -63,14 +68,30 @@ export class TableCell extends React.Component<TableCellProps> {
...
     });
     const content = displayBooleans(displayBoolean, title || children);
 
+    if (isResizable){
+      return (
+        <Resizable
+          as="div"
+          onClick={this.onClick}
+          className = {classNames}
+          enable={{top:false, right:true, bottom:false, left:false, topRight:false, bottomRight:false, bottomLeft:false, topLeft:false }}
```
Nokel81: I cannot currently try this out as I get an error, but it seems strange that we can only resize to the right.

src/renderer/components/item-object-list/item-list-layout.tsx
Nokel81: This (the current version) currently throws as error immediately.

Nokel81 (changes requested):

src/renderer/components/table/table-cell.tsx
pashevskii: What do you think about to refactor the existing "table" (grid of divs)  to something more flexible like [this](https://autodesk.github.io/react-base-table/docs/get-started)

Nokel81: I don't see how that particular example is "more flexible". If you look at the rendered DOM then you will see it too is a grid of divs.

src/renderer/components/table/table-cell.tsx
pashevskii: IMHO, Adding one control element is enough to manage the border between headers
I fixed an error (it was related to reading nullable props), could you please to check the feature, it should be working

src/renderer/components/table/table-cell.tsx
pashevskii: Yeah, it is, I mean It offers a more configurable and whole solution, it already contains a lot of features that we can use on the project. The current PR still contains some number of bugs that require global changes (like 2 scrollbars or getting calculated widths of columns), not sure that fixing these ones is easier than full refactoring of tables

src/renderer/components/table/table-cell.tsx
pashevskii: Do you mean [HTMLAttributes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0893371fea43bfdf1777b6d835424961ba0d1dbb/types/react/index.d.ts#L1765)? DOMAttributes doesn't contain this prop

Nokel81: Ah okay, then please change this to `React.CSSProperties`

jakolehm:
@pashevskii please resolve `ItemListLayout` conflicts so that we can move code reviews further.

pashevskii:
@jakolehm done!

src/common/user-store.ts
```diff
@@ -28,7 +28,8 @@ export interface UserPreferences {
   downloadBinariesPath?: string;
   kubectlBinariesPath?: string;
   openAtLogin?: boolean;
-  hiddenTableColumns?: Record<string, string[]>
```
Nokel81, line 31: Please rebase correctly, hiding tables was added in a different PR.

Nokel81 (changes requested):

src/common/user-store.ts
pashevskii: I've merged hidden columns and column sizes into storeTableConfig, every column in this config has 2 properties, visible and width. Also, I've added index column, now it is just for simplifying the search of the columns, but it could be useful for the future ( I would like to add the ability to change the order of columns)

Nokel81:
It is possible to resize the second last column weirdly:

Screen Shot 2021-03-10 at 9 58 02 AM

Also sometimes the grab handle doesn't seem to be in the expected spot:

https://user-images.githubusercontent.com/8225332/110649891-0ab37300-8188-11eb-90e0-9055b6c78ef3.mov

Finally, I know that this is might be out of scope for this PR and might be better for a future PR but the behaviour of the drag-handle around the warning column seems weird when no warnings exist. Would it be possible to specify fixed width columns?

pashevskii:
> It is possible to resize the second last column weirdly:

There is a similar PR #2146, does it make sense to support two PR at the same time? Probably we should close this one?

Nokel81:
Sounds good thanks for the initial PR though.

PR #1828, opened 2020-12-21
Fix: missing dependent types for Select and Slider components

### LINKED ISSUE #1824: Types for Select and Slider components are not fully exported
This is caused mostly because used types from external package haven't included during extensions-api types generation:

```ts
import ReactSelect, { ActionMeta, components, Props as ReactSelectProps, Styles } from "react-select"

export interface SelectProps<T = any> extends ReactSelectProps<T>, CreatableProps<T>
```

**Proof:**
Screenshot 2020-12-21 at 15 22 59
Screenshot 2020-12-21 at 15 23 29

ixrock:
close #1824

### DIFF
--- a/package.json
+++ b/package.json
@@ -290,7 +290,6 @@
     "@types/jsonpath": "^0.2.0",
     "@types/lodash": "^4.14.155",
     "@types/marked": "^0.7.4",
-    "@types/material-ui": "^0.21.7",
     "@types/md5-file": "^4.0.2",
     "@types/mini-css-extract-plugin": "^0.9.1",
     "@types/mock-fs": "^4.10.0",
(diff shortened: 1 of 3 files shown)


Nokel81 (approved):

package.json
```diff
@@ -290,7 +290,6 @@
     "@types/jsonpath": "^0.2.0",
     "@types/lodash": "^4.14.155",
     "@types/marked": "^0.7.4",
-    "@types/material-ui": "^0.21.7",
```
jakolehm, line 293: Not needed?

ixrock: Yes, this is not used. All relevant types inside package itself (`@material-ui/core`)

jakolehm (approved):

PR #1829, opened 2020-12-21
Workloads overview: don't block on store load

jakolehm:
- renders overview immediately (and updates graphs on-demand), no need to wait for all related stores to be fully loaded
- removes concurrent load of stores, this causes too much load on k8s api on big clusters

### DIFF
--- a/src/renderer/components/+workloads-overview/overview.tsx
+++ b/src/renderer/components/+workloads-overview/overview.tsx
@@ -60,10 +60,12 @@ export class WorkloadsOverview extends React.Component<Props> {
       stores.push(eventStore);
     }
 
+    const unsubscribeList: Array<() => void> = [];
+
     for (const store of stores) {
       await store.loadAll();
+      unsubscribeList.push(store.subscribe());
     }
-    const unsubscribeList = stores.map(store => store.subscribe());
 
     await when(() => this.isUnmounting);
     unsubscribeList.forEach(dispose => dispose());
--- a/src/renderer/components/+workloads-overview/overview.tsx
+++ b/src/renderer/components/+workloads-overview/overview.tsx
@@ -14,7 +14,6 @@ import { statefulSetStore } from "../+workloads-statefulsets/statefulset.store";
 import { replicaSetStore } from "../+workloads-replicasets/replicasets.store";
 import { jobStore } from "../+workloads-jobs/job.store";
 import { cronJobStore } from "../+workloads-cronjobs/cronjob.store";
-import { Spinner } from "../spinner";
 import { Events } from "../+events";
 import { KubeObjectStore } from "../../kube-object.store";
 import { isAllowedResource } from "../../../common/rbac";
@@ -24,7 +23,6 @@ interface Props extends RouteComponentProps<IWorkloadsOverviewRouteParams> {
 
 @observer
 export class WorkloadsOverview extends React.Component<Props> {
-  @observable isReady = false;
   @observable isUnmounting = false;
 
   async componentDidMount() {
@@ -61,9 +59,10 @@ export class WorkloadsOverview extends React.Component<Props> {
     if (isAllowedResource("events")) {
       stores.push(eventStore);
     }
-    this.isReady = stores.every(store => store.isLoaded);
-    await Promise.all(stores.map(store => store.loadAll()));
-    this.isReady = true;
+
+    for (const store of stores) {
+      await store.loadAll();
+    }
     const unsubscribeList = stores.map(store => store.subscribe());
 
     await when(() => this.isUnmounting);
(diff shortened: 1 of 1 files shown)


src/renderer/components/+workloads-overview/overview.tsx
```diff
@@ -61,9 +59,10 @@ export class WorkloadsOverview extends React.Component<Props> {
     if (isAllowedResource("events")) {
       stores.push(eventStore);
     }
-    this.isReady = stores.every(store => store.isLoaded);
-    await Promise.all(stores.map(store => store.loadAll()));
-    this.isReady = true;
+
+    for (const store of stores) {
+      await store.loadAll();
+    }
```
ixrock: This might lead to memory leaks and/or errors IMHO.
It creates sequential loading for all stores, so the code **after** for-cycle might be evaluated when user already has left current page (but stores still loading data). For big clusters could be real-case scenario?

ixrock: Or actually I might be wrong.. since unsubscribing would happen anyway:
```ts
await when(() => this.isUnmounting);
```

ixrock (dismissed):

src/renderer/components/+workloads-overview/overview.tsx
ixrock: But still for big clusters could be a traffic issue since it's continue to load all stores even if user left the page.

src/renderer/components/+workloads-overview/overview.tsx
```diff
@@ -24,7 +23,6 @@ interface Props extends RouteComponentProps<IWorkloadsOverviewRouteParams> {
 
 @observer
 export class WorkloadsOverview extends React.Component<Props> {
-  @observable isReady = false;
   @observable isUnmounting = false;
```
Nokel81, line 26: I think that this is the wrong way to have this, it should be a list of disposers which is called in the `componentWillUnmount` function.

It is useful to still have this variable but I don't think it should be observable since it is only read from `componentDidMount` and only written to in `componentWillUnmount`.

src/renderer/components/+workloads-overview/overview.tsx
```diff
@@ -61,9 +59,10 @@ export class WorkloadsOverview extends React.Component<Props> {
     if (isAllowedResource("events")) {
       stores.push(eventStore);
     }
```
Nokel81, line 61:
```suggestion
    if (isAllowedResource("events") && !this.isUnmounting) {
      await eventStore.loadAll()
      this.disposers.push(eventStore.subscribe());
    }
```

Though TBH I think that `subscribe` should `loadAll` if it hasn't (which in and of itself is also an optimization).

src/renderer/components/+workloads-overview/overview.tsx
```diff
@@ -74,11 +73,7 @@ export class WorkloadsOverview extends React.Component<Props> {
     this.isUnmounting = true;
   }
```
Nokel81, line 76:
```suggestion
    this.isUnmounting = true;
    this.disposers.forEach(disposer => disposer());
  }
```

src/renderer/components/+workloads-overview/overview.tsx
ixrock: The problem with `componentWillUnmount` that it might not work as expected cause of async `componentDidMount`.. So this is better solution in that case..

src/renderer/components/+workloads-overview/overview.tsx
jakolehm: Maybe but that's out of scope for this PR 🙂

src/renderer/components/+workloads-overview/overview.tsx
jakolehm: Probably not worse than the current implementation which triggers all requests in parallel?

src/renderer/components/+workloads-overview/overview.tsx
Nokel81: True, but I definitely believe that awaiting the load and pushing the subscribe in each of the `if`'s is within scope.

jakolehm: Ok, I somehow missed the point in "suggested change". I agree, would be better to do both loadAll & subscribe within if.

src/renderer/components/+workloads-overview/overview.tsx
ixrock: Yeah, for sure

jakolehm:
@nevalla @ixrock @aleksfront ptal

aleksfront (approved):
LGTM

nevalla (approved):

PR #1830, opened 2020-12-21
Optimise Cluster.getAllowedResources()

jakolehm:
Previous implementation caused unnecessary load on Kubernetes API on every refresh. This PR adds caching and also checks more than just first namespace.

### DIFF
--- a/src/main/cluster.ts
+++ b/src/main/cluster.ts
@@ -11,7 +11,7 @@ import { Kubectl } from "./kubectl";
 import { KubeconfigManager } from "./kubeconfig-manager";
 import { loadConfig } from "../common/kube-helpers";
 import request, { RequestPromiseOptions } from "request-promise-native";
-import { apiResources } from "../common/rbac";
+import { apiResources, KubeApiResource } from "../common/rbac";
 import logger from "./logger";
 import { VersionDetector } from "./cluster-detectors/version-detector";
 import { detectorRegistry } from "./cluster-detectors/detector-registry";
@@ -78,6 +78,7 @@ export class Cluster implements ClusterModel, ClusterState {
   protected kubeconfigManager: KubeconfigManager;
   protected eventDisposers: Function[] = [];
   protected activated = false;
+  private resourceAccessStatuses: Map<KubeApiResource, boolean> = new Map();
 
   whenInitialized = when(() => this.initialized);
   whenReady = when(() => this.ready);
@@ -379,6 +380,7 @@ export class Cluster implements ClusterModel, ClusterState {
     this.accessible = false;
     this.ready = false;
     this.activated = false;
+    this.resourceAccessStatuses.clear();
     this.pushState();
   }
 
@@ -483,6 +485,8 @@ export class Cluster implements ClusterModel, ClusterState {
       const versionData = await versionDetector.detect();
 
       this.metadata.version = versionData.value;
+
+      this.failureReason = null;
 
       return ClusterStatus.AccessGranted;
     } catch (error) {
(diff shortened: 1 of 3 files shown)


src/main/cluster.ts
```diff
@@ -643,17 +647,26 @@ export class Cluster implements ClusterModel, ClusterState {
...
-          namespace: this.allowedNamespaces[0]
-        }))
-      );
+
+      const resources = apiResources.filter((resource) => this.resourceAccessStatuses.get(resource) === undefined);
+
+      for (const apiResource of resources) {
+        for (const namespace of this.allowedNamespaces.slice(0, 10)) {
+          if (!this.resourceAccessStatuses.get(apiResource)) {
+            const result = await this.canI({
```
ixrock: Why not parallel loading? Won't hurt for first-time loading?

src/main/cluster.ts
```diff
@@ -643,17 +647,26 @@ export class Cluster implements ClusterModel, ClusterState {
...
+              resource: apiResource.resource,
+              group: apiResource.group,
+              verb: "list",
+              namespace
+            });
+
+            this.resourceAccessStatuses.set(apiResource, result);
```
Nokel81: This is quite broken, since it is "greedy" and is no better then assuming that this first namespace is the "namespace to rule all namespaces". The only benefit I can see from trying multiple namespaces in this fashion is being able to find resources that only exist in some namespaaces.

Nokel81 (dismissed):

src/main/cluster.ts
jakolehm: Just to not cause too many parallel request to k8s api, these requests might be heavy (depending on Kubernetes distro / config).

Nokel81: Can we specialize this? Namely, for the time being have MKE do it serially and others do it up to 10 requests at a time?

src/main/cluster.ts
jakolehm: I disagree. Going through multiple namespaces is better than checking just the first one.

src/main/cluster.ts
jakolehm: @Nokel81 no we can't really. I tested this on MKE but we don't really know how each Kubernetes cluster is configured.

nevalla:
If there are some changes in RBAC, does this actually refresh the allowed resources? Let's say some resource is not allowed -> `this.resourceAccessStatuses.set(apiResource, false);`. Next some one will add RBAC rule that allows user to see the given resource, but when refreshing allowed resources we are checking only those resources that are `undefined`. Or did I read this wrong?

jakolehm:
> If there are some changes in RBAC, does this actually refresh the allowed resources? Let's say some resource is not allowed -> `this.resourceAccessStatuses.set(apiResource, false);`. Next some one will add RBAC rule that allows user to see the given resource, but when refreshing allowed resources we are checking only those resources that are `undefined`. Or did I read this wrong?

You are correct, currently it only refreshes those if user disconnect/connect. It's a tradeoff. We could refresh all which returns falsey but then we keep hammering k8s api.

nevalla:
> You are correct, currently it only refreshes those if user disconnect/connect. It's a tradeoff. We could refresh all which returns falsey but then we keep hammering k8s api.

Okay, makes sense.

jakolehm:
@ixrock @Nokel81 @nevalla PTAL

nevalla (approved):
LGTM

Nokel81 (commented):
I still think that this should be redone every so often, even if that is once every half hour or one every hour. Also, it would be nice if the max concurrent requests was specialized on the cluster type (namely unknown or MKE being 5 and others being 10).

jakolehm:
> I still think that this should be redone every so often, even if that is once every half hour or one every hour. Also, it would be nice if the max concurrent requests was specialized on the cluster type (namely unknown or MKE being 5 and others being 10).

Maybe those can be done in follow-up PRs?

src/main/cluster.ts
```diff
@@ -643,17 +648,30 @@ export class Cluster implements ClusterModel, ClusterState {
...
       }
-      const resourceAccessStatuses = await Promise.all(
-        apiResources.map(apiResource => this.canI({
-          resource: apiResource.resource,
-          group: apiResource.group,
-          verb: "list",
-          namespace: this.allowedNamespaces[0]
-        }))
-      );
+      const resources = apiResources.filter((resource) => this.resourceAccessStatuses.get(resource) === undefined);
```
ixrock, line 651: Why to filter here if we check it below with `!this.resourceAccessStatuses.get(apiResource)`?

ixrock (approved):
LGTM

src/main/cluster.ts
jakolehm: This filter returns only resources which don't have any result. Later we check only falsey values.

Nokel81: I think that is probably needless complexity since the check below will just succeed quickly.

jakolehm: > I think that is probably needless complexity since the check below will just succeed quickly.

Agreed, it might be unnecessary optimisation but should not hurt either. We are probably going to improve this in follow-up PRs where we have new possibilities for cleanup.

Issue #1768, opened 2020-12-14
Prompt for auto-update


panuhorsmalahti:
Changes the auto-update to not be automatic.

* If a new version is found during the application runtime, the user is asked _once_ in a prompt, e.g. "New version available. Would you like to update"?
* Adds auto-update section to e.g. Preferences page, which displays a button to update if a version is available.

Fixes #1701, #1689 and the groundwork for #1601.

panuhorsmalahti:
This should be fairly straightforward to implement: https://www.electron.build/auto-update

Nokel81:
I am targeting this at 3.6.10 so that users don't have to not quit lens to stay at that version. Will then cherry pick it to master.

jakolehm:
Fixed in #1831.

Nokel81:
No, there is still a PR open for 4.1

jakolehm:
@Nokel81 ok, I'll change milestone.

Nokel81:
finished by #1941 (though I still need to forward port the UI changes from 3.6 to that PR)

PR #1831, opened 2020-12-21
Fix auto-update to use Lens notifications, and add update confirmation

Nokel81:
- Add settings in user preferences for auto-updating (default false) and
  for allowing pre-release versions (default false)

- Use in-Lens notifications instead of OS notifications as those were
  found to be flaky

- Add rudimentary main->renderer notification system.

a partial fix for #1768


src/renderer/bootstrap.tsx
```diff
@@ -30,6 +31,10 @@ export async function bootstrap(App: AppComponent) {
   // Register additional store listeners
   clusterStore.registerIpcListener();
 
+  if (process.isMainFrame) {
```
jakolehm: Should this be in `LensApp`?

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -194,6 +194,24 @@ export class Preferences extends React.Component {
...
+          <h2><Trans>Updates</Trans></h2>
+          <Checkbox
+            label={<Trans>Allow auto updates</Trans>}
+            value={preferences.allowAutoUpdates}
+            onChange={v => preferences.allowAutoUpdates = v}
+          />
+          <small className="hint">
+            <Trans>Allow Lens to auto update itself to the latest version. Lens checks on startup and then once a day.</Trans>
+          </small>
+          <Checkbox
```
jakolehm: Pre-release should be a separate PR. I'm not sure if we really want it.

src/renderer/theme.store.ts
```diff
@@ -29,12 +29,8 @@ export class ThemeStore {
     { id: "kontena-light", type: ThemeType.LIGHT },
   ];
 
-  @computed get activeThemeId() {
-    return userStore.preferences.colorTheme;
-  }
-
   @computed get activeTheme(): Theme {
-    const activeTheme = this.themes.find(theme => theme.id === this.activeThemeId) || this.themes[0];
+    const activeTheme = this.themes.find(theme => theme.id === userStore.preferences.colorTheme) || this.themes[0];
```
jakolehm, line 33: How are these related to this PR?

src/common/user-store.ts
```diff
@@ -59,6 +61,8 @@ export class UserStore extends BaseStore<UserStoreModel> {
     colorTheme: UserStore.defaultTheme,
     downloadMirror: "default",
     downloadKubectlBinaries: true,  // Download kubectl binaries matching cluster version
+    allowAutoUpdates: false,
```
jakolehm: Default should be `true`.

jakolehm (dismissed):

src/common/user-store.ts
Nokel81: Doesn't that defeat the purpose of this PR? Especially since the logic of this PR to auto update and quit immediately upon finding an update.

src/renderer/bootstrap.tsx
Nokel81: Good idea.

src/common/user-store.ts
jakolehm: Wait, I thought we should just show a notification and allow user to quit+update from there? Why would we want to quit immediately if update is found?

Nokel81: So if you look at the code, when this option is false then a notification with "Yes install now", "Yes install later", and "No" is displayed to the user. If a user wants to always install immediately then they would toggle this.

jakolehm: Ok, got it... name confused me. Could we then remove this option completely (I think notification is enough)?

Nokel81: Yeah sure, I thought that we wanted auto updates (meaning installing too) which is why I added this.

src/renderer/theme.store.ts
Nokel81: I did this because styling is broken if the `colorTheme` in preferences isn't valid. (And we changed the names in v4)

src/renderer/components/+preferences/preferences.tsx
Nokel81: I'll remove it then.

src/main/app-updater.ts
```diff
@@ -1,19 +1,170 @@
...
+              background: "green",
+              marginRight: "10px"
+            }
+          },
+          {
+            label: "No",
+            backchannel: noChannel,
+            accent: true
+          }
+        ],
```
ixrock: Styles and translatable texts better to keep in renderer components.

package.json
```diff
@@ -171,13 +172,15 @@
...
     "@types/node": "^12.12.45",
+    "@types/node-notifier": "^8.0.0",
     "@types/proper-lockfile": "^4.1.1",
     "@types/react-beautiful-dnd": "^13.0.0",
     "@types/tar": "^4.0.3",
     "array-move": "^3.0.0",
     "chalk": "^4.1.0",
     "conf": "^7.0.1",
     "crypto-js": "^4.0.0",
+    "dateformat": "^4.3.1",
```
ixrock: We have `moment` already for working with `Date` or what's the purpose to add this?

src/common/user-store.ts
```diff
@@ -2,7 +2,7 @@ import type { ThemeId } from "../renderer/theme.store";
 import { app, remote } from 'electron';
 import semver from "semver"
 import { readFile } from "fs-extra"
-import { action, observable, reaction, toJS } from "mobx";
+import { action, IReactionDisposer, observable, reaction, toJS } from "mobx";
```
ixrock: `IReactionDisposer` is non used?

src/main/app-updater.ts
```diff
@@ -1,19 +1,170 @@
...
+            timeout: 5000,
+          }]
+        })
+      } catch (error) {
+        logger.error("[UPDATE CHECKER]: notification failed", { error: String(error) })
+      }
+    })
+
+  async function helper() {
+    while (true) {
```
ixrock, line 132: `while(true){}` always looks scary :D Why this is needed?

src/common/user-store.ts
Nokel81: Thanks

src/main/app-updater.ts
Nokel81: Styles yes, I can move those. However, I don't see why we should have the text in the renderer since its source is `main`.

src/main/app-updater.ts
Nokel81: I don't want to use `setTimeout` as that doesn't allow for catching errors (easily, we could use promises again I guess). This is essentially the same as calling `setTimeout` multiple times but it is more explicit. The `delay` function makes this "long polling".

src/main/app-updater.ts
ixrock: It doesn't matter where the "source" of texts. It's a part of layout now (especially when you show notifications in HTML-block).
For `i18n` support it's easier to keep it in `.tsx` for extracting to dictionary.
But, nevermind, localization support will required quite a lot of changes anyway.

src/main/index.ts
```diff
@@ -75,6 +73,12 @@ async function main() {
 
   // create window manager and open app
   windowManager = new WindowManager(proxyPort);
+
+  /**
+   * This depends on:
+   * 1. windowManager: it will send IPC to the main window for notifications
+   */
+  startUpdateChecking();
```
ixrock: Maybe better to `startUpdateChecking(windowManager)` for explicit dependency and no need in extra comments?

src/renderer/components/notifications/notifications.store.tsx
```diff
@@ -19,6 +23,39 @@ export interface INotification {
...
+    return null;
+  }
+
+  return (
+    <>
+      <br />
+      <div className="flex row align-right box grow">
+        {buttons.map(({ backchannel, ...props}) => (
+          <Button {...props} onClick={() => {
+            console.log(`sending to ${backchannel}`)
```
ixrock: Forgotten `console.log`?

src/renderer/components/notifications/notifications.store.tsx
```diff
@@ -19,6 +23,39 @@ export interface INotification {
...
+  body: string;
+  buttons?: ({
+    backchannel: IpcChannel;
+  } & ButtonProps)[];
+  status: NotificationStatus;
+  timeout?: number;
+  closeChannel?: IpcChannel;
+}
+
+function renderButtons(id: IpcChannel, buttons?: MainNotification["buttons"]): React.ReactNode {
```
ixrock: Probably this should be react-component, like: `function RenderButtons(props: {id: IpcChannel, buttons?: MainNotification["buttons"]})`

src/renderer/components/notifications/notifications.store.tsx
```diff
@@ -27,6 +64,29 @@ export class NotificationsStore {
 
   protected autoHideTimers = new Map<IMessageId, number>();
 
+  registerIpcListener(): void {
+    logger.info(`[NOTIFICATION-STORE] start to listen for notifications requests from main`);
+    ipcRenderer.on(NotificationChannelAdd, (event, model: MainNotification) => {
+      console.log(model);
```
ixrock: console.log

ixrock (approved):
Otherwise LGTM :)

src/main/index.ts
Nokel81: Agreed, though I initially had it as a comment as it was more of a transitive dependency (my function used `broadcastIpc`).

package.json
```diff
@@ -196,6 +197,7 @@
     "mobx-observable-history": "^1.0.3",
     "mock-fs": "^4.12.0",
     "node-machine-id": "^1.1.12",
+    "node-notifier": "^9.0.0",
```
jakolehm: Where is this used?

Nokel81: Sorry, thought I removed all the packages I added while iterating. Will remove

package.json
```diff
@@ -195,6 +195,7 @@
     "mobx": "^5.15.5",
     "mobx-observable-history": "^1.0.3",
     "mock-fs": "^4.12.0",
+    "moment": "^2.26.0",
```
ixrock, line 201: Now, since it's required in both envs pls remove it from `devDependencies`

src/main/app-updater.ts
```diff
@@ -1,19 +1,144 @@
...
+import { ipcMain } from "electron";
+import { isDevelopment } from "../common/vars";
+import { SemVer } from "semver";
+import moment from "moment";
+import { WindowManager } from "./window-manager";
 
-export default class AppUpdater {
+function delay(duration: number): Promise<void> {
+  return new Promise(resolve => setTimeout(resolve, duration));
+}
```
ixrock: Probably should be moved to `common/utils/delay.ts`

src/main/app-updater.ts
```diff
@@ -1,19 +1,144 @@
...
+
+  async function helper() {
+    while (true) {
+      await autoUpdater.checkForUpdates();
+      await delay(interval);
+    }
   }
+
+  helper()
+    .catch(error => logger.error("[UPDATE CHECKER]: failed with an error", { error: String(error) }));
```
ixrock: So it seems we're interested only in first failure? And in case of error it will stop interval checking totally?

Nokel81: Yes that is the current behaviour, do you think we should try again if it throws an error?

jakolehm: I think we should retry on next interval. Is there any reason not to retry?

Nokel81: Nope, (and I thought this too haha so it currently has that change)

src/main/app-updater.ts
```diff
@@ -1,19 +1,141 @@
...
-      autoUpdater.checkForUpdatesAndNotify()
-    }, this.updateInterval)
+  autoUpdater.logger = logger;
 
-    return autoUpdater.checkForUpdatesAndNotify()
+  autoUpdater
+    .on("update-available", async (args: UpdateInfo) => {
+      try {
+        const releaseDate = moment(args.releaseDate);
+        const body = `Version ${args.version} was release on ${releaseDate.format("dddd, mmmm dS, yyyy")}.`;
```
jim-docker: ...was released on...

src/main/app-updater.ts
jim-docker: `while(true){}` is only scary when you expect execution to eventually break out of the loop

src/main/window-manager.ts
```diff
@@ -7,7 +7,7 @@ import { initMenu } from "./menu";
 import { tracker } from "../common/tracker";
 
 export class WindowManager {
-  protected mainView: BrowserWindow;
+  public readonly mainView: BrowserWindow;
```
jim-docker, line 10: Is WindowManager exported to extensions api? Just wondering if it is better in terms of api hardening to add a getter for `mainView` instead.

Nokel81: The generated JS (even for a getter) is no different. As is easily seen by what we do in some of our tests by casting to any (or just by using JS instead of TS).

If we really wanted to be hardened we would use the new JS `#name` private identifiers. But I don't know if we want to do that.

jim-docker: So it won't break a TS extension if one day we needed to change `mainView` field back to being `protected`?

Nokel81: I don't understand, if we have `protected _mainView` and a `public get mainView()`. Then that is completely identical to `public readonly`. Modulo what the field names are.

src/main/app-updater.ts
```diff
@@ -86,24 +87,24 @@ async function autoUpdateCheck(windowManager: WindowManager): Promise<void> {
...
   autoUpdater.logger = logger;
+  autoUpdater.autoDownload = false;
+  autoUpdater.autoInstallOnAppQuit = false;
 
   autoUpdater
     .on("update-available", async (args: UpdateInfo) => {
       try {
         const releaseDate = moment(args.releaseDate);
-        const body = `Version ${args.version} was release on ${releaseDate.format("dddd, mmmm dS, yyyy")}.`;
+        const body = `Version ${args.version} was release on ${releaseDate.format("dddd, MMMM Do, yyyy")}.`;
```
jim-docker:
```suggestion
        const body = `Version ${args.version} was released on ${releaseDate.format("dddd, MMMM Do, yyyy")}.`;
```

jim-docker (approved):

jim-docker (approved):

Nokel81:
And now you work?

edit: apparently no. Checking the logs it actually failed. But my script change wasn't outputted but it did "run" since the exit code is now "good"

PR #1854, opened 2020-12-23
remove array desctructor of Object.values, which is brittle

Nokel81:

### DIFF
--- a/src/renderer/components/+nodes/node-charts.tsx
+++ b/src/renderer/components/+nodes/node-charts.tsx
@@ -7,6 +7,7 @@ import { IResourceMetricsValue, ResourceMetricsContext } from "../resource-metri
 import { observer } from "mobx-react";
 import { ChartOptions, ChartPoint } from "chart.js";
 import { themeStore } from "../../theme.store";
+import { mapValues } from "lodash";
 
 type IContext = IResourceMetricsValue<Node, { metrics: IClusterMetrics }>;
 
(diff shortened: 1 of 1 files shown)


src/renderer/components/+nodes/node-charts.tsx
```diff
@@ -17,26 +18,26 @@ export const NodeCharts = observer(() => {
   const id = object.getId();
   const { chartCapacityColor } = themeStore.activeTheme.colors;
 
-  if (!metrics) return null;
-  if (isMetricsEmpty(metrics)) return <NoMetrics/>;
+  if (!metrics) {
+    return null;
+  }
 
-  const values = Object.values(metrics).map(metric =>
```
ixrock, line 21: Why it's brittle? And what the reason to use `mapValues` from `lodash`? 
I would consider to use `Object.entries(metrics)` for mapping instead of extra lodash-helper.

ixrock (dismissed):

src/renderer/components/+nodes/node-charts.tsx
Nokel81: It is brittle because it depends on the order that the fields were created. Which I don't believe that we guarantee anywhere.

As for using `mapValues`. It is much shorter, does what I want, and we are already using `lodash` so I don't think the argument about an extra helper holds water.

jim-docker (dismissed):
I agree with @ixrock about the extra lodash-helper. The more packages we depend on the more brittle our build becomes, plus it requires us to know more functions and their behaviours (from these packages).
Having said that lodash has been around a long time and as @Nokel81 said, we use it elsewhere already. But one day we may decide we need to scale back our dependencies...

good catch @Nokel81

jim-docker (approved):

ixrock (approved):

PR #1860, opened 2020-12-24
Add suspend/resume button for CronJobs

### LINKED ISSUE #1694: Add suspend/resume button for CronJobs
It would be nice to have a button to suspend / resume CronJobs.

**Why is this needed**:
For example if your CronJob is missbehaving you might want to have a fast, easy and convenient button to just suspend the execution without having to google for what label to patch.

- Kubernetes distribution: On prem (1.19)
- Desktop OS: Ubuntu Linux

vshakirova:
Hi, team. Here is just a small feature. 
`Bug`: I noticed that cronJob `Suspend` field not updated immediately on UI after clicking `Suspend` button. After some research, I found that `kube-watch-api` couldn't find an appropriate api. And console error message(picture 2) related to this issue. I also noticed that all resources that have complex apiVersion not updated on UI without refresh. After debugging, I saw that a server response contains a full (e.g.`batch/v1beta1`) field, but we compare it only with `apiVersion` field. My suggestion is https://github.com/lensapp/lens/compare/master...vshakirova:add-suspend-resume-cronjob-btn?expand=1#diff-bffc0ab32510dc127d9efc665b77c2cda1fc3733f980029a8fc5189e9240ff95R21. Maybe it is not a good solution but it helped me to fix console and UI ussies.
Screen Shot 2020-12-24 at 4 42 44 PM

Screen Shot 2020-12-24 at 4 40 26 PM

Fixes https://github.com/lensapp/lens/issues/1694

### DIFF
--- a/src/renderer/api/api-manager.ts
+++ b/src/renderer/api/api-manager.ts
@@ -18,7 +18,7 @@ export class ApiManager {
   }
 
   getApiByKind(kind: string, apiVersion: string) {
-    return Array.from(this.apis.values()).find((api) => api.kind === kind && api.apiVersion === apiVersion);
+    return Array.from(this.apis.values()).find((api) => api.kind === kind && api.apiVersionWithGroup === apiVersion);
   }
 
   registerApi(apiBase: string, api: KubeApi) {
--- a/src/renderer/api/endpoints/cron-job.api.ts
+++ b/src/renderer/api/endpoints/cron-job.api.ts
@@ -5,6 +5,38 @@ import { formatDuration } from "../../utils/formatDuration";
 import { autobind } from "../../utils";
 import { KubeApi } from "../kube-api";
 
+export class CronJobApi extends KubeApi<CronJob> {
+  suspend(params: { namespace: string; name: string }) {
+    return this.request.patch(this.getUrl(params), {
+      data: {
+        spec: {
+          suspend: true
+        }
+      }
+    },
+    {
+      headers: {
+        "content-type": "application/strategic-merge-patch+json"
+      }
+    });
+  }
+
+  resume(params: { namespace: string; name: string }) {
+    return this.request.patch(this.getUrl(params), {
+      data: {
+        spec: {
+          suspend: false
+        }
+      }
+    },
+    {
+      headers: {
+        "content-type": "application/strategic-merge-patch+json"
+      }
+    });
+  }
+}
+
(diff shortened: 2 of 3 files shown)


aleksfront:
@vshakirova Thanks for your contribution! Can you please remove i18n additions and dismiss all of the changes inside dictionary files? We're removing current internationalisation feature from Lens and there's no need anymore to use `_i18n` and `<Trans />` anymore. Also, this will fix tests.

vshakirova:
Hi, @aleksfront. Thank you for your review, I'll remove i18n and the <Trans> tag from all my opened PRs.

src/renderer/api/api-manager.ts
```diff
@@ -18,7 +18,7 @@ export class ApiManager {
   }
 
   getApiByKind(kind: string, apiVersion: string) {
-    return Array.from(this.apis.values()).find((api) => api.kind === kind && api.apiVersion === apiVersion);
+    return Array.from(this.apis.values()).find((api) => api.kind === kind && api.apiVersionWithGroup === apiVersion);
```
jakolehm, line 21: I think we need to backport this fix to 4.0.x series.

jakolehm (dismissed):

src/renderer/api/api-manager.ts
vshakirova: Hi @jakolehm. Could you please elaborate? Should I create a separate PR fox this fix?

jakolehm: Sorry for not being clear.. it was just a reminder for maintainers that we need to backport this fix 🙂 I think maintainers can do it because it needs to target different branch etc.

jakolehm (approved):

src/renderer/api/api-manager.ts
vshakirova: got it, thank you 🙂

src/renderer/api/endpoints/cron-job.api.ts
```diff
@@ -90,8 +122,12 @@ export class CronJob extends KubeObject {
 
     return day > daysInMonth[month - 1];
   }
+
+  isSuspend() {
```
ixrock, line 126:
```suggestion
get  isSuspend(): boolean {
```

ixrock (approved):
LGTM

PR #1873, opened 2020-12-28
Applying dark colors to dock regardless of active theme

### LINKED ISSUE #1724: Background color of log and shell windows

I would love to have the background color of the log (and maybe shell) window configurable. I'm using the light theme, but that gives the shell and log windows a white background color since version 4.

As we are using colored output on the logfiles for example, the white background makes it hard to read:

Screenshot 2020-12-09 at 20 39 39

In version 3, the background color was black, regardless of the theme you chose. I found that much easier to read.

**Why is this needed**:

Because not everybody is fan of the dark theme, but still wants to have the logfiles in a readable manner.

- Kubernetes distribution: Digital Ocean
- Desktop OS: macOS

aleksfront:
In order to keep good reading experience, all dock tab contents are dark now regardless of active theme type. Also, permanent dark theme applied to all ace editor instances in general.

add clusters
terminal
edit resource
dark logs

Fixes #1724 


jakolehm:
@aleksfront linter still complains.

jakolehm (approved):

nevalla (approved):

PR #1874, opened 2020-12-28
Remove lingui

jakolehm:
We haven't exposed translations to users and now that Lens has extensions we have to rethink how translations should be handled (if at all). Lingui makes build configs and development in general harder -> I'm proposing that we remove everything related to Lingui.

Also removes Babel as build dependency (only Lingui needed it).


src/renderer/components/+config-maps/config-map-details.tsx
```diff
@@ -43,7 +42,7 @@ export class ConfigMapDetails extends React.Component<Props> {
       await configMapsStore.update(configMap, { ...configMap, data: this.data.toJSON() });
       Notifications.ok(
         <p>
-          <Trans>ConfigMap <b>{configMap.getName()}</b> successfully updated.</Trans>
+          <>ConfigMap <b>{configMap.getName()}</b> successfully updated.</>
```
aleksfront, line 45: No need for `<>` wrapper here.

aleksfront (approved):
LGTM, just need to replace template string literal quotes to double quotes when no variable used inside string. For example, replace \`Select custom kubeconfig file\` to "Select custom kubeconfig file"

ixrock (approved):

src/renderer/components/+config-maps/config-map-details.tsx
jakolehm: I'll do this in a separate PR.

Issue #1775, opened 2020-12-14
cluster nodes stale


matti:
- create a new k3s cluster at 12:00:00
- add it to lens
- delete the cluster
- create cluster with the same name at 12:30:00
- lens shows master node Ready from 30minutes ago, not 20s ao

cmd+r fixes

jim-docker:
I reproduced this using a kind cluster. To Lens this probably just looks like a connection issue when the cluster disappears then reappears. Maybe we need to do something like test if the cluster id or uuid has changed on connection to catch this.

PR #1897, opened 2021-01-04
Reset stores before loading them

### LINKED ISSUE #1892: Possible to get stale data on resource lists
**Describe the bug**

It's still possible to get stale data on any resource listings.

**To Reproduce**

- open workloads -> pods
- open nodes
- open again workloads -> pods but this time k8s api throws an error
- listed data is old

**Expected behavior**

Maybe Lens should clear the old data before fetching new? This way user would not see a stale data and it would be pretty clear that fetch failed because spinner would keep spinning.

jakolehm:
Should fix #1892


jim-docker (approved):
I thought this should fix #1775 but it doesn't. Not sure how to test it otherwise

nevalla (approved):

PR #1908, opened 2021-01-07
Fix Windows tray icon click

jakolehm:
On Windows it's common that clicking a tray icon opens the app window. Currently click does nothing which feels like a bug.

### DIFF
--- a/src/main/tray.ts
+++ b/src/main/tray.ts
@@ -10,7 +10,7 @@ import { workspaceStore } from "../common/workspace-store";
 import { preferencesURL } from "../renderer/components/+preferences/preferences.route";
 import { clusterViewURL } from "../renderer/components/cluster-manager/cluster-view.route";
 import logger from "./logger";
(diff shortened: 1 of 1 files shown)


Nokel81:
I don't agree that this is always the case, take for instance the `Bluetooth` tray icon for windows 10. Clicking it brings up a menu which you then have to click something else to open the Bluetooth settings window.

jim-docker:
> I don't agree that this is always the case, take for instance the `Bluetooth` tray icon for windows 10. Clicking it brings up a menu which you then have to click something else to open the Bluetooth settings window.

It certainly varies what happens but it's probably better to do something than nothing

Nokel81:
Does the tray menu not get opened on Windows like it does on Linux and macOS?

jim-docker:
On Windows Lens tray icon right click brings up context menu, left click does nothing

Nokel81 (approved):

nevalla (approved):

PR #1914, opened 2021-01-07
Fixing Singleton typing to correctly return child class

Nokel81:
- Add distinction between `getInstance` and `getInstanceOrCreate` since it is not always possible to create an instance (since you might not know the correct arguments)

- Remove all the `export const *Store = *Store.getInstance<*Store>();` calls as it defeats the purpose of `Singleton`. Plus with the typing changes the appropriate `*Store.getInstance()` is "short enough".

- Special case the two extension export facades to not need to use `getInstanceOrCreate`. Plus since they are just facades it is always possible to create them.

- Singleton constructors have to be `public` to be used until https://github.com/microsoft/TypeScript/issues/30991 is fixed.

- To combat the above (at a runtime level only currently 😞) `Singleton`'s constructor now knows whether it is being called from `getInstanceOrCreate` and throws an error if you try and call `new <T extends Singleton>()` directly. This is a **_hack_**, but I think the typing benefits outweight this needing to be verified by code reviewers.

- Move some other types to be also `Singleton`'s: `ExtensionLoader`, `ExtensionDiscovery`, `ThemeStore`, ...

- Fixed dev-run always using the same port with electron inspect


docs/extensions/guides/stores.md
```diff
@@ -50,7 +50,7 @@ export class ExamplePreferencesStore extends Store.ExtensionStore<ExamplePrefere
   }
 }
 
-export const examplePreferencesStore = ExamplePreferencesStore.getInstance<ExamplePreferencesStore>();
+export const examplePreferencesStore = ExamplePreferencesStore.getInstanceOrCreate();
```
jakolehm: Do we want to introduce this kind of breaking change in a minor version? 🤔

jakolehm: Hmm... or is this PR actually backwards compatible?

Nokel81: I guess it is a breaking change because I seperated out when an instance of a class would be created. I did this so that the program doesn't get subtly broken if someone calls `resetInstance`. I guess in the general case we could call the constructor if it has not arguments declared.

Note: I removed the `export const clusterStore = ...`, `export const userStore = ...`, and others because those are inherantly not singletons at all. That is technically a breaking change. Several places I have had to fix some implicit assumptions about the order in which items are created because of this. But I think that this leaves the codebase in a better and more robust state.

Lastly, if we wanted to we could hold onto this until after 4.1 (since 4.1 includes the auto update confirmation dialog PR). And then immediatly release 5.0 since this would be allowed for a major version.

jakolehm: > Lastly, if we wanted to we could hold onto this until after 4.1 (since 4.1 includes the auto update confirmation dialog PR). And then immediatly release 5.0 since this would be allowed for a major version.

We are not planning to release 5.0 that fast. Any possibility to make this without breaking the extension API compatibility?

Nokel81: I think that undo the breaking change is to re-add the `export const clusterStore` et al. but mark them as deprecated. The other parts of this PR are not actual breaking the API, just when things get instantiated.

Nokel81: Actually, now that I think about it. I don't think we can really go about adding back the `export const clusterStore` et al. Because those would have to be `export const clusterStore = ClusterStore.getInstanceOrCreate();` which somewhat nullifies these changes (since imo it is preferable to have as few places as possible for the creation of these singletons)

Nokel81:
@jakolehm Can we see about getting this PR in too, it is also rather large. I didn't convert your new types to be singletons in this PR but I think they should be done in a follow up PR.

docs/extensions/guides/stores.md
jakolehm: PR now targets 5.0 so we can break the extension API compatibility.

extensions/kube-object-event-status/package-lock.json
```diff
@@ -6,7 +6,691 @@
   "dependencies": {
     "@k8slens/extensions": {
       "version": "file:../../src/extensions/npm/extensions",
-      "dev": true
+      "dev": true,
```
jakolehm: Why so big diff? 🤔  Different npm version used?

Nokel81: Oh bother these keep slipping in.

Nokel81: Might be that but also I am changing our package contents.

Nokel81:
@jakolehm would you be able to take a look at this again?

.eslintrc.js
```diff
@@ -79,6 +79,8 @@ module.exports = {
         sourceType: "module",
       },
       rules: {
+        "no-invalid-this": "off",
```
jakolehm, line 82: What's the reason for these changes?

jim-docker (dismissed):
acting as proxy for @jakolehm

nevalla (dismissed):

jim-docker (dismissed):
proxy rubber stamp

jim-docker (approved):
rubber stamp

src/common/utils/singleton.ts
```diff
@@ -5,25 +5,39 @@
...
+
+  constructor() {
+    if (Singleton.creating.length === 0) {
+      throw new TypeError("A singleton class must be created by getInstanceOrCreate()");
+    }
+  }
 
-  // todo: improve types inferring
-  static getInstance<T>(...args: ConstructorParameters<Constructor<T>>): T {
+  static getInstanceOrCreate<T, R extends any[]>(this: StaticThis<T, R>, ...args: R): T {
```
jakolehm, line 20: Not sure if `createInstance` would make more sense here.

jakolehm (approved):

.eslintrc.js
Nokel81: Do you mean the off or this rule, I found that while refactoring having this rule (the typescript version that is) is *very* useful (and ran into it while doing this PR).

src/common/utils/singleton.ts
Nokel81: Followup up PR as discussed.

src/renderer/bootstrap.tsx
```diff
@@ -50,17 +51,26 @@ export async function bootstrap(App: AppComponent) {
...
   // preload common stores
   await Promise.all([
     userStore.load(),
     hotbarStore.load(),
     clusterStore.load(),
     extensionsStore.load(),
-    filesystemProvisionerStore.load(),
+    filesystemStore.load(),
     themeStore.init(),
+    helmRepoManager.init(),
```
jakolehm, line 73: @Nokel81  was there a reason for this? Now renderer boot is very slow (blank screen visible for few seconds).

/cc @jim-docker

Nokel81: No, we can move that line to outside the array

PR #1916, opened 2021-01-08
Disable telemetry extension during integration tests

jakolehm:


Nokel81 (approved):
Seems a bit heavy handed 😆 but it works.

nevalla (approved):

Issue #1898, opened 2021-01-04
Watch api does not work for non-admins with lots of namespaces


jakolehm:
**Describe the bug**

Watch api does not work at all for non-admin users which have access to lots of namespaces (tested with ~200 namespaces). 

Caused by too big GET requests:

```
431 (Request Header Fields Too Large)
```

**To Reproduce**

- create > 200 namespaces
- create non-admin user which has access to all namespaces
- go to workloads -> pods

PR #1918, opened 2021-01-08
Load k8s resources only for selected namespaces

ixrock:
- loading k8s resources into stores only for context namespaces (selected by user)
- fix: context namespaces not shared between pages (#1897)


src/renderer/components/+apps-releases/release.store.ts
```diff
@@ -60,30 +60,24 @@ export class ReleaseStore extends ItemStore<HelmRelease> {
   @action
   async loadAll() {
     this.isLoading = true;
-    let items;
+    let items: HelmRelease[];
```
Nokel81: Since we don't access this outside the `try` block, we don't need this here.
```suggestion
```

src/renderer/components/+apps-releases/release.store.ts
```diff
@@ -60,30 +60,24 @@ export class ReleaseStore extends ItemStore<HelmRelease> {
...
-
-      items = await this.loadItems(!isAdmin ? allowedNamespaces : null);
-    } finally {
-      if (items) {
-        items = this.sortItems(items);
-        this.items.replace(items);
-      }
+      items = await this.loadItems(namespaceStore.getContextNamespaces());
+      items = this.sortItems(items);
+      this.items.replace(items);
```
Nokel81:
```suggestion
      const items = await this.loadItems(namespaceStore.getContextNamespaces());
      this.items.replace(this.sortItems(items));
```

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -21,32 +21,103 @@ export const namespaceUrlParam = createPageParam<string[]>({
...
+    if (lastUsedNamespaces.size) {
+      return Array.from(lastUsedNamespaces);
+    }
+    // otherwise select "default" or first allowed namespace
+    else {
+      if (allowedNamespaces.has("default")) {
+        return ["default"];
+      } else if (allowedNamespaces.size) {
+        return [Array.from(allowedNamespaces)[0]];
+      }
```
Nokel81: Why only one?

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -21,32 +21,103 @@ export const namespaceUrlParam = createPageParam<string[]>({
...
+
+    if (!namespaces.length) {
+      return [...this.allowedNamespaces]; // show all namespaces when nothing selected
+    }
+
+    return namespaces;
   }
 
+  /**
+   * @deprecated
```
Nokel81: Could you add some help text to this deprecation label?

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -61,18 +132,22 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
...
-    } else {
-      return super.loadItems();
+    if (!isAllowedResource("namespaces")) {
+      return namespaces.map(this.getDummyNamespace);
     }
+
+    return Promise.all(namespaces.map(name => this.api.get({ name })));
   }
 
   protected getDummyNamespace(name: string) {
```
Nokel81: I don't think this should be a method, can you refactor this into a free-function?

src/renderer/components/+workloads-overview/overview.tsx
```diff
@@ -17,81 +17,68 @@ import { cronJobStore } from "../+workloads-cronjobs/cronjob.store";
...
+      isAllowedResource("statefulsets") && statefulSetStore,
+      isAllowedResource("replicasets") && replicaSetStore,
+      isAllowedResource("jobs") && jobStore,
+      isAllowedResource("cronjobs") && cronJobStore,
+      isAllowedResource("events") && eventStore,
+    ].filter(Boolean);
+
+    const unsubscribeMap = new Map<KubeObjectStore, Function>(
+      stores.map(store => [store, Function])
+    );
```
Nokel81:
```suggestion
    const unsubscribeMap = new Map<KubeObjectStore, () => void>();
```

src/renderer/components/+workloads-overview/overview.tsx
```diff
@@ -17,81 +17,68 @@ import { cronJobStore } from "../+workloads-cronjobs/cronjob.store";
...
+    const loadStores = async () => {
+      this.isLoading = true;
+
+      for (const store of stores) {
+        if (this.isUnmounting) break;
+
+        try {
+          store.reset();
+          await store.loadAll();
+          unsubscribeMap.get(store)(); // unsubscribe previous watcher
```
Nokel81: It seems weird to populate a map with a bunch of noops. I think conditionally calling the functions is better.

```suggestion
          unsubscribeMap.get(store)?.(); // unsubscribe previous watcher
```

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -111,28 +113,49 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
+    disposeOnUnmount(this, [
+      namespaceStore.onContextChange(() => this.loadStores(), {
+        fireImmediately: true,
+      })
+    ]);
+  }
+
+  async componentWillUnmount() {
+    this.isUnmounting = true;
+    await when(() => this.isLoaded);
```
Nokel81: This seems unnecessary, especially since JS is single threaded.

src/renderer/kube-object.store.ts
```diff
@@ -71,42 +77,45 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
   @action
-  async loadAll() {
+  async loadAll(params: { namespaces?: string[] } = {}) {
     this.isLoading = true;
     let items: T[];
 
     try {
-      const { allowedNamespaces, accessibleNamespaces, isAdmin } = getHostedCluster();
+      const { namespaceStore } = await import("./components/+namespaces/namespace.store");
+      const contextNamespaces = params.namespaces || namespaceStore.getContextNamespaces();
```
Nokel81: I think this is a bad idea. It implies that there is a circular dependency here. Probably because the namespace store is within the `+namespaces` component folder. I think we should move that out to remove this circular dependency.

src/renderer/item.store.ts
```diff
@@ -9,7 +9,7 @@ export interface ItemObject {
 
 @autobind()
 export abstract class ItemStore<T extends ItemObject = ItemObject> {
-  abstract loadAll(): Promise<void>;
+  abstract loadAll(...args: any[]): Promise<any>;
```
Nokel81: Why do we want `loadAll` to return anything?
```suggestion
  abstract loadAll<T extends any[]>(...args: T): Promise<any>;
```

src/renderer/components/+namespaces/namespace.store.ts
ixrock: Because some clusters might be huge. What you suggest?

src/renderer/components/+namespaces/namespace.store.ts
ixrock: It's not used in the codebase, but because `NamespaceStore` is exported to `extension-api` removing it would be breaking change.. Since api is not documented I'm doubt it's used in extensions already..
Therefore I suggest to remove this method without deprecation.

src/renderer/components/item-object-list/item-list-layout.tsx
ixrock: Well, this is partially true imo.. Just removing this would be not enough.. 
Consider a case:

1. `componentWillUnmount` called in the middle of stores loading (e.g. ns context changed) => unsubscribe is called
2. last actively loading store has finished and it adds own subscription
3. next iteration cycle reaches `break`, nothing new is loading but we have a stale subscription from previous step..

So I've added extra `unsubscribeStores()` when cycle is broken.

src/renderer/kube-object.store.ts
ixrock: This is trade-off and there is no circular dependency when loaded via dynamic `import`..
Ideally we should extract logic from namespace-store related to context namespaces, but it means to break current api or create code-duplication.

src/renderer/components/+namespaces/namespace.store.ts
Nokel81: The function is called `initNamespaces`, is that "initial" or "initialize"

src/renderer/kube-object.store.ts
Nokel81: Okay, sounds good. Maybe add a `TODO` comment about refactoring that out.

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -111,28 +112,49 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
   }
 
   async componentDidMount() {
+    disposeOnUnmount(this, [
+      namespaceStore.onContextChange(() => this.loadStores(), {
```
jim-docker: Aren't these changes pertinent only to `kube-object-list-layout`? `item-list-layout` can be used generically for items that have nothing to do with namespaces, right? Better to put the changes in `kube-object-list-layout.tsx`, if possible.

Hmm, I see namespaces are already a part of `item-list-layout`. Does this limit the use of `item-list-layout` to kube objects? (started using `item-list-layout` for workspaces in #1927)

ixrock: `ItemListLayout` definitely needs to be refactored, but not as part of this PR.

jim-docker: But the changes you're adding in this PR, can they be done in `kube-object-list-layout`?

src/renderer/kube-object.store.ts
```diff
@@ -71,42 +77,45 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
+      .then(items => items.flat());
   }
 
   protected filterItemsOnLoad(items: T[]) {
     return items;
   }
 
   @action
-  async loadAll() {
+  async loadAll(params: { namespaces?: string[] } = {}) {
```
aleksfront: Maybe using `namespaces` `string[]` argument instead of `params` object for sake of simplicity?

src/renderer/kube-object.store.ts
```diff
@@ -71,42 +77,45 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
     this.isLoading = true;
     let items: T[];
 
     try {
-      const { allowedNamespaces, accessibleNamespaces, isAdmin } = getHostedCluster();
+      const { namespaceStore } = await import("./components/+namespaces/namespace.store");
```
aleksfront: Just to double check - shouldn't we wait for `whenReady` here?

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -18,36 +18,111 @@ export const namespaceUrlParam = createPageParam<string[]>({
...
-    }, {
-      fireImmediately: true,
+  onContextChange(callback: (contextNamespaces: string[]) => void, opts: IReactionOptions = {}): IReactionDisposer {
+    return reaction(() => this.contextNs.toJS(), callback, {
       equals: comparer.identity,
+      ...opts,
     });
   }
 
+  private async init() {
```
aleksfront, line 47: Please consider how this function could be simlified.
1. It does a lot of mixed things - checks for hosted cluster, sets ready flag, introduces and creates disposers... It should be separated by couple of independent functions (I know they will be called only from here) to shrink cognitive load. This will allow also to remove comments. For example comment 
```// save selected namespaces to local-storage and update URL```
can be replaced by functions `saveNamespacesToStorage()`  and `updateURL()` etc.
2. Function with name `init()` shouldn't return anything? If it does, it should be called like `getNamespaceDisposers` or something like that?
3. Why we return disposers at the end if we don't use them anywhere?

File: src/renderer/kube-object.store.ts (hunk 12 above)
Comment (ixrock) [reply src/renderer/kube-object.store.ts]: Nope, this is intentional :D 
With argument as object you can easily add some new fields the object in child class, but adding a new argument will change method declaration and might require some overloading.

File: src/renderer/components/+namespaces/namespace.store.ts (hunk 14 above)
Comment (ixrock) [reply src/renderer/components/+namespaces/namespace.store.ts:47]: 1 - maybe
2,3 - `return` currently does nothing, but imho it's always better when function return something to cancel some operation running from the call (disposer, event stop, etc.)

File: src/renderer/components/item-object-list/item-list-layout.tsx (hunk 11 above)
Comment (ixrock) [reply src/renderer/components/item-object-list/item-list-layout.tsx]: Mostly no.. since `namespace` already introduced in `ItemsListLayout` and what's done in PR is just a small refactoring for loading stores + added "event-handler" for updating data related again to namespaces:

```ts
    disposeOnUnmount(this, [
      namespaceStore.onContextChange(() => this.loadStores(), {
        fireImmediately: true,
      })
    ]);
```

src/renderer/kube-object.store.ts
ixrock: Yes, good catch!

src/renderer/components/+namespaces/namespace.store.ts
ixrock: It's a `getter` function, so I guess by logic it's "initial"

ixrock:
@lensapp/lens-maintainers PTAL

src/renderer/api/kube-watch-api.ts
```diff
@@ -62,27 +62,41 @@ export class KubeWatchApi {
...
+  protected async getQuery() {
+    const { namespaceStore } = await import("../components/+namespaces/namespace.store");
+
+    await namespaceStore.whenReady;
+    const { isAdmin } = getHostedCluster();
 
     return {
       api: this.activeApis.map(api => {
-        if (isAdmin) return api.getWatchUrl();
+        if (isAdmin && !api.isNamespaced) {
```
jakolehm, line 76: This does not feel right. I think it's possible for non-admins to have an access to non-namespaced resources.

src/renderer/api/kube-watch-api.ts
```diff
@@ -158,6 +172,10 @@ export class KubeWatchApi {
 
   addListener(store: KubeObjectStore, callback: (evt: IKubeWatchEvent) => void) {
     const listener = (evt: IKubeWatchEvent<KubeJsonApiData>) => {
+      if (evt.type === "ERROR") {
```
jakolehm, line 175: Does this fix a bug? Should we backport it to 4.0.x?

src/renderer/kube-object.store.ts
```diff
@@ -71,42 +77,51 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
-      if (isAdmin && accessibleNamespaces.length == 0) {
-        items = await this.loadItems();
-      } else {
-        items = await this.loadItems(allowedNamespaces);
+        await namespaceStore.whenReady;
+        contextNamespaces = namespaceStore.getContextNamespaces();
       }
 
+      items = await this.loadItems({
+        isAdmin: getHostedCluster().isAdmin,
```
jakolehm: Just a note that we should avoid `getHostedCluster()` as much as possible. It works only if code is executed within a cluster context (read: cluster frame). It's like a magic global variable which feels wrong.

src/renderer/api/kube-watch-api.ts
ixrock: Remove `isAdmin` or what you suggest?

src/renderer/api/kube-watch-api.ts
ixrock: Maybe, but with watch-api related work for #1898 this will be removed soon.

src/renderer/api/kube-watch-api.ts
ixrock: Or maybe we should use `isAllowedResource` helper from `common/rbac.ts`?
Will be that correct?

src/renderer/kube-object.store.ts
ixrock: Alright

src/renderer/api/kube-watch-api.ts
ixrock: Added сheck ^ please re-review again

src/common/rbac.ts
```diff
@@ -7,38 +7,49 @@ export type KubeResource =
...
+export function isAllowedResourceType(kind: string): boolean {
+  const apiResource = apiResources.find(resource => resource.kind === kind);
+
+  if (apiResource) {
+    return getHostedCluster().allowedResources.includes(apiResource.resource);
+  }
+
+  return true; // allowed by default for other resources
+}
+
```
Nokel81: Why is this not a method on `Cluster`? Then the consumer can decide where the cluster comes from. For instance `getHostedCluster().isAlloedResourceType(kind)`

ixrock: Because https://github.com/lensapp/lens/pull/1918#discussion_r556272697

Nokel81: No, this function uses `getHostedCluster()` internally so it is no better (imo).

ixrock: Ok, anyway could be added in separated PR as needed.

Nokel81: I don't see why it should wait for a followup PR. This work is being done in this PR and as Jari noted we should be moving away from `getHostedCluster`.

ixrock: Alright! 👨🏼‍💻

ixrock: Added new methods: 

- `Cluster.isAllowedResource(kind: string): boolean`
- `KubeObjectStore.resolveCluster(): Promise<Cluster>`

Anyway.. `getHostedCluster()` dynamically imported in `resolveCluster` to get access to `isAllowedResource`, but now if store is used outside of cluster frame context, end user can override method and provide own cluster in different env.

Nokel81 (dismissed):

ixrock:
@Nokel81 PTAL again

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -106,8 +104,14 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
   getContextNamespaces(): string[] {
     const namespaces = this.contextNs.toJS();
 
+    // show all namespaces when nothing selected
     if (!namespaces.length) {
-      return [...this.allowedNamespaces]; // show all namespaces when nothing selected
+      if (this.isLoaded) {
+        // return actual namespaces list since "allowedNamespaces" updating every 30s in cluster and thus might be stale
+        return this.items.map(namespace => namespace.getName());
```
Nokel81, line 111: Why is this more up to date? This seems to be following a bad precedent of multiple sources of truth.

src/renderer/api/endpoints/helm-charts.api.ts
```diff
@@ -86,7 +86,7 @@ export class HelmChart {
   tillerVersion?: string;
 
   getId() {
-    return this.digest;
+    return `${this.apiVersion}/${this.name}@${this.getAppVersion()}`;
```
nevalla: just curious, what is the motivation of this change? is `this.digest` broken somehow?

ixrock: Yes, it's the same for all charts (at least was some days ago)

jakolehm: Let's split this into separate PR (and/or issue).

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -98,12 +98,14 @@ interface ItemListLayoutUserSettings {
...
+
+  private watchDisposers: IReactionDisposer[] = [];
+
   @observable hiddenColumnNames = new Set<string>();
   @observable isUnmounting = false;
 
-  // default user settings (ui show-hide tweaks mostly)
   @observable userSettings: ItemListLayoutUserSettings = {
-    showAppliedFilters: false,
+    showAppliedFilters: true,
```
jakolehm: Should be `false` by default.

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -120,31 +122,52 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
-      const subscriptions = stores.map(store => store.subscribe());
+  @computed get stores() {
+    const { store, dependentStores, tableId } = this.props;
 
-      await when(() => this.isUnmounting);
-      subscriptions.forEach(dispose => dispose()); // unsubscribe all
-    } catch (error) {
-      console.log("catched", error);
+    if (this.canBeConfigured) {
+      this.hiddenColumnNames = new Set(userStore.preferences?.hiddenTableColumns?.[tableId]);
```
jakolehm: This line raises an error on pod list when changing namespace filter.

jakolehm (changes requested):

src/renderer/components/item-object-list/item-list-layout.tsx
ixrock: I put it to `true` since it might be confusing now for end users because not all resources loaded and you see something like: **Filtered: 30/30** where numbers are mostly always the same (unless you change the search filter).

I remember even one issue about "how to reset all selected namespaces". So with this thing open by default it's more clear imho.

jakolehm: Let's not change this behaviour in this PR 🙂

src/renderer/components/item-object-list/item-list-layout.tsx
ixrock: Thanks, will fix!

src/renderer/api/endpoints/helm-charts.api.ts
ixrock: Ok

src/renderer/components/+namespaces/namespace.store.ts
ixrock: Bug scenario:

1. You have "All namespaces" selected filter
2. New namespace created and added via watch-api to the store (but it's not included yet to `cluster.allowedNamespaces` since it comes from `ipcMain` and updating every 30s)
3. You're switching to other resource tab and don't see anything about that namespace

src/renderer/components/item-object-list/item-list-layout.tsx
ixrock: Alright! Should I create another PR for that?

jakolehm (approved):

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -502,24 +493,47 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
+    const hiddenColumns = new Set(this.hiddenColumns);
+
+    if (!isVisible) {
+      hiddenColumns.add(columnId);
+    } else {
+      hiddenColumns.delete(columnId);
+    }
+
+    userStore.setHiddenTableColumns(this.props.tableId, hiddenColumns);
+  }
```
Nokel81, line 520: I don't think these changes belong in this PR.

ixrock: it belongs since fixing a bug related to table's `hiddenColumns` with some refactoring applied

Nokel81: Ah sorry, will look again then.

ixrock: otherwise i would not touch it at all.. :)

ixrock: Or you know, actually it was some bad merge from master which involved this  refactoring..  but initially it was looking like a bug.

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -347,16 +329,11 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
+    return <PageFiltersList filters={filters}/>;
   }
 
   renderNoItems() {
-    const { allItems, items, filters } = this;
-    const allItemsCount = allItems.length;
-    const itemsCount = items.length;
-    const isFiltered = filters.length > 0 && allItemsCount > itemsCount;
-
-    if (isFiltered) {
```
aleksfront, line 359: Nice! 👍

aleksfront (approved):

PR #1927, opened 2021-01-10
Workspace overview (WIP)

### LINKED ISSUE #1864: Workspace overview

Workspace overview:
- overview of workspace clusters
- extension api for registing additional content

jim-docker:
This is a mockup of how the workspace UX could be refactored. I came to this idea while thinking about how the user would bring up a workspace overview. This would make the UX for adding/deleting workspaces similar to that of kube resources (like namespaces) and the overview for a workplace would simply be the familiar details page. Customization of the workspace overview would also be a familiar experience for the extension developer as it would be analogous to adding kubeObjectDetailItems.

There could also be a quick link to the current workspace's details/overview on the dashboard somewhere.

https://user-images.githubusercontent.com/40840436/104137183-feb65b00-5368-11eb-9484-df261c6ec614.mov

Todo:
- [ ] implement business logic (use existing) for adding/deleting/editing/activating workspaces
- [ ] figure out why the ItemListLayout only fills about a third of the width (css? argh)
- [ ] flesh out the details/overview content (sortable/searchable list of clusters, editable description etc)
- [ ] add extension api support for details/overview customization

fixes #1864

### DIFF
--- a/src/renderer/components/+workspaces/workspace-list.tsx
+++ b/src/renderer/components/+workspaces/workspace-list.tsx
@@ -1,0 +1,44 @@
+import "./workspaces.scss";
+
+import React from "react";
+import { TabLayout } from "../layout/tab-layout";
+import { Badge } from "../badge";
+import { ItemListLayout, ItemListLayoutProps } from "../item-object-list/item-list-layout";
+import { Workspace, WorkspaceId } from "../../../common/workspace-store";
+import { WorkspaceItem, workspaceListStore } from "./workspace-list.store";
+
+enum sortBy {
+  name = "name",
+  id = "id",
+}
+
+export class WorkspaceList extends React.Component {
+  componentDidMount() {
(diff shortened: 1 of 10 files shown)


jakolehm:
@jim-docker I agree that this is much better UI/UX for workspaces but it does not really fix #1864 . I think we need an another PR for "a workspace overview".

src/renderer/components/+workspaces/workspace-list.tsx
```diff
@@ -0,0 +1,90 @@
...
+  };
+
+  hideDetails = () => {
+    this.showDetails(null);
+  };
+
+  render() {
+    return (
+      <TabLayout>
+        <ItemListLayout
```
jim-docker, line 56: @aleksfront or @ixrock, any idea why this element only fills about a third of the width available (see video)? @Nokel81 and I couldn't figure it out...

PR #1940, opened 2021-01-11
Fix fill the gaps logic in normalize metrics

jakolehm:


src/renderer/api/endpoints/metrics.api.ts
```diff
@@ -78,10 +78,15 @@ export function normalizeMetrics(metrics: IMetrics, frames = 60): IMetrics {
       result.forEach(res => {
         if (!res.values || !res.values.length) return;
 
+        let now = moment().unix();
```
Nokel81: Shouldn't this be above line 78?

src/renderer/api/endpoints/metrics.api.ts
```diff
@@ -78,10 +78,15 @@ export function normalizeMetrics(metrics: IMetrics, frames = 60): IMetrics {
...
+
         while (res.values.length < frames) {
-          const timestamp = moment.unix(res.values[0][0]).subtract(1, "minute").unix();
+          const timestamp = moment.unix(now).subtract(1, "minute").unix();
 
-          res.values.unshift([timestamp, "0"]);
+          if (!res.values.find((value) => value[0] === timestamp)) {
```
Nokel81, line 95: I wonder how the comparison between `moment`'s work? Does `unix()` return a number? 

```suggestion
          if (!res.values.find(([value]) => value === timestamp)) {
```

Won't this sometimes break if the metrics values are from "now - (5 minutes + 3 seconds)" to "now"?

src/renderer/api/endpoints/metrics.api.ts
```diff
@@ -78,10 +78,15 @@ export function normalizeMetrics(metrics: IMetrics, frames = 60): IMetrics {
       result.forEach(res => {
         if (!res.values || !res.values.length) return;
 
+        let now = moment().unix();
+
         while (res.values.length < frames) {
-          const timestamp = moment.unix(res.values[0][0]).subtract(1, "minute").unix();
+          const timestamp = moment.unix(now).subtract(1, "minute").unix();
```
Nokel81: Also, why is it better to `unshift` values from "now" instead of the first time position?

src/renderer/api/endpoints/metrics.api.ts
jakolehm: `unix()` returns a number. "5 minutes + 3 seconds" is a problem, we should probably round the initial value properly.

src/renderer/api/endpoints/metrics.api.ts
jakolehm: First time position might be "X minutes ago", we need to normalize these metrics from "now".

Nokel81: So why not instead of using `now` we just take `res.values[0][0]` and start subtracting from there?

jakolehm: > So why not instead of using `now` we just take `res.values[0][0]` and start subtracting from there?

Now it uses `res.values[0][0]` so that we don't hit into clock sync issues and uses both subtract & add to fill gaps both ways.

aleksfront (approved):
We definitely need to add test coverage for that.

nevalla (approved):

src/renderer/api/endpoints/metrics.api.ts
```diff
@@ -78,10 +78,24 @@ export function normalizeMetrics(metrics: IMetrics, frames = 60): IMetrics {
...
+        let now = moment().startOf("minute").subtract(1, "minute").unix();
+        let timestamp = res.values[0][0];
+
+        while (timestamp <= now) {
+          timestamp = moment.unix(timestamp).add(1, "minute").unix();
+
+          if (!res.values.find((value) => value[0] === timestamp)) {
+            res.values.push([timestamp, "0"]);
+          }
+        }
```
Nokel81, line 90: I still don't see how this fixes the issue. If `timestamp` is not `startOf("minute")` then line 87 will always lead to line 88.

Maybe I am not fulling understanding exactly what the goal is. But I definitely agree with @aleksfront that there should have been tests added with this change.

PR #1941, opened 2021-01-11
Add auto-update notifications and confirmation

### LINKED ISSUE #1768: Prompt for auto-update

Nokel81:
This is mostly a cherry pick of #1831 with small changes to make it work with other changes made so far in 4.x.

Final PR in the auto-updater saga.

Closes #1768


jimmyesquire (dismissed):

scripts/integration.sh
```diff
@@ -0,0 +1,31 @@
...
+    find ~/.config/Logs/Lens -type f -name *.log -delete
+  ;;
+  win)
+    find %APPDATA%/Logs/Lens -type f -name *.log -delete
+  ;;
+esac
+
+make build-extension-types build-extensions
+yarn build:$1
+DEBUG=true yarn integration-runner
```
jim-docker: what is `integration-runner`?

src/renderer/themes/kontena-dark.json
```diff
@@ -0,0 +1,112 @@
+{
```
jim-docker: renamed from lens-dark.json(?)

Nokel81: Oops thanks for the catch

scripts/integration.sh
Nokel81: See `package.json`

jim-docker: I don't see it. What am I looking for?

Nokel81: Oh I see, I forgot to cherry pick that in, good catch.

src/main/app-updater.ts
```diff
@@ -1,20 +1,158 @@
...
+  autoUpdater.autoDownload = false;
+  autoUpdater.autoInstallOnAppQuit = false;
+
+  autoUpdater
+    .on("update-available", async (args: UpdateInfo) => {
+      try {
+        const releaseDate = moment(args.releaseDate);
+        const body = `Version ${args.version} was released on ${releaseDate.format("dddd, MMMM Do, yyyy")}.`;
+
+        windowManager.sendToView({
```
jakolehm: Why not `broadcastMessage()`?

src/renderer/components/notifications/notifications.store.tsx
```diff
@@ -26,6 +62,29 @@ export class NotificationsStore {
 
   protected autoHideTimers = new Map<NotificationId, number>();
 
+  registerIpcListener(): void {
+    logger.info(`[NOTIFICATION-STORE] start to listen for notifications requests from main`);
+    ipcRenderer.on(NotificationChannelAdd, (event, model: MainNotification) => {
```
jakolehm: Why not `subscribeToBroadcast()`?

src/renderer/components/notifications/notifications.store.tsx
```diff
@@ -18,6 +22,38 @@ export interface Notification {
...
+    return null;
+  }
+
+  return (
+    <>
+      <br />
+      <div className="ButtonPannel flex row align-right box grow">
+        {buttons.map(({ backchannel, ...props }) => (
+          <Button key={backchannel} {...props} onClick={() => {
+            ipcRenderer.send(backchannel);
```
jakolehm: Does this mean that it's only possible to send notifications from `main`?  (`broadcastMessage()` probably would solve this)

Nokel81: I don't understand, `renderer` already has a way to display notifications.

src/renderer/components/notifications/notifications.store.tsx
Nokel81: So either each `iframe` has its own `NotificationStore` (or that used to be the case in 3.X) and we don't want notifications from main to be duplicated or there is only one `NotifcationStore` for all of the `renderer` and this works (without do all the extra work of sending to the other frames). 

Lastly, I don't think that `main` should have to care about `NotificationId` to avoid duplicate rendered notifications.

src/renderer/components/notifications/notifications.store.tsx
jakolehm: But that's different way, right? Would it make sense to have just one way to trigger notifications (might be out of scope of this PR).

Nokel81: I think that is out of scope for this PR. Yes it is a different way. They call `Notification.(info|error|success)(...)`

src/main/app-updater.ts
```diff
@@ -1,20 +1,124 @@
...
-    autoUpdater.logger = logger;
+        resolve();
+      })
+      .on(noChannel, () => {
+        logger.info("[UPDATE CHECKER]: User chose not to update");
+        cleanupChannels();
+        resolve();
+      });
+
+    windowManager.sendToView({
```
jakolehm: I feel this is tight coupling of main & renderer & ui. Could we just broadcast a message that an update is available (with update info). Based on that message renderer could show a notification. On notification action renderer could send a message back to main. Meaning: main should not need to know how update information is drawn, it just provides the information and channel for triggering an update process.

Nokel81: Do you mean the `status` ad `action` fields below? Because the rest of that data is quite required. I don't think that it is tight coupling give that `main` the source of a notification should decided what is displayed on it.

Though if you are talking about the "sendToView" vs "broadcast". I don't think we should broadcast because then `main` has to worry about setting an ID so that the notification is only displayed once (which I don't think `main` should need to care about)

jakolehm: > Do you mean the `status` ad `action` fields below? Because the rest of that data is quite required. I don't think that it is tight coupling give that `main` the source of a notification should decided what is displayed on it.

Not really those specific fields but this logic in general. Why can't `main` just broadcast:
```ts
broadcastMessage("app:update-available", updateInfo);
```

and a component in renderer could listen this and act accordingly? Why `main` needs to know about buttons, labels etc.. it should be `renderer` decision to draw information.


> Though if you are talking about the "sendToView" vs "broadcast". I don't think we should broadcast because then `main` has to worry about setting an ID so that the notification is only displayed once (which I don't think `main` should need to care about)

IMHO `main` should not care who is listening these kind of messages as long as these messages stay within core application. Even if we send this to just a view how can `main` be sure that it's displayed only once?

nevalla: > Meaning: main should not need to know how update information is drawn, it just provides the information and channel for triggering an update process.

IMHO this sounds reasonable. Main just send information and renderer decides how the notification looks like. Ideally we should have only one way to trigger those notifications.

scripts/integration.sh
```diff
@@ -0,0 +1,31 @@
...
+  linux)
+    find ~/.config/Logs/Lens -type f -name *.log -delete
+  ;;
+  win)
+    find %APPDATA%/Logs/Lens -type f -name *.log -delete
+  ;;
+esac
+
+make build-extension-types build-extensions
+yarn build:$1
```
jakolehm: This feels a bit odd.. could you explain why we refactored away from top-level make targets?

src/common/ipc.ts
```diff
@@ -29,16 +29,21 @@ export function broadcastMessage(channel: string, ...args: any[]) {
...
+  views.forEach(async webContent => {
     const type = webContent.getType();
 
     logger.silly(`[IPC]: broadcasting "${channel}" to ${type}=${webContent.id}`, { args });
     webContent.send(channel, ...args);
-    getSubFrames().then((frames) => {
-      frames.map((frameInfo) => {
+
+    try {
+      const frames = await getSubFrames();
```
jakolehm: This actually changes how message ordering goes (just a bit but it might break expected behaviour).. Please revert unless there is a real reason why this is done (I tried to make similar change in #1957 but it did break some of the app functionality).

src/renderer/components/notifications/notifications.store.tsx
jakolehm: Related to my other review comment (https://github.com/lensapp/lens/pull/1941#r565863019)): could we just use `Notification` component to show a notification in `renderer`? If no -> could we enhance it so that it can handle these kind of use cases. We really should have just one way to trigger notifications (note: `Notification` is also exposed via extension API).

scripts/integration.sh
Nokel81: Because I don't think that makefile supports what I am doing with lines 17-31. Specifically checking the return.

src/common/ipc.ts
Nokel81: How does this change the messaging order? I didn't make line 32 a `for-loop` so these will still happen asynchronously (though actually concurrently because JS is single-treaded)

jakolehm: Based on discussions & testing it seems that this does not have side-effects. So probably fine, although I still don't see how this is related to this specific PR.

scripts/integration.sh
Nokel81: However, some searching has found that I can do it

src/common/ipc.ts
Nokel81: I don't remember either. Will pull out and put into own PR

Nokel81:
Here is a verification image of the notification styling:

Screen Shot 2021-01-29 at 2 26 52 PM

Note: macOS only has "yes" and "no" whereas linux and windows will have "yes, now", "yes, on quit", and "no"

Makefile
```diff
@@ -38,18 +38,21 @@ test: binaries/client
 
 .PHONY: integration-linux
 integration-linux: build-extension-types build-extensions
+	-find ~/.config/Logs/Lens -type f -name *.log -delete
```
jakolehm: These are probably debug leftovers?

src/main/app-updater.ts
```diff
@@ -1,20 +1,71 @@
...
-    autoUpdater.logger = logger;
+  autoUpdater.logger = logger;
+  autoUpdater.autoDownload = false;
+  autoUpdater.autoInstallOnAppQuit = false;
+
+  autoUpdater
+    .on("update-available", (args: UpdateInfo) => {
+      try {
+        // use a UUID so that this back-channel is harder to discover
+        const backchannel = uuid.v4();
```
jakolehm: Why do we need "hidden" backchannel? Could backchannel be just  something like `auto-update:${version}`.

Makefile
Nokel81: I sort of wanted them to help with debugging in the future, but I'll open a separate PR for that after 4.1

src/main/app-updater.ts
Nokel81: This is to try and discourage malicious extensions from spamming this backchannel

jakolehm:
@Nokel81 please resolve conflicts.

src/main/app-updater.ts
jakolehm: How could extensions spam this? They don't have access to this IPC.

Nokel81:
@jakolehm done

package.json
```diff
@@ -212,6 +212,7 @@
     "mobx-observable-history": "^1.0.3",
     "mobx-react": "^6.2.2",
     "mock-fs": "^4.12.0",
+    "moment": "^2.26.0",
```
aleksfront: Why we're moved out `moment` from `devDependencies`?

src/common/ipc/update-available/index.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export const UpdateAvailableChannel = "update-available";
+export const AutoUpdateLogPrefix = "[UPDATE-CHECKER]";
+
+/**
+ * [<back-channel>, <update-info>]
+ */
+export type UpdateAvailableFromMain = [string, UpdateInfo];
+
+export function areArgsUpdateAvailableFromMain(args: unknown[]): args is UpdateAvailableFromMain {
```
aleksfront, line 11: What exactly these typing mean? `args is UpdateAvailableFromMain` and at the same time `args: unknown[]`? Is it possible to determine argument types instead of `unknown`?

src/common/ipc/update-available/index.ts
```diff
@@ -0,0 +1,48 @@
...
+  if (args.length !== 2) {
```
aleksfront, line 12: Probably need to export to separate function because there's no clear description of this and following conditions. For example, why we return `false` if there's less or more 2 arguments? Separating into function with meaningful name will show the condition intent.

src/renderer/components/button/button-panel.tsx
```diff
@@ -0,0 +1,13 @@
+import "./button-panel.scss";
+import React from "react";
+
+export function ButtonPannel(props: React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement>) {
```
aleksfront: If I understand it right, this `ButtonPanel` shows buttons with a `margin-right` added to last button. Is it reasonable to create separate component for such a small purpose? I suggest to add flex styles in place where it's needed without creating `ButtonPanel`

src/renderer/components/button/button.scss
```diff
@@ -45,6 +51,7 @@
   &.outlined {
     color: inherit;
     background: transparent;
+    border: 1px solid;
```
aleksfront: Please check if apperance of button inside `<InfoPanel/>` stays same because of this change. Search shows that there at least 1 button that using `outlined` style in our codebase.

src/common/ipc/ipc.ts
```diff
@@ -0,0 +1,146 @@
...
+  source.on(channel, handler);
+}
+
+/**
+ * Adds a listener to `source` that checks to verify the arguments before calling the handler.
+ * @param channel The channel to be listened on
+ * @param listener The function for the channel to be called if the args of the correct type
+ * @param verifier The function to be called to verify that the args are the correct type
+ */
+export function onCorrect<
```
aleksfront: Do this function is necessary to be added in current PR? I'm worried that such addition would make auto-update issue quite complex.

aleksfront (dismissed):

package.json
Nokel81: I think I initially did so because we were formatting a date for the notification. However, looking now I see that we use it all over the place in `renderer` so we need it at runtime and thus should be part of `deps`

src/common/ipc/update-available/index.ts
Nokel81: The is what Typescipt calls a "user defined type-guard" (https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards).

Namely, it is a function that is used to validate that the inputs are a certain type and thus narrow them (if the function returns `true`).

src/common/ipc/update-available/index.ts
Nokel81: I completely disagree the function name is not meaningful and signature explains exactly what this function is trying to do.

IPC calls can happen from anywhere so the handler that is called gets some array of elements (namely `unknown[]`). I use `unknown` here instead of `any` because of the difference of the coercion rules. Plus typescipt now recommends to use `unknown` where you don't know the types.

Finally, `UpdateAvailableFromMain` is by defenition a tuple (or an array of length 2). So if the arguments passed to the fuction are not of length 2 we know that they are invalid and thus `!(args is UpdateAvailableFromMain)`.

src/main/app-updater.ts
Nokel81: Do extensions not have access to `ipcRenderer`?

src/renderer/components/button/button-panel.tsx
Nokel81: The `margin-right` is added to all buttons except that last one.

I was under the impression that such small components were exactly what React's design methodology was about, which is why I created one.

src/main/app-updater.ts
jakolehm: Lens is not exposing any `ipc` stuff to extensions. When (if) we expose `ipc` to extensions it most likely will be prefixed so that extensions cannot cause trouble to Lens internal communication (like stores).

Nokel81: Okay I can change over to an incrementing channel

src/common/ipc/ipc.ts
Nokel81: How does this add undue complexity. I find the use of this function much less complex because it helps enforce typing through the IPC channel (and logs when they differ).

src/renderer/ipc/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+function UpdateAvailableHandler(event: IpcRendererEvent, ...[backchannel, updateInfo]: UpdateAvailableFromMain): void {
+  const notificationId = uuid.v4();
+
+  function sendToBackchannel(data: BackchannelArg): void {
+    notificationsStore.remove(notificationId);
+    console.log("sending to backchanel", { backchannel, data });
+    ipcRenderer.send(backchannel, data);
+  }
+
+  function renderYesButtons() {
```
ixrock: Could be better to put `RenderYesButtons` name and use as `<ReactComponent/>` inside notification?

Nokel81: Probably, good idea

src/renderer/ipc/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+      id: notificationId,
+      onClose() {
+        sendToBackchannel({ doUpdate: false });
+      }
+    }
+  );
+}
+
+export function registerIpcHandlers() {
+  onCorrect(ipcRenderer, UpdateAvailableChannel, UpdateAvailableHandler, areArgsUpdateAvailableFromMain);
```
ixrock: I would more expect to see something:

```ts
return unregisterIpcHandlers(() => onCorrect(ipcRenderer, UpdateAvailableChannel, UpdateAvailableHandler, areArgsUpdateAvailableFromMain)); // returns unsubscribe: () => void
```

src/common/ipc/ipc.ts
```diff
@@ -0,0 +1,146 @@
...
+// https://www.electronjs.org/docs/api/ipc-main
+// https://www.electronjs.org/docs/api/ipc-renderer
+
+import { ipcMain, ipcRenderer, webContents, remote } from "electron";
+import { toJS } from "mobx";
+import { EventEmitter } from "ws";
+import logger from "../../main/logger";
+import { ClusterFrameInfo, clusterFrameMap } from "../cluster-frames";
+
+const subFramesChannel = "ipc:get-sub-frames";
```
ixrock, line 10: Shouldn't we prefix `subFramesChannel` with corresponding frame-id?

src/common/ipc/ipc.ts
```diff
@@ -0,0 +1,146 @@
...
+  source: EM,
+  channel: string | symbol,
+  listener: L,
+  verifier: (args: unknown[]) => args is T
+): void {
+  function handler(event: HandlerEvent<EM>, ...args: unknown[]): void {
+    if (verifier(args)) {
+      source.removeListener(channel, handler); // remove immediately
+
+      Promise.resolve(listener(event, ...args)) // might return a promise
```
ixrock: If `listener(event, ...args)` will throw would it be caught in `.catch()`?

src/renderer/ipc/index.tsx
Nokel81: Why do you expect that the handlers will be removed? These are supposed to be present for the entire lifetime of the `renderer`.

src/common/ipc/ipc.ts
Nokel81: Did I copy this over wrong (let me check)?

src/common/ipc/ipc.ts
Nokel81: No, (though I thought it did). Will fix, thanks.

src/renderer/ipc/index.tsx
ixrock: Btw, there is not big correlation with `onCorrect` arguments and their names.

Would be better/more readable if `onCorrect` could accept object-map (and this declaration actually easier to extend later when needed, especially critical when you export something like to extensions-api).

```ts
onCorrect({
  source: ipcRenderer,
  channel: UpdateAvailableChannel,
  listener: UpdateAvailableHandler,
  verifier: areArgsUpdateAvailableFromMain
})
```

Also when I open a file with `areArgsUpdateAvailableFromMain` there is also nothing about that it's some kinda `verifier` argument. Maybe better to name files with a suffix `*.verifier.ts`? Wdyt?

Nokel81: Do you mean a verifier function? I figured that it was obvious from the signature (specifically its declared return value). But I can move it to a different function.

As for taking an object instead of positional args. Sure.

src/renderer/components/button/button.scss
```diff
@@ -21,10 +21,16 @@
   &.primary {
     background: $buttonPrimaryBackground;
   }
+
   &.accent {
     background: $buttonAccentBackground;
   }
 
+  &.light {
+    background-color: white;
```
ixrock: Is it works good with dark/light themes in Lens?

Nokel81: Will check, this was copied directly from the changes Alexei made for the 3.6.10 release. Will verify.

src/renderer/ipc/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { Button, ButtonPannel } from "../components/button";
+import { isMac } from "../../common/vars";
+import * as uuid from "uuid";
+
+function UpdateAvailableHandler(event: IpcRendererEvent, ...[backchannel, updateInfo]: UpdateAvailableFromMain): void {
+  const notificationId = uuid.v4();
+
+  function sendToBackchannel(data: BackchannelArg): void {
+    notificationsStore.remove(notificationId);
+    console.log("sending to backchanel", { backchannel, data });
```
ixrock: Forgotten `console.log`?

src/common/ipc/ipc.ts
```diff
@@ -0,0 +1,146 @@
...
+import { ipcMain, ipcRenderer, webContents, remote } from "electron";
+import { toJS } from "mobx";
+import { EventEmitter } from "ws";
+import logger from "../../main/logger";
+import { ClusterFrameInfo, clusterFrameMap } from "../cluster-frames";
+
+const subFramesChannel = "ipc:get-sub-frames";
+
+export type HandlerEvent<EM extends EventEmitter> = Parameters<Parameters<EM["on"]>[1]>[0];
+export type EventListener<E extends EventEmitter, T extends any[]> = (event: HandlerEvent<E>, ...args: T) => any;
```
ixrock: Seems not used `type` anywhere.

src/common/ipc/ipc.ts
Nokel81: This is copied directly from `master` and I haven't touched the code there.

src/renderer/components/button/button.scss
Nokel81: I removed it because it increases the height by 2px.

src/renderer/components/button/button.scss
Nokel81: It looks sort of okay. I think I'll change it to "invert".

src/renderer/components/button/button.scss
Nokel81: So the old `outlined` styling only worked if the button wasn't also styled the same as the background (namely you couldn't have a `primary outlined` button on `primary` notification. That is why I initially had this change.

src/renderer/components/button/button-panel.tsx
aleksfront: For creating gaps (`margin-right`) between buttons you can use following classes on the parent element
```
<div className="flex gaps">...</div>
```
The `.gaps` rule would add neede margins.

Please look if `ButtonPanel` component could be avoided.

Nokel81: Tested and removed.

package.json
jakolehm: Renderer deps are bundled with webpack. Most likely this should be under `devDependencies`.

src/common/ipc/ipc.ts
jakolehm: Why did we move `ipc` here?

src/common/ipc/update-available/index.ts
```diff
@@ -0,0 +1,48 @@
+import { UpdateInfo } from "electron-updater";
```
jakolehm, line 1: I'm not sure why we need to put these modules under `ipc/`? If we go with this route every module that uses `ipc` should end up here (which sounds wrong).

Nokel81: So you would prefer having these checking function under where? I put them here because these IPC types should be "public" to both `renderer` and `main` (as the receiver and sender) and I figured that the verification functions should be beside the type definitions.

src/common/ipc/ipc.ts
Nokel81: Do you mean why was `src/common/ipc.ts` was moved to `src/common/ipc/ipc.ts`? That is because I added another module file. This doesn't change any imports because of how `index.ts` is handled.

jakolehm: Ok, asked because it's quite hard to spot differences when whole file is moved and modified 🙂

src/common/ipc/update-available/index.ts
jakolehm: Ok, makes sense if this is just shared functionality.

src/common/ipc/ipc.ts
jakolehm: `onCorrect` sounds a bit abstract.. maybe we can improve names in future? (naming is hard)

Nokel81: Sounds good about improving the names in the future. For reference I chose this name because it is sort of equivalent to `ipc.on(...)` but guarantees `correct` arguments.

package.json
Nokel81: Okay will do

src/common/ipc/ipc.ts
Nokel81: Fair, should I move the argument checking functions to a different file so that this file is just a rename in the git diff?

Nokel81: Nvm, I'll do that anyway

Nokel81:
@aleksfront and @jakolehm please re review, I move some code around to other files so that it is easier to review.

jakolehm (approved):
LGTM

src/common/utils/delay.ts
```diff
@@ -1,6 +1,8 @@
-// Create async delay for provided timeout in milliseconds
-
-export async function delay(timeoutMs = 1000) {
-  if (!timeoutMs) return;
-  await new Promise(resolve => setTimeout(resolve, timeoutMs));
+/**
```
ixrock, line 1: Why this change is needed? :) 
Also I would prefer to keep `timeoutMs` as it is more clear/descriptive without consulting the docs..

ixrock (approved):

src/common/utils/delay.ts
Nokel81: I had this function as part of my code base before it go merged to master. I can revert the `timeoutMs` in a followup PR if you want.

Though side note, what editor are you using? I use vscode and it always shows the full docs (including the tsdocs) when I hover over a function.

ixrock: I use `WebStorm` and this is what I see when typing:

image

ixrock: Hovering works, but usually I don't use it (if I need to read docs I'd better navigate to the source):

image

hansiemithun:
This is not working anymore, the Lens is automatically updated on reopen. 

**Environment:** mac

Issue #428, opened 2020-06-10
Metrics history beyond 1h


jakolehm:

cosmonova:
Could be configured under Cluster Settings.

janerikmai:
I suggest providing zoomable graphs f.ex. to see long term container memory consumption (leaks) like in a grafana dashboard.

bertrandp:
Really need this feature

darkfessing:
+ very need

jakolehm:
This is already implemented.

PR #1944, opened 2021-01-12
Improve how metrics are queried & displayed

jakolehm:
- use regex for queries so that we can also include proper history (matching pod names gives only results from currently available pods)
- remove limits/requests from pod overview (those are really confusing and basically misleading information)
- add metrics to Jobs and Namespaces

image

Initial work towards #428


nevalla (dismissed):
LGTM

src/renderer/components/+namespaces/namespace-details.tsx
```diff
@@ -33,9 +36,18 @@ export class NamespaceDetails extends React.Component<Props> {
...
     return (
       <div className="NamespaceDetails">
+        {namespaceStore.isLoaded && (
+          <ResourceMetrics
+            loader={() => namespaceStore.loadMetrics(namespace)}
+            tabs={podMetricTabs} object={namespace} params={{ metrics }}
+          >
+            <PodCharts />
+          </ResourceMetrics>
+        )}
```
Nokel81, line 89: Looking at the code for `loadMetrics` it doesn't look like it is at all connected to `namespaceStore` being loaded.

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -22,12 +22,17 @@ export const namespaceUrlParam = createPageParam<string[]>({
...
 
   constructor() {
     super();
     this.init();
   }
 
+  async loadMetrics(namespace: Namespace) {
+    this.metrics = await namespacesApi.getMetrics(namespace.getName(), "");
+  }
+
```
Nokel81: This function seems completely disconnected from that the rest of `NamespaceStore` is. Why not just have people call the API directly?

aleksfront: Main reason for `loadMetrics` method is necessity to keep metric data in `@observable metrics` variable. This is dictated by `<ResouceMetrics>` component. It receives metrics as prop. So, currently we need to keep metrics variable in store or in the `<NamespaceDetails/>` component itself. I guess store is a better place because of quote from mobx documentation: 
> Make sure that everything that can be derived from the application state, will be derived

But I agree that we can refactor `<ResouceMetrics/>` for eliminating store `loadMetrics` calls.

aleksfront (dismissed):
LGTM

src/renderer/components/+namespaces/namespace.store.ts
Nokel81: My concern is that we seem to be pilling a lot onto these stores (especially `NamespaceStore`) and given that it is a public field of a class that is exported we cannot just go and remove it later.

Finally, it isn't really being derived from application state since we are calling the `loadMetrics` function on every single render.

I really think a better solution would be to have a metrics store that is query-able by resource type

aleksfront (dismissed):

Nokel81:
It seems like the diff is dirty because some of the changes currently in `master` show up as "additions". Maybe a rebase is in order?

aleksfront:
> It seems like the diff is dirty because some of the changes currently in `master` show up as "additions". Maybe a rebase is in order?

I checked `Files changed` and it's contains only changes made by current PR. What additions did you spot?

Nokel81:
@aleksfront I don't remember... will give this a review again

src/renderer/components/+namespaces/namespace-details.tsx
Nokel81: To reiterate, not only does the method `namespaceStore.loadMetrics` have nothing with `namespaceStore.isLoaded` the store doesn't even track which namespace the metrics are from. So on initial load (and display) the user could be seeing metrics from a previously selected namespace.

Nokel81 (changes requested):

src/renderer/components/+namespaces/namespace.store.ts
jakolehm: Originally I did this because it was how other components already worked. That said I completely agree that we shouldn't pollute all of these store classes -> I'll refactor this PR.

ixrock (dismissed):

src/renderer/components/+namespaces/namespace-details.tsx
jakolehm: Good catch, fixed.

src/renderer/components/+namespaces/namespace.store.ts
jakolehm: Refactored metrics stuff to standalone functions. This way we don't pollute extension api.

src/renderer/components/+cluster/cluster-overview.store.ts
```diff
@@ -78,7 +78,7 @@ export class ClusterOverviewStore extends KubeObjectStore<Cluster> implements Cl
     const { masterNodes, workerNodes } = nodesStore;
     const nodes = this.metricNodeRole === MetricNodeRole.MASTER && masterNodes.length ? masterNodes : workerNodes;
 
-    this.metrics = await clusterApi.getMetrics(nodes.map(node => node.getName()), params);
+    this.metrics = await getMetricsForNodeNames(nodes.map(node => node.getName()), params);
```
Nokel81, line 102: Nit: maybe `getMetricByNodeNames`?

Nokel81 (dismissed):

jakolehm:
@nevalla @aleksfront @Nokel81 ptal

src/renderer/components/+namespaces/namespace-details.tsx
Nokel81: Why was this resolved?

I still don't understand why metrics should be stored in the stores? It isn't like they are being kept in a map based on namespace or anything?

Why can't they just be part of this component and then go away when it is unmounted?

aleksfront:
Closing this in favor of https://github.com/lensapp/lens/pull/3424 and https://github.com/lensapp/lens/pull/3369.

PR #1945, opened 2021-01-12
In-app survey extension

nevalla:


extensions/survey/src/survey-preference.tsx
```diff
@@ -0,0 +1,27 @@
+import { Component } from "@k8slens/extensions";
+import React from "react";
+import { observer } from "mobx-react";
```
Nokel81, line 3: How is this being imported if it is not listed as a dep? Also, are we not exporting this as part of `@k8slens`?

extensions/survey/src/survey-preference.tsx
```diff
@@ -0,0 +1,27 @@
...
+@observer
+export class SurveyPreferenceInput extends React.Component<{survey: SurveyPreferencesStore}, {}> {
+  render() {
+    const { survey } = this.props;
+
+    return (
+      <Component.Checkbox
+        label="Allow in-app surveys"
+        value={survey.enabled}
+        onChange={v => { survey.enabled = v; }}
```
Nokel81: Other places don't use all of those, I don't think it is needed here either.
```suggestion
        onChange={v => survey.enabled = v}
```

extensions/survey/src/survey.ts
```diff
@@ -0,0 +1,46 @@
+import { Util } from "@k8slens/extensions";
+import { machineIdSync } from "node-machine-id";
```
Nokel81: Shouldn't this also be a dep? Along with `got` and `refiner-js`? Or are those only devDeps because they are compiled into the JS during build?

extensions/survey/src/survey.ts
```diff
@@ -0,0 +1,46 @@
...
+        id: surveyId,
+      });
+
+    }
+  }
+
+  async fetchSurveyId() {
+    try {
+      const surveyApi = process.env.SURVEY_API_URL || "https://survey.k8slens.dev";
+      const anonymousId = machineIdSync();
```
Nokel81: Shouldn't this use the async version?
```suggestion
      const anonymousId = await machineId();
```

extensions/survey/src/survey.ts
nevalla: yes those are compiled into the JS

extensions/survey/src/survey.ts
nevalla: Yep that would work too

extensions/survey/src/survey-preference.tsx
nevalla: https://github.com/lensapp/lens/blob/master/src/renderer/bootstrap.tsx#L26-L33

and:
```javascript
// webpack.config.js
...
externals: [
      {
        "@k8slens/extensions": "var global.LensExtensions",
        "react": "var global.React",
        "mobx": "var global.Mobx",
        "mobx-react": "var global.MobxReact"
      }
    ],
```

extensions/survey/src/survey-preference.tsx
nevalla: Changed

extensions/survey/src/survey.ts
nevalla: Changed

Nokel81 (dismissed):

extensions/survey/src/survey.ts
```diff
@@ -0,0 +1,46 @@
+import { Util } from "@k8slens/extensions";
+import { machineId } from "node-machine-id";
```
jim-docker, line 2: I noticed `node-machine-id` has this caveat:

Image-based environments have usually the same machine-id

jim-docker (dismissed):
Tested on Mac/Win/linux. Survey preference is respected. 
I did notice that the (same) survey is presented more than once when it is initially not responded to or left open. I started Lens, saw the survey come up, and then left it alone for more than 10 minutes. Then I completed the survey, and dismissed the "Thanks a lot for your response!" dialog by clicking the "X". A few seconds later the survey was presented again. I completed the survey and then let the "Thanks a lot for your response!" dialog dismiss itself after a few seconds. Because I waited more than 10 minutes initially I anticipated that the survey might be presented a third time but it was not. (Refiner followup mode was enabled on the backend, for up to 3 survey showings, 5 minutes between followups). This is probably an issue with Refiner behaviour...

jakolehm (dismissed):
LGTM

jakolehm:
Screenshot for other reviewers:

image (3)

aleksfront:
Spotted same behavior like @jim-docker before. Survey appeared 2 times in a row. Second one was right behind first one.

aleksfront (dismissed):
I was worried how it look when light theme enabled, but it seems okay
lens survey

jim-docker (approved):

aleksfront (approved):

PR #1949, opened 2021-01-12
Add lens:// protocol handling with a routing mechanism

### LINKED ISSUE #1288: Add a special "lens" native OS protocol handler to enable additional extension use cases
Add a special native OS protocol handler (`lens://...`) to make it possible to "open in Lens" from the browser (for example).

This would enable reverse-integration use cases like "open in Lens" from the browser, and could be very powerful, especially for extensions since they will unlikely bring a lot of third-party connections _into_ Lens.

This mechanism should also support request parameters to enable even deeper integration: `lens://clusters/{id}?custom=value` would have Lens open and navigate to that cluster.

Extensions could then hook into "URL Events" to handle them before Lens does its default behavior, and the extension could get all the GET parameters in order to do additional special integration behavior...

When handling a "URL Event", an extension could call `event.preventDefault()` if it handled the event, and also `event.stopPropagation()` to imperatively prevent any other extension from handling the request; otherwise, following extensions would be able to check `event.defaultPrevented()` to see if the event has already been handled.

Nokel81:
Things that still need to be done:

- [x] Testing on linux
- [x] Testing on windows
- [x] Add markdown docs

fixes #1288


src/extensions/lens-main-extension.ts
```diff
@@ -16,4 +18,21 @@ export class LensMainExtension extends LensExtension {
...
+   *
+   * See https://www.npmjs.com/package/path-to-regexp. To use this the link
+   * `lens://extensions/<org-id>/<extension-name>/your/defined/path?with=query`
+   * or `lens://extensions/<extension-name>/your/defined/path?with=query`
+   * (if this extension is not packaged behind an organization) needs to be
+   * opened.
+   * @param pathSchema The path schema for the route.
+   * @param handler The function to call when this route has been matched
+   */
+  onProtocol(pathSchema: string, handler: RouteHandler): void {
```
jakolehm: Needs also something to `docs/`.

jakolehm (commented):
Looks pretty good already 👍 Could we add integration tests for protocol handler?

src/extensions/lens-main-extension.ts
Nokel81: Agreed will do.

Nokel81:
I will write some integration tests.

docs/extensions/guides/protocol-handlers.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Registering A Protocol Handler
+
+The method `onProtocolRequest` exists both on [`LensMainExtension`](extensions/api/classes/lensmainextension/#onprotocolrequest) and on [`LensRendererExtension`](extensions/api/classes/lensrendererextension/#onprotocolrequest).
+This is how, as an extension developer, you can register handlers for your extension.
+The `pathSchema` argument must comply with the [path-to-regexp](https://www.npmjs.com/package/path-to-regexp) package's `compileToRegex` function.
+
+Once you have registered a handler it will be called when a user opens a link on their computer.
+The routing mechanism for extensions is quite straight forward.
+For example consider an extension `example-extension` which is publish by the `@mirantis` org.
```
jim-docker:
```suggestion
For example consider an extension `example-extension` which is published by the `@mirantis` org.
```

docs/extensions/guides/protocol-handlers.md
```diff
@@ -0,0 +1,47 @@
...
+If it were to register a handler with `"/display/:type"` as its corresponding then we would match the following URI like this:
```
jim-docker:
```suggestion
If it were to register a handler with `"/display/:type"` as its corresponding link then we would match the following URI like this:
```

stefcameron:
Great to see this going in! @Nokel81 are there any changes from the original implementation?

src/common/protocol-handler/router.ts
```diff
@@ -0,0 +1,205 @@
...
+  if (!hasOwnProperties(args, "extensionId")) {
+    return false;
+  }
+
+  if (typeof args.extensionId !== "string" || args.extensionId.length === 0) {
+    // ipcChannel is required, must be a string, must be non-empty
+    return false;
+  }
+
+  return false;
```
jim-docker:
```suggestion
  return true;
```

Nokel81:
> Great to see this going in! @Nokel81 are there any changes from the original implementation?

From a users perspective no. So your extension technically doesn't have to be even updated. However, from an implementation standpoint. This now supports handlers from both `renderer` and `main`. Our hackweek version only supported `renderer` handlers.

stefcameron:
> > Great to see this going in! @Nokel81 are there any changes from the original implementation?
> 
> From a users perspective no. So your extension technically doesn't have to be even updated. However, from an implementation standpoint. This now supports handlers from both `renderer` and `main`. Our hackweek version only supported `renderer` handlers.

That's cool! I remember you wanted to add support to `main` as well but didn't have time then. As far as the API goes, though, no changes?

Nokel81:
> > > Great to see this going in! @Nokel81 are there any changes from the original implementation?
> > 
> > 
> > From a users perspective no. So your extension technically doesn't have to be even updated. However, from an implementation standpoint. This now supports handlers from both `renderer` and `main`. Our hackweek version only supported `renderer` handlers.
> 
> That's cool! I remember you wanted to add support to `main` as well but didn't have time then. As far as the API goes, though, no changes?

None, the method on `LensRendererExtension` (and `LensMainExtension` for that matter) is still `onProtocolRequest`.

Nokel81:
I even checked what your extension was expecting to make sure I didn't change anything 😉

stefcameron:
> I even checked what your extension was expecting to make sure I didn't change anything ;)

Fantastic! Thanks for testing it out. 😄

src/extensions/lens-main-extension.ts
```diff
@@ -16,4 +18,15 @@ export class LensMainExtension extends LensExtension {
...
+  /**
+   * Registers a handler to be called when a `lens://` link is called.
+   * @param pathSchema The path schema for the route.
+   * @param handler The function to call when this route has been matched
+   */
+  onProtocolRequest(pathSchema: string, handler: RouteHandler): void {
+    const lprm = LensProtocolRouterMain.getInstance<LensProtocolRouterMain>();
+
+    lprm.extensionOn(this.name, pathSchema, handler);
+  }
```
jim-docker: Can this be defined in `LensExtension` class instead? (and not duplicated in `LensRendererExtension`)

src/main/protocol-handler/__test__/router.test.ts
```diff
@@ -0,0 +1,155 @@
...
+      expect(await lpr.route(Url("lens://internal/page/foo"))).toBeUndefined();
+    } catch (error) {
+      expect(throwIfDefined(error)).not.toThrow();
+
+    }
+
+    expect(called).toBe("foo");
+  });
+
+  it("should call most exact handler for an extensions", async () => {
```
jim-docker:
```suggestion
  it("should call most exact handler for an extension", async () => {
```

Nokel81:
@nevalla any ideas about why the integration tests for mac and linux are not being reported as failing in the CI, even though if you look at the logs there are failed tests?

src/common/protocol-handler/router.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  return true;
+}
+
+/**
+ * This function validates that `args` is at least `DeregisterParams`
+ * @param args a deserialized value
+ */
+export function validateDeregisterParams(args: unknown): args is DeregisterParams {
+  if (args == null || typeof args !== "object") {
```
stefcameron: A much safer check would be `if (!args || typeof args !== "object")`

src/common/protocol-handler/router.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  return false;
+}
+
+/**
+ * This function validates that `args` is at least `RouteParams`
+ * @param args a deserialized value
+ */
+export function validateRouteParams(args: unknown): args is RouteParams {
+  if (args == null || typeof args !== "object") {
```
stefcameron: Better as `if (!args || typeof args !== "object")`

jim-docker:
> @nevalla any ideas about why the integration tests for mac and linux are not being reported as failing in the CI, even though if you look at the logs there are failed tests?

For this PR the only failed test I see for windows and mac is the protocol app start test. There's no minikube set up on Windows or mac so those tests are skipped. Or do you see other failures for other builds?

jim-docker:
> > @nevalla any ideas about why the integration tests for mac and linux are not being reported as failing in the CI, even though if you look at the logs there are failed tests?
> 
> For this PR the only failed test I see for windows and mac is the protocol app start test. There's no minikube set up on Windows or mac so those tests are skipped. Or do you see other failures for other builds?

Oh, but the CI reports success. Ruh Roh.

maraslibay:
15 Oca 2021 Cum, saat 22:41 tarihinde Stefan Cameron <
notifications@github.com> şunu yazdı:

> Great to see this going in! @Nokel81 are there any changes from the
> original implementation?
>
> From a users perspective no. So your extension technically doesn't have to
> be even updated. However, from an implementation standpoint. This now
> supports handlers from both renderer and main. Our hackweek version only
> supported renderer handlers.
>
> That's cool! I remember you wanted to add support to main as well but
> didn't have time then. As far as the API goes, though, no changes?
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.
>
İyi akşamlar benle iletişime geçermisizin çok acil

nevalla:
> @nevalla any ideas about why the integration tests for mac and linux are not being reported as failing in the CI, even though if you look at the logs there are failed tests?

Yes those should have been reported as failed. Hard to say why it's not the case here, need to investigate further.

nevalla:
Okay...I think I got it. We are running `git checkout` as the last step and exit code of that command is returned:
```
 - bash: |
      rm -rf extensions/telemetry
      make integration-win
      git checkout extensions/telemetry
    displayName: Run integration tests
```

 We might want to return the exit code from `make integration-*` command.

src/extensions/lens-main-extension.ts
Nokel81: No, because the renderer needs to call the `LensProtocolRouterRenderer`'s method instead of the `LensProtocolRouterMain`'s method like here.

src/common/protocol-handler/router.ts
Nokel81: How is that safer? `typeof args === "object"` implies either args is `{...}` or `null`.

Nokel81:
> Okay...I think I got it. We are running `git checkout` as the last step and exit code of that command is returned:
> 
> ```
>  - bash: |
>       rm -rf extensions/telemetry
>       make integration-win
>       git checkout extensions/telemetry
>     displayName: Run integration tests
> ```
> 
> We might want to return the exit code from `make integration-*` command.

Thanks will fix that.

Nokel81:
@stefcameron I didn't test your extension. I only made sure that the API was compatible.

src/common/protocol-handler/router.ts
stefcameron: But the preceding `!arg` will catch the `null`, `undefined`, `""`, whatever flasy case.

Note that `typeof [] === 'object'` too, so if you're wanting this to be bulletproof, there's still more to do, but technically, an Array is an object and you can safely get/set any property on it, so I figured that wasn't a concern here.

Technically, `arg == null` should also catch the `undefined` case; it's just so foreign to see a `==` in a JS code base. 😄 Implicit coercion of values on comparison is the source of much trouble...

stefcameron:
> @stefcameron I didn't test your extension. I only made sure that the API was compatible.

Oh, thanks for clarifying that. I'll git it a try against this branch then to make sure it still works.

src/common/protocol-handler/router.ts
jim-docker: I agree, I like `!args` better than `args == null`. I don't like using the truthy operator `==`. Whenever I see it I wonder if they meant to use `===`.

jim-docker:
I think the protocol name shoud be something less generic than `lens`, like `k8slens`, since it is global (`lens` protocol could already be in use somewhere?)

src/common/protocol-handler/router.ts
Nokel81: Yeah it isn't a concern to me that `typeof [] === "object"`, also `!args` is still a form of implicit coercion. I'll change it if you guys want though. I just felt that this was more explicit about what I wanted.

Nokel81:
> I think the protocol name shoud be something less generic than `lens`, like `k8slens`, since it is global (`lens` protocol could already be in use somewhere?)

As seen on slack, it has been decided to keep it as `lens`

docs/extensions/guides/protocol-handlers.md
```diff
@@ -45,3 +45,9 @@ On the other hand, the subpath `"/display/notification"` would be routed to #3.
 
 The URI is routed to the most specific matching `pathSchema`.
 This way the `"/"` (root) `pathSchema` acts as a sort of catch all or default route if no other route matches.
+
+### Cleaning Up
+
+Currently there is not way to remove a protocol handler once it has been registered.
```
jim-docker: ...no way to remove...

Why say "currently", are we planning to add a way?

I'm not sure this will fly, there may be use cases for removing a handler while an extension is active. Why don't you want to provide `removeProtocolHandlers()` method in the extension classes as in your hack-week project?

What about a handler being removed by calling `onProtocolRequest()` with the desired pathSchema and null for the handler?

Nokel81: `removeProtocolHandlers` removes *all* handlers. Yes calling `onProtocolRequest` would remove the handler. I didn't add a removal mechanism because I didn't think that was in the current design (though TBH I was never really given a design, I just asked some questions).

The alternative that I made for the old `removeProtocolHandlers` was to just call it directly for the extension developer.

stefcameron: +1 from me! I think the ability to remove is something we need to provide. Can you add an event handler without ever being able to remove it...? If an extension wanted to stop listening to protocol requests, they'd have to implement some sort of mechanism to know to _ignore_ the requests. That's not very nice.

My vote would be for not overloading the meaning of `onProtocolRequest()` to also handle the removal. I think we should make removal explicit with a "remove" method.

Actually, it feels very much like adding/removing event handlers. The most intuitive thing there would be to rename `onProtocolRequest()` (which, frankly, sounds like a _handler_ itself, not the method you call to _add_ a handler) to `addProtocolHandler()`, and then provide a `removeProtocolHandler()` method.

Nokel81: Sounds good I can do that. Though I still think that we should remove the handlers when an extension is deactivated.

stefcameron: `onProtocolRequest` is more the name you'd give to a React prop on a component that provides a protocol event. But there, you assign your handler to it, and if you don't want to handle the event anymore, you assign `null/undefined` to it.

In this case, it's a method, so I think a different naming scheme would be more appropriate...

Here's what it looks like in my extension, for example:
https://github.com/Mirantis/lens-extension-cc/blob/master/src/renderer.tsx#L128-L131

It would feel more natural to me if it were:
```javascript
      this.addProtocolHandler(
        `/${EXT_EVENT_ACTIVATE_CLUSTER}`,
        this.handleProtocolActivateCluster
      );
      ...
      this.removeProtocolHandler(this.handleProtocolActivateCluster);
```

Oh, and I guess I'm advocating for a way to remove __individual__ handlers, not just all handlers (the hackweek branch just gave us `removeProtocolHandlers()` -- like all or nothing).

Nokel81: I know you are asking for removing individual handlers. However, I won't support removing handler by the function because I am sure that it might come up that an extension wants to use the same handler for multiple routes.

stefcameron: Then the `removeProtocolHandler()` can require the route as well.

Nokel81:
Update: after a design session with Jari and Jim there is a decent amount of more work to do and a significant set of changes. So I will set this to draft mode for the time being.

stefcameron:
> Update: after a design session with Jari and Jim there is a decent amount of more work to do and a significant set of changes. So I will set this to draft mode for the time being.

I appreciate the update! 🙂 Hoping to see this land soon.

docs/extensions/guides/protocol-handlers.md
```diff
@@ -0,0 +1,57 @@
+# Lens Protocol Handlers
+
+Lens has a file association with the `lens://` protocol.
+This means that Lens can be opened by external programs by providing a link that has `lens` as its protocol.
+Lens provides a routing mechanism that extensions can use to register custom handlers.
+
+## Registering A Protocol Handler
+
+The field `protocolhandlers` exists both on [`LensMainExtension`](extensions/api/classes/lensmainextension/#protocolhandlers) and on [`LensRendererExtension`](extensions/api/classes/lensrendererextension/#protocolhandlers).
```
stefcameron: Isn't it `protocolHandlers`? (capital "H") Looks like it in `src/extensions/lens-extension.ts`

src/extensions/registries/protocol-handler-registry.ts
```diff
@@ -0,0 +1,10 @@
+import { RouteHandler } from "../../common/protocol-handler";
+
+/**
+ * ProtocolHandlerRegistration is the data required for an extension to register
+ * a handler to a specific path or dynamic path.
+ */
+export interface ProtocolHandlerRegistration {
+  pathSchema: string;
```
stefcameron, line 6: Seems a bit over the top to name it "pathSchema". How about just "path"?

stefcameron (commented):
@Nokel81 I'm trying to run Lens from this branch against my extension to test. I've made necessary updates to my extension for the new API. I can't run my extension, though, because of a mobx "max callstack side exceeded" extension the second Lens tries to load my extension:
```
Uncaught (in promise) RangeError: Maximum call stack size exceeded
    at isObservableObject (mobx.module.js:4409)
    at _isObservable (mobx.module.js:2550)
    at isObservable (mobx.module.js:2560)
    at deepEnhancer (mobx.module.js:401)
    at new ObservableValue (mobx.module.js:1031)
    at ObservableObjectAdministration../node_modules/mobx/lib/mobx.module.js.ObservableObjectAdministration.addObservableProp (mobx.module.js:4209)
    at mobx.module.js:453
    at decorate (mobx.module.js:363)
    at decoratorFactory (mobx.module.js:384)
    at observableDecorator (mobx.module.js:463)
```
I can't figure out what this is coming from other than somewhere inside mobx. My extension works fine with Lens 4.0.8. Any ideas?

docs/extensions/guides/protocol-handlers.md
Nokel81: It is, but the api docs generator makes everything lowercase because URLs are case-insensitive.

src/extensions/registries/protocol-handler-registry.ts
Nokel81: Why is it over the top? it is a schema in so far as you can describe the shape of a URL.

For instance this is one that I use internally:

```
/:extPubName(\@[A-Za-z0-9_]+)?/:extName`
```

which defines a two part segment where the first must match the regex `/@[A-Za-z0-9_]+/` but is also optional.

Nokel81:
@stefcameron are those changes on a branch that I can use to debug the issue? Thanks for checking it out though.

docs/extensions/guides/protocol-handlers.md
stefcameron: But this isn't in a URL. See "The field -> protocolhandlers <- exists both on..."

Nokel81: True, however it is an "inter-doc" link pointing to a generated markdown file. I figured out what it would be by looking at https://docs.k8slens.dev/latest/extensions/api/classes/lensmainextension/#appmenus which is also all lower case even though if you go to it you will see that the field itself is `appMenus`

src/extensions/registries/protocol-handler-registry.ts
stefcameron: I get why it's not _just_ a plain path and there's a schema to it, but react-router, for example, still just calls it a "path": `<Route path="/users/profile/:id">` Just seems simpler in terms of naming. I thought I'd point that out as a comparison.

stefcameron:
> @stefcameron are those changes on a branch that I can use to debug the issue? Thanks for checking it out though.

Yes, they're on the [protocol](https://github.com/Mirantis/lens-extension-cc/tree/protocol) branch.

docs/extensions/guides/protocol-handlers.md
stefcameron: 🤔 I'm not sure we're looking at the same piece of text. I acknowledge that the _anchor_ or _fragment_ in `lensmainextension/#protocolhandlers` and `lensrendererextension/#protocolhandlers` will be all lowercase, but you're telling me that the property name in plain text that precedes it in `` "The field `protocolhandlers` exists both on" `` also has to be lowercase because that somehow becomes a link? I don't mean to press the issue; just seems strange, and looks like a typo, but if that really is the case, then okidoki. 🤷‍♂️

Nokel81: oh, sorry my bad. I was too narrow focused on the fragment.

Nokel81:
@stefcameron Just removed the assignment to the `protocolHandlers` field in your extension and I still see that mobx error. I will do a bit more investigation tomorrow. But I am very confident that it doesn't have something to do with this PR as none of my code is being run until a protocol `open` event happens. But then again I might be wrong.

stefcameron:
> @stefcameron Just removed the assignment to the `protocolHandlers` field in your extension and I still see that mobx error. I will do a bit more investigation tomorrow. But I am very confident that it doesn't have something to do with this PR as none of my code is being run until a protocol `open` event happens. But then again I might be wrong.

Thanks for looking. I did some more debugging and I was finally able to isolate the issue to this piece of code in my extension:

https://github.com/Mirantis/lens-extension-cc/blob/protocol/src/renderer.tsx#L44-L51

I have no idea why this is causing an issue with mobx on your branch but doesn't with a Prod build of Lens (e.g. 4.0.8 all the way back to 4.0.0 alphas and betas before that). It seems mobx goes berserk here trying to drill into the child tree.

At first, I thought it was maybe something to do with Emotion, since `<Item>` there is an Emotion styled component, so I took Emotion out of the equation:
```
  statusBarItems = [
    {
      item: (
        <div
          className="flex align-center gaps"
          style={{
            color: itemColor,
            cursor: 'pointer',
            fontSize: 'var(--font-size-small)',
            padding: '2px 4px', // same as used for active workspace indicator on the left corner
          }}
          onClick={() => this.navigate(addRoute)}
        >
          {/* <a
            style={{
              // Lens app.css adds a bottom border to anchors without href attributes
              borderBottom: 'none'
            }}
            title={strings.extension.statusBar['label']()}
          >
            <ContainerCloudIcon fill={itemColor} />
          </a> */}
          foo
        </div>
      ),
    },
  ];
```
But that still fails:
```
Error: [mobx] Dynamic observable objects cannot be frozen
    at invariant (http://localhost:64999/build/LensDev.js:339434:15)
    at fail (http://localhost:64999/build/LensDev.js:339429:5)
    at Object.preventExtensions (http://localhost:64999/build/LensDev.js:342393:9)
    at Function.freeze (<anonymous>)
    at setInitialDOMProperties (http://localhost:64999/build/LensDev.js:430040:18)
    at setInitialProperties (http://localhost:64999/build/LensDev.js:430281:3)
    at finalizeInitialChildren (http://localhost:64999/build/LensDev.js:431347:3)
    at completeWork (http://localhost:64999/build/LensDev.js:440616:17)
    at completeUnitOfWork (http://localhost:64999/build/LensDev.js:443956:16)
    at performUnitOfWork (http://localhost:64999/build/LensDev.js:443928:5)
```
Turns out mobx doesn't like the `style` object it gets. If I change the code to this:
```
  statusBarItems = [
    {
      item: (
        <div
          className="flex align-center gaps"
          // style={{
          //   color: itemColor,
          //   cursor: 'pointer',
          //   fontSize: 'var(--font-size-small)',
          //   padding: '2px 4px', // same as used for active workspace indicator on the left corner
          // }}
          onClick={() => this.navigate(addRoute)}
        >
          <a
            // style={{
            //   // Lens app.css adds a bottom border to anchors without href attributes
            //   borderBottom: 'none'
            // }}
            title={strings.extension.statusBar['label']()}
          >
            <ContainerCloudIcon fill={itemColor} />
          </a>
        </div>
      ),
    },
  ];
```
It works. I get my status bar icon item back (but it's not styled properly).

This is new behavior, and I can't explain it. But my code certainly hasn't changed. So something else has changed under it (i.e. in Lens).

stefcameron:
Looks like I can bring it down to this:
```
  statusBarItems = [
    {
      item: (
        <div
          className="flex align-center gaps"
          title={strings.extension.statusBar['label']()}
          onClick={() => this.navigate(addRoute)}
        >
          <ContainerCloudIcon fill={itemColor} />
        </div>
      ),
    },
  ];
```
And the styling looks right, so I'm able to update my code to _avoid_ the issue, but I think there's a deeper bug there that might be a problem. Seems like the easiest way to reproduce it is to just add some inline styles to the `<div>` and you get a nice exception on extension load thanks to mobx and whatever it's trying to do...

stefcameron:
After that, I'm happy to report that WRT to this PR, the updated API is working with my extension (with my updates on the `protocol` branch). 🎉

Nokel81:
> After that, I'm happy to report that WRT to this PR, the updated API is working with my extension (with my updates on the `protocol` branch). 🎉 

Thanks for the detailed right up. I'll see if I can make a small reproduction of this error tomorrow and then file a bug.

electron-builder.yml
```diff
@@ -0,0 +1,2 @@
+fileAssociations:
+  - lens
```
jakolehm: Need to verify if this is enough for `snap` build.

src/extensions/extension-loader.ts
```diff
@@ -52,6 +52,20 @@ export class ExtensionLoader {
...
+
+    for (const [, val] of this.instances.toJS()) {
+      if (val.isBundled) {
+        continue;
+      }
+
+      res.set(val.manifest.name, val);
+    }
+
+    return res;
```
jakolehm:
```suggestion
    const extensions = new Map();

    for (const [, val] of this.instances.toJS()) {
      if (val.isBundled) {
        continue;
      }

      extensions.set(val.manifest.name, val);
    }

    return extensions;
```

src/extensions/lens-renderer-extension.ts
```diff
@@ -29,8 +29,7 @@ export class LensRendererExtension extends LensExtension {
   /**
    * Defines if extension is enabled for a given cluster. Defaults to `true`.
    */
-  // eslint-disable-next-line unused-imports/no-unused-vars-ts
   async isEnabledForCluster(cluster: Cluster): Promise<Boolean> {
-    return true;
+    return (void cluster) || true;
```
jakolehm, line 35: Is this just to make `eslint` happy?

src/main/index.ts
```diff
@@ -34,126 +35,132 @@ let clusterManager: ClusterManager;
...
+  .on("will-quit", (event) => {
+  // Quit app on Cmd+Q (MacOS)
+    logger.info("APP:QUIT");
+    appEventBus.emit({name: "app", action: "close"});
+    event.preventDefault(); // prevent app's default shutdown (e.g. required for telemetry, etc.)
+    clusterManager?.stop(); // close cluster connections
+
+    return; // skip exit to make tray work, to quit go to app's global menu or tray's menu
+  })
+  .on("open-url", (event, rawUrl) => {
```
jakolehm: This kind of change to coding style makes huge diff which is really hard to review. Maybe keep the old style just to make the diff smaller?

src/main/protocol-handler/router.ts
```diff
@@ -0,0 +1,103 @@
...
+      const url = new Url(rawUrl, true);
+
+      if (url.protocol.toLowerCase() !== "lens:") {
+        throw new proto.RoutingError(proto.RoutingErrorType.INVALID_PROTOCOL, url);
+      }
+
+      logger.info(`${proto.LensProtocolRouter.LoggingPrefix}: routing ${url.toString()}`);
+
+      switch (url.host) {
+        case "internal":
```
jakolehm: Does this mean non-extension protocol URLs are like `lens://internal`? Maybe consider more "friendly" name because these links might also be published / used from outside of the app. One candidate is that we route everything except `extensions` to internal.

src/extensions/lens-renderer-extension.ts
Nokel81: yes

src/main/index.ts
Nokel81: Sounds good, I can make a small PR later

src/main/protocol-handler/router.ts
Nokel81: do you mean ignoring the `host` of the URI if it is not `extension` or have some way to route based on `host` as well as `pathname`? Or could we change this to something like `app`? The non-extension protocol URIs are `lens://app`.

src/extensions/extension-loader.ts
Nokel81: Sounds good

electron-builder.yml
Nokel81: Is `snap` linux or windows? There is an integration test (that should be run on all platforms) that verifies that the application in indeed getting the `open-url` event on all platforms. But yes more manual testing would be good.

jakolehm: Snap as in https://snapcraft.io/. Snap apps are sandboxed and might require additional configuration for host level protocol handlers (we can create a separate issue about verifying this).

src/main/protocol-handler/router.ts
jakolehm: I meant route based on `host` + `pathname` but if that's complicated then `lens://app` sounds also good.

Nokel81: It is more complicated since `host` follow different rules from `pathname` and it doesn't gain us anything in terms of flexibility (imo).

jakolehm: Ok, let's use `lens://app`.

electron-builder.yml
Nokel81: Okay sounds good.

src/main/protocol-handler/router.ts
stefcameron: 👍

jakolehm:
@Nokel81 tests are still failing.

jakolehm:
@Nokel81 integration tests seem to fail (timeout).

Nokel81:
Update: After reading a *lot* of github issues about why this wasn't working for linux I have come to the conclusion that it is tricky and needs a lot of OS specific items.

https://github.com/electron/electron/issues/10837 says that `setAsDefaultProtocolClient` works for `deb` and `rpm` packages but not for `AppImage`.

Though https://github.com/shiftkey/desktop/issues/10 might imply that it indeed can be done even for AppImages.

I need to test this out a bit more though and it seems like just running the the unpacked version might also not do the work. Will continue to investigate on Monday.

Nokel81:
Update: I have done manual testing and see that the interception of protocol routing is working on `linux` (tested debian) and on `macOS`. Still need to test if my implementation works on `windows`.

Given how early in the process's lifetime the routing takes place there is almost certainly a race condition between the event happening and the routing taking place. This probably is a candidate for a followup PR.

Lastly, I have not been able to get the integration test to pass on `linux`. It passes on `macOS` but whenever I run the integration tests I get multiple instances of Lens running at the same time (even if only the protocol handler describe is being run by jest (with a `describe.only(...)`). This definitely needs to be figured out before merging this PR.

integration/__tests__/app.tests.ts
```diff
@@ -87,6 +113,19 @@ describe("Lens integration tests", () => {
       await app.client.waitUntilTextExists("h2", "Add Cluster");
     });
 
+    describe("protocol app start", () => {
+      it("should handle opening lens:// links", async () => {
+        const res = await open("lens://app/foobar");
+
+        console.log(res);
+
+        await utils.waitForLogsToContain(app, {
```
jakolehm: This might not be a good idea. Infinite loop behind this util will cause problems. Could we have few (one?) real handlers that we can test here. For example `lens://app/preferences` which would open Preferences view.

package.json
```diff
@@ -210,8 +217,10 @@
     "mobx-observable-history": "^1.0.3",
     "mobx-react": "^6.2.2",
     "mock-fs": "^4.12.0",
+    "moment": "^2.26.0",
     "node-pty": "^0.9.0",
     "npm": "^6.14.8",
+    "open": "^7.3.1",
```
jakolehm: Should this be in `devDependencies`?

Nokel81: Good idea

integration/__tests__/app.tests.ts
Nokel81: How is the infinite loop any worse then for `app.client.waitUntilTextExists`? Sure I can add a single real handler too.

jakolehm: I see lot's of integration failures that leave an app instance behind (we can really have just one instance running)... I'm suspecting it's because this loop never dies.

Nokel81: having commented out the tests that have this code I still see those instances behind so I doubt that it has to do with that code (plus I have added a fail fast technique to stop the loop when the test ends (from timeout))

docs/extensions/guides/protocol-handlers.md
```diff
@@ -0,0 +1,57 @@
+# Lens Protocol Handlers
```
jim-docker, line 1: This is not really a guide (there's no code here showing how to add a handler to a main or renderer extension class). Could be done in another PR though.

extensions/example-extension/package-lock.json
```diff
@@ -626,7 +626,684 @@
     },
     "@k8slens/extensions": {
       "version": "file:../../src/extensions/npm/extensions",
-      "dev": true
+      "dev": true,
```
jim-docker: Is all this necessary for protocol handlers?

src/common/protocol-handler/router.ts
```diff
@@ -0,0 +1,253 @@
...
+import { LensExtension } from "../../extensions/lens-extension";
+
+// IPC channel for protocol actions. Main broadcasts the open-url events to this channel.
+export const ProtocolHandlerIpcPrefix = "protocol-handler";
+
+export const ProtocolHandlerInternal = `${ProtocolHandlerIpcPrefix}:internal`;
+export const ProtocolHandlerExtension= `${ProtocolHandlerIpcPrefix}:extension`;
+
+/**
+ * These two names are long and cubersome by design so as to decrease the chances
```
jim-docker: cumbersome (sp)

src/main/protocol-handler/router.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    return broadcastMessage(proto.ProtocolHandlerExtension, rawUrl);
+  }
+
+  /**
+   * Add a function to the list which will be sequentially called if an extension
+   * is not found while routing to the extensions
+   * @param handler A function that tries to find an extension
+   */
+  public addMissingExtensionHandler(handler: FallbackHandler): void {
```
jim-docker, line 109: I don't understand the "missing extension handlers" concept, and this function is not called in the existing code base. When would this be called and by whom?

src/main/window-manager.ts
```diff
@@ -15,6 +15,9 @@ export class WindowManager extends Singleton {
   protected windowState: windowStateKeeper.State;
   protected disposers: Record<string, Function> = {};
 
+  @observable mainViewInitiallyLoaded = false;
+  whenLoaded = when(() => this.mainViewInitiallyLoaded);
```
jim-docker: Is this related to protocol handlers? Maybe you can add a comment here about what impact this has (i.e. what happens when `this.mainViewInitiallyLoaded` gets set to true)

jim-docker (commented):
Seems to match an extension that doesn't exist. For
```
open "lens://extension/nosuchextension/nosuchpath"
```
the logs show:
```
[0] info: [PROTOCOL ROUTER]: routing lens://extension/nosuchextension/nosuchpath
[0] info: [PROTOCOL ROUTER]: Extension nosuchextension matched, but not installed
```

src/main/protocol-handler/router.ts
Nokel81: This is something that existed from the hackweek, it is true that `addMissingExtensioHandler` isn't called but the list of handlers is.

This is there to facilitate Lens trying to figure out why an extension doesn't have any handlers. Such as:
- it isn't enabled
- it exists (in some extension library) but isn't installed

I made the decision not to include any of those in this PR as I felt it was out of scope of those. But I left the mechanism in as it was a good constraint for me while making sure that the internal design and implementation went smoothly.

src/main/window-manager.ts
Nokel81: Yup, this is needed to make sure that the `renderer` has loaded before broadcasting to the renderer because before this is set the renderer is not guaranteed to be in existence.

Sure I can add that as a comment.

extensions/example-extension/package-lock.json
Nokel81: Nope, thanks for the catch

docs/extensions/guides/protocol-handlers.md
Nokel81: I have added a code example. Thanks for catching this (as I had not yet exported the types at all)

docs/extensions/guides/protocol-handlers.md
```diff
@@ -15,6 +15,23 @@ Handlers will be run in both `main` and `renderer` in parallel with no synchroni
...
+}
+
+export default class ExampleExtensionMain extends LensMainExtension {
+  protocolHandlers = [
+    pathSchema: "/",
+    handler: rootHandler,
+  ]
+}
+```
+
```
jim-docker, line 34: You could also mention the specific URI that would result in calling rootHandler, like using
```
open "lens://extension/example_extension/
```
giving an end-to-end example

docs/getting-started/README.md
```diff
@@ -28,6 +28,28 @@ Review the [System Requirements](/supporting/requirements/) to check if your com
 
 See the [Download Lens](https://github.com/lensapp/lens/releases) page for a complete list of available installation options.
 
+After installing Lens manually (not using a package manager file such as `.deb` or `.rpm`) the following will need to be done in allow protocol handling.
```
jim-docker:
```suggestion
After installing Lens manually (not using a package manager file such as `.deb` or `.rpm`) the following will need to be done to allow protocol handling.
```

integration/helpers/utils.ts
```diff
@@ -110,6 +110,7 @@ export async function tearDown(app: Application) {
 
   try {
     process.kill(pid, "SIGKILL");
+    spawnSync("killall lens");
```
jim-docker: Will this kill my Lens 4.0.8 that I might have open while running the integration tests?

jim-docker (dismissed):

integration/helpers/utils.ts
Nokel81: Yes, thanks for reminding me. I added this for testing purposes and should remove it.

jim-docker (dismissed):

jakolehm (dismissed):

jim-docker (dismissed):

src/main/app-updater.ts
```diff
@@ -26,7 +27,7 @@ function handleAutoUpdateBackChannel(event: Electron.IpcMainEvent, ...[arg]: Upd
  * starts the automatic update checking
  * @param interval milliseconds between interval to check on, defaults to 24h
  */
-export function startUpdateChecking(interval = 1000 * 60 * 60 * 24): void {
+export const startUpdateChecking = once(function (interval = 1000 * 60 * 60 * 24): void {
```
jim-docker, line 30: Does this fix an unrelated issue? Or just refactoring while accommodating protocol handling?

src/main/index.ts
```diff
@@ -37,30 +38,54 @@ let windowManager: WindowManager;
 
 app.setName(appName);
 
+logger.info("📟 Setting as Lens as protocol client for lens://");
```
jim-docker, line 41:
```suggestion
logger.info("📟 Setting Lens as protocol client for lens://");
```

jim-docker (approved):

src/main/app-updater.ts
Nokel81: It was fixing an issue due to the changes related to `src/main/index.ts:158`. Namely, if `"renderer:loaded"` gets called multiple times `startUpdateChecking` shouldn't be called on the subsequent events.

src/main/index.ts
Nokel81: Good catch, though I will do that in a follow up PR so as to not need new approves.

src/main/app-updater.ts
jim-docker: I figured, but is it still a bug otherwise? If so, shouldn't it really be in its own PR?

Nokel81: I figured that I should fix it in this PR since it was in this PR that I changed how the `renderer` lets `main` know it has been loaded.

stefcameron:
🎉 It's finally GREEN!!! 💚

jakolehm (approved):
LGTM. Let's keep improving this in follow-up PRs.

PR #1950, opened 2021-01-13
Prevent initializing clusters multiple times

nevalla:
`cluster.init()` takes some time and that leads to the situation where `ClusterManager` can init the same cluster multiple times. This PR will add boolean `initializing` property to `Cluster` and `ClusterManager` checks that too when initializing a cluster.

### DIFF
--- a/src/main/cluster-manager.ts
+++ b/src/main/cluster-manager.ts
@@ -14,7 +14,7 @@ export class ClusterManager extends Singleton {
     // auto-init clusters
     autorun(() => {
       clusterStore.enabledClustersList.forEach(cluster => {
-        if (!cluster.initialized) {
+        if (!cluster.initialized && !cluster.initializing) {
           logger.info(`[CLUSTER-MANAGER]: init cluster`, cluster.getMeta());
           cluster.init(port);
         }
--- a/src/main/cluster.ts
+++ b/src/main/cluster.ts
@@ -37,7 +37,6 @@ export type ClusterRefreshOptions = {
 };
 
 export interface ClusterState {
-  initializing: boolean;
   initialized: boolean;
   enabled: boolean;
   apiUrl: string;
@@ -579,7 +578,6 @@ export class Cluster implements ClusterModel, ClusterState {
    */
   getState(): ClusterState {
     const state: ClusterState = {
-      initializing: this.initializing,
       initialized: this.initialized,
       enabled: this.enabled,
       apiUrl: this.apiUrl,
--- a/src/main/cluster.ts
+++ b/src/main/cluster.ts
@@ -37,6 +37,7 @@ export type ClusterRefreshOptions = {
 };
 
 export interface ClusterState {
+  initializing: boolean;
   initialized: boolean;
   enabled: boolean;
   apiUrl: string;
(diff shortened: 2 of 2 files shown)


src/main/cluster.ts
```diff
@@ -37,6 +37,7 @@ export type ClusterRefreshOptions = {
 };
 
 export interface ClusterState {
+  initializing: boolean;
```
jakolehm: Do we really need to expose this state?

nevalla: Seems that we don't. Will remove it.

nevalla: Removed.

jakolehm (dismissed):

src/main/cluster.ts
```diff
@@ -76,6 +76,7 @@ export class Cluster implements ClusterModel, ClusterState {
    * If extension sets this it needs to also mark cluster as enabled on activate (or when added to a store)
    */
   public ownerRef: string;
+  public initializing = false;
```
ixrock: Probably should be observable field since participating in `autorun` in cluster-manager.

ixrock (dismissed):

src/main/cluster-manager.ts
```diff
@@ -14,7 +14,7 @@ export class ClusterManager extends Singleton {
     // auto-init clusters
     autorun(() => {
       clusterStore.enabledClustersList.forEach(cluster => {
-        if (!cluster.initialized) {
+        if (!cluster.initialized && !cluster.initializing) {
```
ixrock, line 17: Can't get why only `initialized` not enough? They are both called in `cluster.init()` after some `await`s..

src/main/cluster.ts
```diff
@@ -273,10 +274,12 @@ export class Cluster implements ClusterModel, ClusterState {
    */
   @action async init(port: number) {
     try {
+      this.initializing = true;
       this.contextHandler = new ContextHandler(this);
       this.kubeconfigManager = await KubeconfigManager.create(this, this.contextHandler, port);
       this.kubeProxyUrl = `http://localhost:${port}${apiKubePrefix}`;
       this.initialized = true;
+      this.initializing = false;
```
jim-docker: set this in a `finally` block? As is, if `KubeconfigManager.create()` throws an error `this.initializing` will stay as `true`.

Nokel81: That is probably desired (imo) since if it fails to be created then it isn't finished being `initialized`

jim-docker: That doesn't seem to be the spirit of how `initializing` is used here. In `cluster-manager.ts` it is used to avoid calling `cluster.init()` while it is already running. If `this.initializing` remains as `true` after an error then `cluster.init()` will not get called again (according to the logic added in `cluster-manager.ts`). It's moot if it doesn't retry, or the `cluster` object gets recreated before trying again. I'm not sure what it does.

jakolehm: I agree with @jim-docker , `finally` block would be better.

nevalla: Yes, good catch! Fixed.

src/main/cluster.ts
nevalla: Changed as observable.

nevalla:
@jakolehm @ixrock I've changed the implementation based on the feedback, PTAL.

ixrock (approved):

Nokel81 (approved):

PR #1957, opened 2021-01-14
Initial command palette feature

### LINKED ISSUE #1872: Refactor workspace management to "command palette" like window

Workspaces could be managed from a "command palette" (https://code.visualstudio.com/docs/getstarted/userinterface#_command-palette). This would be also initial work for full-blown command palette framework which would provider better keyboard support.

jakolehm:
Adds command palette UI & extension API (similar to what VS Code has). Also adds some common commands to global & cluster scope so that we can actually see the idea is working.

image

Fixes #1872


integration/__tests__/command-palette.tests.ts
```diff
@@ -0,0 +1,36 @@
+import { Application } from "spectron";
```
jakolehm, line 1: Refactored integration tests to separate files so that they are easier to understand & run in a logical units.

src/common/ipc.ts
```diff
@@ -15,30 +18,34 @@ export async function requestMain(channel: string, ...args: any[]) {
...
-export function broadcastMessage(channel: string, ...args: any[]) {
+export async function broadcastMessage(channel: string, ...args: any[]) {
   const views = (webContents || remote?.webContents)?.getAllWebContents();
 
   if (!views) return;
 
-  views.forEach(webContent => {
+  let subFrames: Promise<ClusterFrameInfo[]>;
+
+  if (ipcRenderer) {
```
jakolehm, line 29: Ipc did not broadcast to frames if broadcaster was renderer. This fixes it.

src/common/workspace-store.ts
```diff
@@ -243,6 +243,7 @@ export class WorkspaceStore extends BaseStore<WorkspaceStoreModel> {
       throw new Error(`workspace ${id} doesn't exist`);
     }
     this.currentWorkspaceId = id;
+    clusterStore.setActive(null);
```
jakolehm: We have to reset active cluster if workspace changes. Otherwise UX is really confusing if active cluster is from non-active workspace.

src/extensions/renderer-api/components.ts
```diff
@@ -13,6 +13,9 @@ export * from "../../renderer/components/select";
 export * from "../../renderer/components/slider";
 export * from "../../renderer/components/input/input";
 
+// command-overlay
+export { CommandOverlay } from "../../renderer/components/command-palette";
```
jakolehm, line 17: `CommandOverlay` or `commandOverlay` (singleton) 🤔 .... naming is hard.

Nokel81: `CommandOverlay`, since imo we shouldn't have `commandOverlay` as that basically nullifies our "singleton" code.

src/common/workspace-store.ts
Nokel81: I thought we had a PR that saved the previous active workspace's active cluster

src/extensions/renderer-api/components.ts
jakolehm: I meant that should we export `CommandOverlay` (class) or `commandOverlay` (singleton class instance).

Nokel81: So looking at `renderer/components/command-palette` I don't even see an instance. Also given that `CommandOverlay` only has static function it won't work at all to export an instance.

integration/__tests__/command-palette.tests.ts
Nokel81: Shouldn't that be its own PR?

src/common/ipc.ts
```diff
@@ -15,30 +18,34 @@ export async function requestMain(channel: string, ...args: any[]) {
...
+    subFrames = requestMain(subFramesChannel);
+  } else {
+    subFrames = getSubFrames();
+  }
```
Nokel81:
```suggestion
  const subFrames: Promise<ClusterFrameInfo[]> = ipcRenderer
    ? requestMain(subFramesChannel)
    : getSubFrames();
```

src/extensions/registries/command-registry.ts
```diff
@@ -0,0 +1,37 @@
...
+  title: string;
+  scope: "cluster" | "global";
+  action: (context: CommandContext) => void;
+  isActive?: (context: CommandContext) => boolean;
+}
+
+export class CommandRegistry extends BaseRegistry<CommandRegistration> {
+  @action
+  add(items: CommandRegistration | CommandRegistration[], extension?: LensExtension) {
+    const itemArray = [items].flat() as CommandRegistration[];
```
Nokel81: Does typescript not correctly infer `CommadRegistration[]`?
```suggestion
    const itemArray = [items].flat();
```

src/renderer/components/+cluster-settings/components/cluster-workspace-setting.tsx
```diff
@@ -18,10 +16,7 @@ export class ClusterWorkspaceSetting extends React.Component<Props> {
       <>
         <SubTitle title="Cluster Workspace"/>
         <p>
-          Define cluster{" "}
-          <Link to={workspacesURL()}>
-            workspace
-          </Link>.
+          Define cluster workspace.
```
Nokel81, line 19: Do we not want changing workspaces here?

src/renderer/components/+workspaces/edit-workspace.tsx
```diff
@@ -0,0 +1,82 @@
...
+  condition: ({ required }) => required,
+  message: () => `Workspace with this name already exists`,
+  validate: (value) => {
+    const current = workspaceStore.currentWorkspace;
+
+    if (current.name === value.trim()) {
+      return true;
+    }
+
+    return !workspaceStore.enabledWorkspacesList.find((workspace) => workspace.name === value);
```
Nokel81, line 18: Why should we allow creating of a namespace with the same name as one that isn't enabled? That seems weird since when the disabled namespace is finally enabled there would be two.

src/renderer/components/+workspaces/edit-workspace.tsx
```diff
@@ -0,0 +1,82 @@
...
+  }
+};
+
+interface EditWorkspaceState {
+  name: string;
+}
+
+@observer
```
Nokel81, line 26:
```suggestion
```
This class has no `observables` so it shouldn't be `observer`

src/renderer/components/+workspaces/remove-workspace.tsx
```diff
@@ -0,0 +1,72 @@
...
+        label: `Remove Workspace`,
+        primary: false,
+        accent: true,
+      },
+      ok: () => {
+        workspaceStore.removeWorkspace(workspace);
+
+        if (workspace.id === workspaceStore.currentWorkspaceId) {
+          workspaceStore.setActive(workspaceStore.enabledWorkspacesList[0].id);
+        }
```
Nokel81: This duplicates work done in `removeWorkspace`, is there a good reason why it was added?

src/renderer/components/+workspaces/workspace-menu.tsx
```diff
@@ -1,66 +0,0 @@
-import "./workspace-menu.scss";
```
Nokel81, line 1: These chages to workspaces seems to nulify the changes to #1927 is that true?

src/renderer/components/command-palette/command-container.tsx
```diff
@@ -0,0 +1,97 @@
...
+        console.log("run-action", commandId);
+        const command = this.findCommandById(commandId);
+
+        if (command) {
+          this.runCommand(command);
+        }
+      });
+    } else {
+      subscribeToBroadcast("command-palette:open", () => {
+        CommandOverlay.open(<CommandDialog />);
```
Nokel81, line 69: Why have a variable that stores a react node instead of conditionally rendering `ComandDialog`? (And thus having "@observable isVisible: boolean" field).

src/renderer/components/command-palette/command-dialog.tsx
```diff
@@ -0,0 +1,81 @@
...
+      }
+
+      if (!command.isActive) {
+        return true;
+      }
+
+      try {
+        return command.isActive(context);
+      } catch(e) {
+        console.error(e);
```
Nokel81, line 36: Maybe add some context:

```suggestion
      } catch(error) {
        console.error("[COMMAND-DIALOG]: filtering comand failed", { error } );
```

src/renderer/components/+workspaces/edit-workspace.tsx
jakolehm: This is a bug, good catch!

src/renderer/components/+workspaces/workspace-menu.tsx
jakolehm: Kind of, as the to-be-fixed issue states this PR refactors workspace management to command palette. We need to think again if we really need also a more traditional UI for this.

src/renderer/components/+workspaces/remove-workspace.tsx
jakolehm: I think this duplicate code was there because this is just c&p from the existing logic. But I'll cleanup this 👍

src/renderer/components/command-palette/command-container.tsx
jakolehm: `@observable visible` is a leftover, I'll remove it.

src/renderer/components/+cluster-settings/components/cluster-workspace-setting.tsx
jakolehm: Previously the link was to the workspace page which is now gone. It didn't really add any value to this view.

src/renderer/components/+workspaces/edit-workspace.tsx
jakolehm: But it uses observables coming outside.. I thought we always need `observer` if a component needs react to _any_ observable 🤔 A snippet from Mobx docs:

> For observer to work, it doesn't matter how the observables arrive in the component, only that they are read. 

@ixrock wdyt?

src/renderer/components/command-palette/command-dialog.tsx
jakolehm: Yes, added.

src/extensions/registries/command-registry.ts
jakolehm: Idk why I had this, fixed.

src/common/ipc.ts
jakolehm: Much better, changed.

src/extensions/renderer-api/components.ts
jakolehm: > So looking at `renderer/components/command-palette` I don't even see an instance. Also given that `CommandOverlay` only has static function it won't work at all to export an instance.

It does not have an instance because I chose to use static functions. I was just wondering should it have 🙂 We probably should use whatever is more consistent with our current extension API.

integration/__tests__/command-palette.tests.ts
jakolehm: I would prefer to keep it here. It was almost impossible to refactor/add integration tests when they were in one monolithic test suite (which was also broken at the time of writing).

src/renderer/components/command-palette/command-container.tsx
```diff
@@ -0,0 +1,96 @@
...
+    commandDialogBus.emit({ component });
+  }
+
+  static close() {
+    commandDialogBus.emit({ component: null });
+  }
+}
+
+@observer
+export class CommandContainer extends React.Component<{cluster?: Cluster}> {
```
aleksfront, line 32: Maybe it's better to rename component to `CommandPalette` to avoid generic `Container` part?

aleksfront (commented):
In overall, the palette working great. 2 things I noticed:
* It doesn't close Preferences or Settings page if navigation command executed from Palette
* Shouldn't we go to last active cluster when workspace is switched from Palette?

src/extensions/renderer-api/components.ts
Nokel81: For something like that I think that static function are best

src/common/ipc.ts
```diff
@@ -15,30 +18,30 @@ export async function requestMain(channel: string, ...args: any[]) {
 }
 
 async function getSubFrames(): Promise<ClusterFrameInfo[]> {
```
Nokel81: Why is this function still async? Actually, to be fair why was this function ever async? The very first version of this function was as an async function in PR #1225 but even then it never had any awaits in it.

src/renderer/components/command-palette/command-container.tsx
jakolehm: This is actually an empty container for any custom command-palette look-alike implementations. Not sure about naming here 🤔 .

src/common/ipc.ts
jakolehm: Seems to be a leftover from the original refactoring. Original implementation had to be async but not anymore.. I'll refactor this even more to be in-line with #2046 . The slight issue here is that we need async call for in a `renderer`, sync call can be only used in `main`.

src/common/workspace-store.ts
jakolehm: Good catch! Fixed.

jakolehm:
> In overall, the palette working great. 2 things I noticed:
> 
> * It doesn't close Preferences or Settings page if navigation command executed from Palette

Good catch! Fixed.

> * Shouldn't we go to last active cluster when workspace is switched from Palette?

Fixed.

src/main/cluster-manager.ts
```diff
@@ -12,14 +12,14 @@ export class ClusterManager extends Singleton {
   constructor(public readonly port: number) {
     super();
     // auto-init clusters
-    autorun(() => {
-      clusterStore.enabledClustersList.forEach(cluster => {
+    reaction(() => clusterStore.enabledClustersList, (clusters) => {
```
jakolehm, line 15: I can slice this to a separate PR if needed. Autorun causes side-effects because it triggers on "any" change related to enabled clusters. Reaction only runs when the list itself changes. 

Caused massive headache when adding a cluster suddenly stopped working after I removed a cluster kubeconfig from a filesystem (while testing this PR locally).

src/renderer/components/+workspaces/edit-workspace.tsx
ixrock: Yes, `observer` is required when current class is using any observables in `render()`.

src/renderer/lens-app.tsx
```diff
@@ -11,6 +11,8 @@ import { Notifications } from "./components/notifications";
 import { ConfirmDialog } from "./components/confirm-dialog";
 import { extensionLoader } from "../extensions/extension-loader";
 import { broadcastMessage } from "../common/ipc";
+import {  } from "./components/command-palette/command-dialog";
```
ixrock: What's this? :) Maybe better to create a bundle with re-exports for dialog + container in `components/command-palette/index.ts`?
So import would be `import {CommandContainer, CommandDialog} from "./components/command-palette"`

src/renderer/components/app.tsx
```diff
@@ -196,6 +197,8 @@ export class App extends React.Component {
   }
 
   render() {
+    const cluster = getHostedCluster();
```
ixrock, line 200: Probably we should provide `getHostedCluster()` to react components via https://reactjs.org/docs/context.html
But this is out of PR's scope..

src/common/ipc.ts
```diff
@@ -73,3 +77,9 @@ export function unsubscribeAllFromBroadcast(channel: string) {
     ipcMain.removeAllListeners(channel);
   }
 }
+
+export function bindBroadcastHandlers() {
+  handleRequest(subFramesChannel, async () => {
+    return toJS(await getSubFrames(), { recurseEverything: true });
```
ixrock: I think it's better to use `toJS` in `getSubFrames()`

src/extensions/registries/command-registry.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export class CommandRegistry extends BaseRegistry<CommandRegistration> {
+  @action
+  add(items: CommandRegistration | CommandRegistration[], extension?: LensExtension) {
+    const itemArray = [items].flat();
+
+    const newIds = itemArray.map((item) => item.id);
+    const currentIds = this.getItems().map((item) => item.id);
+
+    const filteredIds = newIds.filter((id) => !currentIds.includes(id));
```
ixrock, line 30: So, currently it's just skipping already registered ids, right? If we gonna expose it to `extension-api` we should validate them here and/or throw error when same id faced?

src/main/cluster-manager.ts
```diff
@@ -12,14 +12,14 @@ export class ClusterManager extends Singleton {
...
+      clusters.forEach((cluster) => {
         if (!cluster.initialized && !cluster.initializing) {
           logger.info(`[CLUSTER-MANAGER]: init cluster`, cluster.getMeta());
           cluster.init(port);
         }
       });
-    });
+    }, { fireImmediately: true });
```
ixrock, line 22:
```suggestion
    }, { 
    fireImmediately: true,
    equals: comparer.structural,
 });
```

src/renderer/components/+workspaces/add-workspace.tsx
```diff
@@ -0,0 +1,57 @@
...
+const uniqueWorkspaceName: InputValidator = {
+  condition: ({ required }) => required,
+  message: () => `Workspace with this name already exists`,
+  validate: value => !workspaceStore.workspacesList.find((workspace) => workspace.name === value),
+};
+
+@observer
+export class AddWorkspace extends React.Component {
+  onSubmit(name: string) {
+    if (name.trim() === "") {
```
ixrock:
```suggestion
    if (!name.trim()) {
```

src/renderer/components/+workspaces/edit-workspace.tsx
```diff
@@ -0,0 +1,82 @@
...
+export class EditWorkspace extends React.Component<{}, EditWorkspaceState> {
+
+  state: EditWorkspaceState = {
+    name: ""
+  };
```
ixrock, line 31:
```suggestion
  @observable name = "";
```

ixrock: And then below:
```suggestion
          onChange={(v) => this.name = v}
```

src/renderer/components/command-palette/command-container.tsx
```diff
@@ -0,0 +1,96 @@
...
+  @observable commandComponent: React.ReactElement;
```
ixrock:
```suggestion
  @observable.ref commandComponent: React.ReactElement;
```

src/renderer/components/command-palette/command-container.tsx
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  private runCommand(command: CommandRegistration) {
+    command.action({
+      cluster: clusterStore.active,
+      workspace: workspaceStore.currentWorkspace
+    });
+  }
+
+  onClusterAction(commandId: string) {
```
ixrock: Not used method.. Some leftovers?

src/renderer/components/command-palette/command-container.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+    if (command) {
+      this.runCommand(command);
+    }
+  }
+
+  componentDidMount() {
+    if (this.props.cluster) {
+      subscribeToBroadcast(`command-palette:run-action:${this.props.cluster.id}`, (event, commandId: string) => {
+        console.log("run-action", commandId);
```
ixrock: Forgotten `console.log`?

src/renderer/components/command-palette/command-container.tsx
```diff
@@ -0,0 +1,96 @@
...
+  private findCommandById(commandId: string) {
+    return commandRegistry.getItems().find((command) => command.id === commandId);
+  }
+
+  private runCommand(command: CommandRegistration) {
+    command.action({
+      cluster: clusterStore.active,
+      workspace: workspaceStore.currentWorkspace
+    });
+  }
```
ixrock, line 56:
```suggestion
  static findCommandById(commandId: string) {
    return commandRegistry.getItems().find((command) => command.id === commandId);
  }

  static runCommand(commandId: string) {
    this.findCommandById(commandId)?.action({
      cluster: clusterStore.active,
      workspace: workspaceStore.currentWorkspace
    });
  }
```

ixrock:
Looks better if we restrict `background` to `<Select/>` area only, wdyt?

**Before:**
image

**After:**
image

src/renderer/components/+workspaces/edit-workspace.tsx
ixrock: I meant yes, if it uses _direct_ observable(s) in `render()` (coming from props/class fields or from upper js-scope). 
Not just other `<SomeChildComponent/>` with observables in render()..

src/renderer/components/command-palette/command-container.tsx
jakolehm: I prefer to keep these private (at least for now).

src/renderer/components/command-palette/command-container.tsx
jakolehm: Yes, removed 👍

src/main/cluster-manager.ts
jakolehm: Why do we need `structural`? I think it should be enough to react only when array changes.

jakolehm:
> Looks better if we restrict background to `<Select/>` area only, wdyt?

Background is also used in other command UIs so it kinda needs to be there. I agree that it looks a bit ugly but we can fix that in a follow up PR.

src/main/cluster-manager.ts
ixrock: Yes, you're right because `clusterStore.enabledClustersList` is `@computed` observable class field.

src/common/ipc.ts
```diff
@@ -80,6 +80,6 @@ export function unsubscribeAllFromBroadcast(channel: string) {
 
 export function bindBroadcastHandlers() {
   handleRequest(subFramesChannel, async () => {
-    return toJS(await getSubFrames(), { recurseEverything: true });
+    return await getSubFrames(), { recurseEverything: true };
```
ixrock: Shouldn't it be just `return getSubFrames()`?
Or even `handleRequest(subFramesChannel, getSubFrames)`.

jakolehm: Oh yes, I was just moving pieces too fast. Thanks for noticing 👍 (it's now fixed)

src/renderer/components/command-palette/command-container.tsx
Nokel81: You can make them `private static`

ixrock: @Nokel81 you cannot use `private static findCommandById` in `private static runCommand`

Nokel81: Yes you can

https://www.typescriptlang.org/play?#code/MYGwhgzhAEBiD29oG8BQ0PQA4CcCWAbmAC4Cm0ExJew0AQmDgBQCUAXNAfHgCYqrpMAXwGZs+ImQpViNenggALVhy69+YsQngA6BsxYBuQRhEnoAO0Y5EAI3grO3Pmk2ZtehcqPmRQoA

src/renderer/lens-app.tsx
nevalla: What is imported on this case?

src/renderer/components/+workloads/workloads.route.ts
```diff
@@ -80,3 +80,5 @@ export const workloadURL: Partial<Record<KubeResource, ReturnType<typeof buildUR
   "jobs": jobsURL,
   "cronjobs": cronJobsURL,
 };
+
+
```
nevalla: nit: extra empty line

nevalla (dismissed):
Generally looking good. Added on minor comment about the code I did not totally understand but if it's working, I'm fine with it now.

src/renderer/components/+workspaces/add-workspace.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { v4 as uuid } from "uuid";
+import { commandRegistry } from "../../../extensions/registries/command-registry";
+import { Input, InputValidator } from "../input";
+import { navigate } from "../../navigation";
+import { CommandOverlay } from "../command-palette/command-container";
+
+const uniqueWorkspaceName: InputValidator = {
+  condition: ({ required }) => required,
+  message: () => `Workspace with this name already exists`,
+  validate: value => !workspaceStore.workspacesList.find((workspace) => workspace.name === value),
```
aleksfront:
```suggestion
  validate: value => !workspaceStore.getByName(value),
```

src/renderer/components/+workspaces/add-workspace.tsx
```diff
@@ -0,0 +1,57 @@
...
+  onSubmit(name: string) {
+    if (!name.trim()) {
+      return;
+    }
+    const workspace = workspaceStore.addWorkspace(new Workspace({
+      id: uuid(),
+      name
+    }));
+
+    workspaceStore.setActive(workspace.id);
```
aleksfront, line 33: Need to check for `undefined` value of `workspace` variable since `addWorkspace` method currently can return `undefined`  which is also not good. So I suggest to refactor `addWorkspace` or put some check before setting active workspace.

The reason is an error occured when trying to create workspace with already occupied name (eg `default`)
Screenshot 2021-02-02 at 10 17 40

src/renderer/components/+workspaces/remove-workspace.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { Select } from "../select";
+import { CommandOverlay } from "../command-palette/command-container";
+
+@observer
+export class RemoveWorkspace extends React.Component {
+  @computed get options() {
+    return workspaceStore.enabledWorkspacesList.filter((workspace) => workspace.id !== WorkspaceStore.defaultId).map((workspace) => {
+      return { value: workspace.id, label: workspace.name };
+    });
+  }
```
aleksfront, line 16:
```suggestion
    return workspaceStore.enabledWorkspacesList.filter((workspace) => workspaceStore.isDefault(workspace.id).map(({ id, name }) => {
      return { value: id, label: name };
    });
  }
```

src/renderer/components/+workspaces/workspaces.tsx
```diff
@@ -1,216 +1,89 @@
...
+import { landingURL } from "../+landing-page";
+import { clusterViewURL } from "../cluster-manager/cluster-view.route";
 
 @observer
-export class Workspaces extends React.Component {
-  @observable editingWorkspaces = observable.map<WorkspaceId, Workspace>();
+export class ChooseWorkspace extends React.Component {
+  private static addActionId = "__add__";
+  private static removeActionId = "__remove__";
+  private static editActionId = "__edit__";
```
aleksfront, line 19: Probably matter of taste but we can use `enum` instead of private static methods here.
https://www.typescriptlang.org/docs/handbook/enums.html

src/renderer/components/+workspaces/workspaces.tsx
```diff
@@ -1,216 +1,89 @@
...
-  };
-
-  onInputKeypress = (evt: React.KeyboardEvent<any>, workspaceId: WorkspaceId) => {
-    if (evt.key == "Enter") {
-      // Trigget input validation
-      evt.currentTarget.blur();
-      evt.currentTarget.focus();
-      this.saveWorkspace(workspaceId);
+    if (clusterId) {
+      navigate(clusterViewURL({ params: { clusterId } }));
```
aleksfront, line 62: Something odd with workspace creation/navigation is happening. Probably this comment relates more to `AddWorkspace` component.

After creating new workspace - a previous one somehow sticked to it. So when you switch to newly created workspace you'll see some cluster opened. Better show than explain:

https://user-images.githubusercontent.com/9607060/106567226-89424280-6542-11eb-9ef0-292d4a5a61de.mp4

src/renderer/components/app.tsx
aleksfront: Good idea about context 👍

src/renderer/components/command-palette/command-container.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { CommandRegistration, commandRegistry } from "../../../extensions/registries/command-registry";
+import { clusterStore } from "../../../common/cluster-store";
+import { workspaceStore } from "../../../common/workspace-store";
+import { Cluster } from "../../../main/cluster";
+
+export type CommandDialogEvent = {
+  component: React.ReactElement
+};
+
+const commandDialogBus = new EventEmitter<[CommandDialogEvent]>();
```
aleksfront, line 19: I'm okay with this for now, but we should reconsider the way how to change children within `CommandOverlay`. It feels a bit strange that we have `commandDialogBus` and some event emitters for this purpose. 

@ixrock Do you see how we can pass children to this component without inner `CommandDialogEvent` manipulations?

src/renderer/components/command-palette/command-dialog.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+        return false;
+      }
+    }).map((command) => {
+      return { value: command.id, label: command.title };
+    }).sort((a, b) => a.label > b.label ? 1 : -1);
```
aleksfront, line 42: Would it be enough to use `sort` without comparator function? We don't using numbers here and sorting items alphabetically.

I mean replace `.sort((a, b) => a.label > b.label ? 1 : -1)` to just `.sort()`

src/renderer/lens-app.tsx
aleksfront: Is it possible to remove this line? Pretty confusing to have it without explanation.

src/renderer/components/command-palette/command-container.tsx
jakolehm: "Public" api for this is `CommandOverlay.open(<component)`. Bus is only used internally.

src/renderer/components/command-palette/command-dialog.tsx
jakolehm: > Would it be enough to use `sort` without comparator function? We don't using numbers here and sorting items alphabetically.
> 
> I mean replace `.sort((a, b) => a.label > b.label ? 1 : -1)` to just `.sort()`

Plain `.sort()` does not sort by labels, or does it?

src/renderer/components/+workspaces/workspaces.tsx
jakolehm: Are we sure this behaviour is not in the master branch? I have seen similar behaviour in the past and that's why this PR originally always reseted active cluster to `null` when active workspace changed (but I reverted that because of code review comments).

src/renderer/components/+workspaces/add-workspace.tsx
jakolehm: Hmm.. onSubmit should not happen if the value is not valid. Will try to reproduce.

src/renderer/components/+workspaces/add-workspace.tsx
jakolehm: Thanks, fixed!

src/renderer/components/+workspaces/add-workspace.tsx
jakolehm: Reproduced, for some reason input still triggers `onSubmit` even if validation fails. Added more protection to add workspace logic.

src/renderer/components/+workspaces/workspaces.tsx
jakolehm: This PR now explicitly sets active cluster to `null` when workspace is added. I feel that we still have a bug somewhere else in workspace switch logic but that probably is a topic for an another PR.

jakolehm:
@nevalla @aleksfront ptal again.

src/renderer/components/+workspaces/workspaces.tsx
aleksfront: I've checked `master` branch. New workspace created without any cluster attached to it.

https://user-images.githubusercontent.com/9607060/106575957-ae887e00-654d-11eb-9ad7-addac55b80d9.mp4

src/renderer/components/command-palette/command-dialog.tsx
aleksfront: Oh, of course, we don't sort raw strings here.

aleksfront (approved):
LGTM

nevalla (approved):
LGTM

aleksfront:
🎆

src/renderer/components/command-palette/command-container.tsx
ixrock: @Nokel81 yes, seems like it introduced in _Typescript 4.1+_, currently is:

image

ixrock: Oh, right, I just checked in editor. Wrong assumption :)

PR #1960, opened 2021-01-14
do not assume that bitnami is guanteed to be in the helm repo list

Nokel81:

### DIFF
--- a/integration/__tests__/app.tests.ts
+++ b/integration/__tests__/app.tests.ts
@@ -5,8 +5,11 @@
   cluster and vice versa.
 */
 import { Application } from "spectron";
-import * as util from "../helpers/utils";
-import { spawnSync } from "child_process";
+import * as utils from "../helpers/utils";
+import { spawnSync, exec } from "child_process";
+import * as util from "util";
+
+export const promiseExec = util.promisify(exec);
 
 jest.setTimeout(60000);
 
@@ -16,7 +19,7 @@ describe("Lens integration tests", () => {
   const BACKSPACE = "\uE003";
   let app: Application;
   const appStart = async () => {
-    app = util.setup();
+    app = utils.setup();
     await app.start();
     // Wait for splash screen to be closed
     while (await app.client.getWindowCount() > 1);
@@ -71,7 +74,7 @@ describe("Lens integration tests", () => {
 
     afterAll(async () => {
       if (app?.isRunning()) {
-        await util.tearDown(app);
+        await utils.tearDown(app);
       }
     });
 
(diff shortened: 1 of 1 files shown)


jim-docker:
I think we should be able to assume bitnami. Is this just a workaround for spectron flakiness?

Nokel81:
No it is a work around for people (including developers) that have manually run `helm repo add ...` before the switch to `bitnami`. Namely, the code in the settings only runs that command for `bitnami` if there are no helm repos listed at all. But I think previously we did that with `stable`. I ran into this on my machine for instance.

jim-docker:
> No it is a work around for people (including developers) that have manually run `helm repo add ...` before the switch to `bitnami`. Namely, the code in the settings only runs that command for `bitnami` if there are no helm repos listed at all. But I think previously we did that with `stable`. I ran into this on my machine for instance.

ah, I see. It would be ideal if the test was dynamic, detecting if there were repo(s) already present, and looking for bitnami if not, which is what the CI would encounter and would be a stronger test.

Nokel81:
> > No it is a work around for people (including developers) that have manually run `helm repo add ...` before the switch to `bitnami`. Namely, the code in the settings only runs that command for `bitnami` if there are no helm repos listed at all. But I think previously we did that with `stable`. I ran into this on my machine for instance.
> 
> ah, I see. It would be ideal if the test was dynamic, detecting if there were repo(s) already present, and looking for bitnami if not, which is what the CI would encounter and would be a stronger test.

Agreed good idea. Will do that

Nokel81:
@jim-docker Done, care to take a look again? Thanks.

integration/__tests__/app.tests.ts
```diff
@@ -93,7 +96,10 @@ describe("Lens integration tests", () => {
       });
 
       it("ensures helm repos", async () => {
-        await app.client.waitUntilTextExists("div.repos #message-bitnami", "bitnami"); // wait for the helm-cli to fetch the bitnami repo
+        const { stdout: reposJson } = await promiseExec("helm repo list -o json");
```
jim-docker, line 99: better than `spawnSync()`, which is already used elsewhere in this file?

integration/__tests__/app.tests.ts
```diff
@@ -93,7 +96,10 @@ describe("Lens integration tests", () => {
...
+        const repos = JSON.parse(reposJson);
+
+        await app.client.waitUntilTextExists("div.repos #message-bitnami", repos[0].name); // wait for the helm-cli to fetch the any repo
```
jim-docker:
```suggestion
        await app.client.waitUntilTextExists("div.repos #message-bitnami", repos[0].name); // wait for the helm-cli to fetch the repo(s)
```

jim-docker (dismissed):

integration/__tests__/app.tests.ts
Nokel81: We probably should be using `exec` more. See this:

https://stackoverflow.com/questions/48698234/node-js-spawn-vs-execute

jim-docker (dismissed):

Nachasic (approved):

jim-docker (approved):

jakolehm:
This is actually throwing errors in our pipeline.

image (2)

Should we revert?

Nokel81:
Let me investigate, I think we can ignore this error and then test if `bitnami` is present specifically. If you want to revert then sure, but I think we should have some modification to this test.

PR #1964, opened 2021-01-15
Fix: chart.digest is the same for all charts and not suited as unique id

ixrock:
**Before:**
image

**After:**
image

### DIFF
--- a/src/renderer/api/endpoints/helm-charts.api.ts
+++ b/src/renderer/api/endpoints/helm-charts.api.ts
@@ -86,7 +86,7 @@ export class HelmChart {
   tillerVersion?: string;
 
   getId() {
-    return this.digest;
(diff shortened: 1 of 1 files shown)


nevalla (approved):

nevalla:
BTW can't reproduce the issue on `master` branch.

image

ixrock:
> BTW can't reproduce the issue on `master` branch.

this is weird.. any ideas why it could happen? (checked again in `master` - still doesn't work as expected)

Nokel81:
Screen Shot 2021-01-15 at 3 43 30 PM

This is what I see. It is mostly a `"stable"` repo problem but we should still fix it.

Nokel81 (approved):

PR #1965, opened 2021-01-15
Enable default workspace on first boot

jakolehm:
Currently first boot does not set `default`  workspace as enabled -> it's not visible in workspace switcher.

### DIFF
--- a/src/common/workspace-store.ts
+++ b/src/common/workspace-store.ts
@@ -58,20 +58,16 @@ export class Workspace implements WorkspaceModel, WorkspaceState {
    * @observable
    */
   @observable ownerRef?: string;
-  /**
-   * Is workspace enabled
-   *
-   * Workspaces that don't have ownerRef will be enabled by default. Workspaces with ownerRef need to explicitly enable a workspace.
-   *
-   * @observable
-   */
-  @observable enabled: boolean;
+
   /**
    * Last active cluster id
    *
    * @observable
    */
   @observable lastActiveClusterId?: ClusterId;
+
+
+  @observable private _enabled: boolean;
 
   constructor(data: WorkspaceModel) {
     Object.assign(this, data);
@@ -81,6 +77,21 @@ export class Workspace implements WorkspaceModel, WorkspaceState {
         this.pushState();
       });
     }
+  }
+
+  /**
+   * Is workspace enabled
+   *
+   * Workspaces that don't have ownerRef will be enabled by default. Workspaces with ownerRef need to explicitly enable a workspace.
+   *
+   * @observable
+   */
+  get enabled(): boolean {
+    return !this.isManaged || this._enabled;
+  }
+
+  set enabled(enabled: boolean) {
+    this._enabled = enabled;
   }
 
   /**
@@ -141,13 +152,11 @@ export class WorkspaceStore extends BaseStore<WorkspaceStoreModel> {
     super({
       configName: "lens-workspace-store",
     });
-    const defaultWorkspace = new Workspace({
+
+    this.workspaces.set(WorkspaceStore.defaultId, new Workspace({
       id: WorkspaceStore.defaultId,
       name: "default"
-    });
-
-    defaultWorkspace.enabled = true;
-    this.workspaces.set(WorkspaceStore.defaultId, defaultWorkspace);
+    }));
   }
 
   async load() {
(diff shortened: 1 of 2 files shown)


nevalla (dismissed):

src/common/workspace-store.ts
```diff
@@ -138,6 +138,7 @@ export class WorkspaceStore extends BaseStore<WorkspaceStoreModel> {
     super({
       configName: "lens-workspace-store",
     });
+    this.workspaces.get(WorkspaceStore.defaultId).enabled = true;
```
Nokel81: Do you mean, "on every boot"?

jim-docker: I think this only happens the first time after Lens install

Nokel81: No, it happens every time Lens is full quit (meaning including the tray icon is gone). A better solution would be to have the default workspace default to true.

jim-docker: By this, I mean the bug

Nokel81: Yes I see, (and I don't remember if a user is currently allowed to disable the default workspace).

Nokel81:
Maybe this is the time to discuss if we really should have a hard, always present, workspace. Why do we not just enforce the requirement that there is always at least one workspace in existence?

src/common/workspace-store.ts
jim-docker: You could initialize it when it is created, on line 193

jim-docker (dismissed):

src/common/workspace-store.ts
jakolehm: Not sure about that.. we cannot pass ` enabled` to `Workspace` constructor because it eats only data that will be persisted.

Nokel81: Well you could do the following:

```typescript
  @observable workspaces = observable.map<WorkspaceId, Workspace>({
    [WorkspaceStore.defaultId]: () => {
      const res = new Workspace({
        id: WorkspaceStore.defaultId,
        name: "default"
      });
      res.enabled = true;
      
      return res;
    }()
  });
```

Nokel81: Or better yes, initialize this in the constructor instead of at the declaration.

jim-docker: I don't get it. `Workspace` has an `enabled` field, and `this.workspaces` is a map of `Workspace`s. Is it because it's observable?

JS/TS complexity starts at 30 for a single statement so it's no wonder that VSCode extension scores our work so high :)

Nokel81: I don't understand. The constructor for `Workspace` does not take `enabled` (though the impl would set it). This is to work around that. I don't know why that constructor doesn't take enabled. It seems strange that it doesn't.

jim-docker: Oh I see, it copies over only `WorkspaceModel` fields

Nokel81: This should be done in the initializer and `Workspace` constructor should take `enabled` as a field.

Nokel81 (dismissed):

src/common/workspace-store.ts
Nokel81: Obstinately yes, though since the code is `Object.assign` it actually copied over everything.

src/common/workspace-store.ts
```diff
@@ -134,10 +134,20 @@ export class WorkspaceStore extends BaseStore<WorkspaceStoreModel> {
...
     super({
       configName: "lens-workspace-store",
     });
+    const defaultWorkspace = new Workspace({
+      id: WorkspaceStore.defaultId,
+      name: "default"
+    });
+
+    defaultWorkspace.enabled = true;
+    this.workspaces.set(WorkspaceStore.defaultId, defaultWorkspace);
```
Nokel81: Definitely better, though I still don't understand why the `Workspace` constructor doesn't allow setting the `enabled` flag to true.

src/common/workspace-store.ts
jim-docker: more hidden complexity :)

Nokel81: Yeah... I don't particularly like `Object.assign` since in the final JS it doesn't do any type checking.

jim-docker (dismissed):
tested

jakolehm:
Borrowed get/set idea from #1967, should be quite clean now.

nevalla (approved):

jim-docker (approved):

PR #1967, opened 2021-01-15
add computed isEnabled getter to Workspace, use it everywhere

Nokel81:

alternative to #1965

### DIFF
--- a/src/common/workspace-store.ts
+++ b/src/common/workspace-store.ts
@@ -84,6 +84,13 @@ export class Workspace implements WorkspaceModel, WorkspaceState {
   }
 
   /**
+   * Is this workspace not manged by an extension or is it managed and explicitly enabled
+   */
+  get isEnabled(): boolean {
+    return !this.isManaged || this.enabled;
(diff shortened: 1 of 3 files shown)


Nokel81:
closes #1965

src/common/workspace-store.ts
```diff
@@ -83,6 +83,13 @@ export class Workspace implements WorkspaceModel, WorkspaceState {
     }
   }
 
+  /**
+   * Is this workspace not manged by an extension or is it managed and explicitly enabled
+   */
+  get isEnabled(): boolean {
```
jakolehm, line 89: Why different name? Should we use `get enabled` and `set enabled` so that api stays the same?

Nokel81: I left the `@observable enabled: boolean` as the current method of setting this. Having an `is***` getter seemed to me to be more in line with how our other getters are named. But I can change it if you want.

jakolehm:
Duplicate of #1965 .

Issue #1968, opened 2021-01-15
Custom installation path


PhilRanzato:
**What would you like to be added**:

Ability to change default installation path

**Why is this needed**:

To install in a custom drive

**Environment you are Lens application on:**

- Kubernetes distribution: []
- Desktop OS: Windows

carlreid:
This is needed in my case also, since I can install development tools in a special folder that won't be blocked by enterprise rules. A workaround for me was to install as normal then copy everything from the default path to my special folder and run it that way.

jakolehm:
Fixed in #1978

PR #1978, opened 2021-01-18
Allow for changing installation directory on Windows

carlreid:
At the cost of the one-click installer, allow for choosing a custom install path.

Solves: #1968

### DIFF
--- a/package.json
+++ b/package.json
@@ -154,7 +154,9 @@
       ]
     },
     "nsis": {
-      "include": "build/installer.nsh"
+      "include": "build/installer.nsh",
+      "oneClick": false,
+      "allowToChangeInstallationDirectory": true	  
     },
     "publish": [
       {


Nokel81 (approved):

package.json
```diff
@@ -154,7 +154,9 @@
       ]
     },
     "nsis": {
-      "include": "build/installer.nsh"
+      "include": "build/installer.nsh",
+      "oneClick": false,
```
jakolehm, line 158: I'm not sure what side-effects this will have (if any)? Can package still be installed without admin rights?

carlreid: It should just be as normal apart from the install wizard showing. According to the [config here](https://www.electron.build/configuration/nsis) and how I read it, then you would be asked for elevation if it was required (because `allowElevation` is true by default).

I'm not sure about any other possible path related issues in the rest of the application though. Unfortunately I couldn't get a local build working to try out the install on my own machine, and was hoping that the CI would spit out an installer I could try which it either doesn't or I couldn't find it.

jakolehm: Ok, let's take this into 4.1 pre-releases so that it can be tested more easily.

jakolehm (approved):

Issue #272, opened 2020-04-18
Logs per workload


aexvir:
**Make logging available per workload**

**Why is this needed**:
Most deployments have more than a single pod, usually, when searching for something, or making some test, it's not always known in which exact pod the request will end up, so it's really useful to tail the logs from all the pods of a workload

It would actually be really nice if it could have all the options that [stern](https://github.com/wercker/stern) has, for example, like filtering per container or allowing to specify exclusions

**Environment you are Lens application on:**

- Kubernetes distribution: GKE, Rancher and EKS, but I don't think it matters
- Desktop OS: macOS

pieterclaerhout:
+1

povils:
Agree, it would be a killer feature!

pauln:
Even a quick way to switch between pods (within a deployment) when viewing logs would be a great help - having to backtrack to the deployment, scroll back down to the list of pods and then drill into the next one and hit the logs again isn't ideal, especially since the list of pods doesn't seem to have a stable sort order.  If your pod names are long enough that they get truncated, it's easy to end up checking the same pod twice instead of checking the next one.

matthewhartstonge:
The one thing I _really_ miss from Kubernetes Dashboard is being able to switch between logs within a deployment.

For example, here we have the dashboard view of logs for our datadog agent deployment:
![image](https://user-images.githubusercontent.com/6119549/91499644-ea237a00-e915-11ea-9f98-b301035e28e3.png)

We can choose between the specific named containers from within the deployment:
![image](https://user-images.githubusercontent.com/6119549/91499700-0cb59300-e916-11ea-97e2-b83a06bf047d.png)

And then easily flick between the logs for a specific container instance:
![image](https://user-images.githubusercontent.com/6119549/91499769-2f47ac00-e916-11ea-8a9f-43ad30b8f02d.png)

hadim:
For what it's worth I am also very interested in that feature.

aleksfront:
Related to #430

JasonGilholme:
+1 for this.  Aggregated logs across all pods in a deployment (or even namespace) would be fantastic.  I just tried out some other offerings that have this feature and it was very easy to see that one pod wasn't co-operating like it should.

VagyokC4:
> +1 for this. Aggregated logs across all pods in a deployment (or even namespace) would be fantastic. I just tried out some other offerings that have this feature and it was very easy to see that one pod wasn't co-operating like it should.

[...]

PR #1984, opened 2021-01-19
Generic logs view with Pod selector

aleksfront:
Adding pod selector grouping by common owner in Log view. This will make easier to look through deployment logs for example by switching between its pods.

https://user-images.githubusercontent.com/9607060/105027050-57b87a00-5a60-11eb-8957-d2bf4a2e8712.mp4

Part of #272


src/renderer/components/+workloads-pods/pods.store.ts
```diff
@@ -32,15 +31,15 @@ export class PodsStore extends KubeObjectStore<Pod> {
...
+  getPodsByOwnerId(workloadId: string): Pod[] {
+    if (!workloadId) return [];
 
     return this.items.filter(pod => {
       const owners = pod.getOwnerRefs();
 
       if (!owners.length) return;
 
-      return owners.find(owner => owner.uid === workload.getId());
+      return owners.find(owner => owner.uid === workloadId);
```
Nokel81:
```suggestion
      return pod.getOwnerRefs().find(owner => owner.uid === workloadId);
```

src/renderer/components/dock/__test__/log-resource-selector.test.tsx
```diff
@@ -0,0 +1,198 @@
+import React from "react";
+import "@testing-library/jest-dom/extend-expect";
+import { render } from "@testing-library/react";
+import selectEvent from "react-select-event";
+
+import { Pod } from "../../../api/endpoints";
+import { LogResourceSelector } from "../log-resource-selector";
+import { LogTabData } from "../log-tab.store";
+
+const dummyPod = {
```
Nokel81: These dummy pods should be extracted out of these test files since they seem to be the same.

src/renderer/components/dock/log-tab.store.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+export class LogTabStore extends DockTabStore<LogTabData> {
+  constructor() {
+    super({
+      storageName: "pod_logs"
+    });
+  }
+
+  public createPodTab({ selectedPod, selectedContainer }: PodLogsTabData) {
```
Nokel81:
```suggestion
  public createPodTab({ selectedPod, selectedContainer }: PodLogsTabData): void {
```

src/renderer/components/dock/log-tab.store.ts
```diff
@@ -0,0 +1,100 @@
...
+    const title = `Pod ${selectedPod.getName()}`;
+
+    this.createLogsTab(title, {
+      pods: pods.length ? pods : [selectedPod],
+      selectedPod,
+      selectedContainer
+    });
+  }
+
+  public createWorkloadTab({ workload }: WorkloadLogsTabData) {
```
Nokel81:
```suggestion
  public createWorkloadTab({ workload }: WorkloadLogsTabData): void {
```

src/renderer/components/dock/log-resource-selector.tsx
```diff
@@ -30,36 +31,73 @@ export const LogResourceSelector = observer((props: Props) => {
...
+      return;
+    }
+
+    const { getContainers, getInitContainers, getAllContainers } = selectedPod;
+
+    save({
+      selectedPod,
+      containers: getContainers(),
+      initContainers: getInitContainers(),
+      selectedContainer: getAllContainers()[0]
```
Nokel81: Why store `containers` and `initContainers` at all? Since we can always get them on demand from `selectedPod`.

src/renderer/components/dock/log-tab.store.ts
aleksfront: I don't think it's necessary. This `void`type is determined automatically by typescript so its visible in IDE. And I'm trying currently to stick with clear names, meaning if function is called like `createPodTab` then it shouldn't return anything. Otherwise it would be named like `getPodTab()` or similar. But I can add `:void` notation by hand if you want.

src/renderer/components/dock/log-resource-selector.tsx
aleksfront: Good call. I’ll make that change.

src/renderer/components/dock/log-tab.store.ts
Nokel81: I guess I view it as part of the function contract. Namely, changing the return value is generally a breaking change so I am of the opinion that they should be always explicitly marked (except for functions like `render**` or maybe `toString`)

jim-docker (dismissed):
LGTM.
I tried it out. I know the pod name is in the (terminal?) tab's title bar so I can always know which pod's log I'm looking at, but the pod tab "select" field cuts the name off if it's too long. Could the full name be shown in a tooltip when you hover over it? And/or, when you click the pod select field to show the dropdown menu could the currently selected one be the initially highlighted item in the dropdown (instead of always highlighting the first item in the dropdown)? Not a big deal...

image

jim-docker:
I noticed when I scaled up my deployment with the log window open, the pod tab did not list the new pod. Also, when I scaled down down it still showed all the removed pods in the dropdown, and when I selected one it did nothing. Probably expected behaviour? No crashes anyway. Closing the log tab and opening again gave expected behaviour.

aleksfront:
@jim-docker Thanks for noticing this. I will add pod watching behavior into this PR.

aleksfront:
Added watcher checking pods store changes. If pod amount gets changed, the tab data gets refreshed also.

* If there's no selected pod available anymore, first pod in list selected
* If there's no any pods left in owner, log tab gets closed.

https://user-images.githubusercontent.com/9607060/105700548-0bff4800-5f1a-11eb-8a4e-e025de309f81.mp4

https://user-images.githubusercontent.com/9607060/105700551-0d307500-5f1a-11eb-9785-bc055dffceb1.mp4

@jim-docker This will fix your scailing issue. Please take another look.

src/renderer/components/+workloads-pods/pods.store.ts
```diff
@@ -32,15 +31,9 @@ export class PodsStore extends KubeObjectStore<Pod> {
     }
   }
 
-  getPodsByOwner(workload: WorkloadKubeObject): Pod[] {
```
Nokel81: Breaking change

src/renderer/components/dock/log-resource-selector.tsx
```diff
@@ -30,36 +33,61 @@ export const LogResourceSelector = observer((props: Props) => {
...
+  const getSelectOptions = (items: string[]) => {
+    return items.map(item => {
       return {
-        value: container.name,
-        label: container.name
+        value: item,
+        label: item
       };
     });
   };
```
Nokel81, line 50: This seems like a piece of code that would be used in multiple places. Should it be moved to `renderer/utils`?

Nokel81 (changes requested):

src/renderer/components/+workloads-pods/pods.store.ts
Nokel81: Namely, `PodsStore` is an exported type.

src/renderer/components/dock/log-resource-selector.tsx
jim-docker: Good idea, but maybe outside scope for this PR? I've seen it used inline even, wonder if it could be incorporated as default behaviour of `Select` component when adding options?

jim-docker:
looks good with deployment scaling now but pod log title is no longer being updated with selected pod's name?


https://user-images.githubusercontent.com/40840436/105767517-1d038580-5f29-11eb-8d9f-5dbb9af6a9aa.mov

src/renderer/components/+workloads-pods/pods.store.ts
aleksfront: Sure thing! I will return an old function and add `getPodsByOwnerId()` as a separate one.

aleksfront:
> looks good with deployment scaling now but pod log title is no longer being updated with selected pod's name?

Fixed

src/renderer/components/dock/log-resource-selector.tsx
aleksfront: Agreed to consider this issue in the separate PR.

Nokel81 (approved):

src/renderer/components/+workloads-pods/pods.store.ts
```diff
@@ -31,6 +32,18 @@ export class PodsStore extends KubeObjectStore<Pod> {
...
+  getPodsByOwner(workload: WorkloadKubeObject): Pod[] {
+    if (!workload) return [];
+
+    return this.items.filter(pod => {
+      const owners = pod.getOwnerRefs();
+
+      if (!owners.length) return;
+
+      return owners.find(owner => owner.uid === workload.getId());
+    });
```
jim-docker:
```suggestion
    return getPodsByOwnerId(workload.getId());
```

jim-docker (approved):

src/renderer/components/+workloads-pods/pods.store.ts
aleksfront: Of course! Brilliant!

PR #1985, opened 2021-01-19
[3.6] Show single update notification

jakolehm:
image

### DIFF
--- a/src/main/app-updater.ts
+++ b/src/main/app-updater.ts
@@ -18,9 +18,9 @@ class NotificationBackchannel {
 
 const title = "Lens Updater";
 
-async function autoUpdateCheck(windowManager: WindowManager): Promise<void> {
+async function autoUpdateCheck(windowManager: WindowManager, args: UpdateInfo): Promise<void> {
   return new Promise(async resolve => {
-    const body = "Install and restart Lens?";
+    const body = `Version ${args.version} of Lens IDE is now available. Would you like to update?`;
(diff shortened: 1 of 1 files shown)


src/main/app-updater.ts
```diff
@@ -18,9 +18,9 @@ class NotificationBackchannel {
 
 const title = "Lens Updater";
 
-async function autoUpdateCheck(windowManager: WindowManager): Promise<void> {
+async function autoUpdateCheck(windowManager: WindowManager, args: UpdateInfo): Promise<void> {
   return new Promise(async resolve => {
-    const body = "Install and restart Lens?";
+    const body = `There is an update available (${args.version}). Update and restart Lens?`;
```
Nokel81: I think knowing when a release was made public can be useful. Since I am sure some people will make their decision on whether to update based on how old the release is.

```suggestion
    const body = `There is an update (released on ${releaseDate.format("dddd, MMMM Do, yyyy")}) available (${args.version}). Update and restart Lens?`;
```

Nokel81 (commented):
Other than the date looks good.

jakolehm:
Update:
image

src/main/app-updater.ts
jakolehm: We decided not to include release date here.

nevalla (approved):
LGTM

jakolehm:
Update (after rebase):
image

PR #1990, opened 2021-01-19
Watch-api streaming reworks

### LINKED ISSUE #1898: Watch api does not work for non-admins with lots of namespaces

ixrock:
- refactored and improved stability of watch-api client
- concurrent watch-api requests limited by `5` at a time with `100ms` delay before next call
- fix: close previous hanging watch-api request sooner than by timeout end
- fix: `ServiceDetails` component has forgotten unwatch endpoints-api

**Demo:**

https://user-images.githubusercontent.com/6377066/105852071-a760ed80-5fec-11eb-852d-8984c25d61e6.mp4

**Crash test**:
```bash
for i in {1..200}; do kubectl create ns test-${i}; done
```

fix #1898


src/main/routes/watch-route.ts
```diff
@@ -65,50 +89,87 @@ class ApiWatcher {
...
+      while ((apiGroupCall = apiGroupCalls.shift())) {
+        try {
+          if (isWatchRequestEnded) break;
+          await apiGroupCall();
+          await new Promise(resolve => setTimeout(resolve, 150)); // delay between watch group calls
+        } catch (error) {
+          logger.error(error);
+        }
+      }
+    }
```
Nokel81: Jari, I think, had a different system for throttling these connections. Using `plimit`. Should we not use that here too?

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -117,15 +117,15 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
...
   }
 
-  subscribe(apis = [this.api]) {
+  getSubscribeApis() {
     const { accessibleNamespaces } = getHostedCluster();
 
     // if user has given static list of namespaces let's not start watches because watch adds stuff that's not wanted
     if (accessibleNamespaces.length > 0) {
-      return Function; // no-op
+      return []; // no-op
```
Nokel81:
```suggestion
      return [];
```

src/renderer/components/+network-services/service-details.tsx
```diff
@@ -1,28 +1,30 @@
...
 @observer
 export class ServiceDetails extends React.Component<Props> {
   componentDidMount() {
-    if (!endpointStore.isLoaded) {
-      endpointStore.loadAll();
-    }
-    endpointApi.watch();
+    disposeOnUnmount(this, [
+      kubeWatchApi.subscribeStores([endpointStore], {
+        preload: !endpointStore.isLoaded,
```
Nokel81: This seems to be the only place where you don't say `preload: true`. Is there a reason why we don't want to do that here too? Or is it just that loading endpoints is rather intensive?

src/renderer/components/app.tsx
```diff
@@ -75,38 +76,21 @@ export class App extends React.Component {
...
+        broadcastMessage(`cluster-warning-event-count:${getHostedCluster().id}`, count);
+      }),
+    ]);
   }
 
-  @computed
-  get warningsCount() {
-    let warnings = sum(nodesStore.items
-      .map(node => node.getWarningConditions().length));
+  // todo: move to nodes-store.ts
```
Nokel81: Shouldn't this todo be fixed in this PR (since imo it seems connected)

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -82,6 +83,7 @@ const defaultProps: Partial<ItemListLayoutProps> = {
   isSelectable: true,
   isConfigurable: false,
   copyClassNameFromHeadCells: true,
+  preloadStores: true,
```
Nokel81, line 86: If the default is `true` then why all the options that explicitly set it to `true`?

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
 import { apiManager } from "./api-manager";
-import { apiPrefix, isDevelopment } from "../../common/vars";
-import { getHostedCluster } from "../../common/cluster-store";
 
-export interface IKubeWatchEvent<T = any> {
-  type: "ADDED" | "MODIFIED" | "DELETED" | "ERROR";
-  object?: T;
+export { IKubeWatchEvent, IKubeWatchEventStreamEnd };
+
+export interface IKubeWatchMessage<T extends KubeObject = any> {
```
Nokel81, line 21:
```suggestion
export interface IKubeWatchMessage<T extends KubeObject = KubeObject> {
```

`any` is not `KubeObject`

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
-import { stringify } from "querystring";
-import { autobind, EventEmitter } from "../utils";
-import { KubeJsonApiData } from "./kube-json-api";
+import type { Cluster } from "../../main/cluster";
+import type { IKubeWatchEvent, IKubeWatchEventStreamEnd, IWatchRoutePayload } from "../../main/routes/watch-route";
+import type { KubeObject } from "./kube-object";
 import type { KubeObjectStore } from "../kube-object.store";
-import { ensureObjectSelfLink, KubeApi } from "./kube-api";
+import type { NamespaceStore } from "../components/+namespaces/namespace.store";
+
```
Nokel81, line 9:
```suggestion
```

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
+  private cluster: Cluster;
+  private namespaceStore: NamespaceStore;
+
+  private requestId = 0;
+  private isConnected = false;
+  private reader: ReadableStreamReader<string>;
+  private subscribers = observable.map<KubeApi, number>();
+
+  // events
+  public onMessage = new EventEmitter<[IKubeWatchMessage]>();
```
Nokel81, line 55: I think that this should be `private` with a `public get`-er because I don't think outside users should be able to override this `EventEmitter`.

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
+
+  private async init() {
+    const { getHostedCluster } = await import("../../common/cluster-store");
+    const { namespaceStore } = await import("../components/+namespaces/namespace.store");
+
+    await namespaceStore.whenReady;
+
+    this.cluster = getHostedCluster();
+    this.namespaceStore = namespaceStore;
+    this.bindAutoConnect();
```
Nokel81, line 90: I really don't like the precedent that we are heading towards with all these circular dependencies. Furthermore, I don't like the fact that we are now storing a "singleton" in N places: the singleton store, the `namespaceStore` variable, and now on every `KubeWatchApi` instance.

Please find a way to dentangle `NamespaceStore` from `KubeWatchApis` if possible since we want to be able to limit which namespaces we look at.

Nokel81: Another option would be to do the namespace filtering outside of this class and always pass it in.

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
+    return activeApis.map(api => {
+      if (!cluster.isAllowedResource(api.kind)) {
+        return [];
+      }
+
+      if (api.isNamespaced) {
+        return namespaceStore.getContextNamespaces().map(namespace => api.getWatchUrl(namespace));
+      } else {
+        return api.getWatchUrl();
+      }
```
Nokel81, line 74:
```suggestion
      }
      
      return api.getWatchUrl();
```

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
+      disposers.forEach(dispose => dispose());
+      disposers.length = 0;
+    }
 
-        return [];
-      }).flat()
-    };
+    if (preload) {
+      loading.push(...stores.map(store => store.loadAll()));
+    }
```
Nokel81: Why preload before subscribe?

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
+
+    if (waitUntilLoaded) {
+      Promise.all(loading).then(subscribe, error => {
+        this.log({
+          message: new Error("Loading stores has failed"),
+          meta: { stores, error, options },
+        });
+      });
```
Nokel81: Why not use `await`. Furthermore this isn't waiting at all since the `then` function will be added to the promise chain an then this "returns" immediately.

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
+    } else {
+      subscribe();
```
Nokel81, line 163: This is dangerous since it might reject. If you don't want to wait for the result now, please add a catch handler to log the error.

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
 
-  // todo: maybe switch to websocket to avoid often reconnects
-  @autobind()
-  protected async connect() {
-    if (this.evtSource) this.disconnect(); // close previous connection
+  protected connectionCheck() {
+    this.log({
+      message: "connection check",
+      meta: { connected: this.isConnected },
+    });
```
Nokel81: Why log before attempting the check?

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
 
-    const query = await this.getQuery();
+    if (this.isConnected) return;
+
+    return this.connect();
+  }
+
+  protected async connect(apis = this.apis) {
+    this.disconnect(); // close active connections first
```
Nokel81, line 188: This is weird. It seems that we always reconnect then. Which given the name of the method is unexpected.

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
+  protected connectionCheck() {
+    this.log({
+      message: "connection check",
+      meta: { connected: this.isConnected },
+    });
 
-    const query = await this.getQuery();
+    if (this.isConnected) return;
+
+    return this.connect();
```
Nokel81: It seems very dangerous that something called "check" would disconnect all current connections.

src/renderer/api/kube-watch-api.ts
Nokel81: Also, it seems very strange not to `subscribe` as well when waiting for preload.

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
+    const { preload = true, waitUntilLoaded = true } = options;
+    const loading: Promise<any>[] = [];
+    const disposers: Function[] = [];
+    let isDisposed = false;
 
-    await namespaceStore.whenReady;
-    const { isAdmin } = getHostedCluster();
+    async function subscribe() {
+      if (isDisposed) return;
+      const unsubscribeList = await Promise.all(stores.map(store => store.subscribe()));
```
Nokel81: Would storing this promise instead of a list of disposers be safer, so that `unsubscribe` on line 134 could do the cleanup always (by waiting for the promise to resolve).

Nokel81: Also, shouldn't this be `allSettled` so that even the stores that successfully subscribe can be `unsubscribed` even when others fail?

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
 
-  protected disconnect() {
-    if (!this.evtSource) return;
-    this.evtSource.close();
-    this.evtSource.onmessage = null;
-    this.evtSource = null;
-  }
+      // request above is stale since new request-id has been issued
+      if (this.requestId !== requestId) {
+        abortController.abort();
```
Nokel81, line 216: I don't think that this line is doing what you expect since the above request has already completed. (The promise has resolved).

All the abort controller does is tell the enclosing fetch to reject with a specific error (and halt the request too). But it doesn't do anything if the request has already finished.

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
+    if (this.isConnected) return;
+
+    return this.connect();
+  }
+
+  protected async connect(apis = this.apis) {
+    this.disconnect(); // close active connections first
+
+    if (!navigator.onLine || !apis.length) {
+      this.isConnected = false;
```
Nokel81, line 191: Shouldn't having no apis imply that you are connected to all of the vacuously?

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
+  // process received stream events, returns unprocessed buffer chunk if any
+  protected processBuffer(events: string[]): string {
+    for (const json of events) {
+      try {
+        const kubeEvent: IKubeWatchEvent = JSON.parse(json);
+        const message = this.getMessage(kubeEvent);
+
+        this.onMessage.emit(message);
+      } catch (error) {
+        return json;
```
Nokel81, line 259: Why do we ignore the rest of the events if one fails to parse?

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
-        }
+      this.isConnected = true;
+      this.reader = reader;
+
+      while (true) {
+        const { done, value } = await reader.read();
+
+        if (done) break; // exit
+
+        const events = (jsonBuffer + value).split("\n");
```
Nokel81, line 233: Are the json record guaranteed to not include new lines?

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
+        this.log({
+          message: new Error(`Failed to connect on single stream end: ${error}`),
+          meta: { event, error },
+        });
+
+        if (this.isActive) {
+          setTimeout(() => this.onServerStreamEnd(event), 1000);
+        }
+      }
+    }
```
Nokel81: IMO less indentation is generally easier to read.

```suggestion
    if (!api) {
      return;
    }
    
    try {
      await api.refreshResourceVersion({ namespace });
      this.connect();
    } catch (error) {
      this.log({
        message: new Error(`Failed to connect on single stream end: ${error}`),
        meta: { event, error },
      });

      if (this.isActive) {
        setTimeout(() => this.onServerStreamEnd(event), 1000);
      }
    }
```

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
+        await api.refreshResourceVersion({ namespace });
+        this.connect();
+      } catch (error) {
+        this.log({
+          message: new Error(`Failed to connect on single stream end: ${error}`),
+          meta: { event, error },
+        });
+
+        if (this.isActive) {
+          setTimeout(() => this.onServerStreamEnd(event), 1000);
```
Nokel81: This could be unbounded recursion, should there be a limit on this?

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
+        }
+      }
+    }
+  }
+
+  protected log({ message, meta = {} }: IKubeWatchLog) {
+    if (isProduction) {
+      return;
+    }
```
Nokel81: To help with subtle issues, I think we should also allow debug mode.
```suggestion
    if (isProduction && !isDebug) {
      return;
    }
```

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,326 @@
...
+
+    const logMessage = `%c[KUBE-WATCH-API]: ${String(message).toUpperCase()}`;
+    const isError = message instanceof Error;
+    const textStyle = `font-weight: bold; ${isError ? "color: red;" : ""}`;
+    const time = new Date().toLocaleString();
+
+    if (isError) {
+      console.error(logMessage, textStyle, { time, ...meta });
```
Nokel81, line 342: Aren't renderer errors already printed with colour red?

src/main/routes/watch-route.ts
ixrock: Alright, I'll try to reuse logic from `cluster.ts` with `plimit`

src/renderer/components/+network-services/service-details.tsx
ixrock: Well, dunno. I just moved previous logic which was that:

```ts
  componentDidMount() {
    if (!endpointStore.isLoaded) {
      endpointStore.loadAll();
    }
    endpointApi.watch(); // here was a bug - forgotten unwatch not called on unmounting component
  }
```

src/renderer/components/app.tsx
ixrock: Sounds good :)

src/renderer/components/item-object-list/item-list-layout.tsx
ixrock: It's used with `preloadStores={false}` in `KubeObjectListLayout` since loadings stores handled by `kube-watch-api.ts`

ixrock: And by default it's true for all `ItemListLayout` components

src/renderer/api/kube-watch-api.ts
ixrock: Good question. Have to double check, but I assume it probably escaped somehow with `JSON.stringify`?

src/renderer/api/kube-watch-api.ts
ixrock: Maybe. I just moved previous logic. Need to test more.

src/renderer/api/kube-watch-api.ts
ixrock: Yup, why not.

src/renderer/api/kube-watch-api.ts
ixrock: Yes, good point. Seems redundant.

src/renderer/components/+network-services/service-details.tsx
Nokel81: Fair reason, maybe do a bit blame on that line? If not ask on slack and then still if no reason remove it? That way things can either be documented or standardized.

src/renderer/api/kube-watch-api.ts
Nokel81: I don't think json has the concept of multiple "documents" in a single file unlike yaml

ixrock: Actually all good. Try this: 
```js
JSON.stringify(x={test: "multi\nline"}).includes("\n")
```
So it's always single line for receiving json from the stream (it's strinfigied in watch-route).

src/renderer/components/+network-services/service-details.tsx
ixrock: What to remove? It's already fixed in my PR. Added info to PR's description.

src/renderer/api/kube-watch-api.ts
Nokel81: I guess I meant inter-record new lines. Since it seems like the stream is a bunch of JSON records concatenated together with new lines. Thanks for pointing out that in `watch-route` we are not adding any new lines to the serialization.

ixrock:
@Nokel81 PTAL

src/renderer/components/+cluster/cluster-overview.tsx
```diff
@@ -3,59 +3,43 @@ import "./cluster-overview.scss";
...
-    }
+  componentDidMount() {
+    this.metricPoller.start(true);
 
-    await Promise.all(this.stores.map(store => store.loadAll()));
-    this.loadMetrics();
+    disposeOnUnmount(this, [
+      kubeWatchApi.subscribeStores([nodesStore, podsStore, eventStore], {
+        preload: true,
+      }),
```
Nokel81, line 33: preload defaults to true

src/renderer/components/+namespaces/namespace-select.tsx
```diff
@@ -28,17 +29,13 @@ const defaultProps: Partial<Props> = {
...
-    this.unsubscribe = namespaceStore.subscribe();
-  }
-
-  componentWillUnmount() {
-    this.unsubscribe();
+  componentDidMount() {
+    disposeOnUnmount(this, [
+      kubeWatchApi.subscribeStores([namespaceStore], {
+        preload: !namespaceStore.isLoaded,
+      })
```
Nokel81: Given that `subscribeStores` only has complete oversite it shouldn't erroneously load this value again if it was recently loaded. We shouldn't have to do this sort of fine control outside of `kubeWatchApi`

src/renderer/api/kube-watch-api.ts
Nokel81: Is there a reason why we also want it to be bold?

src/renderer/components/app.tsx
```diff
@@ -75,42 +75,20 @@ export class App extends React.Component {
...
   }
 
-  async componentDidMount() {
-    const cluster = getHostedCluster();
-    const promises: Promise<void>[] = [];
+  componentDidMount() {
+    disposeOnUnmount(this, [
+      kubeWatchApi.subscribeStores([podsStore, nodesStore, eventStore], {
+        preload: true,
+      }),
```
Nokel81, line 82: default is preload=true

src/renderer/components/+network-services/service-details.tsx
Nokel81: See my comments on `namespaceStore` for `NamespaceSelect`

src/renderer/components/app.tsx
ixrock: yes, here is used only for better visibility of what's going on

ixrock: > Is there a reason why we also want it to be bold?

I just used common style for both logging outputs, so no special reason as well as why not?

src/renderer/components/+cluster/cluster-overview.tsx
ixrock: see other comment

src/renderer/components/+namespaces/namespace-select.tsx
ixrock: Again, same logic with loading was there before PR and I don't see any reason to change it.

src/renderer/components/app.tsx
Nokel81: Given that, should we just require that `preload` is always specified?

src/renderer/components/+network-services/service-details.tsx
ixrock: Logic updated to `preload: true`, thanks.

src/renderer/components/app.tsx
ixrock: Currently yes in all used places.

src/renderer/components/+namespaces/namespace-select.tsx
ixrock: updated to `preload: true`

src/renderer/api/kube-watch-api.ts
ixrock: Oops, answered in other thread.

src/renderer/components/app.tsx
Nokel81: I meant, changing `src/renderer/api/kube-watch-api.ts:28` to be `preload: boolean` instead of `preload?: boolean`.

ixrock: Why? It should not be a required field / not critical imho.. I put explicit `preload: true` for better visibility, we can remove it all as well since it's default value.

Nokel81: What I mean is, I accept your reasoning that they should be explicit (for visibility reasons) but then I think that we should always require that. Just a suggestion though.

src/renderer/api/kube-watch-api.ts
Nokel81: What do you mean by common style?

ixrock: css-style for `console.log`

Nokel81: No I guess what I mean is, why do we want them to be styled at all, shouldn't the prefix be good enough?

jakolehm:
@ixrock please resolve conflicts.

src/renderer/kube-object.store.ts
```diff
@@ -195,29 +194,32 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
+    });
+
+    reaction(() => this.eventsBuffer.length > 0, this.updateFromEventsBuffer, {
       delay
     });
   }
 
-  subscribe(apis = [this.api]) {
-    return KubeApi.watchAll(...apis);
+  getSubscribeApis(): KubeApi[] {
```
jakolehm, line 210: Could you explain why `getSubscribeApis()` is introduced in this PR? Is it more convenient way to override subscribed apis per store?

src/renderer/components/+workloads-overview/overview.tsx
```diff
@@ -15,60 +14,23 @@ import { replicaSetStore } from "../+workloads-replicasets/replicasets.store";
...
-    await when(() => this.isUnmounting && !this.isLoading);
-    unsubscribeMap.forEach(dispose => dispose());
-    unsubscribeMap.clear();
-  }
-
-  componentWillUnmount() {
-    this.isUnmounting = true;
+  componentDidMount() {
+    disposeOnUnmount(this, [
+      kubeWatchApi.subscribeStores([
```
jakolehm, line 27: Why aren't we checking `isAllowedResource` anymore?

src/main/routes/watch-route.ts
```diff
@@ -65,50 +90,72 @@ class ApiWatcher {
...
 
-    response.setHeader("Content-Type", "text/event-stream");
+    this.setResponse(response);
+    response.setHeader("Content-Type", "application/json");
     response.setHeader("Cache-Control", "no-cache");
     response.setHeader("Connection", "keep-alive");
     logger.debug(`watch using kubeconfig:${JSON.stringify(cluster.getProxyKubeconfig(), null, 2)}`);
 
+    // limit concurrent k8s requests to avoid possible ECONNRESET-error
+    const requests = plimit(5);
```
aleksfront, line 126: Is there a specific reason to use exactly 5 parallel connections, no more no less?

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,335 @@
...
-      delay: 500,
+      equals: comparer.structural,
     });
-  }
 
-  @computed get activeApis() {
-    return Array.from(this.subscribers.keys());
+    window.addEventListener("online", () => this.connect());
+    window.addEventListener("offline", () => this.disconnect());
+    setInterval(() => this.connectionCheck(), 60000 * 5); // every 5m
```
aleksfront, line 103: A suggestion: should we also check if window loses its focus or gets minimized? We can use `win.isFocused()` and `win.isMinimized` electron methods
https://www.electronjs.org/docs/api/browser-window#winisfocused
https://www.electronjs.org/docs/api/browser-window#winisminimized

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,335 @@
...
   }
 
   getSubscribersCount(api: KubeApi) {
     return this.subscribers.get(api) || 0;
   }
 
-  subscribe(...apis: KubeApi[]) {
+  subscribeApi(api: KubeApi | KubeApi[]) {
```
aleksfront: For such cases where an argument could be 1 item or an array of items we can introduce 2 different methods:
`subscribeApi(api: KubeApi)` and `subscribeApis(apis: KubeApi[])`. This will allow us also to eliminate next line
`const apis: KubeApi[] = [api].flat();`

src/renderer/api/kube-watch-api.ts
```diff
@@ -1,201 +1,335 @@
...
-    const { namespaceStore } = await import("../components/+namespaces/namespace.store");
+  subscribeStores(stores: KubeObjectStore[], options: IKubeWatchSubscribeStoreOptions = {}): () => void {
+    const { preload = true, waitUntilLoaded = true } = options;
+    const loading: Promise<any>[] = [];
+    const disposers: Function[] = [];
+    let isDisposed = false;
 
-    await namespaceStore.whenReady;
-    const { isAdmin } = getHostedCluster();
+    async function subscribe() {
```
aleksfront: This method should be refactored or at least renamed. The reason is function `subscribe()` does a unsubscription at the end 
```
disposers.push(...unsubscribeList);
if (isDisposed) unsubscribe();
```
this is the opposite of what function name declares.

src/renderer/api/kube-watch-api.ts
aleksfront: I agree, `return this.connect()` should be moved into another function.

src/renderer/api/kube-watch-api.ts
aleksfront: Probably better to log connection check result, not just the intention.

src/renderer/kube-object.store.ts
ixrock: Yes, it better to have "in place" since some stores might subscribe for multiple apis, e.g. `RolesStore`, `RoleBindingsStore`.
Previously handled in `subscribe()` but having apis separately gives more freedom of using this (initially refactored to  be used in `KubeWatchApi`).

src/renderer/components/+workloads-overview/overview.tsx
ixrock: Because now it's handled under the hood in `KubeObjectStore.loadItems()`.

src/main/routes/watch-route.ts
ixrock: Nope. Just randomly picked magic number to avoid making k8s angry :D

src/renderer/api/kube-watch-api.ts
ixrock: If app is minimized you're suggesting to what exactly to do? How it should be related to connection?

src/renderer/api/kube-watch-api.ts
ixrock: Whaaaat? :D 
Definitely no. Api should be convenient and simple to use.
Only reason to do that in case of hard breaking changes for exposed extension-apis.

src/renderer/api/kube-watch-api.ts
ixrock: This is for a reason, because loading is happens in `async` and (un)subscribing is `sync`-mode.
For large clusters when you visit a page you might go another one before even loading all stores and unsubscribing must happen in sync with component unmount.

Nokel81: Could we instead pass in an `AbortController` which aborts the subscribes if they haven't finished yet?

src/renderer/api/kube-watch-api.ts
ixrock: This is done for better visibility in chrome's `devtools` console. Why you think to simplify it? Info about watch-api not so important to notify for debugging?

Nokel81: in devtools, `console.error` already shows up as "red" is that not visible enough?

ixrock: Yes, but it's same for all kinds of error in output. 
Why you against bold-text in error? Red color already removed from text-style.

src/renderer/api/kube-watch-api.ts
ixrock: this all refactored, please re-review again

src/renderer/api/kube-watch-api.ts
Nokel81: you haven't added it

src/renderer/api/kube-watch-api.ts
Nokel81: I guess what I am really asking about is why having the function at all, yes `logger` shouldn't be in `main` because it gets used in `renderer` too but what it wrong with just calling that?

We already have a decently common style of having a static for the logging prefix, adding metadata already works.

jakolehm (approved):
LGTM. I think we should get this in asap and fix remaining issues (if any) in followup PRs. This way we can actually test this more before final 4.1 release.

src/renderer/api/kube-watch-api.ts
ixrock: I guess this is kinda fixed in #2033 ?

src/renderer/api/kube-watch-api.ts
ixrock: `src/main/logger` doesn't work well in `rendrerer` or it might need some tweaks / additional transports..

src/renderer/api/kube-watch-api.ts
jakolehm: I think @aleksfront has a point here.. we could optimize this so that Lens stops watching if window is unfocused (or if frame is not active). Anyway it's out of scope for this PR.

src/renderer/api/kube-watch-api.ts
Nokel81: Having looked around in that PR I don't see where. Care to point to it specifically?

src/renderer/api/kube-watch-api.ts
Nokel81: How does it not work? We have been (ab)using it for several releases. Namely there are currently 13 files in renderer who import `main/logger`

Nokel81: Anyway wouldn't the better solution just to make a new winston logger instead of hand rolling our own?

aleksfront (approved):

src/renderer/api/kube-watch-api.ts
ixrock: Yes, will be fixed in follow-up PR(s).

src/renderer/api/kube-watch-api.ts
ixrock: This was added:

```ts
    if (isProduction && !isDebugging) {
      return;
    }
```

src/renderer/api/kube-watch-api.ts
ixrock: This api not exposed to "outside users" (only kube-watch-api is a "user").
So I would prefer to do that later if any reason.

src/renderer/api/kube-watch-api.ts
ixrock: This was refactored in https://github.com/lensapp/lens/pull/2033 I guess not relevant anymore.
Please see that PR and comment / approve.

src/renderer/api/kube-watch-api.ts
ixrock: That way `resourceVersion` will be updated first and subsribe(s) would happen without overhead k8s traffic.

src/renderer/api/kube-watch-api.ts
ixrock: To use `await` subscribe method must be an `async` which is not desirable in a way how we use it with `react`-components.

src/renderer/api/kube-watch-api.ts
ixrock: In that PR `subscribe()` cannot reject (at least in theory).

src/renderer/api/kube-watch-api.ts
ixrock: No, since apis is watching by `mobx`-reaction.. 
`connect` is called every time it needs new or refresh previous connection.

src/renderer/api/kube-watch-api.ts
Nokel81: Wouldn't the subscribes need to do that work anyway?

src/renderer/api/kube-watch-api.ts
Nokel81: We should be trying to do defence in depth

src/renderer/api/kube-watch-api.ts
ixrock: Why do you think that `request` is finished at this stage? 
`connect`  is debounced to `1s` so when clicking fast btw pages you might have a situation with a stale request, this should halt it faster.

src/renderer/api/kube-watch-api.ts
ixrock: No. Subscribe might happen without `resourceVersion` at all.

src/renderer/api/kube-watch-api.ts
ixrock: Name is expected. Disconnect is happening here because we keep only single stream-connection and it comes mostly from `mobx`-reaction that triggered when observable watch-apis changed. 
Disconnect itself also used separately in other place.

src/renderer/api/kube-watch-api.ts
ixrock: Because it might happen only for last broken json from the stream (we manage that stream in `watch-route`)

src/renderer/api/kube-watch-api.ts
Nokel81: If `apis` is watched by `mobx` then `isConnected` is unrelated to that.

src/renderer/api/kube-watch-api.ts
ixrock: Tried to replace to `common/logger` but have not satisfied with the result:
This is funny, but there is no official support for browsers for `winston` since 2013 ([issue](https://github.com/winstonjs/winston/issues/287)) and customizing `transport` is a bit tricky (I don't want to install extra npm package for that).

**Before:**
Screenshot 2021-02-01 at 17 25 13

**After:**
Screenshot 2021-02-01 at 17 23 39

PR #1996, opened 2021-01-20
[4.0] Fix extension cluster submenu re-render

jakolehm:
It seems that https://github.com/lensapp/lens/pull/1688 was only a partial fix. We need to keep generated routes in memory so that we don't accidentally instantiate components over and over again.

### DIFF
--- a/src/renderer/components/app.tsx
+++ b/src/renderer/components/app.tsx
@@ -51,6 +51,8 @@ import { sum } from "lodash";
 
 @observer
 export class App extends React.Component {
+  private extensionRoutes: Map<ClusterPageMenuRegistration, JSX.Element> = new Map();
+
   static async init() {
     const frameId = webFrame.routingId;
     const clusterId = getHostedClusterId();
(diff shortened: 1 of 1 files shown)


src/renderer/components/app.tsx
```diff
@@ -145,21 +147,27 @@ export class App extends React.Component {
...
+        return route;
+      }
 
+      const tabRoutes = this.getTabLayoutRoutes(menu);
       if (tabRoutes.length > 0) {
         const pageComponent = () => <TabLayout tabs={tabRoutes} />;
-
-        return <Route key={`extension-tab-layout-route-${index}`} component={pageComponent} path={tabRoutes.map((tab) => tab.routePath)} />;
+        route = <Route key={`extension-tab-layout-route-${index}`} component={pageComponent} path={tabRoutes.map((tab) => tab.routePath)} />;
+        this.extensionRoutes.set(menu, route);
```
ixrock: Probably this better to do not in `render()`. Also what if menu item tabs would change, but it's cached already?

src/renderer/components/app.tsx
```diff
@@ -51,6 +51,8 @@ import { sum } from "lodash";
 
 @observer
 export class App extends React.Component {
+  private extensionRoutes: Map<ClusterPageMenuRegistration, JSX.Element> = new Map();
```
ixrock:
```suggestion
  private extensionRoutes: Map<ClusterPageMenuRegistration, React.ReactNode> = new Map();
```

src/renderer/components/app.tsx
jakolehm: > Probably this better to do not in `render()`. 

What alternatives do we have?

> Also what if menu item tabs would change, but it's cached already?

We don't currently support dynamic menus (from an extension). If extension is updated then `menu` will be a different instance and it might cause issues (duplicate route?). Still I feel that it's bigger issue to recreate route on each re-render.

jakolehm: @ixrock now it's refactored to use `reaction()` which happens outside of `render()`.

ixrock (approved):
LGTM

nevalla (approved):

PR #2001, opened 2021-01-21
Fix lens node shell

### LINKED ISSUE #1857: Lens Node Shell - Node-Shell Pod Stuck in Pending Due to Mismatched Node Selector
**Describe the bug**
When attempting to open a shell at the node level, the "node-shell" pod gets stuck in pending and never starts.

**To Reproduce**
1. Go to Nodes, select a Node and open "Shell"
2. Terminal window opens below with "Connecting ..." and never moved ahead from that.
3. From the Pods view, it shows a "node-shell" pod in Pending status, but never changes or starts.

**Expected behavior**
I expect to get a terminal window running on the Node.

- Lens Version: 3.6.7
- OS: Windows 10
- K8s environment: AWS EKS

**Additional context**
I expect it to be a mismatch in the pod's node affinity. 
My node has the following properties: 
- Name: "ip-10-86-40-168.us-west-2.compute.internal".
- Label: kubernetes.io/hostname=ip-10-86-40-168.mydomain.com

However, Lens creates the pod with the following nodeSelector:
```
  nodeSelector:
    kubernetes.io/hostname: ip-10-86-40-168.us-west-2.compute.internal
```

It seems to be using the **Name** of the node in the selector, yet matching against the label **kubernetes.io/hostname** which doesn't always match the Name of the Node, which in my case, it does not match, causing the pod to never start because the matching node doesn't exist.

vshakirova:
Hi guys, just a small fix. 
I'd like to suggest the idea to replace the `nodeSelector` with `nodeName` (https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodename) 
There is another problem. If something wrong with pod, a terminal hangs with "Connecting..." message and never moved ahead from that. It happens because we pass code=1001 and WebSocket reconnects again and again. I found `reconnectDelaySeconds` field, but it seems I can't set it to 0 only for nodeShellSession. Need your advice. Thank you

Fixes: https://github.com/lensapp/lens/issues/1857

### DIFF
--- a/src/main/node-shell-session.ts
+++ b/src/main/node-shell-session.ts
@@ -26,7 +26,7 @@ export class NodeShellSession extends ShellSession {
 
     if (this.createNodeShellPod(this.podId, this.nodeName)) {
       await this.waitForRunningPod(this.podId).catch(() => {
-        this.exit(1001);
+        this.exitAndClean(1001);
       });
     }
     args = ["exec", "-i", "-t", "-n", "kube-system", this.podId, "--", "sh", "-c", "((clear && bash) || (clear && ash) || (clear && sh))"];
@@ -49,11 +49,14 @@ export class NodeShellSession extends ShellSession {
     appEventBus.emit({name: "node-shell", action: "open"});
   }
 
-  protected exit(code = 1000) {
+  protected exitAndClean(code = 1000) {
     if (this.podId) {
       this.deleteNodeShellPod();
     }
-    super.exit(code);
+
+    if (code != 1000) {
+      this.sendResponse("Error occurred. ");
+    }
   }
 
   protected async createNodeShellPod(podId: string, nodeName: string) {
@@ -65,6 +68,7 @@ export class NodeShellSession extends ShellSession {
         namespace: "kube-system"
       },
       spec: {
+        nodeName,
         restartPolicy: "Never",
         terminationGracePeriodSeconds: 0,
         hostPID: true,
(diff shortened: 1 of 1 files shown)


src/main/node-shell-session.ts
```diff
@@ -50,10 +50,19 @@ export class NodeShellSession extends ShellSession {
...
+    let timeout;
+
     if (this.podId) {
       this.deleteNodeShellPod();
     }
-    super.exit(code);
+
+    if (code != 1000) {
+      this.sendResponse("Error occurred. Terminal will auto-close in 1 seconds ...");
+      timeout = 1*1000;
```
Nokel81: I don't think we should auto close the terminal. Saying an error occurred (with the error code). Should be good enough.

vshakirova: @Nokel81 Could you please elaborate, do you mean to print 1001 error code to a terminal?

Nokel81: That is what I was suggesting. Though now that I have done a bit more research it seems at auto closing the terminal might be a reasonable action, though 1s is a bit fast.

Would you be able to increase the time to 15s?

vshakirova: @Nokel81 Yeah..I tried to do it, but it seems that all next functions e.g. `closeWebsocketOnProcessExit();` also are run if we have the timeout > 1s. I can see two identical messages on my terminal, I mean `Terminal will auto-close in 15 seconds`.

Nokel81: If it is a duplicate wouldn't not having this message here then work or should we just always auto-close in 15s and not print this message?

Nokel81 (approved):

src/main/node-shell-session.ts
vshakirova: I found one solution.. We don't need to override parent's `exit()` because in this case we call `exit()` before `closeWebsocketOnProcessExit()`. And message `Terminal will auto-close in 15 seconds` won't be shown.  Now it looks like:
Screen Shot 2021-03-01 at 6 21 14 PM

ixrock (approved):

PR #2003, opened 2021-01-21
Fix pod loading when under restrictive RBAC

### LINKED ISSUE #1820: RBAC user with restricted user permission stoped listing pods in 4.x
**Describe the bug**
I have created a user using [Permission Manager](https://github.com/sighupio/permission-manager). It is listing the pods just fine in the Lens version3.6. Kubectl get pods using this restricted permission user is showing listing the pods as well.

After auto-updating to 4.0.4, suddenly stopped working. 

See the attachment for more info.

**To Reproduce**
1. Create a user with limited RBAC permission using the Permission Manager.
2. Download the kubeconfig file for the  Permission Manager user
3. Add cluster using Lens 3.6 and kubeconfig from step 3
4. See the pod listing under lens UI Workloads-Pods
5. Try the same using Lens 4.0..4

**Screenshots**

- Lens Version: 4.0.4
- OS:  Mac OSX

lns bug 2
lens 4 0 4 bug

### LINKED ISSUE #1961: Lens 4* versions + RBAC + pods don't appear in the dashboard
**Describe the bug**
Hello Lensapp devs

I have a user with only ClusterRoles and Role for specific Namespaces
and the user is part of the group system:basic-user.

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: aws-auth
  namespace: kube-system
data:
  mapRoles: |
    - userarn: arn:aws:iam::88888888888:user/Jolla
      username: Jolla
      groups:
        - system:basic-user
```

I can exec all the pods, see the logs in all pods in the specific namespaces with kubectl or Lens-3.6.9.AppImage and Lens-3.6.10.AppImage 

I tested all Lens 4.0 version and the issue persists 

How can I debug this and see why the RBAC config isn't working with Lens 4.*

- Lens Version:
  Lens-3.6.9.AppImage
  Lens-3.6.10.AppImage
  Lens-4.0.1.AppImage
  Lens-4.0.2.AppImage
  Lens-4.0.3.AppImage 
  Lens-4.0.4.AppImage
  Lens-4.0.5.AppImage
  Lens-4.0.6.AppImage
  Lens-4.0.7.AppImage
  Lens-4.0.8.AppImage
- OS: 
  linux distros: ubuntu, SB 33, Fedora 33, archlinux
  AppImanges  in all linux distros and in ubuntu snap too

### LINKED ISSUE #2227: "Cannot read property 'getId' of undefined" when clicking on a pod
**Describe the bug**
When clicking on a pod of our production cluster, the left pane starts to show up and it's immediately replaced by the error screen showing the error above. It does not happen with pods of our dev cluster

**To Reproduce**
1. Go to Pods
2. Click on a pod

**Expected behavior**
I expected to see the pod details

- Lens Version: 4.1.3
- OS: ubuntu
- Installation method (e.g. snap or AppImage in Linux): DEB

**Logs:**
This what I get from the terminal
```
warn: PrometheusLens: failed to list services: services "prometheus" not found
warn: PrometheusLens: failed to list services: services "prometheus" not found
warn: PrometheusLens: failed to list services: services "prometheus-server" not found
warn: PrometheusLens: failed to list services: services "prometheus-server" not found
warn: PrometheusLens: failed to list services: services "prometheus" not found
warn: PrometheusLens: failed to list services: services "prometheus" not found
warn: PrometheusLens: failed to list services: services "prometheus-server" not found
warn: PrometheusLens: failed to list services: services "prometheus-server" not found
```

This is from the error page
```
Component stack:


    at QI (http://249ff66b-b091-40f1-804b-a670a6c70fd8.localhost:40019/build/Lens.js:2:8013519)
    at span
    at div
    at jd (http://249ff66b-b091-40f1-804b-a670a6c70fd8.localhost:40019/build/Lens.js:2:7756151)
    at div
    at kI (http://249ff66b-b091-40f1-804b-a670a6c70fd8.localhost:40019/build/Lens.js:2:8016142)
    at Details
    at div
    at div
    at div
    at Bp (http://249ff66b-b091-40f1-804b-a670a6c70fd8.localhost:40019/build/Lens.js:2:7728188)
    at Qd (http://249ff66b-b091-40f1-804b-a670a6c70fd8.localhost:40019/build/Lens.js:2:7753647)
    at sy (http://249ff66b-b091-40f1-804b-a670a6c70fd8.localhost:40019/build/Lens.js:2:7874585)
    at pa (http://249ff66b-b091-40f1-804b-a670a6c70fd8.localhost:40019/build/Lens.js:2:7560311)
    at t (http://249ff66b-b091-40f1-804b-a670a6c70fd8.localhost:40019/build/Lens.js:2:191498)
    at eT (http://249ff66b-b091-40f1-804b-a670a6c70fd8.localhost:40019/build/Lens.js:2:8261853)
```
```
Error stack:

 
TypeError: Cannot read property 'getId' of undefined
    at http://249ff66b-b091-40f1-804b-a670a6c70fd8.localhost:40019/build/Lens.js:2:8013840
    at http://249ff66b-b091-40f1-804b-a670a6c70fd8.localhost:40019/build/Lens.js:2:42956
    at Array.map (<anonymous>)
    at Proxy.forEach.Array.<computed>.pr.<computed> (http://249ff66b-b091-40f1-804b-a670a6c70fd8.localhost:40019/build/Lens.js:2:42931)
    at QI.render (http://249ff66b-b091-40f1-804b-a670a6c70fd8.localhost:40019/build/Lens.js:2:8013805)
    at Ce (http://249ff66b-b091-40f1-804b-a670a6c70fd8.localhost:40019/build/Lens.js:2:13973)
    at http://249ff66b-b091-40f1-804b-a670a6c70fd8.localhost:40019/build/Lens.js:2:7444538
    at ce (http://249ff66b-b091-40f1-804b-a670a6c70fd8.localhost:40019/build/Lens.js:2:12079)
    at e.track (http://249ff66b-b091-40f1-804b-a670a6c70fd8.localhost:40019/build/Lens.js:2:23443)
    at QI.s [as render] (http://249ff66b-b091-40f1-804b-a670a6c70fd8.localhost:40019/build/Lens.js:2:7444486)
```

**Kubeconfig:**
```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: DATA
    server: https://IP
  name: dev
- cluster:
    certificate-authority-data: DATA
    server: https://IP
  name: prod
contexts:
- context:
    cluster: dev
    user: dev
  name: dev
- context:
    cluster: prod
    user: prod
  name: prod
current-context: prod
kind: Config
preferences: {}
users:
- name: dev
  user:
    auth-provider:
      config:
        access-token: TOKEN
        cmd-args: config config-helper --format=json
        cmd-path: /usr/lib/google-cloud-sdk/bin/gcloud
        expiry: "2021-02-25T09:12:24Z"
        expiry-key: '{.credential.token_expiry}'
        token-key: '{.credential.access_token}'
      name: gcp
- name: prod
  user:
    auth-provider:
      config:
        access-token: TOKEN
        cmd-args: config config-helper --format=json
        cmd-path: /usr/lib/google-cloud-sdk/bin/gcloud
        expiry: "2021-02-25T09:12:24Z"
        expiry-key: '{.credential.token_expiry}'
        token-key: '{.credential.access_token}'
      name: gcp
```

**Additional context**

Nokel81:
- Correctly handle the case where `undefined` is returned by the fetch because of a 403: Unauthorized being returned from the Kube Api due to only having list permissions in some of the namespaces

- Fix rendering of Pod Secrets in the details view to handle correctly the case where the user doesn't have list or get secrets permissions

- Remove CancelablePromise<T> as it was only used in one place and the return types where wrong causing a crash while debugging the above.

- Move HelmChartDetails to create its own AbortController for cancelling its request itself as that was the only instance of the cancel() method being called

- Significantly improve the typing of `isJsonApiData` and `isJsonApiDataList` so that they actually make sure that the form of the data matches the assertion they make. This also removes a crash from incorrectly assuming that `any` could not be `undefined`.

- Add tests for the above two functions

fixed #1961 , fixes #2227


src/renderer/api/endpoints/pods.api.ts
```diff
@@ -346,27 +346,17 @@ export class Pod extends WorkloadKubeObject {
...
-        }
-
-        if (state.terminated) {
-          const { reason } = state.terminated;
-
-          message = reason ? reason : "Terminated";
-        }
-      });
+      if (state.terminated) {
+        return state.terminated.reason || "Waiting";
```
jim-docker: Little bit of refactoring on the side?

```suggestion
        return state.terminated.reason || "Terminated";
```

src/renderer/api/endpoints/pods.api.ts
```diff
@@ -477,10 +467,7 @@ export class Pod extends WorkloadKubeObject {
   }
 
   getSelectedNodeOs() {
-    if (!this.spec.nodeSelector) return;
-    if (!this.spec.nodeSelector["kubernetes.io/os"] && !this.spec.nodeSelector["beta.kubernetes.io/os"]) return;
-
-    return this.spec.nodeSelector["kubernetes.io/os"] || this.spec.nodeSelector["beta.kubernetes.io/os"];
+    return this.spec.nodeSelector?.["kubernetes.io/os"] || this.spec.nodeSelector?.["beta.kubernetes.io/os"];
```
jim-docker: more side refactoring. Shouldn't it be:
```suggestion
    return this.spec?.nodeSelector?.["kubernetes.io/os"] || this.spec?.nodeSelector?.["beta.kubernetes.io/os"];
```

src/renderer/api/kube-object.ts
```diff
@@ -55,12 +56,233 @@ export class KubeObject implements ItemObject {
...
+      return false;
+    }
+
+    if (!hasOwnProperties(object, "kind", "apiVersion", "metadata")) {
+      return false;
+    }
+
+    if (typeof object.kind !== "string"
+      || typeof object.apiVersion !== "string"
+      || !object.metadata || typeof object.metadata !== "object"
```
jim-docker: is testing `!object.metadata` necessary since you already checked it with `hasOwnProperties()` call? Or you're trying to catch if it has value `null` or `undefined` or 0 or false...

src/renderer/api/kube-object.ts
```diff
@@ -55,12 +56,233 @@ export class KubeObject implements ItemObject {
...
+      && typeof metadata.continue !== "string"
+      && typeof metadata.continue !== "undefined"
+    ) {
+      return false;
+    }
+
+    if (hasOwnProperty(metadata, "finalizers")
+      && typeof metadata.finalizers !== "undefined"
+      && !(Array.isArray(metadata.finalizers)
+      && metadata.finalizers.every(finalizer => typeof finalizer === "string"))
```
jim-docker: I almost missed that this is a compound conditional, should/could this line be indented to make it more obvious? Or something...

src/renderer/api/kube-object.ts
```diff
@@ -55,12 +56,233 @@ export class KubeObject implements ItemObject {
...
+    }
+
+    labels: if (hasOwnProperty(metadata, "labels")) {
+      if (typeof metadata.labels === "undefined" || (typeof metadata.labels === "object" && metadata.labels !== null && !Array.isArray(metadata.labels))) {
+        if (metadata.labels) {
+          for (const label in metadata.labels) {
+            if (!hasOwnProperty(metadata.labels, label) || typeof metadata.labels[label] !== "string") {
+              return false;
+            }
+          }
```
jim-docker: Would `Object.entries(metadata.labels).forEach(...)` eliminate the need to use `hasOwnProperty()`?

src/renderer/api/kube-object.ts
```diff
@@ -55,12 +56,233 @@ export class KubeObject implements ItemObject {
...
+        }
+      }
+
+      return false;
+    }
+
+    return true;
+  }
+
+  static isPartialJsonApiData(object: unknown): object is Partial<KubeJsonApiData> {
```
jim-docker, line 115: I don't get the difference between this and `isJsonApiData()`. Can duplicated code be shared? They look the same except for what `object` is.

src/renderer/item.store.ts
```diff
@@ -28,8 +28,8 @@ export abstract class ItemStore<T extends ItemObject = ItemObject> {
   }
 
   @action
-  protected sortItems(items: T[] = this.items, sorting?: ((item: T) => any)[], order?: "asc" | "desc"): T[] {
-    return orderBy(items, sorting || this.defaultSorting, order);
+  protected sortItems(items: T[] = this.items, sorting: ((item: T) => any)[] = [this.defaultSorting], order?: "asc" | "desc"): T[] {
```
jim-docker, line 53: Either way this is unreadable to me. Modern JavaScript/Typescript is for the young.

jim-docker (commented):
Nice quick work. Looks like strong bullet-proofing but touches some core code, need to test thoroughly for unexpected side-effects

src/renderer/api/kube-object.ts
Nokel81: I am trying to catch `null` (and would have used `object !== null` but you and Stefan both said you don't like that. 

For historical reasons `typeof null` is `"object"` so we have to add this check here.

src/renderer/api/kube-object.ts
Nokel81: The `every` call?

src/renderer/api/kube-object.ts
Nokel81: This check technically verifies that there the values are not "up the prototype" chain. Furtherore, we cannot use `forEach` we would have to use `every` again because you cannot return out of several functions at once.

src/renderer/api/kube-object.ts
Nokel81: The difference is subtle yes, but then again so is the difference between `T` and `Partial<T>`. Namely, this function still returns true if any of the top level fields are `undefined` or not present. That is the only difference.

Unfortunately, there is no easy way to have a `isPartial` argument. However, I might be able to simplify the above by first calling this and then verifying that none of the fields are `undefined` or missing.

src/renderer/item.store.ts
Nokel81: I'll had some doc comments to help clear this up.

src/renderer/api/endpoints/pods.api.ts
Nokel81: I think that this is fine to be in this PR since I am already touching the `pods.api.ts` file.

With regards to your comment. No, spec is not an optional field so there is no point conditionally checking it.

src/renderer/api/kube-object.ts
jim-docker: It's `==` that I said I didn't like. `!==` is fine.

Nokel81: oh I see

src/renderer/api/kube-object.ts
jim-docker: Something like this?
```
    if (hasOwnProperty(metadata, "finalizers")
      && typeof metadata.finalizers !== "undefined"
      && !(
        Array.isArray(metadata.finalizers)
        && metadata.finalizers.every(finalizer => typeof finalizer === "string")
      )
    ) {
 ```
Maybe young eyes don't need such assistance :)

Comment (Nokel81) [reply src/renderer/api/kube-object.ts]: Oh yes that would be better

File: src/renderer/api/kube-object.ts (hunk 5 above)
Comment (jim-docker) [reply src/renderer/api/kube-object.ts]: I know, but it's 2021, we still have to rely on `hasOwnProperty`...

File: src/renderer/api/kube-object.ts (hunk 3 above)
Comment (jim-docker) [reply src/renderer/api/kube-object.ts]: But `!object.metadata` is fine here, just wanted to make sure you needed to do the check.

File: src/renderer/api/endpoints/pods.api.ts (hunk 2 above)
Comment (jakolehm) [reply src/renderer/api/endpoints/pods.api.ts]: Maybe add a test for this while touching this method? 🙂

File: src/renderer/components/+apps-helm-charts/helm-chart-details.tsx [hunk 8]
```diff
@@ -26,10 +25,11 @@ export class HelmChartDetails extends Component<Props> {
   @observable readme: string = null;
   @observable error: string = null;
 
-  private chartPromise: CancelablePromise<{ readme: string; versions: HelmChart[] }>;
+  private chartPromise: Promise<{ readme: string; versions: HelmChart[] }>;
```
jakolehm, line 28: Not sure how these changes are related to the PR title?

Nokel81: Specifically for this, see note in PR description

> Remove CancelablePromise as it was only used in one place and the return types where wrong causing a crash while debugging the above.

If you want I can move the out to a separate PR but I did encounter an exception while debugging this issue.

src/renderer/api/endpoints/pods.api.ts
Nokel81: Sounds good

src/renderer/components/+apps-helm-charts/helm-chart-details.tsx
jakolehm: Ok, missed the description. Probably fine to keep here as it's somewhat related.

extensions/example-extension/package-lock.json
```diff
@@ -626,7 +626,644 @@
     },
     "@k8slens/extensions": {
       "version": "file:../../src/extensions/npm/extensions",
-      "dev": true
+      "dev": true,
```
jakolehm: Please remove `extensions/*/package-lock.json` changes.

src/renderer/api/endpoints/pods.api.ts
```diff
@@ -267,60 +267,55 @@ export class Pod extends WorkloadKubeObject {
   }
 
   getAllContainers() {
-    return this.getContainers().concat(this.getInitContainers());
+    return [...this.getContainers(), ...this.getInitContainers()];
```
jakolehm: Why are we changing this? Please revert unless really related to the PR title.

src/renderer/api/endpoints/pods.api.ts
```diff
@@ -267,60 +267,55 @@ export class Pod extends WorkloadKubeObject {
...
   }
 
   getRunningContainers() {
-    const statuses = this.getContainerStatuses();
-
-    return this.getAllContainers().filter(container => {
-      return statuses.find(status => status.name === container.name && !!status.state["running"]);
-    }
+    const runningContainerNames = new Set(
```
jakolehm: Why are we changing this? Please revert unless really related to the PR title.

src/renderer/api/endpoints/pods.api.ts
```diff
@@ -267,60 +267,55 @@ export class Pod extends WorkloadKubeObject {
...
   }
 
   getContainerStatuses(includeInitContainers = true) {
-    const statuses: IPodContainerStatus[] = [];
-    const { containerStatuses, initContainerStatuses } = this.status;
-
-    if (containerStatuses) {
-      statuses.push(...containerStatuses);
-    }
+    const { containerStatuses = [], initContainerStatuses = [] } = this.status ?? {};
```
jakolehm: Why are we changing this? Please revert unless really related to the PR title.

src/renderer/api/endpoints/pods.api.ts
```diff
@@ -267,60 +267,55 @@ export class Pod extends WorkloadKubeObject {
...
+      return [...containerStatuses, ...initContainerStatuses];
     }
 
-    return statuses;
+    return [...containerStatuses];
   }
 
   getRestartsCount(): number {
-    const { containerStatuses } = this.status;
+    const { containerStatuses = [] } = this.status ?? {};
```
jakolehm: Why are we changing this? Please revert unless really related to the PR title.

src/renderer/api/endpoints/pods.api.ts
```diff
@@ -267,60 +267,55 @@ export class Pod extends WorkloadKubeObject {
...
 
-    if (!containerStatuses) return 0;
-
-    return containerStatuses.reduce((count, item) => count + item.restartCount, 0);
+    return containerStatuses.reduce((totalCount, { restartCount }) => totalCount + restartCount, 0);
   }
 
   getQosClass() {
-    return this.status.qosClass || "";
+    return this.status?.qosClass || "";
```
jakolehm: Why are we changing this? Please revert unless really related to the PR title.

src/renderer/api/endpoints/pods.api.ts
```diff
@@ -267,60 +267,55 @@ export class Pod extends WorkloadKubeObject {
...
   }
 
   getReason() {
-    return this.status.reason || "";
+    return this.status?.reason || "";
```
jakolehm: Why are we changing this? Please revert unless really related to the PR title.

src/renderer/api/endpoints/pods.api.ts
```diff
@@ -267,60 +267,55 @@ export class Pod extends WorkloadKubeObject {
...
 
-  // Returns one of 5 statuses: Running, Succeeded, Pending, Failed, Evicted
-  getStatus() {
+  getStatus(): PodStatus {
     const phase = this.getStatusPhase();
     const reason = this.getReason();
-    const goodConditions = ["Initialized", "Ready"].every(condition =>
-      !!this.getConditions().find(item => item.type === condition && item.status === "True")
-    );
+    const trueConditionTypes = new Set(this.getConditions()
```
jakolehm: Why are we changing this? Please revert unless really related to the PR title.

src/renderer/api/endpoints/pods.api.ts
```diff
@@ -346,37 +341,27 @@ export class Pod extends WorkloadKubeObject {
...
+      }
     }
-    if (message) return message;
 
     return this.getStatusPhase();
   }
 
   getStatusPhase() {
-    return this.status.phase;
+    return this.status?.phase;
```
jakolehm: Why are we changing this? Please revert unless really related to the PR title.

src/renderer/api/endpoints/pods.api.ts
```diff
@@ -346,37 +341,27 @@ export class Pod extends WorkloadKubeObject {
...
   }
 
   getConditions() {
-    return this.status.conditions || [];
+    return this.status?.conditions || [];
```
jakolehm: Why are we changing this? Please revert unless really related to the PR title.

src/renderer/api/endpoints/pods.api.ts
```diff
@@ -406,20 +389,19 @@ export class Pod extends WorkloadKubeObject {
   }
 
   hasIssues() {
-    const notReady = !!this.getConditions().find(condition => {
-      return condition.type == "Ready" && condition.status !== "True";
-    });
-    const crashLoop = !!this.getContainerStatuses().find(condition => {
-      const waiting = condition.state.waiting;
+    for (const { type, status } of this.getConditions()) {
```
jakolehm: Why are we changing this? Please revert unless really related to the PR title.

src/renderer/api/endpoints/pods.api.ts
```diff
@@ -473,14 +455,11 @@ export class Pod extends WorkloadKubeObject {
   }
 
   getNodeName() {
-    return this.spec?.nodeName;
+    return this.spec.nodeName;
```
jakolehm: Why are we changing this? Please revert unless really related to the PR title.

src/renderer/api/endpoints/pods.api.ts
```diff
@@ -473,14 +455,11 @@ export class Pod extends WorkloadKubeObject {
...
   }
 
-  getSelectedNodeOs() {
-    if (!this.spec.nodeSelector) return;
-    if (!this.spec.nodeSelector["kubernetes.io/os"] && !this.spec.nodeSelector["beta.kubernetes.io/os"]) return;
-
-    return this.spec.nodeSelector["kubernetes.io/os"] || this.spec.nodeSelector["beta.kubernetes.io/os"];
+  getSelectedNodeOs(): string | undefined {
+    return this.spec.nodeSelector?.["kubernetes.io/os"] || this.spec.nodeSelector?.["beta.kubernetes.io/os"];
```
jakolehm: Why are we changing this? Please revert unless really related to the PR title.

Nokel81:
@jakolehm I think I ran into some exception in `pods.api.ts` which I why I did that refactoring and cleanup

But I don't remember exactly so I'll move it out.

src/renderer/api/endpoints/helm-charts.api.ts
```diff
@@ -31,11 +31,11 @@ export const helmChartsApi = {
       });
   },
 
-  get(repo: string, name: string, readmeVersion?: string) {
+  get(repo: string, name: string, readmeVersion?: string, reqInit?: RequestInit) {
```
aleksfront, line 31: Currently we're not following any guidlines regarding to this, so nothing critical here, but I beleive it should be considered twice because of this quote from Robert C. Martin "Clean Code".

> Functions should have a small number of arguments. No arguments is best, followed by one, two, and three. More than three is very questionable and should be avoided with prejudice.

src/renderer/api/json-api.ts
```diff
@@ -119,48 +118,53 @@ export class JsonApi<D = JsonApiData, P extends JsonApiParams = JsonApiParams> {
       reqInit,
     };
 
-    return cancelableFetch(reqUrl, reqInit).then(res => {
-      return this.parseResponse<D>(res, infoLog);
-    });
```
aleksfront, line 124: Any specific reason we're changing promises to async/await functions?

src/renderer/api/kube-api.ts
```diff
@@ -302,55 +294,85 @@ export class KubeApi<T extends KubeObject = any> {
...
+      return object;
+    }
+
     // custom apis might return array for list response, e.g. users, groups, etc.
     if (Array.isArray(data)) {
       return data.map(data => new KubeObjectConstructor(data));
     }
 
-    return data;
+    return null;
```
aleksfront, line 318: Why we returning `null` here instead of `data`? Couldn't unprocessed `data` be used in some places?

src/common/utils/type-narrowing.ts
```diff
@@ -0,0 +1,89 @@
...
+  return typeof val === "string";
+}
+
+/**
+ * checks if val is of type object and isn't null
+ * @param val the value to be checked
+ */
+export function isObject(val: unknown): val is object {
+  return typeof val === "object" && val !== null;
+}
```
aleksfront, line 77: Arrays will pass here since `typeof [1, 2] === object`.

src/renderer/components/dock/log.store.ts
```diff
@@ -65,72 +73,73 @@ export class LogStore {
...
-  };
+
+    return result.trimEnd().split("\n");
+  }
 
   /**
    * Converts logs into a string array
-   * @returns {number} Length of log lines
+   * @returns Length of log lines
    */
```
aleksfront, line 121: Another signal that such comments should be avoided in overall. Once described they tempted to get away from code and started to provide lies. Moreover, this description was invalid right from the start.

The function code was changed, but description comment left untouched.

src/renderer/components/dock/log.store.ts
```diff
@@ -65,72 +73,73 @@ export class LogStore {
...
   @computed
-  get lines() {
-    const id = dockStore.selectedTabId;
-    const logs = this.podLogs.get(id);
-
-    return logs ? logs.length : 0;
+  get lines(): number {
+    return this.logs.length;
```
aleksfront, line 124: Very nice! 👍

src/renderer/api/json-api.ts
Nokel81: Please see the comments that I made in the initial post for this PR. It explains why I have done these changes.

src/renderer/api/kube-api.ts
Nokel81: It could but that is not the point of a validation function, I guess the other option would be the throw an error.

src/common/utils/type-narrowing.ts
Nokel81: I know, but generally you will want to do additional validation, even if it is just that all keys and values are strings (for `Record<string, string>`).

src/renderer/components/dock/log.store.ts
Nokel81: I *emphatically* do not agree that we should avoid comments because the code might be updated.

We should as a team require in code reviews that all comments should be updated or at least marked as checked whenever we change the functions. This codebase is slowly getting larger and we should leave helpful guides to future maintainers about what the code does.

src/renderer/api/endpoints/helm-charts.api.ts
Nokel81: Maybe, the last two arguments because they are both optional.

Nokel81: I think that this work should be done in a seperate PR

src/renderer/kube-object.store.ts
```diff
@@ -272,17 +279,21 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
-    if (this.context.cluster?.isGlobalWatchEnabled && namespaces.length === 0) {
-      apis.forEach(api => this.watchNamespace(api, "", abortController));
-    } else {
-      apis.forEach(api => {
-        this.loadedNamespaces.forEach((namespace) => {
-          this.watchNamespace(api, namespace, abortController);
-        });
-      });
-    }
+    // This waits for
```
jim-docker: what?

src/renderer/api/json-api.ts
```diff
@@ -119,48 +117,53 @@ export class JsonApi<D = JsonApiData, P extends JsonApiParams = JsonApiParams> {
...
+    let data;
+
+    try {
+      data = text ? JSON.parse(text) : ""; // DELETE-requests might not have response-body
+    } catch (e) {
+      data = text;
+    }
+
+    if (status >= 200 && status < 300) {
+      console.log(data, res);
```
jim-docker: is this for debugging?

src/renderer/api/json-api.ts
```diff
@@ -119,48 +117,53 @@ export class JsonApi<D = JsonApiData, P extends JsonApiParams = JsonApiParams> {
...
+    if (log.method === "GET" && res.status === 403) {
+      this.writeLog({ ...log, error: data });
+      throw data;
+    } else {
+      const error = new JsonApiErrorParsed(data, this.parseError(data, res));
+
+      this.onError.emit(error, res);
+      this.writeLog({ ...log, error });
+      throw error;
+    }
```
jim-docker:
```suggestion
    }
    
    const error = new JsonApiErrorParsed(data, this.parseError(data, res));

    this.onError.emit(error, res);
    this.writeLog({ ...log, error });
    throw error;
```

src/renderer/api/kube-api.ts
```diff
@@ -303,55 +300,85 @@ export class KubeApi<T extends KubeObject = any> {
...
     await this.checkPreferredVersion();
 
-    return this.request
-      .get(this.getUrl({ namespace, name }), { query })
-      .then(this.parseResponse);
+    const res = await this.request.get(this.getUrl({ namespace, name }), { query });
+
+    const parsed = this.parseResponse(res);
+
+    if (Array.isArray(parsed)) {
```
jim-docker, line 346: Remind me, what does it mean if `parsed` is an array? I know `get()` should return a single item but what if the array has only one item?

src/renderer/components/+namespaces/add-namespace-dialog.tsx
```diff
@@ -33,20 +33,18 @@ export class AddNamespaceDialog extends React.Component<Props> {
...
     const { namespace } = this;
     const { onSuccess, onError } = this.props;
 
     try {
-      await namespaceStore.create({ name: namespace }).then(onSuccess);
-      this.close();
+      const created = await namespaceStore.create({ name: namespace });
+
+      onSuccess?.(created);
+      AddNamespaceDialog.close();
```
jim-docker, line 44: should this be in a `finally` block instead?

src/renderer/kube-object.store.ts
```diff
@@ -291,60 +302,47 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
-          if (timedRetry) clearTimeout(timedRetry);
-          timedRetry = setTimeout(() => {
-            api.watch({
-              namespace,
-              abortController,
-              callback
-            });
-          }, 5000);
+          clearTimeout(timedRetry);
+          timedRetry = setTimeout(watch, 5000);
```
jim-docker: don't need this in `else` block

src/renderer/components/+namespaces/add-namespace-dialog.tsx
Nokel81: No, because the user might want to fix what caused the error to happen and try again.

src/renderer/components/+user-management/user-management.tsx
```diff
@@ -13,29 +13,36 @@ import { isAllowedResource } from "../../../common/rbac";
...
         title: "Role Bindings",
         component: RoleBindings,
         url: roleBindingsURL({ query }),
         routePath: roleBindingsRoute.path.toString(),
-      },
-      {
+      });
+    }
+
+    if (isAllowedResource("role")) {
```
jim-docker: what about "clusterrole"?

Nokel81: Ah thanks forgot about that one.

Nokel81:
Closing this in favour of #2256 #2257 #2258 #2259 #2260 #2261 #2262 #2263 #2264 #2265 #2266 #2267 

As all the changes of this PR are now in those 11.

PR #2013, opened 2021-01-24
Use import aliases

jakolehm:
Just for better readability.


ixrock:
Is it might kill auto-import if not supported by IDE/editor?
In general I don't like this idea with folder shortcuts and I would prefer real system relative paths..

Nokel81:
That is something we should be able to check. I tend to find this better since we don't have a bunch of `../../../../`'s in the code (which don't really help with readability). Since I am not generally thinking about how any folder "up" to the common parent when I am looking at a file, I generally reason about it "down" from one of these four root folders.

I like this idea (if auto importing still works)

jakolehm:
Too many conflicts, closing.

PR #2015, opened 2021-01-25
Fix Helm repositories and pod logs integration tests

nevalla:

### DIFF
--- a/integration/__tests__/app.tests.ts
+++ b/integration/__tests__/app.tests.ts
@@ -96,8 +96,7 @@ describe("Lens integration tests", () => {
       });
 
       it("ensures helm repos", async () => {
-        const reposJson = await listHelmRepositories();
-        const repos = JSON.parse(reposJson);
+        const repos = await listHelmRepositories();
 
         if (!repos[0]) {
           fail("Lens failed to add Bitnami repository");
--- a/integration/__tests__/app.tests.ts
+++ b/integration/__tests__/app.tests.ts
@@ -6,10 +6,10 @@
 */
 import { Application } from "spectron";
 import * as utils from "../helpers/utils";
-import { spawnSync, exec } from "child_process";
-import * as util from "util";
+import { spawnSync } from "child_process";
+import { listHelmRepositories } from "../helpers/utils";
+import { fail } from "assert";
 
-export const promiseExec = util.promisify(exec);
 
 jest.setTimeout(60000);
 
@@ -96,8 +96,12 @@ describe("Lens integration tests", () => {
       });
 
       it("ensures helm repos", async () => {
-        const { stdout: reposJson } = await promiseExec("helm repo list -o json");
+        const reposJson = await listHelmRepositories();
         const repos = JSON.parse(reposJson);
+
+        if (!repos[0]) {
+          fail("Lens failed to add Bitnami repository");
+        }
 
         await app.client.waitUntilTextExists("div.repos #message-bitnami", repos[0].name); // wait for the helm-cli to fetch the repo(s)
         await app.client.click("#HelmRepoSelect"); // click the repo select to activate the drop-down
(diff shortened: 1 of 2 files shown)


integration/helpers/utils.ts
```diff
@@ -39,3 +42,19 @@ export async function tearDown(app: Application) {
...
+      return reposJson;
+    } catch {
+      await new Promise(r => setTimeout(r, 2000)); // if no repositories, wait for Lens adding bitnami repository
+
+      return await listHelmRepositories((retries + 1));
+    }
+  }
+
+  return "[]";
+}
```
Nokel81: I really think that this should return `any[]` or maybe even `{ name: string, url: string}[]`

integration/__tests__/app.tests.ts
```diff
@@ -96,9 +96,13 @@ describe("Lens integration tests", () => {
       });
 
       it("ensures helm repos", async () => {
-        const { stdout: reposJson } = await promiseExec("helm repo list -o json");
+        const reposJson = await listHelmRepositories();
         const repos = JSON.parse(reposJson);
 
+        if (!repos[0]) {
+          fail("Lens failed to add Bitnami repository");
+        }
```
Nokel81, line 103: This will just fail the tests on the CLI, see #2016 for my solution.

integration/helpers/utils.ts
jakolehm: I agree with @Nokel81 , feels weird to return `string` and not array of objects (basically parsed json).

nevalla: Actually it's not parsed json that is returned. Should we return parsed json instead?

jakolehm: > Actually it's not parsed json that is returned. Should we return parsed json instead?

I think it would make more sense to return parsed json.

integration/__tests__/app.tests.ts
```diff
@@ -505,19 +509,35 @@ describe("Lens integration tests", () => {
...
         await app.client.click(".list .TableRow:first-child");
         await app.client.waitForVisible(".Drawer");
         await app.client.click(".drawer-title .Menu li:nth-child(2)");
         // Check if controls are available
-        await app.client.waitForVisible(".Logs .VirtualList");
+        await app.client.waitForVisible(".LogList .VirtualList");
         await app.client.waitForVisible(".LogResourceSelector");
-        await app.client.waitForVisible(".LogResourceSelector .SearchInput");
-        await app.client.waitForVisible(".LogResourceSelector .SearchInput input");
+        //await app.client.waitForVisible(".LogSearch .SearchInput");
```
aleksfront, line 535: A line to remove probably.

integration/helpers/utils.ts
nevalla: changed

integration/__tests__/app.tests.ts
```diff
@@ -505,19 +508,35 @@ describe("Lens integration tests", () => {
...
         await app.client.waitUntilTextExists('a[href^="/pods"]', "Pods");
         await app.client.click('a[href^="/pods"]');
+        await app.client.click(".NamespaceSelect");
+        await app.client.keys("kube-system");
+        await app.client.keys("Enter");// "\uE007"
         await app.client.waitUntilTextExists("div.TableCell", "kube-apiserver");
+        let podMenuItemEnabled = false;
+
+        // Wait until extensions are enabled on renderer
+        while (!podMenuItemEnabled) {
```
jakolehm, line 518: Fine for this PR but I think we need to have a better fix for this in future. Buttons should appear when an extension gets loaded.

jakolehm (approved):

integration/__tests__/app.tests.ts
nevalla: Agree

aleksfront (approved):

integration/__tests__/app.tests.ts
nevalla: Oh yes...forgot to remove that line after testing that it's redundant.

aleksfront: Maybe remove it next time (trying not to dismiss an approvements).

PR #2020, opened 2021-01-26
Allow namespace to specified on role creation

jalaziz:
* Allow users to specify the namespace when creating a custom role.
* Default role namespace remains the "global" namespace allowing for
  cluster role creatoin.

### DIFF
--- a/src/renderer/components/+user-management-roles/add-role-dialog.tsx
+++ b/src/renderer/components/+user-management-roles/add-role-dialog.tsx
@@ -5,9 +5,11 @@ import { observable } from "mobx";
 import { observer } from "mobx-react";
 import { Dialog, DialogProps } from "../dialog";
 import { Wizard, WizardStep } from "../wizard";
+import { SubTitle } from "../layout/sub-title";
 import { Notifications } from "../notifications";
 import { rolesStore } from "./roles.store";
 import { Input } from "../input";
+import { NamespaceSelect } from "../+namespaces/namespace-select";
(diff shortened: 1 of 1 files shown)


src/renderer/components/+user-management-roles/add-role-dialog.tsx
```diff
@@ -64,13 +68,20 @@ export class AddRoleDialog extends React.Component<Props> {
             nextLabel="Create"
             next={this.createRole}
           >
+            <SubTitle title="Role Name" />
             <Input
               required autoFocus
-              placeholder={`Role name`}
```
Nokel81, line 69: I prefer the old placeholder as it is descriptive instead of instructive.

jalaziz: I was simply following the convention that is used elsewhere when there's a subtitle and an input (particularly, creating a service account).

Nokel81: Looking at the rest of the repo, I would say that slight majority is actually descriptive instead.

jalaziz: Changed it to descriptive, but not fully the same name as the `Subtitle`. This is similar to what is shown when creating secrets.

Nokel81 (approved):

jakolehm (approved):
LGTM

PR #2027, opened 2021-01-27
Fix extension cluster submenu re-render (port #1996 to master)

jakolehm:
Port of #1996

### DIFF
--- a/src/renderer/components/app.tsx
+++ b/src/renderer/components/app.tsx
@@ -42,7 +42,7 @@ import { ClusterPageMenuRegistration, clusterPageMenuRegistry } from "../../exte
 import { TabLayout, TabLayoutRoute } from "./layout/tab-layout";
 import { StatefulSetScaleDialog } from "./+workloads-statefulsets/statefulset-scale-dialog";
 import { eventStore } from "./+events/event.store";
-import { computed, reaction } from "mobx";
+import { reaction, computed, observable } from "mobx";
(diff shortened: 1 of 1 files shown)


nevalla (approved):

src/renderer/components/app.tsx
```diff
@@ -143,22 +151,38 @@ export class App extends React.Component {
...
         }
       }
     });
+
+    Array.from(this.extensionRoutes.keys()).forEach((menu) => {
+      if (!rootItems.includes(menu)) {
+        this.extensionRoutes.delete(menu);
+      }
+    });
+  }
```
Nokel81, line 182: We shouldn't create an array just to iterate through it once and the throw it away.
```suggestion
    for (const meny of this.extensionRoutes.keys()) {
      if (!rootItems.includes(menu)) {
        this.extensionRoutes.delete(menu);
      }
    }
```

Nokel81 (approved):
Approving, though I don't know if you also want review commets (so I will leave some anyway).

src/renderer/components/app.tsx
jakolehm: I'll create a separate PR about cleanup. I never remember that we can use `for` in these cases.

Nokel81: Ah I see, that is why `forEach` is so common. Though remember that `forEach` is one of the only "methods" on interators, so you don't need to convert to an array to use that method.

PR #2031, opened 2021-01-27
Better extensionRoutes.keys() iteration

jakolehm:


src/renderer/components/app.tsx
```diff
@@ -174,11 +174,11 @@ export class App extends React.Component {
       }
     });
 
-    Array.from(this.extensionRoutes.keys()).forEach((menu) => {
+    for (const menu of this.extensionRoutes.keys()) {
```
ixrock, line 177: How it's better in particular case? There is no `break`-exit when it makes sense for example? 🤔

jakolehm: This is a follow-up PR to review comment here: https://github.com/lensapp/lens/pull/2027#discussion_r565309202

ixrock: Yes, but it was just a code suggestion without any reason (from comment) or I missing something?

jakolehm: `Array.from` creates an additional object for no reason. Does not really matter performance wise but it looks a bit cleaner with plain `for`.

ixrock: Ok, readability it's matter of choice (for me functional-style is more readable)

ixrock (approved):

Nokel81 (approved):

src/renderer/components/app.tsx
Nokel81: I agree that sometimes functional style is more readable, however for the `forEach` case I generally always disagree since it has more syntax and is slower and lacks `break` functionality.

PR #2033, opened 2021-01-27
Detach NamespaceStore from KubeWatchApi, proper store.loadAll()

ixrock:
- Detached `NamespaceStore` from `KubeWatchApi`
- Loading all namespaces by default with `KubeObjectStore.loadAll()`
- Added `KubeObjectStore.loadSelectedNamespaces()`


src/renderer/api/kube-watch-api.ts
```diff
@@ -41,13 +40,13 @@ export interface IKubeWatchLog {
...
-  private cluster: Cluster;
-  private namespaceStore: NamespaceStore;
-
   private requestId = 0;
-  private isConnected = false;
   private reader: ReadableStreamReader<string>;
-  private subscribers = observable.map<KubeApi, number>();
+
+  @observable.ref private getCluster: () => Cluster;
+  @observable.ref private getNamespaces: () => string[];
```
Nokel81: These don't have an initial value so they should be marked as optional

```suggestion
  @observable.ref private getCluster?: () => Cluster;
  @observable.ref private getNamespaces?: () => string[];
```

src/renderer/api/kube-watch-api.ts
```diff
@@ -57,16 +56,13 @@ export class KubeWatchApi {
...
-      if (!cluster.isAllowedResource(api.kind)) {
+    return Array.from(this.subscribers.keys()).map(api => {
+      if (!this.getCluster?.().isAllowedResource(api.kind)) {
         return [];
       }
 
       if (api.isNamespaced) {
-        return namespaceStore.getContextNamespaces().map(namespace => api.getWatchUrl(namespace));
+        return this.getNamespaces().map(namespace => api.getWatchUrl(namespace));
       } else {
```
Nokel81: If the getter isn't defined we should either throw a defined error or return `[]` instead of having the error be "cannot call undefined"

src/renderer/api/kube-watch-api.ts
```diff
@@ -77,14 +73,16 @@ export class KubeWatchApi {
...
-    const { getHostedCluster } = await import("../../common/cluster-store");
-    const { namespaceStore } = await import("../components/+namespaces/namespace.store");
+  setupCluster(getter: () => Cluster) {
+    this.getCluster = getter;
+  }
 
-    await namespaceStore.whenReady;
+  setupWatchingNamespaces(getter: () => string[]) {
+    this.getNamespaces = getter;
+  }
```
Nokel81: imo these should be setters

src/renderer/api/kube-watch-api.ts
```diff
@@ -143,7 +141,7 @@ export class KubeWatchApi {
     }
 
     if (preload) {
-      loading.push(...stores.map(store => store.loadAll()));
+      loading.push(...stores.map(store => store.loadAll(this.getNamespaces())));
```
Nokel81: `getNamespaces` might be undefined here

src/renderer/api/kube-watch-api.ts
ixrock: Maybe, but it's used only internally and not critical imho.

src/renderer/api/kube-watch-api.ts
ixrock: Yes, will fix.

ixrock:
@Nokel81 please re-review again

src/renderer/api/kube-watch-api.ts
Nokel81: This was not changed or commented on, unresolving.

src/renderer/api/kube-watch-api.ts
```diff
@@ -43,50 +42,59 @@ export interface IKubeWatchLog {
...
-  // events
   public onMessage = new EventEmitter<[IKubeWatchMessage]>();
 
+  @observable.ref private getCluster: () => Cluster;
+  @observable.ref private getNamespaces: () => string[];
+  @observable isConnected = false;
+  @observable subscribers = observable.map<KubeApi, number>();
+
+  @computed get isReady(): boolean {
+    return !!(this.cluster && this.namespaces);
```
Nokel81: I have generally come to prefer `Boolean(...)` vs `!!...` especially in places like this that are encapsulation points.

src/renderer/api/kube-watch-api.ts
```diff
@@ -43,50 +42,59 @@ export interface IKubeWatchLog {
...
     return this.apis.length > 0;
   }
 
   @computed get apis(): string[] {
-    const { cluster, namespaceStore } = this;
-    const activeApis = Array.from(this.subscribers.keys());
-
-    return activeApis.map(api => {
-      if (!cluster.isAllowedResource(api.kind)) {
+    return Array.from(this.subscribers.keys()).map(api => {
```
Nokel81, line 67: I recently learned about this, which I think is probably faster since it doesn't have to create the intermediate array.
```suggestion
    return Array.from(this.subscribers.keys(), api => {
```

src/renderer/api/kube-watch-api.ts
```diff
@@ -43,50 +42,59 @@ export interface IKubeWatchLog {
...
+      if (!this.isReady || !this.isAllowedApi(api)) {
```
Nokel81: the `!this.isReady` check should probably be hoisted of the map function.

src/renderer/api/kube-watch-api.ts
ixrock: I believe that such micro-optimizations we should avoid in advance. In most cases you won't notice any performance difference..
If it brings more readability or reasonable performance boost, only then apply (but _it is _not__ in that case + here you gets increased cognitive load for rare-used second argument of `Array.from`).

src/renderer/api/kube-watch-api.ts
ixrock: Why?

ixrock: done

src/renderer/api/kube-watch-api.ts
ixrock: done

src/renderer/api/kube-watch-api.ts
ixrock: fixed, PTAL

src/main/routes/watch-route.ts
```diff
@@ -65,50 +90,72 @@ class ApiWatcher {
...
 
       return;
     }
 
-    response.setHeader("Content-Type", "text/event-stream");
+    this.setResponse(response);
+    response.setHeader("Content-Type", "application/json");
     response.setHeader("Cache-Control", "no-cache");
     response.setHeader("Connection", "keep-alive");
     logger.debug(`watch using kubeconfig:${JSON.stringify(cluster.getProxyKubeconfig(), null, 2)}`);
```
Nokel81: Logger already has this functionality and logging should be single lines
```suggestion
    logger.debug("watch using kubeconfig", { kubeconfig: cluster.getProxyKubeconfig() });
```

Nokel81:
@ixrock Please rebase (and maybe even squash) this against master. It has several conflicting versions (compared to master at least) which makes it hard to review. Thanks.

src/renderer/components/app.tsx
```diff
@@ -177,7 +177,7 @@ export class App extends React.Component {
       }
     });
 
-    for (const menu of this.extensionRoutes.keys()) {
```
Nokel81: Why was this changed? It is both slower and more verbose.

ixrock: Wrong merge, you're faster than my revert-fix :D

Nokel81: lol

ixrock:
rebased to https://github.com/lensapp/lens/pull/2063

PR #2034, opened 2021-01-27
Fix unit tests and make tests fail if a promise rejects after the test harness is finishes

Nokel81:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,6 @@ else
     DETECTED_OS := $(shell uname)
 endif
 
-.PHONY: binaries/client
 binaries/client: node_modules
 	yarn download-bins
 
--- a/Makefile
+++ b/Makefile
@@ -16,16 +16,16 @@ node_modules: yarn.lock
 	yarn install --frozen-lockfile
 	yarn check --verify-tree --integrity
 
-static/build/LensDev.html:
+static/build/LensDev.html: node_modules
 	yarn compile:renderer
 
 .PHONY: compile-dev
-compile-dev:
+compile-dev: node_modules
 	yarn compile:main --cache
 	yarn compile:renderer --cache
 
 .PHONY: dev
-dev: node_modules binaries/client build-extensions static/build/LensDev.html
+dev: binaries/client build-extensions static/build/LensDev.html
 	yarn dev
 
 .PHONY: lint
@@ -77,7 +77,7 @@ $(extension_dists): src/extensions/npm/extensions/dist
 	cd $(@:/dist=) && npm run build
 
 .PHONY: build-extensions
-build-extensions: $(extension_node_modules) $(extension_dists)
+build-extensions: node_modules $(extension_node_modules) $(extension_dists)
 
 .PHONY: test-extensions
 test-extensions: $(extension_node_modules)
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,8 @@
+CMD_ARGS = $(filter-out $@,$(MAKECMDGOALS))
+
+%:
+  @:
+
 EXTENSIONS_DIR = ./extensions
 extensions = $(foreach dir, $(wildcard $(EXTENSIONS_DIR)/*), ${dir})
 extension_node_modules = $(foreach dir, $(wildcard $(EXTENSIONS_DIR)/*), ${dir}/node_modules)
@@ -9,6 +14,7 @@ else
     DETECTED_OS := $(shell uname)
 endif
 
+.PHONY: binaries/client
 binaries/client: node_modules
 	yarn download-bins
 
@@ -34,7 +40,7 @@ lint:
 
 .PHONY: test
 test: binaries/client
-	yarn test
+	yarn run jest $(or $(CMD_ARGS), "src")
 
 .PHONY: integration-linux
 integration-linux: binaries/client build-extension-types build-extensions
@@ -57,10 +63,6 @@ integration-win: binaries/client build-extension-types build-extensions
 	# rm %APPDATA%/Lens
 	yarn build:win
 	yarn integration
-
-.PHONY: test-app
-test-app:
-	yarn test
 
 .PHONY: build
 build: node_modules binaries/client build-extensions
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
     "build:linux": "yarn run compile && electron-builder --linux --dir -c.productName=Lens",
     "build:mac": "yarn run compile && electron-builder --mac --dir -c.productName=Lens",
     "build:win": "yarn run compile && electron-builder --win --dir -c.productName=Lens",
-    "test": "scripts/test.sh",
     "integration": "jest --runInBand integration",
     "dist": "yarn run compile && electron-builder --publish onTag",
     "dist:win": "yarn run compile && electron-builder --publish onTag --x64 --ia32",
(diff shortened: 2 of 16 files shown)


package.json
```diff
@@ -25,7 +25,7 @@
     "build:linux": "yarn run compile && electron-builder --linux --dir -c.productName=Lens",
     "build:mac": "yarn run compile && electron-builder --mac --dir -c.productName=Lens",
     "build:win": "yarn run compile && electron-builder --win --dir -c.productName=Lens",
-    "test": "jest --env=jsdom src $@",
+    "test": "scripts/test.sh",
```
Nokel81: This helps support `yarn test <file-name>` only testing one file

src/renderer/api/__tests__/kube-api-parse.test.ts
```diff
@@ -1,134 +1,115 @@
...
 
 describe("parseApi unit tests", () => {
-  for (const { url, expected } of tests) {
-    test(`testing "${url}"`, () => {
-      expect(parseKubeApi(url)).toStrictEqual(expected);
-    });
-  }
+  it.each(tests)("testing %s", (url, expected) => {
+    expect(parseKubeApi(url)).toStrictEqual(expected);
+  });
```
Nokel81, line 114: This just makes use of the jest table testing format, these specific tests weren't causing the problems.

package.json
jakolehm: I don't really like that we split these into yet-another script (worst case: make -> yarn -> bash). How about:

```
    "jest-jsdom": "jest --env=jsdom",
    "test": "yarn run jest-jsdom src",
```

Now you could run custom targets using `yarn jest-jsdom path/to/file.test.ts`.

src/common/base-store.ts
```diff
@@ -19,7 +19,7 @@ export interface BaseStoreParams<T = any> extends ConfOptions<T> {
  * Note: T should only contain base JSON serializable types.
  */
 export abstract class BaseStore<T = any> extends Singleton {
-  protected storeConfig: Config<T>;
+  protected storeConfig?: Config<T>;
```
jakolehm, line 22: Why are we modifying `BaseStore`?

src/renderer/components/dock/__test__/dock-tabs.test.tsx
```diff
@@ -8,6 +8,20 @@ import { createResourceTab } from "../create-resource.store";
 import { createTerminalTab } from "../terminal.store";
 import { observable } from "mobx";
 
+Object.defineProperty(window, "matchMedia", {
```
jakolehm: This is the actual fix for unhandled exceptions, right?

Nokel81: To some of them. I was also getting "KubeApi.parseApi is not a function" which is why there are those mocks in "kube-api-parse.test.ts".

Nokel81:
@jakolehm I don't know why I cannot respond to your comment about the test script (github won't let me). However, I don't see why having another `package.json` "scripts" entry is better then a script file? Plus the script file method is strictly more powerful since a developer doesn't need to remember to switch to `jest-jsdom` just because they want to test a specific file.

src/common/base-store.ts
Nokel81: Because one of the errors I encountered was due to a call to `onModelChange` in the `user-store.test.ts` happening after the test suite was finished (because we don't handle rejected promises very well) and since in that test `isLoaded` was never true the function `enableSync` wasn't called so `storeConfig` was never set.

These changes merely make `BaseStore` more robust by marking the types that get late initialized as such.

package.json
aleksfront: Probably we should't do such changes. If we eliminate `"testEnvironment": "node",` from our `package.json`, tests still works okay and `jsdom` set as default value. So we don't need passing extra vars `--env=jsdom` or
```
/**
 * @jest-environment jsdom
 */
```
declaration in specific files.

scripts/test.sh
```diff
@@ -0,0 +1 @@
+jest --env=jsdom ${1:-src}
```
aleksfront: No need for this file (see my previous comment).

src/renderer/components/dock/__test__/dock-tabs.test.tsx
aleksfront: This test is fixed without needed to mock `matchMedia`.
https://github.com/lensapp/lens/pull/2037. But we can remeber such cases where cleaning test file isn't enough.

aleksfront (changes requested):
Most of the errors are gone which is really great! But I've got another one introduced (related to xterm somehow). They're appeared after few test files (not only one presented in screenshot).

xterm errors

@Nokel81 Can you please run tests again and check this one?

package.json
Nokel81: Ah so we could just remove the "test" script all together then, brilliant!

Nokel81:
@aleksfront Yes I know about those warnings. It has to do with `jest-canvas-mock` being very simplistic. See my comments on https://github.com/hustcc/jest-canvas-mock/issues/60. Basically, `xterm.js` uses canvas to do its css colour parsing but `jest-canvas-mock` always returns `[0, 0, 0, 0]` which triggers this warning.

I decided against trying to get `node-canvas` working within `jest` since it involves a bunch of other depenencies but I guess I could if we wanted to. I looked into trying to mock that function to return something like `[0, 0, 0, 255]` which would block those warnings.

scripts/test.sh
Nokel81: Sounds good

Nokel81:
@aleksfront @jakolehm would you be able to re-review?

src/common/__tests__/cluster-store.test.ts
```diff
@@ -35,7 +35,8 @@ jest.mock("electron", () => {
     app: {
       getVersion: () => "99.99.99",
       getPath: () => "tmp",
-      getLocale: () => "en"
+      getLocale: () => "en",
+      setLoginItemSettings: jest.fn(),
```
aleksfront, line 39: Can you please remember why this line needed? I've checked test with/without it, and results are equal.

src/renderer/components/dock/__test__/dock-tabs.test.tsx
aleksfront: Again, this should be removed for now. Test passes without these changes ^

aleksfront (commented):
Posted minor cleaning comments.

@Nokel81 Changes introduced in `Makefile` isn't unclear to me. Can you please explain why we added `node_modules` part to most of the commands, and why we need to add `CMD_ARGS`?

src/common/__tests__/cluster-store.test.ts
Nokel81: This is required if a `mobx` reaction fires to a change on the userPreferences.openAtLogin. Since that reaction is classified as "fireImmediately" and is in an async function, it could fire at any time.

src/renderer/components/dock/__test__/dock-tabs.test.tsx
Nokel81: Done.

Nokel81:
@aleksfront I have made the requested changes.

aleksfront:
@jakolehm Can you please re-review changes in `Makefile`? I'm about to approve this PR if changes there are okay.

Makefile
```diff
@@ -9,23 +14,24 @@ else
     DETECTED_OS := $(shell uname)
 endif
 
+.PHONY: binaries/client
```
jakolehm: Why do we want to download client binaries on every run?

Nokel81: I think I added this because it doesn't doesn't check correctly the folder modification date. Plus the downloaders already check and verify the current versions before downloading again.

Nokel81:
@jakolehm and @aleksfront removed the one requested change from the Makefile

aleksfront (approved):

jakolehm (approved):

PR #2036, opened 2021-01-28
Add the ability to hide metrics from the UI

### LINKED ISSUE #761: Allow user to disable the Metrics panel in the pod view.
Hi!

I did some digging and it doesn't seem to be possible. I'd like to disable the metrics feature as it's not something my current K8s cluster can offer.

I can take a look at hacking my way to get that possible 🤔 

Thanks!

### LINKED ISSUE #1558: Ability to hide metrics from the UI
A setting so that metrics are hidden from the UI.

**Why is this needed**:
The metrics support in Lens is very buggy and does not play well with existing deployments of prometheus. There is next to no documentation on how to get your deployed prometheus to work with Lens so we end up not using it. There are many issues about this already and has not been fixed. Would like an option to just hide the empty graphs all together.

The most "docs" I've seen is random replacement regex's in github issues that some people said worked for them and others said didn't work for them.

- Desktop OS: Ubuntu 20.04

vshakirova:
Hi, lens team. Just a small feature. Not sure about the layout. 
Screen Shot 2021-02-25 at 5 04 37 PM

Fixes: https://github.com/lensapp/lens/issues/1558 
Fixes: https://github.com/lensapp/lens/issues/761

### DIFF
--- a/src/common/cluster-store.ts
+++ b/src/common/cluster-store.ts
@@ -15,6 +15,7 @@ import { handleRequest, requestMain, subscribeToBroadcast, unsubscribeAllFromBro
 import _ from "lodash";
 import move from "array-move";
 import type { WorkspaceId } from "./workspace-store";
+import { ResourceType } from "../renderer/components/+cluster-settings/components/cluster-metrics-setting";
 
 export interface ClusterIconUpload {
   clusterId: string;
@@ -34,7 +35,7 @@ export type ClusterPrometheusMetadata = {
 
 export interface ClusterStoreModel {
   activeCluster?: ClusterId; // last opened cluster
-  clusters?: ClusterModel[]
+  clusters?: ClusterModel[];
 }
 
 export type ClusterId = string;
@@ -70,12 +71,13 @@ export interface ClusterModel {
   kubeConfig?: string; // yaml
 }
 
-export interface ClusterPreferences extends ClusterPrometheusPreferences{
+export interface ClusterPreferences extends ClusterPrometheusPreferences {
   terminalCWD?: string;
   clusterName?: string;
   iconOrder?: number;
   icon?: string;
   httpsProxy?: string;
+  hiddenMetrics?: string[];
 }
 
 export interface ClusterPrometheusPreferences {
@@ -209,6 +211,10 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
 
   isActive(id: ClusterId) {
     return this.activeCluster === id;
+  }
+
+  isMetricHidden(resource: ResourceType) {
+    return Boolean(this.active?.preferences.hiddenMetrics?.includes(resource));
   }
 
   @action
--- a/src/renderer/components/+cluster-settings/components/cluster-metrics-setting.tsx
+++ b/src/renderer/components/+cluster-settings/components/cluster-metrics-setting.tsx
@@ -1,0 +1,118 @@
+import "./cluster-metrics-setting.scss";
+
+import React from "react";
+import { disposeOnUnmount, observer } from "mobx-react";
+import { Select, SelectOption } from "../../select/select";
+import { Icon } from "../../icon/icon";
+import { Button } from "../../button/button";
+import { SubTitle } from "../../layout/sub-title";
+import { Cluster } from "../../../../main/cluster";
+import { observable, reaction } from "mobx";
+
+interface Props {
+  cluster: Cluster;
+}
+
+export enum ResourceType {
+  Cluster = "Cluster",
+  Node = "Node",
+  Pod = "Pod",
+  Deployment = "Deployment",
+  StatefulSet = "StatefulSet",
+  Container = "Container",
+  Ingress = "Ingress",
+  VolumeClaim = "VolumeClaim",
+  ReplicaSet = "ReplicaSet",
+  DaemonSet = "DaemonSet",
+}
+
+@observer
+export class ClusterMetricsSetting extends React.Component<Props> {
+  @observable hiddenMetrics = observable.set<string>();
+
+  componentDidMount() {
+    this.hiddenMetrics = observable.set<string>(this.props.cluster.preferences.hiddenMetrics ?? []);
+
+    disposeOnUnmount(this, [
+      reaction(() => this.props.cluster.preferences.hiddenMetrics, () => {
+        this.hiddenMetrics = observable.set<string>(this.props.cluster.preferences.hiddenMetrics ?? []);
+      }),
+    ]);
+  }
+
+  save = () => {
+    this.props.cluster.preferences.hiddenMetrics = Array.from(this.hiddenMetrics);
+  };
+
+  onChangeSelect = (values: SelectOption<ResourceType>[]) => {
+    for (const { value } of values) {
+      if (this.hiddenMetrics.has(value)) {
+        this.hiddenMetrics.delete(value);
+      } else {
+        this.hiddenMetrics.add(value);
+      }
+    }
+    this.save();
+  };
+
+  onChangeButton = () => {
+    Object.keys(ResourceType).map(value =>
+      this.hiddenMetrics.add(value)
+    );
+    this.save();
+  };
+
+  reset = () => {
+    this.hiddenMetrics.clear();
+    this.save();
+  };
+
+  formatOptionLabel = ({ value: resource }: SelectOption<ResourceType>) => (
+    <div className="flex gaps align-center">
+      <span>{resource}</span>
+      {this.hiddenMetrics.has(resource) && <Icon smallest material="check" className="box right" />}
+    </div>
+  );
+
+  renderMetricsSelect() {
+
+    return (
+      <>
+        <Select
+          className="box grow"
+          placeholder="Select metrics to hide..."
+          isMulti
+          isSearchable
+          onMenuClose={this.save}
+          closeMenuOnSelect={false}
+          controlShouldRenderValue={false}
+          options={Object.values(ResourceType)}
+          onChange={this.onChangeSelect}
+          formatOptionLabel={this.formatOptionLabel}
+        />
+        <Button
+          primary
+          label="Hide all metrics"
+          onClick={this.onChangeButton}
+        />
+        <Button
+          primary
+          label="Reset"
+          onClick={this.reset}
+        />
+      </>
+    );
+  }
+
+  render() {
+
+    return (
+      <div className="MetricsSelect">
+        <SubTitle title={"Hide metrics from the UI"}/>
+        <div className="flex gaps">
+          {this.renderMetricsSelect()}
+        </div>
+      </div>
+    );
+  }
+}
--- a/src/renderer/components/+cluster-settings/components/show-metrics.tsx
+++ b/src/renderer/components/+cluster-settings/components/show-metrics.tsx
@@ -1,0 +1,63 @@
+import "./cluster-metrics-setting.scss";
+
+import React from "react";
+import { disposeOnUnmount, observer } from "mobx-react";
+import { Cluster } from "../../../../main/cluster";
+import { observable, reaction } from "mobx";
+import { Badge } from "../../badge/badge";
+import { Icon } from "../../icon/icon";
+
+interface Props {
+  cluster: Cluster;
+}
+
+@observer
+export class ShowMetricsSetting extends React.Component<Props> {
+  @observable hiddenMetrics = observable.set<string>();
+
+  componentDidMount() {
+    this.hiddenMetrics = observable.set<string>(this.props.cluster.preferences.hiddenMetrics ?? []);
+
+    disposeOnUnmount(this, [
+      reaction(() => this.props.cluster.preferences.hiddenMetrics, () => {
+        this.hiddenMetrics = observable.set<string>(this.props.cluster.preferences.hiddenMetrics ?? []);
+      }),
+    ]);
+  }
+
+  removeMetric(metric: string) {
+    this.hiddenMetrics.delete(metric);
+    this.props.cluster.preferences.hiddenMetrics = Array.from(this.hiddenMetrics);
+  }
+
+  renderMetrics() {
+
+    return (
+
+      Array.from(this.hiddenMetrics).map(name => {
+        const tooltipId = `${name}`;
+
+        return (
+          <Badge key={name}>
+            <span id={tooltipId}>{name}</span>
+            <Icon
+              smallest
+              material="clear"
+              onClick={() => this.removeMetric(name)}
+              tooltip="Remove"
+            />
+          </Badge>
+        );
+      })
+    );
+  }
+
+  render() {
+
+    return (
+      <div className="MetricsSelect flex wrap gaps">
+        {this.renderMetrics()}
+      </div>
+    );
+  }
+}
--- a/src/renderer/components/+cluster/cluster-overview.scss
+++ b/src/renderer/components/+cluster/cluster-overview.scss
@@ -7,7 +7,9 @@
   display: grid;
   grid-gap: $gridGap;
   grid-template-columns: repeat(auto-fit, minmax(500px, 1fr));
-  grid-template-rows: 1fr 1fr;
+  grid-template-rows:
+          [row1-start] 1fr
+          [row2-start] 1fr;
 
   @include media(">1600px") {
     grid-template-columns: 1fr 1fr;
--- a/src/renderer/components/+cluster/cluster-overview.tsx
+++ b/src/renderer/components/+cluster/cluster-overview.tsx
@@ -5,7 +5,7 @@ import { reaction } from "mobx";
 import { disposeOnUnmount, observer } from "mobx-react";
 import { nodesStore } from "../+nodes/nodes.store";
 import { podsStore } from "../+workloads-pods/pods.store";
-import { getHostedCluster } from "../../../common/cluster-store";
+import { clusterStore, getHostedCluster } from "../../../common/cluster-store";
 import { interval } from "../../utils";
 import { TabLayout } from "../layout/tab-layout";
 import { Spinner } from "../spinner";
@@ -13,6 +13,7 @@ import { ClusterIssues } from "./cluster-issues";
 import { ClusterMetrics } from "./cluster-metrics";
 import { clusterOverviewStore } from "./cluster-overview.store";
 import { ClusterPieCharts } from "./cluster-pie-charts";
+import { ResourceType } from "../+cluster-settings/components/cluster-metrics-setting";
 
 @observer
 export class ClusterOverview extends React.Component {
@@ -37,19 +38,40 @@ export class ClusterOverview extends React.Component {
     this.metricPoller.stop();
   }
 
+  renderMetrics(isMetricsHidden: boolean) {
+    if (isMetricsHidden) {
+      return null;
+    }
+
+    return (
+      <>
+        <ClusterMetrics/>
+        <ClusterPieCharts/>
+      </>
+    );
+  }
+
+  renderClusterOverview(isLoaded: boolean, isMetricsHidden: boolean) {
+    if (!isLoaded) {
+      return <Spinner center/>;
+    }
+
+    return (
+      <>
+        {this.renderMetrics(isMetricsHidden)}
+        <ClusterIssues className={isMetricsHidden ? "OnlyClusterIssues" : ""}/>
+      </>
+    );
+  }
+
   render() {
     const isLoaded = nodesStore.isLoaded && podsStore.isLoaded;
+    const isMetricsHidden = clusterStore.isMetricHidden(ResourceType.Cluster);
 
     return (
       <TabLayout>
         <div className="ClusterOverview">
-          {!isLoaded ? <Spinner center/> : (
-            <>
-              <ClusterMetrics/>
-              <ClusterPieCharts/>
-              <ClusterIssues/>
-            </>
-          )}
+          {this.renderClusterOverview(isLoaded, isMetricsHidden)}
         </div>
       </TabLayout>
     );
--- a/src/renderer/components/chart/pie-chart.scss
+++ b/src/renderer/components/chart/pie-chart.scss
@@ -1,6 +1,7 @@
 .PieChart {
   .chart-container {
     width: 120px;
+    min-height: 150px;
   }
 
   .legend {
--- a/src/common/user-store.ts
+++ b/src/common/user-store.ts
@@ -29,7 +29,7 @@ export interface UserPreferences {
   downloadBinariesPath?: string;
   kubectlBinariesPath?: string;
   openAtLogin?: boolean;
-  hiddenTableColumns?: Record<string, string[]>
+  hiddenTableColumns?: Record<string, string[]>;
 }
 
 export class UserStore extends BaseStore<UserStoreModel> {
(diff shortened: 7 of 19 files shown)


aleksfront:
@vshakirova Please do not forget about other pages where metrics is presented.
* Cluster overview page `src/renderer/components/+cluster/cluster-overview.tsx`
* Node list `src/renderer/components/+nodes/nodes.tsx`

vshakirova:
> @vshakirova Please do not forget about other pages where metrics is presented.
> 
> * Cluster overview page `src/renderer/components/+cluster/cluster-overview.tsx`
> * Node list `src/renderer/components/+nodes/nodes.tsx`

@aleksfront yes, thank you. I'll do, but PR is still in progress

Nokel81:
If the PR is still in progress, can you please mark it as a draft PR?

vshakirova:
> If the PR is still in progress, can you please mark it as a draft PR?

It is ready for review 🙂

src/renderer/components/+preferences/select-metrics-dialog.tsx
```diff
@@ -0,0 +1,101 @@
...
+    SelectMetricsDialog.isOpen = false;
+  }
+
+  close = () => {
+    SelectMetricsDialog.close();
+  };
+
+  @action
+  onChange(value: ResourceType) {
+    userStore.hiddenMetrics.has(value) ? userStore.hiddenMetrics.delete(value) : userStore.hiddenMetrics.add(value);
```
Nokel81: Given that we are not using the value of this action. I don't see why we need to use the ternary.

```suggestion
    if (userStore.hiddenMetrics.has(value)) {
      userStore.hiddenMetrics.delete(value);
    } else {
      userStore.hiddenMetrics.add(value);
    }
```

src/renderer/components/+preferences/select-metrics-dialog.tsx
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  reset = () => {
+    userStore.hiddenMetrics.clear();
+    userStore.preferences.hideMetrics = false;
+  };
+
+  render() {
+    const { ...dialogProps } = this.props;
+    let placeholder = <>Select metrics..</>;
```
Nokel81: please extract to `getPlaceHolder()` function. For both readability and for removing a brace-less if statement.

Nokel81: Also this should use a full ellipsis: `...`

src/renderer/components/+preferences/select-metrics-dialog.tsx
```diff
@@ -0,0 +1,101 @@
...
+      >
+        <Wizard header={header} done={this.close}>
+          <WizardStep contentClass="flow column" nextLabel={"Reset"} next={this.reset} prevLabel={"Close"}>
+            <div>
+              <Select
+                placeholder={placeholder}
+                isMulti
+                closeMenuOnSelect={false}
+                isDisabled={userStore.preferences.hideMetrics}
+                options={Object.values(ResourceType).map(res => res)}
```
Nokel81: Is the `map` required? It seems that it wouldn't do anything.

src/renderer/components/+preferences/select-metrics-dialog.tsx
```diff
@@ -0,0 +1,101 @@
...
+                onChange={([{ value }]: SelectOption[]) => {
+                  this.onChange(value);
+                }}
```
Nokel81: I think it would be better if `this.onChange` accepted multiple options itself, so it could just be passed in here.

vshakirova: Could you please elaborate a bit more? Do you mean in `onChange(resource: ResourceType[])`?

src/renderer/components/+preferences/select-metrics-dialog.tsx
vshakirova: thank you, I'll fix it

vshakirova: thank you, I'll fix it

src/renderer/components/+preferences/select-metrics-dialog.tsx
Nokel81: Yeah that is what the signature should be

vshakirova: yeah.. I tried it and unfortunately `Select` didn't work with `onChange(resource: ResourceType[]`
 I am confused about this moment.

vshakirova: It seems I understand what happens. `([{ value }]: SelectOption[])` gets and passes only single element to `onChange(resource: ResourceType[])` func. That is why ` ResourceType[]` doesn't work. 
I changed inside `Select`
`           onChange={(value: SelectOption[]) => {
                  this.onChange(value.map(e => e.value));
                }}`
and `onChange(resource: ResourceType[])` works fine.

src/renderer/components/+cluster/cluster-overview.scss
```diff
@@ -7,7 +7,7 @@
   display: grid;
   grid-gap: $gridGap;
   grid-template-columns: repeat(auto-fit, minmax(500px, 1fr));
-  grid-template-rows: 1fr 1fr;
+  grid-template-rows: auto;
```
vshakirova: Not sure it is a good solution. Just want to automatically expand `Warnings` and use all available space if cluster metrics are hidden. 

Screen Shot 2021-02-16 at 1 35 48 PM

src/renderer/components/+cluster/cluster-overview.tsx
```diff
@@ -39,14 +41,19 @@ export class ClusterOverview extends React.Component {
...
-              <ClusterPieCharts/>
+              {!isMetricHidden &&
+                <ClusterMetrics/>
+              }
+              {!isMetricHidden &&
+                <ClusterPieCharts/>
+              }
               <ClusterIssues/>
             </>
           )}
```
Nokel81: This is getting rather complex for inline conditional rendering, would you mind refactoring it out into a seperate function, say `renderClusterOverview()`?

```ts
renderMetrics(isMetricsHidden: boolean) {
  if (isMetricsHidden) {
    return null;
  }

  return (
    <>
      <ClusterMetrics/>
      <ClusterPieCharts/>
    </>
  );
}

renderClusterOverview(isLoaded: boolean, isMetricHidden: boolean) {
  if (!isLoaded) {
    return <Spinner center/>;
  }

  return (
    <>
      {this.renderMetrics(isMetricsHidden)}
      <ClusterIssues/>
    </>
  );
}
```

src/renderer/components/+preferences/select-metrics-dialog.tsx
```diff
@@ -0,0 +1,124 @@
...
+    SelectMetricsDialog.isOpen = true;
+  }
+
+  static close() {
+    SelectMetricsDialog.isOpen = false;
+  }
+
+  close = () => {
+    SelectMetricsDialog.close();
+  };
```
Nokel81: Do we need a method as well as the static function?

src/renderer/components/+preferences/select-metrics-dialog.tsx
```diff
@@ -0,0 +1,124 @@
...
+        {...dialogProps}
+        isOpen={SelectMetricsDialog.isOpen}
+        close={this.close}
+      >
+        <Wizard header={header} done={this.close}>
+          <WizardStep contentClass="flow column" nextLabel={"Reset"} next={this.reset} prevLabel={"Close"}>
+            <div>
+              <Select
+                className={"SelectMetricsDialog"}
+                components={{ Placeholder }}
```
Nokel81: Why are there double `{` here? Does that have special meaning, if not please remove.

src/renderer/components/+cluster/cluster-overview.scss
Nokel81: A better solution would be to name grid row areas and switch between them in JS.

src/renderer/components/+cluster/cluster-overview.tsx
vshakirova: thank you, I'll do

src/renderer/components/+preferences/select-metrics-dialog.tsx
vshakirova: I can delete `close` method and change in dialog     
```
  return (
      <Dialog  
         .........
          close={() => SelectMetricsDialog.close()}
      >
      ......
)
```
 Wdyt?

src/renderer/components/+preferences/select-metrics-dialog.tsx
vshakirova: If I'm not mistaken first {} for JS expression, second {}  for { key: value }  JS object
Here is an example 
https://react-select.com/components#defining-components

src/renderer/components/+preferences/select-metrics-dialog.tsx
Nokel81:
```typescript
return (
      <Dialog  
         .........
          close={SelectMetricsDialog.close}
      >
      ......
)
```

Static functions don't need "this" so they are fine to be used in a point-free style

src/renderer/components/+preferences/select-metrics-dialog.tsx
Nokel81: Yes thanks

src/renderer/components/+preferences/select-metrics-dialog.tsx
vshakirova: thank you. I'll fix

src/renderer/components/+cluster/cluster-overview.scss
```diff
@@ -7,7 +7,9 @@
   display: grid;
   grid-gap: $gridGap;
   grid-template-columns: repeat(auto-fit, minmax(500px, 1fr));
-  grid-template-rows: 1fr 1fr;
+  grid-template-rows:
```
vshakirova, line 10: @Nokel81 Please let me know what do you think about this. I tried to name grid row areas

src/renderer/components/chart/pie-chart.scss
```diff
@@ -1,6 +1,7 @@
 .PieChart {
   .chart-container {
     width: 120px;
+    min-height: 150px;
```
vshakirova, line 4: I noticed an issue. When I turn on the metrics again pie-charts are not displayed (without refresh). I see in devTools that their height is equal to 0. Could  I set `min-height` to fix it? It works for me 
Screen Shot 2021-02-18 at 10 21 23 PM

src/renderer/components/+preferences/select-metrics-dialog.tsx
```diff
@@ -0,0 +1,120 @@
...
+              />
+            </div>
+            <Checkbox
+              label="Hide metrics for all resources"
+              value={userStore.preferences.hideMetrics}
+              onChange={v => userStore.preferences.hideMetrics = v}
+            />
+          </WizardStep>
+        </Wizard>
+      </Dialog>
```
Nokel81: Sorry for not noticing this sooner. In trying to find a better solution to the `min-height` problem on "re-showing" the metrics. I had to use this a lot and I personally find it very cumbersome.

Would you be able to remake this aspect of the PR more in line with how the rest of settings is set out:

1. No dialog or Wizard
2. Should be able to see which metrics are disabled
3. The title should be "Hide Metrics" or something to that affect
4. The placeholder text should correctly indicate what is happening
5. Maybe use something like what is used for Helm repos, which lets the user easily clear multiple items.

vshakirova: Thank you for your review and time.  I will do my best to make it more usable. I've changed it completely. Wdyt? (I'm not sure that we have to show hidden metrics in placeholder, because they are placed below)
One more question. What do you think about `min-height` and named grid row areas? I just want to fix all comments at once

Screen Shot 2021-02-19 at 4 01 35 PM
Screen Shot 2021-02-19 at 4 04 37 PM

Nokel81: 1. From my investigation the `min-height` solution works (and I wasn't able to figure out why it wasn't working) so lets leave it.
2. The named `grid-rows` is good
3. I agree that not having the names in the select is good.
4. The `X`'s beside the hidden metric names below the select seem to be slightly larger (taller) than the font size of the names. Could you make them the same?

Other than that looks good.

Nokel81: Finally, after some discussion we think that this would be better as a cluster setting rather than a lens wide setting (since metrics might be available for some clusters but not for others), wdyt?

vshakirova: Agree. I will move this feature to cluster settings

src/renderer/components/+cluster-settings/components/cluster-metrics-setting.tsx
```diff
@@ -0,0 +1,118 @@
...
+export class ClusterMetricsSetting extends React.Component<Props> {
+  @observable hiddenMetrics = observable.set<string>();
+
+  componentDidMount() {
+    this.hiddenMetrics = observable.set<string>(this.props.cluster.preferences.hiddenMetrics);
+  }
+
+  save = () => {
+    this.props.cluster.preferences.hiddenMetrics = Array.from(this.hiddenMetrics);
+  };
```
Nokel81: Because `preferences` is already an observable. I know that other settings have done it in a similar way but we can make this simpler.
```suggestion
  @computed get hiddenMetrics(): Set<string> {
    return new Set(this.props.cluster.preferences.hiddenMetrics);
  }
  
  componentDidMount() {
    disposeOnUnmount(this, [
      reaction(() => this.hiddenMetrics, newHiddenMetrics => {
        this.props.cluster.preferences.hiddenMetrics = Array.from(newHiddenMetrics);
      }),
    ]);
  }
```

src/renderer/components/+cluster-settings/components/cluster-metrics-setting.tsx
```diff
@@ -0,0 +1,118 @@
...
+  onChangeSelect(values: ResourceType[]) {
+    values.map(value => {
+      if (this.hiddenMetrics.has(value)) {
+        this.hiddenMetrics.delete(value);
+      } else {
+        this.hiddenMetrics.add(value);
+      }
+    });
+    this.save();
+  }
```
Nokel81: The above would make this function:
```suggestion
  @action
  onChangeSelect(values: ResourceType[]) {
    values.map(value => {
      if (this.hiddenMetrics.has(value)) {
        this.hiddenMetrics.delete(value);
      } else {
        this.hiddenMetrics.add(value);
      }
    });
  }
```

Because the `reaction` "saves" it and the `@action` means that the reaction only gets called once after the function ends.

src/renderer/components/+cluster-settings/components/cluster-metrics-setting.tsx
```diff
@@ -0,0 +1,118 @@
...
+    Object.keys(ResourceType).map(value =>
+      this.hiddenMetrics.add(value)
+    );
+    this.save();
+  };
+
+  reset = () => {
+    this.hiddenMetrics.clear();
+    this.save();
+  };
```
Nokel81, line 68: These two functions don't need `this.save()` anymore

src/renderer/components/+cluster-settings/components/show-metrics.tsx
```diff
@@ -0,0 +1,58 @@
...
+  @observable hiddenMetrics = observable.set<string>();
+
+  componentDidMount() {
+    this.hiddenMetrics = observable.set<string>(this.props.cluster.preferences.hiddenMetrics);
+  }
+
+  removeMetric = (metric: string) => {
+    this.hiddenMetrics.delete(metric);
+    this.props.cluster.preferences.hiddenMetrics = Array.from(this.hiddenMetrics);
+  };
```
Nokel81: The same here can be done as in the hide-metrics component.

src/renderer/components/+cluster-settings/components/show-metrics.tsx
```diff
@@ -0,0 +1,58 @@
...
+            <Icon
+              small
+              material="clear"
+              onClick={() => this.removeMetric(name)}
+              tooltip="Remove"
+            />
+          </Badge>
+        );
+      })
+    );
```
Nokel81: For rendering we should be using the set. 

```suggestion
    return Array.from(this.hiddenMetrics)
      .map(name => (
        <Badge key={name}>
          <span id={name}>{name}</span>
          <Icon
            smallest
            material="clear"
            onClick={() => this.removeMetric(name)}
            tooltip="Remove"
          />
        </Badge>
      ));
```

Nokel81 (commented):
Thanks for those changes. Very close. Only minor comments with the settings components themselves.

src/renderer/components/+cluster-settings/components/cluster-metrics-setting.tsx
aleksfront: @Nokel81 Please reconsider if this new decision will get things simpler. At first glance yes, we eliminate one `save` method, but we also introducing `disposeOnUnmount` here with a circular dependency (as it looks for me at first glance). We react on `this.hiddenMetrics` change and then actually rewrite same value we react on?

Also, not quite clear how we can make some operations like `this.hiddenMetrics.add(value)` on `computed` value (which is getter). In complex this adds a more congnitive load, than previous decison.

src/common/cluster-store.ts
```diff
@@ -211,6 +213,10 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
     return this.activeCluster === id;
   }
 
+  isMetricHidden(resource: ResourceType) {
+    return clusterStore.getById(clusterStore.activeCluster).preferences.hiddenMetrics.includes(resource);
```
Nokel81: We shouldn't be referencing an instance variable within the class, plus `active` might be `null` and `hiddenMetrics` might be `undefined`.
```suggestion
    return Boolean(this.active?.preferences.hiddenMetrics?.includes(resource));
```

src/renderer/components/+cluster-settings/components/cluster-metrics-setting.tsx
Nokel81: You are correct. Thanks for catching this. I shall reject these comments and make better ones (since I did find some errors while trying this out myself).

src/renderer/components/+cluster-settings/components/cluster-metrics-setting.tsx
```diff
@@ -0,0 +1,118 @@
...
+  ReplicaSet = "ReplicaSet",
+  DaemonSet = "DaemonSet",
+}
+
+@observer
+export class ClusterMetricsSetting extends React.Component<Props> {
+  @observable hiddenMetrics = observable.set<string>();
+
+  componentDidMount() {
+    this.hiddenMetrics = observable.set<string>(this.props.cluster.preferences.hiddenMetrics);
```
Nokel81: Could be undefined

```suggestion
    this.hiddenMetrics = observable.set<string>(this.props.cluster.preferences.hiddenMetrics ?? []);
```

src/renderer/components/+cluster-settings/components/cluster-metrics-setting.tsx
Nokel81: This way we don't need the lamba in two places. Plus we really shouldn't be using `Array.map` here.

```suggestion
  onChangeSelect = (values: SelectOption<ResourceType>[]) => {
    for (const { value } of values) {
      if (this.hiddenMetrics.has(value)) {
        this.hiddenMetrics.delete(value);
      } else {
        this.hiddenMetrics.add(value);
      }
    }
    this.save();
  }
```

src/renderer/components/+cluster-settings/components/cluster-metrics-setting.tsx
```diff
@@ -0,0 +1,118 @@
...
+  formatOptionLabel = ({ value: resource }: SelectOption) => {
+    const isSelected = this.hiddenMetrics.has(resource);
+
+    return (
+      <div className="flex gaps align-center">
+        <span>{resource}</span>
+        {isSelected && <Icon small material="check" className="box right"/>}
+      </div>
+    );
+  };
```
Nokel81: This fixes the height issue (the change to `smallest`)

```suggestion
  formatOptionLabel = ({ value: resource }: SelectOption<ResourceType>) => (
    <div className="flex gaps align-center">
      <span>{resource}</span>
      {this.hiddenMetrics.has(resource) && <Icon smallest material="check" className="box right" />}
    </div>
  );
```

src/renderer/components/+cluster-settings/components/cluster-metrics-setting.tsx
```diff
@@ -0,0 +1,118 @@
...
+          placeholder="Select metrics to hide..."
+          isMulti
+          isSearchable
+          onMenuClose={this.save}
+          closeMenuOnSelect={false}
+          controlShouldRenderValue={false}
+          options={Object.values(ResourceType)}
+          onChange={(options: SelectOption[]) => {
+            this.onChangeSelect(options.map(option => option.value));
+          }}
```
Nokel81:
```suggestion
          onChange={this.onChangeSelect}
```

src/renderer/components/+cluster-settings/components/show-metrics.tsx
```diff
@@ -0,0 +1,58 @@
...
+interface Props {
+  cluster: Cluster;
+}
+
+@observer
+export class ShowMetricsSetting extends React.Component<Props> {
+  @observable hiddenMetrics = observable.set<string>();
+
+  componentDidMount() {
+    this.hiddenMetrics = observable.set<string>(this.props.cluster.preferences.hiddenMetrics);
```
Nokel81: Can be undefined.

```suggestion
    this.hiddenMetrics = observable.set<string>(this.props.cluster.preferences.hiddenMetrics ?? []);
```

Nokel81 (commented):
Sorry about the previous round of suggestions. I was getting ahead of myself. Thanks @aleksfront for catching that. 

Here are some much simpler suggestions and bug fixes.

vshakirova:
> Sorry about the previous round of suggestions. I was getting ahead of myself. Thanks @aleksfront for catching that.
> 
> Here are some much simpler suggestions and bug fixes.
 
 Thank you guys for your time. Finally, I fixed all comments. But I found one more issue. It seems that observable variable (hidden metrics) has to be updated all the time. Otherwise, I get inconsistency between `select-metrics` and `show-metrics`. My suggestion is to use `reaction`. Wdyt?

src/renderer/components/+cluster-settings/components/cluster-metrics-setting.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+@observer
+export class ClusterMetricsSetting extends React.Component<Props> {
+  @observable hiddenMetrics = observable.set<string>();
+
+  componentDidMount() {
+    this.hiddenMetrics = observable.set<string>(this.props.cluster.preferences.hiddenMetrics ?? []);
+
+    disposeOnUnmount(this, [
+      reaction(() => this.props.cluster.preferences.hiddenMetrics, () => {
```
vshakirova, line 37: To avoid inconsistency if you remove some metrics via `show-metrics` component.

Nokel81: I think that is the correct decision

Nokel81 (dismissed):

Nokel81 (dismissed):

aleksfront (approved):
Looking good!

Nokel81 (approved):

PR #2046, opened 2021-01-28
Remove async getSubFrames and move sendToFrame loop

Nokel81:
- getSubFrames never needed to be async and was needlessly converting it to an array

- instead of relying on a bunch of promises that will get executed sometime in the future (after the current function ends). Which is a strange and unexpected use of the JS runtime behaviour. This makes it more explicit.

I have done some testing to make sure that this still works. Namely, I have made sure that adding and removing clusters still works.

### DIFF
--- a/src/common/ipc.ts
+++ b/src/common/ipc.ts
@@ -4,7 +4,7 @@
 
 import { ipcMain, ipcRenderer, webContents, remote } from "electron";
 import logger from "../main/logger";
-import { ClusterFrameInfo, clusterFrameMap } from "./cluster-frames";
+import { clusterFrameMap } from "./cluster-frames";
 
 export function handleRequest(channel: string, listener: (...args: any[]) => any) {
   ipcMain.handle(channel, listener);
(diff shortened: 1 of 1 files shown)


src/common/ipc.ts
```diff
@@ -14,38 +14,31 @@ export async function requestMain(channel: string, ...args: any[]) {
...
+    const type = view.getType();
+
+    logger.silly(`[IPC]: broadcasting "${channel}" to ${type}=${view.id}`, { args });
+    view.send(channel, ...args);
+
+    try {
+      for (const frameInfo of clusterFrameMap.values()) {
+        view.sendToFrame([frameInfo.processId, frameInfo.frameId], channel, ...args);
+      }
+    } catch (error) {
```
jim-docker, line 38: this gives up trying to send messages after the first error (I guess that was the old behaviour? Maybe not quite?). Should the `try`/`catch` be inside the `for ()` ?

Nokel81: Might be a good idea. Also need to rebase this so one moment.

jakolehm: I think this functionality was included in #1957.

Nokel81: It was, closing this PR

PR #2047, opened 2021-01-28
Basic workspace overview

### LINKED ISSUE #1864: Workspace overview

jim-docker:
Section for extension items will come in subsequent PR. Also needs to be integrated with the command palette (#1957), since it supersedes the workspace menu.

https://user-images.githubusercontent.com/40840436/106213129-06b03080-619a-11eb-90d3-05abb5075e24.mov

current look using PageLayout component for common look-and-feel:

image

closes #1864


src/main/cluster.ts
```diff
@@ -245,7 +245,7 @@ export class Cluster implements ClusterModel, ClusterState {
    * Kubernetes version
    */
   get version(): string {
-    return String(this.metadata?.version) || "";
+    return String(this.metadata?.version || "");
```
jim-docker, line 255: Before this change this getter resulted in `undefined` being output (why?) when the cluster in question was not active

src/renderer/components/+landing-page/landing-page.tsx
```diff
@@ -1,40 +1,23 @@
...
-import { observable } from "mobx";
 import { observer } from "mobx-react";
-import { clusterStore } from "../../../common/cluster-store";
-import { workspaceStore } from "../../../common/workspace-store";
-
+import { Workspace, workspaceStore } from "../../../common/workspace-store";
+import { WorkspaceOverview } from "./workspace-overview";
+import { PageLayout } from "../layout/page-layout";
 @observer
 export class LandingPage extends React.Component {
```
jim-docker, line 13: Will change the naming (easier to track differences here by not changing the name/filename yet)

src/renderer/components/+workspaces/workspace-menu.tsx
```diff
@@ -44,8 +43,8 @@ export class WorkspaceMenu extends React.Component<Props> {
         open={() => this.menuVisible = true}
         close={() => this.menuVisible = false}
       >
-        <Link className="workspaces-title" to={workspacesURL()}>
-          Workspaces
+        <Link className="workspaces-title" to={landingURL()}>
```
jim-docker: This was convenient for testing, this whole workspace menu is changing in #1957

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -331,7 +332,7 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
           {this.isReady && info}
         </div>
         {filters}
-        {search}
+        {isSearchable && searchFilters && search}
```
jim-docker, line 331: seems to make sense to not show the search widget if the list is not searchable.

src/main/cluster.ts
Nokel81: because `String(undefined) === "undefined"`

jim-docker: ah, of course. JavaScript. I guess that could've been the expected behaviour? Probably not.

Nokel81:
```js
console.log(String())
console.log(String(undefined))
console.log(String(""))
console.log(String(0))
console.log(String(false))
console.log(String(null))
```

produces:
```js
""
"undefined"
""
"0"
"false"
"null"
```

src/renderer/components/+workspaces/workspace-menu.tsx
jim-docker: #1957 is merged, this is now gone

src/renderer/components/+landing-page/landing-page.tsx
```diff
@@ -1,40 +1,24 @@
...
+import { WorkspaceOverview } from "./workspace-overview";
+import { PageLayout } from "../layout/page-layout";
 
 @observer
 export class LandingPage extends React.Component {
-  @observable showHint = true;
 
+  get workspace(): Workspace {
+    return workspaceStore.currentWorkspace;
+  }
```
Nokel81, line 18: Why is this preferable over just passing `workspaceStore.currentWorkspace` in the two places it is used?

src/renderer/components/+landing-page/workspace-cluster-menu.tsx
```diff
@@ -0,0 +1,76 @@
...
+    );
+  }
+
+
+  renderContent() {
+    const { toolbar } = this.props;
+
+    return (
+      <>
+        {
```
Nokel81: Are these `{` and `}` meaningful? I have never seen them before

src/renderer/components/+landing-page/workspace-cluster-menu.tsx
```diff
@@ -0,0 +1,76 @@
...
+          <MenuItem onClick={this.settings}>
+            <Icon material="settings" interactive={toolbar} title={`Settings`}/>
```
Nokel81:
```suggestion
            <Icon material="settings" interactive={toolbar} title="Settings"/>
```

src/renderer/components/+landing-page/workspace-cluster.store.ts
```diff
@@ -0,0 +1,64 @@
+import { WorkspaceId } from "../../../common/workspace-store";
+import { Cluster } from "../../../main/cluster";
+import { clusterStore } from "../../../common/cluster-store";
+import { ItemObject, ItemStore } from "../../item.store";
+import { autobind } from "../../utils";
+
+export class ClusterItem implements ItemObject {
+  cluster: Cluster;
```
Nokel81:
```suggestion
  constructor(public cluster: Cluster) {
    super();
  }
```

src/renderer/components/+landing-page/workspace-cluster.store.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  loadAll() {
+    return this.loadItems(() => clusterStore.getByWorkspaceId(this.workspaceId).map(cluster => {
+      const clusterItem = new ClusterItem();
+
+      clusterItem.cluster = cluster;
+
+      return clusterItem;
+    }));
+  }
```
Nokel81:
```suggestion
  loadAll() {
    return this.loadItems(
      () => (
        clusterStore
          .getByWorkspaceId(this.workspaceId)
          .map(cluster => new ClusterItem(cluster))
      )
    );
  }
```

src/renderer/components/+landing-page/workspace-cluster.store.ts
```diff
@@ -0,0 +1,64 @@
...
+  async remove(clusterItem: ClusterItem) {
+    const { cluster } = clusterItem;
+
+    if (cluster.isManaged) {
+      return;
+    }
+
+    const clusterId = cluster.id;
+
+    return super.removeItem(clusterItem, async () => {
```
Nokel81: I don't think this function needs to be async
```suggestion
    return super.removeItem(clusterItem, () => {
```

src/renderer/components/+landing-page/workspace-cluster.store.ts
```diff
@@ -0,0 +1,64 @@
...
+      if (clusterStore.activeClusterId === clusterId) {
+        clusterStore.setActive(null);
+      }
+      clusterStore.removeById(clusterId);
```
Nokel81: Isn't this work already done in `removeById`? and isn't there already a `clusterStore.remove(cluster: Cluster)` method?

src/renderer/components/+landing-page/workspace-cluster.store.ts
```diff
@@ -0,0 +1,64 @@
...
+    });
+  }
+  
+  async removeSelectedItems() {
+    if (!this.selectedItems.length) {
+      return;
+    }
```
Nokel81: This is not needed as Promise.all over `[]` results immediately.

src/renderer/components/+landing-page/workspace-overview.tsx
```diff
@@ -0,0 +1,69 @@
...
+        renderTableContents={(item: ClusterItem) => [
+          item.getName(),
+          item.cluster.contextName,
+          item.cluster.version,
+          item.cluster.online ? "online" : "offline"
+        ]}
+        onDetails={this.showCluster}
+        renderItemMenu={(clusterItem: ClusterItem) => {
+          return <WorkspaceClusterMenu clusterItem={clusterItem} workspace={workspace} workspaceClusterStore={workspaceClusterStore}/>;
+        }}
```
Nokel81:
```suggestion
        renderItemMenu={(clusterItem: ClusterItem) => (
          <WorkspaceClusterMenu clusterItem={clusterItem} workspace={workspace} workspaceClusterStore={workspaceClusterStore}/>
        )}
```

Nokel81 (commented):
Looks good so far

aleksfront:
@jakolehm @jim-docker I have some overall structure comments.

1. We shouldn't remove Landing page since it provides friendly and expected step for user before adding new cluster. Now user just goes to empty list. Also thinking about what user will see on the very first app start.
2. There should be "Add cluster" / "Remove cluster" features added (probably in next PRs, but could be in current one also). Add cluster can get advantage from our blue plus button we're using in Secrets for example
Screenshot 2021-02-04 at 10 25 16
3. "Workspace overview" item must be added to command palette, because currently there is no way to open it but create new workspace.
4. Maybe add "Workspaces" item in electron menu along with "Extentions"?
5. Minor thing: probably better to remove `background-color` from `ItemListLayout` for this page - it will look better.

src/main/cluster.ts
aleksfront: Why we ever need to wrap cluster version into `String()`? Can it have any other type?

jim-docker:
Thanks for your comments @aleksfront:

> @jakolehm @jim-docker I have some overall structure comments.
> 
> 1. We shouldn't remove Landing page since it provides friendly and expected step for user before adding new cluster. Now user just goes to empty list. Also thinking about what user will see on the very first app start.

I left the friendly elements of the landing page in my first workspace overview iteration, I could add them back here, when the workspace has no clusters.

> 2. There should be "Add cluster" / "Remove cluster" features added (probably in next PRs, but could be in current one also). Add cluster can get advantage from our blue plus button we're using in Secrets for example

I used the `ItemListLayout` add/remove buttons in the second iteration of workspace overview, but we decided to remove them in favour of the Add Cluster button in the clusters menu, and remove menu item (under the vertical ...) beside each cluster item.

> 3. "Workspace overview" item must be added to command palette, because currently there is no way to open it but create new workspace.

I've been thinking about this too. It probably should be added o the command palette but I think there should also be a way to get to it from a mouse click. @jakolehm suggested possibly an icon at the top of the clusters menu. I was thinking we could split the workspace status bar elements:
image
click the workspace icon to get the workspace command palette, click the name (i.e. "default") to go to the workspace overview. I still think it is a little odd that we force a mouse user to bring up the command palette (really meant for a keyboard user) to access workspace functions.

> 4. Maybe add "Workspaces" item in electron menu along with "Extentions"?

agreed

> 5. Minor thing: probably better to remove `background-color` from `ItemListLayout` for this page - it will look better.

will do

src/renderer/components/+landing-page/landing-page.tsx
jim-docker: Because it's one less time I have to write out `workspaceStore.currentWorkspace`? :)
I had this class doing more but refactored it out. May add some back after reviews...

src/renderer/components/+landing-page/workspace-cluster.store.ts
jim-docker: `ItemObject` is an `interface` so no `super()` to call...

src/renderer/components/+landing-page/workspace-cluster.store.ts
jim-docker: It has to return a `Promise` so specifying `async` is typical, right?

src/renderer/components/+landing-page/workspace-cluster.store.ts
jim-docker: good point about `removeById`, but there's no `clusterStore.remove(cluster: Cluster)`. There's `clusterStore.removeCluster(model: ClusterModel)`, which calls `removeById()`...

aleksfront:
> I've been thinking about this too. It probably should be added o the command palette but I think there should also be a way to get to it from a mouse click. @jakolehm suggested possibly an icon at the top of the clusters menu. I was thinking we could split the workspace status bar elements...

@jakolehm Maybe we can return workspace menu for mouse users? I'm talking about bottom menu eliminated in https://github.com/lensapp/lens/pull/1957

jakolehm:
> > I've been thinking about this too. It probably should be added o the command palette but I think there should also be a way to get to it from a mouse click. @jakolehm suggested possibly an icon at the top of the clusters menu. I was thinking we could split the workspace status bar elements...
> 
> @jakolehm Maybe we can return workspace menu for mouse users? I'm talking about bottom menu eliminated in #1957

#1957 design (and idea) came from @miskun and it follows the same pattern what VS Code has. I would not go back to workspace menu if we are going to push Command Palette further. But let's wait @miskun feedback on this.

jakolehm:
> > > I've been thinking about this too. It probably should be added o the command palette but I think there should also be a way to get to it from a mouse click. @jakolehm suggested possibly an icon at the top of the clusters menu. I was thinking we could split the workspace status bar elements...
> > 
> > 
> > @jakolehm Maybe we can return workspace menu for mouse users? I'm talking about bottom menu eliminated in #1957
> 
> #1957 design (and idea) came from @miskun and it follows the same pattern what VS Code has. I would not go back to workspace menu if we are going to push Command Palette further. But let's wait @miskun feedback on this.

Had a discussion with @miskun about this: we should just put "Workspace: Overview" to the command palette (for now).

src/renderer/components/+landing-page/workspace-cluster.store.ts
Nokel81: Ah oops, then I guess it would just be `constructor(public cluster: Cluster) { }`

src/renderer/components/+landing-page/workspace-cluster.store.ts
Nokel81: I don't think so, one can return promise without `async`. However, an `async` function always wraps its return value in a promise. 

Furthermore, you are not actually calling any other functions that return promises. Finally, async functions always add their "body" to the event loop.

src/renderer/components/+landing-page/workspace-cluster.store.ts
Nokel81: I think just calling `clusterStore.removeById(clusterId);` should be good enough.

jim-docker:
> Had a discussion with @miskun about this: we should just put "Workspace: Overview" to the command palette (for now).

I think it's a little awkward, but we can start there.

src/main/cluster.ts
jim-docker: I guess just by virtue of `this.metadata`'s generic definition?

```
export interface ClusterMetadata {
  [key: string]: string | number | boolean | object;
}
```

stefcameron:
@jim-docker This is looking like it would have potential to fill a gap I've identified through customer feedback: https://github.com/Mirantis/lens-extension-cc/issues/121

But current designs make this look like it's scoped to a single workspace. I would like to see a __full__ overview, across workspaces, of all clusters in all workspaces.

As an extension _extending_ this view, I'd want to highlight specific clusters and workspaces so that I can flag them as "stuff that was just added" when my extension navigates to this overview page after adding a bunch of clusters in a bunch of new workspaces...

jim-docker:
> @jim-docker This is looking like it would have potential to fill a gap I've identified through customer feedback: [Mirantis/lens-extension-cc#121](https://github.com/Mirantis/lens-extension-cc/issues/121)
> 
> But current designs make this look like it's scoped to a single workspace. I would like to see a **full** overview, across workspaces, of all clusters in all workspaces.
> 
> As an extension _extending_ this view, I'd want to highlight specific clusters and workspaces so that I can flag them as "stuff that was just added" when my extension navigates to this overview page after adding a bunch of clusters in a bunch of new workspaces...

This PR has specific design requirements meant to coordinate with the workspace command palette functionality. So as is it may not provide what you're looking for. Of course an extension can put basically any content it likes under the workspace overview's cluster list, such as a flat list of all clusters, and their workspace affiliation. Or buttons to take you to other workspaces... 
Otherwise I think we'd need to add a new **full** cluster overview feature.

stefcameron:
> This PR has specific design requirements meant to coordinate with the workspace command palette functionality. So as is it may not provide what you're looking for. Of course an extension can put basically any content it likes under the workspace overview's cluster list, such as a flat list of all clusters, and their workspace affiliation. Or buttons to take you to other workspaces...

🤔 I'm not sure how I would extend a page like that as an extension, at least not for the use case I'm trying to solve. I'd be interested to see in what ways you anticipate extensions would want to extend this page you're creating. I could see wanting to add columns to show more data for each listed cluster, but to just add seemingly "random" UI to other parts of the page, not sure.

Is this going to be a "master/details" pattern where you select a cluster row and you get a details panel in the lower half of the page that shows more stuff about the selected cluster? That could be interesting to extend as an extension.

Will the list allow CRUD operations on clusters, e.g. select many to delete in one step instead of "right-click + delete + confirm" on each individual cluster from the sidebar? Rename clusters? Move them to other workspaces?

> Otherwise I think we'd need to add a new **full** cluster overview feature.

Yes, please! 😁

src/renderer/components/+landing-page/workspace-cluster-menu.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  @autobind()
+  remove() {
+    const { clusterItem, workspaceClusterStore } = this.props;
+    
+    return workspaceClusterStore.remove(clusterItem);
+  }
+
+  @autobind()
+  settings() {
```
Nokel81: This sounds like a getter, maybe rename to `gotoSettings` or `navigateToSettings` or `showSettings`?

src/renderer/components/+landing-page/workspace-cluster.store.ts
```diff
@@ -0,0 +1,55 @@
...
+  async remove(clusterItem: ClusterItem) {
+    const { cluster } = clusterItem;
+
+    if (cluster.isManaged) {
+      return;
+    }
+
+    const clusterId = cluster.id;
+
+    return super.removeItem(clusterItem, () => clusterStore.removeById(clusterId));
```
Nokel81:
```suggestion
    const { cluster: { isManaged, id: clusterId } = clusterItem;

    if (!isManaged) {
      return super.removeItem(clusterItem, () => clusterStore.removeById(clusterId));
    }
```

src/renderer/components/+landing-page/workspace-cluster.store.ts
```diff
@@ -0,0 +1,55 @@
...
+export class ClusterItem implements ItemObject {
+  constructor(public cluster: Cluster) {}
+  
+  getName() {
+    return this.cluster.name;
+  }
+
+  getId() {
+    return this.cluster.id;
+  }
```
Nokel81: Having these also as getters would allow for some other code to be good.
```suggestion
  get name() {
    return this.cluster.name;
  }

  getName() {
    return this.name;
  }
  
  get id() {
    return this.cluster.id;
  }
  
  get clusterId() {
    return this.cluster.id;
  }

  getId() {
    return this.id;
  }
```

src/renderer/components/+landing-page/workspace-overview.tsx
```diff
@@ -0,0 +1,69 @@
...
+}
+
+@observer
+export class WorkspaceOverview extends Component<Props> {
+
+  showCluster = (clusterItem: ClusterItem) => {
+    const clusterId = clusterItem.getId();
+
+    navigate(clusterViewURL({ params: { clusterId } }));
+  };
```
Nokel81:
```suggestion
  showCluster = ({ clusterId }: ClusterItem) => {
    navigate(clusterViewURL({ params: { clusterId } }));
  };
```

src/renderer/components/+landing-page/workspace-overview.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  render() {
+    const { workspace } = this.props;
+    const workspaceClusterStore = new WorkspaceClusterStore(workspace.id);
+    
+    workspaceClusterStore.loadAll();
```
Nokel81: The creation probably should be done in a field initialize. And the `loadAll` in `componentDidMount`?

jim-docker: I had it that way at one point but it wasn't enough.   Need to update the store each time `render()` is called.

Nokel81: That sounds very strange. Like the loading should be reactive to something else.

src/main/cluster.ts
Nokel81:
```suggestion
    return String(this.metadata?.version ?? "");
```

So that if for whatever reason `version` is `false` or `0` that is passed through.

Nokel81 (approved):
Maybe a followup PR could be adding a right click menu similar to the vertical dots menu on the overview. But other than that looks very good.

jakolehm (approved):

PR #2053, opened 2021-02-01
Add deb & rpm packages

### LINKED ISSUE #2052: Add deb & rpm packages

Packaging from `deb` & `rpm`.

**Why is this needed**:

Works better with the upcoming #1949 feature.

jakolehm:
Fixes #2052

### DIFF
--- a/integration/helpers/utils.ts
+++ b/integration/helpers/utils.ts
@@ -4,7 +4,7 @@ import { exec } from "child_process";
 
 const AppPaths: Partial<Record<NodeJS.Platform, string>> = {
   "win32": "./dist/win-unpacked/Lens.exe",
-  "linux": "./dist/linux-unpacked/kontena-lens",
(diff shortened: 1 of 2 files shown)


integration/helpers/utils.ts
```diff
@@ -4,7 +4,7 @@ import { exec } from "child_process";
 
 const AppPaths: Partial<Record<NodeJS.Platform, string>> = {
   "win32": "./dist/win-unpacked/Lens.exe",
-  "linux": "./dist/linux-unpacked/kontena-lens",
+  "linux": "./dist/linux-unpacked/lens",
```
nevalla, line 7: does this have any side effect?

jakolehm: Changing the executable name might cause a problem with `snap` but I cannot be sure before we do a pre-release.

Nokel81 (approved):

nevalla (approved):

PR #2054, opened 2021-02-01
Fix: charts stripes alignment

### LINKED ISSUE #2008: Chart stripes are misaligned from bars
For some reason, "zebra stripes" are not aligned strictly on bar corners. They should cover some amount of bars positioned strictly on their edges.

image

image

aleksfront:
* Updating chart stripes width and position on every dataset update
* Hide stripes if less than 10 minutes presented (one stripe covers 10 min period normally)

fixed stripes
fixed stripes 15 minutes

Fixes #2008 

### DIFF
--- a/src/renderer/components/chart/bar-chart.tsx
+++ b/src/renderer/components/chart/bar-chart.tsx
@@ -136,7 +136,8 @@ export class BarChart extends React.Component<Props> {
       },
       plugins: {
         ZebraStripes: {
-          stripeColor: chartStripesColor
+          stripeColor: chartStripesColor,
+          interval: chartData.datasets.length ? chartData.datasets[0].data.length : 61
         }
       }
     };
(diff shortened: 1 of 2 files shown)


src/renderer/components/chart/bar-chart.tsx
```diff
@@ -136,7 +136,8 @@ export class BarChart extends React.Component<Props> {
       },
       plugins: {
         ZebraStripes: {
-          stripeColor: chartStripesColor
+          stripeColor: chartStripesColor,
+          interval: chartData.datasets.length ? chartData.datasets[0].data.length : 61
```
Nokel81: I see that `61` is used in other places too. However it seems like a strange number. Maybe add a comment about why it was chosen?

aleksfront: Yes, you're right. Actually, we don't need any predefined numbers in here since `datasets[0].data.length` always contains needed value. Fixed.

jakolehm (approved):

Nachasic (approved):
Aside from the issue with magic number "61" pointed out above, LGTM

PR #2063, opened 2021-02-02
KubeObjectStore & KubeWatchApi fixes and optimizations

ixrock:
_Highlights:_

- Detached `NamespaceStore` from `KubeWatchApi`
- Loading all namespaces by default with `KubeObjectStore.loadAll()`
- Added `KubeObjectStore.loadSelectedNamespaces()`
- Watch-api requests optimization (#2066)

rebased from #2033

### DIFF
--- a/src/main/cluster.ts
+++ b/src/main/cluster.ts
@@ -178,7 +178,7 @@ export class Cluster implements ClusterModel, ClusterState {
    */
   @observable isAdmin = false;
   /**
-   * Does watch-api could be used for all resources, e.g. "/api/v1/services?watch=1"
+   * Global watch-api accessibility , e.g. "/api/v1/services?watch=1"
    *
    * @observable
    */
--- a/src/main/cluster.ts
+++ b/src/main/cluster.ts
@@ -588,7 +588,7 @@ export class Cluster implements ClusterModel, ClusterState {
   /**
    * @internal
    */
-  private async canUseWatchApi(customizeResource: V1ResourceAttributes = {}): Promise<boolean> {
+  async canUseWatchApi(customizeResource: V1ResourceAttributes = {}): Promise<boolean> {
     return this.canI({
       verb: "watch",
       resource: "*",
--- a/src/main/cluster.ts
+++ b/src/main/cluster.ts
@@ -48,6 +48,7 @@ export interface ClusterState {
   isAdmin: boolean;
   allowedNamespaces: string[]
   allowedResources: string[]
+  isGlobalWatchEnabled: boolean;
 }
 
 /**
@@ -90,7 +91,6 @@ export class Cluster implements ClusterModel, ClusterState {
    * @observable
    */
   @observable initializing = false;
-
 
   /**
    * Is cluster object initialized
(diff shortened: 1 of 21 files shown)


jakolehm (dismissed):

ixrock:
@lensapp/lens-maintainers PTAL

src/main/cluster.ts
```diff
@@ -433,6 +435,18 @@ export class Cluster implements ClusterModel, ClusterState {
...
   }
 
+  /**
+   * @internal
+   */
+  private async refreshAccessibility(): Promise<void> {
+    this.isAdmin = await this.isClusterAdmin();
+    this.isGlobalWatchEnabled = await this.canUseWatchApi({ resource: "*" });
+
+    await this.refreshAllowedResources();
```
Nokel81, line 445: If these fail, wouldn't that imply that the cluster is no longer ready?

jakolehm (approved):

src/renderer/components/+apps-releases/release.store.ts
```diff
@@ -82,23 +86,23 @@ export class ReleaseStore extends ItemStore<HelmRelease> {
   async create(payload: IReleaseCreatePayload) {
     const response = await helmReleasesApi.create(payload);
 
-    if (this.isLoaded) this.loadAll();
+    if (this.isLoaded) this.loadSelectedNamespaces();
```
Nokel81, line 89: I suggest adding a comment that this is supposed to be a refreshing call

src/renderer/kube-object.store.ts
```diff
@@ -106,17 +106,18 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
     this.isLoading = true;
 
     try {
-      if (!contextNamespaces) {
+      if (!namespaces.length) {
         const { namespaceStore } = await import("./components/+namespaces/namespace.store");
 
-        contextNamespaces = namespaceStore.getContextNamespaces();
+        // load all available namespaces by default
+        namespaces.push(...namespaceStore.allowedNamespaces);
```
Nokel81, line 117: First of all this doesn't need to be a `push(...)`, it could be an assign.

Second of all, why is `NamespaceStore` getting a "get-er" that merely delegates to `getHostedCluster()`? I don't think we should have that (it doesn't make sense for `NamespaceStore` to have that function) and it adds this circular dependency.

Lastly, I don't think this is how we should be "hiding getHostedCluster".

ixrock: Yes, this will be refactored similarly as in `KubeWatchApi`:

```
  @observable.ref private cluster: Cluster;
  @observable.ref private namespaces: string[] = [];
```

Anyway it's not critical for now and could be improved in follow-up PR(s).

src/renderer/components/+apps-releases/release.store.ts
ixrock: Probably should be refactored to separated method

src/renderer/kube-object.store.ts
```diff
@@ -131,6 +132,12 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
     }
   }
 
+  async loadSelectedNamespaces(): Promise<void> {
```
jim-docker, line 135: The name seems a bit misleading, it's really `loadFromSelectedNamespaces` or `loadForSelectedNamespaces`. But those names are getting too long. Maybe not worth changing...

jim-docker (approved):

src/renderer/kube-object.store.ts
ixrock: how about `loadContextNamespaces()`?

PR #2065, opened 2021-02-02
Change Cluster Settings button to be consistent with cluster icon menu

### LINKED ISSUE #1895: Replace top-right "settings" menu icon with three-dot menu

In master branch (4.1) we have settings menu icon in the top-right corner. We should replace this with three-dot menu which reveals same items as cluster icon context menu (settings, disconnect, remove).

**Why is this needed**:

For consistency.

Nachasic:
The top-right button in the cluster frame now has vertical-dots icon, clicking on it now produces the same menu, as right-clicking on the cluster icon. Popup at this button now says "Cluster actions". Closes #1895 
Screen Shot 2021-02-02 at 6 44 27 PM

### DIFF
--- a/src/common/cluster-store.ts
+++ b/src/common/cluster-store.ts
@@ -225,6 +225,12 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
     workspaceStore.setLastActiveClusterId(clusterId);
   }
 
+  deactivate(id: ClusterId) {
+    if (this.isActive(id)) {
+      this.setActive(null);
+    }
+  }
+
   @action
   swapIconOrders(workspace: WorkspaceId, from: number, to: number) {
     const clusters = this.getByWorkspaceId(workspace);
--- a/src/renderer/components/cluster-manager/cluster-actions.tsx
+++ b/src/renderer/components/cluster-manager/cluster-actions.tsx
@@ -21,7 +21,7 @@ const deactivateCluster = (clusterId: string) => {
 /**
  * Creates handlers for high-level actions
  * that could be performed on an individual cluster
- * @param cluster Cluster>
+ * @param cluster Cluster
  */
 export const ClusterActions = (cluster: Cluster) => ({
   "SHOW_SETTINGS": () => navigate(clusterSettingsURL({
--- a/src/renderer/components/cluster-manager/cluster-actions.tsx
+++ b/src/renderer/components/cluster-manager/cluster-actions.tsx
@@ -11,13 +11,6 @@ import { Cluster } from "../../../main/cluster";
 const navigate = (route: string) =>
   broadcastMessage("renderer:navigate", route);
 
-const deactivateCluster = (clusterId: string) => {
-  if (clusterStore.isActive(clusterId)) {
-    clusterStore.setActive(null);
-  }
-  navigate(landingURL());
-};
-
 /**
  * Creates handlers for high-level actions
  * that could be performed on an individual cluster
@@ -28,7 +21,8 @@ export const ClusterActions = (cluster: Cluster) => ({
     params: { clusterId: cluster.id }
   })),
   disconnect: async () => {
-    deactivateCluster(cluster.id);
+    clusterStore.deactivate(cluster.id);
+    navigate(landingURL());
     await requestMain(clusterDisconnectHandler, cluster.id);
   },
   remove: () => ConfirmDialog.open({
@@ -38,8 +32,9 @@ export const ClusterActions = (cluster: Cluster) => ({
       label: "Remove"
     },
     ok: () => {
-      deactivateCluster(cluster.id);
+      clusterStore.deactivate(cluster.id);
       clusterStore.removeById(cluster.id);
+      navigate(landingURL());
     },
     message: <p>Are you sure want to remove cluster <b title={cluster.id}>{cluster.contextName}</b>?</p>,
   })
--- a/src/renderer/components/cluster-manager/cluster-actions.tsx
+++ b/src/renderer/components/cluster-manager/cluster-actions.tsx
@@ -13,9 +13,9 @@ const navigate = (route: string) =>
 
 const deactivateCluster = (clusterId: string) => {
   if (clusterStore.isActive(clusterId)) {
-    navigate(landingURL());
     clusterStore.setActive(null);
   }
+  navigate(landingURL());
 };
 
 /**
--- a/src/renderer/components/cluster-manager/cluster-actions.tsx
+++ b/src/renderer/components/cluster-manager/cluster-actions.tsx
@@ -1,0 +1,46 @@
+import React from "react";
+import { clusterSettingsURL } from "../+cluster-settings";
+import { landingURL } from "../+landing-page";
+
+import { clusterStore } from "../../../common/cluster-store";
+import { broadcastMessage, requestMain } from "../../../common/ipc";
+import { clusterDisconnectHandler } from "../../../common/cluster-ipc";
+import { ConfirmDialog } from "../confirm-dialog";
+import { Cluster } from "../../../main/cluster";
+
+const navigate = (route: string) =>
+  broadcastMessage("renderer:navigate", route);
+
+const deactivateCluster = (clusterId: string) => {
+  if (clusterStore.isActive(clusterId)) {
+    navigate(landingURL());
+    clusterStore.setActive(null);
+  }
+};
+
+/**
+ * Creates hanndlers for high-level actions
+ * that could be performed on an individual cluster
+ * @param cluster Cluster>
+ */
+export const ClusterActions = (cluster: Cluster) => ({
+  "SHOW_SETTINGS": () => navigate(clusterSettingsURL({
+    params: { clusterId: cluster.id }
+  })),
+  "DISCONNECT": async () => {
+    deactivateCluster(cluster.id);
+    await requestMain(clusterDisconnectHandler, cluster.id);
+  },
+  "REMOVE": () => ConfirmDialog.open({
+    okButtonProps: {
+      primary: false,
+      accent: true,
+      label: "Remove"
+    },
+    ok: () => {
+      deactivateCluster(cluster.id);
+      clusterStore.removeById(cluster.id);
+    },
+    message: <p>Are you sure want to remove cluster <b title={cluster.id}>{cluster.contextName}</b>?</p>,
+  })
+});
--- a/src/renderer/components/cluster-manager/cluster-actions.tsx
+++ b/src/renderer/components/cluster-manager/cluster-actions.tsx
@@ -24,14 +24,14 @@ const deactivateCluster = (clusterId: string) => {
  * @param cluster Cluster
  */
 export const ClusterActions = (cluster: Cluster) => ({
-  "SHOW_SETTINGS": () => navigate(clusterSettingsURL({
+  showSettings: () => navigate(clusterSettingsURL({
     params: { clusterId: cluster.id }
   })),
-  "DISCONNECT": async () => {
+  disconnect: async () => {
     deactivateCluster(cluster.id);
     await requestMain(clusterDisconnectHandler, cluster.id);
   },
-  "REMOVE": () => ConfirmDialog.open({
+  remove: () => ConfirmDialog.open({
     okButtonProps: {
       primary: false,
       accent: true,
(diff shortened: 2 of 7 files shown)


aleksfront (changes requested):
I'm not sure should we use native menu or our inner Menu component here. For example, we're using internal `<Menu/>` (src/renderer/components/menu/menu.tsx) for all stuff around the dashboard. For example for terminal context menu

Screenshot 2021-02-03 at 15 55 12

Also, it seems that code duplicates the one from `src/renderer/components/cluster-manager/clusters-menu.tsx`. Maybe there is a way to move at least click event handlers to one shared place?

In addition, please check `src/renderer/components/layout/__test__/main-layout-header.test.tsx` which is checking previous gear icon. It's a good place to test the new menu I guess.

Nachasic:
After initial implementation I've found a bug with cluster menu.

**Prerequisites:**
Lens app is connected to cluster.

**Steps to reproduce:**
1. Open cluster actions menu by clicking on a button in top-right corner of the UI. Cluster actions menu has 3 actions: "Settings", "Disconnect" and "Remove";
1. Click "Disconnect";
1. After the lens is disconnected, navigate back to this cluster using clusters menu and wait for Lens to re-connect;
1. Open cluster actions menu again.

**Expected behavior:** Cluster actions menu has 3 actions: "Settings", "Disconnect" and "Remove".
**Actual behavior:** Cluster actions menu has only one action: "Settings".

I've marked this PR as draft, until I manage to fix this issue.

Nachasic:
> After initial implementation I've found a bug with cluster menu.
> 
> **Prerequisites:**
> Lens app is connected to cluster.
> 
> **Steps to reproduce:**
> 
> 1. Open cluster actions menu by clicking on a button in top-right corner of the UI. Cluster actions menu has 3 actions: "Settings", "Disconnect" and "Remove";
> 2. Click "Disconnect";
> 3. After the lens is disconnected, navigate back to this cluster using clusters menu and wait for Lens to re-connect;
> 4. Open cluster actions menu again.
> 
> **Expected behavior:** Cluster actions menu has 3 actions: "Settings", "Disconnect" and "Remove".
> **Actual behavior:** Cluster actions menu has only one action: "Settings".
> 
> I've marked this PR as draft, until I manage to fix this issue.

Decided not to fix this explicitly, since the menu in question is only accessible when the cluster is online and connected to the app - removed unnecessary checks instead.

aleksfront:
Thanks for the changes.

However, there're bit more issues appeared.
1. Maybe we don't need tooltip `Cluster actions`. I guess the dots menu is self-explanatory.
2. Menu not always clickable. I feel there's something about triggering it right. Maybe it's better to use `<MenuActions />` component which already has dots icon in it (src/renderer/components/menu/menu-actions.tsx)
not clickable menu
3. Cluster disconnection not always happens right. From the second time it just reconnects.
reconnecting cluster

src/renderer/components/cluster-manager/cluster-actions.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { broadcastMessage, requestMain } from "../../../common/ipc";
+import { clusterDisconnectHandler } from "../../../common/cluster-ipc";
+import { ConfirmDialog } from "../confirm-dialog";
+import { Cluster } from "../../../main/cluster";
+
+const navigate = (route: string) =>
+  broadcastMessage("renderer:navigate", route);
+
+const deactivateCluster = (clusterId: string) => {
+  if (clusterStore.isActive(clusterId)) {
```
aleksfront: This deactivation can be moved to `ClusterStore`.

aleksfront (dismissed):
LGTM

src/renderer/components/cluster-manager/cluster-actions.tsx
Nokel81: Isn't this code already in `ClusterStore`?

src/renderer/components/cluster-manager/cluster-actions.tsx
```diff
@@ -0,0 +1,46 @@
...
+  navigate(landingURL());
+};
+
+/**
+ * Creates handlers for high-level actions
+ * that could be performed on an individual cluster
+ * @param cluster Cluster
+ */
+export const ClusterActions = (cluster: Cluster) => ({
+  "SHOW_SETTINGS": () => navigate(clusterSettingsURL({
```
Nokel81: Why are this all uppercase? If you want them to be private then you should define some non-exported symbols above and use those. 

Furthermore, there is no need for them to be quoted as they are valid identifiers.

src/renderer/components/cluster-manager/cluster-actions.tsx
```diff
@@ -0,0 +1,46 @@
...
+    okButtonProps: {
+      primary: false,
+      accent: true,
+      label: "Remove"
+    },
+    ok: () => {
+      deactivateCluster(cluster.id);
+      clusterStore.removeById(cluster.id);
+    },
+    message: <p>Are you sure want to remove cluster <b title={cluster.id}>{cluster.contextName}</b>?</p>,
```
Nokel81: If you want to have a hover, we have a better hover mechanic already in this repo, I suggest you use it (namely `Tooltip`

jim-docker (commented):
I'm seeing a strange behaviour, sometimes I have to click multiple times to close the settings page. From 0 (settings page closes on its own, right after opening) to 4 clicks required. Thought it was related to how many clusters were connected but not so sure after continued testing...

https://user-images.githubusercontent.com/40840436/108233991-55eede80-7112-11eb-864f-7c224a16e8e1.mov

Nokel81:
Also, given that the cluster menu over the icon is currently an OS menu, why shouldn't this other one be as well? @aleksfront

aleksfront:
> Also, given that the cluster menu over the icon is currently an OS menu, why shouldn't this other one be as well? @aleksfront 

We use common menu component in different places within dashboard. See context menu for workloads, for example. It looks reasonable to use same `<MenuActions/>` for this menu also.

Nokel81:
@aleksfront About the menu being an HTML based one vs a system one. For menus that come up from RMB or LMB clicks it seems like it would give a better experience to use the system one since it can render outside the viewport and will look more consistent with the rest of the operating system.

src/renderer/components/cluster-manager/cluster-actions.tsx
Nokel81: I guess it isn't in `ClusterStore` but I agree with @aleksfront that it should be.

src/renderer/components/cluster-manager/cluster-actions.tsx
Nokel81: bump

src/renderer/components/cluster-manager/cluster-actions.tsx
Nokel81: Still relevant though `Badge` might be a better choice

Nokel81 (commented):
Still some actions not acted upon or responded too.

jim-docker:
still strange side effects, multiple clicks sometimes required to close settings page. Also encountered this after disconnecting a cluster (after clicking on other cluster icons):

image

src/renderer/components/cluster-manager/cluster-actions.tsx
Nachasic: It wasn't, moved it over there

src/renderer/components/cluster-manager/cluster-actions.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Creates handlers for high-level actions
+ * that could be performed on an individual cluster
+ * @param cluster Cluster
+ */
+export const ClusterActions = (cluster: Cluster) => ({
+  showSettings: () => navigate(clusterSettingsURL({
+    params: { clusterId: cluster.id }
+  })),
```
Nokel81, line 24: This doesn't seem to work correctly. (This also didn't work before). But calling this code from within the iframe the `clusterSettingsURL` doesn't exist so we are redirected to `/` or the landing page.

It would be great if once the user exited out of the settings the view would go back to what they were doing before.

Nokel81 (changes requested):

src/renderer/components/cluster-manager/cluster-actions.tsx
Nachasic: This seems to have been fixed in #2330

src/renderer/components/cluster-manager/cluster-actions.tsx
Nachasic: I've experimented with `Badge` component - doesn't look to nice without custom styles, going on with `Tooltip`

src/common/cluster-store.ts
```diff
@@ -225,6 +225,12 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
     workspaceStore.setLastActiveClusterId(clusterId);
   }
 
+  deactivate(id: ClusterId) {
+    if (this.isActive(id)) {
+      this.setActive(null);
+    }
+  }
```
Nokel81, line 232: Note that once #2355 is merged, this will have to change. But we will resolve that race condition in the future.

src/renderer/components/layout/main-layout-header.tsx
```diff
@@ -1,33 +1,34 @@
...
+  const renderMenu = () => <MenuActions autoCloseOnSelect className="ClusterActionsMenu">
+    <MenuItem onClick={actions.showSettings}>
+      <span>Settings</span>
+    </MenuItem>
+    <MenuItem onClick={actions.disconnect}>
+      <span>Disconnect</span>
+    </MenuItem>
+    { !cluster.isManaged && <MenuItem onClick={actions.remove}>
+      <span>Remove</span>
+    </MenuItem> }
```
Nokel81: nit: I find the following slightly more readable.
```suggestion
    {
       !cluster.isManaged && (
         <MenuItem onClick={actions.remove}>
	       <span>Remove</span>
	     </MenuItem> 
       )
    }
```

src/renderer/components/layout/main-layout-header.tsx
```diff
@@ -1,33 +1,34 @@
...
+import { MenuActions, MenuItem } from "../menu";
 
 interface Props {
   cluster: Cluster
   className?: string
 }
 
 export const MainLayoutHeader = observer(({ cluster, className }: Props) => {
+  const actions = ClusterActions(cluster);
+  const renderMenu = () => <MenuActions autoCloseOnSelect className="ClusterActionsMenu">
```
Nokel81: nit: I find starting the root tag on a new line, slightly easier to read:
```suggestion
  const renderMenu = () => (
    <MenuActions autoCloseOnSelect className="ClusterActionsMenu">
```

Nokel81 (dismissed):

Nokel81:
I have noticed that the menu in the workspace overview is not the same as the one here. Would you be able to make that the same too? Maybe add it as a right click menu on the cluster bar as well?

Nachasic:
> I have noticed that the menu in the workspace overview is not the same as the one here. Would you be able to make that the same too? Maybe add it as a right click menu on the cluster bar as well?

I thought about this too. Don't you think it should be in its own PR though?

Nokel81:
A separate PR sounds good.

src/common/cluster-store.ts
Nachasic: I'll get familiar with the changes in the meantime.

Nokel81 (approved):

aleksfront (approved):
LGTM!

Nokel81:
I'll take care of merging this when I do the rc.2 work today.

src/renderer/components/cluster-manager/cluster-actions.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+import { clusterStore } from "../../../common/cluster-store";
+import { broadcastMessage, requestMain } from "../../../common/ipc";
+import { clusterDisconnectHandler } from "../../../common/cluster-ipc";
+import { ConfirmDialog } from "../confirm-dialog";
+import { Cluster } from "../../../main/cluster";
+import { Tooltip } from "../../components//tooltip";
+
+const navigate = (route: string) =>
+  broadcastMessage("renderer:navigate", route);
```
ixrock, line 14: IPC-event name could be re-used from here after merge https://github.com/lensapp/lens/pull/2363/files#diff-1e3a2e0b62c1d7d8887b79611121f74db6b5d18f7de8b9627dffa107a939a9caR10

Nokel81: Good catch, may need to do some follow up tweeks before rc.2

Nokel81:
@Nachasic I think I will merge this (because it is already approved) and then we can fix up #2363

PR #2066, opened 2021-02-02
Watch-api requests optimization

ixrock:
Pods (200+ namespaces selected):

**Before:**
Screenshot 2021-02-02 at 18 46 59

**After:**
Screenshot 2021-02-02 at 18 47 55

requires #2063

### DIFF
--- a/src/main/cluster.ts
+++ b/src/main/cluster.ts
@@ -48,7 +48,7 @@ export interface ClusterState {
   isAdmin: boolean;
   allowedNamespaces: string[]
   allowedResources: string[]
-  watchApiAllEnabled: boolean;
+  isGlobalWatchEnabled: boolean;
 }
 
 /**
@@ -182,7 +182,7 @@ export class Cluster implements ClusterModel, ClusterState {
    *
    * @observable
    */
-  @observable watchApiAllEnabled = false;
+  @observable isGlobalWatchEnabled = false;
   /**
    * Preferences
    *
@@ -440,7 +440,7 @@ export class Cluster implements ClusterModel, ClusterState {
    */
   async refreshAccessibility(): Promise<void> {
     this.isAdmin = await this.isClusterAdmin();
-    this.watchApiAllEnabled = await this.canUseWatchApi({ resource: "*" });
+    this.isGlobalWatchEnabled = await this.canUseWatchApi({ resource: "*" });
 
     await this.refreshAllowedResources();
 
@@ -629,7 +629,7 @@ export class Cluster implements ClusterModel, ClusterState {
       isAdmin: this.isAdmin,
       allowedNamespaces: this.allowedNamespaces,
       allowedResources: this.allowedResources,
-      watchApiAllEnabled: this.watchApiAllEnabled,
+      isGlobalWatchEnabled: this.isGlobalWatchEnabled,
     };
 
     return toJS(state, {
--- a/src/main/cluster.ts
+++ b/src/main/cluster.ts
@@ -48,6 +48,7 @@ export interface ClusterState {
   isAdmin: boolean;
   allowedNamespaces: string[]
   allowedResources: string[]
+  watchApiAllEnabled: boolean;
 }
 
 /**
@@ -90,7 +91,6 @@ export class Cluster implements ClusterModel, ClusterState {
    * @observable
    */
   @observable initializing = false;
-
 
   /**
    * Is cluster object initialized
@@ -177,6 +177,12 @@ export class Cluster implements ClusterModel, ClusterState {
    * @observable
    */
   @observable isAdmin = false;
+  /**
+   * Does watch-api could be used for all resources, e.g. "/api/v1/services?watch=1"
+   *
+   * @observable
+   */
+  @observable watchApiAllEnabled = false;
   /**
    * Preferences
    *
@@ -353,9 +359,7 @@ export class Cluster implements ClusterModel, ClusterState {
     await this.refreshConnectionStatus();
 
     if (this.accessible) {
-      await this.refreshAllowedResources();
-      this.isAdmin = await this.isClusterAdmin();
-      this.ready = true;
+      await this.refreshAccessibility();
       this.ensureKubectl();
     }
     this.activated = true;
@@ -410,13 +414,11 @@ export class Cluster implements ClusterModel, ClusterState {
     await this.refreshConnectionStatus();
 
     if (this.accessible) {
-      this.isAdmin = await this.isClusterAdmin();
-      await this.refreshAllowedResources();
+      await this.refreshAccessibility();
 
       if (opts.refreshMetadata) {
         this.refreshMetadata();
       }
-      this.ready = true;
     }
     this.pushState();
   }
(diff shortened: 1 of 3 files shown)


Nokel81:
Why is increasing the number of Apis an "optimization"?

ixrock:
> Why is increasing the number of Apis an "optimization"?

lol. wrong images order :D thanks for noticing.

src/main/cluster.ts
```diff
@@ -177,6 +177,12 @@ export class Cluster implements ClusterModel, ClusterState {
    * @observable
    */
   @observable isAdmin = false;
+  /**
+   * Does watch-api could be used for all resources, e.g. "/api/v1/services?watch=1"
+   *
+   * @observable
+   */
+  @observable watchApiAllEnabled = false;
```
jakolehm: Maybe `isGlobalWatchEnabled` ?

src/renderer/api/kube-watch-api.ts
```diff
@@ -51,6 +52,12 @@ export class KubeWatchApi {
   @observable subscribers = observable.map<KubeApi, number>();
   @observable isConnected = false;
 
+  @computed get watchAllEnabled(): boolean {
+    const { isAdmin, watchApiAllEnabled } = this?.cluster;
```
jakolehm: I think we don't need `isAdmin` check here.

src/main/cluster.ts
ixrock: Yes, sounds a bit better.

src/renderer/api/kube-watch-api.ts
ixrock: Fixed.

src/main/cluster.ts
```diff
@@ -433,6 +435,18 @@ export class Cluster implements ClusterModel, ClusterState {
     this.metadata = Object.assign(existingMetadata, metadata);
   }
 
+  /**
+   * @internal
+   */
+  async refreshAccessibility(): Promise<void> {
```
jakolehm, line 441: This probably could be `private` because it should not be called from outside.

src/main/cluster.ts
```diff
@@ -571,6 +585,17 @@ export class Cluster implements ClusterModel, ClusterState {
     });
   }
 
+  /**
+   * @internal
+   */
+  async canUseWatchApi(customizeResource: V1ResourceAttributes = {}): Promise<boolean> {
```
jakolehm, line 591: This could be also `private`.

src/main/cluster.ts
```diff
@@ -410,13 +414,11 @@ export class Cluster implements ClusterModel, ClusterState {
...
 
     if (this.accessible) {
-      this.isAdmin = await this.isClusterAdmin();
-      await this.refreshAllowedResources();
+      await this.refreshAccessibility();
 
       if (opts.refreshMetadata) {
         this.refreshMetadata();
       }
-      this.ready = true;
```
jakolehm, line 419: This slightly changes timing but perhaps it's better now.

jakolehm (approved):

src/main/cluster.ts
ixrock: It should not since `this.refreshMetadata();` was not awaited.

src/main/cluster.ts
```diff
@@ -177,6 +177,12 @@ export class Cluster implements ClusterModel, ClusterState {
    * @observable
    */
   @observable isAdmin = false;
+  /**
+   * Does watch-api could be used for all resources, e.g. "/api/v1/services?watch=1"
```
Nokel81, line 181: This comment makes no grammatical sense.

Do you mean "Is watch-api able to be used ..."

src/main/cluster.ts
Nokel81: Agreed

src/renderer/api/kube-watch-api.ts
```diff
@@ -127,45 +128,66 @@ export class KubeWatchApi {
...
+    for (const store of stores) {
+      preloading.push(limitRequests(async () => {
+        if (store.isLoaded && loadOnce) return; // skip
+
+        return store.loadAll(this.namespaces);
+      }));
+    }
+
+    return {
+      loading: Promise.all(preloading),
```
Nokel81, line 144: This should probably be `Promise.allSettled`

src/main/cluster.ts
ixrock: Fixed in #2063

src/main/cluster.ts
ixrock: Fixed in #2063

src/renderer/api/kube-watch-api.ts
```diff
@@ -127,45 +128,66 @@ export class KubeWatchApi {
     };
   }
 
-  subscribeStores(stores: KubeObjectStore[], options: IKubeWatchSubscribeStoreOptions = {}): () => void {
-    const { preload = true, waitUntilLoaded = true, cacheLoading = false } = options;
+  preloadStores(stores: KubeObjectStore[], { loadOnce = false } = {}) {
     const limitRequests = plimit(1); // load stores one by one to allow quick skipping when fast clicking btw pages
```
Nokel81, line 132: If you want fast skipping then wouldn't be better to support passing in an `AbortController`?

Nokel81 (commented):
Why was this PR merged @ixrock there was only one approval

jakolehm:
> Why was this PR merged @ixrock there was only one approval

Target branch was not master -> only one approve is needed.

Nokel81:
Ah I see, thanks for the info. I'll just make my comments on the target PR then.

src/renderer/api/kube-watch-api.ts
ixrock: No, it's not related to watch-api request itself... (it's preloading stores)

src/main/cluster.ts
ixrock: fixed in target PR

src/renderer/api/kube-watch-api.ts
ixrock: Fixed

src/renderer/api/kube-watch-api.ts
Nokel81: Yes I understand that this function is for preloading stores. But I don't understand what that has to do with this call? The comment implies that we only do one request at a time to allow for quick skipping. Which I think implies stopping the previous preload and starting another one.

If the above is true then that is a perfect use case for something like `AbortController`

ixrock: - `AbortController` aborts only request `POST /api/watch` (watch-route.ts)
- `plimit` aborts other k8s requests, e.g `GET /api/v1/pods`

Do you suggest to provide extra `AbortController` to k8s requests to abort them all at once?

Nokel81: Yes I do since `plimit` doesn't actually abort the requests.

ixrock: Yes, but since requests are made one by one not so critical (only last active load will be aborted).
Could be improved anyway in follow-up PR.

Nokel81: I meant in conjunction with raising the `plimit` value

Issue #1040, opened 2020-10-07
Switch fast to all-namespaces


yagilm:
**What would you like to be added**:
Easily go back to "all-namespaces". (maybe "uncheck all namespaces", or "all namespaces")

**Why is this needed**:
When selecting a/some namespace(s) to perform a task or check someting you need to uncheck them all to go to the default (all-namespaces). When working with many namespaces you would need to scroll endlessly and uncheck everything.

**Environment you are Lens application on:**
- Kubernetes distribution: multiple
- Desktop OS: Linux, Gnome

ixrock:
@yagilm you can do it easily, just click to `Filtered` + `Reset`:

![image](https://user-images.githubusercontent.com/6377066/95320426-1f6ea080-08a2-11eb-910b-9577bcb7cc0e.png)

yagilm:
Oh my.. totaly missed that.
thanks @ixrock and thanks @lens_team!

adammw:
This wasn't clear for me as well especially when the "currently applied filters" is hidden by default, any reason not to have "All namespaces" in the dropdown menu?

PR #2068, opened 2021-02-02
Allow to quick select/deselect all namespaces in `NamespaceSelect`

ixrock:
image

> This wasn't clear for me as well especially when the "currently applied filters" is hidden by default, any reason not to have "All namespaces" in the dropdown menu?

_Originally posted by @adammw in https://github.com/lensapp/lens/issues/1040#issuecomment-771993237_

### DIFF
--- a/src/renderer/components/+namespaces/namespace-select.tsx
+++ b/src/renderer/components/+namespaces/namespace-select.tsx
@@ -13,6 +13,8 @@ import { kubeWatchApi } from "../../api/kube-watch-api";
 
 interface Props extends SelectProps {
   showIcons?: boolean;
+  showClusterOption?: boolean; // show "Cluster" option on the top (default: false)
+  showAllNamespacesOption?: boolean; // show "All namespaces" option on the top (default: false)
   customizeOptions?(options: SelectOption[]): SelectOption[];
 }
 
@@ -34,8 +36,14 @@ export class NamespaceSelect extends React.Component<Props> {
   }
 
   @computed get options(): SelectOption[] {
-    const { customizeOptions } = this.props;
+    const { customizeOptions, showClusterOption, showAllNamespacesOption } = this.props;
     let options: SelectOption[] = namespaceStore.items.map(ns => ({ value: ns.getName() }));
+
+    if (showAllNamespacesOption) {
+      options.unshift({ label: "All Namespaces", value: "" });
+    } else if (showClusterOption) {
+      options.unshift({ label: "Cluster", value: "" });
+    }
 
     if (customizeOptions) {
       options = customizeOptions(options);
@@ -57,7 +65,7 @@ export class NamespaceSelect extends React.Component<Props> {
   };
 
   render() {
-    const { className, showIcons, showClusterOption, clusterOptionLabel, customizeOptions, ...selectProps } = this.props;
+    const { className, showIcons, customizeOptions, ...selectProps } = this.props;
 
     return (
       <Select
@@ -77,12 +85,14 @@ export class NamespaceSelectFilter extends React.Component {
     const namespaces = namespaceStore.getContextNamespaces();
 
     switch (namespaces.length) {
+      case namespaceStore.allowedNamespaces.length:
+        return <>All namespaces</>;
+      case 0:
+        return <>Select a namespace</>;
       case 1:
         return <>Namespace: {namespaces[0]}</>;
-      case 2:
-        return <>Namespaces: {namespaces.join(", ")}</>;
       default:
-        return "All Namespaces";
+        return <>Namespaces: {namespaces.join(", ")}</>;
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/renderer/components/+namespaces/namespace-select.tsx
```diff
@@ -13,17 +13,12 @@ import { kubeWatchApi } from "../../api/kube-watch-api";
 
 interface Props extends SelectProps {
   showIcons?: boolean;
-  showClusterOption?: boolean; // show cluster option on the top (default: false)
-  clusterOptionLabel?: React.ReactNode; // label for cluster option (default: "Cluster")
```
ixrock, line 17: Not used in project.

src/renderer/components/+namespaces/namespace-select.tsx
```diff
@@ -80,32 +73,61 @@ export class NamespaceSelect extends React.Component<Props> {
...
-    let placeholder = <>All namespaces</>;
+  @computed get placeholder(): React.ReactNode {
+    const namespaces = namespaceStore.getContextNamespaces();
+
+    switch (namespaces.length) {
+      case 1:
+        return <>Namespace: {namespaces[0]}</>;
+      case 2:
+        return <>Namespaces: {namespaces.join(", ")}</>;
+      default:
```
adammw, line 94: Shouldn't this be for `case 0`, and the join the default?

src/renderer/components/+namespaces/namespace-select.tsx
Nokel81: This is a breaking change

src/renderer/components/+namespaces/namespace-select.tsx
```diff
@@ -80,32 +73,61 @@ export class NamespaceSelect extends React.Component<Props> {
...
+      { value: "", label: "All Namespaces" },
+      ...options
+    ];
+  };
 
+  onChange = ([{ value: namespace }]: SelectOption[]) => {
+    if (namespace) {
+      namespaceStore.toggleContext(namespace);
+    } else {
+      namespaceStore.toggleAll(); // "All namespaces" option clicked
```
Nokel81, line 119: Given that this could toggle many, many namespaces I think that the ability to toggle "on" is a bit dangerous. Though I think a "clearer" way of "clearing all selected" might be a good idea instead.

src/renderer/components/+namespaces/namespace-select.tsx
```diff
@@ -80,32 +73,61 @@ export class NamespaceSelect extends React.Component<Props> {
...
+    }
+  };
+
+  render() {
     return (
       <NamespaceSelect
-        placeholder={placeholder}
+        isMulti={true}
```
Nokel81, line 126:
```suggestion
        isMulti
```

src/renderer/components/+namespaces/namespace-select.tsx
ixrock: It wasn't even documented.

Nokel81: Oh I see, sorry. It is exposed but it hasn't been exposed in a release.

src/renderer/components/+namespaces/namespace-select.tsx
jakolehm: @Nokel81 how would you select n-hundred namespaces without "all"? IMHO "all" is quite critical for cluster-admins who really want to see everything happening in a cluster.

Nokel81: Fair I guess. Maybe limit this to people who we detect are admins?

src/renderer/components/+namespaces/namespace-select.tsx
ixrock: Anyway, thanks for pointing it out. It seems that `showClusterOption` is used internally in `<AddRoleBindingDialog/>`, so this one is reverted.

src/renderer/components/+namespaces/namespace-select.tsx
Nokel81: Why was this resolved?

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -34,7 +34,7 @@ export function getDummyNamespace(name: string) {
 export class NamespaceStore extends KubeObjectStore<Namespace> {
   api = namespacesApi;
 
-  @observable contextNs = observable.array<string>();
+  @observable private contextNs = observable.set<string>();
```
ixrock, line 37: Making it as `private` and updating to `observable.set` is a breaking change.
Public version for getting context namespaces is `NamespaceStore.getContextNamespaces(): string[]`

ixrock:
@jakolehm @Nokel81 please re-review

Nokel81 (approved):

Nokel81:
Though I would like to know why you resolved a review comment about an style change without responding

jakolehm (approved):

src/renderer/components/+namespaces/namespace-select.tsx
ixrock: Because I prefer as it was :) (more verbose)
Please don't be surprised for silent resolves of _code-suggestions_.

Nokel81: How does this not go against the PR recommendations of not changing code style (especially without discussion).

And no I don't think I should be expecting code suggestions to just be silently ignored.

PR #2083, opened 2021-02-04
Watch-api fixes

ixrock:
_Issue:_
> it seems that we are missing some watch events, sometimes it works, often not

_Demo_

https://user-images.githubusercontent.com/6377066/107210881-6cf44900-6a0d-11eb-9b26-764c61c2bd67.mov

### DIFF
--- a/src/renderer/api/kube-api.ts
+++ b/src/renderer/api/kube-api.ts
@@ -1,4 +1,5 @@
 // Base class for building all kubernetes apis
+// Docs: https://kubernetes.io/docs/reference/using-api/api-concepts
 
 import merge from "lodash/merge";
 import { stringify } from "querystring";
@@ -7,7 +8,7 @@ import logger from "../../main/logger";
 import { apiManager } from "./api-manager";
 import { apiKube } from "./index";
 import { createKubeApiURL, parseKubeApi } from "./kube-api-parse";
-import { KubeJsonApi, KubeJsonApiData, KubeJsonApiDataList } from "./kube-json-api";
+import { KubeJsonApi, KubeJsonApiData, KubeJsonApiDataList, KubeJsonApiListMetadata, KubeJsonApiListMetadataParsed, KubeJsonApiResponse } from "./kube-json-api";
 import { IKubeObjectConstructor, KubeObject } from "./kube-object";
 import { kubeWatchApi } from "./kube-watch-api";
 
@@ -233,8 +234,43 @@ export class KubeApi<T extends KubeObject = any> {
     return this.resourceVersions.get(namespace);
   }
 
-  async refreshResourceVersion(params?: { namespace: string }) {
-    return this.list(params, { limit: 1 });
+  async listMetadata({ namespace = "" } = {}): Promise<KubeJsonApiListMetadataParsed> {
+    const response = await this.rawList({ namespace }, {
+      limit: 1, // specifying a limit to get metadata.remainingItemCount in response
+      resourceVersion: this.getResourceVersion(namespace) ?? "1",
+      resourceVersionMatch: "NotOlderThan",
+    });
+
+    const { remainingItemCount, ...metadata } = this.parseMetadata(namespace, response.metadata);
+
+    return {
+      ...metadata,
+      itemsCount: remainingItemCount + 1, // +1 from limit=1
+    };
+  }
+
+  async getItemsCount(): Promise<number> {
+    try {
+      const { itemsCount } = await this.listMetadata(); // list request for all namespaces
+
+      return itemsCount;
+    } catch (error) {
+      logger.error(`[KUBE-API]: getItemsTotal() has failed: ${error}`);
+    }
+
+    return 0;
+  }
+
+  async refreshResourceVersion(params?: { namespace: string }): Promise<string> {
+    try {
+      const { resourceVersion } = await this.listMetadata(params);
+
+      return resourceVersion;
+    } catch (error) {
+      logger.error(`[KUBE-API]: refreshing resourceVersion has failed: ${error}`, { params });
+    }
+
+    return "";
   }
 
   getUrl({ name = "", namespace = "" } = {}, query?: Partial<IKubeApiQueryParams>) {
@@ -261,7 +297,7 @@ export class KubeApi<T extends KubeObject = any> {
     return query;
   }
 
-  protected parseResponse(data: KubeJsonApiData | KubeJsonApiData[] | KubeJsonApiDataList, namespace?: string): any {
+  protected parseResponse(data: KubeJsonApiResponse, namespace?: string): any {
     const KubeObjectConstructor = this.objectConstructor;
 
     if (KubeObject.isJsonApiData(data)) {
@@ -276,9 +312,8 @@ export class KubeApi<T extends KubeObject = any> {
     if (KubeObject.isJsonApiDataList(data)) {
       const { apiVersion, items, metadata } = data;
 
-      // save "resourceVersion" metadata from list requests
-      this.setResourceVersion(namespace, metadata.resourceVersion);
-      this.setResourceVersion("", metadata.resourceVersion);
+      // parse & process metadata
+      this.parseMetadata(namespace, metadata);
 
       return items.map((item) => {
         const object = new KubeObjectConstructor({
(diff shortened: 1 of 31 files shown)


src/renderer/kube-object.store.ts
```diff
@@ -237,20 +240,22 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
-          }
+          this.items.remove(item);
           break;
       }
     }
 
-    // update items
-    this.items.replace(this.sortItems(items.slice(-this.bufferSize)));
+    // sort and limit items to store's maximum
+    const newItems = this.sortItems(this.items.slice(-this.bufferSize));
```
Nokel81: Should there be a logged warning if the store's max buffer size would have been overrun?

src/renderer/item.store.ts
```diff
@@ -27,6 +27,10 @@ export abstract class ItemStore<T extends ItemObject = ItemObject> {
     return this.items.find(item => item.getName() === name);
   }
 
+  getIndex(item: T): number {
+    return this.items.findIndex(i => i === item);
+  }
```
Nokel81: A more general form would be more useful
```suggestion
  getIndexById(id: string): number {
    return this.items.findIndex(i => i.getId() === id);
  }
```

src/renderer/kube-object.store.ts
```diff
@@ -224,11 +224,14 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
 
-    for (const { type, object } of this.eventsBuffer.clear()) {
-      const index = items.findIndex(item => item.getId() === object.metadata?.uid);
-      const item = items[index];
+    if (!eventsBuffer.length) {
+      return;
+    }
+
+    for (const { type, object } of eventsBuffer) {
+      const item = this.getById(object.metadata?.uid);
```
Nokel81: We don't use the item at all, so we might as well go straight for finding the index.
```suggestion
      const itemIndex = this.getIndexById(object.metadata?.uid);
```

src/renderer/kube-object.store.ts
```diff
@@ -237,20 +240,22 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
           const newItem = new api.objectConstructor(object);
 
           if (!item) {
-            items.push(newItem);
+            this.items.push(newItem);
           } else {
-            items.splice(index, 1, newItem);
+            const index = this.getIndex(item);
+
+            this.items.splice(index, 1, newItem);
```
Nokel81: When we already have the index we can just override directly.

```js
if (itemIndex < 0) { // not found
  this.items.push(newItem);
} else {
  this.items[index] = newItem;
}
```

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -103,8 +103,20 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
     return [];
   }
 
-  public getContextNamespaces(): string[] {
-    return Array.from(this.contextNs);
+  getContextNamespaces(): string[] {
```
Nokel81: This function isn't just getting the namespaces listed in the context, its name should be changed

ixrock:
@lensapp/lens-maintainers please re-review, did some refactoring, probably fixing the issue or close enough

src/renderer/kube-object.store.ts
ixrock: Maybe, currently out of PR's scope

src/renderer/components/+namespaces/namespace.store.ts
ixrock: Not really. It differs from `this.contexNs` list. Returns all allowed namespaces list when, for example, nothing selected (aka "All namespaces")

ixrock:
_Test_:

`kubectl scale --replicas=3 deployment/coredns -n=kube-system`

https://user-images.githubusercontent.com/6377066/106932090-f8c15900-671f-11eb-87c8-37df9efb29e3.mov

src/renderer/components/+namespaces/namespace.store.ts
Nokel81: Yes I know it differs from that list, but it isn't really just getting that list (which is what I would guess from the name).

How about something like `getCurrentSelectedNamespaces()`?

ixrock: Sounds good. But it more aligns with context meaning of namespaces over just selection.

jakolehm (dismissed):

src/renderer/kube-object.store.ts
```diff
@@ -24,6 +26,36 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
+
+  private async resolveCluster(): Promise<Cluster> {
+    const { getHostedCluster } = await import("../common/cluster-store");
+
+    this.cluster = getHostedCluster();
+
+    return this.cluster;
+  }
+
+  // TODO: figure out how to transparently replace with this.items
```
ixrock, line 52: Not sure if it's still needed. Probably not since default `replace=false` for `loadAll()`?

ixrock: Anyway, this is currently not used anywhere `computed` field.

src/renderer/api/kube-watch-api.ts
```diff
@@ -274,12 +275,29 @@ export class KubeWatchApi {
       try {
         const kubeEvent: IKubeWatchEvent = JSON.parse(json);
         const message = this.getMessage(kubeEvent);
+        const { data, namespace } = message;
 
-        if (!this.namespaces.includes(message.namespace)) {
-          continue; // skip updates from non-watching resources context
+        if (!this.isAllowedApi(message.api)) {
+          return;
```
jim-docker: `return "";` ?

I find this method confusing. If this event is for an api that is not allowed then no message is emitted for it and all remaining events in `events` are abandoned? Similarly, if there is an error, it's caught below and this event is returned but again all remaining events in `events` are abandoned?

src/renderer/kube-object.store.ts
```diff
@@ -132,10 +155,30 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
+
+    if (filter) items = this.filterItemsOnLoad(items);
+    if (sort) items = this.sortItems(items);
+    if (updateStore) this.items.replace(items);
+
+    return items;
+  }
 
-    return this.loadAll(namespaceStore.getContextNamespaces());
+  async loadContextNamespaces(): Promise<void> {
```
jim-docker: My confusion with the naming is that this sounds like it loads namespaces when it's actually loading items found in the namespaces. `loadAllFromContextNamespaces`? Or `loadAllFromContextNs`?

src/renderer/api/kube-watch-api.ts
ixrock: Yes, you're right. Probably should be remove or replaced with `continue`.

src/renderer/kube-object.store.ts
ixrock: Updated to `loadAllFromContextNamespaces()`

src/renderer/api/kube-watch-api.ts
ixrock: Removed.

ixrock:
ready for re-review

src/renderer/kube-object.store.ts
```diff
@@ -106,23 +131,20 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
   }
 
   @action
-  async loadAll(namespaces: string[] = []): Promise<void> {
+  async loadAll(namespaces?: string[], { replace = false /*partial update*/ } = {}): Promise<void> {
```
jakolehm: I'm not completely sold on the `replace` parameter idea here. Could we just always remove all items that match given namespaces before we call `loadItems()`? By doing that we would always have a fresh data for given namespaces and store would still keep other namespaces in memory.

```ts
      if (this.api.isNamespaced) {
        // clear any existing items from given namespaces
        const leftItems = this.items.toJS().filter(item => !namespaces.includes(item.getNs()));

        this.items.replace(leftItems);
      } else {
        this.items.replace([]);
      }

      const items = await this.loadItems({ namespaces, api: this.api });
      
      this.items.replace([...this.items.toJS(), ...items]);
```

ixrock: We cannot replace _all items_ in store when loading partial namespaces. (since store is singleton and extensions might override items). By clearing items from just  freshly loaded namespaces is a good idea.

ixrock:
@jakolehm @Nokel81 @jim-docker please test again, seems to work.

_Changes:_
- store keep only relevant context items managed in `NamespaceSelect` (in other words by app's global filters)
- for extension users: by default `loadAll()` loads all available namespaces and updates the store (to avoid replacing items in store `loadAll({merge=false})` must be used, it will return loaded items instead)
- detached `NamespaceStore` from `KubeObjectStore` (all relevant context data injected in `app.tsx`)

src/renderer/kube-object.store.ts
ixrock: Reverted logic back as was in 4.0. 
Now it replaces all items in the store and reflects with `NamespaceSelect` state.

src/renderer/api/kube-api.ts
```diff
@@ -35,6 +36,7 @@ export interface IKubeApiOptions<T extends KubeObject> {
 export interface IKubeApiQueryParams {
   watch?: boolean | number;
   resourceVersion?: string;
+  resourceVersionMatch?: "Exact" | "NotOlderThan" // see also: https://kubernetes.io/docs/reference/using-api/api-concepts/#resourceversion-in-metadata
```
jakolehm, line 39: I think we cannot trust these on older kube versions.

src/renderer/api/kube-api.ts
```diff
@@ -232,8 +234,43 @@ export class KubeApi<T extends KubeObject = any> {
...
+      limit: 1, // specifying a limit to get metadata.remainingItemCount in response
+      resourceVersion: this.getResourceVersion(namespace) ?? "1",
+      resourceVersionMatch: "NotOlderThan",
+    });
+
+    const { remainingItemCount, ...metadata } = this.parseMetadata(namespace, response.metadata);
+
+    return {
+      ...metadata,
+      itemsCount: remainingItemCount + 1, // +1 from limit=1
```
jakolehm, line 248: `remainingItemCount` is interesting idea but count is not updated "live". Also I'm not sure if it's available on older kube versions.

src/renderer/api/kube-api.ts
ixrock: This should be backward compatible as far as I know 🤔

src/renderer/api/kube-api.ts
ixrock: in addition we could calculate actual loaded items in store per `namespace` for better results and/or refresh this a bit more often (it's lightweight request)

PR #2084, opened 2021-02-04
Fix: select all namespaces

ixrock:
Previously "All namespaces" option was working as select/deselect all.
Now it's always reset to all namespaces (empty list).

**Demo**

https://user-images.githubusercontent.com/6377066/106899604-3ad9a300-66fe-11eb-94e2-0ade59d3cdb5.mp4

### DIFF
--- a/src/renderer/api/kube-watch-api.ts
+++ b/src/renderer/api/kube-watch-api.ts
@@ -70,6 +70,7 @@ export class KubeWatchApi {
         return [];
       }
 
+      // TODO: optimize - check when all namespaces are selected and then request all in one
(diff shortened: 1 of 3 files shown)


Nokel81 (approved):

jim-docker (approved):
It's a little weird that when you select all the namespaces manually, they all have check marks (Lens recognizes you've selected all the namespaces with "All Namespaces" as selected) but then if you click "All Namespaces" the check marks disappear.


https://user-images.githubusercontent.com/40840436/106929732-ba597900-66e2-11eb-83ed-0889e6139ea7.mov

PR #2085, opened 2021-02-04
Add more details to Helm rollback window

### LINKED ISSUE #583: Helm rollback window with more details
When we use helm rollback via Apps -> Releases it's hard to tell to which revision we would like to roll it back as there is only release number and chart version - would be great to add to the dropdown menu App Version and update date same as the output of "helm history xyz"

**Why is this needed**:
To make Lens more useful in regards to helm

- Kubernetes distribution: AWS/Kops
- Desktop OS: Fedora

vshakirova:
Screen Shot 2021-02-04 at 12 22 15 PM

Fixes: https://github.com/lensapp/lens/issues/583

### DIFF
--- a/src/renderer/api/endpoints/helm-releases.api.ts
+++ b/src/renderer/api/endpoints/helm-releases.api.ts
@@ -57,6 +57,7 @@ export interface IReleaseRevision {
   updated: string;
   status: string;
   chart: string;
+  app_version: string;
   description: string;
 }
 
--- a/src/renderer/components/+apps-releases/release-rollback-dialog.tsx
+++ b/src/renderer/components/+apps-releases/release-rollback-dialog.tsx
@@ -77,7 +77,8 @@ export class ReleaseRollbackDialog extends React.Component<Props> {
           themeName="light"
           value={revision}
           options={revisions}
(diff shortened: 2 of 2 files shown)


src/renderer/api/endpoints/helm-releases.api.ts
```diff
@@ -57,6 +57,7 @@ export interface IReleaseRevision {
   updated: string;
   status: string;
   chart: string;
+  app_version: string;
```
Nokel81, line 60: We use camel case for interface field names
```suggestion
  appVersion: string;
```

Nokel81 (commented):
Looks good except for this minor change

src/renderer/api/endpoints/helm-releases.api.ts
vshakirova: @Nokel81 I tried `appVersion` but it didn't work. I checked API response and in response they used app_version
Screen Shot 2021-02-04 at 7 45 22 PM
Tell me please if I miss smth

Nokel81: Oh I didn't realize that this interface was representing the data gotten directly.

Nokel81 (approved):

ixrock (approved):

PR #2089, opened 2021-02-05
Revert store and watch refactorings

jakolehm:
I think we hit into limits of current kube-object-store architecture and we need to go back to the drawing board. Basically current stores were designed to keep full-state of given kubernetes api in-memory. We tried to change this so that we could keep only subset (selected namespaces) but we learned that it has unexpected side-effects because a store can have multiple consumers (with different requirements). 

I hope we can bring watch api fixes back as a separate PR.

Reverts #1918, #1990, #2063 and #2084 . 

Closes #2083


nevalla (approved):

ixrock:
Why "close #2083" when it could be re-merged from master and merged back eventually?

Issue #2077, opened 2021-02-03
Verify that existing extensions work with 4.1


jakolehm:
**Starboard Extension:**
- https://github.com/aquasecurity/starboard-lens-extension/issues/48
- https://github.com/aquasecurity/starboard-lens-extension/issues/47

jakolehm:
https://github.com/aquasecurity/starboard-lens-extension/issues/47 does not seem to be a blocker.

PR #2091, opened 2021-02-08
Fix: saving apiManager store keys with string type

### LINKED ISSUE #48: App crash when editing VulnerabilityReports
When testing the extension with Lens 4.1.0-alpha.1 release I've figured out that editing VulnerabilityReports sometimes causes app crash due to an uncaught error. 

Bildschirmfoto 2021-02-01 um 09 45 29

**Issue can be reproduced with the following versions of Lens:**
- Lens 4.0.8
- Lens 4.0.1-alpha.1
- Lens 4.0.1-alpha.2

**Steps to reproduce**
1. Install and Enable Starboard extension + Starboard Operator
2. Open a Starboard Vulnerability Report 
3. App Crash (see screenshot)
4. Disable Starboard Extension
5. Close Lens 
6. Open Lens From Menu Icon
7. Enable Starboard Extension
8. App Crash (see screenshot)
9. Goto step 4)

**OS:** Mac OS 10.14.6 (Mojave)

aleksfront:
Using `string` instead of `KubeApi` type for keys in `stores` map.

This allows extensions to rewrite stores while registering them. Previously shallow comparison decided that `KubeApi` passing as key is always a new object and should be added to the end of `stores` map.

Fixes https://github.com/aquasecurity/starboard-lens-extension/issues/48

Relates to #2077 

### DIFF
--- a/src/renderer/api/api-manager.ts
+++ b/src/renderer/api/api-manager.ts
@@ -7,7 +7,7 @@ import { KubeApi, parseKubeApi } from "./kube-api";
 @autobind()
 export class ApiManager {
   private apis = observable.map<string, KubeApi>();
-  private stores = observable.map<KubeApi, KubeObjectStore>();
+  private stores = observable.map<string, KubeObjectStore>();
 
   getApi(pathOrCallback: string | ((api: KubeApi) => boolean)) {
     if (typeof pathOrCallback === "string") {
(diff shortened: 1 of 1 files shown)


Nokel81 (dismissed):

src/renderer/api/api-manager.ts
```diff
@@ -46,12 +46,12 @@ export class ApiManager {
   @action
   registerStore(store: KubeObjectStore, apis: KubeApi[] = [store.api]) {
     apis.forEach(api => {
-      this.stores.set(api, store);
+      this.stores.set(api.kind, store);
```
jakolehm: I think we cannot use just `kind` here because multiple apis might have the same kind.

jakolehm (changes requested):

src/renderer/api/api-manager.ts
aleksfront: Switched to `apiBase` field. We consider it unique since storing `@observable apis` keys with same field
registered apis

jakolehm (approved):

Nokel81 (approved):

PR #2095, opened 2021-02-08
Refactor watches to use native k8s api

jakolehm:
* Changes visible total counts for a user: now it shows total count for a given context namespaces
* Refactors away custom watch api implementation: now uses native k8s api (each watch api is a separate request). Should help with debug etc because no requests are "hidden" behind main process

Based on #2083

Closes #2083
Closes #2089


jakolehm:
Blocker: browser `fetch` can only do limited number of concurrent requests.

src/renderer/api/json-api.ts
```diff
@@ -55,6 +55,20 @@ export class JsonApi<D = JsonApiData, P extends JsonApiParams = JsonApiParams> {
     return this.request<T>(path, params, { ...reqInit, method: "get" });
   }
 
+  getReadableStream(path: string, params?: P, init: RequestInit = {}): Promise<Response> {
```
ixrock: Maybe `getResponse()` since returns `Promise<Response>`?

src/renderer/api/kube-api.ts
```diff
@@ -9,7 +9,9 @@ import { apiKube } from "./index";
 import { createKubeApiURL, parseKubeApi } from "./kube-api-parse";
 import { KubeJsonApi, KubeJsonApiData, KubeJsonApiDataList } from "./kube-json-api";
 import { IKubeObjectConstructor, KubeObject } from "./kube-object";
-import { kubeWatchApi } from "./kube-watch-api";
+import byline from "byline";
+import { ReadableWebToNodeStream } from "readable-web-to-node-stream";
```
ixrock, line 13: What does this package(s)?
Maybe it's possible to use own custom helper/util based on https://developer.mozilla.org/en-US/docs/Web/API/Streams_API for that purpose?

src/renderer/api/kube-api.ts
```diff
@@ -91,6 +93,12 @@ export function ensureObjectSelfLink(api: KubeApi, object: KubeJsonApiData) {
   }
 }
 
+type KubeApiWatchOptions = {
+  namespace: string;
```
ixrock, line 97: Could be optional. Is it makes sense to pass `namespace: ""` for cluster-wide items?

src/renderer/api/kube-api.ts
```diff
@@ -357,8 +366,96 @@ export class KubeApi<T extends KubeObject = any> {
...
+        return;
+      }
+      const nodeStream = new ReadableWebToNodeStream(response.body);
+      const stream = byline(nodeStream);
+
+      stream.on("data", (line) => {
+        try {
+          const data: IKubeWatchEvent = JSON.parse(line);
+
+          console.log("data", data);
```
ixrock: I liked more logging output with color identification for type of event and time output (could help for example to detect how often watch-api reconnects / connection dropping from clusters).

_Before:_
Screenshot 2021-02-08 at 21 56 56

_Now:_
Screenshot 2021-02-08 at 21 57 13

src/renderer/kube-object.store.ts
```diff
@@ -219,7 +267,32 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
+    const callback = (data: IKubeWatchEvent) => {
+      this.eventsBuffer.push(data);
+    };
+
+    if (this.context.cluster.isGlobalWatchEnabled) {
+      disposers = apis.map(api => api.watch({
+        abortController,
+        namespace: "",
+        callback: (data) => callback(data),
+      }));
```
ixrock:
```suggestion
      disposers = apis.map(api => api.watch({abortController, callback}));
```

ixrock (dismissed):
Why don't we load all pods on app start? In `app.tsx` stores `podsStore, nodesStore, eventStore` should be preloaded with all items by default, right?

Screenshot 2021-02-08 at 21 54 30

Also whenever I change a namespace select I see always single number of items regarding to context filter. Is it feature or bug? Previously where like `Pods: 1 / 10`

Otherwise LGTM 👍
Tested with 5 namespaces, watch-api seems to work.

src/renderer/api/json-api.ts
jakolehm: Yeah, makes sense (I'm refactoring this anyway because of browser connection limits).

src/renderer/api/kube-api.ts
jakolehm: We can probably write our own helpers for this (or port logic from master branch). Biggest issue I see here is that we are using `fetch` which does not return "nodejs" compatible stream. `byline` wants nodejs stream, `ReadableWebToNodeStream` converts browser stream to nodejs stream ... it's a hack.

src/renderer/api/kube-api.ts
jakolehm: Optional is probably better.

src/renderer/api/kube-api.ts
jakolehm: Yes, logging is still WIP.. but I'm not sure if it's a good idea to log all events by default (it might be too noisy).

src/renderer/api/kube-api.ts
```diff
@@ -357,8 +366,96 @@ export class KubeApi<T extends KubeObject = any> {
...
-  watch(): () => void {
-    return kubeWatchApi.subscribeApi(this);
+  watch(opts: KubeApiWatchOptions = { namespace: "" }): () => void {
+    if (!opts.abortController) {
+      opts.abortController = new AbortController();
+    }
+    const { abortController, namespace, callback } = opts;
+
+    // FIXME
+    const watchUrl = `http://${this.kind}-watch.${window.location.host}${apiKubePrefix}${this.getWatchUrl(namespace)}`;
```
jakolehm: I'm out of ideas how to customise request host without passing full url. There must be a better way.

jakolehm: This is needed as a workaround for browser connection limits. It seems that chromium allows only 6 or so connections to a host. If we generate different subdomain for each watched kind we can overcome this limitation.

jim-docker:
Lens deployment scaling dialog shows 0 replicas (there are 4) and "Scale" button is grayed out

image

If I cancel then bring the dialog back up it shows 4 replicas but the "Scale" button remains grayed out, even when I move the slider

src/renderer/api/kube-api.ts
ixrock: Btw, since in `renderer` we have enabled `nodeIntegration: true` could we use native [node-js stream-api](https://nodejs.org/dist/latest-v12.x/docs/api/stream.html) in both processes?

src/renderer/api/kube-api.ts
ixrock: It was enabled by default under `vars.isDevelopment || vars.isDebugging` flags.

jakolehm:
> If I cancel then bring the dialog back up it shows 4 replicas but the "Scale" button remains grayed out, even when I move the slider

@jim-docker I assume you tested a version which did not have a workaround for browser connection limits? That caused request to block until watches were reseted (chromium allows only 6 connections per domain).

jakolehm:
I'm quite sure integration tests will pass now 😎

nevalla (approved):
LGTM

src/renderer/api/kube-api.ts
jakolehm: We could use native but `fetch` does not return nodejs native streams. That's why I'm converting streams here. Probably story for an another PR (we could probably replace fetch with nodejs library, like `got`)

src/renderer/api/kube-api.ts
jakolehm: I'll open an issue about watch api params once this is merged. I think this needs to be fixed before 4.1.0.

src/renderer/api/kube-api.ts
jakolehm: Ok, that makes sense. Let's fix it in follow-up PR.

src/renderer/api/kube-api.ts
ixrock: I meant if we could throw away `fetch` and use native node.js stream for both `main` and `renderer` => no transform stream(s) would be needed at the client-side.

jakolehm: > I meant if we could throw away `fetch` and use native node.js stream for both `main` and `renderer` => no transform stream(s) would be needed at the client-side.

Yeah, sounds like a plan. Only downside would be that browser devtools would not show up requests if we use nodejs libraries... but that probably is acceptable tradeoff.

src/renderer/api/kube-api.ts
```diff
@@ -357,8 +366,88 @@ export class KubeApi<T extends KubeObject = any> {
...
+      const nodeStream = new ReadableWebToNodeStream(response.body);
+      const stream = byline(nodeStream);
+
+      stream.on("data", (line) => {
+        try {
+          const event: IKubeWatchEvent = JSON.parse(line);
+
+          this.modifyWatchEvent(event);
+
+          if (callback) {
```
ixrock, line 406: Probably should be filtered in accordance with `context.allNamespaces` from `src/renderer/components/context.ts`?

ixrock (approved):
LGTM

src/renderer/api/kube-api.ts
jakolehm: I'm not sure about that.. if yes -> we should also limit all api queries to those namespaces 🤔

PR #2098, opened 2021-02-08
Use allocatable instead of capacity for node resources.

### LINKED ISSUE #2097: Memory metrics are misleading and not consistent with kubectl.
**Describe the bug**
The Cluster and Node Memory metrics are not intuitive. The statistics shown are using `kube_node_status_capacity` instead of `kube_node_status_allocatable` which can be highly misleading for certain cloud providers/node VM sizes (due to provider specific resource reservations described [here](https://learnk8s.io/allocatable-resources)). In my circumstance using Azure Kubernetes Service with Standard_B2s results in a reported utilization of ~55% when it is actually ~98% according to `kubectl describe node`. This difference makes it very difficult to understand why a cluster would report `insufficient memory` when scheduling pods. Memory consumption relative to the cluster or node's total Allocatable Memory is a much more useful metric for most users who are generally only concerned with the resources available for consumption by pods/workloads. This is supported by kubectl's default reporting of node memory (i.e. relative to the allocatable memory).

**To Reproduce**
1. Connect to a cloud hosted cluster (I am familiar with Azure, but any cluster with non-negligible built-in per-node resource reservation should suffice)
2. Install the prometheus extension in Lens on the cluster.
3. Deploy a pod with a substantial memory request (e.g. 1-2GB).
3. View the memory pie chart for the cluster in Lens.
4. Observe the reported memory request percentage.
5. Run `kubectl describe node` against the same cluster
6. Observe that the reported memory request percentage is different than in step 4.

**Expected behavior**
For the memory request metrics to reflect memory available for use by Pods (the allocatable memory), not total memory capacity.

**Screenshots**
Screenshots of the views reporting memory capacity .
Memory Pie Chart in the Cluster view:
image
Memory Bar Chart in the Nodes list:
image

- Lens Version: 4.0.8
- OS: Windows 10 64-bit
- Installation method (e.g. snap or AppImage in Linux): Windows setup exe.

mcpappas:
Fixes #2097.

The Cluster and Node Memory metrics are not intuitive. The statistics shown are using kube_node_status_capacity instead of kube_node_status_allocatable which can be highly misleading for certain cloud providers/node VM sizes (due to provider specific resource reservations described here). In my circumstance using Azure Kubernetes Service with Standard_B2s results in a reported utilization of ~55% when it is actually ~98% according to kubectl describe node. This difference makes it very difficult to understand why a cluster would report insufficient memory when scheduling pods. Memory consumption relative to the cluster or node's total Allocatable Memory is a much more useful metric for most users who are generally only concerned with the resources available for consumption by pods/workloads. This is supported by kubectl's default reporting of node memory (i.e. relative to the allocatable memory).

### DIFF
--- a/src/renderer/components/+cluster/cluster-pie-charts.tsx
+++ b/src/renderer/components/+cluster/cluster-pie-charts.tsx
@@ -45,20 +45,20 @@ export const ClusterPieCharts = observer(() => {
 
   const renderCharts = () => {
     const data = getMetricLastPoints(clusterOverviewStore.metrics);
-    const { memoryUsage, memoryRequests, memoryCapacity, memoryLimits } = data;
-    const { cpuUsage, cpuRequests, cpuCapacity, cpuLimits } = data;
-    const { podUsage, podCapacity } = data;
-    const cpuLimitsOverload = cpuLimits > cpuCapacity;
-    const memoryLimitsOverload = memoryLimits > memoryCapacity;
+    const { memoryUsage, memoryRequests, memoryAllocatableCapacity, memoryCapacity, memoryLimits } = data;
+    const { cpuUsage, cpuRequests, cpuAllocatableCapacity, cpuCapacity, cpuLimits } = data;
+    const { podUsage, podAllocatableCapacity, podCapacity } = data;
+    const cpuLimitsOverload = cpuLimits > cpuAllocatableCapacity;
+    const memoryLimitsOverload = memoryLimits > memoryAllocatableCapacity;
     const defaultColor = ThemeStore.getInstance().activeTheme.colors.pieChartDefaultColor;
-
-    if (!memoryCapacity || !cpuCapacity || !podCapacity) return null;
+    
+    if (!memoryCapacity || !cpuCapacity || !podCapacity || !memoryAllocatableCapacity || !cpuAllocatableCapacity || !podAllocatableCapacity) return null;
     const cpuData: ChartData = {
       datasets: [
         {
           data: [
             cpuUsage,
-            cpuUsage ? cpuCapacity - cpuUsage : 1,
+            cpuUsage ? cpuAllocatableCapacity - cpuUsage : 1,
           ],
           backgroundColor: [
             "#c93dce",
@@ -70,7 +70,7 @@ export const ClusterPieCharts = observer(() => {
         {
           data: [
             cpuRequests,
-            cpuRequests ? cpuCapacity - cpuRequests : 1,
+            cpuRequests ? cpuAllocatableCapacity - cpuRequests : 1,
           ],
           backgroundColor: [
             "#4caf50",
@@ -82,7 +82,7 @@ export const ClusterPieCharts = observer(() => {
         {
           data: [
             cpuLimits,
-            cpuLimitsOverload ? 0 : cpuCapacity - cpuLimits,
+            cpuLimitsOverload ? 0 : cpuAllocatableCapacity - cpuLimits,
           ],
           backgroundColor: [
             "#3d90ce",
@@ -96,6 +96,7 @@ export const ClusterPieCharts = observer(() => {
         `Usage: ${cpuUsage ? cpuUsage.toFixed(2) : "N/A"}`,
         `Requests: ${cpuRequests ? cpuRequests.toFixed(2) : "N/A"}`,
         `Limits: ${cpuLimits ? cpuLimits.toFixed(2) : "N/A"}`,
+        `Allocatable Capacity: ${cpuAllocatableCapacity || "N/A"}`,
         `Capacity: ${cpuCapacity || "N/A"}`
       ]
     };
@@ -104,7 +105,7 @@ export const ClusterPieCharts = observer(() => {
         {
           data: [
             memoryUsage,
-            memoryUsage ? memoryCapacity - memoryUsage : 1,
+            memoryUsage ? memoryAllocatableCapacity - memoryUsage : 1,
           ],
           backgroundColor: [
             "#c93dce",
@@ -116,7 +117,7 @@ export const ClusterPieCharts = observer(() => {
         {
           data: [
             memoryRequests,
-            memoryRequests ? memoryCapacity - memoryRequests : 1,
+            memoryRequests ? memoryAllocatableCapacity - memoryRequests : 1,
           ],
           backgroundColor: [
             "#4caf50",
@@ -128,7 +129,7 @@ export const ClusterPieCharts = observer(() => {
         {
           data: [
             memoryLimits,
-            memoryLimitsOverload ? 0 : memoryCapacity - memoryLimits,
+            memoryLimitsOverload ? 0 : memoryAllocatableCapacity - memoryLimits,
           ],
           backgroundColor: [
             "#3d90ce",
@@ -142,7 +143,8 @@ export const ClusterPieCharts = observer(() => {
         `Usage: ${bytesToUnits(memoryUsage)}`,
         `Requests: ${bytesToUnits(memoryRequests)}`,
         `Limits: ${bytesToUnits(memoryLimits)}`,
-        `Capacity: ${bytesToUnits(memoryCapacity)}`,
+        `Allocatable Capacity: ${bytesToUnits(memoryAllocatableCapacity)}`,
+        `Capacity: ${bytesToUnits(memoryCapacity)}`
       ]
     };
     const podsData: ChartData = {
@@ -150,7 +152,7 @@ export const ClusterPieCharts = observer(() => {
         {
           data: [
             podUsage,
-            podUsage ? podCapacity - podUsage : 1,
+            podUsage ? podAllocatableCapacity - podUsage : 1,
           ],
           backgroundColor: [
             "#4caf50",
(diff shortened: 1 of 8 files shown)


Nokel81:
Would it make sense to have both? Is there a reason why `podCapacity` is not down based on "allocatable" as well?

mcpappas:
Ah, I didn't think PodCapacity had an "allocatable" allegory, but it appears I was wrong according to `kubectl describe node`'s output.

mcpappas:
> Would it make sense to have both?

I don't think it doesn't make sense to include both. In the case of Azure Kubernetes Service (which I believe is a representative example), not all of that total capacity is accessible for use by workloads, and Lens is only able to report on usage/requests/limits by proper workloads. The portion that isn't accessible to workloads is consumed by "other processes" like kubelet itself, and the memory usage of those other processes isn't something Lens can access afaik. 

TL;DR; Since Lens can't account for the usage of that reserved memory, I believe it would only be confusing to report on that capacity.

mcpappas:
I've had a change of heart. In debugging locally, I realized that there are scenarios in which the allocatable memory capacity paints an incomplete picture also. I'm going to convert this PR into a draft while I come up with a way to accurately represent cluster/node memory conditions for all scenarios.

Basically, I was operating under a false premise that the "allocatable" memory was _reserved_ for workloads just as the "non-allocatable" memory is reserved for kubelet and other  "non-workloads", but that isn't the case at all. That means that there are circumstances where the memory actually available for workloads is less than the "allocatable" memory (especially in clusters that don't really reserve memory for kubelet at all, allocatable == total capacity). 
I am thinking that means that it would be most accurate to display a breakdown of memory that distinguishes between "workload" and "non-workload" (things like kubelet) consumption as well as displaying "allocatable", "non-allocatable" and "total capacity".

danderbury:
Did you get anywhere with this? Its nearly really closely bitten us.

jonstelly:
> I've had a change of heart...

I think how you phrased it makes perfect sense and I found this issue looking around about my AKS cluster so I may be biased, but the allocatable value seems to make much more sense for what I think people using lens are interested in or would expect, right?

> especially in clusters that don't really reserve memory for kubelet at all, allocatable == total capacity

I don't know how all different flavors and deployments of Kubernetes do this but if that's the worst case then wouldn't this change as you initially implemented it not really impact those folks?  I imagine there are no sane deployments where allocatable > total node memory but even if there were (maybe some dynamic-memory VM?) wouldn't showing the allocatable still feel like it makes more sense, assuming the person specifically configuring it that way may best know the specifics of their environment?

It seems like deploying this would be a reasonable interim/incremental improvement, so I guess I'm trying to suggest or ask that you ship it.

I can appreciate not wanting to introduce churn if you feel like you're going to have to redo this commit at some point in the near future, but in the interim it's very difficult to use lens to get an accurate view of my AKS cluster.

JackSinclairT:
Where are we with this? We've had some nasty surprises on our production clusters with this bug

mcpappas:
Sorry, I didn't realize there was interest in this issue/PR because github notifications were going to my junk folder for some reason.

I had set this PR aside when I realized the relationship between Allocatable/Non-Allocatable/Workload/Non-Workload resources were more complex than I thought, but I will try to take another look at how to display a more accurate representation of node resources in the next week or so.

jonstelly:
Thanks, I appreciate you taking the time for another look.

The AKS memory situation on smaller nodes is a bit silly, 4GB VM only has 2.2GB allocatable so I see evictions when it looks like I'm only using ~60% of a node's memory.  Honestly, this is more of an AKS problem than a Lens problem but I'm sure AKS users everywhere would rejoice with some sort of fix.

mcpappas:
> The AKS memory situation on smaller nodes is a bit silly, 4GB VM only has 2.2GB allocatable so I see evictions when it looks like I'm only using ~60% of a node's memory. Honestly, this is more of an AKS problem than a Lens problem but I'm sure AKS users everywhere would rejoice with some sort of fix.

That was exactly our scenario. We were using the Standard_B2s class also when we came across this and it took some time to figure out so when I finally did I flew around the internet raising issues on all the tools that failed me. I raised this to Microsoft because even their graphs in the Azure Portal are misleading in the same way. They reached out to me to ask some questions (which I thought was encouraging at the time), but it doesn't seem like they are going to address it. I think I made an issue on the Kubernetes Dashboard project on GitHub also but I don't recall if I kept up with that one either 😬.

mcpappas:
Okay, it took some minor surgery to catch up to master, but I realized I actually had some (uncommitted!!) changes locally that I think actually portrays things pretty accurately. I must have made them months ago though, so I don't recall if I tracked down all the locations where the memory graphs show up in the UI to make sure they all make sense. 

I think I am comfortable setting this as ready for review at this point. 🤞

danderbury:
Really good to read this grabbing traction again. 

I agree it’s more of an AKS translation issue than lens, but we spend our days in lens monitoring (fab fab tool) so it would save us running frequent commands. 

I imagine there’s many like use that got bit by thinking we had more resources than we did. 

Thanks!

Nokel81:
@mcpappas So far this looks good, we plan on merging #2707 so this PR will need to change. Just FYI since it hasn't been merged yet.

mcpappas:
@Nokel81 Thanks for the heads up, I’ll take a look when I get a chance.

nevalla (dismissed):
I think the changes are reasonable. LGTM.

Nokel81:
@mcpappas Once the conflicts are resolved this should be good to go.

src/renderer/components/+cluster/cluster-pie-charts.tsx
```diff
@@ -45,20 +45,20 @@ export const ClusterPieCharts = observer(() => {
...
-    const memoryLimitsOverload = memoryLimits > memoryCapacity;
+    const { memoryUsage, memoryRequests, memoryAllocatableCapacity, memoryCapacity, memoryLimits } = data;
+    const { cpuUsage, cpuRequests, cpuAllocatableCapacity, cpuCapacity, cpuLimits } = data;
+    const { podUsage, podAllocatableCapacity, podCapacity } = data;
+    const cpuLimitsOverload = cpuLimits > cpuAllocatableCapacity;
+    const memoryLimitsOverload = memoryLimits > memoryAllocatableCapacity;
     const defaultColor = ThemeStore.getInstance().activeTheme.colors.pieChartDefaultColor;
-
-    if (!memoryCapacity || !cpuCapacity || !podCapacity) return null;
+
```
BlackHole1, line 54: Excess space

Nokel81:
@mcpappas Please fix the conflicts.

jonstelly:
I took a stab at resolving conflicts: https://github.com/kymaria/lens/tree/feature/allocatable

I'm unable to test locally...  I can't figure out how to install the lens-metrics prometheus resources in lens 5 and neither the beta 8 build or my development build of the branch above seems to be able to connect to my custom prometheus deployments.

But I hope this helps, the merge conflicts seemed pretty straight-forward to resolve and I'd really like to see this moved along in time for 5.0.

mcpappas:
@jonstelly, @Nokel81, sorry, work has kept me very busy lately so I haven’t had time to take a stab myself. I can run Jon’s changes locally and adopt them if it checks out. I’ll shoot to take a look this afternoon.

jonstelly:
No need to apologize, I appreciate all your work on this.

src/renderer/components/+nodes/nodes.tsx
```diff
@@ -97,7 +97,7 @@ export class Nodes extends React.Component<Props> {
   }
 
   renderMemoryUsage(node: Node) {
-    const metrics = nodesStore.getLastMetricValues(node, ["memoryUsage", "memoryCapacity"]);
+    const metrics = nodesStore.getLastMetricValues(node, ["workloadMemoryUsage", "memoryAllocatableCapacity"]);
```
mcpappas, line 100: This is likely misleading when `(memoryAllocatableCapacity - memoryCapacity) < (memoryUsage - workloadMemoryUsage)`. I considered revising it, but I am pressed for time.

Nokel81: What do you think it should be instead during when that is true?

Nokel81 (approved):
Thank you very much for this improvement. If you want to improve it further down the line. More PRs would be very much appreciated.

src/renderer/components/+nodes/nodes.tsx
mcpappas: I think my first comment had a mistake (was rushing to catch a train). When `memoryCapacity - memoryAllocatableCapacity < memoryUsage - workloadMemoryUsage` Then it probably makes more sense to compare `memoryUsage - (memoryCapacity - memoryAllocatableCapacity)` to `memoryAllocatableCapacity`.

That probably seems a bit like magic math, so to be clear: 
`(memoryCapacity - memoryAllocatableCapacity)` represents the portion of memory that is reserved for non-workloads (kubelet, etc.). If those “non-workloads” exceed the reserved amount (or there was no reservation at all) then they begin to consume `memoryAllocatableCapacity`.

Nokel81: Ah that makes sense. But can be done in the future.

mcpappas: Agreed, the other charts should provide sufficient clarity for now, and would be a small change.

PR #2104, opened 2021-02-09
Fix: preventing namespace filter jumping scroll

aleksfront:
It reduces the number of unnecessary rerenders of `React Select` component which caused jumping scroll behavior.

* Moving out placeholder updates to separate component (thanks React Select for such customization ability)
* Not reacting if no namespaces changes

Here is the fix working:

https://user-images.githubusercontent.com/9607060/107356088-69ca8d00-6ae1-11eb-9791-738af69f204f.mp4


src/renderer/components/+namespaces/namespace-select-filter.tsx
```diff
@@ -0,0 +1,77 @@
...
+          <span>{namespace}</span>
+          {isSelected && <Icon small material="check" className="box right"/>}
+        </div>
+      );
+    }
+
+    return label;
+  }
+
+  onChange([{ value: namespace }]: SelectOption[]) {
```
vshakirova, line 55: Hi @aleksfront. Could you please explain why do we use SelectOption[] as an array but `toggleContext()` accepts only one string?
 I checked parent `onChange(value: SelectOption, meta: ActionMeta<any>){  if (this.props.onChange) {
      this.props.onChange(value, meta);
    } }` 
    func and it always contains one element array and then it passes this element as a string(if i am not mistaken) to custom `onChange` func().

Nokel81: `onChange`'s argument is `SelectOption[]` because we are dealing with an `isMulti={true}` table.

src/renderer/components/+namespaces/namespace-select-filter.tsx
```diff
@@ -0,0 +1,77 @@
...
+    if (namespace) {
+      namespaceStore.toggleContext(namespace);
+    } else {
+      namespaceStore.resetContext(); // "All namespaces" clicked, empty list considered as "all"
+    }
```
Nokel81, line 60: Shouldn't we deal with all namespaces that are selected (though I don't know how you would click multiple)

Nokel81 (dismissed):

jakolehm:
@aleksfront pls resolve conflicts.

src/renderer/components/+namespaces/namespace-select-filter.tsx
aleksfront: @Nokel81 Thanks for describing the issue ^

src/renderer/components/+namespaces/namespace-select.tsx
```diff
@@ -35,7 +33,7 @@ export class NamespaceSelect extends React.Component<Props> {
     ]);
   }
 
-  @computed get options(): SelectOption[] {
+  @computed.struct get options(): SelectOption[] {
```
ixrock, line 36: > In practice, computed.struct is less useful than it sounds. Only use it if changes in the underlying observables can still lead to the same output. For example, if we were rounding the coordinates first, the rounded coordinates might be equal to the previously rounded coordinates even though the underlying values aren't.

from here: https://mobx.js.org/computeds.html#computed-struct

src/renderer/components/+namespaces/namespace-select-filter.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  render() {
+    return (
+      <NamespaceSelect
+        isMulti={true}
+        components={{ Placeholder }}
+        showAllNamespacesOption={true}
+        closeMenuOnSelect={false}
+        controlShouldRenderValue={false}
+        placeholder={""}
```
ixrock, line 71: Why it's needed?

ixrock (approved):

src/renderer/components/+namespaces/namespace-select-filter.tsx
ixrock: Currently it works as `namespaceStore.toggleAll(true)` I guess (but empty namespaces list we consider as "all")

jim-docker:
I observed this peculiar behaviour, but it may have more to do with the recent watch optimizations(?)


https://user-images.githubusercontent.com/40840436/107444204-25061c80-6b08-11eb-874f-76cfaa59262f.mov

Selecting many namespaces one at a time, occasionally the namespace dropdown menu disappears for a second then reappears. There is a corresponding error observed in the dev tools window

src/renderer/components/+namespaces/namespace-select-filter.tsx
aleksfront: As it turned out, changes of `placeholder` prop caused React Select to rerender (and "jumping" scroll happens). That's why I left placeholder untouched and used dedicated `Placeholder` component which can be freely rerendered wihout affecting dropdown itself.

Nokel81 (approved):

src/renderer/components/+namespaces/namespace-select-filter.tsx
vshakirova: Thank you, guys

src/renderer/components/+namespaces/namespace-select.tsx
aleksfront: We have exact case when namespace store produces exactly same list on every context change. We shouldn't rerender then.

PR #2105, opened 2021-02-09
Fix: export Dialog to extensions-api

ixrock:
_Intro:_

For some reasons generating type from `ReactDOM.createPortal` is failing and stops producing `dialog.d.ts`

### DIFF
--- a/webpack.extensions.ts
+++ b/webpack.extensions.ts
@@ -1,9 +1,9 @@
 
 import path from "path";
 import webpack from "webpack";
-import { sassCommonVars } from "./src/common/vars";
+import { sassCommonVars, isDevelopment } from "./src/common/vars";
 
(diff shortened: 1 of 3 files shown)


webpack.extensions.ts
```diff
@@ -22,6 +22,10 @@ export default function (): webpack.Configuration {
       // e.g. require('@k8slens/extensions')
       libraryTarget: "commonjs"
     },
+    cache: isDevelopment,
+    optimization: {
+      minimize: false, // speed up types compilation
```
ixrock, line 27: On my machine first time compilation was `x2` faster! (~30s vs 60s+)

Nokel81 (approved):

jakolehm (approved):

PR #2108, opened 2021-02-09
Remove unnecessary subscribeStores from cluster overview

jakolehm:
Same stores are already loaded in `app.tsx`.


Nokel81 (approved):

Nokel81:
Though shouldn't additional subscribes not matter? Especially since these will be disposed of before any dispose within `app.tsx`?

ixrock:
Should we also move subscribing to `namespaceStore` from `namespace-select.tsx` to `app.tsx`? (to avoid reconnects btw pages)

```ts
      kubeWatchApi.subscribeStores([namespaceStore], {
        preload: true,
      })
```

image

ixrock (approved):

PR #2112, opened 2021-02-10
Improve kube watch error handling

jakolehm:
- handles cases where kubernetes api goes offline
- handles cases where kubernetes api returns 200 with status message with code 410
- handles cases where kubernetes watch api is ended (http 200 without response)
- handles cases where computer resumes from sleep
- fallbacks to retry if state is unknown (like during network issue etc).

Follow-up for #2095

### DIFF
--- a/src/renderer/api/kube-api.ts
+++ b/src/renderer/api/kube-api.ts
@@ -8,7 +8,7 @@ import { apiManager } from "./api-manager";
 import { apiKube } from "./index";
 import { createKubeApiURL, parseKubeApi } from "./kube-api-parse";
 import { KubeJsonApi, KubeJsonApiData, KubeJsonApiDataList } from "./kube-json-api";
-import { IKubeObjectConstructor, KubeObject } from "./kube-object";
+import { IKubeObjectConstructor, KubeObject, KubeStatus } from "./kube-object";
 import byline from "byline";
 import { ReadableWebToNodeStream } from "readable-web-to-node-stream";
 import { IKubeWatchEvent } from "./kube-watch-api";
@@ -93,9 +93,11 @@ export function ensureObjectSelfLink(api: KubeApi, object: KubeJsonApiData) {
   }
 }
 
-type KubeApiWatchOptions = {
+export type KubeApiWatchCallback = (data: IKubeWatchEvent, error: any) => void;
+
+export type KubeApiWatchOptions = {
   namespace: string;
-  callback?: (data: IKubeWatchEvent) => void;
+  callback?: KubeApiWatchCallback;
   abortController?: AbortController
 };
 
@@ -370,8 +372,8 @@ export class KubeApi<T extends KubeObject = any> {
     if (!opts.abortController) {
       opts.abortController = new AbortController();
     }
+    let errorReceived = false;
     const { abortController, namespace, callback } = opts;
-
     const watchUrl = this.getWatchUrl(namespace);
     const responsePromise = this.request.getResponse(watchUrl, null, {
       signal: abortController.signal
@@ -379,53 +381,63 @@ export class KubeApi<T extends KubeObject = any> {
 
     responsePromise.then((response) => {
       if (!response.ok && !abortController.signal.aborted) {
-        if (response.status === 410) { // resourceVersion has gone
-          setTimeout(() => {
-            this.refreshResourceVersion().then(() => {
-              this.watch({...opts, abortController});
-            });
-          }, 1000);
-
-        } else if (response.status >= 500) { // k8s is having hard time
-          setTimeout(() => {
-            this.watch({...opts, abortController});
-          }, 5000);
+        if (callback) {
+          callback(null, response);
         }
 
         return;
       }
       const nodeStream = new ReadableWebToNodeStream(response.body);
+
+      nodeStream.on("end", () => {
+        if (errorReceived) return; // kubernetes errors should be handled in a callback
+
+        setTimeout(() => { // we did not get any kubernetes errors so let's retry
+          if (abortController.signal.aborted) return;
+
+          this.watch({...opts, namespace, callback});
+        }, 1000);
+      });
+
       const stream = byline(nodeStream);
 
       stream.on("data", (line) => {
         try {
           const event: IKubeWatchEvent = JSON.parse(line);
 
+          if (event.type === "ERROR" && event.object.kind === "Status") {
+            console.error(event);
+            errorReceived = true;
+            callback(null, new KubeStatus(event.object as any));
+
+            return;
+          }
+
           this.modifyWatchEvent(event);
 
           if (callback) {
-            callback(event);
+            callback(event, null);
           }
         } catch (ignore) {
           // ignore parse errors
         }
       });
-
-      stream.on("close", () => {
-        setTimeout(() => {
-          if (!abortController.signal.aborted) this.watch({...opts, namespace, callback});
-        }, 1000);
-      });
     }, (error) => {
       if (error instanceof DOMException) return; // AbortController rejects, we can ignore it
 
-      console.error("watch rejected", error);
+      if (callback) {
+        callback(null, error);
+      }
     }).catch((error) => {
-      console.error("watch error", error);
+      if (callback) {
+        callback(null, error);
+      }
     });
 
     const disposer = () => {
-      abortController.abort();
+      if (!abortController.signal.aborted) {
+        abortController.abort();
+      }
     };
 
     return disposer;
--- a/src/renderer/api/kube-api.ts
+++ b/src/renderer/api/kube-api.ts
@@ -406,7 +406,6 @@ export class KubeApi<T extends KubeObject = any> {
           const event: IKubeWatchEvent = JSON.parse(line);
 
           if (event.type === "ERROR" && event.object.kind === "Status") {
-            console.error(event);
             errorReceived = true;
             callback(null, new KubeStatus(event.object as any));
 
--- a/src/renderer/api/kube-api.ts
+++ b/src/renderer/api/kube-api.ts
@@ -381,9 +381,7 @@ export class KubeApi<T extends KubeObject = any> {
 
     responsePromise.then((response) => {
       if (!response.ok && !abortController.signal.aborted) {
-        if (callback) {
-          callback(null, response);
-        }
+        callback?.(null, response);
 
         return;
       }
@@ -424,19 +422,13 @@ export class KubeApi<T extends KubeObject = any> {
     }, (error) => {
       if (error instanceof DOMException) return; // AbortController rejects, we can ignore it
 
-      if (callback) {
-        callback(null, error);
-      }
+      callback?.(null, error);
     }).catch((error) => {
-      if (callback) {
-        callback(null, error);
-      }
+      callback?.(null, error);
     });
 
     const disposer = () => {
-      if (!abortController.signal.aborted) {
-        abortController.abort();
-      }
+      abortController.abort();
     };
 
     return disposer;
--- a/src/renderer/api/kube-object.ts
+++ b/src/renderer/api/kube-object.ts
@@ -40,6 +40,26 @@ export interface IKubeObjectMetadata {
   }[];
 }
 
+export interface IKubeStatus {
+  kind: string;
+  apiVersion: string;
+  code: number;
+  message?: string;
+  reason?: string;
+}
+
+export class KubeStatus implements IKubeStatus {
+  public readonly kind: string;
+  public readonly apiVersion: string;
+  public readonly code: number;
+  public readonly message: string;
+  public readonly reason: string;
+
+  constructor(data: IKubeStatus) {
+    Object.assign(this, data);
+  }
+}
+
 export type IKubeMetaField = keyof IKubeObjectMetadata;
 
 @autobind()
--- a/src/renderer/api/kube-object.ts
+++ b/src/renderer/api/kube-object.ts
@@ -48,8 +48,8 @@ export interface IKubeStatus {
   reason?: string;
 }
 
-export class KubeStatus implements IKubeStatus {
-  public readonly kind: string;
+export class KubeStatus {
+  public readonly kind = "Status";
   public readonly apiVersion: string;
   public readonly code: number;
   public readonly message: string;
--- a/src/renderer/api/kube-object.ts
+++ b/src/renderer/api/kube-object.ts
@@ -56,7 +56,10 @@ export class KubeStatus {
   public readonly reason: string;
 
   constructor(data: IKubeStatus) {
-    Object.assign(this, data);
+    this.apiVersion = data.apiVersion;
+    this.code = data.code;
+    this.message = data.message || "";
+    this.reason = data.reason || "";
   }
 }
 
--- a/src/renderer/api/kube-watch-api.ts
+++ b/src/renderer/api/kube-watch-api.ts
@@ -35,14 +35,6 @@ export class KubeWatchApi {
 
   contextReady = when(() => Boolean(this.context));
 
-  constructor() {
-    this.init();
-  }
-
-  private async init() {
-    await this.contextReady;
-  }
-
   isAllowedApi(api: KubeApi): boolean {
     return Boolean(this.context?.cluster.isAllowedResource(api.kind));
   }
--- a/src/renderer/api/kube-watch-api.ts
+++ b/src/renderer/api/kube-watch-api.ts
@@ -12,7 +12,7 @@ import { KubeJsonApiData } from "./kube-json-api";
 import { isDebugging, isProduction } from "../../common/vars";
 
 export interface IKubeWatchEvent<T = KubeJsonApiData> {
-  type: "ADDED" | "MODIFIED" | "DELETED";
+  type: "ADDED" | "MODIFIED" | "DELETED" | "ERROR";
   object?: T;
 }
 
@@ -32,8 +32,6 @@ export interface IKubeWatchLog {
 @autobind()
 export class KubeWatchApi {
   @observable context: ClusterContext = null;
-  @observable subscribers = observable.map<KubeApi, number>();
-  @observable isConnected = false;
 
   contextReady = when(() => Boolean(this.context));
 
--- a/src/renderer/kube-object.store.ts
+++ b/src/renderer/kube-object.store.ts
@@ -178,10 +178,10 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
 
     // update existing items
     if (!replace) {
-      const partialIds = partialItems.map(item => item.getId());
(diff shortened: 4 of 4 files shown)


src/renderer/api/kube-object.ts
```diff
@@ -40,6 +40,26 @@ export interface IKubeObjectMetadata {
   }[];
 }
 
+export interface IKubeStatus {
```
ixrock, line 43: Probably having these props definitions on the class should be enough (class could be used as a type as well)

src/renderer/api/kube-object.ts
```diff
@@ -40,6 +40,26 @@ export interface IKubeObjectMetadata {
...
+  kind: string;
+  apiVersion: string;
+  code: number;
+  message?: string;
+  reason?: string;
+}
+
+export class KubeStatus implements IKubeStatus {
+  public readonly kind: string;
```
ixrock:
```suggestion
  public readonly kind = "Status";
```

src/renderer/api/kube-object.ts
```diff
@@ -40,6 +40,26 @@ export interface IKubeObjectMetadata {
...
+  public readonly apiVersion: string;
+  public readonly code: number;
+  public readonly message: string;
+  public readonly reason: string;
+
+  constructor(data: IKubeStatus) {
```
ixrock, line 58: This will allow to avoid type check from raw data on constructing object.

```suggestion
  constructor(data: IKubeStatus | any) {
```

src/renderer/kube-object.store.ts
```diff
@@ -149,8 +149,10 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
       if (merge) {
         this.mergeItems(items, { replace: false });
       } else {
-        return items;
+        this.mergeItems(items, { replace: true });
```
ixrock, line 152: Why `replace: true`? 
This way random extension can screw up loaded items with `loadAll({namespaces: ["default"]})`?

src/renderer/api/kube-api.ts
```diff
@@ -370,62 +372,71 @@ export class KubeApi<T extends KubeObject = any> {
...
       const nodeStream = new ReadableWebToNodeStream(response.body);
+
+      nodeStream.on("end", () => {
+        if (errorReceived) return; // kubernetes errors should be handled in a callback
+
+        setTimeout(() => { // we did not get any kubernetes errors so let's retry
+          if (abortController.signal.aborted) return;
+
+          this.watch({...opts, namespace, callback});
+        }, 1000);
```
Nokel81, line 397:
```suggestion
        const rewatch = setTimeout(() => { // we did not get any kubernetes errors so let's retry
          this.watch({...opts, namespace, callback});
        }, 1000);
        
        abortController.addEventListener("abort", () => clearTimeout(rewatch));
```

src/renderer/kube-object.store.ts
```diff
@@ -267,31 +269,69 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
+    const callback = (data: IKubeWatchEvent, error: any) => {
+      if (!this.isLoaded || abortController.signal.aborted) return;
+
+      if (error instanceof Response) {
+        if (error.status === 404) {
+          // api has gone, let's not retry
+          return;
+        } else { // not sure what to do, best to retry
+          setTimeout(() => {
+            if (abortController.signal.aborted) return;
```
ixrock: Already used in line 293, so doesn't make sense.

```ts
if (!this.isLoaded || abortController.signal.aborted) return;
```

src/renderer/api/kube-api.ts
```diff
@@ -370,62 +372,71 @@ export class KubeApi<T extends KubeObject = any> {
     if (!opts.abortController) {
       opts.abortController = new AbortController();
     }
+    let errorReceived = false;
     const { abortController, namespace, callback } = opts;
```
Nokel81, line 376: Is there a reason why we are using a callback here? Does it make the API easier to use, vs a promise?

src/renderer/api/kube-api.ts
```diff
@@ -93,9 +93,11 @@ export function ensureObjectSelfLink(api: KubeApi, object: KubeJsonApiData) {
   }
 }
 
-type KubeApiWatchOptions = {
+export type KubeApiWatchCallback = (data: IKubeWatchEvent, error: any) => void;
```
ixrock, line 96:
```suggestion
export type KubeApiWatchCallback = (res: {data?: IKubeWatchEvent, error?: any}) => void;
```

src/renderer/api/kube-api.ts
```diff
@@ -370,62 +372,71 @@ export class KubeApi<T extends KubeObject = any> {
...
-      console.error("watch error", error);
+      if (callback) {
+        callback(null, error);
+      }
     });
 
     const disposer = () => {
-      abortController.abort();
+      if (!abortController.signal.aborted) {
+        abortController.abort();
```
Nokel81: There is no need to do this check as the abort controller does it internally

src/renderer/kube-object.store.ts
```diff
@@ -267,31 +269,69 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
       });
     }
 
-    return () => disposers.forEach(dispose => dispose());
+    return () => {
+      abortController.abort();
+    };
+  }
+
+  private watchNamespace(api: KubeApi<T>, namespace: string, abortController: AbortController) {
```
ixrock, line 290: It feels like it should be method of `KubeApi`.. Wdyt?

**UPD:** actually `KubeApi.watch()` support `namespace` in args. Maybe just rename to `KubeObjectStore.watch(namespaces?: string[])`?

src/renderer/api/kube-api.ts
```diff
@@ -370,62 +372,71 @@ export class KubeApi<T extends KubeObject = any> {
...
-          if (!abortController.signal.aborted) this.watch({...opts, namespace, callback});
-        }, 1000);
-      });
     }, (error) => {
       if (error instanceof DOMException) return; // AbortController rejects, we can ignore it
 
-      console.error("watch rejected", error);
+      if (callback) {
+        callback(null, error);
+      }
```
Nokel81: Personally I prefer the `callback?.(null, error)` but that is entirely a style preference.

src/renderer/api/kube-api.ts
ixrock: @Nokel81 I guess it's done to have `KubeApi.watch()` as `sync` method.

jakolehm: I went with callback because it's backward compatible (we are returning disposer functions).

src/renderer/api/kube-api.ts
jakolehm: Ok, will cleanup.

src/renderer/kube-object.store.ts
jakolehm: This happens somewhere in future and that's why it needs to be checked again, right?

src/renderer/api/kube-object.ts
Nokel81: I disagree that we should do this as it would decrease the benefit of typescript and typing in general. If you don't want to write the type checking specifically the a static function that takes `any` and does the type checking using functions in #2003.

src/renderer/api/kube-object.ts
```diff
@@ -40,6 +40,26 @@ export interface IKubeObjectMetadata {
...
+    Object.assign(this, data);
+  }
+}
+
```
Nokel81: What is the benefit of this class vs the interface above? also this contructor is incorrect as `message` and `reason` should either be optional or be defaulted to `""`

jakolehm: Class because then user can check `xyz instanceof KubeStatus` ... no other reason really.

src/renderer/kube-object.store.ts
Nokel81: Why not just clear the timeout?

src/renderer/api/kube-object.ts
Nokel81: Okay sounds good, though I still think that we shouldn't use `Object.assign` in constructors.

src/renderer/kube-object.store.ts
ixrock: No, because you get access to the same `abortController` from the scope above and if it was aborted the code below will be not reached? Or I'm missing something? 🤔

Nokel81: What I mean is why not (give line 300 having `const retry = setTimeout(...`) have the following:

```js
abortController.signal.addEventHandler("abort", () => clearTimeout(retry));
```

src/renderer/api/kube-api.ts
jakolehm: I'm not sure why this suggestion 🤔

src/renderer/kube-object.store.ts
jakolehm: @Nokel81 I like your suggestion, will cleanup.

src/renderer/api/kube-api.ts
Nokel81: I think it is so that you don't have to specify `null` when producing an error. Also is the standard order `(error: any, data: IKubeWatchEvent)`?

src/renderer/kube-object.store.ts
jakolehm: They can only screw up if they explicitly set `merge: false`, right? Maybe I don't understand what you originally meant with `merge` option. Anyway it's not obvious for a user of this function if `merge: false` does not update the store. 

That said: I can revert this change if needed but the watch api reload needs a way to force a fresh items array.

src/renderer/api/kube-api.ts
jakolehm: Added abort listener.

src/renderer/kube-object.store.ts
jakolehm: I think it needs to live within a store because it needs access to store internals. I chose private so that we don't add this to public extension api (we can move it later if needed).

src/renderer/api/kube-api.ts
jakolehm: Yes, fixed 👍

src/renderer/api/kube-object.ts
jakolehm: What should we use instead?

jakolehm:
@ixrock @Nokel81 ptal again.

src/renderer/api/kube-object.ts
Nokel81: Assigning directly,

```typescript
export class KubeStatus implements IKubeStatus {
  public readonly kind: string;
  public readonly apiVersion: string;
  public readonly code: number;
  public readonly message: string;
  public readonly reason: string;

  constructor(data: IKubeStatus) {
    this.kind = data.kind;
    this.apiVersion = data.apiVersion;
    this.code = data.code;
    this.message = data.message || "";
    this.reason = data.reason || "";
  }
}
```

We only have to write this code once, so I don't see why we should be copying over more than what we actually want.

jakolehm: Agreed and fixed.

src/renderer/api/kube-object.ts
```diff
@@ -56,7 +56,10 @@ export class KubeStatus {
   public readonly reason: string;
 
   constructor(data: IKubeStatus) {
-    Object.assign(this, data);
```
Nokel81: Should we check that `data.kind` is actually `Status` here and if not throw an error?

src/renderer/api/kube-watch-api.ts
```diff
@@ -32,8 +32,6 @@ export interface IKubeWatchLog {
 @autobind()
 export class KubeWatchApi {
   @observable context: ClusterContext = null;
-  @observable subscribers = observable.map<KubeApi, number>();
-  @observable isConnected = false;
 
   contextReady = when(() => Boolean(this.context));
```
Nokel81, line 37: Maybe this isn't for this PR, but what exactly is the constructor "supposed" to be doing here? But besides that, what is with the habit of calling async functions in a constructor?

src/renderer/kube-object.store.ts
ixrock: Yeah, right. Since by default `merge = true`

src/renderer/kube-object.store.ts
```diff
@@ -267,31 +269,68 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
+    let timedRetry: NodeJS.Timeout;
+
+    abortController.signal.addEventListener("abort", () => clearTimeout(timedRetry));
```
ixrock, line 293: Shouldn't we use nowadays something for unsubscribing  "abort" event?

```ts
    const onAbort = () => clearTimeout(timedRetry);
    abortController.signal.addEventListener("abort", onAbort);

    // ..and call this at some point? or it's automatically garbage-collected?
    abortController.signal.removeEventListener("abort", onAbort);
```

src/renderer/kube-object.store.ts
```diff
@@ -267,31 +269,68 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
+    const callback = (data: IKubeWatchEvent, error: any) => {
+      if (!this.isLoaded || abortController.signal.aborted) return;
+
+      if (error instanceof Response) {
+        if (error.status === 404) {
+          // api has gone, let's not retry
+          return;
+        } else { // not sure what to do, best to retry
+          timedRetry = setTimeout(() => {
+            api.watch({
```
ixrock, line 305: `api.watch()` returns a disposer for `unwatch: () => void` but it's not collected anywhere here?
yes, it could be aborted/disposed from above i guess, but it's not obvious

src/renderer/kube-object.store.ts
```diff
@@ -267,31 +269,68 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
+              namespace,
+              abortController,
+              callback
+            });
+          }, 5000);
+        }
+      } else if (error instanceof KubeStatus && error.code === 410) {
+        // resourceVersion has gone, let's try to reload
+        timedRetry = setTimeout(() => {
+          (namespace === "" ? this.loadAll({ merge: false }) : this.loadAll({namespaces: [namespace]})).then(() => {
```
ixrock, line 316: Would be this more descriptive?

```suggestion
          const loadAndReplaceAll = () => this.loadAll({ merge: false });
          const loadAndMerge = (namespace: string) => this.loadAll({namespaces: [namespace]}));
          (namespace ? loadAndMerge(namespace) : loadAndReplaceAll()).then(() => {
```

src/renderer/api/kube-api.ts
```diff
@@ -370,58 +372,59 @@ export class KubeApi<T extends KubeObject = any> {
...
+        setTimeout(() => { // we did not get any kubernetes errors so let's retry
+          if (abortController.signal.aborted) return;
+
+          this.watch({...opts, namespace, callback});
+        }, 1000);
+      });
+
       const stream = byline(nodeStream);
 
       stream.on("data", (line) => {
```
ixrock, line 402: Should we use `stream.off("data")` or it's auto-garbage collected?

src/renderer/kube-object.store.ts
jakolehm: IMHO that's not needed because abort happens only once (when disposer is called). And when that happens garbage collector should clean things.

src/renderer/kube-object.store.ts
jakolehm: First I collected all disposers but then I realised that it's duplicate work because we pass `abortController` reference to watches. When abort happens it will clean things.

src/renderer/api/kube-watch-api.ts
jakolehm: Removed, should have been removed already in #2095 .

nevalla (dismissed):
Tested and seems to recover nicely from various error scenarios.

ixrock (approved):

src/renderer/api/kube-api.ts
```diff
@@ -370,58 +372,59 @@ export class KubeApi<T extends KubeObject = any> {
...
       const nodeStream = new ReadableWebToNodeStream(response.body);
+
+      nodeStream.on("end", () => {
+        if (errorReceived) return; // kubernetes errors should be handled in a callback
+
+        setTimeout(() => { // we did not get any kubernetes errors so let's retry
+          if (abortController.signal.aborted) return;
+
+          this.watch({...opts, namespace, callback});
+        }, 1000);
```
Nokel81, line 397: You said that you would change this is clearing the timeout, but haven't

src/renderer/api/kube-object.ts
```diff
@@ -40,6 +40,29 @@ export interface IKubeObjectMetadata {
...
+}
+
+export class KubeStatus {
+  public readonly kind = "Status";
+  public readonly apiVersion: string;
+  public readonly code: number;
+  public readonly message: string;
+  public readonly reason: string;
+
+  constructor(data: IKubeStatus) {
```
Nokel81, line 58: This constructor should verify `data.kind` is actually `"Status"`. As this is a good point to verify our assumptions about the data being passed in.

src/renderer/kube-object.store.ts
```diff
@@ -267,31 +269,80 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
+            api.watch({
+              namespace,
+              abortController,
+              callback
+            });
+          });
+        }, 1000);
+      } else if(error) { // not sure what to do, best to retry
+        if (timedRetry) clearTimeout(timedRetry);
```
Nokel81, line 325: No need for this `if` as `clearTimeout(...) ignores invalid values.

nevalla:
Now after the sleep only those watches that have not received any data were retried:

image

src/renderer/api/kube-api.ts
ixrock: In general I would propose always to use first argument as plain object.

_Benefits:_ 

- allows to pass args in any order
- better extensibility when overriding a method
- easier to collect and pass bunch of params with object's destruct/rest, e.g. `func({param = "default"} = {})`

jakolehm:
> Now after the sleep only those watches that have not received any data were retried:

@nevalla do you mean that this PR somehow handles those worse vs master branch?

src/renderer/api/kube-api.ts
jakolehm: I changed those within `kube-object.store.ts`, not here. I would prefer to cleanup in a separate PR so that we can get these fixes merged sooner.

nevalla (approved):

PR #2113, opened 2021-02-10
Fix: compact-mode for Workloads overview -> Events

ixrock:
- fix: compact-mode for `<Events/>` page component in workloads overview (`limit = 10` by default)
- fix: type of event previously has shown as `involvedObject.kind` (moved to `Involved Object` column)
- show `warning` items on the top (default sorting: `event.type:desc`)

image

close #2109

PR #2114, opened 2021-02-10
fix problem with compact events

### LINKED ISSUE #2109: Workload overview event list is not compact
I'm not sure if this is actually a bug in `Event` component but 4.1.0-beta.2 shows all events from eventStore which makes this page really heavy if store has lots of items.

jim-docker:
fixes underlying problem with `ItemListLayout` and the `filterItems` prop.  May not be the best way, or there are conditions on using `filterItems` in conjunction with `items`?

This PR could be merged instead of #2113, or as well as, since #2113 doesn't fix the underlying problem.

fixes #2109

### DIFF
--- a/src/renderer/components/item-object-list/item-list-layout.tsx
+++ b/src/renderer/components/item-object-list/item-list-layout.tsx
@@ -215,7 +215,7 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
       }
     });
 
-    const items = this.props.items ?? allItems;
+    const items = this.applyFilters(this.props.filterItems, this.props.items) ?? allItems;
 
     return this.applyFilters(filterItems, items);
   }


src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -215,7 +215,7 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
       }
     });
 
-    const items = this.props.items ?? allItems;
```
jim-docker, line 218: the only time `this.props.filterItems` is applied is in the `allItems` getter, which doesn't get applied here if `this.props.items` is set.

aleksfront:
I'm not sure we should mix `this.props.items` and `allItems` together. I suggest following @ixrock path by keeping single source of truth and removing any usage of `filterItems` prop from `ItemListLayout`.

See #2123

jim-docker:
> I'm not sure we should mix `this.props.items` and `allItems` together. I suggest following @ixrock path by keeping single source of truth and removing any usage of `filterItems` prop from `ItemListLayout`.
> 
> See #2123

Sounds good, the `filterItems` prop looked to be a bit of a hack anyway. The one advantage of this PR over #2113 is that it touches only a single line of code, less likely to introduce side effects just as we are about to make a final release...

PR #2121, opened 2021-02-11
Don't show warning badge if cluster is offline

jakolehm:
Because it's stale/misleading info if a cluster is not connected.


nevalla (approved):
LGTM

aleksfront (approved):

PR #2123, opened 2021-02-11
Removing "filterItems" prop from ItemListLayout

aleksfront:
This PR follows this one https://github.com/lensapp/lens/pull/2113 and removes all usage of `filterItems` prop described in `ItemListComponent`.

This allows to remove dualism from choosing items to present inside `ItemListLayout` component here https://github.com/lensapp/lens/pull/2114 and moves filtering responsibility to specified components.

### DIFF
--- a/src/renderer/components/+custom-resources/crd-list.tsx
+++ b/src/renderer/components/+custom-resources/crd-list.tsx
@@ -33,6 +33,15 @@ export class CrdList extends React.Component {
     return crdGroupsUrlParam.get();
   }
 
+  get items() {
+    const selectedGroups = this.groups;
+    const storeItems = crdStore.items;
+
+    return selectedGroups.length ?
+      storeItems.filter(item => selectedGroups.includes(item.getGroup())) :
+      storeItems;
+  }
+
   onSelectGroup(group: string) {
     const groups = new Set(this.groups);
 
@@ -62,11 +71,7 @@ export class CrdList extends React.Component {
         store={crdStore}
         sortingCallbacks={sortingCallbacks}
         searchFilters={Object.values(sortingCallbacks)}
-        filterItems={[
-          (items: CustomResourceDefinition[]) => {
-            return selectedGroups.length ? items.filter(item => selectedGroups.includes(item.getGroup())) : items;
-          }
-        ]}
+        items={this.items}
         renderHeaderTitle="Custom Resources"
         customizeHeader={() => {
           let placeholder = <>All groups</>;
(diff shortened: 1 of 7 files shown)


src/renderer/components/+custom-resources/crd-list.tsx
```diff
@@ -33,6 +33,15 @@ export class CrdList extends React.Component {
     return crdGroupsUrlParam.get();
   }
 
+  get items() {
+    const selectedGroups = this.groups;
+    const storeItems = crdStore.contextItems;
+
+    return selectedGroups.length ?
+      storeItems.filter(item => selectedGroups.includes(item.getGroup())) :
+      storeItems;
```
Nokel81, line 42: I find having the `?` and `:` at the front of the line easier to read in passing.
```suggestion
    return selectedGroups.length
      ? storeItems.filter(item => selectedGroups.includes(item.getGroup()))
      : storeItems;
```

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -45,7 +45,6 @@ export interface ItemListLayoutProps<T extends ItemObject = ItemObject> {
   isClusterScoped?: boolean;
   hideFilters?: boolean;
   searchFilters?: SearchFilter<T>[];
-  filterItems?: ItemsFilter<T>[];
```
Nokel81, line 48: Are you sure that this is not a breaking change. Shouldn't we mark it as depcricated (or is `ItemListLayout` not exported in 4.0.X?)

ixrock: Yes, it is breaking change.. Initially I thought it's not, but:

`export interface KubeObjectListLayoutProps extends ItemListLayoutProps`
and we re-export `KubeObjectListLayout` to extensions-api..

So, even if we don't export `ItemListLayout` directly it's `Props`-interface leaking to `KubeObjectList`.

ixrock:
@aleksfront please merge `master` since there are some unrelated to PR changes in diff.

jim-docker:
Unless this fixes a blocking issue I think this should be merged after the 4.1 release...

src/renderer/components/item-object-list/item-list-layout.tsx
Nokel81: So furthermore we cannot remove this functionality at all because that is a breaking change.

Nokel81 (changes requested):
I am in favour of removing our uses of `filterItems` but we either have to add it to `KubeObjectListLayout` and mirror the functionality there. Or just mark it as `@deprecated`.

jim-docker:
> I am in favour of removing our uses of `filterItems` but we either have to add it to `KubeObjectListLayout` and mirror the functionality there. Or just mark it as `@deprecated`.

I agree, let's remove our uses and deprecate it

aleksfront:
Closing this in favour of new table component that will be introduced in `5.x.x`.

PR #2127, opened 2021-02-11
Fix: incorrectly filtered events with compact=true

ixrock:
Since #2113 we're passing items to `KubeObjectList` from above and the items must be sorted in advance to work with `compact=true`

_Before:_
image

_After:_
image

_Events page:_
image

### DIFF
--- a/src/renderer/components/+events/events.tsx
+++ b/src/renderer/components/+events/events.tsx
@@ -1,9 +1,11 @@
 import "./events.scss";
 
 import React, { Fragment } from "react";
+import { computed } from "mobx";
 import { observer } from "mobx-react";
+import { orderBy } from "lodash";
 import { TabLayout } from "../layout/tab-layout";
-import { eventStore } from "./event.store";
+import { EventStore, eventStore } from "./event.store";
 import { getDetailsUrl, KubeObjectListLayout, KubeObjectListLayoutProps } from "../kube-object";
 import { KubeEvent } from "../../api/endpoints/events.api";
 import { Tooltip } from "../tooltip";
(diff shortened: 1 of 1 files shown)


Nokel81 (commented):
This is moving quite fast for me, what is "compact" mode supposed to do? If it is merely displaying the entries in a compact mode. What does that have to do with sorting. I think this is the wrong direction to take.

ixrock:
> This is moving quite fast for me, what is "compact" mode supposed to do? If it is merely displaying the entries in a compact mode. What does that have to do with sorting. I think this is the wrong direction to take.

"Compact-mode" is used only in Workloads overview..  Please see for more details #2113 #2123
This is done because items sorting is handled inside `ItemsListLayout` currently (yes, this whole component bloated and needs refactoring, but this is another story).

src/renderer/components/+events/events.tsx
```diff
@@ -37,12 +40,20 @@ const defaultProps: Partial<Props> = {
 export class Events extends React.Component<Props> {
   static defaultProps = defaultProps as object;
 
-  get store() {
+  private defaultSorting: TableSortParams = {
+    sortBy: columnId.type,
+    orderBy: "desc", // show "Warning" events at the top
```
jim-docker, line 45: Sorting by type here corresponds to what's done on the Events page but subsequent sorting via the UI doesn't work the same. For age sorting on the workloads overview page it only sorts by age the 10 events that were first sorted by type.

jim-docker (approved):

src/renderer/components/+events/events.tsx
```diff
@@ -37,12 +40,20 @@ const defaultProps: Partial<Props> = {
 export class Events extends React.Component<Props> {
   static defaultProps = defaultProps as object;
 
-  get store() {
+  private defaultSorting: TableSortParams = {
+    sortBy: columnId.type,
```
jakolehm, line 44: Missed this in #2113 but shouldn't we sort events by "age"?

aleksfront (approved):

ixrock:
closing in favour #2141

PR #2128, opened 2021-02-11
Fix auto update on quit with newer version

Nokel81:
- If the user specifies that Lens should auto update on quit to a specific version, and before they quit the auto-updater finds a newer version. Then disregard the previous request to update

### DIFF
--- a/src/main/app-updater.ts
+++ b/src/main/app-updater.ts
@@ -54,11 +54,10 @@ export function startUpdateChecking(interval = 1000 * 60 * 60 * 24): void {
        * didn't ask for.
        */
       autoUpdater.autoInstallOnAppQuit = false;
+      installVersion = args.version;
 
       try {
         const backchannel = `auto-update:${args.version}`;
-
-        installVersion = args.version;
 
         ipcMain.removeAllListeners(backchannel); // only one handler should be present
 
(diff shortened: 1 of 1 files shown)


src/main/app-updater.ts
```diff
@@ -38,9 +40,26 @@ export function startUpdateChecking(interval = 1000 * 60 * 60 * 24): void {
...
+       * This should be always set to false here because it is the reasonable
+       * default. Namely, if a don't auto update to a version that the user
+       * didn't ask for.
+       */
+      autoUpdater.autoInstallOnAppQuit = false;
+
       try {
         const backchannel = `auto-update:${args.version}`;
 
+        installVersion = args.version;
```
jim-docker: Should `installVersion` be reset in the catch block below?

jim-docker (dismissed):

src/main/app-updater.ts
Nokel81: Good point it probably should be

jakolehm (approved):

jim-docker (approved):
hard to test this. I did notice this, which is probably not related to this PR:
Screen Shot 2021-02-22 at 4 33 26 PM
This likely isn't encountered very often since the check interval is 24hr, but probably shouldn't happen?

Nokel81:
Thanks, will fix that with a follow up PR, but yes very unlikely for that to happen in "real life".

PR #2130, opened 2021-02-11
Fix: loading in namespaces under reduced RBAC

### LINKED ISSUE #2111: Accessible namespaces functionality is broken in 4.1.0-beta2
**Describe the bug**
Accessible namespaces functionality is broken in 4.1.0-beta2
**To Reproduce**
1. Add a new cluster and configure Accessible Namespaces
2. Click on cluster

```
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {
    
  },
  "status": "Failure",
  "message": "namespaces is forbidden: User \"https://sts.windows.net/fa7b1b5a-7b34-4387-94ae-d2c178decee1/#nqPjN1BDJSi4LUEkIYa8fR1C1-mZpoOiACmZXDYMKjE\" cannot list resource \"namespaces\" in API group \"\" at the cluster scope",
  "reason": "Forbidden",
  "details": {
    "kind": "namespaces"
  },
  "code": 403
}
```

Nokel81:
- don't assume that loading "all namespaces" that lens knows about is the same as loading "all namespaces" that exist on the cluster

- Track when actually listing namespaces

Fixes #2111

### DIFF
--- a/src/main/cluster.ts
+++ b/src/main/cluster.ts
@@ -49,6 +49,7 @@ export interface ClusterState {
   allowedNamespaces: string[]
   allowedResources: string[]
   isGlobalWatchEnabled: boolean;
+  listedNamespaces: boolean;
 }
 
 /**
@@ -201,6 +202,15 @@ export class Cluster implements ClusterModel, ClusterState {
    * @observable
    */
   @observable allowedNamespaces: string[] = [];
+
+  /**
+   * Is true if `allowedNamespaces` was filled by a successful kubeAPI list
+   * namespaces request.
+   *
+   * @observable
+   */
+  @observable listedNamespaces = false;
+
   /**
    * List of allowed resources
    *
@@ -630,6 +640,7 @@ export class Cluster implements ClusterModel, ClusterState {
       allowedNamespaces: this.allowedNamespaces,
       allowedResources: this.allowedResources,
       isGlobalWatchEnabled: this.isGlobalWatchEnabled,
+      listedNamespaces: this.listedNamespaces,
     };
 
     return toJS(state, {
@@ -669,6 +680,8 @@ export class Cluster implements ClusterModel, ClusterState {
 
   protected async getAllowedNamespaces() {
     if (this.accessibleNamespaces.length) {
+      this.listedNamespaces = false;
+
       return this.accessibleNamespaces;
     }
 
@@ -677,8 +690,11 @@ export class Cluster implements ClusterModel, ClusterState {
     try {
       const namespaceList = await api.listNamespace();
 
+      this.listedNamespaces = true;
+
       return namespaceList.body.items.map(ns => ns.metadata.name);
     } catch (error) {
+      this.listedNamespaces = false;
       const ctx = this.getProxyKubeconfig().getContextObject(this.contextName);
 
       if (ctx.namespace) return [ctx.namespace];
--- a/src/renderer/api/kube-watch-api.ts
+++ b/src/renderer/api/kube-watch-api.ts
@@ -78,11 +78,11 @@ export class KubeWatchApi {
       }
     };
 
-    let subscribeP: Promise<void>;
+    let subscribing: Promise<void> = Promise.resolve();
 
     if (preloading) {
       if (waitUntilLoaded) {
-        subscribeP = preloading.loading
+        subscribing = preloading.loading
           .then(() => subscribe(ac.signal))
           .catch(error => {
             this.log({
@@ -91,7 +91,7 @@ export class KubeWatchApi {
             });
           });
       } else {
-        subscribeP = subscribe(ac.signal);
+        subscribing = subscribe(ac.signal);
       }
 
       // reload stores only for context namespaces change
@@ -100,14 +100,14 @@ export class KubeWatchApi {
           preloading.cancelLoading();
         }
         ac.abort();
-        subscribeP.then(() => {
+        subscribing.then(() => {
           unsubscribeList.forEach(unsubscribe => unsubscribe());
           unsubscribeList.length = 0;
 
           ac = new AbortController();
           preloading = load(namespaces);
           preloading.loading
-            .then(() => subscribeP = subscribe(ac.signal));
+            .then(() => subscribing = subscribe(ac.signal));
         });
       }, {
         equals: comparer.shallow,
@@ -124,7 +124,7 @@ export class KubeWatchApi {
         preloading.cancelLoading();
       }
       ac.abort();
-      subscribeP.then(() => {
+      subscribing.then(() => {
         unsubscribeList.forEach(unsubscribe => unsubscribe());
         unsubscribeList.length = 0;
       });
--- a/src/renderer/api/kube-watch-api.ts
+++ b/src/renderer/api/kube-watch-api.ts
@@ -64,36 +64,51 @@ export class KubeWatchApi {
     let isUnsubscribed = false;
 
     const load = (namespaces = subscribingNamespaces) => this.preloadStores(stores, { namespaces, loadOnce });
-    let preloading = preload && load();
+    let preloading: boolean | ReturnType<typeof load> = preload && load();
     let cancelReloading: IReactionDisposer = noop;
+    let ac = new AbortController();
 
-    const subscribe = () => {
-      if (isUnsubscribed) return;
+    const subscribe = async (signal: AbortSignal) => {
+      if (isUnsubscribed || signal.aborted) return;
 
-      stores.forEach((store) => {
-        unsubscribeList.push(store.subscribe());
-      });
+      for (const store of stores) {
+        if (!signal.aborted) {
+          unsubscribeList.push(await store.subscribe());
+        }
+      }
     };
+
+    let subscribeP: Promise<void>;
 
     if (preloading) {
       if (waitUntilLoaded) {
-        preloading.loading.then(subscribe, error => {
-          this.log({
-            message: new Error("Loading stores has failed"),
-            meta: { stores, error, options: opts },
+        subscribeP = preloading.loading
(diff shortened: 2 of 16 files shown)


src/renderer/api/kube-watch-api.ts
```diff
@@ -64,36 +64,51 @@ export class KubeWatchApi {
     let isUnsubscribed = false;
 
     const load = (namespaces = subscribingNamespaces) => this.preloadStores(stores, { namespaces, loadOnce });
-    let preloading = preload && load();
+    let preloading: boolean | ReturnType<typeof load> = preload && load();
```
Nokel81, line 67: This is the correct typing of this value. I assume that because we aren't using `strict: true` it was typed as only `ReturnType<typeof load>`

src/renderer/api/kube-watch-api.ts
```diff
@@ -64,36 +64,51 @@ export class KubeWatchApi {
...
+      if (isUnsubscribed || signal.aborted) return;
 
-      stores.forEach((store) => {
-        unsubscribeList.push(store.subscribe());
-      });
+      for (const store of stores) {
+        if (!signal.aborted) {
+          unsubscribeList.push(await store.subscribe());
+        }
+      }
```
Nokel81, line 78: fail quickly if we shouldn't load the stores

src/renderer/api/kube-watch-api.ts
```diff
@@ -64,36 +64,51 @@ export class KubeWatchApi {
...
-        preloading = load(namespaces);
-        preloading.loading.then(subscribe);
+      cancelReloading = reaction(() => this.context?.selectedNamespaces, namespaces => {
+        if (typeof preloading === "object") {
+          preloading.cancelLoading();
+        }
+        ac.abort();
+        subscribeP.then(() => {
+          unsubscribeList.forEach(unsubscribe => unsubscribe());
+          unsubscribeList.length = 0;
```
Nokel81, line 105: This list is only "full" after `subscribeP` has "finished"

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -108,9 +112,11 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
     return namespaces;
   }
 
-  getSubscribeApis() {
+  async getSubscribeApis() {
+    await this.contextReady;
+
     // if user has given static list of namespaces let's not start watches because watch adds stuff that's not wanted
-    if (this.context?.cluster.accessibleNamespaces.length > 0) {
+    if (this.context.cluster.accessibleNamespaces.length > 0) {
```
Nokel81, line 115: For this to be actually correct we need to always wait for the context to be ready. Otherwise we would return an API if this function was called before the context was ready.

src/renderer/components/context.ts
```diff
@@ -5,7 +5,7 @@ import { namespaceStore } from "./+namespaces/namespace.store";
 export interface ClusterContext {
   cluster?: Cluster;
   allNamespaces?: string[]; // available / allowed namespaces from cluster.ts
-  contextNamespaces?: string[]; // selected by user (see: namespace-select.tsx)
+  selectedNamespaces?: string[]; // selected by user (see: namespace-select.tsx)
```
Nokel81, line 8: I can revert this change if wanted. But `contextNamespaces` is a very not descriptive name IMO.

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -108,9 +112,11 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
     return namespaces;
   }
 
-  getSubscribeApis() {
+  async getSubscribeApis() {
```
ixrock, line 111: Why it's became `async`?

src/main/cluster.ts
```diff
@@ -201,6 +202,15 @@ export class Cluster implements ClusterModel, ClusterState {
...
    */
   @observable allowedNamespaces: string[] = [];
+
+  /**
+   * Is true if `allowedNamespaces` was filled by a successful kubeAPI list
+   * namespaces request.
+   *
+   * @observable
+   */
+  @observable listedNamespaces = false;
```
ixrock, line 212: Why do we need this when we have already `cluster.accessibleNamespaces` and `cluster.allowedNamespaces`?
If `accessibleNamespaces` is provided by user it must be used first.

Otherwise maybe better to use this as I understand it correct it serves same purpose?

```ts
  @computed get listedNamespaces(): boolean {
    return !this.accessibleNamespaces.length;
  }
```

src/renderer/api/kube-watch-api.ts
```diff
@@ -64,36 +64,51 @@ export class KubeWatchApi {
...
-        unsubscribeList.push(store.subscribe());
-      });
+      for (const store of stores) {
+        if (!signal.aborted) {
+          unsubscribeList.push(await store.subscribe());
+        }
+      }
     };
 
+    let subscribeP: Promise<void>;
```
ixrock:
```suggestion
    let subscribePromise: Promise<void>;
    // OR
    let subscribing: Promise<void>;
```

src/renderer/api/kube-watch-api.ts
```diff
@@ -64,36 +64,51 @@ export class KubeWatchApi {
...
-      cancelReloading = reaction(() => this.context?.contextNamespaces, namespaces => {
-        preloading?.cancelLoading();
-        unsubscribeList.forEach(unsubscribe => unsubscribe());
-        unsubscribeList.length = 0;
-        preloading = load(namespaces);
-        preloading.loading.then(subscribe);
+      cancelReloading = reaction(() => this.context?.selectedNamespaces, namespaces => {
+        if (typeof preloading === "object") {
+          preloading.cancelLoading();
+        }
```
ixrock, line 101:
```suggestion
        preloading?.cancelLoading();
```

src/renderer/kube-object.store.ts
```diff
@@ -264,11 +264,15 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
     });
   }
 
-  getSubscribeApis(): KubeApi[] {
+  async getSubscribeApis(): Promise<KubeApi[]> {
     return [this.api];
   }
 
-  subscribe(apis = this.getSubscribeApis()) {
+  async subscribe(apis?: KubeApi[]): Promise<Disposer> {
```
ixrock, line 271: Why moving to `async`?

src/renderer/utils/index.ts
```diff
@@ -2,6 +2,8 @@
 
 export const isElectron = !!navigator.userAgent.match(/Electron/);
 
+export type Disposer = () => void;
```
ixrock: Maybe better to create `utils/common-types.ts`?

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -65,7 +65,11 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
   }
 
   private autoLoadAllowedNamespaces(): IReactionDisposer {
-    return reaction(() => this.allowedNamespaces, namespaces => this.loadAll({ namespaces }), {
+    return reaction(() => this.allowedNamespaces, namespaces => {
+      console.log("autoLoadAllowedNamespaces", namespaces);
```
ixrock: Forgotten `console.log`

ixrock (dismissed):

src/main/cluster.ts
Nokel81: Because `allowedNamespaces` might also be from the context. See line 700

src/renderer/components/+namespaces/namespace.store.ts
Nokel81: Because we need to wait for `contextReady`

src/renderer/api/kube-watch-api.ts
Nokel81: See my note about the actually typing of `preloading` and why the old version was actually wrong.

src/renderer/kube-object.store.ts
Nokel81: See my note

src/renderer/utils/index.ts
Nokel81: sure, good idea

src/renderer/components/+namespaces/namespace.store.ts
Nokel81: yes thanks

src/main/cluster.ts
ixrock: I tought `allowedNamespaces` _must_ be _only_ from the context? No?

Nokel81: Or by listing via the kube api ( which I believe is the most common)

ixrock: Also there are some inconsistency between 2 methods imho:

```ts
  protected async getAllowedNamespaces() { // get allowed?
    if (this.accessibleNamespaces.length) {
      return this.accessibleNamespaces; // No, get accessible!
    }

  protected async getAllowedResources() {
    try {
      if (!this.allowedNamespaces.length) { // why not checking accessible as well?
        return [];
      }
```

Nokel81: Agreed, but if `getAllowedResources` is only ever called after `getAllowedNamespaces` it "just works"

Nokel81: There are *lots* of things I would like to change to `cluster.ts` and the above it part of it but I don't think that it is in scope for this PR.

jakolehm: I don't really like the idea of yet-another-public-observable because it will go to the extension API and is really hard to take away later. Any possibility that we could resolve this from the existing information (as we did before)?

PR #2132, opened 2021-02-11
Fix(rbac): pdp should have policy group

meier-christoph:
For rbac the `PodDisruptionBudget` should use the `policy` group.


Nokel81 (approved):
According to this document this is correct:

https://schoolofdevops.github.io/ultimate-kubernetes-bootcamp/configuring_authentication_and_authorization/

jakolehm (approved):

PR #2136, opened 2021-02-12
Handle suspend/resume error from watch stream read

### LINKED ISSUE #2133: Kubernetes watch does not always retry after resume from sleep
**Describe the bug**

Watch that already has received data does not retry if computer goes to sleep for a moment and resumes so fast that network does not go offline during the sleep. 

Note: retry works if watch didn't get any data before computer goes to sleep.

jakolehm:
Had to copy `ReadableWebToNodeStream` to here because we cannot catch read errors with the upstream version. We should probably contribute a fix to upstream.

Fixes #2133

### DIFF
--- a/src/renderer/api/kube-api.ts
+++ b/src/renderer/api/kube-api.ts
@@ -10,8 +10,8 @@ import { createKubeApiURL, parseKubeApi } from "./kube-api-parse";
 import { KubeJsonApi, KubeJsonApiData, KubeJsonApiDataList } from "./kube-json-api";
 import { IKubeObjectConstructor, KubeObject, KubeStatus } from "./kube-object";
 import byline from "byline";
-import { ReadableWebToNodeStream } from "readable-web-to-node-stream";
 import { IKubeWatchEvent } from "./kube-watch-api";
+import { ReadableWebToNodeStream } from "../utils/readableStream";
 
 export interface IKubeApiOptions<T extends KubeObject> {
   /**
@@ -373,7 +373,13 @@ export class KubeApi<T extends KubeObject = any> {
       opts.abortController = new AbortController();
     }
     let errorReceived = false;
+    let timedRetry: NodeJS.Timeout;
     const { abortController, namespace, callback } = opts;
+
+    abortController.signal.addEventListener("abort", () => {
+      clearTimeout(timedRetry);
+    });
+
     const watchUrl = this.getWatchUrl(namespace);
     const responsePromise = this.request.getResponse(watchUrl, null, {
       signal: abortController.signal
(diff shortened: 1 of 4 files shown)


nevalla (approved):
LGTM

src/renderer/api/kube-api.ts
```diff
@@ -387,14 +393,17 @@ export class KubeApi<T extends KubeObject = any> {
...
-        if (errorReceived) return; // kubernetes errors should be handled in a callback
+      ["end", "close", "error"].forEach((eventName) => {
+        nodeStream.on(eventName, () => {
+          if (errorReceived) return; // kubernetes errors should be handled in a callback
 
-        setTimeout(() => { // we did not get any kubernetes errors so let's retry
-          if (abortController.signal.aborted) return;
+          clearTimeout(timedRetry);
+          timedRetry = setTimeout(() => { // we did not get any kubernetes errors so let's retry
+            if (abortController.signal.aborted) return;
```
Nokel81, line 402: Why do we need this give lines 379-381 above, if the signal has already aborted, shouldn't the watch just stop anyway?

src/renderer/utils/readableStream.ts
```diff
@@ -0,0 +1,87 @@
...
+
+      if (data.done || this.released) {
+        this.push(null); // Signal EOF
+      } else {
+        this.bytesRead += data.value.length;
+        this.push(data.value); // Push new data to the queue
+      }
+    } catch(error) {
+      this.push(null); // Signal EOF
+    }
```
Nokel81, line 62: This is the change? Why do we want to signal `EOF` on an error? Shouldn't we log these errors anyway?

src/renderer/utils/readableStream.ts
```diff
@@ -0,0 +1,87 @@
+import { Readable } from "readable-stream";
```
ixrock, line 1:
```suggestion
import { Readable } from "stream";
```

> This package is a mirror of the streams implementations in Node.js.
>
> If you want to guarantee a stable streams base, regardless of what version of Node you, or the users of your libraries are using, use readable-stream only and avoid the "stream" module in Node-core, for background see this blogpost.

Since we bundle app with Node.js `v12.x` maybe we could use native `stream` module?
https://nodejs.org/dist/latest-v12.x/docs/api/stream.html#stream_readable_streams

ixrock (approved):

src/renderer/api/kube-api.ts
```diff
@@ -387,14 +393,17 @@ export class KubeApi<T extends KubeObject = any> {
       }
       const nodeStream = new ReadableWebToNodeStream(response.body);
 
-      nodeStream.on("end", () => {
-        if (errorReceived) return; // kubernetes errors should be handled in a callback
+      ["end", "close", "error"].forEach((eventName) => {
+        nodeStream.on(eventName, () => {
+          if (errorReceived) return; // kubernetes errors should be handled in a callback
```
Nokel81, line 398: I think this comment is wrong. Shouldn't it read something like:

```
// kubernetes error status events should prevent auto retrying of the watch.
```

src/renderer/api/kube-api.ts
jakolehm: You are right, we don't need this.

src/renderer/utils/readableStream.ts
jakolehm: We cannot even catch errors that are thrown from here. If we could I would not have copied this library to our codebase.

All other error cases end up with EOF so it makes sense to push this error to the similar path. I really would like to get this fix into upstream and that means this class itself won't be logging anything.

jakolehm:
@ixrock @Nokel81 I'll open a separate PR based on suggestions.

src/renderer/utils/readableStream.ts
jakolehm: Makes sense, I'll do that in a separate PR.

PR #2141, opened 2021-02-12
Fix: events sorting with compact=true is broken

ixrock:
**Fixes:**

- _Proper default sorting by age (Last seen / `event.creationTimestamp:asc`) as a **number**:_

image

- _Proper sorting of limited items with `compact=true` at **Workloads -> Overview** (sorted by `event.type:desc`):_

Screenshot 2021-02-13 at 0 09 12


src/renderer/components/+events/event.store.ts
```diff
@@ -20,8 +20,8 @@ export class EventStore extends KubeObjectStore<KubeEvent> {
 
   protected sortItems(items: KubeEvent[]) {
     return super.sortItems(items, [
-      event => event.metadata.creationTimestamp
```
ixrock, line 23: This was broken since `creationTimestamp` is a date-`string`

src/renderer/api/kube-object.ts
```diff
@@ -122,12 +122,15 @@ export class KubeObject implements ItemObject {
...
 
-  // todo: refactor with named arguments
+  getTimeDiffFromNow(): number {
+    return new Date().getTime() - new Date(this.metadata.creationTimestamp).getTime();
+  }
+
   getAge(humanize = true, compact = true, fromNow = false) {
     if (fromNow) {
-      return moment(this.metadata.creationTimestamp).fromNow();
+      return moment(this.metadata.creationTimestamp).fromNow(); // "string"
```
Nokel81: Why do we need this comment, is `getAge` supposed to return only `number` or `number | string`. Wouldn't it be better to explicitly mark the return type?

src/renderer/components/+events/events.tsx
```diff
@@ -37,23 +40,60 @@ const defaultProps: Partial<Props> = {
...
   static defaultProps = defaultProps as object;
 
-  get store() {
+  private sortingCallbacks: TableSortCallbacks = {
+    [columnId.namespace]: (event: KubeEvent) => event.getNs(),
+    [columnId.type]: (event: KubeEvent) => event.type,
+    [columnId.object]: (event: KubeEvent) => event.involvedObject.name,
+    [columnId.count]: (event: KubeEvent) => event.count,
+    [columnId.age]: (event: KubeEvent) => event.getTimeDiffFromNow(),
+  };
```
Nokel81, line 55: We are supposed to be only fixing stuff. Please back out all refactors.

Nokel81 (changes requested):

src/renderer/components/+events/events.scss
```diff
@@ -1,4 +1,10 @@
 .Events {
+  h5.title {
+    .events-count {
+      font-size: $font-size;
+    }
+  }
+
```
aleksfront: No need to change font size. It's better to move `events-count` out from `<h5>`
events table

src/renderer/components/+events/event.store.ts
```diff
@@ -20,8 +20,8 @@ export class EventStore extends KubeObjectStore<KubeEvent> {
...
-    ], "desc");
+      event => event.getTimeDiffFromNow(), // keep events order as timeline ("fresh" on top)
+    ], "asc");
```
aleksfront, line 24: Why it's better than sorting by `creationTimestamp`? Other places still use `creationTimestamp` for sorting by age. Pods, for example 
```
[columnId.age]: (pod: Pod) => pod.metadata.creationTimestamp
```

src/renderer/components/+events/events.tsx
ixrock: All these relevant to actual fix and `compact=true` from previous PR(s).

src/renderer/components/+events/event.store.ts
ixrock: Sorting is broken with raw `creationTimestamp` (it's date string, not timestamp)

src/renderer/api/kube-object.ts
ixrock: This is done more to inform why not used `getTimeDiffFromNow(): number` from above.

src/renderer/components/+events/events.tsx
Nokel81: It seems like a lot of work, why not just correctly handle slicing in `KubeObjectListLayout`? Namely sorting and then only displaying a slice?

ixrock: Because I need access to all items at `Events` page (e.g. for customizing header)

src/renderer/components/+events/events.scss
ixrock: Fixed / moved to `customizeHeader` prop.

src/renderer/api/kube-object.ts
```diff
@@ -122,12 +122,15 @@ export class KubeObject implements ItemObject {
     return this.metadata.namespace || undefined;
   }
 
-  // todo: refactor with named arguments
-  getAge(humanize = true, compact = true, fromNow = false) {
+  getTimeDiffFromNow(): number {
+    return new Date().getTime() - new Date(this.metadata.creationTimestamp).getTime();
```
Nokel81, line 126: No need to create a new `Date` instance if all that is being done is calling `getTime()`

```suggestion
    return Date.now() - new Date(this.metadata.creationTimestamp).getTime();
```

src/renderer/components/+events/events.tsx
```diff
@@ -37,23 +41,79 @@ const defaultProps: Partial<Props> = {
...
-    const { compact, compactLimit, className, ...layoutProps } = this.props;
-    const visibleItems = compact ? items.slice(0, compactLimit) : items;
+  @computed get visibleItems(): KubeEvent[] {
+    const { compact, compactLimit } = this.props;
+
+    if (compact) {
+      return this.items.slice(0, compactLimit);
+    }
+
+    return this.items;
```
Nokel81, line 83: Since this component is not exported we can change the API. Would switching to having `compactLimit > 0` implying `compact` be a good idea. 
```suggestion
    const { compactLimit = this.items.length } = this.props;

    return this.items.slice(0, compactLimit);
```

src/renderer/components/table/table.tsx
```diff
@@ -24,11 +25,9 @@ export interface TableProps extends React.DOMAttributes<HTMLDivElement> {
   selectable?: boolean; // Highlight rows on hover
   scrollable?: boolean; // Use scrollbar if content is bigger than parent's height
   storageKey?: string;  // Keep some data in localStorage & restore on page reload, e.g sorting params
-  sortable?: {
-    // Define sortable callbacks for every column in <TableHead><TableCell sortBy="someCol"><TableHead>
-    // @sortItem argument in the callback is an object, provided in <TableRow sortItem={someColDataItem}/>
-    [sortBy: string]: TableSortCallback;
-  };
+  // Define sortable callbacks for every column in <TableHead><TableCell sortBy="someCol"><TableHead>
+  // @sortItem argument in the callback is an object, provided in <TableRow sortItem={someColDataItem}/>
```
Nokel81: Multi line comments, especially with (js/ts)doc formatting should be done with a multi line comment.
```suggestion
  /**
   * Define sortable callbacks for every column in <TableHead><TableCell sortBy="someCol"><TableHead>
   * @sortItem argument in the callback is an object, provided in <TableRow sortItem={someColDataItem}/>
   */
```

Nokel81:
Looking good

src/renderer/api/kube-object.ts
ixrock: I guess technically it's the same :) So, not a big deal.

src/renderer/components/+events/events.tsx
ixrock: We don't really need to `.slice()` items every time imho..
Also I would like to keep some defaults for `compactLimit={10}` and enable it just by `compact` prop. 
Otherwise I should always specify some value for `compactLimit` which should be `> 0 && !== items.length`?

Nokel81: okay sounds good.

src/renderer/components/table/table.tsx
ixrock: Fixed

src/renderer/api/kube-object.ts
Nokel81: True but it is another allocation (I think)

Nokel81 (approved):

src/renderer/api/kube-object.ts
jim-docker: `Date.now()` is clearer and more concise

jim-docker (approved):

src/renderer/api/kube-object.ts
ixrock: Okay, you are right, guys. 
Will make a separate PR for that (to avoid reset approvals / speed up code-flow)

PR #2138, opened 2021-02-12
Fix accessible namespaces functionality under restrictive RBAC

jakolehm:
Fixes #2111

Closes #2130

PR #2142, opened 2021-02-12
Fix: list each namespace seperately if from accessibleNamespaces

Nokel81:
don't use list all optimization when operating in accessible namespaces mode

Follow up from #2138


jakolehm (approved):
Tested, fixes the issue where user has access to only one namespace.

ixrock (approved):

PR #2143, opened 2021-02-12
return undefined instead of localhost from getClusterIdFromHost

Nokel81:

### DIFF
--- a/src/common/__tests__/cluster-store.test.ts
+++ b/src/common/__tests__/cluster-store.test.ts
@@ -448,11 +448,25 @@ describe("pre 3.6.0-beta.1 config with an existing cluster", () => {
 });
 
 describe("getClusterIdFromHost", () => {
+  const clusterFakeId = "fe540901-0bd6-4f6c-b472-bce1559d7c4a";
+
   it("should return undefined for non cluster frame hosts", () => {
     expect(getClusterIdFromHost("localhost:45345")).toBeUndefined();
   });
 
-  it("should return undefined for non cluster frame hosts", () => {
-    expect(getClusterIdFromHost("fe540901-0bd6-4f6c-b472-bce1559d7c4a.localhost:59110")).toBe("fe540901-0bd6-4f6c-b472-bce1559d7c4a");
+  it("should return ClusterId for cluster frame hosts", () => {
+    expect(getClusterIdFromHost(`${clusterFakeId}.localhost:59110`)).toBe(clusterFakeId);
+  });
+
+  it("should return ClusterId for cluster frame hosts with additional subdomains", () => {
+    expect(getClusterIdFromHost(`abc.${clusterFakeId}.localhost:59110`)).toBe(clusterFakeId);
+    expect(getClusterIdFromHost(`abc.def.${clusterFakeId}.localhost:59110`)).toBe(clusterFakeId);
+    expect(getClusterIdFromHost(`abc.def.ghi.${clusterFakeId}.localhost:59110`)).toBe(clusterFakeId);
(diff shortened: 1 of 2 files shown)


src/common/__tests__/cluster-store.test.ts
```diff
@@ -446,3 +446,13 @@ describe("pre 3.6.0-beta.1 config with an existing cluster", () => {
...
   });
 });
+
+describe("getClusterIdFromHost", () => {
+  it("should return undefined for non cluster frame hosts", () => {
+    expect(getClusterIdFromHost("localhost:45345")).toBeUndefined();
+  });
+
+  it("should return undefined for non cluster frame hosts", () => {
+    expect(getClusterIdFromHost("fe540901-0bd6-4f6c-b472-bce1559d7c4a.localhost:59110")).toBe("fe540901-0bd6-4f6c-b472-bce1559d7c4a");
```
ixrock:
```suggestion
    const clusterFakeId = "fe540901-0bd6-4f6c-b472-bce1559d7c4a"
    expect(getClusterIdFromHost(`${clusterFakeId}.localhost:59110`)).toBe(clusterFakeId);
```

src/common/cluster-store.ts
```diff
@@ -354,18 +354,23 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
...
 export const clusterStore = ClusterStore.getInstance<ClusterStore>();
 
-export function getClusterIdFromHost(hostname: string): ClusterId {
-  const subDomains = hostname.split(":")[0].split(".");
+export function getClusterIdFromHost(host: string): ClusterId | undefined {
+  // e.g host == "%clusterId.localhost:45345"
+  const subDomains = host.split(":")[0].split(".");
 
-  return subDomains.slice(-2)[0]; // e.g host == "%clusterId.localhost:45345"
+  if (subDomains.length >= 2) {
```
ixrock: We need more restrictions on this. Otherwise could be easily break at some point..
Or maybe better to return `subDomains.slice(-2)[0]`?

```suggestion
  if (subDomains.length === 2) {
```

ixrock (dismissed):

src/common/cluster-store.ts
Nokel81: Well the old version didn't work so we cannot return to it. I figured that we want the first part of the hostname unless the hostname had only one part

src/common/__tests__/cluster-store.test.ts
jakolehm: We need test also for `xyz.fe540901-0bd6-4f6c-b472-bce1559d7c4a.localhost:59110` because watch api uses random subdomain as a workaround for http1.1 limitations

Nokel81: Good test case, thanks

src/common/cluster-store.ts
```diff
@@ -357,12 +357,11 @@ export const clusterStore = ClusterStore.getInstance<ClusterStore>();
...
   const subDomains = host.split(":")[0].split(".");
+  const lastTwoParts = subDomains.slice(-2);
 
-  if (subDomains.length >= 2) {
-    return subDomains[0];
-  }
+  lastTwoParts.pop(); // remove last entry (namely "localhost")
 
-  return undefined;
+  return lastTwoParts[0]; // ClusterId or undefined
```
ixrock:
```suggestion
  return subDomains.slice(-2, -1); // ClusterId or undefined
```

Nokel81: I guess I could do `subDomains.slice(-2, -1)[0]` if you really want.

ixrock (approved):
LGTM

nevalla (approved):

PR #2146, opened 2021-02-12
Add a possibility to resize table colums

### LINKED ISSUE #601: Allow to expand/resize column headers

Nachasic:
Closes #601
**Changes made in this PR:**
`<ItemListLayout>`:
1. Add `resizable` property to the component;
2. Create a state for this component of type `ItemListLayoutState` to remember cell sizes set by user - it's an `Array` of cell widths with number of items equal to number of columns in the table;
3. Based on `resizable` property and availability of saved set sizes, tell each rendered `<TableCell>` its size;
4. Forward `resizable` property to `<TableCell>`

`<TableCell>`
1. Add `resizable` property to the component;
2. Add `size` property to the component;
2. Add internal `_onResize?(width: number): void` callback property;
3. Add `resizeHandlerState` to keep track of last known mouse position when resizing;
4. Add `addMouseListeners()` member function to listen for `mousemove` and `mouseup` events;
5. Add `removeMouseListeners()` member function to clean up event listeners;
6. Utilize `componentWillUnmount()` lifecycle callback to perform additional cleanup for event listeners;
7. Add `span.resize-anchor` for resizable instances of `<TableCell>` for visual representation and initial capture of mouse events.
8. If defined, forward `size` property to styles.

**TODO:**
1. Styling;
2. Code cleanup;

**Demo:**
Screen Recording 2021-02-12 at 5 44 51 PM

### DIFF
--- a/src/common/__tests__/user-store.test.ts
+++ b/src/common/__tests__/user-store.test.ts
@@ -45,18 +45,6 @@ describe("user store tests", () => {
       expect(us.seenContexts.has("bar")).toBe(true);
     });
 
-    it("allows setting and getting preferences", () => {
-      const us = UserStore.getInstance<UserStore>();
-
-      us.preferences.httpsProxy = "abcd://defg";
-
-      expect(us.preferences.httpsProxy).toBe("abcd://defg");
-      expect(us.preferences.colorTheme).toBe(UserStore.defaultTheme);
-
-      us.preferences.colorTheme = "light";
-      expect(us.preferences.colorTheme).toBe("light");
-    });
-
     it("correctly resets theme to default value", async () => {
       const us = UserStore.getInstance<UserStore>();
 
@@ -74,6 +62,32 @@ describe("user store tests", () => {
 
       us.lastSeenAppVersion = (new SemVer(electron.app.getVersion())).inc("major").format();
       expect(us.isNewVersion).toBe(false);
+    });
+  });
+
+  describe("preferences", () => {
+    it("allows setting and getting preferences", () => {
+      const us = UserStore.getInstance<UserStore>();
+
+      us.preferences.httpsProxy = "abcd://defg";
+
+      expect(us.preferences.httpsProxy).toBe("abcd://defg");
+      expect(us.preferences.colorTheme).toBe(UserStore.defaultTheme);
+
+      us.preferences.colorTheme = "light";
+      expect(us.preferences.colorTheme).toBe("light");
+    });
+
+    it("allows setting and getting table column sizes", () => {
+      const us = UserStore.getInstance<UserStore>();
+      const [clusterId, tableId] = ["foo", "bar"];
+
+      expect( us.getTableSizing(clusterId, tableId) ).toStrictEqual([]);
+
+      const sizes = [100, 150, 70, 400];
+
+      us.setTableSizing(clusterId, tableId, sizes);
+      expect( us.getTableSizing(clusterId, tableId) ).toStrictEqual(sizes);
     });
   });
 
--- a/src/common/user-store.ts
+++ b/src/common/user-store.ts
@@ -93,6 +93,23 @@ export class UserStore extends BaseStore<UserStoreModel> {
   }
 
   @action
+  setTableSizing(clusterId: string, tableId: string, sizes: number[]) {
+    const { tableSizeConfig: config } = this.preferences;
+    
+    if (!config[clusterId]) {
+      config[clusterId] = { [tableId]: sizes };
+    } else {
+      config[clusterId][tableId] = sizes;
+    }
+  }
+
+  getTableSizing(clusterId: string, tableId: string): number[] {
+    const { tableSizeConfig: config } = this.preferences;
+
+    return config[clusterId]?.[tableId] || [];
+  }
+
+  @action
   setHiddenTableColumns(tableId: string, names: Set<string> | string[]) {
     this.preferences.hiddenTableColumns[tableId] = Array.from(names);
   }
--- a/src/common/user-store.ts
+++ b/src/common/user-store.ts
@@ -30,8 +30,8 @@ export interface UserPreferences {
   downloadBinariesPath?: string;
   kubectlBinariesPath?: string;
   openAtLogin?: boolean;
-  tableSizeConfig?: TableSizeConfig;
   hiddenTableColumns?: Record<string, string[]>
+  tableSizeConfig?: TableSizeConfig;
 }
 
 export class UserStore extends BaseStore<UserStoreModel> {
--- a/src/common/user-store.ts
+++ b/src/common/user-store.ts
@@ -9,6 +9,7 @@ import { getAppVersion } from "./utils/app-version";
 import { kubeConfigDefaultPath, loadConfig } from "./kube-helpers";
 import { appEventBus } from "./event-bus";
 import logger from "../main/logger";
+import { TableSizeConfig } from "./table-size-config";
 import path from "path";
 
 export interface UserStoreModel {
@@ -29,7 +30,8 @@ export interface UserPreferences {
   downloadBinariesPath?: string;
   kubectlBinariesPath?: string;
   openAtLogin?: boolean;
-  hiddenTableColumns?: Record<string, string[]>;
+  tableSizeConfig?: TableSizeConfig;
+  hiddenTableColumns?: Record<string, string[]>
 }
 
 export class UserStore extends BaseStore<UserStoreModel> {
@@ -57,6 +59,7 @@ export class UserStore extends BaseStore<UserStoreModel> {
     downloadKubectlBinaries: true,  // Download kubectl binaries matching cluster version
     openAtLogin: false,
     hiddenTableColumns: {},
+    tableSizeConfig: {},
   };
 
   protected async handleOnLoad() {
--- a/src/renderer/components/item-object-list/item-list-layout.scss
+++ b/src/renderer/components/item-object-list/item-list-layout.scss
@@ -29,6 +29,21 @@
     min-height: 200px;
   }
 
+  .TableHead {
+    .TableCell {
+      display: flex;
+      justify-content: space-between;
+    }
+
+    &:hover {
+      .TableCell {
+        .resize-anchor {
+          opacity: 1;
+        }
+      }
+    }
+  }
+
   .TableCell {
     position: relative;
 
@@ -36,14 +51,17 @@
       @include table-cell-action;
     }
     .resize-anchor {
-      position: absolute;
-      top: 0;
+      transition: opacity 0.3s;
+      opacity: 0;
       right: 0;
-      min-width: 5px;
       user-select: none;
       cursor: col-resize;
-      background-color: gray;
-      height: -webkit-fill-available;
+      transform: rotate(90deg);
+      color: var(--textColorPrimary);
+
+      &:hover {
+        color: white;
+      }
     }
   }
 }
--- a/src/renderer/components/item-object-list/item-list-layout.scss
+++ b/src/renderer/components/item-object-list/item-list-layout.scss
@@ -30,8 +30,20 @@
   }
 
   .TableCell {
+    position: relative;
+
     &.menu {
       @include table-cell-action;
+    }
+    .resize-anchor {
+      position: absolute;
+      top: 0;
+      right: 0;
+      min-width: 5px;
+      user-select: none;
+      cursor: col-resize;
+      background-color: gray;
+      height: -webkit-fill-available;
     }
   }
 }
--- a/src/renderer/components/item-object-list/item-list-layout.tsx
+++ b/src/renderer/components/item-object-list/item-list-layout.tsx
@@ -44,7 +44,7 @@ export interface ItemListLayoutProps<T extends ItemObject = ItemObject> {
   preloadStores?: boolean;
   isClusterScoped?: boolean;
   hideFilters?: boolean;
-  resizable?: boolean;
+  isResizable?: boolean;
   searchFilters?: SearchFilter<T>[];
   /** @deprecated */
   filterItems?: ItemsFilter<T>[];
@@ -232,7 +232,7 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps, ItemLis
   getRow(uid: string) {
     const {
       isSelectable, renderTableHeader, renderTableContents, renderItemMenu,
-      store, hasDetailsView, onDetails, resizable,
+      store, hasDetailsView, onDetails, isResizable: resizable,
       copyClassNameFromHeadCells, customizeTableRowProps, detailsItem,
     } = this.props;
     const { isSelected } = store;
@@ -277,7 +277,7 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps, ItemLis
             }
 
             if (!headCell || !this.isHiddenColumn(headCell)) {
-              return <TableCell resizable={resizable} key={index} {...cellProps} />;
+              return <TableCell isResizable={resizable} key={index} {...cellProps} />;
             }
           })
         }
@@ -410,7 +410,7 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps, ItemLis
   }
 
   renderTableHeader() {
-    const { customizeTableRowProps, renderTableHeader, isSelectable, isConfigurable, store, resizable } = this.props;
+    const { customizeTableRowProps, renderTableHeader, isSelectable, isConfigurable, store, isResizable: resizable } = this.props;
 
     if (!renderTableHeader) {
       return;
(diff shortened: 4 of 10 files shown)


src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -44,6 +44,7 @@ export interface ItemListLayoutProps<T extends ItemObject = ItemObject> {
   preloadStores?: boolean;
   isClusterScoped?: boolean;
   hideFilters?: boolean;
+  resizable?: boolean;
```
ixrock: Maybe better to name `isResizable: boolean` to be more aligned with other `is*`-props?

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -111,6 +116,11 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
     const defaultUserSettings = toJS(this.userSettings);
     const storage = createStorage<ItemListLayoutUserSettings>("items_list_layout", defaultUserSettings);
 
+    // this.cellSizes = new Array<number>(props.renderTableHeader.length);
+    this.state = {
```
ixrock: I would recommed to use `mobx` instead, e.g. `@observable resizeData = {cellSizes: 20}`

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -220,11 +230,18 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
     return this.applyFilters(filterItems, items);
   }
 
+  handleCellResize(index: number, width: number) {
```
ixrock: Could be simplified if we would work with sync-state management with `@observable` instead of default async `this.setState()`

src/renderer/components/table/table-cell.tsx
```diff
@@ -20,9 +23,30 @@ export interface TableCellProps extends React.DOMAttributes<HTMLDivElement> {
...
+  mousePosX?: number,
+};
+
+/**
+ * Ergonomic rate for throttling mouse events
+ * when resizing, in miliseconds.
+ * Represents optimal balance between user-facing latency
+ * and rendering performance 
+ */
+const ERGONOMIC_THROTTLE_RATE = 10;
```
ixrock: Can we extract all this logic with ergonomics to separated file/module?

src/renderer/components/table/table-cell.tsx
```diff
@@ -20,9 +23,30 @@ export interface TableCellProps extends React.DOMAttributes<HTMLDivElement> {
...
+
+export class TableCell extends React.Component<TableCellProps, ResizeHandlerState> {
+  private resizeHandlerState?: ResizeHandlerState;
+  private cellContainer: React.RefObject<HTMLDivElement> = React.createRef();
+
+  constructor(props: TableCellProps) {
+    super(props);
+    this.state = {};
```
ixrock: Not sure, but maybe it's better to use `mobx` everywhere from this point for state management in react-components?

ixrock (commented):
LGTM! :)
What about saving state of columns after app reload? Is it handled somehow already?

src/renderer/components/item-object-list/item-list-layout.tsx
Nachasic: I though of this, but then I realized `resizable` is probably going to be used as a JSX-attribute, so this should be within convention. If the rest of the codebase formulates boolean attributes with `is*` suffix, sure.

src/renderer/components/table/table-cell.tsx
Nachasic: I'll be sure to adopt it as a general practice in scope of the project, however right here it turned out to be more beneficial to keep `resizeHandlerState` outside component lifecycle all-together, since it is updated by global DOM-native events.

Nachasic:
> LGTM! :)
> What about saving state of columns after app reload? Is it handled somehow already?

Not yet, I've been thinking about this

aleksfront:
@Nachasic Here is the issues I've discovered:
1. Console error received from time to time ```react-dom.development.js:67 Warning: Received `true` for a non-boolean attribute `resizable`.```
resizable error
2. There's no resizing anchor avaiable on table head? Is that expected? In description gif shows that it was there at least.
3. Resizing shouldn't change anchor located on left. Demo:

https://user-images.githubusercontent.com/9607060/108066833-f8ef1c00-7070-11eb-8a9b-5876b2fef00f.mp4

4. Default drawer opening should be disabled when releasing mouse button.
5. Anchor not following the mouse cursor. 

https://user-images.githubusercontent.com/9607060/108067144-68650b80-7071-11eb-9341-ec814ef8052a.mp4

pashevskii:
@Nachasic  @aleksfront  Hi guys, a couple of months ago I pushed [similar pr](https://github.com/lensapp/lens/pull/1823). Some things like storing configs and left columns freezing were already implemented. Probably it could be userful to you. Feel free to use anything

aleksfront:
@Nachasic I've come to this styling of resizing anchor. If we'll stick with it, need to remove those anchore from all table cells except `TableHead` descendants. Wdyt?

https://user-images.githubusercontent.com/9607060/108507049-cfc9c800-72ca-11eb-835d-7f9a83513922.mp4

Nachasic:
@aleksfront looks nice, I'll go with it.

nevalla:
I have a feeling that container bricks are slightly off now.

src/common/__tests__/user-store.test.ts
```diff
@@ -45,18 +45,6 @@ describe("user store tests", () => {
...
-
-      us.preferences.httpsProxy = "abcd://defg";
-
-      expect(us.preferences.httpsProxy).toBe("abcd://defg");
-      expect(us.preferences.colorTheme).toBe(UserStore.defaultTheme);
-
-      us.preferences.colorTheme = "light";
-      expect(us.preferences.colorTheme).toBe("light");
-    });
-
```
Nokel81, line 59: Why was this test removed?

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -255,13 +283,23 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
             const headCell = renderTableHeader?.[index];
+            const cellSize = this.cellSizes[index];
 
             if (copyClassNameFromHeadCells && headCell) {
               cellProps.className = cssNames(cellProps.className, headCell.className);
             }
 
+            if (cellSize !== undefined) {
+              cellProps.size = cellSize;
+            }
```
Nokel81: If it is undefined then it won't change the meaning of the value. So this can be unconditional.

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -221,11 +241,19 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
+  handleCellResize(index: number, width: number) {
+    this.cellSizes[index] = width;
+  }
+
   @autobind()
   getRow(uid: string) {
     const {
       isSelectable, renderTableHeader, renderTableContents, renderItemMenu,
-      store, hasDetailsView, onDetails,
+      store, hasDetailsView, onDetails, isResizable: resizable,
```
Nokel81: Why the name change if you are just going to change the name back again?

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -255,13 +283,23 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
             }
 
+            if (cellSize !== undefined) {
+              cellProps.size = cellSize;
+            }
+
+            if (resizable) {
+              cellProps._onResize = width => this.handleCellResize(index, width);
+              cellProps._onResizeComplete = () => this.persistCellSizes();
+            }
```
Nokel81: Given that there is an `isResizable` prop these should be able to be passed in unconditionally.

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -411,8 +449,17 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
         )}
         {renderTableHeader.map((cellProps, index) => {
+          const _cellProps = isResizable ?
+            { 
+              ...cellProps, 
+              _onResize: (width: number) => this.handleCellResize(index, width),
+              _onResizeComplete: () => this.persistCellSizes(),
+              size: this.cellSizes[index],
+              isResizable 
+            } : cellProps;
```
Nokel81: Same with my arguments above, the truthiness of `isResizable` should determine whether the two callback are used so might as well just pass them in always.

src/renderer/components/table/table-cell.tsx
```diff
@@ -20,9 +24,98 @@ export interface TableCellProps extends React.DOMAttributes<HTMLDivElement> {
   _sorting?: Partial<TableSortParams>; // <Table> sorting state, don't use this prop outside (!)
   _sort?(sortBy: TableSortBy): void; // <Table> sort function, don't use this prop outside (!)
   _nowrap?: boolean; // indicator, might come from parent <TableHead>, don't use this prop outside (!)
+
+  /**
+   * Triggers every time this cell is resized by user
+   * @param width desired width of this cell
+   */
+  _onResize?(width: number): void;
```
Nokel81: Why are these underscore prefixed, that visually seems to imply that they are some sort of "private" or for "internal testing only"

src/renderer/components/table/table-cell.tsx
```diff
@@ -20,9 +24,98 @@ export interface TableCellProps extends React.DOMAttributes<HTMLDivElement> {
...
+   * Appends static event proxies to document `mousemove` and `mouseup` events for resizing,
+   * sets a particulate instance of `<TableCell>` to handle these events
+   * @param handlingInstance instance of `<TableCell>` responsible for handling events
+   */
+  private static addMouseEventListeners(handlingInstance: TableCell) {
+    TableCell.currentEventHandlingInstance = handlingInstance;
+
+    document.addEventListener("mousemove", TableCell.mouseMoveEventProxy);
+    document.addEventListener("mouseup", TableCell.mouseUpEventProxy);
+  }
```
Nokel81: All this work has already been done with `src/renderer/components/resizing-anchor.tsx`. Is there a good reason not to reuse that?

Nokel81 (commented):
Even though this PR works as intended it redoes work already in the repo.

src/common/__tests__/user-store.test.ts
Nachasic: I moved it to a separate `preferences` test suite,
https://github.com/lensapp/lens/pull/2146/files#diff-6d7101eddd71988a42a32649e56b79e1afdaf7d7f60d005024d4ca2a06d47f40R68

src/renderer/components/table/table-cell.tsx
Nachasic: After multiple refactors of this part I can no longer recall my motivation for this kind of naming tbh, changing it back to normal.

src/renderer/components/table/table-cell.tsx
Nachasic: Wasn't really aware of this component's existence in our codebase, lol
Looking into it now

src/renderer/components/table/table-cell.tsx
```diff
@@ -20,9 +24,98 @@ export interface TableCellProps extends React.DOMAttributes<HTMLDivElement> {
...
+};
+
 export class TableCell extends React.Component<TableCellProps> {
+  private resizeHandlerState?: ResizeHandlerState;
+  private cellContainer: React.RefObject<HTMLDivElement> = React.createRef();
+
+  /**
+   * Stores an instance of `<TableCell>` that is currently responsible for handling resize-related events
+   */
+  private static currentEventHandlingInstance?: TableCell;
```
ixrock: What the point of having `private static` instead of just `private`?

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -105,13 +107,31 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
+    const { clusterId, tableId, renderTableHeader: { length } } = this.props;
+    const savedSizes = userStore.getTableSizing(clusterId, tableId);
+
+    return savedSizes.length !== length ? new Array<number>(length) : savedSizes;
+  }
+
+  set persistentCellSizes(sizes: number[]) {
+    const { clusterId, tableId } = this.props;
+  
+    userStore.setTableSizing(clusterId, tableId, sizes);
```
ixrock: Maybe better to work with `clusterStore` instead of `userStore` to persist data? Therefore would no need in introducing `props.clusterId` to `ItemsListLayout` since you can get `getHostedClusterId()` from `cluster-store.ts` to manage proper place to keep in cluster-state.

src/common/user-store.ts
```diff
@@ -85,6 +88,23 @@ export class UserStore extends BaseStore<UserStoreModel> {
     return semver.gt(getAppVersion(), this.lastSeenAppVersion);
   }
 
+  @action
+  setTableSizing(clusterId: string, tableId: string, sizes: number[]) {
```
ixrock, line 96: Besides `tableId` probably better to keep in format `{[columnId: sting]: number /*size*/}`. Wdyt?

ixrock:
This is weird. Should the `warning` column (icon's container) have resizing capabilities too?

image

Also please notice how after resizing warning column it's impossible to resize namespace column (icon to resize just hidden even if I resize another column.

https://user-images.githubusercontent.com/6377066/111186802-0a6a0c00-85bc-11eb-99ff-bd9318b5c5aa.mov

src/renderer/components/table/table-cell.tsx
Nachasic: So `currentEventHandlingInstance` could be accessed independently from it's own lifecycle (creation/destruction), otherwise it would leak event handlers (will not properly handle listeners removal - event listeners would persist after destruction of a `TableCell` that was handling the events)

Nachasic: I'm currently working on integrating a resizing component made by @Nokel81 that would handle this logic on it's own.

src/renderer/components/item-object-list/item-list-layout.tsx
Nachasic: My logic was that column sizes are more of a fit for user preferences.

Nokel81: Would it make sense for an initial version to be "global" or use the same column widths over all clusters? @ixrock what do you think?

src/common/user-store.ts
```diff
@@ -29,7 +30,8 @@ export interface UserPreferences {
   downloadBinariesPath?: string;
   kubectlBinariesPath?: string;
   openAtLogin?: boolean;
-  hiddenTableColumns?: Record<string, string[]>;
+  tableSizeConfig?: TableSizeConfig;
+  hiddenTableColumns?: Record<string, string[]>
```
Nokel81: nit: can you reverse these to minimize the diff? (highly minor)

src/common/user-store.ts
```diff
@@ -89,6 +92,23 @@ export class UserStore extends BaseStore<UserStoreModel> {
     return semver.gt(getAppVersion(), this.lastSeenAppVersion);
   }
 
+  @action
+  setTableSizing(clusterId: string, tableId: string, sizes: number[]) {
```
Nokel81, line 96: Actually, this really isn't a setting because you don't change it in the user settings. This should utilize #2279

src/renderer/components/item-object-list/item-list-layout.scss
```diff
@@ -29,10 +29,40 @@
...
+      opacity: 0;
+      right: 0;
+      user-select: none;
+      cursor: col-resize;
+      transform: rotate(90deg);
+      color: var(--textColorPrimary);
+
+      &:hover {
+        color: white;
+      }
```
Nokel81, line 64: Should any of these be in resize-anchor proper?

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -37,13 +37,15 @@ export interface IHeaderPlaceholders {
 
 export interface ItemListLayoutProps<T extends ItemObject = ItemObject> {
   tableId?: string;
+  clusterId?: string;
```
Nokel81, line 40: This component shouldn't care about what cluster it is in, and wouldn't need to if you use `createStorage`

Nokel81 (commented):
Some more comments (haven't looked at your TableCell changes because you are still investigating ResizingAnchor I think).

src/renderer/components/item-object-list/item-list-layout.tsx
Nachasic: I thought having resized columns on one table then to see same result being applied to unrelated tables would be annoying for users. Besides implementing independent sizing for each particular table in each cluster didn't seem too difficult.

Nokel81: I meant the same table type has the same settings, not every table has the same settings. So I think we are on the same page.

src/common/user-store.ts
Nachasic: Noticed this too, working on it

Nachasic:
This functionality is targeted for 5.0.0, and we're planning to use another set of components to handle all table-related matters. Closing

PR #2147, opened 2021-02-12
Listen for before-quit-for-update to allow will-quit to use default flow

### LINKED ISSUE #2134: Accepting application update notification does not close Lens app and do the update
When accepting Lens app update notification, app updater tries to quit the app by closing app windows and restart the app. However, because of tray icon, application is not closed and the restart never happens.

```typescript
/**
     * Restarts the app and installs the update after it has been downloaded.
     * It should only be called after `update-downloaded` has been emitted.
     *
     * **Note:** `autoUpdater.quitAndInstall()` will close all application windows first and only emit `before-quit` event on `app` after that.
     * This is different from the normal quit event sequence.
     *
     * @param isSilent *windows-only* Runs the installer in silent mode. Defaults to `false`.
     * @param isForceRunAfter Run the app after finish even on silent install. Not applicable for macOS. Ignored if `isSilent` is set to `false`.
     */
    abstract quitAndInstall(isSilent?: boolean, isForceRunAfter?: boolean): void;
```

Nokel81:
Please @jakolehm or @nevalla test with with a signed local build (pretending to be a version before `rc.1`) to verify that this actually works as intended. 

fixes #2134

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -4,7 +4,7 @@ import "../common/system-ca";
 import "../common/prometheus-providers";
 import * as Mobx from "mobx";
 import * as LensExtensions from "../extensions/core-api";
-import { app, dialog, powerMonitor } from "electron";
+import { app, autoUpdater, dialog, powerMonitor } from "electron";
 import { appName } from "../common/vars";
 import path from "path";
 import { LensProxy } from "./lens-proxy";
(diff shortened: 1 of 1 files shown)


jakolehm (approved):
Seems to close the tray icon on update.

Nokel81:
@jakolehm It sort of has to do that given that it needs to fully quit and stop all running processes to update.

nevalla (approved):

Issue #2125, opened 2021-02-11
Large amount of fetch requests when selecting namespaces


aleksfront:
Every time an item is clicked inside namespace dropdown, large amount of fetch requests executed - items loaded for different stores and for each selected namespace accordingly. So things getting really slow as amount of selected namespaces increases.

https://user-images.githubusercontent.com/9607060/107629254-e3de4b80-6c72-11eb-80af-a1c8c0111494.mp4

Expected behavior: load nothing when choosing namespaces from dropdown.

Nokel81:
I think that this is desired behaviour. Though we shouldn't be cancelling any requests if you only choose another namespace from the drop down (only add new requests).

aleksfront:
Can you elaborate this a bit more? Why we ever need to load/reload something if we just trying to filter out current list view? And I assume that's not expected behavior at least in part of ui sloweness. More namespaces you selected, the slower ui would be.

jakolehm:
Is this already fixed?

Nokel81:
It might be yes given the changes I made to subscribing

PR #2151, opened 2021-02-13
The select all checkbox should not select disabled items

### LINKED ISSUE #2150: Trying to remove terminating namespaces causes frenetic list rendering
The "select all" checkbox on an ItemListLayout seems to select grayed out items. The grayed out items in this video are namespace resources that are stuck in the terminating state. When asking Lens to remove them the UI (namespace list) rerenders rapidly, showing the list alternating with "No items found" message. The kube api appears to be hammered with list namespaces requests (`GET /api-kube/api/v1/namespaces?`) 

https://user-images.githubusercontent.com/40840436/107825809-c207c600-6d51-11eb-9c60-9538aa67daf5.mov

1. "Select all" should not select grayed out items
2. May also need a guard against whatever is happening above (race or infinite loop?)

jim-docker:

fixes #2150

### DIFF
--- a/src/renderer/components/item-object-list/item-list-layout.tsx
+++ b/src/renderer/components/item-object-list/item-list-layout.tsx
@@ -89,7 +89,8 @@ const defaultProps: Partial<ItemListLayoutProps> = {
   filterItems: [],
   hasDetailsView: true,
   onDetails: noop,
-  virtual: true
+  virtual: true,
+  customizeTableRowProps: () => ({} as TableRowProps),
 };
 
 interface ItemListLayoutUserSettings {
@@ -241,7 +242,7 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
         sortItem={item}
         selected={detailsItem && detailsItem.getId() === itemId}
         onClick={hasDetailsView ? prevDefault(() => onDetails(item)) : undefined}
-        {...(customizeTableRowProps ? customizeTableRowProps(item) : {})}
+        {...customizeTableRowProps(item)}
       >
         {isSelectable && (
           <TableCell
@@ -398,9 +399,7 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
       return;
     }
 
-    const enabledItems = customizeTableRowProps
-      ? this.items.filter(item => !customizeTableRowProps(item).disabled)
-      : this.items;
+    const enabledItems = this.items.filter(item => !customizeTableRowProps(item).disabled);
 
     return (
       <TableHead showTopLine nowrap>
--- a/src/renderer/components/item-object-list/item-list-layout.tsx
+++ b/src/renderer/components/item-object-list/item-list-layout.tsx
@@ -233,13 +233,6 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
     if (!item) return;
     const itemId = item.getId();
 
-    const customProps = customizeTableRowProps ? customizeTableRowProps(item) : {};
-
-    // make sure a disabled item is not in the selected state
-    if (customProps.disabled) {
-      store.unselect(item);
-    }
-
     return (
       <TableRow
         key={itemId}
@@ -248,7 +241,7 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
         sortItem={item}
         selected={detailsItem && detailsItem.getId() === itemId}
         onClick={hasDetailsView ? prevDefault(() => onDetails(item)) : undefined}
-        {...customProps}
+        {...(customizeTableRowProps ? customizeTableRowProps(item) : {})}
       >
         {isSelectable && (
           <TableCell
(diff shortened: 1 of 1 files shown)


src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -392,19 +399,24 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
+    const { customizeTableRowProps, renderTableHeader, isSelectable, isConfigurable, store } = this.props;
 
     if (!renderTableHeader) {
       return;
     }
 
+    let items = this.items;
+    if ( customizeTableRowProps ) {
+      items = items.filter((item) => !customizeTableRowProps(item).disabled);
+    }
```
aleksfront: Please consider if ternary operator suits here better
```
const enabledItems = customizeTableRowProps
      ? this.items.filter(item => !customizeTableRowProps(item).disabled)
      : this.items;
```

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -233,6 +233,13 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
     if (!item) return;
     const itemId = item.getId();
 
+    const customProps = customizeTableRowProps ? customizeTableRowProps(item) : {};
+
+    // make sure a disabled item is not in the selected state
+    if (customProps.disabled) {
+      store.unselect(item);
+    }
```
aleksfront: I'm not sure we need these lines. 
1. It doesn't sound right to make side effects (unselecting from store) inside `getRow` prop which should have only 1 rendering purpose.
2. Selection/unselection should be happened from specific places only (checkboxes).
3. It's unclear when this is needed. While testing I didn't spot any changes with or without these lines.

Is there any specific case, where such unselection needed?

aleksfront:
Also, this might be out of scope and related to https://github.com/lensapp/lens/issues/2125, but frenetic rendeding still there. While namespaces gets deleted, new `GET` request pop up and refreshes whole list. Then watch events did they job also.
Demo:

https://user-images.githubusercontent.com/9607060/107920130-7cc7dc00-6f7d-11eb-895d-d3af45f44aff.mp4

But, the PR fixes the issue in title well. Disabled items don't become selected.

src/renderer/components/item-object-list/item-list-layout.tsx
jim-docker: My concern was that if a user "selects all" and there is a disabled item, what happens if that disabled item becomes enabled? I see now this should just result in clearing the "select all" checkbox so no need for this bulletproofing (which I don't think was right anyway :( ). I'll do some more tests...

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -392,19 +392,23 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
+
     return (
       <TableHead showTopLine nowrap>
         {isSelectable && (
           <TableCell
             checkbox
-            isChecked={store.isSelectedAll(this.items)}
-            onClick={prevDefault(() => store.toggleSelectionAll(this.items))}
+            isChecked={store.isSelectedAll(enabledItems)}
+            onClick={prevDefault(() => store.toggleSelectionAll(enabledItems))}
```
jim-docker, line 410: I think there's a flaw in the `store.toggleSelectionAll()` function, it verifies that each item is selected before deselecting them all. But otherwise it selects them all. This is fine if it is always called with a list of items that are either all selected or all not selected. If not then it should verify that they are all not selected before selecting them all.

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -392,19 +392,23 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
   renderTableHeader() {
-    const { renderTableHeader, isSelectable, isConfigurable, store } = this.props;
+    const { customizeTableRowProps, renderTableHeader, isSelectable, isConfigurable, store } = this.props;
 
     if (!renderTableHeader) {
       return;
     }
 
+    const enabledItems = customizeTableRowProps
+      ? this.items.filter(item => !customizeTableRowProps(item).disabled)
```
ixrock: Why do we need to use `customizeTableRowProps` here? I guess it's required only in items list rendering, but not in the header. By default `customizeTableRowProps` is `undefined`, maybe we should add default value?

```ts
const defaultProps: Partial<ItemListLayoutProps> = {
  customizeTableRowProps: (item) => ({} as TableRowProps),
  ...
```

And refactor also `item-list-layout.tsx:244` to:

```ts
 {...(customizeTableRowProps ? customizeTableRowProps(item) : {})} // before
 {...customizeTableRowProps(item)} // with default value
```

Wdyt?

```suggestion
      ? this.items.filter(item => !item.disabled)
```

ixrock (dismissed):
LGTM

src/renderer/components/item-object-list/item-list-layout.tsx
Nokel81: That does sound like a flaw, but tbh the name doesn't really match either of those behaviours.

Nokel81 (approved):

jim-docker:
> Also, this might be out of scope and related to #2125, but frenetic rendeding still there. While namespaces gets deleted, new `GET` request pop up and refreshes whole list. Then watch events did they job also.
> But, the PR fixes the issue in title well. Disabled items don't become selected.

could be same issue as #2150 but should be fixed in another PR

aleksfront (approved):

PR #2153, opened 2021-02-14
Cluster icons as letters

### LINKED ISSUE #1119: Display cluster names in the cluster vertical bar

I would like cluster names to be displayed below cluster icons (or as a replacement of cluster icons) in the cluster vertical bar. 
If multiple mode are to be supported, an option might be added to select how this is displayed (only icon, only name, both icon+name)

**Why is this needed**:

As mentioned in #1011, icons are super cool, but when having more than a couple clusters, finding the proper one is painful with only icons, as you can only hover over each icon to display its name until you find the proper cluster.

Not relevant, I think.

jakolehm:
image

image

### DIFF
--- a/src/renderer/components/cluster-icon/cluster-icon.scss
+++ b/src/renderer/components/cluster-icon/cluster-icon.scss
@@ -7,23 +7,25 @@
   user-select: none;
   cursor: pointer;
 
-  &.interactive {
-    img {
-      opacity: .55;
-    }
+
+  div.MuiAvatar-colorDefault {
+    font-weight:500;
+    text-transform: uppercase;
+  }
+
+  div.active {
+    background-color: var(--primary);
+  }
+
+  div.default {
+    background-color: var(--halfGray);
   }
 
   &.active, &.interactive:hover {
     background-color: #fff;
-
-    img {
-      opacity: 1;
+    div {
+      background-color: var(--primary);
     }
-  }
-
-  img {
-    width: var(--size);
-    height: var(--size);
   }
 
   .Badge {
--- a/src/renderer/components/cluster-icon/cluster-icon.tsx
+++ b/src/renderer/components/cluster-icon/cluster-icon.tsx
@@ -2,14 +2,13 @@ import "./cluster-icon.scss";
 
 import React, { DOMAttributes } from "react";
 import { disposeOnUnmount, observer } from "mobx-react";
-import { Params as HashiconParams } from "@emeraldpay/hashicon";
-import { Hashicon } from "@emeraldpay/hashicon-react";
 import { Cluster } from "../../../main/cluster";
 import { cssNames, IClassName } from "../../utils";
 import { Badge } from "../badge";
 import { Tooltip } from "../tooltip";
 import { subscribeToBroadcast } from "../../../common/ipc";
 import { observable } from "mobx";
+import { Avatar } from "@material-ui/core";
 
 interface Props extends DOMAttributes<HTMLElement> {
   cluster: Cluster;
@@ -19,7 +18,6 @@ interface Props extends DOMAttributes<HTMLElement> {
   showTooltip?: boolean;
   interactive?: boolean;
   isActive?: boolean;
-  options?: HashiconParams;
 }
 
 const defaultProps: Partial<Props> = {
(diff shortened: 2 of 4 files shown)


src/renderer/components/cluster-icon/cluster-icon.scss
```diff
@@ -7,23 +7,25 @@
...
-    img {
-      opacity: 1;
-    }
+  div.default {
+    background-color: var(--halfGray);
   }
 
-  img {
-    width: var(--size);
-    height: var(--size);
```
jim-docker, line 26: you may need to keep img styling for when user sets their own icon. Unless we are eliminating that?

jim-docker (commented):
Though the hashicons are neat I prefer this for differentiating clusters. Maybe consider using 3 characters when a duplicate is encountered? Or put letters over hashicons? Or vary the background colour?

Nokel81:
I like the idea of variable length display names (along with whatever the auto generated icons are called).

Something along the lines of the shortest prefix for each entry such that it is unambiguous (up to some max length of course).

src/renderer/components/cluster-icon/cluster-icon.scss
jakolehm: Oh right, I probably should revert these.

jakolehm:
> Though the hashicons are neat I prefer this for differentiating clusters. Maybe consider using 3 characters when a duplicate is encountered? Or put letters over hashicons? Or vary the background colour?

I tried following things and decided to revert them (they didn't look good):

- 3 characters -> looks a bit odd (might be doable with more tweaking)
- varied background color -> we don't have color palette for this, throwing random colors here looks really bad

nevalla:
I like this. It makes UI more calm.

> * varied background color -> we don't have color palette for this, throwing random colors here looks really bad

Some color palette would be nice in future. Not to throw random colors but allow users to either upload custom cluster icon or select background color?

src/renderer/components/cluster-icon/cluster-icon.scss
Nokel81: @jakolehm are you going to revert them?

src/renderer/components/cluster-icon/cluster-icon.tsx
```diff
@@ -47,9 +45,29 @@ export class ClusterIcon extends React.Component<Props> {
...
+
+    if (splittedName.length === 1) {
+      splittedName = splittedName[0].split("-");
+    }
+
+    if (splittedName.length === 1) {
+      splittedName = splittedName[0].split("@");
+    }
+
+    splittedName = splittedName.map((part) => part.replace(/\W/g, ""));
```
Nokel81, line 59: Why remove non-ascii characters? That might leave a name with nothing?

src/renderer/components/cluster-icon/cluster-icon.tsx
```diff
@@ -67,7 +85,7 @@ export class ClusterIcon extends React.Component<Props> {
           <Tooltip targetId={clusterIconId}>{name}</Tooltip>
         )}
         {icon && <img src={icon} alt={name}/>}
-        {!icon && <Hashicon value={clusterId} options={options}/>}
+        {!icon && <Avatar variant="square" className={isActive ? "active" : "default"}>{this.iconString}</Avatar>}
```
Nokel81, line 88: Maybe move these out?

src/renderer/components/cluster-icon/cluster-icon.tsx
```diff
@@ -47,9 +45,29 @@ export class ClusterIcon extends React.Component<Props> {
     ]);
   }
 
+  get iconString() {
```
aleksfront, line 48: This method is asking to be tested. Maybe moved into some separate place to make unit test for this?

src/renderer/components/cluster-icon/cluster-icon.tsx
aleksfront: If decide be removed, we should also restrict `Cluster name` input field to prevent entering non-ascii characters.

aleksfront (changes requested):
**A bug**: when defining custom cluster icon, icon size didn't changed. In master it works okay, images are fitted in their respective areas.
oversized icons

**A suggestion**: cluster icons as letters should be enabled as a flag in Preferences allowing users to choose between 2 icon styles.

src/renderer/components/cluster-icon/cluster-icon.tsx
Nokel81: In #2331 I specifically support non-ascii names.

Nokel81:
@aleksfront That bug is due to this PR removing the stylings for ``'s

jakolehm:
Closing because this is included in #2418 .

Nokel81:
@jakolehm Did you include my work on background colours?

jakolehm:
> @jakolehm Did you include my work on background colours?

Nope, only included portions of this PR.

PR #2155, opened 2021-02-15
Fix Documentation: fix broken link in getting-started page

rafedramzi:
Hi! It's that 1 liner fix broken link in the docs PR. It bugs me when I found a broken link in the docs, so here's the fix (tested locally against your master branch). I also think you need to cherry-pick this commit into your other release branches as well (maybe v4.xx release?), cheers!

### DIFF
--- a/docs/getting-started/README.md
+++ b/docs/getting-started/README.md
@@ -5,7 +5,7 @@ Lens is lightweight and simple to install. You'll be up and running in just a fe
 
 ## System Requirements
 
-Review the [System Requirements](/supporting/requirements/) to check if your computer configuration is supported.
+Review the [System Requirements](../supporting/requirements.md) to check if your computer configuration is supported.
 
 
 ## macOS


Nokel81 (approved):

docs/getting-started/README.md
```diff
@@ -5,7 +5,7 @@ Lens is lightweight and simple to install. You'll be up and running in just a fe
 
 ## System Requirements
 
-Review the [System Requirements](/supporting/requirements/) to check if your computer configuration is supported.
```
jim-docker, line 8: this seems to work when the docs are served locally...

docs/getting-started/README.md
```diff
@@ -5,7 +5,7 @@ Lens is lightweight and simple to install. You'll be up and running in just a fe
...
+Review the [System Requirements](../supporting/requirements.md) to check if your computer configuration is supported.
```
jim-docker, line 8: ...but this is in line with other links in this file anyway

jim-docker (approved):

docs/getting-started/README.md
Nokel81: Ah so it should be this instead?

jim-docker: ? This is what @rafedramzi proposed, I was merely agreeing it is suitable, even though the original link worked locally.

rafedramzi:
Ok I should probably explain the problem more clearly:

When you're testing locally, the URL is http://localhost:8000/getting-started/ so
/supporting/requirements/ will be evaluated as http://localhost:8000/supporting/requirements/ looks good so far


But on your prod, the URL is https://docs.k8slens.dev/v4.1.1/getting-started/ ans also it has multiple version deployed!
So /supporting/requirements/ will be evaluated as https://docs.k8slens.dev/supporting/requirements/ and that's not correct
The correct URL should be https://docs.k8slens.dev/v4.1.1/supporting/requirements/

Edit: wrong link evaluation, just fixed it

docs/getting-started/README.md
rafedramzi: Sorry I didn't explain the problem, I explain the problem in https://github.com/lensapp/lens/pull/2155#issuecomment-781393646

PR #2157, opened 2021-02-15
Register existing store with new apiBase when re-registering Kube API

nevalla:
This PR fix issue when api version is resolved from Kubernetes API and respective store has been already registered with default api version.

### DIFF
--- a/src/renderer/api/api-manager.ts
+++ b/src/renderer/api/api-manager.ts
@@ -23,6 +23,12 @@ export class ApiManager {
 
   registerApi(apiBase: string, api: KubeApi) {
     if (!this.apis.has(apiBase)) {
+      this.stores.forEach((store) => {
+        if(store.api === api) {
+          this.stores.set(apiBase, store);
+        }
+      });
+
       this.apis.set(apiBase, api);
     }
   }
@@ -41,14 +47,6 @@ export class ApiManager {
 
       if (entry) this.unregisterApi(entry[0]);
     }
-  }
-
-  updateStoreKey(oldKey: string) {
-    const store = this.stores.get(oldKey);
-
-    if (!store) return;
-    this.stores.delete(oldKey);
-    this.registerStore(store, [store.api]);
   }
 
   @action
--- a/src/renderer/api/api-manager.ts
+++ b/src/renderer/api/api-manager.ts
@@ -43,6 +43,14 @@ export class ApiManager {
     }
   }
 
+  updateStoreKey(oldKey: string) {
+    const store = this.stores.get(oldKey);
+
+    if (!store) return;
+    this.stores.delete(oldKey);
+    this.registerStore(store, [store.api]);
+  }
+
   @action
   registerStore(store: KubeObjectStore, apis: KubeApi[] = [store.api]) {
     apis.forEach(api => {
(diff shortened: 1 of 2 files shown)


src/renderer/api/kube-api.ts
```diff
@@ -229,8 +229,11 @@ export class KubeApi<T extends KubeObject = any> {
       });
 
       if (this.apiVersionPreferred) {
-        Object.defineProperty(this, "apiBase", { value: this.getUrl() });
+        const apiBase = this.getUrl();
+
+        Object.defineProperty(this, "apiBase", { value: apiBase });
         apiManager.registerApi(this.apiBase, this);
+        apiManager.updateStoreKey(this.objectConstructor.apiBase);
```
ixrock: Maybe better to do all the job for updating the keys inside `registerApi` since we pass `this` already?
Something like: 

```ts
apiManager.registerApi(this.apiBase, this, {update: true} /*options={}*/);
```

Also, purpose of `updateStoreKey` is not clear for me (why / when should I use it).

jakolehm: Maybe we could do the store key update automatically when `registerApi` is called?

src/renderer/api/kube-api.ts
```diff
@@ -229,7 +229,9 @@ export class KubeApi<T extends KubeObject = any> {
       });
 
       if (this.apiVersionPreferred) {
-        Object.defineProperty(this, "apiBase", { value: this.getUrl() });
+        const apiBase = this.getUrl();
```
jakolehm: Unnecessary change, please revert 🙂

jakolehm (dismissed):

src/renderer/api/kube-api.ts
ixrock: This is what I also proposed.

src/renderer/api/api-manager.ts
```diff
@@ -23,6 +23,12 @@ export class ApiManager {
 
   registerApi(apiBase: string, api: KubeApi) {
     if (!this.apis.has(apiBase)) {
+      this.stores.forEach((store) => {
+        if(store.api === api) {
+          this.stores.set(apiBase, store);
+        }
+      });
+
```
ixrock, line 31:
```suggestion
      const store = this.getStore(api);
      if (store) this.stores.set(apiBase, store);
```

ixrock (approved):

src/renderer/api/api-manager.ts
nevalla: This does not work since api has updated value for `apiBase` and no stores are found with that key.

src/renderer/api/api-manager.ts
```diff
@@ -23,6 +23,12 @@ export class ApiManager {
 
   registerApi(apiBase: string, api: KubeApi) {
     if (!this.apis.has(apiBase)) {
+      this.stores.forEach((store) => {
```
jakolehm, line 26: Test for this would be nice.

jakolehm (approved):

src/renderer/api/api-manager.ts
ixrock: Ok, but sounds like a bug since `apiManager.getStore(api: string | KubeApi)`..
Introduced from here: https://github.com/lensapp/lens/pull/2091

PR #2158, opened 2021-02-16
Fix: deprecated helm chart filtering

### LINKED ISSUE #2156: Helm chart list doesn't filter out deprecated charts
List data received doesn't contain `deprecated` fields, that's why filtration introduced [here](https://github.com/lensapp/lens/blob/master/src/renderer/components/+apps-helm-charts/helm-charts.tsx#L76) doesn't work.

deprecated chart
chart list data

When chart is loaded separately, it has needed flag.
specific chart data

aleksfront:
Now deprecated Helm charts are not showed in list.

no deprecated charts

Fixes #2156

### DIFF
--- a/src/main/helm/__mocks__/helm-chart-manager.ts
+++ b/src/main/helm/__mocks__/helm-chart-manager.ts
@@ -88,14 +88,14 @@ export class HelmChartManager {
             {
               apiVersion: "3.0.0",
               name: "pretzel",
-              version: "1.0.0-beta2",
+              version: "1.0",
               repo: "bitnami",
               digest: "test",
             },
             {
               apiVersion: "3.0.0",
               name: "pretzel",
-              version: "1.0.0",
+              version: "1.0.1",
               repo: "bitnami",
               digest: "test"
             }
--- a/src/main/helm/__mocks__/helm-chart-manager.ts
+++ b/src/main/helm/__mocks__/helm-chart-manager.ts
@@ -1,0 +1,115 @@
+import { HelmRepo, HelmRepoManager } from "../helm-repo-manager";
+
+export class HelmChartManager {
+  private cache: any = {};
+  private repo: HelmRepo;
+
+  constructor(repo: HelmRepo){
+    this.cache = HelmRepoManager.cache;
+    this.repo = repo;
+  }
+
+  public async charts(): Promise<any> {
+    return new Promise((resolve, reject) => {
+      let groups: any = {};
+
+      if (this.repo.name == "stable") {
+        groups = {
+          "apm-server": [
+            {
+              apiVersion: "3.0.0",
+              name: "apm-server",
+              version: "2.1.7",
+              repo: "stable",
+              digest: "test"
+            },
+            {
+              apiVersion: "3.0.0",
+              name: "apm-server",
+              version: "2.1.6",
+              repo: "stable",
+              digest: "test"
+            }
+          ],
+          "redis": [
+            {
+              apiVersion: "3.0.0",
+              name: "apm-server",
+              version: "1.0.0",
+              repo: "stable",
+              digest: "test"
+            },
+            {
+              apiVersion: "3.0.0",
+              name: "apm-server",
+              version: "0.0.9",
+              repo: "stable",
+              digest: "test"
+            }
+          ]
+        };
+      }
+
+      if (this.repo.name == "experiment") {
+        groups = {
+          "fairwind": [
+            {
+              apiVersion: "3.0.0",
+              name: "fairwind",
+              version: "0.0.1",
+              repo: "experiment",
+              digest: "test"
+            },
+            {
+              apiVersion: "3.0.0",
+              name: "fairwind",
+              version: "0.0.2",
+              repo: "experiment",
+              digest: "test",
+              deprecated: true
+            }
+          ]
+        };
+      }
+
+      if (this.repo.name == "bitnami") {
+        groups = {
+          "hotdog": [
+            {
+              apiVersion: "3.0.0",
+              name: "hotdog",
+              version: "1.0.1",
+              repo: "bitnami",
+              digest: "test"
+            },
+            {
+              apiVersion: "3.0.0",
+              name: "hotdog",
+              version: "1.0.2",
+              repo: "bitnami",
+              digest: "test",
+            }
+          ],
+          "pretzel": [
+            {
+              apiVersion: "3.0.0",
+              name: "pretzel",
+              version: "1.0.0-beta2",
+              repo: "bitnami",
+              digest: "test",
+            },
+            {
+              apiVersion: "3.0.0",
+              name: "pretzel",
+              version: "1.0.0",
+              repo: "bitnami",
+              digest: "test"
+            }
+          ]
+        };
+      }
+
+      resolve(groups);
+    });
+  }
+}
--- a/src/main/helm/__mocks__/helm-chart-manager.ts
+++ b/src/main/helm/__mocks__/helm-chart-manager.ts
@@ -10,11 +10,9 @@ export class HelmChartManager {
   }
 
   public async charts(): Promise<any> {
-    return new Promise((resolve, reject) => {
-      let groups: any = {};
-
-      if (this.repo.name == "stable") {
-        groups = {
+    switch (this.repo.name) {
+      case "stable":
+        return Promise.resolve({
           "apm-server": [
             {
               apiVersion: "3.0.0",
@@ -47,11 +45,9 @@ export class HelmChartManager {
               digest: "test"
             }
           ]
-        };
-      }
-
-      if (this.repo.name == "experiment") {
-        groups = {
+        });
+      case "experiment":
+        return Promise.resolve({
           "fairwind": [
             {
               apiVersion: "3.0.0",
@@ -69,11 +65,9 @@ export class HelmChartManager {
               deprecated: true
             }
           ]
-        };
-      }
-
-      if (this.repo.name == "bitnami") {
-        groups = {
+        });
+      case "bitnami":
+        return Promise.resolve({
           "hotdog": [
             {
               apiVersion: "3.0.0",
@@ -106,10 +100,9 @@ export class HelmChartManager {
               digest: "test"
             }
           ]
-        };
-      }
-
-      resolve(groups);
-    });
+        });
+      default:
+        return Promise.resolve({});
+    }
   }
 }
--- a/src/main/helm/helm-chart-manager.ts
+++ b/src/main/helm/helm-chart-manager.ts
@@ -4,9 +4,12 @@ import { HelmRepo, HelmRepoManager } from "./helm-repo-manager";
 import logger from "../logger";
 import { promiseExec } from "../promise-exec";
 import { helmCli } from "./helm-cli";
+import { HelmChart } from "../../renderer/api/endpoints/helm-charts.api";
+
+type HelmGroups = { [key: string]: HelmChart[] };
 
 type CachedYaml = {
-  entries: any; // todo: types
+  entries: HelmGroups
 };
 
 export class HelmChartManager {
@@ -24,7 +27,7 @@ export class HelmChartManager {
     return charts[name];
   }
 
-  public async charts(): Promise<any> {
+  public async charts(): Promise<HelmGroups> {
     try {
       const cachedYaml = await this.cachedYaml();
 
@@ -32,7 +35,7 @@ export class HelmChartManager {
     } catch(error) {
       logger.error(error);
 
-      return [];
+      return {};
     }
   }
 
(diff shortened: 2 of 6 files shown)


src/renderer/api/endpoints/helm-charts.api.ts
```diff
@@ -27,7 +27,7 @@ export const helmChartsApi = {
         return Object
           .values(data)
           .reduce((allCharts, repoCharts) => allCharts.concat(Object.values(repoCharts)), [])
-          .map(HelmChart.create);
+          .map((charts: HelmChart[]) => HelmChart.create(charts[0]));
```
Nokel81:
```suggestion
          .map(([chart]) => HelmChart.create(chart));
```

Nokel81: Though why are we only getting the first entry?

src/renderer/components/+apps-helm-charts/helm-charts.tsx
```diff
@@ -72,9 +72,6 @@ export class HelmCharts extends Component<Props> {
             (chart: HelmChart) => chart.getAppVersion(),
             (chart: HelmChart) => chart.getKeywords(),
           ]}
-          filterItems={[
-            (items: HelmChart[]) => items.filter(item => !item.deprecated)
-          ]}
```
Nokel81, line 77: Why was this not working as intended? Was it filtering too late?

src/main/helm/helm-service.ts
```diff
@@ -95,21 +93,27 @@ class HelmService {
...
-        return entry["deprecated"] != true;
-      });
-    }
+  constructor(groups: HelmChartGroups) {
+    this.items = new Map(Object.entries(groups));
+    this.excludeDeprecatedGroups();
+  }
 
-    return entries;
+  excludeDeprecatedGroups() {
```
Nokel81: This should be private and probably just part of the getter

src/renderer/api/endpoints/helm-charts.api.ts
```diff
@@ -5,7 +5,7 @@ import { autobind } from "../../utils";
 
 interface IHelmChartList {
   [repo: string]: {
-    [name: string]: HelmChart;
+    [name: string]: HelmChart[];
   };
 }
```
Nokel81: This sub type seems to be used in multiple places. Should it just be declared here and imported above?

```suggestion
export type RepoHelmChartList = Record<string, HelmChart[]>;
export type HelmChartList = Record<string, RepoHelmChartList>;
```

src/main/helm/helm-service.ts
```diff
@@ -95,21 +93,27 @@ class HelmService {
...
   }
+}
 
-  protected excludeDeprecated(entries: any) {
-    for (const key in entries) {
-      entries[key] = entries[key].filter((entry: any) => {
-        if (Array.isArray(entry)) {
-          return entry[0]["deprecated"] != true;
-        }
+class ChartGroups {
```
Nokel81: This (imo) is not the sort of reason to introduce a class. This could easily just be a free function given it is immediately deconstructed after `new ChartGroups` call.

src/renderer/components/+apps-helm-charts/helm-charts.tsx
jim-docker: `filterItems` is (being) deprecated from `ItemListLayout` anyway...?

Nokel81: Ah I see makes sense

src/renderer/api/endpoints/helm-charts.api.ts
aleksfront: The list is received in a form of "chart groups", for examples:
```
'apm-server: [
  {
    name: 'apm-server',
    ...
    version: '2.1.7'
  },
  {
    name: 'apm-server',
    ...
    version: '2.1.6'
  }
]
```
So we're choosing only last one (`charts[0]`) to avoid listing same chart many times in a row.

src/renderer/components/+apps-helm-charts/helm-charts.tsx
aleksfront: Yes, `filterItems` would get deprecated in some way, but that's not the only reason. Before this PR we had filters in both places - here in UI in `helm-charts.tsx` and inside `helm-service.ts`. So, keeping it in both places doesn't make any sense and I chose to leave it in "backend" with some refactorings (where my new class was introduced). My decision was dictated by a feeling that react component shouldn't do such actions (maybe in rare cases only).

src/main/helm/helm-service.ts
aleksfront: Set as `private` but not sure about putting it into getter since I find it unexpected - getters shouldn't do any modifitations to entities. Function should do only one thing at best.

src/main/helm/helm-service.ts
aleksfront: Totally agree, but this was my experiment of making code easier to read and understand. `HelmChartGroups` seems like a separate entity to me, so I tried to move it into separate class along with related operations. If it's harder to read/understant, I can make a functions without a class.

src/renderer/api/endpoints/helm-charts.api.ts
aleksfront: Good change, keeping your types. Thanks!

ixrock: Maybe this `type HelmChartList = Record<string, HelmChart | HelmChart[]>`?

src/main/helm/helm-service.ts
```diff
@@ -3,29 +3,27 @@ import logger from "../logger";
 import { repoManager } from "./helm-repo-manager";
 import { HelmChartManager } from "./helm-chart-manager";
 import { releaseManager } from "./helm-release-manager";
+import { HelmChart } from "../../renderer/api/endpoints/helm-charts.api";
+
+import type { HelmChartList, RepoHelmChartList } from "../../renderer/api/endpoints/helm-charts.api";
```
ixrock: You can `import` types from line above since we're importing `HelmChart` class as well.

src/renderer/api/endpoints/helm-charts.api.ts
Nokel81: So it is guaranteed to be sorted in decreasing version order?

src/main/helm/helm-chart-manager.ts
```diff
@@ -24,15 +25,15 @@ export class HelmChartManager {
...
   }
 
-  public async charts(): Promise<any> {
+  public async charts(): Promise<RepoHelmChartList> {
     try {
       const cachedYaml = await this.cachedYaml();
 
       return cachedYaml["entries"];
     } catch(error) {
       logger.error(error);
```
Nokel81:
```suggestion
      logger.error("HELM-CHART-MANAGER]: failed to list charts", { error });
```

src/renderer/components/+apps-helm-charts/helm-charts.tsx
Nokel81: Sounds good

src/main/helm/helm-service.ts
Nokel81: I commend the effort but I really don't think that classes should be used like this. It isn't an object that we are using merely we are using it for filtering.

src/main/helm/helm-chart-manager.ts
aleksfront: Done

src/main/helm/helm-service.ts
aleksfront: Replaced class with common methods.

src/renderer/api/endpoints/helm-charts.api.ts
aleksfront: Added descending sorting by `version` field.

aleksfront:
@ixrock @Nokel81 I've removed dedicated class for sake of common methods + added sorting by version to make sure we will have always latest version + added test coverage for filter/sorting. PTAL again please ^

src/main/helm/__mocks__/helm-chart-manager.ts
```diff
@@ -0,0 +1,115 @@
...
+              version: "1.0.0",
+              repo: "bitnami",
+              digest: "test"
+            }
+          ]
+        };
+      }
+
+      resolve(groups);
+    });
```
Nokel81:
```suggestion
    switch (this.repo.name) {
      case "stable":
        return Promise.resolve({
          "apm-server": [
            {
              apiVersion: "3.0.0",
              name: "apm-server",
              version: "2.1.7",
              repo: "stable",
              digest: "test"
            },
            {
              apiVersion: "3.0.0",
              name: "apm-server",
              version: "2.1.6",
              repo: "stable",
              digest: "test"
            }
          ],
          "redis": [
            {
              apiVersion: "3.0.0",
              name: "apm-server",
              version: "1.0.0",
              repo: "stable",
              digest: "test"
            },
            {
              apiVersion: "3.0.0",
              name: "apm-server",
              version: "0.0.9",
              repo: "stable",
              digest: "test"
            }
          ]
        });
      case "experiment":
        return Promise.resolve({
          "fairwind": [
            {
              apiVersion: "3.0.0",
              name: "fairwind",
              version: "0.0.1",
              repo: "experiment",
              digest: "test"
            },
            {
              apiVersion: "3.0.0",
              name: "fairwind",
              version: "0.0.2",
              repo: "experiment",
              digest: "test",
              deprecated: true
            }
          ]
        });
      case "bitnami":
        return Promise.resolve({
          "hotdog": [
            {
              apiVersion: "3.0.0",
              name: "hotdog",
              version: "1.0.1",
              repo: "bitnami",
              digest: "test"
            },
            {
              apiVersion: "3.0.0",
              name: "hotdog",
              version: "1.0.2",
              repo: "bitnami",
              digest: "test",
            }
          ],
          "pretzel": [
            {
              apiVersion: "3.0.0",
              name: "pretzel",
              version: "1.0.0-beta2",
              repo: "bitnami",
              digest: "test",
            },
            {
              apiVersion: "3.0.0",
              name: "pretzel",
              version: "1.0.0",
              repo: "bitnami",
              digest: "test"
            }
          ]
        });
      default:
        return Promise.resolve({});
    }
```

aleksfront: A bit cleanier!

Nokel81 (dismissed):

ixrock (changes requested):
Tried locally, doesn't work as expected (infinite spinner).

Screenshot 2021-02-19 at 14 36 22

Current `master`:

image

Also noticed this in app logs after navigating to charts page.

```
2.000000. This error will be logged once per view in violation.
(node:34244) UnhandledPromiseRejectionWarning: TypeError: Invalid Version: 0.6
    at new SemVer (/Users/ixrock/Desktop/Projects/lens/node_modules/semver/classes/semver.js:41:13)
    at Object.compare (/Users/ixrock/Desktop/Projects/lens/node_modules/semver/functions/compare.js:3:32)
    at eval (webpack-internal:///./src/main/helm/helm-service.ts:89:70)
    at Array.sort (<anonymous>)
    at HelmService.sortChartsByVersion (webpack-internal:///./src/main/helm/helm-service.ts:88:49)
    at HelmService.listCharts (webpack-internal:///./src/main/helm/helm-service.ts:25:39)
    at async HelmApiRoute.listCharts (webpack-internal:///./src/main/routes/helm-route.ts:12:24)
    at async Router.route (webpack-internal:///./src/main/router.ts:34:13)
(node:34244) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:34244) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

aleksfront:
Added `semver.coercion` for versions before comparing them. Coercion transforms non-semantic versions to semantic ones, for example `v0.6 -> 0.6.0`, `0.1 -> 0.1.0` etc. This solves case founded by @ixrock ^
https://www.npmjs.com/package/semver

Nokel81 (approved):

nevalla (approved):
LGTM

ixrock (approved):
For now I don't have any infinite loading indicators at the page, but also no charts for `stable` repo branch at all.. 
I guess _this is fine_ since on the screenshot you have listed other repos (which I don't have for some reasons).

aleksfront:
> For now I don't have any infinite loading indicators at the page, but also no charts for `stable` repo branch at all..
> I guess _this is fine_ since on the screenshot you have listed other repos (which I don't have for some reasons).

Repos could be added in Preferences
helm repositories

PR #2165, opened 2021-02-16
Fix: keep highlighted table row during navigation in the details panel

ixrock:
@miskun as I remember correct we have had this behaviour in the past, right?

Screenshot 2021-02-16 at 18 07 41

### DIFF
--- a/src/renderer/components/kube-object/kube-object-details.tsx
+++ b/src/renderer/components/kube-object/kube-object-details.tsx
@@ -13,11 +13,21 @@ import { CrdResourceDetails } from "../+custom-resources";
 import { KubeObjectMenu } from "./kube-object-menu";
 import { kubeObjectDetailRegistry } from "../../api/kube-object-detail-registry";
 
+/**
+ * Used to store `object.selfLink` to show more info about resource in the details panel.
+ */
 export const kubeDetailsUrlParam = createPageParam({
   name: "kube-details",
   isSystem: true,
 });
 
+/**
+ * Used to highlight last active/selected table row with the resource.
+ *
+ * @example
+ * If we go to "Nodes (page) -> Node (details) -> Pod (details)",
+ * last clicked Node should be "active" while Pod details are shown).
+ */
 export const kubeSelectedUrlParam = createPageParam({
   name: "kube-selected",
   isSystem: true,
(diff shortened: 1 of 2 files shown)


Nokel81 (approved):

jim-docker (approved):

src/renderer/components/kube-object/kube-object-details.tsx
```diff
@@ -36,18 +46,18 @@ export function hideDetails() {
   showDetails();
 }
 
-export function getDetailsUrl(details: string, resetSelected = false) {
-  const detailsUrl = kubeDetailsUrlParam.toSearchString({ value: details });
+export function getDetailsUrl(selfLink: string, resetSelected = false, mergeGlobals = true) {
```
aleksfront, line 49: Function get much harder to understand with this triadic arguments. Is it possible to avoid `mergeGlobals` at least. And it's always `true` as I see.

src/renderer/components/kube-object/kube-object-details.tsx
```diff
@@ -36,18 +46,18 @@ export function hideDetails() {
...
 
-  if (resetSelected) {
-    const params = new URLSearchParams(detailsUrl);
-
-    params.delete(kubeSelectedUrlParam.name);
+  params.set(kubeDetailsUrlParam.urlName, selfLink);
 
-    return `?${params.toString()}`;
+  if (resetSelected) {
+    params.delete(kubeSelectedUrlParam.urlName);
```
aleksfront, line 55: Is it a bit awkward that we set a `param` and than delete it under condition? It it possible to move first `set` under condition in some way too?

src/renderer/components/kube-object/kube-object-details.tsx
ixrock: Yes, currently it's always `true`. 
I would prefer to extend the function as `getDetailsUrl(selfLink: string, {resetSelected = false, mergeGlobals = true} = {})` but it would be _breaking change_ since it's exported to extensions-api.

src/renderer/components/kube-object/kube-object-details.tsx
ixrock: We set _kube**Details**UrlParam_, but updating _different_ param under condition: _kube**Selected**UrlParam_.

aleksfront: 👍

src/renderer/components/kube-object/kube-object-details.tsx
Nokel81: It is not necessarily a breaking change since you could do `getDetailsUrl(selfLink: string, opts: boolean | DetailsUrlOpts = false)`

PR #2167, opened 2021-02-17
Display environment variables coming from secret in pod details

### LINKED ISSUE #2118: Secrets not shown in Pod Detail Panel
**Describe the bug**
I used to be able to see secrets that are loaded into a pod in the pod detail panel. Albeit, they were hidden and only revealed after an additional click.
Now I don't see them anymore at all although they are correctly loaded into the pod.

**To Reproduce**
1. Inject a secret into a pod with
```
          envFrom:
            - secretRef:
                name: my-secret
```
2. Click on the pod detail view
3. Look at the `Environment` section
4. See that there is no secret

**Expected behavior**
Used to be able to see the secrets (albeit garbled).

- Lens Version: 4.0.7 (also happens on edge version `4.1.0-beta.2`)
- OS: Ubuntu 20.04
- Installation method: snap

nevalla:
Currently if environment variable defined in pod's container spec has secret reference in `valueFrom` property, secret is displayed in pod details. Also, if pod's container spec has `envFrom` property, only variables from configMap reference are displayed. 

This PR will render also variables that are coming from secret reference.

image

image

Fixes #2118

### DIFF
--- a/src/renderer/components/+workloads-pods/pod-container-env.tsx
+++ b/src/renderer/components/+workloads-pods/pod-container-env.tsx
@@ -30,7 +30,11 @@ export const ContainerEnvironment = observer((props: Props) => {
           }
         });
         envFrom && envFrom.forEach(item => {
-          const { configMapRef } = item;
+          const { configMapRef, secretRef } = item;
+
+          if (secretRef && secretRef.name) {
+            secretsStore.load({ name: secretRef.name, namespace });
+          }
 
           if (configMapRef && configMapRef.name) {
             configMapsStore.load({ name: configMapRef.name, namespace });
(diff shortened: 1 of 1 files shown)


aleksfront (approved):

jakolehm (approved):

src/renderer/components/+workloads-pods/pod-container-env.tsx
```diff
@@ -30,7 +30,11 @@ export const ContainerEnvironment = observer((props: Props) => {
           }
         });
         envFrom && envFrom.forEach(item => {
-          const { configMapRef } = item;
+          const { configMapRef, secretRef } = item;
+
+          if (secretRef && secretRef.name) {
```
Nokel81, line 35:
```suggestion
          if (secretRef?.name) {
```

src/renderer/components/+workloads-pods/pod-container-env.tsx
```diff
@@ -89,19 +93,52 @@ export const ContainerEnvironment = observer((props: Props) => {
...
+      <div className="variable" key={name}>
+        <span className="var-name">{name}</span>: {value}
+      </div>
+    ));
+  };
+
+  const renderEnvFromSecret = (secretName: string) => {
+    const secret = secretsStore.getByName(secretName, namespace);
+
+    if (!secret) return;
```
Nokel81, line 121: Shouldn't we render the name only here?

nevalla: Maybe having `envFrom` section in pod details where secretKeyRef is rendered, so not to mix it with environment variables. But that can be done in follow-up PR if we feel it's a good idea.

PR #2168, opened 2021-02-17
Windows: use SHELL for terminal if set

### LINKED ISSUE #2087: Can't change PTYSHELL to wsl.exe, powershell.exe is hardcoded?
**Describe the bug**
I use wsl for kubernetes tools. I don't have even kubectl for Windows installed.

I try to override SHELL or PTYSHELL environment variable to `wsl.exe` and still `powershell.exe` is spawned.

**To Reproduce**
1. Go to pods list
2. Click on 'Pod shell'

**Expected behavior**
bash should be started

- Lens Version: 4.0.8
- OS: Windows 10 with WSL2 and Ubuntu 20.04

jakolehm:
Fixes #2087

### DIFF
--- a/src/main/shell-session.ts
+++ b/src/main/shell-session.ts
@@ -108,7 +108,7 @@ export class ShellSession extends EventEmitter {
 
     if(isWindows) {
       env["SystemRoot"] = process.env.SystemRoot;
-      env["PTYSHELL"] = "powershell.exe";
+      env["PTYSHELL"] = process.env.SHELL || "powershell.exe";
       env["PATH"] = pathStr;
     } else if(typeof(process.env.SHELL) != "undefined") {
       env["PTYSHELL"] = process.env.SHELL;


nevalla (approved):

ixrock (approved):

src/main/shell-session.ts
```diff
@@ -108,7 +108,7 @@ export class ShellSession extends EventEmitter {
 
     if(isWindows) {
       env["SystemRoot"] = process.env.SystemRoot;
-      env["PTYSHELL"] = "powershell.exe";
+      env["PTYSHELL"] = process.env.SHELL || "powershell.exe";
```
Nokel81, line 111: Shouldn't we only optionally override `PTYSHELL`?
```suggestion
      env["PTYSHELL"] ||= process.env.SHELL || "powershell.exe";
```

jakolehm: Maybe, I tried to keep this in line with existing logic. We could open an issue about this if needed.

PR #2173, opened 2021-02-17
Add notification to user to add accessible namespaces

### LINKED ISSUE #1981: Suggest to user to add accessible namespaces
If Lens detects that a user does not have list namespaces RBAC permissions. It should display a notification that suggests to the user to add the namespaces that they have any access to.

**Why is this needed**:
We are getting a lot of bug reports regarding bug reports around not being able to list items (e.g. Pods) in 4.X because we have fixed bugs around RBAC. This should help users know how to help Lens work properly.

Nokel81:
- If cluster's refresh fails due to permissions

- Broadcasting event and then receiving it in renderer

fixes #1981 

This should also be a good nudge for users when a cluster doesn't/no longer shows any info.

### DIFF
--- a/src/common/ipc/cluster.ipc.ts
+++ b/src/common/ipc/cluster.ipc.ts
@@ -6,6 +6,6 @@ export const ClusterListNamespaceForbiddenChannel = "cluster:list-namespace-forb
 
 export type ListNamespaceForbiddenArgs = [clusterId: string];
 
-export function argArgsListNamespaceFordiddenArgs(args: unknown[]): args is ListNamespaceForbiddenArgs {
+export function isListNamespaceForbiddenArgs(args: unknown[]): args is ListNamespaceForbiddenArgs {
   return args.length === 1 && typeof args[0] === "string";
 }
--- a/src/common/ipc/cluster.ipc.ts
+++ b/src/common/ipc/cluster.ipc.ts
@@ -1,0 +1,11 @@
+/**
+ * This channel is broadcast on whenever the cluster fails to list namespaces
+ * during a refresh and no `accessibleNamespaces` have been set.
+ */
+export const ClusterListNamespaceForbiddenChannel = "cluster:list-namespace-forbidden";
+
+export type ListNamespaceForbiddenArgs = [clusterId: string];
+
+export function argArgsListNamespaceFordiddenArgs(args: unknown[]): args is ListNamespaceForbiddenArgs {
+  return args.length === 1 && typeof args[0] === "string";
+}
--- a/src/renderer/components/+cluster-settings/components/cluster-accessible-namespaces.tsx
+++ b/src/renderer/components/+cluster-settings/components/cluster-accessible-namespaces.tsx
@@ -16,7 +16,7 @@ export class ClusterAccessibleNamespaces extends React.Component<Props> {
   render() {
     return (
       <>
-        <SubTitle title="Accessible Namespaces" />
+        <SubTitle title="Accessible Namespaces" id="accessible-namespaces" />
         <p>This setting is useful for manually specifying which namespaces you have access to. This is useful when you do not have permissions to list namespaces.</p>
         <EditableList
           placeholder="Add new namespace..."
--- a/src/renderer/components/app.scss
+++ b/src/renderer/components/app.scss
@@ -50,8 +50,8 @@
   color: white;
 }
 
-:target {
-  color: $targetColor;
+*:target {
+  color: $textColorAccent;
 }
 
 html {
--- a/src/renderer/components/app.scss
+++ b/src/renderer/components/app.scss
@@ -50,6 +50,10 @@
   color: white;
 }
 
+:target {
+  color: $targetColor;
+}
+
 html {
   font-size: 62.5%; // 1 rem == 10px
   color: $textColorPrimary;
--- a/src/renderer/ipc/index.tsx
+++ b/src/renderer/ipc/index.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { ipcRenderer, IpcRendererEvent } from "electron";
-import { areArgsUpdateAvailableFromMain, UpdateAvailableChannel, onCorrect, UpdateAvailableFromMain, BackchannelArg, ClusterListNamespaceForbiddenChannel, argArgsListNamespaceFordiddenArgs, ListNamespaceForbiddenArgs } from "../../common/ipc";
+import { areArgsUpdateAvailableFromMain, UpdateAvailableChannel, onCorrect, UpdateAvailableFromMain, BackchannelArg, ClusterListNamespaceForbiddenChannel, isListNamespaceForbiddenArgs, ListNamespaceForbiddenArgs } from "../../common/ipc";
 import { Notifications, notificationsStore } from "../components/notifications";
 import { Button } from "../components/button";
 import { isMac } from "../../common/vars";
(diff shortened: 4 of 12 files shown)


src/common/ipc/cluster/index.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * This channel is broadcast on whenever the cluster fails to list namespaces
+ * during a refresh and no `accessibleNamespaces` have been set.
+ */
+export const ClusterListNamespaceForbiddenChannel = "cluster:list-namespace-forbidden";
+
+export type ListNamespaceFordiddenArgs = [clusterId: string];
```
Nokel81: I recently learned that you can do this for tuple types. The actual type is still just `[string]` but it is a good documentation in code.

src/common/ipc/cluster/index.ts
```diff
@@ -0,0 +1,11 @@
+/**
```
ixrock: We already have `common/cluster-ipc.ts`. How it differs from that?

Also if you want to have dedicated folder for `ipc` I would prefer to name files as you put for parent folders, instead of `index.ts` => gives better direct navigation/search from IDE.

So the final files would be:
```
common/ipc/cluster.ipc.ts
common/ipc/update-available.ipc.ts
```

Wdyt?

src/common/ipc/cluster/index.ts
ixrock: You have a typo in _ListNamespaceFor**d**iddenArgs_

src/common/ipc/cluster/index.ts
```diff
@@ -0,0 +1,11 @@
...
+ * This channel is broadcast on whenever the cluster fails to list namespaces
+ * during a refresh and no `accessibleNamespaces` have been set.
+ */
+export const ClusterListNamespaceForbiddenChannel = "cluster:list-namespace-forbidden";
```
ixrock: This file contains parts of `type-enforced-ipc.ts`, right?
Is there any special reason why splitting those? Can we wrap it in single "action" an export to just call outside?
Also I think `onCorrect` is too way abstract in general.

ixrock (dismissed):

src/common/ipc/cluster/index.ts
Nokel81: Much better layout, good idea.

src/common/ipc/cluster/index.ts
Nokel81: Thanks

src/renderer/components/+cluster-settings/components/cluster-accessible-namespaces.tsx
```diff
@@ -16,7 +16,7 @@ export class ClusterAccessibleNamespaces extends React.Component<Props> {
   render() {
     return (
       <>
-        <SubTitle title="Accessible Namespaces" />
+        <SubTitle title="Accessible Namespaces" id="accessible-namespaces" />
```
ixrock, line 19: Btw, you can highlight element with CSS when `#accessible-namespaces` in the URL (easier to spot the block after scrolling).

_Example:_

image

Nokel81: Ah excellent, great solution. I had thought that this would be a good idea but didn't know about this solution (I thought it would be a large hassle so I didn't try) 👍

src/common/ipc/cluster/index.ts
Nokel81: Do you mean instead of the raw broadcast and receiving pair being exported, just other items? 

What do you mean exactly bu `onCorrect` (and I assume `onceCorrect` too) are way too abstract? Yeah the types assume that you ignore the first arg but other than that they do the type checking that such an call should have done to it.

Nokel81: I think a different PR is in order for that work to be done, since I don't really see it as part of this PR. But yes I think I can rearrange the code to be easier.

nevalla:
Any ideas what is causing extensions' `package-lock.json` changes?

src/renderer/ipc/index.tsx
```diff
@@ -51,11 +55,53 @@ function UpdateAvailableHandler(event: IpcRendererEvent, ...[backchannel, update
...
+    return;
+  }
+
+  const notificationId = `list-namespaces-forbidden:${clusterId}`;
+
+  Notifications.info(
+    (
+      <div className="flex column gaps">
+        <b>Add Accessible Namespaces</b>
+        <p>Cluster <b>{clusterStore.active.name}</b> does not have permissions to list namespaces. Please add the namespaces you have access to.</p>
```
nevalla, line 79: How about 


> <p>Cluster <b>{clusterStore.active.name}</b> does not have permissions to list namespaces. Please add the namespaces you have access to in cluster settings.</p>

And button text could then simply be `Add` ? Does it make it any better?

Nokel81:
@nevalla Thanks for noticing the `package-lock.json` changes. For me they get updated whenever I run `make dev` or `make build`. I try to always do a `git restore extensions` before committing but I guess I didn't the most recent time.

src/renderer/ipc/index.tsx
Nokel81: I don't know if it makes it better because the button doesn't bring up a dialog but instead it navigates the user to the cluster settings page, scrolls down to the `Accessible Namespaces` section and makes that section easier to find.

src/common/ipc/cluster.ipc.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * This channel is broadcast on whenever the cluster fails to list namespaces
+ * during a refresh and no `accessibleNamespaces` have been set.
+ */
+export const ClusterListNamespaceForbiddenChannel = "cluster:list-namespace-forbidden";
+
+export type ListNamespaceForbiddenArgs = [clusterId: string];
+
+export function argArgsListNamespaceFordiddenArgs(args: unknown[]): args is ListNamespaceForbiddenArgs {
```
aleksfront: I understand it is some form of convention for naming such functions, but there're too many `args` in this name? I believe function name should be easy to pronounce. How about `hasForbiddenArgs`?

src/renderer/themes/theme-vars.scss
```diff
@@ -21,6 +21,7 @@ $colorTerminated: var(--colorTerminated);
 // Layout
 $mainBackground: var(--mainBackground);
 $contentColor: var(--contentColor);
+$targetColor: var(--targetColor);
```
aleksfront: We can use `--accentColor`, it has same color values you provided.

aleksfront (commented):
Minor thing but it doesn't highlight active path after clicking `Cluster Settings` in notification. But it scrolls down to `Accessible Namespaces` though.
cluster settings

src/common/ipc/cluster.ipc.ts
Nokel81: lol, sure I meant to call this `are...` but yeah it is a mouthful. Your name is better.

Nokel81:
@aleksfront Fixed the highlighting issue, changed to using `textAccentColor`

aleksfront (dismissed):
LGTM

jim-docker (dismissed):

src/renderer/components/app.scss
```diff
@@ -50,6 +50,10 @@
   color: white;
 }
 
+*:target {
```
ixrock, line 53: Probably `*` won't needed in that case.

```suggestion
:target {
```

ixrock (dismissed):

src/renderer/components/app.scss
Nokel81: I tested and I did need it.

nevalla (approved):

ixrock (approved):

PR #2181, opened 2021-02-18
Trigger quitAndInstall after update-downloaded event

### LINKED ISSUE #2179: Auto-update fails
**Describe the bug**

If user chooses "Yes, now" the auto-update process fails in the background with following error:

```
warn: No update available, can't quit and install
error: Error: Error: No update available, can't quit and install
    at MacUpdater.quitAndInstall (/Applications/Lens.app/Contents/Resources/app.asar/node_modules/electron-updater/out/MacUpdater.js:185:24)
    at /Applications/Lens.app/Contents/Resources/app.asar/static/build/main.js:1:108165
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:94:5)
2021-02-18 11:34:43.589 Lens[22193:1679199] Download completed to: file:///Users/foo/Library/Caches/com.electron.kontena-lens.ShipIt/update.4wp8sht/1613640882973-4213.zip
```

jakolehm:
Should fix #2179

### DIFF
--- a/src/main/app-updater.ts
+++ b/src/main/app-updater.ts
@@ -9,9 +9,8 @@ function handleAutoUpdateBackChannel(event: Electron.IpcMainEvent, ...[arg]: Upd
   if (arg.doUpdate) {
     if (arg.now) {
       logger.info(`${AutoUpdateLogPrefix}: User chose to update now`);
-      autoUpdater.downloadUpdate()
-        .then(() => autoUpdater.quitAndInstall())
-        .catch(error => logger.error(`${AutoUpdateLogPrefix}: Failed to download or install update`, { error }));
+      autoUpdater.on("update-downloaded", () => autoUpdater.quitAndInstall());
+      autoUpdater.downloadUpdate().catch(error => logger.error(`${AutoUpdateLogPrefix}: Failed to download or install update`, { error }));
     } else {
       logger.info(`${AutoUpdateLogPrefix}: User chose to update on quit`);
       autoUpdater.autoInstallOnAppQuit = true;


src/main/app-updater.ts
```diff
@@ -9,9 +9,8 @@ function handleAutoUpdateBackChannel(event: Electron.IpcMainEvent, ...[arg]: Upd
   if (arg.doUpdate) {
     if (arg.now) {
       logger.info(`${AutoUpdateLogPrefix}: User chose to update now`);
-      autoUpdater.downloadUpdate()
-        .then(() => autoUpdater.quitAndInstall())
-        .catch(error => logger.error(`${AutoUpdateLogPrefix}: Failed to download or install update`, { error }));
+      autoUpdater.on("update-downloaded", () => autoUpdater.quitAndInstall());
```
panuhorsmalahti, line 12: We could also listen for `"error"` event and log those.

panuhorsmalahti (approved):

src/main/app-updater.ts
ixrock, line 12: @jakolehm how it actually fixing the issue? `autoUpdater.quitAndInstall().then(quitAndInstall)` is never called? 
or how it differs from `autoUpdater.on("update-downloaded", quitAndInstall)` event-call?

ixrock (approved):

src/main/app-updater.ts
jakolehm: `autoUpdater.quitAndInstall().then(quitAndInstall)` calls too fast (it does not find the update from filesystem). Docs say that it should be called _after_ `update-downloaded` event.

Nokel81: We should

Issue #2166, opened 2021-02-17
Persist the last selected namespace(s)


mboutet:
**What would you like to be added**:

Persist the last selected namespace(s) (including "All namespaces") in order for the choice to be kept between restarts and reloads. This is a only a problem since v4.1.0.

If this is too difficult to implement, an alternative could be to set the namespace(s) to select in the cluster's settings.

**Why is this needed**:

My network connection sometimes momentarily drops resulting in a reload of Lens and the loss of the selected namespaces. This can lead to missed events in the cluster since they are not shown when in the default namespace. 

Ever since 4.1.1, I find myself constantly making sure that the correct namespaces are selected. (Of course, I have other monitoring and alerting in place. However, there can be delays in the alerts and some unforeseen events might not be covered by them).

**Environment you are Lens application on:**

- Kubernetes distribution: AKS
- Desktop OS: macOS 10.15.7 (19H15)

matthewmgr:
Please, to make this application useful again, MAKE THIS CHANGE.  I monitor, literally, 30+ clusters and every time I toggle between clusters I have to select, for every area, "All Namespaces" because we intentionally do not use default. 

Alternately...  and actually preferred for me, default to "All Namespaces".  Why would you limit what you're viewing?

rittneje:
Agreed, I feel this change in 4.1.0 is a complete misfeature. It's pretty rare that we would even put something in the default namespace. People who want to see only the default namespace should be forced to explicitly select it in the first place, and then it could be persisted.

rflume:
Wouldn't it probably be best to do both:
1. Set "All namespaces" as default value ...
2. ... **and** add a user option to persist last selected namespace(s per area)?

PR #2185, opened 2021-02-18
Fix: don't reset selected namespaces to defaults in case of "All namespaces" on page reload

ixrock:
_Fixes:_

- fix: don't reset selected namespaces in case of "All namespaces" at page reload (Cmd+R)
- fix: don't make k8s api-request(s) for loading namespaces on every page visit / `NamespaceSelect.componentDidMount()`
- partial-fix for #2166

https://user-images.githubusercontent.com/6377066/108360447-e3f1c480-71f9-11eb-8cea-f20ffb41520b.mov

### DIFF
--- a/src/renderer/components/+namespaces/namespace.store.ts
+++ b/src/renderer/components/+namespaces/namespace.store.ts
@@ -5,7 +5,7 @@ import { Namespace, namespacesApi } from "../../api/endpoints/namespaces.api";
 import { createPageParam } from "../../navigation";
 import { apiManager } from "../../api/api-manager";
 
-const storage = createStorage<string[]>("context_namespaces", []);
+const storage = createStorage<string[]>("context_namespaces");
 
 export const namespaceUrlParam = createPageParam<string[]>({
   name: "namespaces",
@@ -74,11 +74,11 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
   @computed
   private get initialNamespaces(): string[] {
     const namespaces = new Set(this.allowedNamespaces);
-    const prevSelected = storage.get().filter(namespace => namespaces.has(namespace));
+    const prevSelectedNamespaces = storage.get();
 
-    // return previously saved namespaces from local-storage
-    if (prevSelected.length > 0) {
-      return prevSelected;
+    // return previously saved namespaces from local-storage (if any)
+    if (prevSelectedNamespaces) {
+      return prevSelectedNamespaces.filter(namespace => namespaces.has(namespace));
     }
 
     // otherwise select "default" or first allowed namespace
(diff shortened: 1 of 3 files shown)


Nokel81 (approved):

mboutet:
Will this fix https://github.com/lensapp/lens/issues/2166?

Nokel81:
@mboutet Only for a single run of Lens, not between closing of windows or restarting Lens

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -5,7 +5,7 @@ import { Namespace, namespacesApi } from "../../api/endpoints/namespaces.api";
 import { createPageParam } from "../../navigation";
 import { apiManager } from "../../api/api-manager";
 
-const storage = createStorage<string[]>("context_namespaces", []);
+const storage = createStorage<string[]>("context_namespaces");
```
jim-docker, line 8: If this were maintained in the active cluster it could persist across Lens restarts?

jim-docker (approved):

src/renderer/components/+namespaces/namespace.store.ts
ixrock: For some reasons saving data to _localStorage_ **doesn't work** and after app restarts _localStorage_ is empty.
Probably need to investigate  `electronjs` docs / tweak some configuration for  [BrowserWindow](https://www.electronjs.org/docs/api/browser-window) or similar. 
@jakolehm any ideas?

Nokel81: Does `localStorage` persist between quitting the final browser instance and starting a new one (I mean for a normal browser)?

ixrock: Tried this, doesn't work:

```ts
      this.mainWindow = new BrowserWindow({
       // ...
        webPreferences: {
          partition: "persist:lens-app",
        },
      });
```

Probably we have to fallback to using plain json files as in `base-store.ts` (`conf` package).

Nokel81: And remove all the instances of `createStorage`? Or create a new abstraction that syncs with both?

jakolehm: > For some reasons saving data to _localStorage_ **doesn't work** and after app restarts _localStorage_ is empty.
> Probably need to investigate `electronjs` docs / tweak some configuration for [BrowserWindow](https://www.electronjs.org/docs/api/browser-window) or similar.
> @jakolehm any ideas?

It seems to persist if tray stays alive... so my guess is that localStorage is wiped because we get a new (random) port on boot.

PR #2188, opened 2021-02-18
add lock around lensViews in initView

Nokel81:

### DIFF
--- a/src/renderer/components/cluster-manager/lens-views.ts
+++ b/src/renderer/components/cluster-manager/lens-views.ts
@@ -2,6 +2,7 @@ import { observable, when } from "mobx";
 import { ClusterId, ClusterStore, getClusterFrameUrl } from "../../../common/cluster-store";
 import { getMatchedClusterId } from "../../navigation";
 import logger from "../../../main/logger";
+import AwaitLock from "await-lock";
 
 export interface LensView {
   isLoaded?: boolean
@@ -10,20 +11,19 @@ export interface LensView {
 }
 
 export const lensViews = observable.map<ClusterId, LensView>();
+const lensViewsLock = new AwaitLock();
 
 export function hasLoadedView(clusterId: ClusterId): boolean {
   return !!lensViews.get(clusterId)?.isLoaded;
 }
 
 export async function initView(clusterId: ClusterId) {
+  await lensViewsLock.acquireAsync();
+
   if (!clusterId || lensViews.has(clusterId)) {
     return;
   }
   const cluster = ClusterStore.getInstance().getById(clusterId);
-
-  if (!cluster) {
-    return;
-  }
 
   logger.info(`[LENS-VIEW]: init dashboard, clusterId=${clusterId}`);
   const parentElem = document.getElementById("lens-views");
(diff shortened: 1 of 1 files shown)


src/renderer/components/cluster-manager/lens-views.ts
```diff
@@ -10,36 +11,43 @@ export interface LensView {
...
+
+    iframe.name = cluster.contextName;
+    iframe.setAttribute("src", getClusterFrameUrl(clusterId));
+    iframe.addEventListener("load", () => {
+      logger.info(`[LENS-VIEW]: loaded from ${iframe.src}`);
+      lensViews.get(clusterId).isLoaded = true;
+    }, { once: true });
+    lensViews.set(clusterId, { clusterId, view: iframe });
+    parentElem.appendChild(iframe);
+    await autoCleanOnRemove(clusterId, iframe);
```
jakolehm: Shouldn't we release lock before this?

Nokel81: Or we should just not `await` here (which is probably what we should do instead)

Nokel81: Fixed

src/renderer/components/cluster-manager/lens-views.ts
```diff
@@ -10,21 +11,20 @@ export interface LensView {
...
 
 export function hasLoadedView(clusterId: ClusterId): boolean {
   return !!lensViews.get(clusterId)?.isLoaded;
 }
 
 export async function initView(clusterId: ClusterId) {
+  await lensViewsLock.acquireAsync();
+
   if (!clusterId || lensViews.has(clusterId)) {
     return;
```
jim-docker: shouldn't the lock be released here before returning? Maybe use try block for all the code in this function, with the release done in finally?
RAII would be nice for releasing the lock :)

jim-docker (commented):
Is this just bulletproofing or is there a known issue you think this might fix?

Nokel81:
This was a potential fix that #2190 also fixed. So currently just bulletproofing. Though I agree with your suggestion.

src/renderer/components/cluster-manager/lens-views.ts
```diff
@@ -2,6 +2,8 @@ import { observable, when } from "mobx";
 import { ClusterId, clusterStore, getClusterFrameUrl } from "../../../common/cluster-store";
 import { getMatchedClusterId } from "../../navigation";
 import logger from "../../../main/logger";
+import AwaitLock from "await-lock";
```
ixrock, line 5: Can you explain in short what it does and why it's needed for lens-views? (it's not obvious and there is no description in PR)

Nokel81: We were observing a race condition (the same as was also fixed with #2190) and this was my proposed solution. I still think it is relevant because we don't really want to have to think about mutable mutable access.

ixrock: Thanks, but I still can't fully get why it might happen? Which process modifies what which leads to race condition?

- `lens-views` is running in single process in main frame of `BrowserWindow`
- `initView` is called once on change `getMatchedClusterId()` and saves view to the map. where is race?

If `initView()` would be a sync function would change it something?

Nokel81: I forget exactly what was causing it, but you could easily get it to happen if any of the async calls took a long time (for instance if we added a 5s delay). It would show multiple clusters at once.

Nokel81:
Closing because we are probably going to remove iframes.

PR #2190, opened 2021-02-18
Fix duplicate cluster iframes

jakolehm:
Fixes regression in #2183

### DIFF
--- a/src/renderer/components/cluster-manager/lens-views.ts
+++ b/src/renderer/components/cluster-manager/lens-views.ts
@@ -25,8 +25,6 @@ export async function initView(clusterId: ClusterId) {
     return;
   }
 
-  logger.info(`[LENS-VIEW]: waiting cluster to be ready, clusterId=${clusterId}`);
-  await cluster.whenReady;
   logger.info(`[LENS-VIEW]: init dashboard, clusterId=${clusterId}`);
   const parentElem = document.getElementById("lens-views");
   const iframe = document.createElement("iframe");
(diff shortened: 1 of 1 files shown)


src/renderer/components/cluster-manager/lens-views.ts
```diff
@@ -39,6 +37,8 @@ export async function initView(clusterId: ClusterId) {
   }, { once: true });
   lensViews.set(clusterId, { clusterId, view: iframe });
   parentElem.appendChild(iframe);
+  logger.info(`[LENS-VIEW]: waiting cluster to be ready, clusterId=${clusterId}`);
+  await cluster.whenReady;
```
jim-docker, line 41: what happens if the cluster is never ready? We have an async function that never resolves?

nevalla (approved):

jim-docker (approved):
Double dashboard is gone, but I did manage to lockup the UI. Probably not related, requires hardening via a separate PR?

https://user-images.githubusercontent.com/40840436/108386769-f6103a80-71da-11eb-9208-93db6a6eb2cf.mov

src/renderer/components/cluster-manager/lens-views.ts
ixrock: @jim-docker why it could be never ready? These views are created when you first time opens a cluster, so it should be ready at some point?

jim-docker: If there's a bug? I was just concerned that if `InitView()` blocks here because of a bug what will the user see?

PR #2193, opened 2021-02-19
Fix NamespaceSelectFilter to correctly track accessibleNamespaces

### LINKED ISSUE #2111: Accessible namespaces functionality is broken in 4.1.0-beta2

Nokel81:
- Consolidate the ClusterContext functions into a class

- Add method for the correct determiation of whether all possible namespaces are selected to ClusterContext

- Use said method in the two API related use cases as well in the above component

fixes #2111 (hopefully for the last time)

### DIFF
--- a/src/renderer/api/kube-watch-api.ts
+++ b/src/renderer/api/kube-watch-api.ts
@@ -6,7 +6,7 @@ import type { ClusterContext } from "../components/context";
 
 import plimit from "p-limit";
 import { comparer, IReactionDisposer, observable, reaction, when } from "mobx";
-import { autobind, noop } from "../utils";
+import { asTuple, autobind, noop } from "../utils";
 import { KubeApi } from "./kube-api";
 import { KubeJsonApiData } from "./kube-json-api";
 import { isDebugging, isProduction } from "../../common/vars";
@@ -88,7 +88,15 @@ export class KubeWatchApi {
       }
 
       // reload stores only for context namespaces change
-      cancelReloading = reaction(() => this.context?.contextNamespaces, namespaces => {
+      cancelReloading = reaction(() => {
+        const namespaces = this.context?.selectedNamespaces;
+
+        /**
+         * react to the changing of "allPossibleNamespaces" so that adding
+         * accessibleNamespaces means that this is restarted
+         */
+        return asTuple([this.context?.isAllPossibleNamespaces(namespaces), namespaces]);
+      }, ([, namespaces]) => {
         preloading?.cancelLoading();
         unsubscribeList.forEach(unsubscribe => unsubscribe());
         unsubscribeList.length = 0;
--- a/src/renderer/components/+namespaces/namespace-select.tsx
+++ b/src/renderer/components/+namespaces/namespace-select.tsx
@@ -11,16 +11,31 @@ import { kubeWatchApi } from "../../api/kube-watch-api";
 import { components, ValueContainerProps } from "react-select";
 
 interface Props extends SelectProps {
+  /**
+   * Show icons preceeding the entry names
+   * @default true
+   */
   showIcons?: boolean;
-  showClusterOption?: boolean; // show "Cluster" option on the top (default: false)
-  showAllNamespacesOption?: boolean; // show "All namespaces" option on the top (default: false)
+
+  /**
+   * show a "Cluster" option above all namespaces
+   * @default false
+   */
+  showClusterOption?: boolean;
+
+  /**
+   * show "All namespaces" option on the top (has precedence over `showClusterOption`)
+   * @default false
+   */
+  showAllNamespacesOption?: boolean;
+
+  /**
+   * A function to change the options for the select
+   * @param options the current options to display
+   * @default passthrough
+   */
   customizeOptions?(options: SelectOption[]): SelectOption[];
 }
-
-const defaultProps: Partial<Props> = {
-  showIcons: true,
-  showClusterOption: false,
-};
 
 function GradientValueContainer<T>({children, ...rest}: ValueContainerProps<T>) {
   return (
@@ -34,7 +49,12 @@ function GradientValueContainer<T>({children, ...rest}: ValueContainerProps<T>) 
 
 @observer
 export class NamespaceSelect extends React.Component<Props> {
-  static defaultProps = defaultProps as object;
+  static defaultProps: Props = {
+    showIcons: true,
+    showClusterOption: false,
+    showAllNamespacesOption: false,
+    customizeOptions: (opts) => opts,
+  };
 
   componentDidMount() {
     disposeOnUnmount(this, [
@@ -47,7 +67,7 @@ export class NamespaceSelect extends React.Component<Props> {
 
   @computed.struct get options(): SelectOption[] {
     const { customizeOptions, showClusterOption, showAllNamespacesOption } = this.props;
-    let options: SelectOption[] = namespaceStore.items.map(ns => ({ value: ns.getName() }));
+    const options: SelectOption[] = namespaceStore.allowedNamespaces.map(ns => ({ value: ns }));
 
     if (showAllNamespacesOption) {
       options.unshift({ label: "All Namespaces", value: "" });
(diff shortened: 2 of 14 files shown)


src/renderer/components/+namespaces/namespace-select.tsx
```diff
@@ -10,20 +10,40 @@ import { namespaceStore } from "./namespace.store";
...
-  showAllNamespacesOption?: boolean; // show "All namespaces" option on the top (default: false)
+
+  /**
+   * show a "Cluster" option above all namespaces
+   * @default false
+   */
+  showClusterOption?: boolean;
+
+  /**
+   * how "All namespaces" option on the top (has precidence over `showClusterOption`)
```
jim-docker:
```suggestion
   * show "All namespaces" option on the top (has precedence over `showClusterOption`)
```

jim-docker (commented):
Something is fishy, can't pick multiple namespaces:

https://user-images.githubusercontent.com/40840436/108559320-2b935180-72c9-11eb-8da0-8ff4ea879e51.mov

Nokel81:
Thanks for catching this, please re-review

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -33,7 +33,7 @@ export function getDummyNamespace(name: string) {
 export class NamespaceStore extends KubeObjectStore<Namespace> {
   api = namespacesApi;
 
-  @observable private contextNs = observable.set<string>();
+  @observable private rawSelectedNamespaces = observable.set<string>();
```
nevalla, line 34: `contextNS` used to be public property. After changing it private extensions developer could not use it anymore. However current extensions still did not break. This change will make also current extensions to break if they are relying on this property.

nevalla:
Not sure if it's out of scope of this PR, but tested this with the following flow:
1. Created service account with `edit` rolebinding in `default` namespaces
2. Added service account kubeconfig to Lens
3. Did not see any any namespaces
4. Added `default` to accessible namespaces in cluster settings
5. Default namespace appeared to namespace selector, but still could not see any resources:
image
6. When added the second namespace, resources are displayed 🤔

src/renderer/components/+namespaces/namespace.store.ts
Nokel81: It maybe have been previously `public` but it currently isn't.

Nokel81:
@nevalla by 6 do you mean you selected a second namespace? Currently the watches are not restarted if `accessibleNamespaces` gets changed. I think you are observing that (which I agree, is out of scope for this PR).

nevalla:
No, I meant that when I added another namespace in accessible namespaces, then resources appeared.

src/renderer/components/+namespaces/namespace.store.ts
nevalla: Yes, it's definitely private now, but it's causing build time failure now for extensions developers, existing lens extension instances are still running fine because they are javascript. 

I'm just saying that if we are going to include this PR in 4.1.3 it will then definitely break those extensions that are already installed in Lens. 

For example used here: https://github.com/nevalla/lens-resource-map-extension/blob/master/components/NamespaceSelect.tsx#L11

Nokel81:
It might be related but I think it is out of scope.

Namely, `subscribe` doesn't react to changes to namespaces (specifically `accessibleNamespaces`). I assume that the reason by you saw adding a second namespace immediately start loading was that it coincided with the 30s refresh. 

It might also be a combination that by default the namespaces are not selected at all which defaults to "selecting" `allNamespaces` (which is why coincidentally if you keep adding namespaces to `accessibleNamespaces` they all appear in the namespace filter. 

Finally, there is another bug related to the kube watch items where Lens assumes that `list` permissions implies `watch` permissions.

src/renderer/components/+namespaces/namespace.store.ts
Nokel81: Ah of course, extensions can ignore our visibility modifiers. Given that `NamespaceSelect` is now public I assume you should use that?

Nokel81:
Further information, `subscribeStores` erroneously considered that if `context.selectedNamespaces` didn't change then neither should the `subscribe`'s. However, that didn't take into consideration the checks for not using the "cluster-scope" request when dealing with `accessibleNamespaces`.

Nokel81:
@nevalla Please take another look. I have fixed one of the bugs that are part of what you observed (the rest I think are out of scope for this PR though they are tracked as part of #2209 and #2207)

src/renderer/components/+namespaces/namespace.store.ts
nevalla: I can do a follow-up PR to target this and not to introduce breaking change.

jakolehm: @Nokel81 any reason why we cannot make this PR so that it does not break extensions (runtime)? If not -> please change this to be compatible with existing extensions.

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -98,8 +98,8 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
     ].flat()));
   }
 
-  @computed get contextNamespaces(): string[] {
-    const namespaces = Array.from(this.contextNs);
+  @computed get selectedNamespaces(): string[] {
```
jakolehm, line 116: Breaking change.

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -135,28 +135,27 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
...
 
-  hasContext(namespaces: string | string[]) {
-    return [namespaces].flat().every(namespace => this.contextNs.has(namespace));
+  @computed get selectedAll(): boolean {
+    return this.context?.isAllPossibleNamespaces(Array.from(this.rawSelectedNamespaces), true) ?? false;
   }
 
-  @computed get hasAllContexts(): boolean {
-    return this.contextNs.size === this.allowedNamespaces.length;
+  @computed private get hasAllContexts(): boolean {
```
jakolehm: Breaking change.

src/renderer/api/kube-watch-api.ts
```diff
@@ -88,7 +88,15 @@ export class KubeWatchApi {
...
       // reload stores only for context namespaces change
-      cancelReloading = reaction(() => this.context?.contextNamespaces, namespaces => {
+      cancelReloading = reaction(() => {
+        const namespaces = this.context?.selectedNamespaces;
+
+        /**
+         * react to the changing of "allPossibleNamespaces" so that adding
+         * accessibleNamespaces means that this is restarted
+         */
+        return [this.context?.isAllPossibleNamespaces(namespaces), namespaces] as const;
```
jakolehm: Really hard to understand, could we make it more easy to read?

src/renderer/components/+namespaces/namespace.store.ts
nevalla: could be something like this:
```typescript
/** 
* @depreated use NamespaceStore.selectedNamespaces instead 
*/
get contextNs() {
    return this.selectedNamespaces;
}
```

Nokel81: That is what I was planning on doing @nevalla

Nokel81:
I am going to close this PR. As other work will probably make this code flow simpler and thus this PR won't be needed.

PR #2194, opened 2021-02-19
Allow to define the path of the shell in app preferences

### LINKED ISSUE #482: Support for alternative terminal shells

Can we please add a user preference/setting for a specific terminal so that when "New Terminal Session" is initiated, Lens launches and preloads my kubecontext into my terminal of choice, rather than only the singular preset? Similar to VSCode, an Electron app that does this quite well at providing added flexibility.

**Why is this needed**:

Currently when I open a New Terminal Session window in Lens while on my Windows 10 machine, I am greeted with a PowerShell prompt that has already wired up my `kubectl` integration - very slick. 

image

However, I prefer to interact with `kubectl` from a *nix-style terminal so that I can use goodies like `jq`, piping, et. al. I have WSL2 installed on my machine, but when I launch the bash prompt, I lose my kubecontext that lens had preset for me. 

image

I suspect I would also hit this issue if I:

* Was using Windows 10 and wanted to use `cmd` rather than `PowerShell` 
* Was using bash but wanted to switch to zsh, etc.

- Kubernetes distribution: Docker Enterprise 3.0
- Desktop OS: Windows 10 (Version 10.0.19041.264)

Thanks!

nevalla:
This PR allows user to define the path of the shell used in Lens terminal.

image

Fixes #482


src/main/shell-session.ts
```diff
@@ -105,13 +105,14 @@ export class ShellSession extends EventEmitter {
   protected async getShellEnv() {
     const env = JSON.parse(JSON.stringify(await shellEnv()));
     const pathStr = [this.kubectlBinDir, this.helmBinDir, process.env.PATH].join(path.delimiter);
+    const shell = userStore.preferences.shell && userStore.preferences.shell !== "" ? userStore.preferences.shell : process.env.SHELL;
```
Nokel81:
```suggestion
    const shell = userStore.preferences.shell || process.env.SHELL;
```

src/main/shell-session.ts
```diff
@@ -105,13 +105,14 @@ export class ShellSession extends EventEmitter {
...
 
     if(isWindows) {
       env["SystemRoot"] = process.env.SystemRoot;
-      env["PTYSHELL"] = process.env.SHELL || "powershell.exe";
+      env["PTYSHELL"] = shell || "powershell.exe";
```
Nokel81, line 112: Maybe we should also respect the user's `PTYSHELL` setting?
```suggestion
      env["PTYSHELL"] ||= shell || "powershell.exe";
```

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -130,7 +131,17 @@ export class Preferences extends React.Component {
         <small className="hint">
           Proxy is used only for non-cluster communication.
         </small>
-
+        <h2>Terminal Shell</h2>
+        <Input
+          theme="round-black"
+          placeholder={process.env.SHELL}
```
Nokel81: Unless I am mistaken about this but the `PTYSHELL` should also be considered, right?

```suggestion
          placeholder={process.env.SHELL || process.env.PTYSHELL}
```

src/main/shell-session.ts
nevalla: that would prevent overriding the shell

Nokel81: How so? Or do you mean that if this setting is set we want to override `PTYSHELL` no matter what?

The solution then would be:

```suggestion
      env["PTYSHELL"] = shell || env["PTYSHELL"] || "powershell.exe";
```

nevalla: Yes, that's the purpose.

Nokel81 (dismissed):

src/main/shell-session.ts
nevalla: `PTYSHELL` shell is now taken into an account earlier.

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -130,7 +131,17 @@ export class Preferences extends React.Component {
         <small className="hint">
           Proxy is used only for non-cluster communication.
         </small>
-
+        <h2>Terminal Shell</h2>
+        <Input
+          theme="round-black"
+          placeholder={process.env.SHELL || process.env.PTYSHELL}
```
jim-docker: the placeholder is not editable, is this intended? (e.g. if it is `/bin/bash` and I want it to be `/bin/zsh` I can't use backspace from the end to delete `bash`) I think other inputs have this "feature" too, so maybe it is by design?

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -130,7 +131,17 @@ export class Preferences extends React.Component {
...
+          value={this.shell}
+          onChange={v => this.shell = v}
+          onBlur={() => preferences.shell = this.shell}
+        />
+        <small className="hint">
+          The path of the shell that the terminal uses.
```
jim-docker, line 153: I noticed that this doesn't take effect until already open Lens terminals are closed. A comment about it here might be good (unless that's a bug)

src/main/shell-session.ts
```diff
@@ -105,13 +105,14 @@ export class ShellSession extends EventEmitter {
...
+    const shell = userStore.preferences.shell || process.env.SHELL || process.env.PTYSHELL;
 
     if(isWindows) {
       env["SystemRoot"] = process.env.SystemRoot;
-      env["PTYSHELL"] = process.env.SHELL || "powershell.exe";
+      env["PTYSHELL"] = shell || "powershell.exe";
       env["PATH"] = pathStr;
-    } else if(typeof(process.env.SHELL) != "undefined") {
-      env["PTYSHELL"] = process.env.SHELL;
+    } else if(typeof(shell) != "undefined") {
```
ixrock:
```suggestion
    } else if(shell !== undefined) {
```

since `typeof undefined === "undefined" // true`

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -25,6 +25,7 @@ export class Preferences extends React.Component {
   @observable helmRepos: HelmRepo[] = [];
   @observable helmAddedRepos = observable.map<string, HelmRepo>();
   @observable httpProxy = userStore.preferences.httpsProxy || "";
+  @observable shell = userStore.preferences.shell || "";
```
ixrock, line 29: Why do we need to copy value from `userStore`? Only because of saving on `blur`?
Other things in preferences, like `colorTheme` are done by updating store directly, e.g. see `preferences.tsx:128`.

src/renderer/components/+preferences/preferences.tsx
ixrock: Could be the placeholder empty btw if both of env.vars are `undefined`?

ixrock (dismissed):

src/renderer/components/+preferences/preferences.tsx
nevalla: Yes... this is because Lens terminal is caching shell environment for faster opening. This applies also to cluster workspace dir and proxy settings.

src/renderer/components/+preferences/preferences.tsx
nevalla: Yeah, I guess it's that we don't want to update store on every key press.

src/main/shell-session.ts
nevalla: Changed.

nevalla:
@ixrock added some changes based on your review comments. PTAL.

src/renderer/components/+preferences/preferences.tsx
nevalla: Added default placeholder.

nevalla: > the placeholder is not editable, is this intended? (e.g. if it is `/bin/bash` and I want it to be `/bin/zsh` I can't use backspace from the end to delete `bash`) I think other inputs have this "feature" too, so maybe it is by design?

Yes it's by design.

Nokel81 (dismissed):
Looks better now, though I agree with Jim that we seem to have two different styles of saving preferences to the store. 

Also, I would move the `placeholder` selection code to a different function, but that is minor.

jakolehm (dismissed):

littleccguy:
Thank you @nevalla for taking this on.  The work is greatly appreciated.

Nokel81:
@nevalla Please resolve conflicts

nevalla:
Conflicts resolved @jakolehm @Nokel81 PTAL again.

jakolehm (approved):

src/renderer/components/+preferences/preferences.tsx
ixrock: Should we then add delay for updating other preferences as well? Like here:

``` 
onChange={({ value }: SelectOption) => preferences.colorTheme = value}
```

ixrock (approved):

PR #2198, opened 2021-02-19
Pass Lens wslenvs to terminal session on Windows

### LINKED ISSUE #2180: KUBECONFIG environment-var missing from the Terminal-process
**Describe the bug**
kubectl in terminal have no access to the kubeconfig-file

**To Reproduce**
Steps to reproduce the behavior (on Windows 10 with WSL as shell)
1. export SHELL=wsl.exe
2. Select a cluster and open the Terminal
3. type 'kubectl clutser-info' (or anything else that accesses the cluster)
4. see "unable to access" or similar error

Reason: the KUBECONFIG is not set to the clusters kubeconfig in the terminal-process.

**Expected behavior**
the env var KUBECONFIG pointing to the file the cluster is using

**Screenshots**
(none)

- Lens Version: 4.1.1
- OS: Windows 10 with WSL2
- Installation method: local user in custom folder

**Logs:**
(N/A)

**Kubeconfig:**
(N/A)

**Additional context**
It's great to actually be able to use WSL as the shell, and I know there are some path-translations that need to happen because of using WSL as SHELL, but please fix this issue.

nevalla:
This PR will pass `KUBECONFIG` environment variable to terminal session running on WSL on Windows enabling terminal sessions points to correct Kubernetes cluster automatically. Also `LENS_SESSION` environment variable is set, so user can configure bash init script based on this, for example:

```bash
if [ -n "$LENS_SESSION" ] ; then
    alias kubectl="kubectl.exe"
fi
```

Which will make `kubectl` command to use correct version of kubectl.exe downloaded by Lens.

Fixes https://github.com/lensapp/lens/issues/2180

### DIFF
--- a/src/main/shell-session.ts
+++ b/src/main/shell-session.ts
@@ -110,6 +110,14 @@ export class ShellSession extends EventEmitter {
       env["SystemRoot"] = process.env.SystemRoot;
       env["PTYSHELL"] = process.env.SHELL || "powershell.exe";
       env["PATH"] = pathStr;
+      env["LENS_SESSION"] = "true";
+      const lensWslEnv = "KUBECONFIG/up:LENS_SESSION/u";
+
+      if (process.env.WSLENV != undefined) {
+        env["WSLENV"] = `${process.env["WSLENV"]}:${lensWslEnv}`;
+      } else {
+        env["WSLENV"] = lensWslEnv;
+      }
     } else if(typeof(process.env.SHELL) != "undefined") {
       env["PTYSHELL"] = process.env.SHELL;
       env["PATH"] = pathStr;


jakolehm (dismissed):
Seems to work 👍

nevalla:
This PR is targeted to `feature/customizable_shell_path` so either I need to tweak this branch so it can be merged directly to master or alternatively we can merge this after #2194 is merged.

jakolehm:
> This PR is targeted to `feature/customizable_shell_path` so either I need to tweak this branch so it can be merged directly to master or alternatively we can merge this after #2194 is merged.

Didn't realize that... but this would actually fix a bug so it would make sense to have this as a separate PR (unless it's too big task).

nevalla:
Targeted to `master` branch now. PTAL again.

src/main/shell-session.ts
```diff
@@ -110,6 +110,14 @@ export class ShellSession extends EventEmitter {
...
       env["PTYSHELL"] = process.env.SHELL || "powershell.exe";
       env["PATH"] = pathStr;
+      env["LENS_SESSION"] = "true";
+      const lensWslEnv = "KUBECONFIG/up:LENS_SESSION/u";
+
+      if (process.env.WSLENV != undefined) {
+        env["WSLENV"] = `${process.env["WSLENV"]}:${lensWslEnv}`;
+      } else {
+        env["WSLENV"] = lensWslEnv;
+      }
```
Nokel81, line 120:
```suggestion
      env["WSLENV"] = [
        process.env.WSLENV,
        "KUBECONFIG/up:LENS_SESSION/u"
      ].filter(Boolean).join(":");
```

Nokel81 (commented):
Looks good

jakolehm (approved):

src/main/shell-session.ts
nevalla: Initially I made something like this but ended up to have the current implementation for the sake of more readable code.

Nokel81: okay fair, I find this more readable because there no `if` that has to be deciphered for the differences and each value is treated the same.

Nokel81 (approved):

PR #2204, opened 2021-02-22
Fix RoleBindingDetails not rendering the name of the role binding

Nokel81:
- Move all strings of the form {`...`} to use just ""

- Fix RoleBindingDetails not rendering the name of the role binding

follow up to #1874

### DIFF
--- a/src/renderer/components/+apps-helm-charts/helm-chart-details.tsx
+++ b/src/renderer/components/+apps-helm-charts/helm-chart-details.tsx
@@ -82,9 +82,9 @@ export class HelmChartDetails extends Component<Props> {
         <div className="intro-contents box grow">
           <div className="description flex align-center justify-space-between">
             {selectedChart.getDescription()}
-            <Button primary label={`Install`} onClick={this.install} />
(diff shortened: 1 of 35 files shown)


jim-docker (approved):

nevalla:
> Move all strings of the form {...} to use just ""

So `{...}` does not work anymore?

> Fix RoleBindingDetails not rendering the name of the role binding

Would be make clearer if this change was in separate PR. Now it's a bit hard to spot that fix.

Nokel81:
@nevalla Using `""` is clearer of intent and is certainly easier to parse for the JS runtime that `{\`\`}`.

As for the other two, I put them in one PR because these were all left over bugs/TODOs from the removal of langci.

nevalla:
Okay...got it now. I was just a bit confused because this PR title says this fixes removal of langci so I initially thought something is broken there. But now I get it that removal of langci PR introduced that bug in RoleBinding details?

Nokel81:
Yes that is correct

jakolehm (approved):

nevalla (approved):

Issue #2207, opened 2021-02-22
Accessible namespaces remain in UI after deletion


jim-docker:
On cluster settings page add a namespace to the accessible namespaces list. Then remove it. Open the namespaces selector on a dashboard page and observe the removed namespace is there. (It isn't added to the cluster store so it is gone after a Lens restart) 

https://user-images.githubusercontent.com/40840436/108782957-8a133680-753a-11eb-982c-e627c64a0192.mov

This is with RBAC, user cannot list namespaces

Nokel81:
The source of the bug is the conflation of a theoretical `NamespaceSelectStore` and `NamespaceStore` (the later is the store for the `namespaceApi`) into just one store.

Specifically `NamespaceStore`'s (the current impl) "loads" the `allowedNamespaces` from the `context` into its `item` store which then sticks around after the `accessibleNamespace` is removed.

The solution (that I think is the most robust) is to separate these two stores and to implement an IPC channel for the `renderer` to request that `main` refresh the namespaces on a cluster "early".

Nokel81:
I have also noticed that if I add an accessible namespace that doesn't actually exist, then it doesn't show up in the dropdown. We shouldn't be doing any validation on those provided items.

PR #2217, opened 2021-02-24
Fix loading all namespaces for users with limited cluster access

nevalla:
`NamespaceStore.loadAll()` fails if user does not have access to list namespaces. This is because kubernetes api returns 403 and no kube object data are returned. Instead error is raised and, thus, no accessible namespaces are added. This PR fixes this by:

1. not to try to parse response if `KubeJsonApiData` returned from kubernetes api request is `undefined`. If not doing this,  error will be raised and `KubeObjectStore.loadAll()` fails.
2. set initial namespaces as an empty array if `NamespaceStore.loadItems()` returns `undefined`, so it can successfully add accessible namespaces to empty items.

### DIFF
--- a/src/renderer/api/kube-api.ts
+++ b/src/renderer/api/kube-api.ts
@@ -272,6 +272,7 @@ export class KubeApi<T extends KubeObject = any> {
   }
 
   protected parseResponse(data: KubeJsonApiData | KubeJsonApiData[] | KubeJsonApiDataList, namespace?: string): any {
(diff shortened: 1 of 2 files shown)


Nokel81:
I would have said that given that my PR #2193 also fixes this we don't need it. But this is a good secondary fix. However, we still should consider how to fix #2207

Nokel81 (approved):

jakolehm (approved):

PR #2221, opened 2021-02-24
Refactor the Extensions settings page

### LINKED ISSUE #2116: Installing extension twice - blue "install" status field remains flashing
**Describe the bug**
blue "install" status field gets stuck flashing

**To Reproduce**
- Open Lens
- Open extensions
- install an extension
- install the exact same extension again

**Expected behavior**
On second install, extension installs as expected and there is no stuck UI behavior.

**Screenshots**
Screen Shot 2021-02-10 at 12 52 49 PM

- Lens Version: Lens: 4.0.8
- OS: OSX 10.15.7
- Installation method (e.g. snap or AppImage in Linux): mac install

Nokel81:
- Simplify the install logic by refactoring out the multi-install logic

- Revamp the `ExtensionInstallStateStore` to more strictly track the lifetime of an install or uninstall request

- Fix the install of an already installed extension just hanging visually

- Display a spinner more often for more visual feedback

fixes #2116

### DIFF
--- a/src/extensions/extension-discovery.ts
+++ b/src/extensions/extension-discovery.ts
@@ -9,23 +9,24 @@ import { broadcastMessage, handleRequest, requestMain, subscribeToBroadcast } fr
 import { getBundledExtensions } from "../common/utils/app-version";
 import logger from "../main/logger";
 import { extensionInstaller, PackageJson } from "./extension-installer";
+import { extensionLoader } from "./extension-loader";
 import { extensionsStore } from "./extensions-store";
 import type { LensExtensionId, LensExtensionManifest } from "./lens-extension";
 
 export interface InstalledExtension {
-    id: LensExtensionId;
+  id: LensExtensionId;
 
-    readonly manifest: LensExtensionManifest;
+  readonly manifest: LensExtensionManifest;
 
-    // Absolute path to the non-symlinked source folder,
-    // e.g. "/Users/user/.k8slens/extensions/helloworld"
-    readonly absolutePath: string;
+  // Absolute path to the non-symlinked source folder,
+  // e.g. "/Users/user/.k8slens/extensions/helloworld"
+  readonly absolutePath: string;
 
-    // Absolute to the symlinked package.json file
-    readonly manifestPath: string;
-    readonly isBundled: boolean; // defined in project root's package.json
-    isEnabled: boolean;
-  }
+  // Absolute to the symlinked package.json file
+  readonly manifestPath: string;
+  readonly isBundled: boolean; // defined in project root's package.json
+  isEnabled: boolean;
+}
 
 const logModule = "[EXTENSION-DISCOVERY]";
 
@@ -236,9 +237,11 @@ export class ExtensionDiscovery {
   /**
    * Uninstalls extension.
    * The application will detect the folder unlink and remove the extension from the UI automatically.
-   * @param extension Extension to unistall.
+   * @param extensionId The ID of the extension to uninstall.
    */
-  async uninstallExtension({ absolutePath, manifest }: InstalledExtension) {
+  async uninstallExtension(extensionId: LensExtensionId) {
+    const { manifest, absolutePath } = this.extensions.get(extensionId) ?? extensionLoader.getExtension(extensionId);
+
     logger.info(`${logModule} Uninstalling ${manifest.name}`);
 
     await this.removeSymlinkByPackageName(manifest.name);
--- a/src/extensions/extension-discovery.ts
+++ b/src/extensions/extension-discovery.ts
@@ -144,9 +144,9 @@ export class ExtensionDiscovery {
       // Extension add is detected by watching "<extensionDir>/package.json" add
       .on("add", this.handleWatchFileAdd)
       // Extension remove is detected by watching "<extensionDir>" unlink
-      .on("unlinkDir", this.handleWatchUnlinkDir)
+      .on("unlinkDir", this.handleWatchUnlinkEvent)
       // Extension remove is detected by watching "<extensionSymLink>" unlink
-      .on("unlink", this.handleWatchUnlinkDir);
+      .on("unlink", this.handleWatchUnlinkEvent);
   }
 
   handleWatchFileAdd =  async (manifestPath: string) => {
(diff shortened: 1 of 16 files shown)


Nokel81:
https://user-images.githubusercontent.com/8225332/109091545-00529d00-76e3-11eb-80af-6d8ee72fa711.mov

aleksfront:
While it works okay in case of twice install, but I found a few issues:
1. It doesn't clear disabled state if wrong link provided
invalid bundle

2. When installed initial extension from a tar file, there's no loading state showed (not inside `Install` button nor with Spinner at the bottom).
no loading state

3. It doesn't clear disabled state if close button clicked inside notification. Spinner is spinning and input keeps disabled.
clicking close

4. When installing same extension twice, green notification about successful uninstall appears. Than nothing happens - no any loading state (but new extension getting installed in background).
no loading

5. Why we replaced "Install state" visualisation from button? I think we don't need to show Spinner instead when installation is in progress. At least we need to keep both.

src/renderer/components/+extensions/extension-install.store.ts
```diff
@@ -1,13 +1,187 @@
...
-@autobind()
-export class ExtensionStateStore extends Singleton {
-  extensionState = observable.map<string, ExtensionState>();
+const Prefix = "[ExtensionInstallationStore]";
+const installingExtensions = observable.set<string>();
+const uninstallingExtensions = observable.set<string>();
+const preInstallIds = observable.set<string>();
+
+export class ExtensionInstallationStateStore {
+  @action static reset() {
```
jakolehm, line 33: I see only `reset` being used .. are other methods going to be used in some other PR?

Nokel81: Do you check `src/renderer/components/+extensions/extensions.tsx`? It is by default not shown because it is considered "large"

jakolehm: Missed that, now it makes more sense 🙂

jakolehm (dismissed):

jim-docker:
> 5. Why we replaced "Install state" visualisation from button? I think we don't need to show Spinner instead when installation is in progress. At least we need to keep both.

If you already have a long list of extensions the spinner is not even visible unless you scroll to the bottom

Nokel81:
I see know, reverting to the button having the animation.

Nokel81:
@aleksfront @jakolehm @jim-docker I have fixed the issues that Aleks discovered and reverted to the older loading animation (for install). PTALA

aleksfront:
@Nokel81 There's still 2 issue available. 1 and 3 from the previous list.
1. It shows `[object Object]` error message and doesn't drop loading state.
object Object

2. It doesn't clear disabled state if close button clicked inside notification.
clicking close

Nokel81:
@aleksfront fixed

src/extensions/extension-discovery.ts
```diff
@@ -190,8 +193,9 @@ export class ExtensionDiscovery {
     // Check that the removed path is directly under this.localFolderPath
```
jim-docker, line 194: Is the comment on line 189 true anymore? (github won't let me comment on that line, this is the nearest I can get)
```
// filePath is the non-symlinked path to the extension folder
```
On line 149 `handleWatchUnlinkDir` is added as a handler for `unlink` (watching `"<extensionSymLink>"`)

src/extensions/extension-discovery.ts
```diff
@@ -272,26 +267,26 @@ export class ExtensionDiscovery {
...
-    const extensions = await this.ensureExtensions();
-
-    this.isLoaded = true;
-
-    return extensions;
+    try {
+      return await this.ensureExtensions();
+    } finally {
+      this.isLoaded = true;
+    }
```
jim-docker: Did you mean to change behaviour here? (previously `this.isLoaded` was not set to `true` if `this.ensureExtensions()` threw an error.

src/extensions/extension-loader.ts
```diff
@@ -290,28 +288,20 @@ export class ExtensionLoader {
...
-        extEntrypoint = path.resolve(path.join(path.dirname(extension.manifestPath), extension.manifest.main));
-      }
+    if (!extensionEntryPointRelativePath) {
+      return null;
+    }
 
-      if (extEntrypoint !== "") {
-        if (!fs.existsSync(extEntrypoint)) {
-          console.log(`${logModule}: entrypoint ${extEntrypoint} not found, skipping ...`);
+    const extensionEntryPointAbsolutePath = path.resolve(path.join(path.dirname(extension.manifestPath), extensionEntryPointRelativePath));
```
jim-docker: these path functions can all throw a `TypeError` so why not keep this statement in the `try` block?

src/renderer/components/+extensions/__tests__/extensions.test.tsx
```diff
@@ -46,84 +47,55 @@ jest.mock("../../../../extensions/extension-loader", () => ({
...
-  });
+  it("disables install button while installing", async () => {
+    const res = render(<Extensions />);
 
-  it("disables install button while installing", () => {
-    render(<Extensions />);
+    (fse.unlink as jest.MockedFunction<typeof fse.unlink>).mockReturnValue(Promise.resolve() as any);
 
-    fireEvent.change(screen.getByPlaceholderText("Path or URL to an extension package", {
+    fireEvent.change(res.getByPlaceholderText("Path or URL to an extension package", {
```
jim-docker, line 81: The placeholder text has changed but I guess this still works because it's not looking for an exact match?

src/renderer/components/+extensions/extension-install.store.ts
```diff
@@ -1,13 +1,218 @@
...
+import { ipcRenderer } from "electron";
 
-interface ExtensionState {
-  displayName: string;
-  // Possible states the extension can be
-  state: "installing" | "uninstalling";
+export enum ExtensionInstallationState {
+  INSTALLING = "installing",
+  UNINSTALLING = "uninstalling",
+  IDLE = "IDLE",
```
jim-docker:
```suggestion
  IDLE = "idle",
```

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -35,458 +57,457 @@ interface InstallRequestPreloaded extends InstallRequest {
...
 
-  getExtensionDestFolder(name: string) {
-    return path.join(this.extensionsPath, sanitizeExtensionName(name));
+  const manifest = await readFileFromTar<LensExtensionManifest>({
+    tarPath: filePath,
+    filePath: manifestLocation,
+    parseJson: true,
+  });
+
+  if (!manifest.lens && !manifest.renderer) {
```
jim-docker: why `manifest.lens` and not `manifest.main`?

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -35,458 +57,457 @@ interface InstallRequestPreloaded extends InstallRequest {
...
-              </div>
-            );
-          }
-        }
-      })
+    return Notifications.error(
+      <div className="flex column gaps">
+        <b>Extension Install Collision:</b>
+        <p>The <em>{name}</em> extension is currently {curState.toLowerCase()}.</p>
+        <p>Will not procede with this current install request.</p>
```
jim-docker:
```suggestion
        <p>Will not proceed with this current install request.</p>
```

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -35,458 +57,457 @@ interface InstallRequestPreloaded extends InstallRequest {
...
           </div>
-        );
+        </div>
+        <Button autoFocus label="Install" onClick={async () => {
+          removeNotification();
+
+          if (await uninstallExtension(validatedRequest.id, validatedRequest.manifest)) {
+            await unpackExtension(validatedRequest, dispose);
+          } else {
+            dispose();
```
jim-docker, line 366: Does the `validatedRequest.tempFile` need to be removed at this point?

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -35,458 +57,457 @@ interface InstallRequestPreloaded extends InstallRequest {
...
+          }
+        }} />
+      </div>,
+      {
+        onClose() {
+          dispose();
```
jim-docker: and here?

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -35,458 +57,457 @@ interface InstallRequestPreloaded extends InstallRequest {
...
 
-      if (unpackedFiles.length === 1) {
-        // check if %extension.tgz was packed with single top folder,
-        // e.g. "npm pack %ext_name" downloads file with "package" root folder within tarball
-        unpackedRootFolder = path.join(unpackingTempFolder, unpackedFiles[0]);
-      }
+    logger.info(`[EXTENSION-INSTALL]: installation has failed: ${message}`, { error, installPath });
+    Notifications.error(<p>Installation has failed: <b>{message}</b></p>);
+  } finally {
+    disposer?.();
```
jim-docker, line 423: doesn't `requestInstall()` call this in the case where `disposer` is not null?

src/extensions/extension-loader.ts
Nokel81: They can throw but only if it is not a string.

src/renderer/components/+extensions/__tests__/extensions.test.tsx
Nokel81: Probably

src/renderer/components/+extensions/extensions.tsx
Nokel81: My mistake

src/renderer/components/+extensions/extensions.tsx
Nokel81: Yes

src/renderer/components/+extensions/extensions.tsx
Nokel81: The disposer that I created can be called multiple times safely.

jim-docker: Could you have added the removal of `validatedRequest.tempFile` to the disposer?

Nokel81: I tried that and I though it was too complicated. I tried it again and it isn't.

jim-docker (approved):

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -366,15 +364,11 @@ async function attemptInstall(request: InstallRequest, d?: Disposer): Promise<vo
...
-            await fse.unlink(validatedRequest.tempFile).catch(noop);
           }
         }} />
       </div>,
       {
-        onClose() {
-          dispose();
-          fse.unlink(validatedRequest.tempFile).catch(noop);
-        }
+        onClose: dispose,
```
jim-docker, line 371: much cleaner!

jakolehm (approved):
LGTM

PR #2229, opened 2021-02-25
Flush response headers always when proxy gets a response

### LINKED ISSUE #1988: Terminal is hanging after the result of a kubectl rollout status command is returned
**Describe the bug**
When issuing a `kubectl rollout status` command the result is returned to the terminal but it doesn't complete, the terminal just hangs for approx 5mins. This means tools like Flux who rely on the result of these commands as part of their bootstrap process fail.

**To Reproduce**
1. Open lens and connect to a cluster
2. Find a deployment object
3. Open a new terminal in Lens and perform the command `kubectl rollout status -n <namespace> deployment <deployment>`
4. The result should be returned but the terminal hangs

**Expected behavior**
The terminal should return the result from the kubectl command and then return to a prompt.

**Screenshots**

- Lens Version: 4.0.6
- OS: OSX

**Logs:**
```
Your logs go here...
```

**Kubeconfig:**
```
your kubeconfig here
```

**Additional context**

jakolehm:
Headers were automatically flushed when response got data. That might not happen with watch requests and that's why we need to force flush.

Fixes #1988

### DIFF
--- a/src/main/lens-proxy.ts
+++ b/src/main/lens-proxy.ts
@@ -120,11 +120,15 @@ export class LensProxy {
   protected createProxy(): httpProxy {
     const proxy = httpProxy.createProxyServer();
 
-    proxy.on("proxyRes", (proxyRes, req) => {
+    proxy.on("proxyRes", (proxyRes, req, res) => {
       const retryCounterId = this.getRequestId(req);
 
       if (this.retryCounters.has(retryCounterId)) {
         this.retryCounters.delete(retryCounterId);
+      }
+
+      if (!res.headersSent) {
+        res.flushHeaders();
       }
     });
 
--- a/src/main/lens-proxy.ts
+++ b/src/main/lens-proxy.ts
@@ -127,8 +127,10 @@ export class LensProxy {
         this.retryCounters.delete(retryCounterId);
       }
 
-      if (!res.headersSent) {
-        res.flushHeaders();
+      if (!res.headersSent && req.url) {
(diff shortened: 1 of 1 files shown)


nevalla (dismissed):
LGTM

Nokel81 (dismissed):

jakolehm:
One integration test seems to fail (even after retry). Maybe this has side-effects 🤔

src/main/lens-proxy.ts
```diff
@@ -127,8 +127,10 @@ export class LensProxy {
         this.retryCounters.delete(retryCounterId);
       }
 
-      if (!res.headersSent) {
-        res.flushHeaders();
+      if (!res.headersSent && req.url) {
+        const url = new URL(req.url, "http://localhost");
+
+        if (url.searchParams.has("watch")) res.flushHeaders();
       }
```
Nokel81:
```suggestion
      if (!res.headersSent && (new URL(req.url, "http://localhost")).searchParams.has("watch")) {
        res.flushHeaders();
      }
```

Nokel81 (approved):
I think that this is slightly clearer since the url will convert to a valid URL (without `watch` as a `searchParam`).

nevalla (approved):

src/main/lens-proxy.ts
jakolehm: I prefer my version of this, it's longer but easier to read 🙂

PR #2233, opened 2021-02-25
Display warning notification if invalid kubeconfig detected

nevalla:
This PR is a follow-up to #1956.  This PR will display error notification if clusters with invalid kubeconfig are detected.

image

### DIFF
--- a/src/common/cluster-store.ts
+++ b/src/common/cluster-store.ts
@@ -11,7 +11,7 @@ import { appEventBus } from "./event-bus";
 import { dumpConfigYaml } from "./kube-helpers";
 import { saveToAppFiles } from "./utils/saveToAppFiles";
 import { KubeConfig } from "@kubernetes/client-node";
-import { handleRequest, requestMain, subscribeToBroadcast, unsubscribeAllFromBroadcast } from "./ipc";
+import { broadcastMessage, handleRequest, requestMain, subscribeToBroadcast, unsubscribeAllFromBroadcast, InvalidKubeconfigChannel } from "./ipc";
 import _ from "lodash";
 import move from "array-move";
 import type { WorkspaceId } from "./workspace-store";
@@ -371,4 +371,10 @@ export function getHostedClusterId() {
 
 export function getHostedCluster(): Cluster {
   return clusterStore.getById(getHostedClusterId());
+}
+
+export function reportDeadClusters() {
+  clusterStore.clustersList.filter(cluster => cluster.isDead).forEach(cluster => {
+    broadcastMessage(InvalidKubeconfigChannel, cluster.id);
+  });
 }
--- a/src/common/cluster-store.ts
+++ b/src/common/cluster-store.ts
@@ -11,7 +11,7 @@ import { appEventBus } from "./event-bus";
 import { dumpConfigYaml } from "./kube-helpers";
 import { saveToAppFiles } from "./utils/saveToAppFiles";
 import { KubeConfig } from "@kubernetes/client-node";
-import { broadcastMessage, handleRequest, requestMain, subscribeToBroadcast, unsubscribeAllFromBroadcast, InvalidKubeconfigChannel } from "./ipc";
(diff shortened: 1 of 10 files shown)


src/renderer/ipc/invalid-kubeconfig-handler.tsx
```diff
@@ -0,0 +1,44 @@
...
+        <b>Cluster with invalid Kubeconfig Detected!</b>
+        <p>Cluster <b>{cluster.name}</b> has invalid kubeconfig and cannot be displayed.
+        Please fix the <a href="#" onClick={(e) => { e.preventDefault(); shell.showItemInFolder(cluster.kubeConfigPath); }}>kubeconfig</a> manually or remove the cluster. </p>
+        <p>Do you want to remove the cluster now?</p>
+        <div className="flex gaps row align-left box grow">
+          <Button active outlined label="Remove" onClick={()=> {
+            clusterStore.removeById(clusterId);
+            notificationsStore.remove(notificationId);
+          }} />
+          <Button active outlined label="Cancel" onClick={() => notificationsStore.remove(notificationId)} />
```
Nokel81, line 35: Shouldn't the notification be removed here too?

nevalla: What do you mean? To remove the notification when clicking `Cancel` button? It should already do that.

Nokel81: Looking at the code, there doesn't seem to be any code inspecting a notification's `message` field for a button labelled with `"Cancel"`. There also doesn't seem to be any other instances of this pattern too (which is why I didn't know about it if it is the case).

Nokel81: The only place where a button has a label of `"Cancel"` is in `InfoPanel` but that is for the hover text

Nokel81: Oh sorry, I completely blanked while reading the code. Sorry.

nevalla: `notificationsStore.remove(notificationId)` should remove the notification or at least it disappears when I clicked the `Cancel` button.

Nokel81: Yes I don't know why I thought otherwise. Sorry.

src/common/cluster-store.ts
```diff
@@ -372,3 +372,9 @@ export function getHostedClusterId() {
 export function getHostedCluster(): Cluster {
   return clusterStore.getById(getHostedClusterId());
 }
+
+export function reportDeadClusters() {
+  clusterStore.clustersList.filter(cluster => cluster.isDead).forEach(cluster => {
```
jakolehm: More about an issue in the target branch/PR but could we actually store "dead state" in `ClusterManager` and not in cluster object itself? That way we would not need to add yet-another-attribute to a (exported) class that's also visible through the extension API.

jakolehm (approved):

src/common/cluster-store.ts
nevalla: Played around this a bit and it's not that trivial. If we raise error from `Cluster` contructor and catch it on `ClusterStore`, cluster object itself will be `undefined` and cannot put in to cluster store items -> it will get lost when persisting cluster store next time.

Nokel81: Exactly why I think this should be stored in `ClusterStore`. A `Cluster` represents a valid kubeconfig.

nevalla: I have one idea that I will try, but I agree it's more about the issue in the target PR, so I propose let's continue discussion there.

nevalla:
I've now updated the notification message, to display context if it's different than cluster name. Also added mention to restart Lens after the kubeconfig is edited manually. @jakolehm @aleksfront PTAL.

jakolehm (approved):

PR #2241, opened 2021-02-26
Apply custom filters after selecting source

### LINKED ISSUE #2228: Group filtering does not work in Custom Resources
**Describe the bug**
Under the Custom Resources / Definitions panel the group filtering has no effect.

**To Reproduce**
1. Go to Custom Resources -> Definitions
2. Click on filter (top right) and select any group name
4. Custom Resources list does not change 

**Expected behavior**
List to show just resources for the group selected.

- Lens Version: 4.1.1
- OS: Ubuntu 20.04.2 

**Logs:**
```
Ran from the command line but no logs generated.
```

pashevskii:
Fixes [this ticket](https://github.com/lensapp/lens/issues/2228)

The problem was in applying custom filters. It is going just within `get allItems()` getter, but this will be not used if this.props.items is not empty. 
Solution: to apply custom filters after selecting data source.

### DIFF
--- a/src/renderer/components/item-object-list/item-list-layout.tsx
+++ b/src/renderer/components/item-object-list/item-list-layout.tsx
@@ -197,16 +197,11 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
     return filters.reduce((items, filter) => filter(items), items);
   }
 
-  @computed get allItems() {
-    const { filterItems, store } = this.props;
-
-    return this.applyFilters(filterItems, store.items);
-  }
-
   @computed get items() {
-    const { allItems, filters, filterCallbacks } = this;
-    const filterItems: ItemsFilter[] = [];
+    const {filters, filterCallbacks } = this;
     const filterGroups = groupBy<Filter>(filters, ({ type }) => type);
+
+    const filterItems: ItemsFilter[] = [];
 
     Object.entries(filterGroups).forEach(([type, filtersGroup]) => {
       const filterCallback = filterCallbacks[type];
@@ -216,9 +211,9 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
       }
     });
 
-    const items = this.props.items ?? allItems;
+    const items = this.props.items ?? this.props.store.items;
 
-    return this.applyFilters(filterItems, items);
+    return this.applyFilters(filterItems.concat(this.props.filterItems), items);
   }
 
   @autobind()


jim-docker:
Hmm, https://github.com/lensapp/lens/pull/2114 fixed this too, but was closed because an alternative PR was applied. Looks like it should've been applied. Maybe https://github.com/lensapp/lens/pull/2114 is better?

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -215,8 +209,7 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
         filterItems.push(filterCallback);
       }
     });
-
-    const items = this.props.items ?? allItems;
+    const items = this.props.items ?? this.props.store.items;
 
     return this.applyFilters(filterItems, items);
   }
```
Nokel81, line 217: We shouldn't be pushing to a prop. Plus I think that `filterItems` will soon going away. However, here is version that doesn't

```typescript
  @computed get items() {
    const { filters, filterCallbacks } = this;
    const filterGroups = groupBy<Filter>(filters, ({ type }) => type);

    const filterItems = Object.keys(filterGroups)
      .map(type => [filterCallbacks[type], filterGroups[type]] as const)
      .filter(([callback, groups]) => callback && groups.length)
      .map(([callback]) => callback)
      .concat(this.props.filterItems);

    const items = this.props.items ?? this.props.store.items;

    return this.applyFilters(filterItems, items);
  }
```

jim-docker:
Also, be aware that `filterItems` is set to be deprecated: https://github.com/lensapp/lens/pull/2123

Nokel81 (dismissed):

src/renderer/components/item-object-list/item-list-layout.tsx
pashevskii: Oh, forgot this. Fixed

pashevskii:
> Also, be aware that `filterItems` is set to be deprecated: #2123

Hmmm, it makes sense, thanks

pashevskii:
So, [#2123](https://github.com/lensapp/lens/pull/2123) fixes the issue too :)  
I'll close my PR

Nokel81:
I don't think that issue will be merged any time soon given that it currently has a breaking change in it. This PR is nice and quick for the meanwhile.

jim-docker (dismissed):

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -197,26 +197,17 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
-
-      if (filterCallback && filtersGroup.length > 0) {
-        filterItems.push(filterCallback);
-      }
-    });
+    const filterItems = Object.keys(filterGroups)
+      .map(type => [filterCallbacks[type], filterGroups[type]] as const)
+      .filter(([callback, groups]) => callback && groups.length)
+      .map(([callback]) => callback)
+      .concat(this.props.filterItems);
```
aleksfront: This refactoring with multiple iterations doesn't give code cleanliness, it's harder to understand what is intent here. Also, `as const` part looks like a forced measure, not the planned thing.

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -197,17 +197,12 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
     return filters.reduce((items, filter) => filter(items), items);
   }
 
-  @computed get allItems() {
```
jim-docker, line 200: Isn't removing this technically a breaking change?

pashevskii: Why? 
allItems was used in one place before the change and contained just one function which, in fact, was moved to after the  data source selecting

jakolehm: @pashevskii this component is exported via extension API. Which means we don't know who uses this.

aleksfront: Probably this isn't a breaking change since `allItems` method is used internally inside component. You can't use it outside, you have to only pass a props listed.

jim-docker: Oh right. I thought it was basically the same issue as with `filterItems` (https://github.com/lensapp/lens/pull/2123#discussion_r574485966)
but that was leaking through `ItemListLayoutProps`. `allItems` is not there

jim-docker (approved):

Nokel81 (approved):

aleksfront (approved):

aleksfront:
Fixes #2291

PR #2245, opened 2021-02-26
make Pages and PageMenus observable and reactive

### LINKED ISSUE #1432: Observable fields in extension types aren't reacted to
**Describe the bug**
Several fields within extension related types (e.g. `LensExtension` or `LensMainExtension`) are marked as `@observable` (currently shallowly). However, if they are changed during execution of the extension (meaning the time between `onActivate` is called and `onDeactivate` is called) it does not propagate. 

For instance, if an extension sets the `appMenus` field to an array and then later modifies it (say by calling `.pop()` on the field), there is no change to the corresponding app menu.

**To Reproduce**
The following code does not work as expected:
```typescript
class MyExtension extends LensMainExtension {
    clusterPageMenus = [
        {
          target: { pageId: "example", params: {} },
          title: "Example Extension",
          components: {
            Icon: ExampleIcon,
          }
        }
      ]

    async onActivate() {
        console.log("hello world")
        setTimeout(() => {
          this.clusterPageMenus.pop()
        }, 5000)
      }
}
```

**Expected behavior**
The "Example Extension" menu should be visible for the first 5000ms of the extensions' lifetime and then disappear

Nokel81:

This is my MVP POC to fix #1432.

The main idea of this solution is to use `computed` values behind symbol private `get`-ers so that extension developers are not able to use them and then build up helper functions that rely on the `ExtensionLoader` to be `reactive` to changes to the values.

This also simplifies the code in so far as we now only have a "push" model (the extension instances push to the consumers) instead of a "pull/push" model (where the stores are pushed the values and then the consumers pull from the stores). Furthermore, it removes the need to have callbacks on initialization because the mere act of adding a `LensExtension` instance to the `ExtensionLoader`'s observable map causes all the computed values to be recalculated.

---

To try it out (and see that it does indeed work) add the following snippet to the `helloworld-sample` in `lens-extension-samples`.

```ts
setTimeout(() => {
  console.log("adding new cluster page menu");

  this.clusterPages.push({
    id: "hello-2", // hello-world:foo
    components: {
      Page: () => <ExamplePage extension={this} />,
    }
  });
  this.clusterPageMenus.push({
    target: { pageId: "hello-2" },
    title: "Hello World 2",
    components: {
      Icon: ExampleIcon,
    }
  });
}, 10000);
```

### DIFF
--- a/src/extensions/core-api/index.ts
+++ b/src/extensions/core-api/index.ts
@@ -1,6 +1,6 @@
 // Lens-extensions api developer's kit
-export * from "../lens-main-extension";
-export * from "../lens-renderer-extension";
+export { LensMainExtension } from "../lens-main-extension";
+export { LensRendererExtension } from "../lens-renderer-extension";
 
 // APIs
 import * as App from "./app";
--- a/src/extensions/extension-loader.ts
+++ b/src/extensions/extension-loader.ts
@@ -14,7 +14,6 @@ import type { LensRendererExtension } from "./lens-renderer-extension";
 import * as registries from "./registries";
 import fs from "fs";
 
-
 export function extensionPackagesRoot() {
   return path.join((app || remote.app).getPath("userData"));
 }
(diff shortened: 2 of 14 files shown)


Nokel81:
Closing as this will be part of DI

Issue #2195, opened 2021-02-19
Lens shows as open and as a process but with no window appearing on screen 


Ryan-Sharpe1:
**Describe the bug**
No window showing when opening the Lens application. I have tried rebooting my mac, deleting the k8s config folder 
 and reinstalling lens but still no window opening when the process is started. It is showing in the toolbar that Lens is 'open' but with no options on many of the menus as shown in the screenshots attached. My colleagues are not having the same issue and seems to only be on my machine, are there any known fixes? 

**To Reproduce**
1. Unsure we can reproduce this error.
2. Download Lens through github
3. Install the application
4. Open the application
5. No window showing but toolbar shows the programme is 'running' 

**Expected behavior**
For the UI to open. 

**Screenshots**
![Screenshot 2021-02-19 at 12 26 34](https://user-images.githubusercontent.com/54810225/108505510-6be5d600-72af-11eb-9073-0472617a4a74.jpg)
![Screenshot 2021-02-19 at 12 26 55](https://user-images.githubusercontent.com/54810225/108505734-c3844180-72af-11eb-92c8-9a6b613de2bf.jpg)
![Screenshot 2021-02-19 at 12 26 56](https://user-images.githubusercontent.com/54810225/108505773-d26af400-72af-11eb-8bdf-1f72c770ba93.jpg)
![Screenshot 2021-02-19 at 12 26 58](https://user-images.githubusercontent.com/54810225/108505797-deef4c80-72af-11eb-9d8a-46c00668109b.jpg)
![Screenshot 2021-02-19 at 12 26 59](https://user-images.githubusercontent.com/54810225/108505823-e6aef100-72af-11eb-958d-b30280335fb4.jpg)
![Screenshot 2021-02-19 at 12 27 01](https://user-images.githubusercontent.com/54810225/108505854-f0385900-72af-11eb-9452-4ffd02ced60f.jpg)

**Environment (please complete the following information):**
- Lens Version: 4.1.2
- OS: MacOS Catalina 10.15.7 

- Installation method (e.g. snap or AppImage in Linux):
- Lens v4.1.2 - MacOS from github

**Logs:**
When you run the application executable from command line you will see some logging output. Please paste them here:
```
Your logs go here...
```

**Kubeconfig:**
Quite often the problems are caused by malformed kubeconfig which the application tries to load. Please share your kubeconfig, remember to remove any secret and sensitive information.
```
your kubeconfig here
```

**Additional context**

arogic:
[...]

PR #2246, opened 2021-02-27
attempt to recover the ui after failing to load cluster store

jim-docker:
Trying to improve the situation when the app starts without the main UI appearing (menu is present) due to an unknown problem with loading the cluster store after an update. This PR recovers the UI (with an empty cluster store) but any clusters that are added can't connect (spins forever)  
```
info: [APP]: Init dashboard, clusterId=1fdef426-95cd-443b-ade0-8d5a4374075c, frameId=5
...
Uncaught (in promise) TypeError: Cannot read property 'whenReady' of undefined
```

addresses #2195

### DIFF
--- a/src/common/cluster-store.ts
+++ b/src/common/cluster-store.ts
@@ -124,36 +124,43 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
   }
 
   async load() {
-    await super.load();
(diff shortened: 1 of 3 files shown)


Nokel81:
Closing due to referring item being closed.

PR #2252, opened 2021-03-01
Add image sha256 in pod inspector

### LINKED ISSUE #600: Add image sha256 in pod inspector
I would like having **sha256** of images inside pod inspector.

**Why is this needed**:
There are situations, like I having right now, when the image's tag not usually modified. My company workflow's fault, though.
Currently, I'm using `imagePullPolicy: Always`, so that new image with the same tag are pulled.
As careless as usual, I need to check the image sha256 so that my deployment success as expected.
I recently moved to Lens, it's a big help when I'm not wasting my time typing any single trivial commands anymore.
But Lens currently's not having sha256 alongside with image name and tag.

I think this are not appropriate to most of users, but that's not bad either to having it.

- Kubernetes distribution: OVH
- Desktop OS: MacOS Catalina

vshakirova:
Fixes: https://github.com/lensapp/lens/issues/600

Screen Shot 2021-03-02 at 12 07 56 PM

### DIFF
--- a/src/renderer/components/+workloads-pods/pod-details-container.tsx
+++ b/src/renderer/components/+workloads-pods/pod-details-container.tsx
@@ -53,6 +53,7 @@ export class PodDetailsContainer extends React.Component<Props> {
     const state = status ? Object.keys(status.state)[0] : "";
     const lastState = status ? Object.keys(status.lastState)[0] : "";
     const ready = status ? status.ready : "";
+    const imageId = status? status.imageID : "";
     const liveness = pod.getLivenessProbe(container);
     const readiness = pod.getReadinessProbe(container);
     const startup = pod.getStartupProbe(container);
@@ -84,7 +85,7 @@ export class PodDetailsContainer extends React.Component<Props> {
         </DrawerItem>
         }
         <DrawerItem name="Image">
-          {image}
+          <Badge label={image} tooltip={imageId}/>
         </DrawerItem>
         {imagePullPolicy && imagePullPolicy !== "IfNotPresent" &&
         <DrawerItem name="ImagePullPolicy">


src/renderer/components/+workloads-pods/pod-details-container.tsx
```diff
@@ -86,6 +87,9 @@ export class PodDetailsContainer extends React.Component<Props> {
         <DrawerItem name="Image">
           {image}
         </DrawerItem>
+        <DrawerItem name="ImageId">
+          {imageId}
+        </DrawerItem>
```
Nokel81: Given that not everyone will need this maybe only have it as a tooltip, wdyt @aleksfront ?

```suggestion
        <DrawerItem name="Image" tooltip={imageId}>
          {image}
        </DrawerItem>
```

BTW, for the above suggestion to work you need to add `, TooltipDecoratorProps` to the `interface DrawerItemProps` and an `@withTooltip` line above the `class DrawerItem` statement.

aleksfront: The question here would users need to copy `ImageId`? If we'll go with tooltip, we can do this by wrapping image value with `<Badge/>`:
```
<DrawerItem name="Image">
  <Badge label={image} tooltip={imageId}/>
</DrawerItem>
```

Nokel81: You can copy from a badge so yes I think that is a better idea.

Nokel81 (dismissed):

src/renderer/components/drawer/drawer-item.tsx
```diff
@@ -11,6 +12,7 @@ export interface DrawerItemProps extends React.HTMLAttributes<any> {
   renderBoolean?: boolean; // show "true" or "false" for all of the children elements are "typeof boolean"
 }
 
+@withTooltip
```
aleksfront: Since we're using `<Badge/>` to show a tooltip, we don't need any changes in current file (`drawer-item`). Badge component itself already has all needed decorators.

aleksfront (changes requested):

src/renderer/components/drawer/drawer-item.tsx
vshakirova: Thank you, I'll remove the decorator and `TooltipDecoratorProps`

Nokel81 (approved):

aleksfront (approved):

PR #2253, opened 2021-03-01
Fix: preventing <App/> to render on cluster refresh

### LINKED ISSUE #2208: Drawer or Modal popup gets recreated when it is connected to Mobx
**Description**
Drawer or Modal popup gets recreated when it is connected to Mobx or redux and this behavious is seen greater version of 4.0.8.

**To Reproduce**

Call a AddDialog box from addorremove property using KubeObjectList.

Mount the AddDialog inside the same file. Trigger the open from KubeListTableControls i.e add icon with mobx. It opens and stays for sometime and gets closed after few seconds. This is happening after >4.0.8 versions.

The behaviour is same Drawer component as well

**Expected behavior**
The drawer which is opened or modal popup should not be closed and the values inside should be retained.

aleksfront:
Removing all references to observable cluster object from `<App/>` render method. This prevents app to rerender (and loose state in extension pages) on cluster refresh. Changed route files because of same references inside `isAllowedResource` method.

Also, adding `observer` wrapper to `<MainLayoutHeader/>` to properly catch cluster renaming from Settings.

Fixes #2208

### DIFF
--- a/src/renderer/components/+config/config.route.ts
+++ b/src/renderer/components/+config/config.route.ts
@@ -1,12 +1,21 @@
 import { RouteProps } from "react-router";
-import { Config } from "./config";
 import { IURLParams } from "../../../common/utils/buildUrl";
-import { configMapsURL } from "../+config-maps/config-maps.route";
+import { configMapsRoute, configMapsURL } from "../+config-maps/config-maps.route";
+import { hpaRoute } from "../+config-autoscalers";
+import { limitRangesRoute } from "../+config-limit-ranges";
+import { pdbRoute } from "../+config-pod-disruption-budgets";
+import { resourceQuotaRoute } from "../+config-resource-quotas";
+import { secretsRoute } from "../+config-secrets";
 
 export const configRoute: RouteProps = {
-  get path() {
-    return Config.tabRoutes.map(({ routePath }) => routePath).flat();
-  }
+  path: [
+    configMapsRoute,
+    secretsRoute,
+    resourceQuotaRoute,
+    limitRangesRoute,
+    hpaRoute,
+    pdbRoute
+  ].map(route => route.path.toString())
 };
 
 export const configURL = (params?: IURLParams) => configMapsURL(params);
--- a/src/renderer/components/+workloads/workloads.route.ts
+++ b/src/renderer/components/+workloads/workloads.route.ts
@@ -1,13 +1,6 @@
 import type { RouteProps } from "react-router";
 import { buildURL, IURLParams } from "../../../common/utils/buildUrl";
 import { KubeResource } from "../../../common/rbac";
-import { Workloads } from "./workloads";
-
-export const workloadsRoute: RouteProps = {
-  get path() {
-    return Workloads.tabRoutes.map(({ routePath }) => routePath).flat();
-  }
-};
 
 // Routes
 export const overviewRoute: RouteProps = {
@@ -33,6 +26,19 @@ export const jobsRoute: RouteProps = {
 };
 export const cronJobsRoute: RouteProps = {
   path: "/cronjobs"
+};
+
+export const workloadsRoute: RouteProps = {
+  path: [
+    overviewRoute,
+    podsRoute,
+    deploymentsRoute,
+    daemonSetsRoute,
+    statefulSetsRoute,
+    replicaSetsRoute,
+    jobsRoute,
+    cronJobsRoute
+  ].map(route => route.path.toString())
 };
 
 // Route params
--- a/src/renderer/components/app.tsx
+++ b/src/renderer/components/app.tsx
@@ -51,6 +51,7 @@ import { CommandContainer } from "./command-palette/command-container";
 import { KubeObjectStore } from "../kube-object.store";
 import { clusterContext } from "./context";
 
+@observer
 export class App extends React.Component {
   static async init() {
     const frameId = webFrame.routingId;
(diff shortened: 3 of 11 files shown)


Nokel81 (dismissed):
Except for the lint catch. Looks good. Nice and simple

Nokel81 (dismissed):

jakolehm (commented):
How can extension route changes work without an observer?

hansiemithun:
Thanks for considering @aleksfront, really appreciate it. Hoping to verify this soon. When we can expect this to be released?

hansiemithun:
@aleksfront  - small lint issue exist for this to close 🙂

aleksfront:
> @aleksfront - small lint issue exist for this to close 🙂

Current fix doesn't solve all of the issues. But the good news is update is on the way. We expecting that it will be fixed soon.

hansiemithun:
> > @aleksfront - small lint issue exist for this to close 🙂
> 
> Current fix doesn't solve all of the issues. But the good news is update is on the way. We expecting that it will be fixed soon.

@aleksfront  - So, can we at-least believe this fixes drawer recreation?

src/renderer/components/app.tsx
```diff
@@ -152,38 +148,6 @@ export class App extends React.Component {
     });
   }
 
-  @observable extensionRoutes: Map<ClusterPageMenuRegistration, React.ReactNode> = new Map();
-
-  generateExtensionTabLayoutRoutes(rootItems: ClusterPageMenuRegistration[]) {
```
Nokel81, line 157: What was this used for previously? It isn't being used anywhere.

Nokel81 (dismissed):
From a code review pov this looks good.

src/renderer/components/command-palette/command-container.tsx
```diff
@@ -29,8 +28,9 @@ export class CommandOverlay {
 }
 
 @observer
-export class CommandContainer extends React.Component<{cluster?: Cluster}> {
+export class CommandContainer extends React.Component {
   @observable.ref commandComponent: React.ReactElement;
+  @observable cluster = getHostedCluster();
```
jakolehm: Why `getHostedCluster` vs passing it as a prop? This is an anti-pattern.

src/renderer/components/+config/config.route.ts
```diff
@@ -1,12 +1,21 @@
...
-    return Config.tabRoutes.map(({ routePath }) => routePath).flat();
-  }
+  path: [
+    configMapsRoute,
+    secretsRoute,
+    resourceQuotaRoute,
+    limitRangesRoute,
+    hpaRoute,
+    pdbRoute
+  ].map(route => route.path.toString())
```
ixrock, line 18:
```suggestion
  path: [
    configMapsRoute.path,
    secretsRoute.path,
    resourceQuotaRoute.path,
    limitRangesRoute.path,
    hpaRoute.path,
    pdbRoute.path,
  ].map(String)
```

src/renderer/components/app.tsx
jakolehm: Wasn't it called on line 95?

src/renderer/components/app.tsx
```diff
@@ -92,24 +92,20 @@ export class App extends React.Component {
         broadcastMessage(`cluster-warning-event-count:${getHostedCluster().id}`, count);
       }),
 
-      reaction(() => clusterPageMenuRegistry.getRootItems(), (rootItems) => {
-        this.generateExtensionTabLayoutRoutes(rootItems);
-      }, {
-        fireImmediately: true
+      reaction(getHostedCluster, () => {
```
ixrock: This looks weird. What it the intention here? 
I suppose we can setup `startUrl` without any reactions since it should be available when app is renders first time.

src/renderer/components/+workloads/workloads.route.ts
```diff
@@ -35,6 +28,19 @@ export const cronJobsRoute: RouteProps = {
   path: "/cronjobs"
 };
 
+export const workloadsRoute: RouteProps = {
```
jakolehm, line 31: Are these refactorings necessary for the fix?

src/renderer/components/app.tsx
Nokel81: It was called yes but the field `extensionRoutes` was never read from outside of that function

src/renderer/components/+workloads/workloads.route.ts
aleksfront: Yes, they're tightly coupled. Previous `workloadsRoute` (and similar) used following sentence:
```
return Workloads.tabRoutes.map(({ routePath }) => routePath).flat();
```
`Workloads.tabRoutes` checks if some resource is allowed then pushed it to `routes` array
```
if (isAllowedResource("pods")) {
      routes.push({
        title: "Pods",
        component: Pods,
        url: podsURL({ query }),
        routePath: podsRoute.path.toString()
      });
    }
```
The key here that `isAllowedResouse` is using observable `getHostedCluster` inside. This causes an `<App/>` to rerender. So, by removing references to observable field we eliminate excessive rerenderings.

src/renderer/components/app.tsx
aleksfront: True. `extensionRoutes` was used for some time, but after [this changes](https://github.com/lensapp/lens/pull/1990/files#diff-bf0690965bbc846224b78fb77bb20125b2c02c3e48553d6c35f825d7b20bd01e) no longer in use

src/renderer/components/+config/config.route.ts
aleksfront: Not sure what option is better. Are we okay with lots of `.path` repeats?

src/renderer/components/app.tsx
aleksfront: Fixed

src/renderer/components/command-palette/command-container.tsx
aleksfront: Fixed, passing cluster id as a prop.

jakolehm (approved):

nevalla (approved):
LGTM

PR #2259, opened 2021-03-01
Fix integration test lifecycle fuctions not failing tests

Nokel81:
- Fail test if an async lifecycle function fails (though a bug in Jest means that the test body is still run)

part of splitting out #2003

### DIFF
--- a/integration/__tests__/app.tests.ts
+++ b/integration/__tests__/app.tests.ts
@@ -3,7 +3,6 @@ import * as utils from "../helpers/utils";
 import { listHelmRepositories } from "../helpers/utils";
 import { fail } from "assert";
 
-
 jest.setTimeout(60000);
 
 // FIXME (!): improve / simplify all css-selectors + use [data-test-id="some-id"] (already used in some tests below)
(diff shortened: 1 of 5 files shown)


integration/__tests__/app.tests.ts
```diff
@@ -3,21 +3,22 @@ import * as utils from "../helpers/utils";
...
-
 jest.setTimeout(60000);
 
 // FIXME (!): improve / simplify all css-selectors + use [data-test-id="some-id"] (already used in some tests below)
 describe("Lens integration tests", () => {
   let app: Application;
 
   describe("app start", () => {
-    beforeAll(async () => app = await utils.appStart(), 20000);
+    beforeAll(utils.wrapJestLifecycle(async () => {
```
jim-docker: You could reduce duplication with `beforeAllWrapped(fn : () => Promise<void>, ...)`
and `afterAllTearDown()`, `afterEachTearDown()`

Nokel81: Good idea

jim-docker (approved):

aleksfront (approved):
@Nokel81 Is this bug you mentioned in description listed somewhere (my quick search not helped). I'm just thinking is it still an issue in latest jest version?

Nokel81:
@aleksfront only the issue that I created: https://github.com/facebook/jest/issues/11060

PR #2265, opened 2021-03-01
Refactor helm-chart.api and improve kube validation and error handling

### LINKED ISSUE #1961: Lens 4* versions + RBAC + pods don't appear in the dashboard

### LINKED ISSUE #2227: "Cannot read property 'getId' of undefined" when clicking on a pod

Nokel81:
- Move to async free functions
- Use AbortController instead of CancaliblePromise
- Significatly improve the runtime checking of values returned by the Kube and KubeJson APIs
- Add tests for KubeObject
- Add error handling for unexpected values from the API

fixed #1961 , fixes #2227


jim-docker:
The title should be more like "Improve kube object validation/kube api error handling plus helm chart refactor" (the helm chart refactor couldn't have been broken out into a separate PR? Guess it would've been dependent on the rest of this PR?

Nokel81:
I initially tried having it as separate PRs but as you have seen they both would have been blocked on the other getting merged. That is a better name though, thanks.

src/common/utils/type-narrowing.ts
```diff
@@ -1,13 +1,89 @@
 /**
  * Narrows `val` to include the property `key` (if true is returned)
  * @param val The object to be tested
- * @param key The key to test if it is present on the object
+ * @param key The key to test if it is present on the object (must be a literal for tsc to do any type meaningful)
```
jim-docker: feels like there's a typo here

jim-docker (approved):

src/renderer/api/json-api.ts
```diff
@@ -119,48 +117,53 @@ export class JsonApi<D = JsonApiData, P extends JsonApiParams = JsonApiParams> {
...
+    if (status >= 200 && status < 300) {
+      this.onData.emit(data, res);
+      this.writeLog({ ...log, data });
+
+      return data;
+    }
+
+    if (log.method === "GET" && res.status === 403) {
+      this.writeLog({ ...log, error: data });
+      throw data;
```
jakolehm, line 146: Why did we change the logic to throw?

src/renderer/api/kube-api.ts
```diff
@@ -271,20 +277,12 @@ export class KubeApi<T extends KubeObject = any> {
...
-
-      ensureObjectSelfLink(this, object);
-
-      return object;
-    }
-
-    // process items list response
-    if (KubeObject.isJsonApiDataList(data)) {
+    // process items list response, check before single item since there is overlap
+    if (KubeObject.isJsonApiDataList(data, KubeObject.isPartialJsonApiData)) {
```
jakolehm, line 285: Why did we swap order of checking list vs data? Can it cause unintended side-effects?

src/renderer/api/kube-object.ts
```diff
@@ -78,15 +79,59 @@ export class KubeObject implements ItemObject {
...
-
-  static isJsonApiDataList(object: any): object is KubeJsonApiDataList {
-    return object.items && object.metadata;
-  }
-
-  static stringifyLabels(labels: { [name: string]: string }): string[] {
+  static isJsonApiData(object: unknown): object is KubeJsonApiData {
+    return (
+      isObject(object)
+      && hasTypedProperty(object, "kind", isString)
```
jakolehm, line 85: Why do we need to test type of these properties? Don't we trust that kubernetes will respond with correct types? 🤔

src/renderer/api/json-api.ts
Nokel81: I changed to throw because that matches how the rest of the 4XX errors are handled and it is what the renderer is expecting. Without this the renderer thinks this call succeeds.

src/renderer/api/kube-api.ts
Nokel81: As the comment tried to convey. I found that if we kept the old order then all the data list requests would be typed as individual item requests.

src/renderer/api/kube-object.ts
Nokel81: I would consider this a form of defence in depth. With this here we are making sure that we are not messing with the data too much with any of the other code. 

Yes I think that kubernetes will probably respond with the correct data fields.

src/renderer/api/json-api.ts
jakolehm: But this change affects _all_ api calls. These messages were ignored on purpose -> if we throw these then user will see a notification every time api call is made where user does not have access to -> lot's of spam.

Nokel81: I didn't find that was the case. Let me check again.

src/renderer/api/kube-object.ts
jakolehm: To me it feels like a bit too much defence... it's like validating data from a database every time it's read. Not sure what others think? /cc @nevalla @jim-docker @ixrock  

If we really want to validate data we probably should use something like https://joi.dev/ .

Nokel81: Why would you not want to validate data from the db? In most languages (except dynamic typed ones) that is basically a necessity. I think that validating data as soon as possible is a great way to prevent bad data from getting through.

jakolehm: Starts to be a bit off-topic but if a database has a column with a type it cannot really be anything else.. so why would you validate the type on read?

Nokel81: To make sure the program requesting the data has the same model as the database?

Plus that only works for strict typed databases. Which an HTTP request is not that sort imo.

ixrock: >  In most languages (except dynamic typed ones) that is basically a necessity.

Maybe then we should re-write Lens on c++? :D

I agree with @jakolehm that it's a bit too much defence..

jim-docker: Is it going to make life more difficult for programmers/users? Is it easy to draw the line at less defence? If so then we should back it off. Otherwise it's pretty thorough but I didn't think it was too much...

Nokel81: I would also like to make the counter point that just because we can do less validation doesn't mean that we should. For instance if something like `selfLink` becoming optional and deprecated happens again then users (or CI tests if we get them that good) will start seeing errors a lot sooner.

Nokel81: @jakolehm Would you prefer if I reworked this PR in terms of `joi` then?

jakolehm: I'm not completely happy that we are checking data types here and doing it with our custom implementation but the implementation is hidden behind functions so we can improve/cleanup later.

src/renderer/api/json-api.ts
jakolehm: Did we ever check this?

jakolehm (approved):

src/renderer/api/json-api.ts
Nokel81: I have, the only change is that the logs to the dev-tools console are now red instead of green. That is all. The reason for this is `fetch` doesn't throw on HTTP errors, only on things like passing it not a string.

PR #2268, opened 2021-03-01
Lens should point to the release docs

### LINKED ISSUE #2254: Lens always links to latest documentation
**Describe the bug**
When selecting "Help" -> "Documentation" Lens navigates to `https://docs.k8slens.dev/` but should navigate to the documentation corresponding to the running version (e.g. `https://docs.k8slens.dev/v4.1.0/`).

Nokel81:

- On production builds Lens should point to its version of the docs.
- On dev builds it should point to latest (as the next version hasn't been released yet)

### DIFF
--- a/src/common/vars.ts
+++ b/src/common/vars.ts
@@ -42,5 +42,8 @@ export const apiKubePrefix = "/api-kube"; // k8s cluster apis
 // Links
 export const issuesTrackerUrl = "https://github.com/lensapp/lens/issues";
 export const slackUrl = "https://join.slack.com/t/k8slens/shared_invite/enQtOTc5NjAyNjYyOTk4LWU1NDQ0ZGFkOWJkNTRhYTc2YjVmZDdkM2FkNGM5MjhiYTRhMDU2NDQ1MzIyMDA4ZGZlNmExOTc0N2JmY2M3ZGI";
-export const docsUrl = "https://docs.k8slens.dev/";
 export const supportUrl = "https://docs.k8slens.dev/latest/support/";
+
+const docsVersion = isProduction ? `v${packageInfo.version}` : "latest";
+
+export const docsUrl = `https://docs.k8slens.dev/${docsVersion}`;
--- a/src/common/vars.ts
+++ b/src/common/vars.ts
@@ -1,5 +1,6 @@
 // App's common configuration for any process (main, renderer, build pipeline, etc.)
 import path from "path";
+import { SemVer } from "semver";
 import packageInfo from "../../package.json";
 import { defineGlobal } from "./utils/defineGlobal";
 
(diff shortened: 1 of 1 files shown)


src/common/vars.ts
```diff
@@ -42,5 +42,8 @@ export const apiKubePrefix = "/api-kube"; // k8s cluster apis
 // Links
 export const issuesTrackerUrl = "https://github.com/lensapp/lens/issues";
 export const slackUrl = "https://join.slack.com/t/k8slens/shared_invite/enQtOTc5NjAyNjYyOTk4LWU1NDQ0ZGFkOWJkNTRhYTc2YjVmZDdkM2FkNGM5MjhiYTRhMDU2NDQ1MzIyMDA4ZGZlNmExOTc0N2JmY2M3ZGI";
-export const docsUrl = "https://docs.k8slens.dev/";
 export const supportUrl = "https://docs.k8slens.dev/latest/support/";
+
+const docsVersion = isProduction ? `v${packageInfo.version}` : "latest";
```
jim-docker: Doesn't work for releases like 4.1.0-beta.2 but probably no big deal?

jim-docker (dismissed):

src/common/vars.ts
Nokel81: A good catch, probably a good idea to strip the suffix

src/common/vars.ts
```diff
@@ -44,6 +45,9 @@ export const issuesTrackerUrl = "https://github.com/lensapp/lens/issues";
 export const slackUrl = "https://join.slack.com/t/k8slens/shared_invite/enQtOTc5NjAyNjYyOTk4LWU1NDQ0ZGFkOWJkNTRhYTc2YjVmZDdkM2FkNGM5MjhiYTRhMDU2NDQ1MzIyMDA4ZGZlNmExOTc0N2JmY2M3ZGI";
 export const supportUrl = "https://docs.k8slens.dev/latest/support/";
 
-const docsVersion = isProduction ? `v${packageInfo.version}` : "latest";
+// This explicitly ignores the prerelease info on the package version
+const { major, minor, patch } = new SemVer(packageInfo.version);
+const mmpVersion = [major, minor, patch].join(".");
```
jim-docker, line 50: The `v4.2.0` docs might not exist until `4.2.0` is actually released(?) This would mean 404 for `4.2.0-rc.1`. Just show latest? I guess either way it's wrong. But what you have here will work if someone installs an rc version after the final release was made.

jim-docker (approved):

src/common/vars.ts
jakolehm: I agree with @jim-docker, pre-releases should point to `latest`.

jakolehm (approved):
I still think pre-release should point to "latest" but we can fix that later.

Nokel81:
Shall I fix that in a follow up PR before we release, or it do it now? I completely forgot to implement that @jakolehm .

jakolehm:
> Shall I fix that in a follow up PR before we release, or it do it now? I completely forgot to implement that @jakolehm .

IMHO this PR is already an improvement so follow up PR is fine.

PR #2269, opened 2021-03-02
Persist local-storage state in external json-file

### LINKED ISSUE #2166: Persist the last selected namespace(s)

ixrock:
- fix: keep state and restore on app start (currently it doesn't work cause of random port between app restarts)
- fix: expanding sidebar by clicking to the arrow-icon (for consistency with `compact` mode where no sub-menus)
- refactored: `Sidebar`, `SidebarItem` and `MainLayout` 
- refactored: `createStorage` helper now is observable

related issue #2166

_Demo:_

https://user-images.githubusercontent.com/6377066/109575652-f1ad2100-7afa-11eb-9938-aaa4a0e8d4bf.mov


src/renderer/utils/storageHelper.ts
```diff
@@ -0,0 +1,170 @@
...
+// TODO: write unit/integration tests
+
+import type { CreateObservableOptions } from "mobx/lib/api/observable";
+import { action, comparer, observable, toJS, when } from "mobx";
+import produce, { Draft, setAutoFreeze } from "immer";
+import { isEmpty, isEqual, isFunction } from "lodash";
+
+setAutoFreeze(false); // allow to merge observables
+
+export interface StorageHelperOptions<T = any> extends StorageConfiguration<T> {
```
Nokel81: First of all I think that we should be moving away from `<T = any>` for several reasons:

1. Unless absolutely necessary we should avoid using `any` as it is too easily coercible
2. For something like a storage system and encapsulation layer it should validate the types it deals with, with defence in depth strategies.
3. Lastly, for types like those found this this file, we should be nudging the users to specify a type either implicitly or explicitly and this default blocks that implicit specification.

src/renderer/utils/storageHelper.ts
```diff
@@ -0,0 +1,170 @@
...
+export interface StorageConfiguration<T = any> {
+  storage?: StorageAdapter<T>;
+  observable?: CreateObservableOptions;
+}
+
+export interface StorageAdapter<T = any, C = StorageHelper<T>> {
+  getItem(this: C, key: string): T | Promise<T>; // import
+  setItem(this: C, key: string, value: T): void; // export
+  onChange?(this: C, value: T, oldValue?: T): void;
+}
```
Nokel81: This should be an `abstract class`, same as above with `T = any`.

src/renderer/utils/storageHelper.ts
```diff
@@ -0,0 +1,170 @@
...
+  getItem(key: string) {
+    return JSON.parse(localStorage.getItem(key));
+  },
+  setItem(key: string, value: any) {
+    if (value != null) {
+      localStorage.setItem(key, JSON.stringify(value));
+    } else {
+      localStorage.removeItem(key);
+    }
+  }
```
Nokel81: This seems like it is overloading the function `setItem` too much. We should have a `deleteItem(key: string)` method instead.

src/renderer/utils/storageHelper.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  @action
+  configureStorage(storage: StorageAdapter<T>) {
+    this.storage = Object.getOwnPropertyNames(storage).reduce((storage, name: keyof StorageAdapter) => {
+      storage[name] = storage[name]?.bind(this); // bind storage-adapter methods to "this"-context
+
+      return storage;
+    }, { ...storage });
+  }
```
Nokel81: With the above being an `abstract class` this won't be necessary. Unless you are trying to make it so that the `StorageAdaptor` thinks it has the same `this` context as the `StorageHelper` which seems very nuanced. Can we do away with that?

src/renderer/utils/storageHelper.ts
```diff
@@ -0,0 +1,170 @@
...
+    this.data = observable.box<T>(this.data.get(), {
+      ...StorageHelper.defaultOptions.observable, // inherit default observability options
+      ...options,
+    });
+    this.data.observe(change => {
+      const { newValue, oldValue } = toJS(change, { recurseEverything: true });
+
+      this.onChange(newValue, oldValue);
+    });
+  }
```
Nokel81: This really should be a `private` function unless there is a good reason to change the observability rules "on the fly". But that feels like it would make for some rather hard to debug bugs.

src/renderer/utils/storageHelper.ts
```diff
@@ -0,0 +1,170 @@
...
+  hasValue(value: T) {
+    return !isEmpty(value);
+  }
+
+  isDefaultValue(value: T) {
+    return isEqual(value, this.defaultValue);
+  }
+
+  @action
+  configure({ storage, observable }: StorageConfiguration<T> = this.options): this {
```
Nokel81: This too should be `private`.

src/renderer/utils/storageHelper.ts
```diff
@@ -0,0 +1,170 @@
...
+      const value = await this.load();
+      const notEmpty = this.hasValue(value);
+      const notDefault = !this.isDefaultValue(value);
+
+      if (notEmpty && notDefault) {
+        this.merge(value);
+      }
+      this.initialized = true;
+    } catch (error) {
+      console.error(`[init]: ${error}`, this);
```
Nokel81, line 78: The `[init]` prefix seems a bit too generic for something like this, imo.

src/renderer/utils/storageHelper.ts
```diff
@@ -0,0 +1,170 @@
...
+    this.storage.setItem(this.key, value);
+  }
+
+  async load(): Promise<T> {
+    return this.storage.getItem(this.key);
+  }
+
+  get(): T {
+    return this.data.get();
+  }
```
Nokel81, line 120: This is wrong if the key has not been set. (at least for the default helper)

src/renderer/utils/storageHelper.ts
```diff
@@ -0,0 +1,170 @@
...
+  proxyTo(object: object) {
+    return new Proxy(object, {
+      get: (target: object, prop: PropertyKey, receiver: any) => {
+        return Reflect.get(target, prop, receiver);
+      },
+      set: (target: object, prop: PropertyKey, value: any, receiver: any) => {
+        return Reflect.set(target, prop, value, receiver);
+      }
+    });
+  }
```
Nokel81: Given that this is not used in this PR, and it is listed as an experiment, I think it should be removed.

integration/__tests__/cluster-pages.tests.ts
```diff
@@ -7,14 +7,9 @@
 import { Application } from "spectron";
```
Nokel81, line 7: Do these changes need to be in this PR. Looking at them, not much seems to have changed.

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -71,24 +73,11 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
...
-    // otherwise select "default" or first allowed namespace
-    if (namespaces.has("default")) {
-      return ["default"];
-    } else if (namespaces.size) {
-      return [Array.from(namespaces)[0]];
-    }
-
-    return [];
+    // return all namespaces (empty list) if "default" namespace doesn't exist
+    return selectedNamespaces.get().filter(namespace => namespaces.includes(namespace));
```
Nokel81: I don't think it is a good idea to return "all" if "default" is not present.

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -71,24 +73,11 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
     });
   }
 
-  @computed
   private get initialNamespaces(): string[] {
-    const namespaces = new Set(this.allowedNamespaces);
-    const prevSelectedNamespaces = storage.get();
+    const namespaces = this.allowedNamespaces;
```
Nokel81: Why remove the set?

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -1,25 +1,31 @@
...
-import { autobind, createStorage } from "../../utils";
+import { createStorage, StorageHelper } from "../../local-storage";
+import { autobind } from "../../utils";
 import { dockStore, TabId } from "./dock.store";
 
 interface Options<T = any> {
-  storageName?: string; // name to sync data with localStorage
-  storageSerializer?: (data: T) => Partial<T>; // allow to customize data before saving to localStorage
+  storageName?: string; // persistent key
+  storageSerializer?: (data: T) => Partial<T>; // allow to customize data before saving
```
Nokel81, line 8: Serializationation implies `(data: T) => string` or maybe `Buffer`

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -34,14 +40,19 @@ export class DockTabStore<T = any> {
...
+      return Object.entries(data).reduce((data, [tabId, tabData]) => {
+        data[tabId] = storageSerializer(tabData) as T;
 
-      return [tabId, tabData];
-    });
+        return data;
+      }, data);
+    }
+
+    return data;
```
Nokel81, line 55: This function is incorrect since it says that the data returned from it is `Record<TabId, T>` where infact it is `Record<TabId, Partial<T>>`

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -1,25 +1,31 @@
...
+    const { storageName: storageKey } = this.options;
 
-      this.data.replace(storage.get());
-      reaction(() => this.serializeData(), (data: T | any) => storage.set(data));
+    // auto-save to local-storage
+    if (storageKey) {
+      this.storage = createStorage(storageKey, {});
+      await this.storage.whenReady;
+      this.data.replace(this.storage.get());
+      reaction(() => this.serializeData(), (data: T | any) => this.storage.set(data));
```
Nokel81: Why the `| any`?

src/renderer/components/dock/dock.store.ts
```diff
@@ -21,64 +22,83 @@ export interface IDockTab {
...
+    localStorage.merge({ isOpen: value });
+  }
+
+  get height(): number {
+    return localStorage.get().height;
+  }
+
+  set height(value: number) {
+    localStorage.merge({ height: value });
+  }
```
Nokel81: This should be private as it bypasses the other height checks

src/renderer/components/dock/dock.store.ts
```diff
@@ -21,64 +22,83 @@ export interface IDockTab {
...
+    return localStorage.get().tabs;
+  }
+
+  set tabs(value: IDockTab[]) {
+    localStorage.merge({ tabs: value });
+  }
+
+  get selectedTabId(): TabId {
+    return localStorage.get().selectedTabId || this.tabs[0]?.id;
+  }
```
Nokel81:
```suggestion
  get selectedTabId(): TabId | undefined {
    return localStorage.get().selectedTabId || this.tabs[0]?.id;
  }
```

src/renderer/components/dock/dock.store.ts
```diff
@@ -21,64 +22,83 @@ export interface IDockTab {
...
     return this.tabs.find(tab => tab.id === this.selectedTabId);
   }
 
-  get defaultHeight() {
-    return Math.round(window.innerHeight / 2.5);
+  constructor() {
+    this.init();
+  }
+
+  private async init() {
```
Nokel81: Why is this `async`?

src/renderer/components/dock/dock.store.ts
```diff
@@ -21,64 +22,83 @@ export interface IDockTab {
...
+  height: 300,
+  tabs: [
+    { id: "terminal", kind: TabKind.TERMINAL, title: "Terminal" },
+  ],
+});
 
-  @observable isOpen = false;
+@autobind()
+export class DockStore implements DockStorageState {
+  readonly minHeight = 100;
```
Nokel81, line 41:
```suggestion
  readonly static minHeight = 100;
```

src/renderer/components/dock/dock.store.ts
```diff
@@ -21,64 +22,83 @@ export interface IDockTab {
...
+    localStorage.merge({ tabs: value });
+  }
+
+  get selectedTabId(): TabId {
+    return localStorage.get().selectedTabId || this.tabs[0]?.id;
+  }
+
+  set selectedTabId(value: TabId) {
+    localStorage.merge({ selectedTabId: value });
+  }
```
Nokel81: This should be `private` as it does no validation.

src/renderer/components/dock/dock.store.ts
```diff
@@ -21,64 +22,83 @@ export interface IDockTab {
...
-  @observable tabs = observable.array<IDockTab>(this.initialTabs);
-  @observable selectedTabId = this.defaultTabId;
+
+  get isOpen(): boolean {
+    return localStorage.get().isOpen;
+  }
+
+  set isOpen(value: boolean) {
+    localStorage.merge({ isOpen: value });
+  }
```
Nokel81:
```suggestion
  set isOpen(isOpen: boolean) {
    localStorage.merge({ isOpen });
  }
```

src/renderer/components/layout/sidebar-item.tsx
```diff
@@ -0,0 +1,92 @@
...
+    });
+
+    this.props.onToggle?.(this.id, {
+      props: this.props,
+      event,
+    });
+  };
+
+  render() {
+    const { isHidden, isActive, subMenus = [], icon, text, children, url } = this.props;
```
Nokel81:
```suggestion
    const { isHidden, isActive, subMenus = [], icon, text, children, url, className } = this.props;
```

src/renderer/components/layout/sidebar-item.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    if (isHidden) return null;
+
+    const { id, expanded, compact } = this;
+    const isExpandable = (subMenus.length > 0 || children) && !compact;
+    const className = cssNames(SidebarItem.displayName, this.props.className, {
```
Nokel81:
```suggestion
    const classNames = cssNames(SidebarItem.displayName, className, {
```

src/renderer/components/layout/sidebar-item.tsx
```diff
@@ -0,0 +1,92 @@
...
+      compact,
+    });
+
+    return (
+      <div className={className} data-test-id={id}>
```
Nokel81:
```suggestion
      <div className={classNames} data-test-id={id}>
```

src/renderer/components/layout/sidebar-item.tsx
```diff
@@ -0,0 +1,92 @@
...
+@observer
+export class SidebarItem extends React.Component<SidebarItemProps> {
+  static displayName = "SidebarItem";
+
+  get id(): string {
+    return this.props.id; // unique id, used in storage and integration tests
+  }
+
+  get expanded(): boolean {
+    return sidebarLocalStorage.get().expanded[this.id];
```
Nokel81:
```suggestion
    return Boolean(sidebarLocalStorage.get().expanded[this.id]);
```

src/renderer/components/layout/sidebar-storage.ts
```diff
@@ -0,0 +1,15 @@
...
+  expanded: {
+    [itemId: string]: boolean;
+  }
+}
+
+export const sidebarLocalStorage = createStorage<SidebarLocalStorageModel>("sidebar", {
+  width: 200,     // sidebar size in non-compact mode
+  compact: false, // compact-mode (icons only)
+  expanded: {},
+});
```
Nokel81, line 15: Given that this PR is now changing storage. We could actually make it so that it supports `Map` and `Set` (maybe even by default). That way `expanded` could be `Set<string>` or maybe since we do this very often a new type `ToggleSet<string>` which is the following:

```ts
class ToggleSet<T> extends Set<T> {
  toggle(key: T) {
    if (this.has(key)) {
      this.delete(key);
    } else {
      this.add(key);
    }
  }
}
```

src/renderer/components/layout/sidebar.tsx
```diff
@@ -131,125 +132,123 @@ export class Sidebar extends React.Component<Props> {
   }
 
   render() {
-    const { toggle, isPinned, className } = this.props;
+    const { toggle, compact, className } = this.props;
```
Nokel81, line 135: Why is passing this down manually better in your opinion. I am not suggesting that you change it but I wonder why you did.

src/renderer/local-storage.ts
```diff
@@ -0,0 +1,86 @@
...
+  async load() {
+    try {
+      await fse.ensureDir(this.folderPath, { mode: 0o755 });
+      const state = await fse.readJson(this.filePath);
+
+      this.state.replace(state);
+    } catch (error) {
+    }
+    this.isLoaded = true;
+  }
```
Nokel81: This should either be private or throw an error if not called before `init`.

src/renderer/local-storage.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  private async save(state: object) {
+    if (isEmpty(state)) {
+      return; // skip empty state on clear
+    }
+
+    try {
+      await fse.writeJson(this.filePath, state, { spaces: 2 });
```
Nokel81, line 57: If this is for local storage and should be opaque, why the formatting?

src/renderer/local-storage.ts
```diff
@@ -0,0 +1,86 @@
...
+    } catch (error) {
+      console.error(`[save]: ${error}`, this);
+    }
+  }
+
+  @action
+  clear() {
+    this.state.clear();
+    fse.unlink(this.filePath).catch(() => null);
```
Nokel81:
```suggestion
    fse.unlink(this.filePath).catch(noop);
```

src/renderer/local-storage.ts
```diff
@@ -0,0 +1,86 @@
...
+        await lensLocalStorage.whenReady;
+
+        return lensLocalStorage.state.get(key);
+      },
+      async setItem(key: string, value: any) {
+        lensLocalStorage.state.set(key, value);
+      },
+    },
+  });
+}
```
Nokel81: Why have this function instead of having default values in the `StorageHelper` class? Maybe change it to `Storage` or something less "Helper"-y?

src/renderer/utils/storageHelper.ts
```diff
@@ -0,0 +1,170 @@
...
+
+export interface StorageConfiguration<T = any> {
+  storage?: StorageAdapter<T>;
+  observable?: CreateObservableOptions;
+}
+
+export interface StorageAdapter<T = any, C = StorageHelper<T>> {
+  getItem(this: C, key: string): T | Promise<T>; // import
+  setItem(this: C, key: string, value: T): void; // export
+  onChange?(this: C, value: T, oldValue?: T): void;
```
Nokel81: Why does the `StorageAdaptor` need a reference to the `StorageHelper`? That seems to be an abstraction leak.

src/renderer/utils/storageHelper.ts
ixrock: This is currently internal-only api/helper. I can change `any` to `object` type, how does it sounds?

src/renderer/utils/storageHelper.ts
ixrock: Why?

src/renderer/utils/storageHelper.ts
ixrock: I'd prefer to keep `setItem` and maybe optional `removeItem` (I did it at first previously)

src/renderer/local-storage.ts
ixrock: Why not? :) If you open file manually it's better for debug.

src/renderer/local-storage.ts
ixrock: By default `storageHelper` has minimal dependencies and using `window.localStorage`.  But maybe you're right, let's wait what other fellows think.

src/renderer/utils/storageHelper.ts
ixrock: For usability and subtle fine-tuning when needed. Don't see  any obstacles, and again this is only internal api.

src/renderer/utils/storageHelper.ts
Nokel81: Why not leave it as just `<T>`?

src/renderer/utils/storageHelper.ts
Nokel81: Is there a good reason why? That assumes that all users will want to treat `null` as the same as "give me default".

src/renderer/utils/storageHelper.ts
Nokel81: Just because it is internal doesn't mean we should have this sort of tight coupling. Can you give a good example where this sort of coupling is beneficial?

src/renderer/utils/storageHelper.ts
ixrock: Ok, sounds good.

src/renderer/utils/storageHelper.ts
ixrock: e.g. to get access to `this.initialized` or other public internals of `StorageHelper`

ixrock:
@lensapp/lens-maintainers any ideas why this is failed on Azure? (locally works fine)

image

src/renderer/utils/storageHelper.ts
ixrock: There is no need in one more inheritance layer IMHO. You can consider `StorageAdapter` as base class.

src/renderer/utils/storageHelper.ts
ixrock: Ok, removed `this` context to for less ambiguity.

src/renderer/utils/storageHelper.ts
Nokel81: Why is an inheritance layer too much but explicitly calling `bind()` not? This seems like a much more complicated way of doing inheritance.

src/renderer/components/+namespaces/namespace.store.ts
jakolehm: I agree with @Nokel81 , we should try to keep the existing behaviour here.

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -93,29 +94,20 @@ const defaultProps: Partial<ItemListLayoutProps> = {
...
-    // keep ui user settings in local storage
-    const defaultUserSettings = toJS(this.userSettings);
-    const storage = createStorage<ItemListLayoutUserSettings>("items_list_layout", defaultUserSettings);
-
-    Object.assign(this.userSettings, storage.get()); // restore
-    disposeOnUnmount(this, [
-      reaction(() => toJS(this.userSettings), settings => storage.set(settings)),
-    ]);
+  set showFilters(showFilters: boolean) {
+    localStorage.merge({ showFilters });
```
jakolehm, line 110: These are changes to public api, right?. We cannot make these in a patch release.

src/renderer/utils/storageHelper.ts
ixrock: Oh, btw I have removed the context for `this` in types, but forgot to remove it here. Thanks for spotting :)

src/renderer/utils/storageHelper.ts
Nokel81: I would say that it is a better descriptor of what we want to have.

src/renderer/components/item-object-list/item-list-layout.tsx
ixrock: Yes, we can. It's react component internals (not public `props`). It matters only if you use `ref={}` to get access to this internals, but I doubt anyone is using it..

src/renderer/components/+namespaces/namespace.store.ts
ixrock: It's only for initial first run. Otherwise picking the first namespace also not the best solution..

Nokel81: Why, I think it is a better default than loading everything if `"default"` isn't there.

jakolehm: Still we shouldn't change this behaviour in a patch release. There is a reason why we don't default to all namespaces -> it might easily burn clusters 🔥 .

src/renderer/components/layout/sidebar-item.tsx
ixrock: You cannot override `const`

src/renderer/components/layout/sidebar.tsx
ixrock: Instead of direct access from `sidebarLocalStorage` you mean? To make `Sidebar` less dependent and testable + `compact` also used in `MainLayout` where it comes from.

src/renderer/components/layout/sidebar-item.tsx
Nokel81: I know, this pattern (calling the return value of a `cssName(..)` call `classNames`) is used in other places in the code base already. I was merely suggesting use it.

src/renderer/utils/storageHelper.ts
ixrock: since it logs `this` at the same time => this is fine :) (you can easily spot class/component for debug)

src/renderer/utils/storageHelper.ts
ixrock: What the key? It returns whole storage state (`defaultValue` by default)

Nokel81: No it calls `get` on the `observable.box` which returns `undefined` when not set.

src/renderer/components/layout/sidebar.tsx
Nokel81: Sounds good.

src/renderer/utils/storageHelper.ts
ixrock: Binding to `this` context has removed.

Nokel81:
I guess a philosophical question is why use `localStorage` at all if we are just persisting the file, why not use use `BaseStore` instead?

ixrock:
> I guess a philosophical question is why use `localStorage` at all if we are just persisting the file, why not use use `BaseStore` instead?

It doesn't work if store used only in `renderer` (saving happens in `main`) + it's cumbersome for such task.
p.s. I tried to use it at first pace.

Nokel81:
I just think that we could do away with have two sources of truth if we only used a file. (Bypassing saving stuff to localStorage)

jakolehm:
This PR is quite big change, probably too big for a patch release. Couldn't we just swap implementation within `StorageHelper` so that it does not require _any_ other changes?

ixrock:
> This PR is quite big change, probably too big for a patch release. Couldn't we just swap implementation within `StorageHelper` so that it does not require _any_ other changes?

Don't think so.. I tried to touch only things related to `createStorage` util during this work (minimal changes + small fixes to UI components). (let's wait not a patch release then)

ixrock:
> I just think that we could do away with have two sources of truth if we only used a file. (Bypassing saving stuff to localStorage)

Why you think there are 2 source-of-truths? This PR doesn't store any data to `window.localStorage`

nevalla:
I think @jakolehm meant that why not just to create storage implementation that stores data into file-system with the same interface as `localStorage` and use that storage instead of `localStorage` in StorageHelper?

```
protected get storage() {
    if (this.options.useSession) return window.sessionStorage;

    return lensLocalStorage; // or whatever implements the storage interface
  }
```

ixrock:
> I think @jakolehm meant that why not just to create storage implementation that stores data into file-system with the same interface as `localStorage` and use that storage instead of `localStorage` in StorageHelper?

Because `localStorage` (WebStorage API methods) are `sync` to fully replace it transparently.  For that reason introduced `StorageAdapter` with similar interfaces, but with `async` + `observable` support which is heavily reducing amount of boilerplate.

jakolehm:
Going to move this to 4.2.0 milestone. If we want to have a fix for 4.1.x release it should be minimal (probably just replace `window.localStorage`).

ixrock:
Rebased to https://github.com/lensapp/lens/pull/2280

PR #2273, opened 2021-03-02
Fix font-size on <code>

### LINKED ISSUE #1683: Default font-size style for <code> elements is invalid
The default `font-size` isn't using CSS vars:

image

It should be

`font-size: calc(var(--font-size) * .9);`

instead.

Nokel81:

fixes #1683

### DIFF
--- a/src/renderer/components/app.scss
+++ b/src/renderer/components/app.scss
@@ -151,7 +151,7 @@ code {
   vertical-align: middle;
   border-radius: $radius;
   font-family: $font-monospace;
-  font-size: calc($font-size * .9);
+  font-size: calc(var(--font-size) * .9);
   color: #b4b5b4;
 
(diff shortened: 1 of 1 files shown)


ixrock (approved):

jakolehm:
@Nokel81 how does this fix change UI? Screenshot would be nice.

Nokel81:
Old:
Screen Shot 2021-03-03 at 9 09 29 AM


Fixed:
Screen Shot 2021-03-03 at 9 08 21 AM

This is because the font size here is actually slightly bigger than the `--font-size-small` that was the fallback. Probably why we didn't notice this. Should I just change it to `--font-size-small` instead?

nevalla (approved):

stefcameron:
Thank you!

PR #2279, opened 2021-03-04
Persist local-storage in external json-file

### LINKED ISSUE #2166: Persist the last selected namespace(s)

ixrock:
- persisting local-storage state in json-files located at `/Users/%USER%/Library/Application Support/LensDev/lens-local-storage/%clusterId%.json`
- refactored `renderer/utils/createStorage` / now it's `observable`
- refactored `Sidebar`, `SidebarItem` and `MainLayout`
- ~~expanding sidebar by clicking arrow-icon (for consistency with compact mode where no sub-menus)~~

close #2166


src/renderer/utils/createStorage.ts
```diff
@@ -1,73 +1,210 @@
...
+    },
+    observable: {
+      deep: true,
+      equals: comparer.shallow,
     }
-  }
+  };
+
+  private data = observable.box<T>();
+  @observable.ref storage: StorageAdapter<T>;
```
Nokel81: Given that this is is only really assigned in the constructor and that it really shouldn't be changed on the fly. I don't think that this should be either `observable` or `public`. Futhermore, given that this class also stores its `options` wholly this might as well just be a getter.
```suggestion
  private get storage(): StorageAdapter<T> {
    return this.options.storage;
  }
```

src/renderer/utils/createStorage.ts
```diff
@@ -1,73 +1,210 @@
...
 
-  set(value: T) {
-    this.storage.setItem(this.key, JSON.stringify(value));
+  @action
+  private configure({ storage, observable }: StorageConfiguration<T> = this.options): this {
+    if (storage) this.storage = storage;
+    if (observable) this.configureObservable(observable);
 
     return this;
   }
```
Nokel81: Given the above suggestion and that this leaves this function only as an `if (observable)` I would just move that to the `constructor`.

Furthermore, given the rules of how spreading multiple objects into a new object works. I would unconditionally call `this.configureObservable`.

src/renderer/utils/createStorage.ts
```diff
@@ -1,73 +1,210 @@
...
+}
 
-    if (this.options.addKeyPrefix) {
-      this.key = StorageHelper.keyPrefix + key;
+export class StorageHelper<T> {
+  static defaultOptions: StorageHelperOptions<any> = {
+    autoInit: true,
+    get storage() {
+      return jsonFileSyncStorageAdapter;
+    },
```
Nokel81: We shouldn't have a default storage adapter here as it doesn't actually track the correct type of `T`

src/renderer/utils/createStorage.ts
```diff
@@ -1,73 +1,210 @@
...
+  getItem(key: string): T; // import
+  setItem(key: string, value: T): void; // export
+  removeItem?(key: string): void; // if not provided setItem(key,undefined) will be used
+  onChange?(data: { key: string, value: T, oldValue?: T }): void;
+}
 
-    if (this.options.addKeyPrefix) {
-      this.key = StorageHelper.keyPrefix + key;
+export class StorageHelper<T> {
+  static defaultOptions: StorageHelperOptions<any> = {
```
Nokel81: Default options should be untyped (see bellow for reason).

src/renderer/utils/createStorage.ts
```diff
@@ -1,73 +1,210 @@
...
+      return jsonFileSyncStorageAdapter;
+    },
+    observable: {
+      deep: true,
+      equals: comparer.shallow,
     }
-  }
+  };
+
+  private data = observable.box<T>();
```
Nokel81: Given that we have default options, we should type them as fully supplied. Furthermore, there shouldn't be a need for an outside inspector (except in tests where we can cast to any anyway) to view the options after construction.
```suggestion

  private options: Required<StorageHelperOptions<T>>;
  private data = observable.box<T>();
```

src/renderer/utils/createStorage.ts
```diff
@@ -1,73 +1,210 @@
...
+export interface StorageHelperOptions<T> extends StorageConfiguration<T> {
+  autoInit?: boolean; // default: true
+}
 
-  static defaultOptions: IStorageHelperOptions = {
-    addKeyPrefix: true,
-    useSession: false,
-  };
+export interface StorageConfiguration<T> {
+  storage?: StorageAdapter<T>;
```
Nokel81: Given that the default storage adaptor in this PR doesn't cover all `T` this field should be made required. (I don't think that is a breaking change since we don't export `StorageHelper`).

src/renderer/utils/createStorage.ts
```diff
@@ -1,73 +1,210 @@
...
+  };
+
+  private data = observable.box<T>();
+  @observable.ref storage: StorageAdapter<T>;
+  @observable initialized = false;
+  whenReady = when(() => this.initialized);
 
-  protected get storage() {
-    if (this.options.useSession) return window.sessionStorage;
+  constructor(readonly key: string, readonly defaultValue?: T, readonly options: StorageHelperOptions<T> = {}) {
```
Nokel81: Given that options now has a required field it should be a required argument.

```suggestion
  constructor(readonly key: string, options: StorageHelperOptions<T>, readonly defaultValue?: T) {
```

src/renderer/utils/createStorage.ts
```diff
@@ -1,73 +1,210 @@
...
+  },
+  setItem(key: string, value: any) {
+    localStorage.setItem(key, JSON.stringify(value));
+  },
+  removeItem(key: string) {
+    localStorage.removeItem(key);
+  }
+};
+
+// TODO: remove after merge https://github.com/lensapp/lens/pull/2269
```
Nokel81: I assume this is the PR this is supposed to relate to?

```suggestion
// TODO: remove after merge https://github.com/lensapp/lens/pull/2280
```

src/renderer/utils/createStorage.ts
```diff
@@ -1,73 +1,210 @@
...
+export const jsonFileSyncStorageAdapter: StorageAdapter<object> = {
+  cwd: path.resolve((app || remote.app).getPath("userData"), "lens-local-storage"),
+  data: observable.map<string, any>([], { deep: false }),
+  initialized: false,
+
+  get filePath() {
+    const clusterId = location.hostname.split(".").slice(-2, -1)[0];
```
Nokel81: This should be gotten from `getHostedClusterId()`

src/renderer/utils/createStorage.ts
```diff
@@ -1,73 +1,210 @@
...
+      }
+    });
+  },
+
+  getItem(key: string) {
+    this.init();
+
+    return this.data.get(key);
+  },
+  setItem(key: string, value: any) {
```
Nokel81: We shouldn't have this sort of coercion in a storage abstraction.
```suggestion
  setItem(key: string, value: object) {
```

src/renderer/utils/createStorage.ts
```diff
@@ -1,73 +1,210 @@
...
+  }
+
+  clear() {
+    this.data.set(null);
   }
 
-  restoreDefaultValue() {
-    return this.set(this.defaultValue);
+  merge(value: Partial<T> | ((draft: Draft<T>) => Partial<T> | void)) {
+    const updater = isFunction(value) ? value : (state: Draft<T>) => {
```
Nokel81: Is there a reason why you don't use `typeof`? We aren't targeting Safari 9 so I don't see the benefit of calling lodash's function here

ixrock: Because I use already `lodash` in that module, otherwise I prefer `typeof`

src/renderer/utils/createStorage.ts
ixrock: Yes, but no.. This is `renderer/utils/*` which should be with minimal dependencies, especially from `main` module.

Nokel81: I guess I didn't realize that `getHostedClusterId()` was in `main`. That seems like a strange place to have it.

ixrock: Technically it's in `src/common/cluster-store.ts` but I consider it more as `main` stuff which should be avoided in `renderer`. After moving `getHostedClusterId` to appropriate place I'll update it here.

Nokel81: How is it `main` then? It is used by `getHostedCluster()` all the time.

src/renderer/utils/createStorage.ts
ixrock: Doesn't really matter since the line cleaned up in follow-up PR (#2280).

src/renderer/utils/createStorage.ts
ixrock: Updated in #2280

Nokel81: May I ask why you are updating a follow-up PR with changes requested on this PR?

ixrock: Because it's not targeted to patch-release fix anymore (suggesting to merge #2280 first)

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -5,14 +5,14 @@ import { Namespace, namespacesApi } from "../../api/endpoints/namespaces.api";
...
-const storage = createStorage<string[]>("context_namespaces");
+const selectedNamespaces = createStorage<string[]>("selected_namespaces");
 
 export const namespaceUrlParam = createPageParam<string[]>({
   name: "namespaces",
   isSystem: true,
   multiValues: true,
   get defaultValue() {
-    return storage.get() ?? []; // initial namespaces coming from URL or local-storage (default)
+    return selectedNamespaces.get() ?? []; // initial namespaces coming from URL or local-storage (default)
```
Nokel81: Why not pass `[]` in as the default value?

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -1,25 +1,30 @@
 import { autorun, observable, reaction } from "mobx";
-import { autobind, createStorage } from "../../utils";
+import { autobind, createStorage, StorageHelper } from "../../utils";
 import { dockStore, TabId } from "./dock.store";
 
 interface Options<T = any> {
-  storageName?: string; // name to sync data with localStorage
-  storageSerializer?: (data: T) => Partial<T>; // allow to customize data before saving to localStorage
+  storageName?: string; // persistent key
```
Nokel81: The key isn't persistent but the data behind the key

```suggestion
  storageName?: string; // persistence key
```

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -1,25 +1,30 @@
...
+  storageSerializer?: (data: T) => Partial<T>; // allow to customize data before saving
```
Nokel81: We aren't doing serialization here.

```suggestion
  storageFinalizer?: (data: T) => Partial<T>; // allow to customize data before saving
```

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -1,25 +1,30 @@
...
 }
 
 @autobind()
 export class DockTabStore<T = any> {
-  protected data = observable.map<TabId, T>([]);
+  private storage?: StorageHelper<Record<TabId, T>>;
```
Nokel81: Do we even need to have this as a field? It seems like it gets created and then can be handed off to the reaction.

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -34,14 +39,19 @@ export class DockTabStore<T = any> {
     });
   }
 
-  protected serializeData() {
+  protected serializeData(): Record<TabId, T> {
```
Nokel81: There is no grantee given the implementation that the values will be complete. We shouldn't mislead future maintainers.

```suggestion
  protected finalizeData(): Record<TabId, Partial<T>> {
```

src/renderer/components/dock/dock.store.ts
```diff
@@ -21,64 +21,87 @@ export interface IDockTab {
...
+  isOpen?: boolean;
+}
+
 @autobind()
-export class DockStore {
-  protected initialTabs: IDockTab[] = [
-    { id: "terminal", kind: TabKind.TERMINAL, title: "Terminal" },
-  ];
+export class DockStore implements DockStorageState {
+  readonly minHeight = 100;
```
Nokel81, line 33:
```suggestion
  static private readonly MinHeight = 100;
```

src/renderer/components/dock/dock.store.ts
```diff
@@ -21,64 +21,87 @@ export interface IDockTab {
...
+  @observable fullSize = false;
 
-  protected storage = createStorage("dock", {}); // keep settings in localStorage
-  public readonly defaultTabId = this.initialTabs[0].id;
-  public readonly minHeight = 100;
+  private storage = createStorage<DockStorageState>("dock", {
+    height: 300,
+    tabs: [
+      { id: "terminal", kind: TabKind.TERMINAL, title: "Terminal" },
+    ],
```
Nokel81, line 40: Why not have tabs be a `Map<string, DockTab>`? That way we don't need to do all those iterations.

src/renderer/components/layout/sidebar-storage.ts
```diff
@@ -0,0 +1,15 @@
+import { createStorage } from "../../utils";
+
+export interface SidebarStorageState {
+  width: number;
+  compact: boolean;
+  expanded: {
+    [itemId: string]: boolean;
+  }
```
Nokel81, line 8: Better describes what we want

```suggestion
  expanded: Set<string>,
```

src/renderer/utils/createStorage.ts
```diff
@@ -1,73 +1,211 @@
...
+    observable: observableOptions,
+    defaultValue,
+  });
 }
 
-export function createStorage<T>(key: string, defaultValue?: T, options?: IStorageHelperOptions) {
-  return new StorageHelper(key, defaultValue, options);
+export type StorageModel = Record<string, any>; // any json-compatible model
+
+export interface StorageAdapter<T = StorageModel> {
```
Nokel81: This doesn't really do what I think you mean. Plus I don't think it should matter to the consumers how to data is serialized.
```suggestion
export interface StorageAdapter<T> {
```

src/renderer/utils/createStorage.ts
```diff
@@ -1,73 +1,211 @@
...
+  toJS() {
+    return toJS(this.get(), { recurseEverything: true });
   }
 }
+
+/**
+ * Keep intended window.localStorage state in external JSON-file.
+ * Reason: app creates random ports between restarts and as a result storage not persistent.
+ */
+export const jsonFileStorageAdapter: StorageAdapter = {
```
Nokel81: I think that this should be a function that takes validators for the deserialization step (from a defence in depth pov).

Furthermore `any` isn't necessarily JSON serializable so I think a better default would be to use a different format. If the data was sent to `main` and then serialized there we could use `v8.serialize` and get most of this for "free". The other option would be to have our own serialization function that intercepts some types and encodes them in a strict sense.

src/renderer/utils/createStorage.ts
```diff
@@ -1,73 +1,211 @@
...
+  },
+
+  async init() {
+    if (this.initialized) return;
+
+    const data = await readJson(this.filePath).catch(noop);
+
+    if (data) {
+      this.data.replace(data);
+    }
```
Nokel81:
```suggestion
    try {
      const data = await readJson(this.filePath);

      this.data.replace(data);
    } catch () {}
```

src/renderer/components/dock/dock-tab.store.ts
ixrock: There is a missing `this.storage?.clear()` in `clear()`, will be added.

Nokel81: okay sounds good.

src/renderer/components/dock/dock.store.ts
ixrock: It sticks to more plain json-format (this is out of scope of PR) / could be improved later (_dock.store.ts_)

Nokel81: okay sounds good

jim-docker:
The sidebar submenus don't open?

https://user-images.githubusercontent.com/40840436/110554436-ce8cfd80-8108-11eb-8af0-6502af23fc2c.mov

ixrock:
> The sidebar submenus don't open?

@jim-docker  from PR's description:

> - expanding sidebar by clicking arrow-icon (for consistency with compact mode where no sub-menus)

src/renderer/components/dock/dock-tab.store.ts
ixrock: Out of PR's scope

src/renderer/components/layout/sidebar-storage.ts
ixrock: Let's not postpone milestone with such micro optimisations? (could be improved later on demand / follow-up PRs)

src/renderer/components/dock/dock-tab.store.ts
Nokel81: I acknowledge that switching to `Set` and `Map` is out of scope, but why is this out of scope?

ixrock: To not postpone **4.2.0**, all these suggestions are non-critical.

ixrock:
@lensapp/lens-maintainers integration tests are fixed, please re-review.

src/renderer/components/dock/dock-tab.store.ts
Nokel81: Why is allowing incorrect names through in a PR postponing the milestone?

ixrock: I have nothing to say anymore. Too many _why_..
Only bugs/critical issues I'm up to discuss and change for this PR.

jim-docker:
> > The sidebar submenus don't open?
> 
> @jim-docker from PR's description:
> 
> > * expanding sidebar by clicking arrow-icon (for consistency with compact mode where no sub-menus)

I saw that but didn't quite understand. (thought I tried clicking arrows too, I will check again)

jim-docker:
> > The sidebar submenus don't open?
> 
> @jim-docker from PR's description:
> 
> > * expanding sidebar by clicking arrow-icon (for consistency with compact mode where no sub-menus)

Sorry, works as advertised. But personally I prefer the way it was before. I don't use compact mode so that consistency is not important to me. However the way it was before is consistent with other drop-downs in the Lens UI such as the namespace select drop-down. Plus I know it's a little late to suggest but this could've/should've been done in a separate PR?

ixrock:
>> Plus I know it's a little late to suggest but this could've/should've been done in a separate PR?

It could be easily fixed, will update soonish.

ixrock:
@jim-docker sidebar's expand logic reverted and tests are green again. Please take a look.

src/renderer/utils/createStorage.ts
```diff
@@ -1,73 +1,211 @@
...
+/**
+ * Keep intended window.localStorage state in external JSON-file.
+ * Reason: app creates random ports between restarts and as a result storage not persistent.
+ */
+export const jsonFileStorageAdapter: StorageAdapter = {
+  cwd: path.resolve((app || remote.app).getPath("userData"), "lens-local-storage"),
+  data: observable.map<string, any>([], { deep: false }),
+  initialized: false,
+
+  get filePath() {
```
jim-docker: Not sure where it would be done but these files (for a cluster) should be deleted when the cluster is removed from Lens. Especially since the same cluster gets a new id every time it is added to Lens.

src/renderer/utils/createStorage.ts
```diff
@@ -1,73 +1,211 @@
...
+
+  async init() {
+    if (this.initialized) return;
+
+    const data = await readJson(this.filePath).catch(noop);
+
+    if (data) {
+      this.data.replace(data);
+    }
+    this.bindAutoSave();
```
jim-docker: this should get disposed when a cluster is removed?

jim-docker (commented):
Looks good but I'd like to test a bit more tomorrow. I noticed my selected namespaces got reset to `[]` when connecting to a flaky kind cluster. Not sure where the flakiness is coming from (seen it on master too, with minikube as well)

src/renderer/utils/createStorage.ts
ixrock: Yes, those are should be deleted, will take a look.

src/renderer/utils/createStorage.ts
ixrock: Not sure, since everything wiping out from "cluster-scope" including `iframe` (where this code actually running).

ixrock:
> I noticed my selected namespaces got reset to [] when connecting to a flaky kind cluster. Not sure where the flakiness is coming from (seen it on master too, with minikube as well)

Would be nice to have reproducable scenario for `minikube` on local machine. (will test more)

jim-docker:
Haven't reproduced the `[]` namespace issue, but did notice this (not major):


https://user-images.githubusercontent.com/40840436/110974041-1344ae80-832c-11eb-98c5-badce5de0df8.mov

Doesn't save the right `selectedTabId` so after reload (or restart) it is not what was last selected

ixrock:
> Doesn't save the right `selectedTabId` so after reload (or restart) it is not what was last selected

Weird, can't reproduce the issue (page reloading same as from the menu, but with `Cmd+R` hotkey):

https://user-images.githubusercontent.com/6377066/110999523-2afe5000-8389-11eb-9ea6-274ab0b4e4d2.mov

But I noticed another issue with that tabs, will investigate it further..

jim-docker:
> Weird, can't reproduce the issue (page reloading same as from the menu, but with `Cmd+R` hotkey):

I can definitely reproduce it, but can't nail it down consistently. I noticed another strange behaviour. I built and ran from master, then returned to this branch, built and ran, and it reverted to default behaviour for all clusters, at least for the dock (no tabs). It shouldn't depend on the build, should it? Should I be testing the dist binary?

jim-docker:
Starting fresh, no files in `~/Library/Application Support/LensDev/lens-local-storage`
run `make dev`, then

https://user-images.githubusercontent.com/40840436/111006176-50c03080-835a-11eb-9c83-1d5fac48b202.mov

The contents of the local storage file just before (and after) reloading is:
```{
  "selected_namespaces": [
    "default"
  ],
  "cluster_overview": {
    "metricType": "cpu",
    "metricNodeRole": "master"
  },
  "dock": {
    "height": 300,
    "tabs": [
      {
        "id": "terminal",
        "kind": "terminal",
        "title": "Terminal"
      },
      {
        "id": "6985603517e2425b77961bf82d4c28b4",
        "kind": "terminal",
        "title": "Terminal (2)"
      }
    ],
    "isOpen": true,
    "selectedTabId": "6985603517e2425b77961bf82d4c28b4"
  }
}
```
but `selectedTabId` should be `terminal`

ixrock:
@jim-docker I did some refactoring, seems to work fine now. 
Also added removing json-file on cluster deletion. Please try it out more / re-review.

src/renderer/utils/createStorage.ts
ixrock: Fixed

ixrock:
> Should I be testing the dist binary?

Nope, development version from the branch should be enough.

jim-docker (dismissed):
Nice refactor!

aleksfront:
@ixrock It works good with main part of the PR. However, I found some Sidebar-refactor-related visual issues. I prefer moving them to another issue/pr, but bug with custom resources menu won't allow me to do that (it must be fixed in here).
1. Custom resource groups lost their labels and they cannot be expanded.
dissapeared groups
2. Submenus need to be moved right slightly to align parent item text.
misplaced submenus
3. While sidebar in it's default width (200px), arrow button covers `Custom Resources` text a bit. This happens when scrollbar appeared. Make icon area smaller a bit?
misplaced chevron

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -1,25 +1,30 @@
...
 
 @autobind()
 export class DockTabStore<T = any> {
-  protected data = observable.map<TabId, T>([]);
+  private storage?: StorageHelper<Record<TabId, T>>;
+  protected data = observable.map<TabId, T>();
 
   constructor(protected options: Options<T> = {}) {
-    const { storageName } = options;
+    this.init();
```
Nokel81: Please catch any possible error here.

src/renderer/components/dock/dock-tab.store.ts
Nokel81: That is not how PRs should work. This current name is highly misleading and we shouldn't not be restrict what we can talk about because a release is coming up. If something isn't good enough then we will just bump it to the next release.

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -34,14 +39,19 @@ export class DockTabStore<T = any> {
...
-  protected serializeData() {
+  protected serializeData(): Record<TabId, Partial<T>> {
+    const data = this.data.toJSON();
     const { storageSerializer } = this.options;
 
-    return Array.from(this.data).map(([tabId, tabData]) => {
-      if (storageSerializer) return [tabId, storageSerializer(tabData)];
+    if (storageSerializer) {
+      return Object.entries(data).reduce((data, [tabId, tabData]) => {
+        data[tabId] = storageSerializer(tabData) as T;
```
Nokel81: this cast is no longer necessary

src/renderer/utils/createStorage.ts
```diff
@@ -1,73 +1,70 @@
...
+    fse.readJson(jsonFilePath)
+      .then((data = {}) => storage.merge(data))
+      .catch(() => null) // ignore empty / non-existing / invalid json files
+      .finally(() => loaded.set(true));
+
+    // bind auto-saving
+    reaction(() => storage.toJSON(), saveFile, { delay: 250 });
+
+    // remove json-file when cluster deleted
+    when(() => clusterStore.removedClusters.has(clusterId)).then(removeFile);
```
Nokel81: We shouldn't have this call if `clusterId` is undefined

Nokel81:
The overall now looks pretty good

src/renderer/components/dock/dock-tab.store.ts
jim-docker: If the name is more appropriate it's an easy change, right? This function is not exported to extensions?

src/renderer/components/dock/dock-tab.store.ts
ixrock: Without casting I got this:

image

src/renderer/utils/createStorage.ts
ixrock: Fixed

src/renderer/components/dock/dock-tab.store.ts
ixrock: Refactored, PTAL

src/renderer/components/dock/dock-tab.store.ts
ixrock: Refactored a bit, PTAL if still needs extra `try-catch`

src/renderer/components/dock/dock-tab.store.ts
ixrock: This functionality moved out from class initialization `options` to `protected` class method (overwritten in `edit-resource.store.ts`)

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -34,14 +49,18 @@ export class DockTabStore<T = any> {
...
+  protected getStorableData(): DockTabStorageState<T> {
+    const allTabsData = this.data.toJSON();
 
-      return [tabId, tabData];
-    });
+    return Object.entries(allTabsData).reduce((data, [tabId, tabData]) => {
+      data[tabId] = this.serializeBeforeSave(tabData);
+
+      return data;
+    }, allTabsData);
```
Nokel81: I would recommend recreating it.
```suggestion
    return Object.fromEntries(
      Object.entries(allTabsData)
        .map(([tabId, tabData]) => ([
          tabId,
          this.finalizeDataForSave(tabData),
        ])
    );
```

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -34,14 +49,18 @@ export class DockTabStore<T = any> {
     });
   }
 
-  protected serializeData() {
-    const { storageSerializer } = this.options;
+  protected serializeBeforeSave(data: T): T {
```
Nokel81: Why do you insist that this is called "serialize"? This function isn't doing the following (from wikipedia)

>  translating a data structure or object state into a format that can be stored

That would be a call to `JSON.serialize` or similar.

Nokel81 (commented):
one last thing

src/renderer/components/dock/dock-tab.store.ts
ixrock: Ok, I'll update the method name

Nokel81 (dismissed):

ixrock:
@Nokel81 @aleksfront @jim-docker 

Thanks for detailed reviews. Ready for review again, hopefully this would be a last fix to PR :)

https://user-images.githubusercontent.com/6377066/111500590-d6722080-874c-11eb-8f95-c1b34a927ffe.mp4

src/renderer/components/+custom-resources/crd-list.tsx
```diff
@@ -60,13 +68,9 @@ export class CrdList extends React.Component {
         className="CrdList"
         isClusterScoped={true}
         store={crdStore}
+        items={items}
         sortingCallbacks={sortingCallbacks}
         searchFilters={Object.values(sortingCallbacks)}
-        filterItems={[
```
ixrock, line 65: Btw, this filtering seems to be broken in `4.1.4`

src/renderer/utils/createStorage.ts
```diff
@@ -1,73 +1,72 @@
...
-  get(): T {
-    const strValue = this.storage.getItem(this.key);
-
-    if (strValue != null) {
-      try {
-        return JSON.parse(strValue);
-      } catch (e) {
-        console.error(`Parsing json failed for pair: ${this.key}=${strValue}`);
-      }
+  async function saveFile(json = {}) {
```
aleksfront, line 40: Maybe `saveToAppFiles()` could be used instead? `src/common/utils/saveToAppFiles.ts`

aleksfront (approved):
Work as expected.

src/renderer/utils/createStorage.ts
ixrock: Maybe, but it uses `fs-extra` sync file operations, e.g. `fse.writeFileSync`

src/renderer/components/+custom-resources/crd-list.tsx
Nokel81: Broken how, like it isn't being called at all? That is not good since we have to keep that prop working as it is exported.

Nokel81 (approved):
Please leave merging to me, I need to coordinate it with the rc.1 PR

ixrock:
> Broken how, like it isn't being called at all? That is not good since we have to keep that prop working as it is exported.

Well, I don't know for sure, but in Lens `4.1.4` it behaves like this for some reasons with `props.filterItems`: 

https://user-images.githubusercontent.com/6377066/111630117-b00abe80-87fa-11eb-9af0-17425ff54b75.mov

_PR's fix_:

https://user-images.githubusercontent.com/6377066/111630633-3cb57c80-87fb-11eb-8550-dfb3e29d18a5.mov

Nokel81:
Ah thanks for the clarification. Not a good showing since we shouldn't break our APIs like that (since I think that was working pre-4.1

PR #2280, opened 2021-03-04
Refactoring usages of renderer/utils/createStorage

ixrock:
Follow-up for #2279 

- refactored usages of `renderer/utils/createStorage`
- refactored `Sidebar`, `SidebarItem` and `MainLayout`
- expanding sidebar by clicking arrow-icon (for consistency with compact mode where no sub-menus)


src/renderer/components/layout/sidebar-item.tsx
```diff
@@ -0,0 +1,90 @@
...
+  icon?: React.ReactNode;
+  isHidden?: boolean;
+  isActive?: boolean;
+  subMenus?: TabLayoutRoute[];
+  onToggle?(id: string, meta: { props: SidebarItemProps, event: React.MouseEvent }): void;
+}
+
+@observer
+export class SidebarItem extends React.Component<SidebarItemProps> {
+  static displayName = "SidebarItem";
```
ixrock, line 25: Probably we should start using this static field in other places too to improve debugging react-components in `production`.

Nokel81: Should that be a PR of its own?

ixrock:
@Nokel81 PTAL / re-review. 

I'm suggesting to merge this first since targeted to #2279 / **4.2.0** (not a patch release anymore) and continue reviewing in target-branch. Wdyt?

Nokel81:
Sounds good, a single place to review is better.

src/renderer/components/layout/sidebar-item.tsx
ixrock: Maybe, it's just an example..

PR #2281, opened 2021-03-05
VSCode launch configurations for debugging

Nachasic:
## Changes
 - `/.gitignore` - Remove `.vscode` directory from;
 - `/.vscode/launch.json` - Add launch configurations for debugging Electron's renderer and main processes, unit and integrations tests;
 - `/package.json` - Add `debug-build` script to build main process for debugging;
 - `/.vscode/tasks.json` - Add a task to build main process for debugging;
 - `/src/renderer/bootstrap.tsx` - Add code to wait for Chrome debugger attachment in development builds.

### Main process
1. In a separate terminal launch dev-server (`yarn dev:renderer`) so Electron can pull renderer resources from it;
1. Set up breakpoints in VSCode or use `debugger` keyword;
1. Use `Main Process` configuration to launch.

### Renderer Process
1. Launch dev build (`make dev`);
1. Set up breakpoints in VSCode or use `debugger` keyword;
1. Use `Renderer Process` configuration to attach debugger.

**Note:** Since VSCode doesn't support background dependent tasks in launch configurations, I couldn't find a setup that would allow to debug both main and renderer processes in Electron simultaneously.

### Integration and Unit Tests
1. Set up breakpoints in VSCode or use `debugger` keyword;
1. Use `Integration Tests` and `Unit Tests` configurations respectively.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -34,8 +34,24 @@
             "name": "Integration Tests",
             "type": "node",
             "request": "launch",
-            "console": "externalTerminal",
-            "runtimeArgs": ["${workspaceRoot}/node_modules/.bin/jest", "--runInBand", "integration"],
+            "console": "integratedTerminal",
+            "runtimeArgs": [
+                "${workspaceFolder}/node_modules/.bin/jest",
+                "--runInBand",
+                "integration"
+            ],
         },
+        {
+            "name": "Unit Tests",
+            "type": "node",
+            "request": "launch",
+            "internalConsoleOptions": "openOnSessionStart",
+            "program": "${workspaceFolder}/node_modules/jest/bin/jest.js",
+            "args": [
+                "--env=jsdom",
+                "-i",
+                "src"
+            ]
+        }
     ],
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,11 +5,23 @@
     "version": "0.2.0",
     "configurations": [
         {
+            "name": "Main Process",
             "type": "node",
             "request": "launch",
-            "name": "Integration tests",
+            "cwd": "${workspaceFolder}",
+            "runtimeExecutable": "${workspaceFolder}/node_modules/.bin/electron",
+            "windows": {
+              "runtimeExecutable": "${workspaceFolder}/node_modules/.bin/electron.cmd"
+            },
+            "args" : ["."],
+            "outputCapture": "std"
+        },
+        {
+            "name": "Integration Tests",
+            "type": "node",
+            "request": "launch",
             "console": "externalTerminal",
             "runtimeArgs": ["${workspaceRoot}/node_modules/.bin/jest", "--runInBand", "integration"],
-        }
+        },
     ]
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -9,12 +9,26 @@
             "type": "node",
             "request": "launch",
             "cwd": "${workspaceFolder}",
+            "protocol": "inspector",
+            "preLaunchTask": "compile-dev",
             "runtimeExecutable": "${workspaceFolder}/node_modules/.bin/electron",
             "windows": {
               "runtimeExecutable": "${workspaceFolder}/node_modules/.bin/electron.cmd"
             },
-            "args" : ["."],
+            "runtimeArgs": [
+                "--remote-debugging-port=9223",
+                "--inspect",
+                "."
+            ],
             "outputCapture": "std"
+        },
+        {
+            "name": "Renderer Process",
+            "type": "pwa-chrome",
+            "request": "attach",
+            "port": 9223,
+            "webRoot": "${workspaceFolder}",
+            "timeout": 30000
         },
         {
             "name": "Integration Tests",
@@ -23,5 +37,5 @@
             "console": "externalTerminal",
             "runtimeArgs": ["${workspaceRoot}/node_modules/.bin/jest", "--runInBand", "integration"],
         },
-    ]
+    ],
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,15 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "Integration tests",
+            "console": "externalTerminal",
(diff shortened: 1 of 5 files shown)


src/renderer/bootstrap.tsx
```diff
@@ -75,3 +87,4 @@ export async function bootstrap(App: AppComponent) {
 
 // run
 bootstrap(process.isMainFrame ? LensApp : App);
+attachChromeDebugger();
```
chenhunghan: forgot to await?
```suggestion
await attachChromeDebugger();
```

src/renderer/bootstrap.tsx
```diff
@@ -19,6 +20,17 @@ import { App } from "./components/app";
...
 import { themeStore } from "./theme.store";
 
+/**
+ * If this is a development buid, wait a second to attach
+ * Chrome Debugger to renderer process
+ * https://stackoverflow.com/questions/52844870/debugging-electron-renderer-process-with-vscode
+ */
+async function attachChromeDebugger() {
+  if (remote.process.defaultApp) {
+    await new Promise(r => setTimeout(r, 1000));
```
chenhunghan: [nit] prefer `resolve` instead of just `r` (can be confusing)
```suggestion
    await new Promise(resolve => setTimeout(resolve, 1000));
```

src/renderer/bootstrap.tsx
```diff
@@ -19,6 +20,17 @@ import { App } from "./components/app";
 import { LensApp } from "./lens-app";
 import { themeStore } from "./theme.store";
 
+/**
+ * If this is a development buid, wait a second to attach
+ * Chrome Debugger to renderer process
+ * https://stackoverflow.com/questions/52844870/debugging-electron-renderer-process-with-vscode
+ */
+async function attachChromeDebugger() {
+  if (remote.process.defaultApp) {
```
chenhunghan: Is this for checking if the app in development mode? (correct me if I am wrong) If so, there is `isDevelopment` from `./src/common/vars`, would that be be better?

Nachasic: Wasn't aware of this one, thanks!

chenhunghan (dismissed):
LGTM

.vscode/launch.json
```diff
@@ -0,0 +1,57 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
Nokel81, line 4: Do we know that the vscode JSON parser is json5 compliant?

src/renderer/bootstrap.tsx
```diff
@@ -19,6 +19,17 @@ import { App } from "./components/app";
...
 import { themeStore } from "./theme.store";
 
+/**
+ * If this is a development buid, wait a second to attach
+ * Chrome Debugger to renderer process
+ * https://stackoverflow.com/questions/52844870/debugging-electron-renderer-process-with-vscode
+ */
+async function attachChromeDebugger() {
+  if (isDevelopment) {
+    await new Promise(resolve => setTimeout(resolve, 1000));
```
Nokel81: We already have a helper function for this. (in `src/common/utils`
```suggestion
    await delay(1000);
```

.vscode/launch.json
Nachasic: Don't know for sure, this comment was generated by VSCode's own templates though

Nokel81: If it works then great!

Nokel81 (dismissed):

src/renderer/bootstrap.tsx
```diff
@@ -19,6 +19,17 @@ import { App } from "./components/app";
...
 import { themeStore } from "./theme.store";
 
+/**
+ * If this is a development buid, wait a second to attach
+ * Chrome Debugger to renderer process
+ * https://stackoverflow.com/questions/52844870/debugging-electron-renderer-process-with-vscode
+ */
+async function attachChromeDebugger() {
+  if (isDevelopment) {
+    await delay(1000);
```
jim-docker, line 30: need to import `delay`...

jim-docker: Then I see this when running `yarn dev:renderer`:
```
    ERROR in ./src/renderer/bootstrap.tsx 71:0
    Module parse failed: Cannot use keyword 'await' outside an async function (71:0)
    File was processed with these loaders:
     * ./node_modules/@pmmmwh/react-refresh-webpack-plugin/loader/index.js
     * ./node_modules/ts-loader/index.js
    You may need an additional loader to handle the result of these loaders.
    | // run
    | bootstrap(process.isMainFrame ? LensApp : App);
    > await attachChromeDebugger();
    | 
    | 
     @ multi ./node_modules/@pmmmwh/react-refresh-webpack-plugin/client/ReactRefreshEntry.js ./node_modules/@pmmmwh/react-refresh-webpack-plugin/client/ErrorOverlayEntry.js ./src/renderer/bootstrap.tsx LensDev[2]
```

Nokel81: Oh that is a different problem. That has to do with the `await attachChromeDebugger()` being called at the end of this file.

src/renderer/bootstrap.tsx
```diff
@@ -75,3 +87,4 @@ export async function bootstrap(App: AppComponent) {
 
 // run
 bootstrap(process.isMainFrame ? LensApp : App);
+await attachChromeDebugger();
```
Nokel81: This is still broken and will never work `async`/`await` doesn't work like this. I also don't understand what you are trying to do? Do you want to block the startup of the application or block the renderering?

Nokel81 (changes requested):

src/renderer/bootstrap.tsx
Nachasic: Yeah, noticed that - already fixed

src/renderer/bootstrap.tsx
Nachasic: The aim was to block startup for 1 second, so chrome debugger has time to attach properly. I moved the function to an appropriate place at the top of `bootstrap()`.

Nokel81 (approved):

jim-docker (approved):

src/renderer/bootstrap.tsx
```diff
@@ -35,6 +47,7 @@ export {
 export async function bootstrap(App: AppComponent) {
   const rootElem = document.getElementById("app");
 
+  await attachChromeDebugger();
```
jim-docker, line 50: Should we not do this for production?

Nachasic: There is a check for development build inside this function, on production builds it doesn't do anything.

PR #2297, opened 2021-03-08
Fix Lens not clearing other KUBECONFIG env vars

### LINKED ISSUE #1969: Using Cordon with multiple clusters always uses one cluster
**Describe the bug**
I have two clusters registered in Lens and it does not appear to be correctly setting up console windows for each cluster. It always seems to have them associated with the first listed cluster.

**To Reproduce**
1. Setup Lens to have two clusters registered
2. Select cluster 1
3. View a node and choose to cordon it
4. A console window opens and performs the correct action
5. Select cluster 2
6. View a node and choose to cordon it
7. A console window opens and it tries to cordon a node on cluster 1

It appears the second console window is still pointing at Cluster 1. This is hugely dangerous as it is applying a change you asked for and confirmed, but on the wrong cluster.

**Expected behavior**
For each console window that opens to be pointing at the api server of the correct cluster.

**Screenshots**
N/A

- Lens Version: 4.0.6
- OS: Windows 10
- Installation method (e.g. snap or AppImage in Linux): Windows installer

**Logs:**

**Kubeconfig:**
They appear to be well formed as all UI operations work correctly on each cluster. It is only command windows that seem to fail.

**Additional context**
Not sure if it is relevant but we do run Windows nodepools on our clusters and in this case we were acting on nodes in these Windows nodepools.

Nokel81:

fixes #1969

### DIFF
--- a/package.json
+++ b/package.json
@@ -223,7 +223,7 @@
     "mobx-react": "^6.2.2",
     "mock-fs": "^4.12.0",
     "moment": "^2.26.0",
-    "node-pty": "^0.10.0",
+    "node-pty": "^0.9.0",
     "npm": "^6.14.8",
     "openid-client": "^3.15.2",
     "p-limit": "^3.1.0",
--- a/package.json
+++ b/package.json
@@ -223,7 +223,7 @@
     "mobx-react": "^6.2.2",
     "mock-fs": "^4.12.0",
     "moment": "^2.26.0",
-    "node-pty": "^0.9.0",
+    "node-pty": "^0.10.0",
     "npm": "^6.14.8",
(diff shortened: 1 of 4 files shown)


jakolehm (dismissed):
LGTM

jim-docker (dismissed):

jim-docker (approved):

nevalla (approved):

Nokel81:
@jakolehm There is something about the VM that runs this PR's azure CI that keeps a previous node_modules around even though the `yarn.lock` should specify that `node-pty@0.10.0` shouldn't be used. Are you fine if I merge this? I removed the accidental version bump several days ago.

PR #2298, opened 2021-03-08
Conditionally render status icon for kube meta 

### LINKED ISSUE #2200: Copying fields from overlay should not include newline

**Why is this needed**:

- Desktop OS: macOS 10.15.7

Nokel81:

There should be no visual changes due to the PR.

### DIFF
--- a/src/renderer/components/kube-object-status-icon/kube-object-status-icon.tsx
+++ b/src/renderer/components/kube-object-status-icon/kube-object-status-icon.tsx
@@ -6,10 +6,6 @@ import { cssNames, formatDuration } from "../../utils";
 import { KubeObject, KubeObjectStatus, KubeObjectStatusLevel } from "../../..//extensions/renderer-api/k8s-api";
 import { kubeObjectStatusRegistry } from "../../../extensions/registries";
 
-export interface ObjectProps {
-  object: KubeObject;
-}
-
 function statusClassName(level: number): string {
   switch (level) {
     case KubeObjectStatusLevel.INFO:
@@ -63,7 +59,11 @@ function splitByLevel(src: KubeObjectStatus[]): SplitStatusesByLevel {
   return { maxLevel, criticals, warnings, infos };
 }
 
-export class KubeObjectStatusIcon extends React.Component<ObjectProps> {
+interface Props {
+  object: KubeObject;
+}
+
+export class KubeObjectStatusIcon extends React.Component<Props> {
   renderStatuses(statuses: KubeObjectStatus[], level: number) {
     const filteredStatuses = statuses.filter((item) => item.level == level);
 
--- a/src/renderer/components/kube-object-status-icon/kube-object-status-icon.tsx
+++ b/src/renderer/components/kube-object-status-icon/kube-object-status-icon.tsx
@@ -6,18 +6,7 @@ import { cssNames, formatDuration } from "../../utils";
 import { KubeObject, KubeObjectStatus, KubeObjectStatusLevel } from "../../..//extensions/renderer-api/k8s-api";
 import { kubeObjectStatusRegistry } from "../../../extensions/registries";
 
-interface BaseProps {
-  statuses?: KubeObjectStatus[];
-  object?: KubeObject;
-}
-
-interface StatusProps extends BaseProps {
-  statuses: KubeObjectStatus[];
-  object?: undefined;
-}
-
-interface ObjectProps extends BaseProps {
-  statuses?: undefined;
+export interface ObjectProps {
   object: KubeObject;
 }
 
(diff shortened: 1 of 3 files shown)


src/renderer/components/kube-object/kube-object-meta.tsx
```diff
@@ -22,10 +23,21 @@ export class KubeObjectMeta extends React.Component<KubeObjectMetaProps> {
...
+  renderObjectName() {
+    const { object } = this.props;
+    const statuses = kubeObjectStatusRegistry.getItemsForObject(object);
+
+    if (statuses.length === 0) {
+      return object.getName();
+    }
+
+    return <>{object.getName()} <KubeObjectStatusIcon key="icon" statuses={statuses} /></>;
+  }
```
aleksfront: Better split it into 2 functions - `renderObjectName()` and `renderStatusIcon()`.

src/renderer/components/kube-object/kube-object-meta.tsx
```diff
@@ -22,10 +23,21 @@ export class KubeObjectMeta extends React.Component<KubeObjectMetaProps> {
...
 
+  renderObjectName() {
+    const { object } = this.props;
+    const statuses = kubeObjectStatusRegistry.getItemsForObject(object);
+
+    if (statuses.length === 0) {
+      return object.getName();
+    }
+
+    return <>{object.getName()} <KubeObjectStatusIcon key="icon" statuses={statuses} /></>;
```
aleksfront: To keep single source of truth rule, we need to allow to pass only `statuses` prop or `object`, not both. I see that `statuses` was introduces for this specific case in `kube-object-meta.tsx`?

aleksfront: This will allow also to remove prop type dualism here 
```
export class KubeObjectStatusIcon extends React.Component<StatusProps | ObjectProps> {
```

aleksfront (changes requested):
I've checked #2200 issue and it's still relevant. Some lines like `Namespace` can be copied without trailing line, but most of the items (e.g. `Name`) are still contains new line if you copy them as described in #2200.

Nokel81:
@aleksfront Do you mean after this PR or before this PR's changes?

aleksfront:
> @aleksfront Do you mean after this PR or before this PR's changes?

Behavior same before and after the PR changes

Nokel81:
Can I have a GIF? I have tried the triple-click as described in the linked issue and without a status icon there isn't the final space.

Nokel81:
Ah now I see, slightly different issue.

Nokel81:
This is very weird, there isn't a new line in the html. And if I try and select it manually I cannot.

https://user-images.githubusercontent.com/8225332/112852530-05a45e00-907a-11eb-97e7-dc95e62974d4.mov

Nokel81:
I have found that changing the `span` to a `div` does nothing and changing it to `p` adds another new line.

Will continue to investigate.

Nokel81:
So it seems like this is how browsers handle triple clicking spans. Also the following should show that chrome adds new lines even if there are none, so I don't quite know how to solve this. Maybe this is what a table should be used for.

Nokel81:
I tried using a table but that just changed it from being a newline to being a tab character. Do you know how this might be fixed @aleksfront ?

aleksfront:
> I tried using a table but that just changed it from being a newline to being a tab character. Do you know how this might be fixed @aleksfront ?

As it turns out, there's no quick and easy way to do this. The triple-click selection behaviour is os-specific. However, my investigation shows that we have at least 2 workarounds, but both of them should be implemented with care as with high possibility of side effects.

1. Set `user-select` property to `all`. It will select all text on *first* click. I would vote for this method but it behaves really strange in chrome causing a lot of selection blinking. https://developer.mozilla.org/en-US/docs/Web/CSS/user-select
2. Explicitly cut last selected character on **clipboard copy** event. https://developer.mozilla.org/en-US/docs/Web/API/Element/copy_event

My suggestion for this PR is to remove `fixes #2200` part from it and point to status icon changes only.

Nokel81:
@aleksfront sounds good. I will remove the issue resolution. Thanks for doing a deeper dive into this.

src/renderer/components/kube-object-status-icon/kube-object-status-icon.tsx
```diff
@@ -2,105 +2,106 @@ import "./kube-object-status-icon.scss";
...
-import { KubeObject } from "../../api/kube-object";
 import { cssNames, formatDuration } from "../../utils";
-import { KubeObjectStatusRegistration, kubeObjectStatusRegistry } from "../../../extensions/registries/kube-object-status-registry";
-import { KubeObjectStatus, KubeObjectStatusLevel } from "../../..//extensions/renderer-api/k8s-api";
-import { computed } from "mobx";
+import { KubeObject, KubeObjectStatus, KubeObjectStatusLevel } from "../../..//extensions/renderer-api/k8s-api";
+import { kubeObjectStatusRegistry } from "../../../extensions/registries";
 
-interface Props {
+export interface ObjectProps {
```
aleksfront: Why renamed to `ObjectProps` and exported? I see that the only component is using it placed inside same file.
About new name: maybe it's just me, but it started to direct my mind to native js objects at first. Maybe no need to rename?

aleksfront (approved):
LGTM with 1 minor question/comment.

src/renderer/components/kube-object-status-icon/kube-object-status-icon.tsx
Nokel81: I really don't remember why I did this. Will revert.

jakolehm (approved):

Issue #1983, opened 2021-01-19
Improve preferences layout


jakolehm:
- left menu should acts as shortcuts to different sections
- all extension registered preferences should be under "Extensions"
- new back/close button

Mockup:
![image (1)](https://user-images.githubusercontent.com/1446224/105025079-82510580-5a55-11eb-92c4-cba13a4a8217.png)

aleksfront:
@Nokel81 This can be added into 4.2.0 milestone when https://github.com/lensapp/lens/pull/2304 will be reviewed.

Nokel81:
Finished without changing the close/back button.

PR #2304, opened 2021-03-09
Navigation menu in Preferences

aleksfront:
* Reordering preferences contents, separating them by `<section/>` tags
* Visual fixes, more bright heading colors
* Adding `<ScrollSpy/>` & `<TreeView/>` components which in conjuction provide navigation shortcuts to available sections.

**ScrollSpy**
Implementation of classic ScrollSpy component but with [IntersectionObserver API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) under the hood. Used to build navigation tree when specific sections become visible on screen or when DOM changes. It parses all available `<section/>` elements to create a tree, so couple of requirements are available for content layout:
* all contents should be carefully placed and separated by sections.
* every section must contain `id` tag. It can be omitted if section intends to be hidden in menu.

**TreeView**
`<ScrollSpy/>` generates hierarchy tree regardless of which component will be used to render navigation. I've chose `Material UI TreeView` as best among competitors. It handles all needed cases but it's still in alpha version though. `TreeView` is responsible to render actual menu component and highlight active item.

https://user-images.githubusercontent.com/9607060/110469754-156bfa80-80eb-11eb-98d8-c10d762d31d0.mp4

light theme nav

Part of #1983


package.json
```diff
@@ -305,6 +306,7 @@
     "concurrently": "^5.2.0",
     "css-element-queries": "^1.2.3",
     "css-loader": "^3.5.3",
+    "deepdash": "^5.3.5",
```
aleksfront, line 321: This is used to recursive search within a complex object. Used inside `<TreeView/>` to look for selected item.

src/renderer/components/+preferences/helm-charts.tsx
```diff
@@ -0,0 +1,139 @@
+import "./helm-charts.scss";
```
aleksfront, line 1: Helm charts moved to own file without changes.

src/renderer/components/scroll-spy/scroll-spy.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { NavigationTree } from "../tree-view";
+
+interface Props extends React.DOMAttributes<HTMLElement> {
+  render: (data: NavigationTree[]) => JSX.Element
+}
+
+export function ScrollSpy(props: Props) {
+  const parent = useRef<HTMLDivElement>();
+  const sections = useRef<NodeListOf<HTMLElement>>();
+  const [tree, setTree] = useState<NavigationTree[]>([]);
```
Nokel81, line 15: How does this tie into `mobx`? (asking mostly for information)

src/renderer/components/tree-view/tree-view.tsx
```diff
@@ -0,0 +1,102 @@
...
+import React, { useEffect, useRef } from "react";
+import { Icon } from "../icon";
+import TreeView from "@material-ui/lab/TreeView";
+import TreeItem from "@material-ui/lab/TreeItem";
+import { cssNames } from "../../utils";
+
+import _ from "lodash";
+import getDeepDash from "deepdash";
+
+const deepDash = getDeepDash(_);
```
Nokel81, line 12: looking at the docs this isn't needed and we should be able to just do the following:
```suggestion
import { findDeep } from "deepdash";
```

src/renderer/components/tree-view/tree-view.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+interface Props {
+  data: NavigationTree[]
+}
+
+const scrollToItem = (id: string) => {
+  const element = document.getElementById(id);
+
+  element?.scrollIntoView();
+};
```
Nokel81: This should be a function
```suggestion
function scrollToItem(id: string): void {
  document
  	.getElementById(id)
    ?.scrollIntoView();
}
```

src/renderer/components/tree-view/tree-view.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+const getSelectedNode = (data: NavigationTree[]) => {
+  return deepDash.findDeep(data, (value, key) => key === "selected" && value === true)?.parent;
+};
```
Nokel81: And this
```suggestion
function getSelectedNode(data: NavigationTree[]): NavigationTree | undefined {
  return findDeep(data, (value, key) => key === "selected" && value === true)?.parent;
}
```

src/renderer/components/tree-view/tree-view.tsx
```diff
@@ -0,0 +1,102 @@
...
+    setExpanded(nodeIds);
+  };
+
+  const expandTopLevelNodes = () => {
+    setExpanded(data.map(node => node.id));
+  };
+
+  const expandParentNode = () => {
+    const node = getSelectedNode(data) as any as NavigationTree;
+    const id = node?.parentId;
```
Nokel81, line 48: The above allows this to be:
```suggestion
    const id = getSelectedNode(data)?.parentId;
```

src/renderer/hooks/useMutationObserver.ts
```diff
@@ -0,0 +1,27 @@
+import { useEffect } from "react";
+
+const config: MutationObserverInit = {
+  subtree: true,
+  childList: true,
+  attributes: false,
+  characterData: false
+};
+
+export function useMutationObserver(
```
Nokel81, line 10: is this instead of `mobx`?

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -33,178 +27,101 @@ export class Preferences extends React.Component {
...
+              return (
+                <section key={index} id={title}>
+                  <h2>{title}</h2>
+                  <Input/>
+                  <small className="hint">
+                    <Hint/>
+                  </small>
+                </section>
+              );
+            })}
```
Nokel81, line 146: This assumes that every extension only has one preference. I think it would be better if each extension had its own tree.

src/renderer/components/scroll-spy/scroll-spy.tsx
aleksfront: Mobx observable state not really useful in functional components. With native hooks like `useState` we do all needed work without any additional decorators. So answering the question - it doesn't tie into `mobx` at all.

https://mobx.js.org/react-integration.html#you-might-not-need-locally-observable-state

src/renderer/components/tree-view/tree-view.tsx
aleksfront: I've tryied this, but we'll suddenly hit typescript complaints.
finddeep error

There's possible to install `deepdash-es` dependency, but it takes also whole new `lodash-es` library as a dependency. And also, there're some test errors occured with `deepdash-es`. So I ended up with current option considering it as necessary evil.

src/renderer/hooks/useMutationObserver.ts
aleksfront: Not really sure if "instead" is a proper word here. This is just new react hook which can be used in any functional component. It doesn't relate to mobx at all.

src/renderer/components/+preferences/preferences.tsx
aleksfront: I was looking at https://user-images.githubusercontent.com/1446224/105025079-82510580-5a55-11eb-92c4-cba13a4a8217.png

Every extension has it's own `section` element along with a link in navigation bar.

src/renderer/components/tree-view/tree-view.tsx
aleksfront: Reverted `getSelectedNode`  into a function but left types untouched since `.parentId` also needs same kind of `as any...` transformations.
```
Property 'parentId' does not exist on type 'unknown'.ts(2339)
```

jim-docker:
Haven't looked at the code yet but played with it. Looks very nice. I noticed some little glitches(?) When at the top of the page, if you click on "Extensions", "Extensions" is highlighted but the "Kubernetes" drawer also opens and "Kubectl binary" is also highlighted. I think I understand why it happens (you can't scroll "Extensions" to the top unless you show some blank space beneath it) but it feels a bit strange. 


https://user-images.githubusercontent.com/40840436/110714402-4cb4d700-81d1-11eb-88ad-d8beafe5bd8c.mov

aleksfront:
@jim-docker Fixed along with additional bugs.


https://user-images.githubusercontent.com/9607060/110753024-102bbe80-8257-11eb-81ee-3918eed00eb6.mp4

jakolehm (dismissed):
LGTM

src/renderer/components/tree-view/tree-view.tsx
Nokel81: oh okay weird. Thanks for the explanation.

src/renderer/components/tree-view/tree-view.tsx
Nokel81: I would say that `getSelectedNode`'s return value being `IDeepEntry` (though if you look at the typings even that is wrong) is an implementation leak. The cast should be within `getSelectedNode`.

src/renderer/components/+preferences/preferences.tsx
Nokel81: How so? Currently none of the built in extensions have more than one but we should support it. This does mean that `appPreferenceRegistry` needs to be updated to keep track of that. Or even better you could do something like I propose in #2245 but that is probably out of scope of this PR.

src/renderer/hooks/useMutationObserver.ts
ixrock: Yep, nothing related to `mobx` here. It's just a`react` hook with https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver API on board.

src/renderer/hooks/useMutationObserver.ts
```diff
@@ -0,0 +1,27 @@
+import { useEffect } from "react";
```
ixrock, line 1: // Observe changes in DOM
// API docs: https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver

src/renderer/components/scroll-spy/scroll-spy.tsx
```diff
@@ -0,0 +1,100 @@
...
+  useMutationObserver(getSectionsParentElement(), updateNavigation);
+
+  return (
+    <div className="ScrollSpy" ref={parent}>
+      {props.render(tree)}
+    </div>
+  );
+}
+
+ScrollSpy.defaultProps = {
```
ixrock: This is one of the reasons why I don't like functional-style react components :D 
Default props should be on the top of component for better readability, imho..

ixrock (dismissed):
Does `content-navigation` have it's own scrollbar? Otherwise I would not block page content area' scrollbar while mouse hovering it. Wdyt? 

Screenshot 2021-03-11 at 18 42 51

src/renderer/components/scroll-spy/__tests__/scroll-spy.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+            <div>description</div>
+          </section>
+        </section>
+      </div>
+    )}/>);
+
+    await waitFor(() => {
+      expect(queryByTitle("Application")).toBeInTheDocument();
+      expect(queryByTitle("Appearance")).toBeInTheDocument();
+      expect(queryByTitle("Theme")).toBeInTheDocument();
```
ixrock, line 100: Where `queryByTitle` does it search? Is it intended to test content inside `<section id="application">` or inside `<nav>`? Otherwise it looks like it just checks static content from `render` above.

src/renderer/components/scroll-spy/__tests__/scroll-spy.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+    // Restore writing to stderr.
+    console.error = err;
+  });
+});
+
+
+describe("<TreeView/> navigation inside <ScrollSpy/>", () => {
+  it("contains links to all sections", async () => {
+    const { queryByTitle } = render(<ScrollSpy render={navigation => (
```
ixrock:
```suggestion
    const { queryByTitle } = render(<ScrollSpy render={dataTree => (
```

src/renderer/components/scroll-spy/scroll-spy.tsx
Nokel81: Well given that a functional component is hoisted (because it is a function) it is possible to place this above. I agree that above it better for visibility and readability.

src/renderer/components/scroll-spy/__tests__/scroll-spy.test.tsx
aleksfront: It checks everything from `render` method. But only navigation links have their own `title` attributes. So finding something with `title=Application` means that there's at least one navigation link inside `nav`.

src/renderer/components/scroll-spy/scroll-spy.tsx
aleksfront: Moved to the top for better readability.

src/renderer/components/tree-view/tree-view.tsx
aleksfront: Yes, unfortunately, deepDash has broken typings. I was using [type assertions[(https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions) here and unable to find an option to set proper types for return value of `getSelectedNode`

aleksfront:
> Does `content-navigation` have it's own scrollbar? Otherwise I would not block page content area' scrollbar while mouse hovering it. Wdyt?
> 
> Screenshot 2021-03-11 at 18 42 51

Good catch! I've added vertical scrollbar
navigation scrollbar

aleksfront:
@ixrock @jakolehm @Nokel81 Fixed lastly founded issues. Please re-review.

src/renderer/components/scroll-spy/scroll-spy.tsx
aleksfront: *Update*. Converted `ScrollSpy` into observable component and now setting defaultProps through default arguments in a manner
```
export const ScrollSpy = observer(({ render, rootMargin = "0px 0px -85%" }: Props) => {
```

src/renderer/components/layout/page-layout.scss
```diff
@@ -54,13 +59,19 @@
...
       max-width: var(--max-width);
-      padding: calc(var(--spacing) * 2);
+      padding: var(--wrapper-padding);
       margin: 0 auto;
 
       .content-navigation {
         > ul {
           position: fixed;
           width: var(--nav-width);
+          height: calc(100vh - var(--header-height) - var(--wrapper-padding) - var(--bottom-bar-height));
```
ixrock: Why do we need so much fixed height calculations?
Couldn't it just work with some proper `display: grid` + `overflow: auto` + column defined with something like `max-content/1fr`?

ixrock (changes requested):
Doesn't work as expected on small screen:

https://user-images.githubusercontent.com/6377066/110936666-ebf4de00-8339-11eb-83de-ee093db74abf.mov

Also, would be nice to highlight relevant preferences section after clicking navigation item (for quick search on the screen):
image

aleksfront:
1. Fixed small window height behavior.
small window navigation

2. Header highlight is a good enhancement but should be moved to issues first (out of scope).
3. I've got some options how to refactor `<PageLayout/>`, but suggest to do this in next PRs.
@ixrock ^

Nokel81:
Is this really what we want to display for extensions that have have multiple preferences?

https://user-images.githubusercontent.com/8225332/110950248-9c9ab780-8311-11eb-8e66-2de99e5db92b.mov

jim-docker:
> Is this really what we want to display for extensions that have have multiple preferences?
> 
I think it is acceptable given that we currently don't visually delineate the preferences by extension. That would be done in a separate PR, then this tree-view could be updated correspondingly.

aleksfront:
@ixrock Regadring to your last founded issues.
1. Fixed intersection observer buggy behavior - not it has clear `rootMargin` prop and work on any viewport height.
2. Simplifyed `<PageView/>` layout, removed excessive height calculations and `position: fixed` for navigation bar.
3. Your highlight active section content proposal will be moved to issue tracker.

preferences page

src/renderer/components/layout/page-layout.tsx
```diff
@@ -80,11 +80,8 @@ export class PageLayout extends React.Component<PageLayoutProps> {
             <RecursiveTreeView data={navigation}/>
           </nav>
         )}
-        <div
-          className={cssNames("content", contentClass)}
-          id="ScrollSpyRoot"
-        >
-          <div className={cssNames(contentGaps && "flex column gaps")}>
+        <div className="content-wrapper" id="ScrollSpyRoot">
```
ixrock, line 83: Why do we have `id="ScrollSpyRoot"` instead of `className`? Is it intended to be single rendered component on the screen?

aleksfront: We have to point ScrollSpy component which element will be used as `root`. Root is used as a target for intersection and mutation observers. So yes, we need only 1 such element, and it should be rendered once.

src/renderer/components/tree-view/tree-view.tsx
Nokel81: I meant having the `as any as NavigationTree` within your helper function but that is of low priority.

Nokel81 (approved):

ixrock (approved):
Tried in branch, navigation seems to work as expected. 
But I would still add css style regarding to clicked item mentioned in https://github.com/lensapp/lens/pull/2304#pullrequestreview-610707766

Nokel81:
@ixrock I think that should be done in a follow up PR so that this can be merged and will definitely get into the RC.

PR #2316, opened 2021-03-11
Add terminal shortcut for mac os

### LINKED ISSUE #2214: cmd + k does not clear the terminal
**Describe the bug**
When I press cmd + k in the terminal window, nothing happens. The default behaviour on Mac to clear a terminal is `cmd + k`. This works in the default terminal, in iTerm etc...

**To Reproduce**
1. Open the terminal
2. kubectl get all -A
3. cmd + k
4. See it do nothing...

**Expected behavior**
I expect the terminal to be cleared.

- Lens Version: 4.1.2
- OS: [e.g. OSX] OSX 10.15.7 (19H524)
- Installation method (e.g. snap or AppImage in Linux): brew install lens

vshakirova:
Wdyt guys? Do we need smth similar for linux? ctrl+k for example 

Fixes: https://github.com/lensapp/lens/issues/2214

### DIFF
--- a/src/renderer/components/dock/terminal.ts
+++ b/src/renderer/components/dock/terminal.ts
@@ -6,6 +6,7 @@ import { dockStore, TabId } from "./dock.store";
 import { TerminalApi } from "../../api/terminal-api";
 import { themeStore } from "../../theme.store";
 import { autobind } from "../../utils";
+import { isMac } from "../../../common/vars";
 
 export class Terminal {
   static spawningPool: HTMLElement;
@@ -182,7 +183,7 @@ export class Terminal {
   };
 
   keyHandler = (evt: KeyboardEvent): boolean => {
-    const { code, ctrlKey, type } = evt;
+    const { code, ctrlKey, type, metaKey } = evt;
 
     // Handle custom hotkey bindings
     if (ctrlKey) {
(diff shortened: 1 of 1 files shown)


src/renderer/components/dock/terminal.ts
```diff
@@ -199,6 +200,15 @@ export class Terminal {
       }
     }
 
+    //Ctrl+K: clear the entire buffer, making the prompt line the new first line on mac os
+    if (isMac && metaKey) {
```
ixrock, line 204: Noticed that `Ctrl+L` clears output on Mac. Do we need one more hotkey?

vshakirova: Yes, it works. But according to the comment, it would be nice to have the ability to "clear the scrollback buffer as well as the screen".
https://github.com/lensapp/lens/issues/2214#issuecomment-787975216

aleksfront: @ixrock See https://github.com/lensapp/lens/issues/2214#issuecomment-787969824

aleksfront (approved):
LGTM

aleksfront:
@Nokel81 Can we include this into 4.2?

ixrock (approved):

Nokel81:
Yes this can be added to 4.2.0. 

However, I would like to note that from a precedent perspective I don't think we should try and special case more keyboard shortcuts. A general passthrough mechanism would be better for the future.

Issue #2315, opened 2021-03-10
Cluster settings page sometimes requires multiple clicks to close


jim-docker:
Cluster settings page sometimes requires multiple clicks to close. It hasn't been reproduced in master but occurs for https://github.com/lensapp/lens/pull/2065 and https://github.com/lensapp/lens/pull/2047. May be related to using a pop-up `Menu` element for accessing the cluster settings page. The page may be getting added multiple times to the navigation history according to @Nachasic 

See the issue here:
https://github.com/lensapp/lens/pull/2065#pullrequestreview-592379673

Nokel81:
This has to do with some odd behaviour around `navigation`. I don't know exactly what is causing it but from adding some inspection it is easy to observe that this behaviour only occurs when more than one cluster iframe is active.

Namely, adding the following code to `src/renderer/navigation/history.ts` can help observe what is going on.

```ts
navigation.listen((location, action) => {
  const clusterId = getHostedClusterId();
  const frame = clusterId ?? "root";

  logger.debug(`[NAVIGATION]: ${action}`, { location, frame });
});
```

jim-docker:
fixed by #2330

PR #2321, opened 2021-03-11
add debug prints for actions against the navigation object

Nokel81:

Will help debug things like #2315 in the future.

### DIFF
--- a/src/renderer/navigation/history.ts
+++ b/src/renderer/navigation/history.ts
@@ -1,6 +1,15 @@
 import { ipcRenderer } from "electron";
 import { createBrowserHistory, createMemoryHistory } from "history";
 import { createObservableHistory } from "mobx-observable-history";
+import { getHostedClusterId } from "../../common/cluster-store";
+import logger from "../../main/logger";
 
 export const history = ipcRenderer ? createBrowserHistory() : createMemoryHistory();
 export const navigation = createObservableHistory(history);
+
+navigation.listen((location, action) => {
+  const clusterId = getHostedClusterId();
+  const frame = clusterId ?? "root";
+
+  logger.debug(`[NAVIGATION]: ${action}`, { location, frame });
+});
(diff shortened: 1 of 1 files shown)


src/renderer/navigation/history.ts
```diff
@@ -1,6 +1,15 @@
 import { ipcRenderer } from "electron";
 import { createBrowserHistory, createMemoryHistory } from "history";
 import { createObservableHistory } from "mobx-observable-history";
+import { getHostedClusterId } from "../../common/cluster-store";
+import logger from "../../main/logger";
```
jim-docker, line 4: unrelated, but shouldn't `logger` be in `src/common`?

jim-docker (dismissed):

src/renderer/navigation/history.ts
Nokel81: Yes it probably should be. I too have thought about that. Lets see about doing that for 4.3

src/renderer/navigation/history.ts
```diff
@@ -1,6 +1,15 @@
...
 
 export const history = ipcRenderer ? createBrowserHistory() : createMemoryHistory();
 export const navigation = createObservableHistory(history);
+
+navigation.listen((location, action) => {
+  const clusterId = getHostedClusterId();
+  const frame = clusterId ?? "root";
+
+  logger.debug(`[NAVIGATION]: ${action}`, { location, frame });
+});
```
ixrock: Do we really need to know `clusterId` in debug info?
Otherwise I would rather to use `location.href` which is more informative (and contains cluster-id) and no need in importing `cluster-store.ts`

```suggestion
navigation.listen((location, action) => {
  const isClusterView = !process.isMainFrame;

  logger.debug(`[NAVIGATION]: ${action}`, { domain: location.href, location, isClusterView });
});
```

Nokel81: I found parsing that object harder to read but sure we can have this instead.

jim-docker (approved):

ixrock (approved):

PR #644, opened 2020-08-05
Added ClusterRole 

just1689:
Greetings

Love the Lens project. Its fantastic. Thank you for creating the project.

Regarding the pull request:

Added ClusterRole that allows for read only access. This ALSO supports Lens metrics. The default **read only** ClusterRoles  do not always allow metrics to show in Lens because they do not explicitly list Kubernetes sub-resources.

See README.md in the commit for more info.

Cheers
Justin

PR #2327, opened 2021-03-12
YAML Templates in Create Resource dock tab

### LINKED ISSUE #995: YAML Templates in Create Resource dock tab
Create resource controls area should contain `Template` dropdown selector with list of predefined workloads. Clicking any of them would paste yaml template into editor.

template dropdown

pashevskii:
The commit adds the ability to paste the template into Create Resource editor from the prepared list.
Moreover, a user can add his own templates by adding yaml files into <userHome>/.k8slens/tempates/{subfolder: optional}

template

Fixes #995


matti:
one (1) beer deserved, :beer:

src/renderer/components/dock/create-resource.tsx
```diff
@@ -21,6 +25,39 @@ interface Props {
 @observer
 export class CreateResource extends React.Component<Props> {
   @observable error = "";
+  templates:GroupSelectOption[] = [];
+
+  componentDidMount(){
+    let rootPaths: any[]= [];
```
ixrock: Maybe `string[]`?

src/renderer/components/dock/create-resource.tsx
```diff
@@ -21,6 +25,39 @@ interface Props {
...
+
+    for (const store of this.templatesStores) {
+      rootPaths = rootPaths.concat(fs.readdirSync(store).map(value => {return {root: store, value};}));
+    }
+    const ungroupedFiles: any[] = [];
```
ixrock: Here too `string[]`?

src/renderer/components/dock/create-resource.tsx
```diff
@@ -21,6 +25,39 @@ interface Props {
...
+        }
+        else {
+          ungroupedFiles.push({value: absRootPath, label: path.parse(absRootPath).name});
+        }
+      }
+    });
+
+    if (ungroupedFiles.length > 0) {
+      this.templates.push({label: "Ungrouped", options: ungroupedFiles});
+    }
```
ixrock: Would be more readable if we move all these logic to 1-2 methods with clear intention. Maybe `loadTemplates()`?

src/renderer/components/dock/create-resource.tsx
```diff
@@ -30,11 +67,26 @@ export class CreateResource extends React.Component<Props> {
     return createResourceStore.getData(this.tabId);
   }
 
+  // templatesStores - provides list of folders with templates in priority order (first overrides others)
+  get templatesStores(): string[] {
```
ixrock: Probably should be `@computed` field.

src/renderer/components/dock/create-resource.tsx
```diff
@@ -30,11 +67,26 @@ export class CreateResource extends React.Component<Props> {
...
+    const userStore = path.join((app || remote.app).getPath("userData"), "templates");
```
ixrock:
```suggestion
    const templatesBaseDir = path.join((app || remote.app).getPath("userData"), "templates");
```

Nokel81: I think that using `.k8slens/templates` would be better

src/renderer/components/dock/create-resource.tsx
```diff
@@ -21,6 +25,39 @@ interface Props {
...
 export class CreateResource extends React.Component<Props> {
   @observable error = "";
+  templates:GroupSelectOption[] = [];
+
+  componentDidMount(){
+    let rootPaths: any[]= [];
+
+    for (const store of this.templatesStores) {
+      rootPaths = rootPaths.concat(fs.readdirSync(store).map(value => {return {root: store, value};}));
+    }
```
Nokel81:
```suggestion
    const rootPaths = this.templatesStores
      .flatMap(store => (
        fs.readdirSync(store)
          .map(value => ({ root: store, value })
      ));
```

src/renderer/components/dock/create-resource.tsx
```diff
@@ -21,6 +25,39 @@ interface Props {
...
+    const ungroupedFiles: any[] = [];
+
+    rootPaths.forEach((rootPath, index) => {
+      const absRootPath = path.join(rootPath.root, rootPath.value);
+
+      const isOverrided :boolean = rootPaths.map(v => v.value).indexOf(rootPath.value) < index;
```
Nokel81: I don't think we should override the defaults. How about just adding them to the end?

src/renderer/components/dock/create-resource.tsx
```diff
@@ -30,11 +67,26 @@ export class CreateResource extends React.Component<Props> {
...
+    return stores;
+  }
+
   onChange = (value: string, error?: string) => {
     createResourceStore.setData(this.tabId, value);
     this.error = error;
   };
 
+  onSelectTemplate = (item: SelectOption) => {
+    createResourceStore.setData(this.tabId, fs.readFileSync(item.value,"utf8"));
```
Nokel81: I would prefer if this was `async` so as to not block the UI.

src/renderer/components/dock/create-resource.tsx
Nokel81: Furthermore, even though this is in `componentDidMount` I would prefer if it was async.

src/renderer/components/dock/create-resource.tsx
```diff
@@ -30,11 +67,26 @@ export class CreateResource extends React.Component<Props> {
     return createResourceStore.getData(this.tabId);
   }
 
+  // templatesStores - provides list of folders with templates in priority order (first overrides others)
+  get templatesStores(): string[] {
+    const userStore = path.join((app || remote.app).getPath("userData"), "templates");
+    const stores = [];
+
+    if (fs.pathExistsSync(userStore)) stores.push(userStore);
```
Nokel81: This could technically break in subtle ways. I would prefer just recreate it if not present since we handle the "no entries" case explicitly anyway.

src/renderer/components/dock/create-resource.tsx
```diff
@@ -67,6 +119,23 @@ export class CreateResource extends React.Component<Props> {
...
+      <div className="flex gaps align-center">
+        <span>Template</span>
+        <Select
+          autoConvertOptions = {false}
+          className="TemplateSelect"
+          options={this.templates}
+          menuPlacement="top"
+          themeName="outlined"
+          onChange={v => this.onSelectTemplate(v)}
+        />
```
Nokel81:
```suggestion
        <Select
          autoConvertOptions = {false}
          className="TemplateSelect"
          placeholder="Select Template ..."
          options={this.templates}
          menuPlacement="top"
          themeName="outlined"
          onChange={v => this.onSelectTemplate(v)}
        />
```

Nokel81 (commented):
This looks like a very good addition. Can you add the templates from #644 too? Since that PR's author is not responding anymore.

src/renderer/components/dock/create-resource.tsx
pashevskii: No, SelectOption :)

src/renderer/components/dock/create-resource.tsx
pashevskii: No, I use {root: string, value: value} interface to handle store (built-in of user paths)  of current folder/file. 
I'll fix it.

src/renderer/components/dock/create-resource.tsx
pashevskii: How about adding a badge to templates created by the user?
image
It seems I should work on alignment, it is just a prototype

src/renderer/components/dock/create-resource.tsx
pashevskii: I just took from [here](https://github.com/lensapp/lens/blob/a9a5766920dc055bdd8e27034fa427728ab1da49/src/renderer/components/dock/install-chart.tsx#L152) 
Should I fix selects here?

src/renderer/components/dock/create-resource.tsx
Nokel81: Could you have both then? `ConfigMap user` and `ConfigMap Lens`?

pashevskii: So built-in templates will be not marked, only created by the user, also the group can include two templates with the same name but are from different sources (lens/user), no overrides at all, right?

Nokel81: Oh I meant the built in ones to be marked but not marked is better, I agree.

If you want to be the user provided ones with a badge instead of at the end, sure and sort them might be good idea.

pashevskii: yes, I think we should highlight only user templates. There is no sense to give a badge to all items because it will make identifying them more difficult. 
About sorting: 
I'm going to place groups on top which was created by the user (and not exist in built-in group list)
In other cases just sorting by the label, what do you think about it?

Nokel81: That sounds good.

src/renderer/components/dock/create-resource.tsx
```diff
@@ -30,11 +84,23 @@ export class CreateResource extends React.Component<Props> {
     return createResourceStore.getData(this.tabId);
   }
 
+  get lensStore():string {
```
aleksfront: Why we're calling paths as stores? Maybe better to call this `lensTemplatesFolder`.

src/renderer/components/dock/create-resource.tsx
```diff
@@ -30,11 +84,23 @@ export class CreateResource extends React.Component<Props> {
...
+    return path.resolve(__static, "../templates/create-resource")
+  }
+
+  get userStore():string {
```
aleksfront: `userStore` -> `userTemplatesFolder()` or `userTemplatesPath`.

src/renderer/components/dock/create-resource.tsx
```diff
@@ -21,6 +26,55 @@ interface Props {
 @observer
 export class CreateResource extends React.Component<Props> {
   @observable error = "";
+  @observable templates:GroupSelectOption[] = [];
+
+  componentDidMount(){
```
aleksfront: I believe all paths/templates manipulations should be done outside of this React component. Basically, all templates should be derived from store. `src/renderer/components/dock/create-resource.store.ts` is a good place to move all logic here.

aleksfront: Also, please spread this code to smaller reusable methods. For example:
* parsing lens and user templates may be separated.
* there is code repetition with
```
fs.readdirSync(this.lensStore).forEach(group => {
```
```
fs.readdirSync(this.userStore).forEach((group) => {
```
This can be extracted into own method.
* we can make code smaller by keeping `React-Select` options preparation and `renderWidthBadge` inside component.

src/renderer/components/dock/create-resource.scss
```diff
@@ -1,2 +1,15 @@
...
+    &.TemplateSelect {
+      min-width: 190px;
+    }
+  }
+}
+
+.select-template-badge {
+  float: right;
+  vertical-align: middle;
+}
```
aleksfront: Please use flexbox for badge positioning. Currently, there is a rendering bug
floated badge

You can reference to `NamespaceSelectFilter` to look for an example (`Badge` can be placed instead of tick icon). `src/renderer/components/+namespaces/namespace-select-filter.tsx`

aleksfront (changes requested):
Great feature, but there's few refactorings to be made.

src/renderer/components/dock/create-resource.store.ts
```diff
@@ -1,14 +1,95 @@
...
+  }
+
+  get userTemplatesFolder():string {
+    return path.join(os.homedir(), ".k8slens", "templates");
+  }
+
+  async getLensTemplates() {
+    const templates:BadgedGroupSelectOption[] = [];
+
+    fs.readdirSync(this.lensTemplatesFolder).forEach((group) => {
```
Nokel81: please don't mix sync and async function. `fse` or `fs-extra` has this function as async.

Nokel81: I think it would be better if we used something like https://www.npmjs.com/package/walk

src/renderer/components/dock/create-resource.store.ts
```diff
@@ -1,14 +1,95 @@
...
+
+    return templates;
+  }
+
+  async addUserTemplates( templates:BadgedGroupSelectOption[]) {
+    const userTemplates:BadgedGroupSelectOption[] = [];
+    const ungroupedTemplates:BadgedSelectOption[] = [];
+
+    fs.ensureDirSync(this.userTemplatesFolder);
+    fs.readdirSync(this.userTemplatesFolder).forEach((group) => {
```
Nokel81: no sync calls in async functions

src/renderer/components/dock/create-resource.store.ts
```diff
@@ -1,14 +1,95 @@
...
+
+    return templates;
+  }
+
+  orderTemplatesGroups(templates:BadgedGroupSelectOption[]): BadgedGroupSelectOption[] {
+    return templates.map( group => {
+      group.options = group.options.sort((a, b) => a.label.toString() > b.label.toString() ? 1 : a.label.toString() < b.label.toString() ? -1 : 0);
+
+      return group;
+    }).sort((a, b) => a.label.toString() > b.label.toString() ? 1 : a.label.toString() < b.label.toString() ? -1 : 0);
```
Nokel81: Please extra this sorting function so that we don't repeat it.

src/renderer/components/dock/create-resource.tsx
```diff
@@ -21,6 +24,22 @@ interface Props {
 @observer
 export class CreateResource extends React.Component<Props> {
   @observable error = "";
+  @observable templates:GroupSelectOption<SelectOption>[] = [];
+
+  async componentDidMount(){
+    let badgetTemplates = await createResourceStore.getLensTemplates();
+
+    badgetTemplates = await createResourceStore.addUserTemplates(badgetTemplates);
```
Nokel81: This seems backwards since it requires getting the lens templates and then badge templates.

I think it would be better to get both in tandem and then perform a merge.

src/renderer/components/dock/create-resource.tsx
```diff
@@ -21,6 +24,22 @@ interface Props {
...
+    badgetTemplates.map((group) => {this.templates.push(this.convertBadgedGroup(group));});
+  }
+
+  convertBadgedGroup(badgedGroup: BadgedGroupSelectOption):GroupSelectOption {
+    const options = badgedGroup.options.map((option:BadgedSelectOption) => {
+      return {label: this.renderSelectOption(option.label.toString(), option.badge), value: option.value};
+    });
```
Nokel81:
```suggestion
    const options = badgedGroup.options.map(({label, badge, value }) => ({
      label: this.renderSelectOption(label.toString(), badge), 
      value,
    }));
```

src/renderer/components/dock/create-resource.tsx
```diff
@@ -67,6 +90,40 @@ export class CreateResource extends React.Component<Props> {
...
+  }
+
+  renderSelectGroup(option: string, badge?: string){
+    return this.renderSelectItem(option,"select-template-group", badge );
+  }
+
+  renderSelectItem(option: string, className: string, badge?: string) {
+    return(
+      <div className={cssNames("flex select-template-item", className)}>
+        <span>{option} </span>
```
Nokel81: Shouldn't we render the label? or is that what this is.

Also we should only render the space if there is a badge.

src/renderer/components/dock/create-resource.tsx
```diff
@@ -21,6 +24,22 @@ interface Props {
 @observer
 export class CreateResource extends React.Component<Props> {
   @observable error = "";
+  @observable templates:GroupSelectOption<SelectOption>[] = [];
```
Nokel81, line 28: I think that we should have a store of some kind that watches for changes (using https://github.com/paulmillr/chokidar). So that if we open this component multiple times we don't reload every time.

Nokel81 (changes requested):
Please incorporate https://github.com/lensapp/lens/pull/644

src/renderer/components/dock/create-resource.tsx
pashevskii: I throw label here as option: string
https://github.com/lensapp/lens/pull/2327/commits/0b1059c240a86e740a8ff5c4b848874d569159da#diff-6035fe6796405ec877bfcd890b6ee764ac0f3edd60dcd38d2c7f5ba6c7b586edR38
I'll rename it on next commit

src/renderer/components/dock/create-resource.tsx
aleksfront: My guess this improvement could be moved to next PR if needed.

Nokel81: I don't think we push that off, this is rather important.

src/renderer/components/dock/create-resource.tsx
pashevskii: Unfortunately, I didn't catch what do you mean
BadgedGroupSelectOption - it is an extended GroupSelectOption class that allows storing labels and badges as separated values. 
Basically, I use it for getting info from create-resource.store.ts without including react libraries into that file, then I convert BadgedGroupSelectOption to GroupSelectOption where GroupSelectOption.label: Node is a compilation of BadgedGroupSelectOption.label:string and BadgedGroupSelectOption.badge:string (via renderSelectItem)

Nokel81: I mean that the building up of `BadgedGroupSelectOption[]` from the user templates folder should be done seperately from the lens template folder, and then merged.

pashevskii: Why? Merging logic is the most part of addUserTemplates, it is because we need to check built-in-template-folder absence before adding a new one in the userTemplatesGroups and marking this one by Badge (or marking all user files in this group if the folder exists) 
We can leave this logic and convert BadgedGroupSelectOption[] to GroupSelectOption[] twice but It looks like some excess to me.

Nokel81: The reason why I am asking about this is that both of them read from the file system. It would be both easier to test and easier to review if those two actions were split.

aleksfront:
@pashevskii @Nokel81 @ixrock What if we remove all built-in groups like `Configuration`, `Workloads` etc and put all built-in yaml templates into one `Lens` group? This will allows us:
1. Remove merge logic (when user created `Configuration` folder, for example. Now user templates gets merged inside built-in)
2. Remove need to `<Badge/>`. Now all user groups are marked with `User`. If we'll keep only 1 built-in group, there's no need to mark user ones.
3. Make list view more simple visually.
4. Let user be agnostic-free from our pre-defined groups.

Nokel81:
Sure sounds good.

src/renderer/components/dock/create-resource.store.ts
```diff
@@ -1,14 +1,57 @@
...
+      ? path.parse(path.relative(templatesPath,v)).dir
+      : defaultGroup)) : {};
+  }
+
+  async getMergedTemplates() {
+    return {...this.getTemplates(this.userTemplatesFolder, "ungrouped"),...this.lensTemplates};
+  }
+
+  async watchUserTemplates(calback: ()=> void){
+    calback();
```
aleksfront: Please change spelling: `calback` -> `callback`.

src/renderer/components/dock/create-resource.store.ts
```diff
@@ -1,14 +1,57 @@
...
+  }
+
+  get lensTemplates() {
+    return this.getTemplates(this.lensTemplatesFolder, "lens");
+  }
+
+  getTemplates(templatesPath: string, defaultGroup: string) {
+    const templates = filehound.create().path(templatesPath).ext("yaml").depth(1).findSync();
+
+    return templates ? (groupBy(templates,(v:string) => path.relative(templatesPath,v).split(path.sep).length>1
```
aleksfront: Maybe it's better to move this grouping trails into separate function with clean name, to better understand the intention.

src/renderer/components/dock/create-resource.tsx
```diff
@@ -21,6 +24,25 @@ interface Props {
...
+
+  async componentDidMount(){
```
aleksfront: `async` can be removed since we're not using `await` here.

src/renderer/components/dock/create-resource.tsx
```diff
@@ -21,6 +24,25 @@ interface Props {
...
+      createResourceStore.getMergedTemplates().then(
+        templatesDictionary => {
+          Object.keys(templatesDictionary).forEach(group => {
+            this.templates.push(this.convertEntryToGroup(group, templatesDictionary[group]));
+          });
+        });
+    });
+  }
+
+  convertEntryToGroup(group:string, items:string[]):GroupSelectOption {
```
aleksfront: Maybe simpler `convertToGroup()`?

aleksfront (changes requested):
Numerous duplicates get added to the Select if some folder gets renamed. Same things happened after remove and return a folder inside `.k8slens/templates`. Here is an example:

https://user-images.githubusercontent.com/9607060/113825339-8570ad80-9789-11eb-82dc-9b7f7ca7b77e.mp4

src/renderer/components/dock/create-resource.tsx
```diff
@@ -21,6 +24,26 @@ interface Props {
...
 export class CreateResource extends React.Component<Props> {
   @observable error = "";
+  @observable templates:GroupSelectOption<SelectOption>[] = [];
+
+  componentDidMount() {
+    createResourceStore.getMergedTemplates().then(v => this.updateGroupSelecOptions(v));
+    createResourceStore.watchUserTemplates(() => createResourceStore.getMergedTemplates().then(v => this.updateGroupSelecOptions(v)));
+  }
+
+  updateGroupSelecOptions(templates :{[x:string]: string[]}) {
```
aleksfront: `updateGroupSelecOptions` -> `updateGroupSelectOptions`

aleksfront:
@pashevskii Found 1 minor thing - the dropdown doesn't clear its state (selected item) when new `Create Resource` tab created. Otherwise, looks good!
select template dropdown

src/renderer/components/dock/create-resource.store.ts
```diff
@@ -1,13 +1,62 @@
...
+  get userTemplatesFolder():string {
+    return path.join(os.homedir(), ".k8slens", "templates");
+  }
+
+  get lensTemplates() {
+    return this.getTemplates(this.lensTemplatesFolder, "lens");
+  }
+
+  getTemplates(templatesPath: string, defaultGroup: string) {
+    const templates = filehound.create().path(templatesPath).ext("yaml").depth(1).findSync();
```
Nokel81: This should be async.

src/renderer/components/dock/create-resource.store.ts
```diff
@@ -1,13 +1,62 @@
...
 import { dockStore, IDockTab, TabKind } from "./dock.store";
 
 @autobind()
 export class CreateResourceStore extends DockTabStore<string> {
+
   constructor() {
     super({
       storageKey: "create_resource"
     });
+    fs.ensureDirSync(this.userTemplatesFolder);
```
Nokel81, line 18: I don't think we should create this if it isn't created.

src/renderer/components/dock/create-resource.tsx
```diff
@@ -20,7 +23,28 @@ interface Props {
...
+    this.templates = [];
+
+    Object.keys(templates).forEach(group => {
+      this.templates.push(this.convertToGroup(group, templates[group]));
+    });
```
Nokel81:
```suggestion
    this.templates = Object.entries(templates)
      .map(([name, grouping]) => this.convertToGroup(name, grouping));
```

src/renderer/components/dock/create-resource.tsx
Nokel81: To add `templates: Record<string, string[]>`

src/renderer/components/dock/create-resource.store.ts
Nokel81: Also, why not also include `json` files. Also this seems like we could just use `fs.promises.readDir` and loose this extra dependency.

src/renderer/components/dock/create-resource.store.ts
pashevskii: https://github.com/lensapp/lens/pull/2327#discussion_r594432805

src/renderer/components/dock/create-resource.store.ts
pashevskii: > This should be async.

Why? 
It is called within the async context + I use it in getter for built-in templates, it allows to exclude re-reading built-in templates directory on any change in the user templates directory

aleksfront (dismissed):

aleksfront (dismissed):

src/renderer/components/dock/create-resource.store.ts
pashevskii: > Also this seems like we could just use `fs.promises.readDir` and loose this extra dependency.

https://github.com/lensapp/lens/pull/2327#discussion_r598697196 Some time ago you offered to use extra library for searching files.

Nokel81: I think you are misunderstanding how async and sync play together, calling a sync function (ignoring computational functions) in an async function basically makes the async function sync.

I don't think that using a `getter` is a good enough reason to not have this being `async`. Furthermore, `getters` should be cachable and **definitely shouldn't** touch the file system or any outside dependencies. I should call that a big anti-pattern.

src/renderer/components/dock/create-resource.store.ts
Nokel81: Yes I know I said that. I guess I wasn't being clear. We can fix that later.

I meant to say that you could do something like the following:

```js
try {
    return await fs.promises.readDir(this.userTemplatesFolder, { withFileTypes: true });
} catch (error) {
    if (error.code === "ENOENT") { // dir doesn't exists
        return [];
    }

    throw error; // else propogate.
}
```

src/renderer/components/dock/create-resource.store.ts
Nokel81: > > Also this seems like we could just use `fs.promises.readDir` and loose this extra dependency.
> 
> [#2327 (comment)](https://github.com/lensapp/lens/pull/2327#discussion_r598697196) Some time ago you offered to use extra library for searching files.

I am sorry for having changing what I was suggesting. I am much more concerned about the `async` vs `sync` then about the extra dependency.

Nokel81 (dismissed):
We can improve this more in later PRs.

src/renderer/components/dock/create-resource.store.ts
pashevskii: You are right, there is some misunderstanding async programming from my side. I've fixed this one. Could you please to check it. I did some tests and it seems it works

aleksfront (approved):
LGTM! @Nokel81 Please take a look on the fixes

Nokel81 (approved):
Looks good.

PR #2331, opened 2021-03-12
Add colour picker for Letter Cluster Icons

### LINKED ISSUE #2379: Option To Select Cluster Icon Color
Option to select the cluster icon color

**Why is this needed**:
Just a nice to have. Not really need, but just would help to identify which cluster is which.

Nokel81:
This is my proposal for augmenting #2153.

1. Add changing background colour
2. Remove default of blue background (the white boarder represents active well enough)

### DIFF
--- a/package.json
+++ b/package.json
@@ -200,6 +200,7 @@
     "electron-window-state": "^5.0.3",
     "filenamify": "^4.1.0",
     "fs-extra": "^9.0.1",
+    "grapheme-splitter": "^1.0.4",
     "handlebars": "^4.7.6",
     "http-proxy": "^1.18.1",
     "js-yaml": "^3.14.0",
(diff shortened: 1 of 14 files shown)


jakolehm:
@Nokel81 could you describe what things are changed and maybe include a screenshot (or video).

Nokel81:
This PR adds the ability to set the icon colour (for when the letters are used). It also changes a bit of the styling around the letters so that it works with all colours.

Nokel81:
https://user-images.githubusercontent.com/8225332/111158857-e6cca480-856e-11eb-99eb-2f953a18a9ba.mov

Nokel81:
I have fixed the sluggishness in the video. And changed how the label is.

Screen Shot 2021-03-15 at 9 55 01 AM

Nokel81:
In the future we could clean up the file picker it self so that the icon could be clicked, but I think that is out of scope.

Nokel81:
Closing because we don't have cluster icons any more and this should be split

PR #2334, opened 2021-03-15
Start hidden on autostart

jakolehm:
Makes autostart lighter & less annoying. Also helps with `main` extension development because a developer can start Lens with `--hidden`.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -5,7 +5,7 @@ import "../common/prometheus-providers";
 import * as Mobx from "mobx";
 import * as LensExtensions from "../extensions/core-api";
 import { app, autoUpdater, ipcMain, dialog, powerMonitor } from "electron";
-import { appName } from "../common/vars";
+import { appName, isMac } from "../common/vars";
 import path from "path";
 import { LensProxy } from "./lens-proxy";
 import { WindowManager } from "./window-manager";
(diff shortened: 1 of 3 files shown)


jakolehm:
@Nokel81 could this be included in 4.2.0?

aleksfront (approved):

src/main/index.ts
```diff
@@ -151,9 +151,17 @@ app.on("ready", async () => {
   extensionLoader.init();
   extensionDiscovery.init();
 
+  // Start the app without showing the main window when auto starting on login
+  // (On Windows and Linux, we get a flag. On MacOS, we get special API.)
+  const startHidden = process.argv.includes("--hidden") || (isMac && app.getLoginItemSettings().wasOpenedAsHidden);
```
ixrock, line 156: Is it electron's flag or specified by Lens? (if by lens => should we collect all available cli flags in some common place, e.g. `vars.ts`?)
Is it different from `--process-start-args: --hidden`? (found in docs [here](https://www.electronjs.org/docs/api/app#appsetloginitemsettingssettings-macos-windows))

ixrock (approved):

Nokel81:
@jakolehm I think that this can be added to 4.2.0

PR #2335, opened 2021-03-15
Restrict what vesions of node-pty can be downloaded

Nokel81:

This is needed because `0.10.0` is for a newer version of node. And CI is currently broken.


jakolehm (approved):

jim-docker (approved):

Nokel81:
After doing some more digging this isn't actually the problem :( Will make a new PR, sorry for the noise.

PR #2342, opened 2021-03-16
Fix workspace overview switching and enabled state storage

Nokel81:
- Workspaces that are managed can now be enabled without the store removing that flag unconditionally via a roundtrip to the FS

- Workspace overview should be reactive to its computed prop

### DIFF
--- a/src/common/workspace-store.ts
+++ b/src/common/workspace-store.ts
@@ -26,6 +26,8 @@ export interface WorkspaceState {
   enabled: boolean;
 }
 
+const updateFromModel = Symbol("updateFromModel");
+
 /**
  * Workspace
  *
@@ -130,7 +132,7 @@ export class Workspace implements WorkspaceModel, WorkspaceState {
     Object.assign(this, state);
   }
 
-  @action updateModel(model: WorkspaceModel) {
+  @action [updateFromModel](model: WorkspaceModel) {
     Object.assign(this, model);
   }
 
@@ -251,10 +253,9 @@ export class WorkspaceStore extends BaseStore<WorkspaceStoreModel> {
 
   @action
   addWorkspace(workspace: Workspace) {
-    workspace.name = workspace.name.trim();
     const { id, name } = workspace;
 
-    if (!name || this.getByName(name)) {
+    if (!name.trim() || this.getByName(name.trim())) {
       return;
     }
     this.workspaces.set(id, workspace);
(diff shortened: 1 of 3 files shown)


src/renderer/components/+landing-page/workspace-overview.tsx
```diff
@@ -23,20 +24,31 @@ enum sortBy {
...
+    ]);
   }
 
-
   showCluster = ({ clusterId }: ClusterItem) => {
     navigate(clusterViewURL({ params: { clusterId } }));
   };
 
   render() {
+    const { workspaceClusterStore } = this;
```
jim-docker, line 42: You can remove `this.` on lines 59 and 84

src/common/workspace-store.ts
```diff
@@ -130,6 +130,10 @@ export class Workspace implements WorkspaceModel, WorkspaceState {
     Object.assign(this, state);
   }
 
+  @action updateModel(model: WorkspaceModel) {
```
jakolehm: Should this be private/protected?

Nokel81: We can make this a unexported symbol if you would like

jim-docker (dismissed):

src/common/workspace-store.ts
```diff
@@ -310,7 +310,7 @@ export class WorkspaceStore extends BaseStore<WorkspaceStoreModel> {
     }
 
     const currentWorkspaces = this.workspaces.toJS();
-    const newWorkspaceIds = new Set<WorkspaceId>();
+    const newWorkspaceIds = new Set<WorkspaceId>([WorkspaceStore.defaultId]); // never delete default
```
jim-docker, line 313: oh yeah, that's important. :)
You could hold changes to this file till 4.2.1 (the other changes fix the problem Mario found, right?)

jim-docker (approved):

src/common/workspace-store.ts
Nokel81: I think so, though I think he was also running into this so I think I'll keep it. I'll open another PR about the other stuff which can be left to 4.2.1

jakolehm: Tests would be nice 🙂

Nokel81: Tests caught this change, do you mean more tests?

edit: I mean before I added the `[WorkspaceStore.defaultId]` to the set constructor the unit tests were failing.

jakolehm: I meant more tests (I don't see any added in this PR).

jakolehm (approved):

PR #2344, opened 2021-03-16
Fix custom resource loading spinner appears above extensions' cluster menus

Nokel81:

Before:
---
https://user-images.githubusercontent.com/8225332/111351143-3f786c00-8659-11eb-9d05-54a9fd610a1b.mov

After:
---
https://user-images.githubusercontent.com/8225332/111350879-f3c5c280-8658-11eb-9a0c-d75ff54a0f5d.mov

### DIFF
--- a/src/renderer/components/layout/sidebar.tsx
+++ b/src/renderer/components/layout/sidebar.tsx
@@ -271,7 +271,7 @@ export class Sidebar extends React.Component<Props> {
             icon={<Icon material="extension"/>}
           >
             {this.renderTreeFromTabRoutes(CustomResources.tabRoutes)}
-            {this.crdSubMenus}
+            {this.renderCustomResources()}
           </SidebarItem>
           {this.renderRegisteredMenus()}
         </div>
--- a/src/renderer/components/layout/sidebar.tsx
+++ b/src/renderer/components/layout/sidebar.tsx
@@ -2,7 +2,6 @@ import "./sidebar.scss";
 import type { TabLayoutRoute } from "./tab-layout";
 
 import React from "react";
-import { computed } from "mobx";
 import { observer } from "mobx-react";
 import { NavLink } from "react-router-dom";
 import { cssNames } from "../../utils";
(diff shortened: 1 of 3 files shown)


src/renderer/components/spinner/spinner.scss
```diff
@@ -35,9 +35,7 @@
   }
 
   &.centerHorizontal {
-    position: absolute;
-    left: 50%;
-    margin-left: calc(var(--spinner-size) / -2);
+    margin-left: 50%;
```
jim-docker: I don't see any other code applying `centralHorizontal` to a `Spinner`, but is this a generally useful change? If I apply `centralHorizontal` to a `Spinner` in another context will it have an expected effect?

It's also hard to see the effect in your videos. The spinner moved lower and to the right a bit? Still seems to be over the extension cluster menu.

jim-docker (dismissed):

src/renderer/components/layout/sidebar.tsx
```diff
@@ -46,7 +46,7 @@ export class Sidebar extends React.Component<Props> {
 
   renderCustomResources() {
     if (crdStore.isLoading) {
-      return <Spinner centerHorizontal/>;
+      return <Spinner centerHorizontal />;
     }
```
aleksfront: The desired effect could be done with
```
if (crdStore.isLoaded) {
      return (
        <div className="flex justify-center">
          <Spinner/>
        </div>
      );
    }
```

Then we can completely remove any references (and styles) for `centerHorizontal`.
sidebar spinner

Nokel81: Do you mean that we should remove the `centerHorizontal` prop from `<Spinner>`?

aleksfront: Yes, we can do this since no one using it.

jim-docker (approved):

aleksfront (approved):

PR #2349, opened 2021-03-17
Fix: closing workspace menu after clicking on iframe

### LINKED ISSUE #2317: WorkspaceMenu does not close after mouse click on iframe
The WorkspaceMenu does not close when the mouse is clicked outside it (specifically when clicking anywhere on dashboard). This looks particularly bad when the cluster settings page is opened from the top right cog, or preferences or extension pages are opened.

Screen Shot 2021-03-11 at 10 42 34 AM

aleksfront:
`iframe` doesn't have `onClick` event handler, but it's possible to work with `window.onBlur` and `document.activeElement` to detect such "clicks".

closing workspace menu

Fixes #2317 

### DIFF
--- a/src/renderer/components/menu/menu.tsx
+++ b/src/renderer/components/menu/menu.tsx
@@ -82,6 +82,7 @@ export class Menu extends React.Component<MenuProps, State> {
     window.addEventListener("click", this.onClickOutside, true);
     window.addEventListener("scroll", this.onScrollOutside, true);
     window.addEventListener("contextmenu", this.onContextMenu, true);
+    window.addEventListener("blur", this.onBlur, true);
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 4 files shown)


aleksfront:
Switched `z-index` for menu and generic pages fixing visible menu when Preferences is open.
menu behind page

src/renderer/components/menu/menu.tsx
```diff
@@ -232,6 +233,12 @@ export class Menu extends React.Component<MenuProps, State> {
     }
   }
 
+  onBlur() {
+    if (document.activeElement?.tagName == "IFRAME") {
```
Nokel81, line 237: Is this guaranteed to be upper case. Would checking the lower case be safer?

Nokel81 (approved):

src/renderer/components/menu/menu.tsx
aleksfront: First iteration was with `toLowerCase` checking. But it's safe while we're using `html` not `xhtml` for our markup.

Nokel81: okay

jim-docker (approved):
The workspace menu should probably also hide when an app menu is clicked on, but this is less serious(?)

Nokel81:
@jim-docker I agree but that is currently the case for all of Lens's `<Menu />`'s

PR #2353, opened 2021-03-17
Save sorting order when changing view

### LINKED ISSUE #763: Sorting order is not preserved when changing views
When selecting a sorting order of a column, the sorting order is removed when switching through views.
Example:
1) Go to view "Pods", select column "Age" to sort the timestamps descending 
2) Go to view "Deployments", the sorting order is now reset for the view "Deployments"
3) Go back to view "Pods", the sorting order is now reset back to default for the view "Pods" as well

It would be preferrable to me to keep the sorting order for the column throughout the Lens session _or_ throughout all subsequent Lens sessions by making Lens save the sorting order (maybe make the latter configurable?) 

**Why is this needed**:
For every view change, the user has to reselect his preferred sorting order

3.6.0-beta.1

not relevant 

Windows 10

vshakirova:
The commit allows you to keep sorting order. It saves params to local storage 

Fixes: https://github.com/lensapp/lens/issues/763

### DIFF
--- a/src/renderer/components/+cluster/cluster-issues.tsx
+++ b/src/renderer/components/+cluster/cluster-issues.tsx
@@ -134,6 +134,7 @@ export class ClusterIssues extends React.Component<Props> {
           <>Warnings: {warnings.length}</>
         </SubHeader>
         <Table
+          tableId="cluster_issues"
           items={warnings}
           virtual
           selectable
--- a/src/renderer/components/+storage-volumes/volume-details-list.tsx
+++ b/src/renderer/components/+storage-volumes/volume-details-list.tsx
@@ -64,6 +64,7 @@ export class VolumeDetailsList extends React.Component<Props> {
       <div className="VolumeDetailsList flex column">
         <DrawerTitle title="Persistent Volumes"/>
         <Table
+          tableId="storage_volume_details_list"
           items={persistentVolumes}
           selectable
           virtual={virtual}
--- a/src/renderer/components/+workloads-pods/pod-details-list.tsx
+++ b/src/renderer/components/+workloads-pods/pod-details-list.tsx
@@ -135,6 +135,7 @@ export class PodDetailsList extends React.Component<Props> {
       <div className="PodDetailsList flex column">
         {showTitle && <DrawerTitle title="Pods"/>}
         <Table
+          tableId="workloads_pod_details_list"
           items={pods}
           selectable
           virtual={virtual}
--- a/src/renderer/components/table/table.tsx
+++ b/src/renderer/components/table/table.tsx
@@ -3,7 +3,6 @@ import "./table.scss";
 import React from "react";
 import { orderBy } from "lodash";
 import { observer } from "mobx-react";
-import { observable } from "mobx";
 import { autobind, cssNames, noop } from "../../utils";
 import { TableRow, TableRowElem, TableRowProps } from "./table-row";
 import { TableHead, TableHeadElem, TableHeadProps } from "./table-head";
@@ -11,6 +10,8 @@ import { TableCellElem } from "./table-cell";
 import { VirtualList } from "../virtual-list";
 import { createPageParam } from "../../navigation";
 import { ItemObject } from "../../item.store";
+import { getSortParams, setSortParams } from "./table.storage";
+import { computed } from "mobx";
 
 export type TableSortBy = string;
 export type TableOrderBy = "asc" | "desc" | string;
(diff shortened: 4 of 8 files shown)


src/renderer/components/+cluster/cluster-issues.tsx
```diff
@@ -134,6 +134,7 @@ export class ClusterIssues extends React.Component<Props> {
           <>Warnings: {warnings.length}</>
         </SubHeader>
         <Table
+          tableId={"cluster_issues"}
```
Nokel81: No need to use `{}`

```suggestion
          tableId="cluster_issues"
```

src/renderer/components/+storage-volumes/volume-details-list.tsx
```diff
@@ -64,6 +64,7 @@ export class VolumeDetailsList extends React.Component<Props> {
       <div className="VolumeDetailsList flex column">
         <DrawerTitle title="Persistent Volumes"/>
         <Table
+          tableId={"storage_volume_details_list"}
```
Nokel81: same here

src/renderer/components/+workloads-pods/pod-details-list.tsx
```diff
@@ -135,6 +135,7 @@ export class PodDetailsList extends React.Component<Props> {
       <div className="PodDetailsList flex column">
         {showTitle && <DrawerTitle title="Pods"/>}
         <Table
+          tableId={"workloads_pod_details_list"}
```
Nokel81: and here

src/renderer/components/table/table.tsx
```diff
@@ -62,6 +64,29 @@ export class Table extends React.Component<TableProps> {
     sortSyncWithUrl: true,
   };
 
+  constructor(props: TableProps) {
+    super(props);
+
+    if (this.props.sortable) {
+      const storage = createStorage<Record<string, Partial<TableSortParams>>>("table_sort_params", {});
```
Nokel81: I think that the storage should probably be created unconditionally. Please refactor this to be more like how #2279 does it.

src/renderer/components/table/table.tsx
```diff
@@ -52,6 +53,16 @@ export const orderByUrlParam = createPageParam({
...
 
+export interface TableStorageSortParams {
+  tables: {
+    [tableId: string]: Partial<TableSortParams>;
+  }
+}
+
+const tableStorage = createStorage<TableStorageSortParams>("table_sort_params", {
+  tables: {}
+});
```
ixrock: I would extract all these logic to separated file as was done here:

https://github.com/lensapp/lens/pull/2392/files#diff-73214131a69d3577e42e23206c7f54b6c614b43693a0f443a8aa7071d704c811

vshakirova: @ixrock yeah..does it make sense because you are going to get rid of `table` in 5.x ?

aleksfront: I think it make sense because:
1. It will be available for use very soon right after `4.2.1`.
2. New table probably won't have save-to-storage features, so we will need it sooner or later.
3. It's almost done, why not merge it?

aleksfront: @Nokel81 Since it's marked as `5.0.0` there's no need to merge the PR... But this changes could be moved to `4.2.1`. Wdyt?

Nokel81: Given that we seem to be doing a slower 4.2.0 release. Sure, we might be able to get most of these things into 4.2.0. Will change the target.

vshakirova: ok, thank you, I will move storage logic to a separated file

src/renderer/components/table/table.tsx
```diff
@@ -62,13 +63,32 @@ export class Table extends React.Component<TableProps> {
...
+    if (!prevSortParams) {
+      setSortParams(this.props.tableId, Object.assign(
+        this.props.sortSyncWithUrl ? {
+          sortBy: sortByUrlParam.get(),
+          orderBy: orderByUrlParam.get(),
+        } : {},
+        this.props.sortByDefault,
+      ));
+    }
+  }
```
ixrock: Most probably we don't need to `sortSyncWithUrl` since params is persistent.
Something like this should be enough:

```suggestion
@computed get sortParams(){
 return Object.assign({}, this.props.sortByDefault, getSortParams(this.props.tableId));
}
```

@Nokel81 @aleksfront wdyt?

Nokel81: Yeah that is probably the better way to do it.

vshakirova: thank you, fixed

ixrock (approved):
LGTM

Nokel81 (approved):

PR #2355, opened 2021-03-17
Fix being able to view clusters outside the current workspace

### LINKED ISSUE #1759: An extension cannot reliably ask Lens to activate a cluster
After some hacking, this seems to work, although it doesn't _always_ work (not sure why):
```javascript
  Store.clusterStore.activeClusterId = clusterId;
  extension.navigate(`/cluster/${clusterId}`);
```

I noticed that `clusterStore` does not provide a `setActive(clusterId)` similar to `workspaceStore.setActive(wsId)` where the latter, when called, activates the workspace AND switches to it.

On the surface, it would be seem logical to have `clusterStore.setActive(clusterId)`, and when called, Lens activates the cluster AND switches the view to it.

Note that today, just doing `Store.clusterStore.activeClusterId = clusterId` will only _activate_ (or at least, seem to) the cluster (by putting a white square behind its icon in the left sidebar, like what happens when you click the icon), however it does NOT also navigate to that cluster.

### LINKED ISSUE #2348: Possible to view clusters not in current workspace
**Describe the bug**
The `ClusterStore.activeCluster` is not always kept up to date with the current workspace. Leading to opening a cluster in one workspace and then switching workspaces and the cluster is still active.

**To Reproduce**
1. Go to the `default` workspace
2. Click on an added cluster
3. Go to another workspace that doesn't have that cluster and doesn't have a last selected cluster
4. See that the same cluster is still visible.

**Expected behavior**
The cluster view frame should be empty

**Extra Information:**
It probably should just be deprecated and delegated to `WorkspaceStore`

Nokel81:
- Completely removes ClusterStore.activeCluster
- Every workspace now tracks it current activeCluster
- If an active cluster is removed then the workspace's activeClusterId is set to undefined
- Only show welcome notification on the first time a non-managed workspace is viewed in the workspace overview
- Add unit tests for the WorkspaceStore
- Add validation that only valid clusters can be set to the activeClusterId field
- Marks the current workspace as disabled in the switch workspace's command

fixes #2348 

### future work:
- I could make it so that if you disconnect from a cluster it pick another one to go to instead of popping you back to the overview screen (see end of video).

### New look:

https://user-images.githubusercontent.com/8225332/111534350-44631b80-873e-11eb-8580-c42cd898a24d.mov

### DIFF
--- a/src/common/workspace-store.ts
+++ b/src/common/workspace-store.ts
@@ -6,8 +6,12 @@ import { appEventBus } from "./event-bus";
 import { broadcastMessage, handleRequest, requestMain } from "../common/ipc";
 import logger from "../main/logger";
 import type { ClusterId } from "./cluster-store";
+import { Cluster } from "../main/cluster";
+import migrations from "../migrations/workspace-store";
 
 export type WorkspaceId = string;
+
+export class InvariantError extends Error {}
 
 export interface WorkspaceStoreModel {
   workspaces: WorkspaceModel[];
@@ -19,7 +23,7 @@ export interface WorkspaceModel {
   name: string;
   description?: string;
   ownerRef?: string;
-  lastActiveClusterId?: ClusterId;
+  activeClusterId?: ClusterId;
 }
 
 export interface WorkspaceState {
@@ -61,18 +65,19 @@ export class Workspace implements WorkspaceModel, WorkspaceState {
    */
   @observable ownerRef?: string;
 
+  @observable private _enabled = false;
+
   /**
-   * Last active cluster id
-   *
-   * @observable
+   * The active cluster within this workspace
    */
-  @observable lastActiveClusterId?: ClusterId;
+  #activeClusterId = observable.box<ClusterId | undefined>();
 
+  get activeClusterId() {
+    return this.#activeClusterId.get();
+  }
 
-  @observable private _enabled: boolean;
-
-  constructor(data: WorkspaceModel) {
-    Object.assign(this, data);
+  constructor(model: WorkspaceModel) {
+    this[updateFromModel](model);
 
     if (!ipcRenderer) {
       reaction(() => this.getState(), () => {
@@ -86,9 +91,9 @@ export class Workspace implements WorkspaceModel, WorkspaceState {
    *
    * Workspaces that don't have ownerRef will be enabled by default. Workspaces with ownerRef need to explicitly enable a workspace.
    *
-   * @observable
+   * @computed
    */
-  get enabled(): boolean {
+  @computed get enabled(): boolean {
     return !this.isManaged || this._enabled;
   }
 
@@ -98,9 +103,82 @@ export class Workspace implements WorkspaceModel, WorkspaceState {
 
   /**
    * Is workspace managed by an extension
+   *
+   * @computed
    */
-  get isManaged(): boolean {
-    return !!this.ownerRef;
+  @computed get isManaged(): boolean {
+    return Boolean(this.ownerRef);
+  }
+
+  @computed get activeCluster(): Cluster | undefined {
+    return clusterStore.getById(this.activeClusterId);
+  }
+
+  /**
+   * Resolves the clusterId or cluster, checking some invariants
+   * @param clusterOrId The ID or cluster object to resolve
+   * @returns A Cluster instance of the specified cluster if it is in this workspace
+   * @throws if provided a falsey value or if it is an unknown ClusterId or if
+   * the cluster is not in this workspace.
+   */
+  private resolveClusterOrId(clusterOrId: ClusterId | Cluster): Cluster {
+    if (!clusterOrId) {
+      throw new InvariantError("Must provide a Cluster or a ClusterId");
+    }
+
+    const cluster = typeof clusterOrId === "string"
+      ? clusterStore.getById(clusterOrId)
+      : clusterOrId;
+
+    if (!cluster) {
+      throw new InvariantError(`ClusterId ${clusterOrId} is invalid`);
+    }
+
+    if (cluster.workspace !== this.id) {
+      throw new InvariantError(`Cluster ${cluster.name} is not in Workspace ${this.name}`);
+    }
+
+    return cluster;
+  }
+
+  /**
+   * Sets workspace's active cluster to resolved `clusterOrId`. As long as it
+   * is valid
+   * @param clusterOrId the cluster instance or its ID
+   */
+  @action setActiveCluster(clusterOrId?: ClusterId | Cluster) {
+    try {
+      if (clusterOrId === undefined) {
+        this.#activeClusterId.set(undefined);
+      } else {
+        this.#activeClusterId.set(this.resolveClusterOrId(clusterOrId).id);
+      }
+    } catch (error) {
+      logger.error("[WORKSPACE]: activeClusterId was attempted to be set to an invalid value", { error, workspaceName: this.name });
+    }
+  }
+
+  /**
+   * Tries to clear the cluster as this workspace's activeCluster.
+   * @param clusterOrId the cluster instance or its ID
+   * @returns true if it matches the `activeClusterId` (and is thus cleared) else false
+   */
+  @action tryClearAsActiveCluster(clusterOrId: ClusterId | Cluster): boolean {
+    const clusterId = typeof clusterOrId === "string"
+      ? clusterOrId
+      : clusterOrId.id;
+
+    const clearActive = this.activeClusterId === clusterId;
+
+    if (clearActive) {
+      this.clearActiveCluster();
+    }
+
+    return clearActive;
+  }
+
+  @action clearActiveCluster() {
+    this.#activeClusterId.set(undefined);
   }
 
   /**
@@ -129,11 +207,15 @@ export class Workspace implements WorkspaceModel, WorkspaceState {
    * @param state workspace state
    */
   @action setState(state: WorkspaceState) {
-    Object.assign(this, state);
+    this.enabled = state.enabled;
   }
 
   [updateFromModel] = action((model: WorkspaceModel) => {
-    Object.assign(this, model);
+    this.id = model.id;
+    this.name = model.name;
+    this.description = model.description;
+    this.ownerRef = model.ownerRef;
+    this.setActiveCluster(model.activeClusterId);
   });
 
   toJSON(): WorkspaceModel {
@@ -142,7 +224,7 @@ export class Workspace implements WorkspaceModel, WorkspaceState {
       name: this.name,
       description: this.description,
       ownerRef: this.ownerRef,
-      lastActiveClusterId: this.lastActiveClusterId
+      activeClusterId: this.activeClusterId,
     });
   }
 }
@@ -152,16 +234,18 @@ export class WorkspaceStore extends BaseStore<WorkspaceStoreModel> {
   private static stateRequestChannel = "workspace:states";
 
   @observable currentWorkspaceId = WorkspaceStore.defaultId;
+
   @observable workspaces = observable.map<WorkspaceId, Workspace>();
 
   private constructor() {
     super({
       configName: "lens-workspace-store",
+      migrations
     });
 
     this.workspaces.set(WorkspaceStore.defaultId, new Workspace({
       id: WorkspaceStore.defaultId,
-      name: "default"
+      name: "default",
     }));
   }
 
@@ -233,6 +317,19 @@ export class WorkspaceStore extends BaseStore<WorkspaceStoreModel> {
     return id === WorkspaceStore.defaultId;
   }
 
+  /**
+   * Checks if `workspaceOrId` represents `WorkspaceStore.currentWorkspaceId`
+   * @param workspaceOrId The workspace or its ID
+   * @returns true if the given workspace is the currently active on
+   */
+  isActive(workspaceOrId: Workspace | WorkspaceId): boolean {
+    const workspaceId = typeof workspaceOrId === "string"
+      ? workspaceOrId
+      : workspaceOrId.id;
+
+    return this.currentWorkspaceId === workspaceId;
+  }
+
   getById(id: WorkspaceId): Workspace {
     return this.workspaces.get(id);
   }
@@ -293,14 +390,20 @@ export class WorkspaceStore extends BaseStore<WorkspaceStoreModel> {
     if (this.currentWorkspaceId === id) {
       this.currentWorkspaceId = WorkspaceStore.defaultId; // reset to default
     }
+
     this.workspaces.delete(id);
+
     appEventBus.emit({name: "workspace", action: "remove"});
     clusterStore.removeByWorkspaceId(id);
   }
 
   @action
-  setLastActiveClusterId(clusterId?: ClusterId, workspaceId = this.currentWorkspaceId) {
-    this.getById(workspaceId).lastActiveClusterId = clusterId;
+  /**
+   * Attempts to clear `cluster` as the `activeCluster` from its own workspace
+   * @returns true if the cluster was previously the active one for its workspace
+   */
+  tryClearAsActiveCluster(cluster: Cluster): boolean {
+    return this.getById(cluster.workspace).tryClearAsActiveCluster(cluster);
   }
 
   @action
--- a/src/common/workspace-store.ts
+++ b/src/common/workspace-store.ts
@@ -8,6 +8,7 @@ import logger from "../main/logger";
 import type { ClusterId } from "./cluster-store";
 import { Cluster } from "../main/cluster";
 import migrations from "../migrations/workspace-store";
+import { clusterViewURL } from "../renderer/components/cluster-manager/cluster-view.route";
 
 export type WorkspaceId = string;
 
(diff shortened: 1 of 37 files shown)


src/common/workspace-store.ts
```diff
@@ -152,16 +241,24 @@ export class WorkspaceStore extends BaseStore<WorkspaceStoreModel> {
   private static stateRequestChannel = "workspace:states";
 
   @observable currentWorkspaceId = WorkspaceStore.defaultId;
+
+  #seenWorkspaces = observable.set<WorkspaceId>();
```
ixrock: Why do we use private fields here btw?

src/renderer/components/cluster-manager/clusters-menu.tsx
```diff
@@ -32,6 +32,9 @@ interface Props {
 @observer
 export class ClustersMenu extends React.Component<Props> {
   @observable workspaceMenuVisible = false;
+  @computed get workspace() {
```
ixrock: There is no need in `@computed` when nothing to calculate/derive inside the getter.

ixrock (dismissed):
LGTM

src/common/workspace-store.ts
Nokel81: Do you mean instead of typescript's `private` modifier? I felt that this was a better way to prevent outside consumers from using the raw values

src/renderer/components/cluster-manager/clusters-menu.tsx
Nokel81: Fair, will remove

Nokel81:
Hopefully this can be in rc.2

ixrock (dismissed):

src/renderer/components/+landing-page/landing-page.tsx
```diff
@@ -11,28 +11,28 @@ import { Icon } from "../icon";
...
       });
     }
+
+    workspaceStore.markSeen(this.workspace);
   }
 
   render() {
-    const showBackButton = this.clusters.length > 0;
-    const header = <><Icon svg="logo-lens" big /> <h2>{workspaceStore.currentWorkspace.name}</h2></>;
+    const showBackButton = this.workspaceClusters.length > 0;
```
jim-docker: I think the back button should only be shown if there is an active cluster in this workspace because this is the landing page, there is no place to go back to from the landing page if there is no active cluster. (unless we go to the previous workspace, if there is one. But I think that would be a weird UX)

src/common/workspace-store.ts
```diff
@@ -6,20 +6,25 @@ import { appEventBus } from "./event-bus";
 import { broadcastMessage, handleRequest, requestMain } from "../common/ipc";
 import logger from "../main/logger";
 import type { ClusterId } from "./cluster-store";
+import { Cluster } from "../main/cluster";
+import migrations from "../migrations/workspace-store";
 
 export type WorkspaceId = string;
 
+export class InvarientError extends Error {}
```
jim-docker:
```suggestion
export class InvariantError extends Error {}
```

jim-docker (commented):
This is a lot of refactoring to introduce between 4.2.0-rc.1 and rc.2. Did you consider doing just a surgical fix for the issue now, and saving this for 4.2.1 (or even 4.3)?

src/common/workspace-store.ts
```diff
@@ -98,9 +104,88 @@ export class Workspace implements WorkspaceModel, WorkspaceState {
...
+      return false;
+    }
+
+    if (this.activeClusterId === clusterOrId.id) {
+      this.clearActiveCluster();
+
+      return true;
+    }
+
+    return false;
```
jim-docker: You could avoid duplicating logic similar to what you did in `isActive()`, `hasBeenSeen()`,`markSeen()`

src/common/workspace-store.ts
```diff
@@ -233,6 +330,44 @@ export class WorkspaceStore extends BaseStore<WorkspaceStoreModel> {
...
+  isActive(workspaceOrId: Workspace | WorkspaceId): boolean {
+    const workspaceId = typeof workspaceOrId === "string"
+      ? workspaceOrId
+      : workspaceOrId.id;
+
+    return this.currentWorkspaceId === workspaceId;
+  }
+
+  /**
+   * Checks to see if the workspace has been to the overview page before
```
jim-docker: workspace-store doesn't really know anything about overview page, does it? Is there another way you could describe this? Or is there a more appropriate place for these methods?

src/common/workspace-store.ts
```diff
@@ -98,9 +104,88 @@ export class Workspace implements WorkspaceModel, WorkspaceState {
...
+      logger.error("[WORKSPACE]: activeClusterId was attempted to be set to an invalid value", { clusterOrId, workspaceName: this.name });
+    }
+  }
+
+  /**
+   * Tries to clear the cluster as this workspace's activeCluster.
+   * @param clusterOrId the cluster instance or its ID
+   * @returns true if it matches the `activeClusterId` (and is thus cleared) else false
+   */
+  @action tryClearAsCurrentActiveCluster(clusterOrId: ClusterId | Cluster): boolean {
```
jim-docker:
```suggestion
  @action tryClearAsActiveCluster(clusterOrId: ClusterId | Cluster): boolean {
```

src/common/workspace-store.ts
```diff
@@ -293,18 +428,23 @@ export class WorkspaceStore extends BaseStore<WorkspaceStoreModel> {
...
   }
 
   @action
-  setLastActiveClusterId(clusterId?: ClusterId, workspaceId = this.currentWorkspaceId) {
-    this.getById(workspaceId).lastActiveClusterId = clusterId;
+  /**
+   * Attempts to clear `cluster` as the `activeCluster` from its own workspace
+   * @returns true if the cluster was previously the active one for its workspace
+   */
+  tryClearAsWorkspaceActiveCluster(cluster: Cluster): boolean {
```
jim-docker: This could have the same name as its `workspace` counterpart?

src/extensions/stores/cluster-store.ts
```diff
@@ -16,16 +17,21 @@ export class ClusterStore extends Singleton {
...
+    return internalWorkspaceStore.currentWorkspace.activeClusterId;
   }
 
   /**
    * Set active cluster id
+   * @deprecated use navigate
    */
   set activeClusterId(id : ClusterId) {
-    internalClusterStore.activeCluster = id;
+    console.warn("Store.ClusterStore.activeClusterId is deprecated. Use LensExtension.navigate()");
```
jim-docker, line 34: Shouldn't this still do something?
`internalWorkspaceStore.currentWorkspace.activeClusterId = id;`?

src/main/index.ts
```diff
@@ -106,11 +106,13 @@ app.on("ready", async () => {
   await Promise.all([
     userStore.load(),
     clusterStore.load(),
-    workspaceStore.load(),
     extensionsStore.load(),
     filesystemProvisionerStore.load(),
   ]);
 
+  // load this after clusterStore, because it does validation on its entries
+  await workspaceStore.load();
```
jim-docker: Can you keep it in the `promise.all()` as `clusterStore.load().then(workspaceStore.load)`?

src/extensions/stores/cluster-store.ts
Nokel81: So the previous code was basically completely broken so I doubt that anyone was really using it and I don't think we need to keep the behaviour. This wrapper class is marked as `@beta` anyway. I kept this setter (marking it as `@deprecated`) so that things will still build but I don't really think we need to maintain the broken code. 

I can be persuaded to the contrary, that is just my current position.

src/main/index.ts
Nokel81: Yes I think so

src/common/workspace-store.ts
Nokel81: Probably

src/common/workspace-store.ts
Nokel81: Maybe as a store like what is in #2279 ?

src/extensions/stores/cluster-store.ts
jim-docker: You're right, it's no big deal.  I leave it up to you. 

But if it's easy to keep supporting that broken behaviour as deprecated then why not? Otherwise someone will say "I know it was broken, but now it's brokener". :)

Nokel81: Okay sounds good. I am also now working on #1759 so in that PR I'll change what the deprecation message is.

Nokel81:
I did consider a more patchwork fix, we could push this to 4.2.1 depending on how other people feel about this.

I think that if we test it enough it should be fine to put it into rc.2 though.

src/extensions/stores/cluster-store.ts
```diff
@@ -18,7 +18,7 @@ export class ClusterStore extends Singleton {
   /**
    * Active cluster id
    *
-   * @deprecated use `workspaceStore.activeCluster`
+   * @deprecated use `workspaceStore.currentWorkspace.activeClusterId`
    */
   get activeClusterId(): string {
     console.warn("get Store.ClusterStore.activeClusterId is deprecated. Use workspace.activeCluster");
```
jim-docker: use workspaceStore.currentWorkspace.activeClusterId

src/renderer/components/+landing-page/landing-page.tsx
```diff
@@ -3,14 +3,17 @@ import React from "react";
...
+import { WorkspaceId, workspaceStore } from "../../../common/workspace-store";
 import { WorkspaceOverview } from "./workspace-overview";
 import { PageLayout } from "../layout/page-layout";
 import { Notifications } from "../notifications";
 import { Icon } from "../icon";
+import { createStorage } from "../../utils";
 
 @observer
 export class LandingPage extends React.Component {
+  private static storage = createStorage<WorkspaceId[]>("landing_page", []);
```
jim-docker: so it will persist, nice, great idea.

src/renderer/components/+landing-page/landing-page.tsx
```diff
@@ -20,14 +23,17 @@ export class LandingPage extends React.Component {
...
+    if (!workspacesSeen.has(this.workspace.id) && this.workspaceClusters.length === 0) {
       Notifications.info(<><b>Welcome!</b><p>Get started by associating one or more clusters to Lens</p></>, {
         timeout: 30_000,
         id: "landing-welcome"
       });
     }
 
-    workspaceStore.markSeen(this.workspace);
+    workspacesSeen.add(this.workspace.id);
+    LandingPage.storage.set(Array.from(workspacesSeen));
```
jim-docker: I think the storage will also need to be updated wherever workspaces are deleted, so they can be removed from this storage.

Nokel81:
I have moved this to target 4.2.1

jim-docker:
> I have moved this to target 4.2.1

this will be a known issue for 4.2.0 then?

Nokel81:
Yes in the final release notes I'll add that.

src/extensions/stores/cluster-store.ts
```diff
@@ -21,7 +21,7 @@ export class ClusterStore extends Singleton {
    * @deprecated use `workspaceStore.currentWorkspace.activeClusterId`
    */
   get activeClusterId(): string {
-    console.warn("get Store.ClusterStore.activeClusterId is deprecated. Use workspace.activeCluster");
+    console.warn("get Store.ClusterStore.activeClusterId is deprecated. Use workspace..currentWorkspace.activeClusterId");
```
jim-docker: two dots? :)

src/renderer/components/+landing-page/landing-page.tsx
```diff
@@ -23,17 +23,23 @@ export class LandingPage extends React.Component {
   }
 
   componentDidMount() {
-    const workspacesSeen = new Set(LandingPage.storage.get());
+    // ignore workspaces that don't exist
+    const seenWorkspaces = new Set(
+      LandingPage
+        .storage
+        .get()
+        .filter(id => workspaceStore.getById(id))
```
jim-docker, line 27: slick!

src/renderer/components/+landing-page/landing-page.tsx
```diff
@@ -23,17 +23,23 @@ export class LandingPage extends React.Component {
...
+    );
 
-    if (!workspacesSeen.has(this.workspace.id) && this.workspaceClusters.length === 0) {
+    if (!seenWorkspaces.has(this.workspace.id) && this.workspaceClusters.length === 0 && !this.workspace.isManaged) {
```
jim-docker: just noticed this might need to be done in a `reaction()`, because this page may already be visible when user creates a new one, or switches to an empty one.

jim-docker (dismissed):
I can't tell what's changed since I last reviewed. I am assuming it's just merge conflicts? I tested again, LGTM

src/renderer/ipc/index.tsx
```diff
@@ -76,7 +76,7 @@ function ListNamespacesForbiddenHandler(event: IpcRendererEvent, ...[clusterId]:
     (
       <div className="flex column gaps">
         <b>Add Accessible Namespaces</b>
-        <p>Cluster <b>{clusterStore.active.name}</b> does not have permissions to list namespaces. Please add the namespaces you have access to.</p>
+        <p>Cluster <b>{clusterStore.getById(clusterId).name}</b> does not have permissions to list namespaces. Please add the namespaces you have access to.</p>
```
ixrock: Could be `clusterStore.getById(clusterId)` undefined?

src/renderer/components/+landing-page/landing-page.tsx
```diff
@@ -1,38 +1,58 @@
...
 import { WorkspaceOverview } from "./workspace-overview";
 import { PageLayout } from "../layout/page-layout";
 import { Notifications } from "../notifications";
 import { Icon } from "../icon";
+import { createStorage } from "../../utils";
 
 @observer
 export class LandingPage extends React.Component {
-  @observable showHint = true;
+  private static storage = createStorage<WorkspaceId[]>("landing_page", []);
```
ixrock: Storage key name doesn't reflect what it stores. 
Why it's tighted to landing-page only? Maybe just `seen_workspaces`?

ixrock (dismissed):

src/renderer/ipc/index.tsx
Nokel81: I guess in theory though this event is currently only sent by the main process while loading a cluster so I doubt that the cluster won't exist for the renderer.

src/renderer/components/+landing-page/landing-page.tsx
Nokel81: Sure I can rename it to `seen_workspaces`. What do you mean by "tighted"? That isn't a word that I know.

ixrock: Ok, I mean `tied` :)

ixrock (dismissed):

src/renderer/components/+landing-page/landing-page.tsx
Nokel81: Ah now I understand the question. This is the only place that `"seen_workspaces"` is used and Jim pointed out (reasonably imo) that storing it in `WorkspaceStore` doesn't make much sense.

jim-docker (dismissed):

stefcameron:
@Nokel81 Can you give me an overview of what the new API to activate a cluster in a workspace looks like now, with changes in this PR, since it sounds like it'll fix #1759 ?

Nokel81:
@stefcameron I think it will be as simple as `<LensExtensionInstance>.navigate("/cluster/<cluster-id>");` but I will verify this in the morning.

Nokel81:
@stefcameron Done, I have added it. I don't know why the `LensExtension.navigate` method is only limited to a specific extension but whatever.

You need to do the following:
```ts
import { Store } from "@k8slens/extensions";

...
const cluster = ...;

await Store.workspaceStore.setActiveCluster(cluster); // or cluster ID
```

stefcameron:
> @stefcameron Done, I have added it. I don't know why the `LensExtension.navigate` method is only limited to a specific extension but whatever.
> 
> You need to do the following:
> 
> ```ts
> import { Store } from "@k8slens/extensions";
> 
> ...
> const cluster = ...;
> 
> await Store.workspaceStore.setActiveCluster(cluster); // or cluster ID
> ```

Thanks for following-up on this, @Nokel81 ! I will watch for this in 4.2.1 (unless I get to it off `master` before then) and will update my code accordingly.

Based on the call, the cluster doesn't have to be in the active workspace. Could be in some other workspace, and Lens will also activate the workspace as a result? Of course, this means Lens now ensures (maybe it always has; I haven't specifically tested) that all clusters have unique IDs...

Nokel81:
No that call will update the active workspace too. Though now that I think about it. I think we should instead have it be `Navigation.navigate("/cluster/<cluster-id");` and instead change the active workspace on the fly. So I might back out this change.

Nokel81:
But I need to think about it. I only mention this because workspace are going away and I would prefer if what I told you will also work in 5.0

stefcameron:
> But I need to think about it. I only mention this because workspace are going away and I would prefer if what I told you will also work in 5.0

Whaaaaat? 😉 OK, I'd like to know a way that will survive, though if workspaces disappear, that will definitely impact my extension.

BTW, `extension.navigate("/cluster/<cluster-id>");` is what I'm already doing, but it's not reliable, which is the reason for #1759

Nokel81:
@stefcameron So as we discussed, I am going to leave this API as is since putting more work into this is misplaced effort.

stefcameron:
> @stefcameron So as we discussed, I am going to leave this API as is since putting more work into this is misplaced effort.

You mean it will be `await Store.workspaceStore.setActiveCluster(cluster); // or cluster ID` for now until Lens 5.0 when workspaces disappear? I assume there will be some equivalent method off the Catalogue API to activate an object at that point.

Nokel81:
Yeah that.

src/renderer/components/select/select.scss
```diff
@@ -104,9 +104,7 @@ html {
 
       &--is-disabled {
         cursor: not-allowed;
-        background: none !important;
```
jakolehm: Are these changes by accident here?

Nokel81: I honestly don't remember. I think I changed this because there are some places where the disabled text looks weird without it (since it changes the text colour).

jakolehm: Should we revert this.. feels a bit dangerous change.

Nokel81: Sure I can, I don't think it is as dangerous as it feels. I'll make a new PR with this change if I come across the need again.

jakolehm (approved):

nevalla (approved):
LGTM

PR #2364, opened 2021-03-19
Fix: editing resource is broken when another resource is being editing

ixrock:
Regression from #2279

_Before:_

https://user-images.githubusercontent.com/6377066/111774474-2c170c00-88b8-11eb-83e3-cdccc906ec3f.mov

_After:_

https://user-images.githubusercontent.com/6377066/111774876-b9f2f700-88b8-11eb-9109-65f80c88b6c1.mov

### DIFF
--- a/src/renderer/components/dock/edit-resource.tsx
+++ b/src/renderer/components/dock/edit-resource.tsx
@@ -1,7 +1,7 @@
 import "./edit-resource.scss";
 
 import React from "react";
-import { observable, when } from "mobx";
+import { action, computed, observable } from "mobx";
 import { observer } from "mobx-react";
 import jsYaml from "js-yaml";
 import { IDockTab } from "./dock.store";
@@ -11,6 +11,7 @@ import { InfoPanel } from "./info-panel";
 import { Badge } from "../badge";
 import { EditorPanel } from "./editor-panel";
 import { Spinner } from "../spinner";
+import { KubeObject } from "../../api/kube-object";
 
 interface Props {
   className?: string;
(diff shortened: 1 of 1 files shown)


Nokel81 (approved):

aleksfront (approved):
LGTM

jim-docker:
This seems weird:

https://user-images.githubusercontent.com/40840436/111790787-f44c9c00-8898-11eb-9eff-b6e46fe440c2.mov

Cursor is placed at bottom of source in the second editor, and when you go back to the first editor, cursor has moved to the bottom there too.

ixrock:
> Cursor is placed at bottom of source in the second editor, and when you go back to the first editor, cursor has moved to the bottom there too.

Good catch! 
Looks like a timing issue for restoring cursor in `ace-editor` on tab switching / componentDidUpdate.

ixrock:
@jim-docker it seems like that behaviour happens only in case for first opened tab for editing (cursor position at `0:0`, then it jumps to the end line for follow-up tabs for for editing). But, actually it doesn't matter since those resources haven't been edited yet and cursor position not saved before. Otherwise I guess it works as expected, last cursor positions are restored per tab on select. Please see the video:

One thing could be improved probably, if we always keep position for freshly opened resources for editing at line 0:0 instead of jumping to the end of contents? (not critical imo)

https://user-images.githubusercontent.com/6377066/112126486-2b8ab800-8bcd-11eb-9dad-110e0e5ed05b.mov

Nokel81:
@ixrock Can you please log a bug with that, I think we should fix it so that it doesn't jump around but I agree that it isn't that critical.

ixrock:
> @ixrock Can you please log a bug with that, I think we should fix it so that it doesn't jump around but I agree that it isn't that critical.

Done ^

PR #2365, opened 2021-03-19
Wrapping ReactSelect styles into CacheProvider

### LINKED ISSUE #2219: Extensions can adversely impact Lens UI elements
After enabling the [lens-extension-cc](https://github.com/Mirantis/lens-extension-cc) extension strange things happen to Lens UI elements:

Namespace selector field doubles in height

https://user-images.githubusercontent.com/40840436/109035944-1fc3d880-7697-11eb-8d98-b58da4fe51f0.mov

Helm repos UI is mangled

https://user-images.githubusercontent.com/40840436/109036237-64e80a80-7697-11eb-8b63-ba81c50a4102.mov

Likely to due CSS configuration in extension spilling over into Lens?

aleksfront:
As it turned out, by default every instance of [emotion library](https://emotion.sh/docs/introduction) shares same `<style>` tag to add their CSS styles. This caused a bug described in #2219 when any extension with `emotion` inside gets added, it eventually erases all shared styles. This, in turn, breaks `ReactSelect` appearance, which also uses styles from Emotion library.

Solving an issue comes in 3 ways:
1. Wrap with `<CacheProvider>` areas where `ReactSelect` is used across the app. Done with this PR ✅ 
2. Wrap with `<CacheProvider>` any extension using Emotion for styling like [MCC extension](https://github.com/Mirantis/lens-extension-cc/)
3. Provide a notice about the issue itself and `<CacheProvider>` in docs to convince developers to wrap their styles properly.

**What CacheProvider does?**
It creates separated `style` tag which totally ignored by other emotion instances, so no styles get harmed inside of it.
separated styles

Now, our dropdowns are safe when MCC extension switched on

https://user-images.githubusercontent.com/9607060/111785120-40b1cf00-88cd-11eb-9564-340e5fcca715.mp4

Fixes #2219 


Nokel81:
Does this also fix https://github.com/Mirantis/lens-extension-cc/blob/master/src/renderer.tsx#L27-L44?

Nokel81 (commented):
Shouldn't we do the extension wrapping in this PR too then?

aleksfront:
> Does this also fix https://github.com/Mirantis/lens-extension-cc/blob/master/src/renderer.tsx#L27-L44?

I guess no, it doesn't change any mobx/emotion component relationships described in this comment.

aleksfront:
> Shouldn't we do the extension wrapping in this PR too then?

https://github.com/Mirantis/lens-extension-cc/pull/205

Nokel81:
I meant that shouldn't we wrap every place we take in user code that like? Maybe that is a bit too much work...

aleksfront:
> I meant that shouldn't we wrap every place we take in user code that like? Maybe that is a bit too much work...

Yes, that's a good question. You pointed right that to be 100% safe we shouldn't rely on extension developers, we must wrap any code goes from extensions within `CacheProvider` with unique `key`. Do you think pointing developers to isolate their `emotion` styles in docs not enough for this time?

Nokel81:
At this time I think it probably is good enough. I guess this is just a problem with using `emotion` and other 3rd party packages. They might have some globals that we cannot control. Though at least `emotion` has this solution.

src/renderer/components/layout/page-layout.tsx
```diff
@@ -80,11 +82,13 @@ export class PageLayout extends React.Component<PageLayoutProps> {
             <RecursiveTreeView data={navigation}/>
           </nav>
         )}
-        <div className="content-wrapper" id="ScrollSpyRoot">
-          <div className={cssNames("content", contentClass, contentGaps && "flex column gaps")}>
-            {children}
+        <CacheProvider value={createCache({ key: "page-layout" })}>
```
ixrock: Why do we wrap `CacheProvider` only in `PageLayout` and `MainLayout`?
Isn't it should be inside `app.tsx` and `lens-app.tsx`? Otherwise how you can be sure where `react-select` actually used internally in app?

aleksfront: Moved `CacheProvider` into `app.tsx` & `lens-app.tsx` covering more components at once.

Nokel81 (dismissed):

aleksfront:
Moved `CacheProvider` one level top - to the `src/renderer/bootstrap.tsx`.

Nokel81 (dismissed):

package.json
```diff
@@ -352,7 +354,7 @@
     "react-beautiful-dnd": "^13.0.0",
     "react-refresh": "^0.9.0",
     "react-router-dom": "^5.2.0",
-    "react-select": "^3.1.0",
+    "react-select": "^4.3.0",
```
ixrock, line 355: Should we also update `@types/react-select` to latest?

ixrock:
@aleksfront what could be wrong with tests?

image

package.json
aleksfront: Updated types to latest version.

src/renderer/components/select/select.tsx
```diff
@@ -22,7 +22,8 @@ export interface SelectOption<T = any> {
   label?: React.ReactNode;
 }
 
-export interface SelectProps<T = any> extends ReactSelectProps<T>, CreatableProps<T> {
+export interface SelectProps<T = any> extends ReactSelectProps<OptionTypeBase, boolean, GroupTypeBase<OptionTypeBase>> {
+  options?: readonly any[];
```
ixrock:
```suggestion
export interface SelectProps<T = OptionTypeBase> extends ReactSelectProps<T> {
  options?: readonly OptionProps<T, boolean>[];

```

src/renderer/components/select/select.tsx
```diff
@@ -68,7 +69,7 @@ export class Select extends React.Component<SelectProps> {
     return this.options.find(opt => opt === value || opt.value === value);
   }
 
-  @computed get options(): SelectOption[] {
+  @computed get options(): readonly (OptionTypeBase | GroupTypeBase<OptionTypeBase>)[] {
```
ixrock: Why do we use here `OptionTypeBase` instead of `OptionType` when:

```
export interface OptionTypeBase {
    [key: string]: any;
}
```

Every `option` must contain at least `value` and `label` fields.

src/renderer/components/select/select.tsx
```diff
@@ -6,8 +6,8 @@ import React, { ReactNode } from "react";
 import { computed } from "mobx";
 import { observer } from "mobx-react";
 import { autobind, cssNames } from "../../utils";
-import ReactSelect, { ActionMeta, components, Props as ReactSelectProps, Styles } from "react-select";
-import Creatable, { CreatableProps } from "react-select/creatable";
+import ReactSelect, { ActionMeta, components, GroupTypeBase, OptionTypeBase, Props as ReactSelectProps, Styles } from "react-select";
```
ixrock: I'd recommend to import such common names as `Styles` from external package as something more distinguishable, e.g. `Styles as ReactSelectStyles`

ixrock (changes requested):

src/renderer/components/select/select.tsx
ixrock: And again, `OptionTypeBase` shouldn't be here..

aleksfront: Thanks for improvements!

aleksfront: We're passing different types into `options` and then converting them inside `this.options()` getter. I suggest in previous iteration an `options` prop was exactly as `any[]`.

aleksfront: The good thing we can do is to get rid from `autoConvert` functions, always pass proper `options` and `value` types to this component. I've tried a couple of iterations - they went much out of scope of current PR.

aleksfront:
Fine-tuned react select types a bit more + fixed `NamespaceSelectFilter` gradients (moved to CSS pseudo elements)

dark gradients
fixed gradients

aleksfront:
Fixed tests by eliminating `emotion` dependencies from our codebase. Used "native" `react-select` `NonceProvider` for same caching purposes.

See about `NonceProvider` https://github.com/JedWatson/react-select/blob/c7041f20ae1eb49a092a98e93a4a1c6ebd171d93/packages/react-select/src/NonceProvider.js

@ixrock @Nokel81 ptal please ^

src/renderer/bootstrap.tsx
```diff
@@ -80,9 +81,14 @@ export async function bootstrap(App: AppComponent) {
       window.location.href = "about:blank";
     }
   });
+
+  const cacheProps = { nonce: "lens", cacheKey: "lens" };
+
   render(<>
     {isMac && <div id="draggable-top" />}
-    <App />
+    <StyleCache {...cacheProps}>
```
aleksfront, line 89: Passing props this way allows to ignore `NonceProvider` stale types. Current version of it requires `cacheKey`, but type definitions doesn't show that.

Nokel81: I have filled https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52073 to fix this, hopefully it gets merged soon.

Nokel81 (approved):

src/renderer/bootstrap.tsx
aleksfront: Thanks! That's cool! 🚀

src/renderer/components/select/select.tsx
```diff
@@ -6,23 +6,29 @@ import React, { ReactNode } from "react";
 import { computed } from "mobx";
 import { observer } from "mobx-react";
 import { autobind, cssNames } from "../../utils";
-import ReactSelect, { ActionMeta, components, Props as ReactSelectProps, Styles } from "react-select";
-import Creatable, { CreatableProps } from "react-select/creatable";
+import ReactSelect, { components } from "react-select";
+import Creatable from "react-select/creatable";
 import { themeStore } from "../../theme.store";
 
+import type { ActionMeta, GroupTypeBase, OptionTypeBase, Props as ReactSelectProps, Styles as ReactSelectStyles } from "react-select";
```
ixrock, line 13: You could import types from above as well:

```
import ReactSelect, { components } from "react-select";
```

ixrock (approved):

PR #2366, opened 2021-03-19
Fix: highlight sidebar's active section

ixrock:
- fix: highlight active sidebar's section (regression from #2166)
- fix: caching crd definitions to prevent sidebar's scroll jumps
- refactored & simplified `Sidebar` and `SidebarItem`

_Before:_

https://user-images.githubusercontent.com/6377066/111791141-9c7b5880-88cb-11eb-901a-bfb925ca5b0e.mov

_After:_

https://user-images.githubusercontent.com/6377066/111791121-95ece100-88cb-11eb-9ba2-e6e4735712f4.mov

### DIFF
--- a/src/renderer/components/layout/sidebar-item.tsx
+++ b/src/renderer/components/layout/sidebar-item.tsx
@@ -13,11 +13,11 @@ interface SidebarItemProps {
   id: string; // unique id, used in storage and integration tests
   url: string;
   className?: string;
-  title: React.ReactNode;
+  text: React.ReactNode;
   icon?: React.ReactNode;
   isHidden?: boolean;
   isActive?: boolean; // marks a link, by default checks `props.url` to identify active state
-  subMenu?: React.ReactNode | React.ComponentType<SidebarItemProps>[]; // props.children could be used too
+  subMenus?: React.ReactNode | React.ComponentType<SidebarItemProps>[]; // props.children could be used too
 }
 
 @observer
@@ -48,7 +48,7 @@ export class SidebarItem extends React.Component<SidebarItemProps> {
       return false; // not available currently
     }
 
-    return Boolean(this.props.subMenu || this.props.children);
+    return Boolean(this.props.subMenus || this.props.children);
   }
 
   toggleExpand = () => {
@@ -58,7 +58,7 @@ export class SidebarItem extends React.Component<SidebarItemProps> {
   };
 
   render() {
-    const { isHidden, icon, title, children, url, className, subMenu } = this.props;
+    const { isHidden, icon, text, children, url, className, subMenus } = this.props;
 
     if (isHidden) return null;
 
@@ -75,7 +75,7 @@ export class SidebarItem extends React.Component<SidebarItemProps> {
           className={cssNames("nav-item flex gaps align-center", { expandable: isExpandable })}
           onClick={isExpandable ? prevDefault(toggleExpand) : undefined}>
           {icon}
-          <span className="link-text box grow">{title}</span>
+          <span className="link-text box grow">{text}</span>
           {isExpandable && <Icon
             className="expand-icon box right"
             material={expanded ? "keyboard_arrow_up" : "keyboard_arrow_down"}
@@ -83,7 +83,7 @@ export class SidebarItem extends React.Component<SidebarItemProps> {
         </NavLink>
         {isExpandable && expanded && (
           <ul className={cssNames("sub-menu", { active: isActive })}>
-            {subMenu}
+            {subMenus}
             {children}
           </ul>
         )}
--- a/src/renderer/components/layout/sidebar-item.tsx
+++ b/src/renderer/components/layout/sidebar-item.tsx
@@ -10,14 +10,23 @@ import { sidebarStorage } from "./sidebar-storage";
 import { isActiveRoute } from "../../navigation";
 
 interface SidebarItemProps {
-  id: string; // unique id, used in storage and integration tests
+  /**
+   * Unique id, used in storage and integration tests
+   */
+  id: string;
   url: string;
   className?: string;
   text: React.ReactNode;
   icon?: React.ReactNode;
   isHidden?: boolean;
-  isActive?: boolean; // marks a link, by default checks `props.url` to identify active state
-  subMenus?: React.ReactNode | React.ComponentType<SidebarItemProps>[]; // props.children could be used too
+  /**
+   * Forces this item to be also show as active or not.
+   *
+   * Default: dynamically checks the location against the `url` props to determine if
+   * this item should be shown as active
+   */
+  isActive?: boolean;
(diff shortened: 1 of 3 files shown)


src/renderer/components/layout/sidebar.tsx
```diff
@@ -44,29 +45,50 @@ export class Sidebar extends React.Component<Props> {
     crdStore.reloadAll();
   }
 
-  renderCustomResources() {
-    if (crdStore.isLoading) {
+  @computed get crdSubMenu(): React.ReactNode {
+    if (!crdStore.isLoaded) {
```
Nokel81: I don't think that we should display the spinner if loading has failed.

src/renderer/components/layout/sidebar-item.tsx
```diff
@@ -6,42 +6,49 @@ import { cssNames, prevDefault } from "../../utils";
...
   url: string;
-  text: React.ReactNode | string;
   className?: string;
+  title: React.ReactNode;
   icon?: React.ReactNode;
   isHidden?: boolean;
-  isActive?: boolean;
-  subMenus?: TabLayoutRoute[];
+  isActive?: boolean; // marks a link, by default checks `props.url` to identify active state
+  subMenu?: React.ReactNode | React.ComponentType<SidebarItemProps>[]; // props.children could be used too
```
Nokel81: Why do we need this change? Isn't the problem that `isActive` wasn't computed?

src/renderer/components/layout/sidebar-item.tsx
```diff
@@ -6,42 +6,49 @@ import { cssNames, prevDefault } from "../../utils";
...
 import { sidebarStorage } from "./sidebar-storage";
 import { isActiveRoute } from "../../navigation";
 
 interface SidebarItemProps {
-  id: string; // Used to save nav item collapse/expand state in local storage
+  id: string; // unique id, used in storage and integration tests
   url: string;
-  text: React.ReactNode | string;
   className?: string;
+  title: React.ReactNode;
```
Nokel81: Changing this has nothing to do with fixing the bug.

Nokel81 (changes requested):
^^^

src/renderer/components/layout/sidebar-item.tsx
ixrock: It's not related to `isActive`. `props.subMenu` has basically same role as `props.children`, could be even removed (currently both adds data to sub-menu container).

src/renderer/components/layout/sidebar-item.tsx
ixrock: And what? I don't calculate changed lines (just re-sorted a bit for better look)

src/renderer/components/layout/sidebar-item.tsx
Nokel81: Yeah I have noticed that they are very similar indeed. But for a bug fix I think it should be a small as possible (within reason).

src/renderer/components/layout/sidebar-item.tsx
Nokel81: You changed a prop field name, that doesn't have anything to do with this bug fix.

src/renderer/components/layout/sidebar.tsx
ixrock: I'd prefer to keep it as is since it fixes a bug (see video1) with reloading CRD definitions each time (otherwise more files needs to be tweaked)

src/renderer/components/layout/sidebar-item.tsx
ixrock: Yes, it fits better imho + it's used currently only in `Sidebar` and `SidebarItem`. So it's a good time to change if we want it.

src/renderer/components/layout/sidebar-item.tsx
ixrock: Renamed back to `props.subMenus`

src/renderer/components/layout/sidebar.tsx
Nokel81: Where in the video?

src/renderer/components/layout/sidebar-item.tsx
```diff
@@ -6,42 +6,49 @@ import { cssNames, prevDefault } from "../../utils";
...
+  id: string; // unique id, used in storage and integration tests
   url: string;
-  text: React.ReactNode | string;
   className?: string;
+  text: React.ReactNode;
   icon?: React.ReactNode;
   isHidden?: boolean;
-  isActive?: boolean;
-  subMenus?: TabLayoutRoute[];
+  isActive?: boolean; // marks a link, by default checks `props.url` to identify active state
```
Nokel81:
```suggestion
  /**
   * Forces this item to be also show as active or not.
   *
   * Default: dynamically checks the location against the `url` props to determine if
   * this item should be shown as active
   */
  isActive?: boolean;
```

src/renderer/components/layout/sidebar-item.tsx
```diff
@@ -76,19 +83,7 @@ export class SidebarItem extends React.Component<SidebarItemProps> {
...
-            {subMenus.map(({ title, routePath, url = routePath }) => {
-              const subItemId = `${id}${routePath}`;
-
-              return (
-                <SidebarItem
-                  key={subItemId}
-                  id={subItemId}
-                  url={url}
-                  text={title}
-                  isActive={isActiveRoute({ path: url, exact: true })}
```
Nokel81, line 88: Shouldn't it be possible to just remove this line and fix the bug?

src/renderer/components/layout/sidebar.tsx
ixrock: When I'm clicking second time at `Definitions` you can see loading indicator and jumping scroll for a short time.

src/renderer/components/layout/sidebar-item.tsx
ixrock: Dunno. CDS's and 3rd level menus are handled a bit differently.

ixrock:
@Nokel81 ready for re-review, PTAL.

aleksfront (approved):
LGTM

Nokel81 (approved):
Looks good, can you do a follow up PR to remove `submenu`? So that they are just `children`?

ixrock:
> Looks good, can you do a follow up PR to remove `submenu`? So that they are just `children`?

Yes!

Nokel81:
Next time please don't merge as it make my work with releases slightly more complicated (this is really only a problem when an active release is in progress). Thanks.

ixrock:
> Next time please don't merge as it make my work with releases slightly more complicated (this is really only a problem when an active release is in progress). Thanks.

Alright! Usually you say it when you need to wait some time before merge (for various reasons).

Nokel81:
True, sorry.

PR #2372, opened 2021-03-22
Fix: Recreate proxy kubeconfig if it is deleted

nevalla:
With this PR, cluster's kubeconfig that points to authentication proxy will be re-created if it's removed from the filesystem. 

For example on Mac, the operating system will garbage collect temporary files and Lens terminal is not able to connect to the cluster anymore. 

Also, requests to fetch allowed namespaces and resources do not work anymore, which will cause some issues for sure.

### DIFF
--- a/src/main/kubeconfig-manager.ts
+++ b/src/main/kubeconfig-manager.ts
@@ -24,17 +24,17 @@ export class KubeconfigManager {
   protected async init() {
     try {
       await this.contextHandler.ensurePort();
-      this.tempFile = this.createProxyKubeconfig();
+      this.tempFile = await this.createProxyKubeconfig();
     } catch (err) {
       logger.error(`Failed to created temp config for auth-proxy`, { err });
     }
   }
 
-  getPath() {
+  async getPath() {
     // create proxy kubeconfig if it is removed
-    if (this.tempFile && !fs.pathExistsSync(this.tempFile)) {
+    if (this.tempFile && !(await fs.pathExists(this.tempFile))) {
       try {
-        this.tempFile = this.createProxyKubeconfig();
+        this.tempFile = await this.createProxyKubeconfig();
       } catch (err) {
         logger.error(`Failed to created temp config for auth-proxy`, { err });
       }
@@ -52,7 +52,7 @@ export class KubeconfigManager {
    * Creates new "temporary" kubeconfig that point to the kubectl-proxy.
    * This way any user of the config does not need to know anything about the auth etc. details.
    */
-  protected createProxyKubeconfig(): string {
+  protected async createProxyKubeconfig(): Promise<string> {
     const { configDir, cluster } = this;
     const { contextName, kubeConfigPath, id } = cluster;
     const tempFile = path.join(configDir, `kubeconfig-${id}`);
@@ -81,8 +81,8 @@ export class KubeconfigManager {
     // write
     const configYaml = dumpConfigYaml(proxyConfig);
 
-    fs.ensureDirSync(path.dirname(tempFile));
-    fs.writeFileSync(tempFile, configYaml, { mode: 0o600 });
+    await fs.ensureDir(path.dirname(tempFile));
+    await fs.writeFile(tempFile, configYaml, { mode: 0o600 });
     logger.debug(`Created temp kubeconfig "${contextName}" at "${tempFile}": \n${configYaml}`);
 
     return tempFile;
--- a/src/main/kubeconfig-manager.ts
+++ b/src/main/kubeconfig-manager.ts
@@ -32,7 +32,7 @@ export class KubeconfigManager {
 
   async getPath() {
     // create proxy kubeconfig if it is removed
-    if ((this.tempFile || this.tempFile === "") && !(await fs.pathExists(this.tempFile))) {
+    if (this.tempFile !== undefined && !(await fs.pathExists(this.tempFile))) {
       try {
         this.tempFile = await this.createProxyKubeconfig();
       } catch (err) {
--- a/src/main/kubeconfig-manager.ts
+++ b/src/main/kubeconfig-manager.ts
@@ -33,7 +33,11 @@ export class KubeconfigManager {
   getPath() {
     // create proxy kubeconfig if it is removed
     if (this.tempFile && !fs.pathExistsSync(this.tempFile)) {
-      this.tempFile = this.createProxyKubeconfig();
+      try {
+        this.tempFile = this.createProxyKubeconfig();
+      } catch (err) {
+        logger.error(`Failed to created temp config for auth-proxy`, { err });
+      }
     }
 
     return this.tempFile;
@@ -77,7 +81,7 @@ export class KubeconfigManager {
     // write
     const configYaml = dumpConfigYaml(proxyConfig);
 
-    fs.ensureDir(path.dirname(tempFile));
+    fs.ensureDirSync(path.dirname(tempFile));
     fs.writeFileSync(tempFile, configYaml, { mode: 0o600 });
     logger.debug(`Created temp kubeconfig "${contextName}" at "${tempFile}": \n${configYaml}`);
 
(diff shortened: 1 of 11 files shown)


src/main/kubeconfig-manager.ts
```diff
@@ -24,14 +24,20 @@ export class KubeconfigManager {
   protected async init() {
     try {
       await this.contextHandler.ensurePort();
-      await this.createProxyKubeconfig();
+      this.tempFile = this.createProxyKubeconfig();
```
nevalla, line 27: this might not needed anymore, since it's created on-demand anyway, but maybe doesn't hurt to have it here.

src/main/kubeconfig-manager.ts
```diff
@@ -42,7 +48,7 @@ export class KubeconfigManager {
    * Creates new "temporary" kubeconfig that point to the kubectl-proxy.
    * This way any user of the config does not need to know anything about the auth etc. details.
    */
-  protected async createProxyKubeconfig(): Promise<string> {
+  protected createProxyKubeconfig(): string {
```
nevalla: does not need to be async

src/main/kubeconfig-manager.ts
```diff
@@ -24,14 +24,20 @@ export class KubeconfigManager {
...
+      this.tempFile = this.createProxyKubeconfig();
     } catch (err) {
       logger.error(`Failed to created temp config for auth-proxy`, { err });
     }
   }
 
   getPath() {
+    // create proxy kubeconfig if it is removed
+    if (this.tempFile && !fs.pathExistsSync(this.tempFile)) {
+      this.tempFile = this.createProxyKubeconfig();
```
jim-docker: try/catch around this, like in `init()`?

src/main/kubeconfig-manager.ts
Nokel81: Why are we going towards calling sync fuctions, that can block the event loop. We should move these towards the async methods instead.

src/main/kubeconfig-manager.ts
```diff
@@ -24,14 +24,24 @@ export class KubeconfigManager {
...
     try {
       await this.contextHandler.ensurePort();
-      await this.createProxyKubeconfig();
+      this.tempFile = this.createProxyKubeconfig();
     } catch (err) {
       logger.error(`Failed to created temp config for auth-proxy`, { err });
     }
   }
 
   getPath() {
```
Nokel81: This will return `""` if `this.tempFile` is `""`. Is that what we want?

src/main/kubeconfig-manager.ts
```diff
@@ -24,14 +24,24 @@ export class KubeconfigManager {
...
+    // create proxy kubeconfig if it is removed
+    if (this.tempFile && !fs.pathExistsSync(this.tempFile)) {
+      try {
+        this.tempFile = this.createProxyKubeconfig();
+      } catch (err) {
+        logger.error(`Failed to created temp config for auth-proxy`, { err });
+      }
```
Nokel81: If there is a problem with the temp files being removed, shouldn't it be better if we store the files under `userData` instead, clearing them out on every "remove cluster" and "start Lens" event?

nevalla: probably both fixes make sense.

src/main/kubeconfig-manager.ts
nevalla: when `this.tempFile` will get ""?

src/main/kubeconfig-manager.ts
nevalla: yes... I uses sync since this method already uses sync methods. I investigated option to switch to use async, but it would require bigger refactoring, so ended up to add smaller fix here.

nevalla: AFAIK the only async call there is `fs.ensureDir` which I think is probably wrong since it's not awaited.

nevalla: Actually I can be wrong, how big effort would it be to switch to use async... will check it again.

src/main/kubeconfig-manager.ts
Nokel81: I don't know but your `if` check checks that `tempFile` is truthy (so not "") before checking if the file exists. If `tempFile` is falsey then it just returns that.

src/main/shell-session.ts
```diff
@@ -65,11 +62,13 @@ export class ShellSession extends EventEmitter {
...
   protected cwd(): string {
-    if(!this.preferences || !this.preferences.terminalCWD || this.preferences.terminalCWD === "") {
+    const { preferences } = this.cluster;
+
+    if(!preferences || !preferences.terminalCWD || preferences.terminalCWD === "") {
       return null;
     }
 
-    return this.preferences.terminalCWD;
+    return preferences.terminalCWD;
```
Nokel81, line 71: how about:

```js
    return preferences?.terminalCWD || null;
```

src/main/kubeconfig-manager.ts
```diff
@@ -24,14 +24,24 @@ export class KubeconfigManager {
...
-      await this.createProxyKubeconfig();
+      this.tempFile = await this.createProxyKubeconfig();
     } catch (err) {
       logger.error(`Failed to created temp config for auth-proxy`, { err });
     }
   }
 
-  getPath() {
+  async getPath() {
+    // create proxy kubeconfig if it is removed
```
Nokel81, line 34: So not create it if not present? Wouldn't it simplify the `init()` function above?

src/main/shell-session.ts
nevalla: I did only minimal change (`this.preferences` -> `this.cluster.preferences`) hesitated to do any additional refactorings.

src/main/kubeconfig-manager.ts
nevalla: Okay, opened the can of worms and I think I've found all the occurrences. WDYT should I keep these changes or revert to use `sync` versions?

Nokel81: I think we should keep the `async`, yes it is larger but I think it is safer to not mix async and sync like we were.

src/main/kubeconfig-manager.ts
nevalla: `tempFile` should contain path to proxy kubeconfig, so I think we want to re-create it if that's `""`.

Nokel81: exactly

src/main/kubeconfig-manager.ts
```diff
@@ -24,14 +24,24 @@ export class KubeconfigManager {
...
+    if ((this.tempFile || this.tempFile === "") && !(await fs.pathExists(this.tempFile))) {
```
Nokel81: Why so complex a check on `tempFile`? Basically you are checking if `tempFile` is `null`, `undefined`, `0`, or `false`.

nevalla: `tempFile` is either string (containing path to Kubeconfig) or `undefined`. It's `undefined` when it's not generated or it's unlinked. After it's unlinked `getPath()` should return `undefined`, at least based on tests. So those are limitations we have. What do you suggest?

Nokel81: From looking at the code `unlink` never gets called. If it can be `undefined` then the field should be marked as optional.

Finally, I would just do a `if (this.tempFile === undefined) { return undefined; }`. Though given all the uses of `getPath()` they all seem to be subtly broken in that none of them (that I can see) check for it being `undefined`.

Nokel81 (approved):
Should the handling of calling `getPath()` after `unlink` be handled in a different PR?

src/main/kubeconfig-manager.ts
nevalla: Refactored. And yes, you are right that unlink is not called anymore. This must be dropped off by some refactoring.

Nokel81:
Doing some checking, `unlink` hasn't been called since before the v3.1.0 tag

nevalla:
> Should the handling of calling `getPath()` after `unlink` be handled in a different PR?

SGTM. Should also call `unlink` properly.

jim-docker (approved):

PR #2384, opened 2021-03-24
Get rid of SidebarItem.props.subMenus in favour of props.children

ixrock:
Follow-up for https://github.com/lensapp/lens/pull/2366#pullrequestreview-619665907

### DIFF
--- a/src/renderer/components/layout/sidebar-item.tsx
+++ b/src/renderer/components/layout/sidebar-item.tsx
@@ -26,7 +26,6 @@ interface SidebarItemProps {
    * this item should be shown as active
    */
   isActive?: boolean;
-  subMenus?: React.ReactNode | React.ComponentType<SidebarItemProps>[];
 }
 
 @observer
@@ -53,11 +52,9 @@ export class SidebarItem extends React.Component<SidebarItemProps> {
   }
 
   @computed get isExpandable(): boolean {
-    if (this.compact) {
-      return false; // not available currently
-    }
+    if (this.compact) return false; // not available in compact-mode currently
 
-    return Boolean(this.props.subMenus || this.props.children);
+    return Boolean(this.props.children);
   }
 
   toggleExpand = () => {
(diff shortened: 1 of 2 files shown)


src/renderer/components/layout/sidebar-item.tsx
```diff
@@ -53,11 +52,9 @@ export class SidebarItem extends React.Component<SidebarItemProps> {
   }
 
   @computed get isExpandable(): boolean {
-    if (this.compact) {
-      return false; // not available currently
-    }
+    if (this.compact) return false; // not available in compact-mode currently
 
-    return Boolean(this.props.subMenus || this.props.children);
+    return Boolean(this.props.children);
```
Nokel81, line 57: how about:

```ts
    return !this.compact || Boolean(this.props.children);
```

src/renderer/components/layout/sidebar.tsx
```diff
@@ -51,25 +51,22 @@ export class Sidebar extends React.Component<Props> {
...
+          {crds.map((crd) => {
+            return (
+              <SidebarItem
+                key={crd.getResourceApiBase()}
+                id={`crd-resource:${crd.getResourceApiBase()}`}
+                url={crd.getResourceUrl()}
+                text={crd.getResourceTitle()}
+              />
+            );
+          })}
```
Nokel81:
```suggestion
          {crds.map(crd => (
            <SidebarItem
              key={crd.getResourceApiBase()}
              id={`crd-resource:${crd.getResourceApiBase()}`}
              url={crd.getResourceUrl()}
              text={crd.getResourceTitle()}
            />
          ))}
```

Nokel81 (commented):
Thanks

src/renderer/components/layout/sidebar-item.tsx
ixrock: I would like keep it as is for reasons:
1. current code imho more readable / intention clear
2. we could (and should) support it for `compact`-mode, so the line will be just removed.

Nokel81: okay

Nokel81: Though why did you format it at all then?

src/renderer/components/layout/sidebar.tsx
ixrock: Updated

Nokel81 (approved):

aleksfront (approved):

PR #2392, opened 2021-03-25
Allow to resize table columns

### LINKED ISSUE #601: Allow to expand/resize column headers

ixrock:
Alternative to #2146 

close #601

_Output format for saving (per `%clusterId%.json`):_

```
  "table_settings": {
    "columnSizes": {
      "network_endpoints": {
        "endpoints": 206,
        "namespace": 122
      }
    }
  }
```

_Demo:_

https://user-images.githubusercontent.com/6377066/112509984-abfe1400-8d99-11eb-89f5-a312e0417aa6.mov

### DIFF
--- a/src/renderer/components/table/table-cell.tsx
+++ b/src/renderer/components/table/table-cell.tsx
@@ -27,9 +27,9 @@ export interface TableCellProps extends React.DOMAttributes<HTMLDivElement> {
   title?: ReactNode;
   /**
    * Allow to resize width and save to local storage, default: true
-   * (applicable only with props.id)
    */
   isResizable?: boolean;
+  onResizeEnd?: () => void;
   checkbox?: boolean; // render cell with a checkbox
   isChecked?: boolean; // mark checkbox as checked or not
   renderBoolean?: boolean; // show "true" or "false" for all of the children elements are "typeof boolean"
@@ -43,7 +43,7 @@ export interface TableCellProps extends React.DOMAttributes<HTMLDivElement> {
 
 @observer
 export class TableCell extends React.Component<TableCellProps> {
-  private elem: HTMLElement;
+  @observable.ref elem?: HTMLElement;
 
   static defaultProps: TableCellProps = {
     isResizable: true,
@@ -97,7 +97,9 @@ export class TableCell extends React.Component<TableCellProps> {
   }
 
   @computed get columnSize(): number {
-    return getColumnSize(this.props.tableId, this.columnId) ?? 0;
+    const savedSize = getColumnSize(this.props.tableId, this.columnId);
+
+    return savedSize ?? this.elem?.offsetWidth;
   }
 
   @computed get isResizable(): boolean {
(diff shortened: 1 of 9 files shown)


Nokel81:
Just from the demo, I really don't like how this works. Other columns should not resize like.

ixrock:
>It's using your component `ResizingAnchor` or what do you mean?  How I can tweak it?

Right now the resizing happens through modification of the `flex-basis` CSS prop. Maybe modifying `width` property instead would produce a more satisfying result?

Nokel81:
You are resizing the `Age` column, it should also resize the `Name` column at the same time.

src/renderer/components/table/table-cell.tsx
```diff
@@ -60,24 +86,87 @@ export class TableCell extends React.Component<TableCellProps> {
...
+      this.columnId,
+    ].every(Boolean);
+  }
+
+  @computed get style(): React.CSSProperties {
+    const styles: React.CSSProperties & Record<string, any> = Object.assign({}, this.props.style);
+
+    if (this.isResizable && this.columnSize) {
+      styles.flexGrow = 0;
+      styles.flexBasis = this.columnSize;
```
Nachasic, line 119: Maybe reassigning `width` CSS property instead of `flex-basis` would produce more satisfying results?

Nachasic (commented):
Overall, I really like your implementation better than mine. Once the behavior specified by @Nokel81 is addressed, and some additional styles for resizing anchor are applied (feel free to use `.resize-anchor` class in `item-list-layout.scss` #2146 for reference), I think we should be good to go.

ixrock:
@lensapp/lens-maintainers PTAL if we want to include this in `4.x`.

Still might be not perfect in some cases (after resize), but for `5.x` we should get rid of `<Table/>` and `<VirtualList/>` in favour `react-table` or something better.

https://user-images.githubusercontent.com/6377066/112889531-6b2a3600-90de-11eb-8d01-fc7a5517b11b.mov

_Explanation for demo:_

By default all columns are flexible (`flex-grow: number` applied) and it means it might get more space after resizing a column (this is by design). When all columns within table resized == fixed width, than auto-correction won't affect column sizes.

P.S. I'm not going to invest time for this issue anymore. cc @Nachasic

Nokel81:
@ixrock Since this is targeted for 5.0.0 I think we should just work on using `react-table` or similar instead and do away with our own hand rolled versions. Thanks for tackling this 2nd attempt though.

scipe:
@Nokel81 
oh guys just wanna forward that staff: https://prnt.sc/11vj793 
so hard to see actual name.. :C 

Any way to help you to improve it?

Nokel81:
@scipe We are planning to still have this functionality. Just not by implementing it ourselves.

scipe:
I saw update regarding to that in beta version of v5. Great!! ❤️

PR #2394, opened 2021-03-25
Add support for registry.npmjs.com URLs in the extension install input

### LINKED ISSUE #1707: Install an extension from a NPM registry url
For example if user wants to install `@mirantis/lens-extension-cc` it could be done by giving an URL http://registry.npmjs.com/@mirantis/lens-extension-cc  (which returns a json object that Lens can use for resolving latest tarball URL).

Nokel81:

fixes #1707

### DIFF
--- a/src/renderer/components/+extensions/extensions.tsx
+++ b/src/renderer/components/+extensions/extensions.tsx
@@ -24,6 +24,28 @@ import { Spinner } from "../spinner/spinner";
 import { TooltipPosition } from "../tooltip";
 import { ExtensionInstallationState, ExtensionInstallationStateStore } from "./extension-install.store";
 
+function getMessageFromError(error: any): string {
+  if (!error || typeof error !== "object") {
+    return "an error has occured";
+  }
+
+  if (error.message) {
+    return String(error.message);
+  }
+
+  if (error.err) {
+    return String(error.err);
+  }
+
+  const rawMessage = String(error);
+
+  if (rawMessage === String({})) {
+    return "an error has occured";
+  }
+
+  return rawMessage;
+}
+
 interface InstallRequest {
   fileName: string;
   filePath?: string;
@@ -58,9 +80,10 @@ async function uninstallExtension(extensionId: LensExtensionId, manifest: LensEx
 
     return true;
   } catch (error) {
-    Notifications.error(
-      <p>Uninstalling extension <b>{displayName}</b> has failed: <em>{error?.message ?? error?.toString() ?? ""}</em></p>
-    );
+    const message = getMessageFromError(error);
+
+    logger.info(`[EXTENSION-UNINSTALL]: uninstalling ${displayName} has failed: ${error}`, { error });
+    Notifications.error(<p>Uninstalling extension <b>{displayName}</b> has failed: <em>{message}</em></p>);
 
     return false;
   } finally {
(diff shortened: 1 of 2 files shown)


Nokel81:
Will add a screenshot of the UI tomorrow. The general UX is Lens asks the user what version they want to install, and the versions are chosen from a dropdown in a notification.

Nokel81:
Screen Shot 2021-03-26 at 8 42 12 AM
Screen Shot 2021-03-26 at 8 42 17 AM

jakolehm:
Maybe we should just install latest version (or version that is given, `package@version`)? Version selector in a notification looks a bit odd.

Nokel81:
The URL that would be provided is `https://registry.npmjs.com/@mirantis/lens-extension-cc` which doesn't include a version.

Sure I can make it so that it auto installed the latest, I just figured that a choice might be better.

jakolehm:
> The URL that would be provided is `https://registry.npmjs.com/@mirantis/lens-extension-cc` which doesn't include a version.
> 
> Sure I can make it so that it auto installed the latest, I just figured that a choice might be better.

Choice is better but we shouldn't put too many things into a notification. I would split out the version selector to a new PR where we can discuss how it should look like.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -308,22 +303,77 @@ async function installOnDrop(files: File[]) {
   await requestInstalls(files.map(({ path }) => path));
 }
 
+async function requestInstallVersion(url: URLParse): Promise<string | null> {
```
jakolehm: Would it be easy to trigger this from a protocol handler (for example `lens://app/extension/from-url/:url`)? If yes, should we add it?

jim-docker: Could handle getting versions other than latest this way too, using a fragment for the version?

Nokel81: I would use a query param but yes that would work.

jim-docker:
This PR depends on another PR? (merging into `twice-install` branch?)

Nokel81:
@jim-docker yes, because I used some of the changes I had in #2221. Both are targeted to 4.2.1 so it shouldn't be an issue with collision (I think).

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -308,22 +303,77 @@ async function installOnDrop(files: File[]) {
...
   try {
     // install via url
     // fixme: improve error messages for non-tar-file URLs
     if (InputValidators.isUrl.validate(installPath)) {
-      const disposer = ExtensionInstallationStateStore.startPreInstall();
-      const { promise: filePromise } = downloadFile({ url: installPath, timeout: 60000 /*1m*/ });
-      const data = await filePromise;
+      const url = new URLParse(installPath);
+
+      if (url.hostname === "registry.npmjs.com") {
```
jim-docker: It would be nice to be able to just specify `@company/ext-name` and have Lens try to get it from `registry.npmjs.com` by default.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -308,22 +303,77 @@ async function installOnDrop(files: File[]) {
...
+    return null;
+  }
+
+  let selected: string;
+  const options = Object.keys(json.versions)
+    .map(version => new SemVer(version, { loose: true, includePrerelease: true }))
+    .sort((a, b) => b.compare(a))
+    .map(a => a.format());
+
+  const removeNotification = Notifications.info(
```
jim-docker: you're probably getting rid of this notification? If not when clicking the X to close the notification (cancel the install) the Install button stays dimmed and the spinner spins and spins...

src/renderer/components/+extensions/extensions.tsx
jim-docker: of course :)

Nokel81:
@jim-docker and @jakolehm PTALA, I have removed the notification and made it so that by default the latest is installed. Though it is still possible to specify a version.

Taking from Jim's suggestion I have made it so that only the name of the extension is needed, that way we can in the future swap out for our own registry without needing the users to know the difference.

PR #2398, opened 2021-03-26
Fix path traversal bug in router

jakolehm:


nevalla (dismissed):
LGTM

ixrock (dismissed):

src/main/router.ts
```diff
@@ -102,7 +108,15 @@ export class Router {
...
-    const asset = path.join(__static, filePath);
+    const asset = path.join(this.staticRootPath, filePath);
+    const normalizedFilePath = path.resolve(asset);
+
+    if (!normalizedFilePath.startsWith(this.staticRootPath)) {
+      res.statusCode = 404;
+      res.end();
+
+      return;
+    }
```
Nokel81, line 119: Should an additional check be made to make sure that none of the path segments are symlinks?

Or do we need to traverse symlinks?

nevalla (approved):

Nokel81 (approved):

PR #2415, opened 2021-03-30
Fix createStorage initializing twice

Nokel81:

This fixes current master.

### DIFF
--- a/src/renderer/utils/storageHelper.ts
+++ b/src/renderer/utils/storageHelper.ts
@@ -5,6 +5,7 @@ import { action, comparer, observable, toJS, when } from "mobx";
 import produce, { Draft, enableMapSet, setAutoFreeze } from "immer";
 import { isEqual, isFunction, isPlainObject } from "lodash";
 import logger from "../../main/logger";
+import { IObservableValue } from "mobx/lib/internal";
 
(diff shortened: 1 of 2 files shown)


src/renderer/utils/storageHelper.ts
```diff
@@ -5,6 +5,7 @@ import { action, comparer, observable, toJS, when } from "mobx";
 import produce, { Draft, enableMapSet, setAutoFreeze } from "immer";
 import { isEqual, isFunction, isPlainObject } from "lodash";
 import logger from "../../main/logger";
+import { IObservableValue } from "mobx/lib/internal";
```
ixrock, line 8: This is available to import from `mobx`

src/renderer/utils/storageHelper.ts
```diff
@@ -33,22 +34,27 @@ export class StorageHelper<T> {
     }
   };
 
-  @observable private data = observable.box<T>();
+  private data: IObservableValue<T>;
```
ixrock, line 37: Why this change is need?

ixrock (commented):
Can't get why so many changes to `storageHelper.ts`. 
Yesterday we agreed that the issue in `dock-tab.store.ts` only. Default value is set in `reset()` on init.

image

PR #2417, opened 2021-03-31
Refactor StorageHelper to be clearer in its constructor

Nokel81:
- Don't create a box and then override it

- Initial value should be the default value

@ixrock This is a simpler form of the fix I think.

### DIFF
--- a/src/renderer/utils/storageHelper.ts
+++ b/src/renderer/utils/storageHelper.ts
@@ -59,14 +59,20 @@ export class StorageHelper<T> {
     }
   }
 
-  /**
-   * This function turns an optionally synchronous call to an asynchronous call
-   * and correctly handles the synchronous call throwing an error
-   * @returns the result of this.storage.getItem
-   */
-  private async getItemFromStorage(): Promise<T> {
-    return this.storage.getItem(this.key);
-  }
+  private onData = (data: T): void => {
+    const notEmpty = data != null;
+    const notDefault = !this.isDefaultValue(data);
+
+    if (notEmpty && notDefault) {
+      this.merge(data);
+    }
+
+    this.initialized = true;
+  };
+
+  private onError = (error: any): void => {
+    logger.error(`[load]: ${error}`, this);
+  };
 
   @action
   init({ force = false } = {}) {
@@ -74,20 +80,17 @@ export class StorageHelper<T> {
       return;
     }
 
-    this.getItemFromStorage()
-      .then(data => {
-        const notEmpty = data != null;
-        const notDefault = !this.isDefaultValue(data);
+    try {
+      const data = this.storage.getItem(this.key);
 
-        if (notEmpty && notDefault) {
-          this.merge(data);
-        }
-
-        this.initialized = true;
-      })
-      .catch(error => {
-        logger.error(`[load]: ${error}`, this);
-      });
+      if (data instanceof Promise) {
+        data.then(this.onData, this.onError);
+      } else {
+        this.onData(data);
+      }
+    } catch (error) {
+      this.onError(error);
+    }
   }
 
   isDefaultValue(value: T): boolean {
--- a/src/renderer/utils/storageHelper.ts
+++ b/src/renderer/utils/storageHelper.ts
@@ -1,7 +1,7 @@
 // Helper for working with storages (e.g. window.localStorage, NodeJS/file-system, etc.)
 
 import type { CreateObservableOptions } from "mobx/lib/api/observable";
-import { action, comparer, observable, toJS, when } from "mobx";
+import { action, comparer, observable, toJS, when, IObservableValue } from "mobx";
 import produce, { Draft, enableMapSet, setAutoFreeze } from "immer";
 import { isEqual, isFunction, isPlainObject } from "lodash";
 import logger from "../../main/logger";
@@ -33,7 +33,7 @@ export class StorageHelper<T> {
     }
   };
 
-  @observable private data = observable.box<T>();
+  private data: IObservableValue<T>;
   @observable initialized = false;
   whenReady = when(() => this.initialized);
 
@@ -101,7 +101,7 @@ export class StorageHelper<T> {
 
   @action
   private configureObservable(options = this.options.observable) {
-    this.data = observable.box<T>(this.data.get(), {
+    this.data = observable.box<T>(this.options.defaultValue, {
       ...StorageHelper.defaultOptions.observable, // inherit default observability options
       ...(options ?? {}),
     });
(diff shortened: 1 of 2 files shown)


src/renderer/utils/storageHelper.ts
```diff
@@ -21,7 +21,7 @@ export interface StorageHelperOptions<T> {
   autoInit?: boolean; // start preloading data immediately, default: true
   observable?: CreateObservableOptions;
   storage: StorageAdapter<T>;
-  defaultValue?: T;
+  defaultValue: T;
```
ixrock: Why it should be required field? Why `defaultValue` cannot be `undefined`?

src/renderer/utils/storageHelper.ts
```diff
@@ -101,7 +101,7 @@ export class StorageHelper<T> {
 
   @action
   private configureObservable(options = this.options.observable) {
-    this.data = observable.box<T>(this.data.get(), {
+    this.data = observable.box<T>(this.options.defaultValue, {
```
ixrock: This looks really wrong imho.
Default value should be set only in `constructor()` and in `reset()`.

Nokel81: I did this because you didn't want me changing `StorageHelper` too much. I would have gotten rid of this function completely otherwise.

src/renderer/utils/storageHelper.ts
Nokel81: Because `undefined` is not a valid `number`.

src/renderer/utils/storageHelper.ts
ixrock:
```
constructor:
    this.configureObservable(); // create observable field
    this.reset(); // default value is set here
```

What is wrong here? Why default value not set properly you think?

ixrock: Otherwise I would better to update it like `configureObservable({value?, options?})` and pass `defaultValue` in constructor.

src/renderer/utils/storageHelper.ts
ixrock: What? Why `number`? It could be `any` (json-serializable)

Nokel81: Typescript does not support conditional optionality. But I have make a `StorageHelper` for `number`'s then we should have a defined default value.

ixrock: Didn't get what you mean about conditional optionality..

When you store a number you should define a `number` as `defaultValue` in place of use if it's cannot be `undefined` initially for your logic.

ixrock: Summoning @jakolehm to the thread. Do you have some thoughts on this?

aleksfront:
@Nokel81 Can you please post a brief description why this change needed? There's no link to issue or reference to a bug.

Nokel81:
@aleksfront This is trying to fix master where a test is failing on linux.

src/renderer/utils/storageHelper.ts
jim-docker: I remember being confused the first time I reviewed this function. It is (only) called by the constructor. So it is setting the initial value of `this.data`. It passes `this.data.get()` to the `observable.box<T>` constructor?? Maybe you intend that `configureObservable()` be called outside the constructor as well sometime(?) but seems weird for the constructor to use it.
Why not call `this.reset()` before `this.configureObservable()` in the constructor?

ixrock: We can remove (move) this function to constructor. It's organized in separated method for readability initially.
And yes, calling `this.reset()` first also make sense if it will help.

Nokel81: If we call reset first, why are we "initializing" it twice anyway?

ixrock: Is it matters? 
The "first init" is more like a definition without any data behind and changing it to `data: IObservableValue` does nothing in general (except saving some cpu maybe).

Have a  look:

```ts
class Test {
  test = console.log("out");
  
  constructor() {
    this.test = console.log("in");
  }
}

new Test(); // prints "out" first 
```

And it means `reset()` inside constructor would correct set default value. 
Or I missing some deeps of TS knowledge?

Nokel81: Yeah I know that it will just be overwritten, but from a maintenance perspective we shouldn't leave these sorts of things around because the next reader will have to figure out if we actually want to use that inline declared value (we don't).

jim-docker: I'm sure I am :)

Nokel81:
So after even more debugging I have come to the conclusion that even though I think this is a worth while refactor, the actual solution is to change `src/renderer/utils/storageHelper.ts:132` to be `return this.data.get() ?? this.defaultValue;`.

src/renderer/utils/storageHelper.ts
ixrock: Ok, let's just get rid of `configureObservable` and put contents inside constructor.
That way init `this.options.defaultValue` make sense (if it's not locked in the method).

Nokel81:
Changed this PR into a refactor, and opened #2421 to actually fix the bug.

ixrock (approved):
LGTM

jim-docker (dismissed):

src/renderer/utils/storageHelper.ts
```diff
@@ -21,46 +21,61 @@ export interface StorageHelperOptions<T> {
...
       this.init();
     }
   }
 
+  /**
+   * This function turns an optionally synchronous call to an asynchronous call
+   * and correctly handles the synchronous call throwing an error
+   * @returns the result of this.storage.getItem
+   */
+  private async getItemFromStorage(): Promise<T> {
```
ixrock: Btw, I guess this is not correct at all and breaks work with `sync` reading from storage.

See a test in `storageHerlper.test.ts`:

```ts
    it("initialized with default value", async () => {
      localStorage.setItem(storageKey, "saved"); // pretending it was saved previously

      expect(storageHelper.key).toBe(storageKey);
      expect(storageHelper.defaultValue).toBe("test");
      expect(storageHelper.get()).toBe("test");

      // here usage of `await` is wrong and if you remove it test will fail which is not happening in previous version
      // for that reason was introduced cumbersome `loadFromStorage()` which is handle it correctly.
      await storageHelper.init();

      expect(storageHelper.key).toBe(storageKey);
      expect(storageHelper.defaultValue).toBe("test");
      expect(storageHelper.get()).toBe("saved");
    });
```

ixrock (changes requested):

Nokel81:
Will look into this thanks.

ixrock (approved):

jim-docker (approved):

PR #2418, opened 2021-03-31
Catalog & Hotbar - initial groundwork

jakolehm:
## TODO

* [x] Remove workspace
* [x] Introduce Catalog api & expose it via extension API
* [x] Replace workspace overview with Catalog
* [x] Dynamic context menu for Catalog entities
* [x] Dynamic context menu for Hotbar items
* [x] Replace cluster menu with Hotbar
* [x] Remove Cluster & Workspace extension apis
* [x] Introduce Catalog extension apis
* [x] Refactor command palette to use "entity" scope
* [x] Fix tests

## Out-of-scope (should be fixed in follow-up PRs)

* UI fine tune
* multiple hotbars
* documentation changes


src/common/catalog-category-registry.ts
```diff
@@ -0,0 +1,56 @@
+import { action, computed, observable, toJS } from "mobx";
+import { CatalogCategory, CatalogEntityData } from "./catalog-entity";
+
+export class CatalogCategoryRegistry {
+  @observable protected categories: CatalogCategory[] = [];
```
ixrock, line 5: If you define it as:

```ts
protected categories = observable.array<CatalogCategory>();
```

you can use extended observable-array api to remove item below as (if those are not cloned objects on removal):

```
this.categories.remove(category);
```

src/common/catalog-entity-registry.ts
```diff
@@ -0,0 +1,32 @@
+import { action, computed, observable } from "mobx";
+import { CatalogEntity } from "./catalog-entity";
+
+export class CatalogEntityRegistry {
+  @observable.deep protected sources: Map<string, CatalogEntity[]> = observable.map(new Map(), { deep: true });
```
ixrock:
```suggestion
  protected sources = observable.map<string, CatalogEntity[]>([], { deep: true });
```

src/common/hotbar-store.ts
```diff
@@ -0,0 +1,67 @@
...
+      accessPropertiesByDotNotation: false, // To make dots safe in cluster context names
+      syncOptions: {
+        equals: comparer.structural,
+      },
+      migrations,
+    });
+  }
+
+  @action protected async fromStore(data: Partial<HotbarStoreModel> = {}) {
+    this.hotbars = data.hotbars || [{
```
ixrock, line 38: If incoming `data` is partial is it correct to overwrite it fully?

src/extensions/registries/command-registry.ts
```diff
@@ -34,4 +35,16 @@ export class CommandRegistry extends BaseRegistry<CommandRegistration> {
   }
 }
 
+export function syncCommandRegistryWithCatalog() {
```
ixrock: Not used anywhere currently.

src/main/catalog-pusher.ts
```diff
@@ -0,0 +1,26 @@
...
+import { broadcastMessage, subscribeToBroadcast } from "../common/ipc";
+import { CatalogEntityRegistry} from "../common/catalog-entity-registry";
+import "../common/catalog-entities/kubernetes-cluster";
+
+export class CatalogPusher {
+  static init(catalog: CatalogEntityRegistry) {
+    new CatalogPusher(catalog).init();
+  }
+
+  constructor(private catalog: CatalogEntityRegistry) {}
```
ixrock: Should be a `private` then if static `init()` is entry point?
Also `init()` probably should return disposer from `autorun` to make clean up at some point?

src/main/cluster-manager.ts
```diff
@@ -1,16 +1,25 @@
...
 import { apiKubePrefix } from "../common/vars";
 import { Singleton } from "../common/utils";
+import { CatalogEntity } from "../common/catalog-entity";
+import { KubernetesCluster } from "../common/catalog-entities/kubernetes-cluster";
+import { catalogEntityRegistry } from "../common/catalog-entity-registry";
+
+const clusterOwnerRef = "ClusterManager";
 
 export class ClusterManager extends Singleton {
+  @observable.deep catalogSource: CatalogEntity[] = [];
```
ixrock, line 17: `deep` is a default modifier for `@observable` (just in case)

src/renderer/components/+catalog/catalog.scss
```diff
@@ -0,0 +1,23 @@
...
+  .content-wrapper {
+    .content {
+      margin: unset;
+      max-width: unset;
+      height: 100%;
+    }
+  }
+}
+
+.TableCell.status {
```
ixrock: Better to move inside some context (e.g. `.CatalogPage > .TableCell`), otherwise it's a global style for whole app.

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -0,0 +1,141 @@
...
+    if (this.catalogEntityStore.items.length === 0) {
+      Notifications.info(<><b>Welcome!</b><p>Get started by associating one or more clusters to Lens</p></>, {
+        timeout: 30_000,
+        id: "catalog-welcome"
+      });
+    }
+  }
+
+  componentWillUnmount() {
+    this.disposers.forEach((d) => d());
```
ixrock: Maybe better to use `disposeOnUnmount` from `mobx-react` to reduce some boilerplate (if no other special purposes for `this.disposers`)

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -0,0 +1,141 @@
...
+      <MenuActions onOpen={() => onOpen()}>
+        <MenuItem key="add-to-hotbar" onClick={() => this.addToHotbar(item) }>
+          <Icon material="add" interactive={true} title="Add to hotbar"/> Add to Hotbar
+        </MenuItem>
+        <MenuItem key="remove-from-hotbar" onClick={() => this.removeFromHotbar(item) }>
+          <Icon material="clear" interactive={true} title="Remove from hotbar"/> Remove from Hotbar
+        </MenuItem>
+        { this.contextMenu.menuItems.map((menuItem) => {
+          return (
+            <MenuItem key={menuItem.title} onClick={() => menuItem.onClick()}>
```
ixrock: I guess `menuItem.title` might be not uniq key? It will lead to `react` errors/warnings.

ixrock (commented):
Looking good! But needs a bit fine-tuning.

Screenshot 2021-04-07 at 13 35 22
Screenshot 2021-04-07 at 13 38 41
Screenshot 2021-04-07 at 13 43 27

_Disconnect from different places is different & flaky_

https://user-images.githubusercontent.com/6377066/113861805-49047800-97b0-11eb-868b-f9835dae2805.mov

src/main/cluster.ts
```diff
@@ -612,7 +605,6 @@ export class Cluster implements ClusterModel, ClusterState {
       id: this.id,
       contextName: this.contextName,
       kubeConfigPath: this.kubeConfigPath,
-      workspace: this.workspace,
```
Nokel81: Won't this remove the workspaces from the store file currently, can we leave this as deprecated for the time being?

src/main/tray.ts
```diff
@@ -78,28 +75,6 @@ function createTrayMenu(windowManager: WindowManager): Menu {
           .catch(error => logger.error(`${TRAY_LOG_PREFIX}: Failed to nativate to Preferences`, { error }));
       },
     },
-    {
```
Nokel81, line 81: Why do we want to remove this?

jakolehm: Because we don't have workspaces/clusters anymore... we can add replacement if/when we know what should be here.

src/main/cluster.ts
jakolehm: I'll add it back.

src/common/hotbar-store.ts
jakolehm: This only overwrites `hotbars` array ... what would be better syntax for "optional" data 🤔

src/main/catalog-pusher.ts
jakolehm: Improved.

src/main/tray.ts
Nokel81: Oh okay

src/renderer/components/+catalog/catalog.scss
jakolehm: Fixed.

src/common/catalog-category-registry.ts
jakolehm: Good to know but I'm not sure if we have exactly the same instance on removal.

src/common/catalog-entity-registry.ts
jakolehm: Thanks!

src/renderer/components/+catalog/catalog.tsx
jakolehm: Yes, fixed.

src/renderer/components/+catalog/catalog.tsx
jakolehm: Yes, fixed.

jakolehm:
Added catalog to bottom bar:
image

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -83,9 +83,9 @@ export class Catalog extends React.Component {
         <MenuItem key="remove-from-hotbar" onClick={() => this.removeFromHotbar(item) }>
           <Icon material="clear" interactive={true} title="Remove from hotbar"/> Remove from Hotbar
         </MenuItem>
-        { this.contextMenu.menuItems.map((menuItem) => {
+        { this.contextMenu.menuItems.map((menuItem, index) => {
           return (
-            <MenuItem key={menuItem.title} onClick={() => menuItem.onClick()}>
+            <MenuItem key={`menuitem-${index}`} onClick={() => menuItem.onClick()}>
```
Nokel81: Why switch this, also this is no different that using the index as the key. Is there a reason to do this formatting on every render?

src/common/hotbar-store.ts
Nokel81: Do you mean all of `HotbarStoreModel` is normal except for `hotbars` which is optional?

I also don't exactly understand what Roman is getting at.

Nokel81: Also to add, why do we need a default hotbar? I thought hotbars were optional to use.

extensions/metrics-cluster-feature/renderer.tsx
```diff
@@ -1,21 +1,55 @@
...
-      },
-      feature: new MetricsFeature()
+  onActivate() {
+    const category = Store.catalogCategories.getForGroupKind<Store.KubernetesClusterCategory>("entity.k8slens.dev", "KubernetesCluster");
+
+    if (!category) {
+      return;
+    }
+
+    category.on("contextMenuOpen", this.clusterContextMenuOpen.bind(this));
```
jim-docker, line 12: This is a mouthful. Maybe eventually we can add a helper method to the `LensExtension` class so it simplifies to something like
```
this->CategoryOn("KubernetesCluster", "contextMenuOpen", this.clusterContextMenuOpen);
```

Nokel81: `->`?!?!

Is your c++ memories leaking \s.

jim-docker: Oops. My memory is leaky in general. And all languages are looking more and more the same.

jakolehm:
Added list of categories to Catalog view:
image

src/renderer/components/+catalog/catalog.tsx
jakolehm: No reason, changed it to use plain `index`.

extensions/metrics-cluster-feature/renderer.tsx
```diff
@@ -1,21 +1,55 @@
...
-          <span>
-            Enable timeseries data visualization (Prometheus stack) for your cluster.
-            Install this only if you don&apos;t have existing Prometheus stack installed.
-            You can see preview of manifests <a href="https://github.com/lensapp/lens/tree/master/extensions/metrics-cluster-feature/resources" rel="noreferrer" target="_blank">here</a>.
-          </span>
-        )
-      },
-      feature: new MetricsFeature()
+  onActivate() {
+    const category = Store.catalogCategories.getForGroupKind<Store.KubernetesClusterCategory>("entity.k8slens.dev", "KubernetesCluster");
```
Nokel81, line 6: While I agree with Jim that this should be simpler. I think we should allow for some API like:

(also, these aren't stores they shouldn't be called such).

```ts
Catalog.byKind("dev.k8slens.entity.KubernetesCluster")
	.on("contextMenuOpen", this.clusterContextMenuOpen.bind(this));
```

Namely, I am thinking that this should "always" work.

1. I don't see the benefit of having separate arguments for the get category request. We can just embed that information in one string.
2. Lens should be defensive here, especially with regards to ordering. If we have two clusters, one that specifies are kind of catalog entry, and the other that consumes it. Lens itself should do the synchronization. Therefore, no matter what the argument to `byKind` is, it should return a valid object which then late binds whenever the future item gets registered. 
3. If after some time that item still isn't registered, we could log a warning to the devtools but that should be for the future.

src/common/catalog-entity.ts
```diff
@@ -0,0 +1,75 @@
...
+  labels: {
+    [key: string]: string;
+  }
+  [key: string]: string | object;
+};
+
+export type CatalogEntityStatus = {
+  phase: string;
+  reason?: string;
+  message?: string;
```
Nokel81, line 36: What is the difference between these two?

src/common/catalog-entity.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+  }
+  getId: () => string;
+}
+
+export type CatalogEntityMetadata = {
+  uid: string;
+  name: string;
+  description?: string;
+  source?: string;
```
Nokel81, line 26: Shouldn't source be the same apiVersion + kind? And we disallow extensions to use `dev.k8slens.entity`?

src/common/catalog-category-registry.ts
Nokel81, line 5: Shouldn't this be a map, given how we are searching through it bellow? I think this is more important given how often we will be doing those actions.

```typescript
protected catagories = observable.map<string, observable.Map<string, CatalogCategory[]>>();
```

src/common/catalog-category-registry.ts
```diff
@@ -0,0 +1,56 @@
...
+    this.categories.push(category);
+  }
+
+  @action remove(category: CatalogCategory) {
+    this.categories = this.categories.filter((cat) => cat.apiVersion !== category.apiVersion && cat.kind !== category.kind);
+  }
+
+  @computed get items() {
+    return toJS(this.categories);
+  }
```
Nokel81, line 17: This seems weird, why do we want to force the `toJS` here, but then also cache its value?

src/common/catalog-category-registry.ts
Nokel81: Maybe also have version as another level of the mapping too.

src/common/catalog-category-registry.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  getForGroupKind<T extends CatalogCategory>(group: string, kind: string) {
+    return this.categories.find((c) => c.spec.group === group && c.spec.names.kind === kind) as T;
+  }
+
+  getEntityForData(data: CatalogEntityData) {
```
Nokel81, line 23: Like I said above, I think we shouldn't ever return `null` from these sorts of things. We should have the the logic here to work with asynchronous adding of new catalogues types.

src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  async onRun(context: CatalogEntityActionContext) {
+    context.navigate(`/cluster/${this.metadata.uid}`);
+  }
+
+  async onDetailsOpen() {
+    //
+  }
+
+  async onContextMenuOpen(context: CatalogEntityContextMenuContext) {
```
Nokel81, line 47: Given that this doesn't make sense to be called on `main`. We shouldn't have this as part of `common`.

src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -0,0 +1,105 @@
...
+      },
+      {
+        icon: "delete",
+        title: "Delete",
+        onClick: async () => clusterStore.removeById(this.metadata.uid),
+        confirm: {
+          message: `Remove Kubernetes Cluster "${this.metadata.name} from Lens?`
+        }
+      },
+    ];
```
Nokel81, line 62: Maybe I am misunderstanding what is happening here. But won't this override what other extensions add too? Or is this the default values?

src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -0,0 +1,105 @@
...
+        title: "Settings",
+        onClick: async () => context.navigate(`/cluster/${this.metadata.uid}/settings`)
+      },
+      {
+        icon: "delete",
+        title: "Delete",
+        onClick: async () => clusterStore.removeById(this.metadata.uid),
+        confirm: {
+          message: `Remove Kubernetes Cluster "${this.metadata.name} from Lens?`
+        }
```
Nokel81, line 60: Also is this temporary, because this really should just emit an event which the source should detect, or maybe call the sources handle. This seems to imply that all `KubernetesCluster`'s will be from Lens.

src/common/catalog-entity-registry.ts
```diff
@@ -0,0 +1,32 @@
+import { action, computed, observable } from "mobx";
+import { CatalogEntity } from "./catalog-entity";
+
+export class CatalogEntityRegistry {
+  protected sources = observable.map<string, CatalogEntity[]>([], { deep: true });
+
+  @action addSource(id: string, source: CatalogEntity[]) {
```
Nokel81, line 7: I think we should require that the sources are observable, or can we not do that?

src/common/catalog-entity-registry.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  @computed get items() {
+    const catalogItems: CatalogEntity[] = [];
+
+    for (const items of this.sources.values()) {
+      items.forEach((item) => catalogItems.push(item));
+    }
+
+    return catalogItems;
```
Nokel81, line 22:
```suggestion
    return Array.from(this.sources.values()).flat();
```

src/common/cluster-store.ts
```diff
@@ -235,22 +234,6 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
     }
   }
 
-  @action
-  swapIconOrders(workspace: WorkspaceId, from: number, to: number) {
```
Nokel81, line 239: Maybe I cannot find it, but we should keep swapping hotbar entries.

src/common/hotbar-store.ts
Nokel81, line 38: I think we should default to no hotbars, and have an easy "grab bar" to add one. That way we don't need to special case the default one.

src/common/hotbar-store.ts
```diff
@@ -0,0 +1,67 @@
...
+  name: string;
+  items: HotbarItem[];
+}
+
+export interface HotbarStoreModel {
+  hotbars: Hotbar[];
+}
+
+export class HotbarStore extends BaseStore<HotbarStoreModel> {
+  @observable hotbars: Hotbar[] = [];
```
Nokel81, line 24: I think this should be a map from IDs to hotbars.

src/common/hotbar-store.ts
```diff
@@ -0,0 +1,67 @@
...
+import { BaseStore } from "./base-store";
+import migrations from "../migrations/hotbar-store";
+
+export interface HotbarItem {
+  entity: {
+    uid: string;
+  };
+  params?: {
+    [key: string]: string;
+  }
```
Nokel81, line 11: I don't think we should care what the param types are.

```suggestion
  params?: Record<string, any>; 
```

src/extensions/core-api/stores.ts
```diff
@@ -1,8 +1,4 @@
...
-
-export { clusterStore, Cluster, ClusterStore } from "../stores/cluster-store";
-export type { ClusterModel, ClusterId } from "../stores/cluster-store";
-
-export { workspaceStore, Workspace, WorkspaceStore } from "../stores/workspace-store";
-export type { WorkspaceId, WorkspaceModel } from "../stores/workspace-store";
-
+export { KubernetesCluster, KubernetesClusterCategory } from "../../common/catalog-entities/kubernetes-cluster";
+export { catalogCategoryRegistry as catalogCategories } from "../../common/catalog-category-registry";
+export { catalogEntities } from "./catalog";
```
Nokel81, line 4: Catalogue isn't a store, it is completely on demand.

src/extensions/registries/command-registry.ts
```diff
@@ -1,25 +1,25 @@
...
   id: string;
   title: string;
-  scope: "cluster" | "global";
+  scope: "entity" | "global";
   action: (context: CommandContext) => void;
   isActive?: (context: CommandContext) => boolean;
 }
 
 export class CommandRegistry extends BaseRegistry<CommandRegistration> {
+  @observable activeEntity: CatalogEntity;
```
Nokel81, line 21: Why do we have this hear? It should be gotten on demand.

src/main/catalog-pusher.ts
```diff
@@ -0,0 +1,32 @@
...
+      this.broadcast();
+    });
+
+    disposers.push(() => unsubscribeFromBroadcast("catalog:broadcast", listener));
+
+    return disposers;
+  }
+
+  broadcast() {
+    broadcastMessage("catalog:items", toJS(this.catalog.items, { recurseEverything: true }));
```
Nokel81, line 30: We should remember to document that catalogue entities *must* be IPC serializable.

src/main/cluster-manager.ts
```diff
@@ -19,8 +28,18 @@ export class ClusterManager extends Singleton {
           cluster.init(port);
         }
       });
+
     }, { fireImmediately: true });
 
+    reaction(() => toJS(clusterStore.enabledClustersList, { recurseEverything: true }), () => {
+      this.updateCatalogSource(clusterStore.enabledClustersList);
```
Nokel81, line 35: Shouldn't we be removing the concept of enabled clusters anyway?
```suggestion
      this.updateCatalogSource(clusterStore.clustersList);
```

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -0,0 +1,61 @@
+import { action, observable } from "mobx";
+import { broadcastMessage, subscribeToBroadcast } from "../../common/ipc";
+import { CatalogCategory, CatalogEntity, CatalogEntityData } from "../../common/catalog-entity";
+import { catalogCategoryRegistry, CatalogCategoryRegistry } from "../../common/catalog-category-registry";
+import "../../common/catalog-entities";
+
+export class CatalogEntityRegistry {
+  @observable protected _items: CatalogEntity[] = observable.array([], { deep: true });
```
Nokel81, line 8: You don't need both the annotation and the call, plus I am pretty sure `deep` is the default.

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  constructor(private categoryRegistry: CatalogCategoryRegistry) {}
```
Nokel81, line 10: Why is the entity registry getting passed the category registry. I figured they would be singletons.

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  init() {
+    subscribeToBroadcast("catalog:items", (ev, items: CatalogEntityData[]) => {
+      this.updateItems(items);
+    });
+    broadcastMessage("catalog:broadcast");
```
Nokel81, line 16:
```suggestion
    broadcastMessage("catalog:initialized");
```

Nokel81: Broadcast is a very "broad" term.

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  @action updateItems(items: CatalogEntityData[]) {
```
Nokel81, line 19: This seems contrary to the ethos of having multiple sources. Shouldn't the registry just combind the arrays that it is given, and let the sources deal with editing?

Nokel81: Plus we really shouldn't care about ordering here. That should be a display problem and we should have a uniform sorting for display.

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -0,0 +1,61 @@
...
+        this._items.push(item);
+      } else {
+        this._items.splice(index, 1, item);
+      }
+    });
+  }
+
+  get items() {
+    return this._items;
+  }
```
Nokel81, line 45: Is there a reason why we don't just have items be public?

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    return items as T[];
+  }
+
+  getItemsForCategory<T extends CatalogEntity>(category: CatalogCategory): T[] {
+    const supportedVersions = category.spec.versions.map((v) => `${category.spec.group}/${v.name}`);
+    const items = this._items.filter((item) => supportedVersions.includes(item.apiVersion) && item.kind === category.spec.names.kind);
+
+    return items as T[];
+  }
```
Nokel81, line 58: I also think that this implies it should be a map of some kind.

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export const catalogEntityRegistry = new CatalogEntityRegistry(catalogCategoryRegistry);
```
Nokel81, line 61: This should be a singleton, see my singleton PR.

src/renderer/components/+catalog/catalog-entity.store.ts
```diff
@@ -0,0 +1,81 @@
...
+  get labels() {
+    const labels: string[] = [];
+
+    Object.keys(this.entity.metadata.labels).forEach((key) => {
+      const value = this.entity.metadata.labels[key];
+
+      labels.push(`${key}=${value}`);
+    });
+
+    return labels;
```
Nokel81, line 40:
```suggestion
    return Object.entries(this.entity.metadata.labels)
      .map(([key, value]) => `${key}=${value}`);
```

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -0,0 +1,195 @@
...
+  @observable private catalogEntityStore?: CatalogEntityStore;
+  @observable.deep private contextMenu: CatalogEntityContextMenuContext;
+  @observable activeTab: string;
+
+  async componentDidMount() {
+    this.contextMenu = {
+      menuItems: [],
+      navigate: (url: string) => navigate(url)
+    };
+    this.catalogEntityStore = new CatalogEntityStore();
```
Nokel81, line 38: Why are we recreating the entity store?

src/renderer/components/+catalog/catalog-entity.store.ts
```diff
@@ -0,0 +1,81 @@
...
+      reaction(() => this.entities, () => this.loadAll()),
+      reaction(() => this.activeCategory, () => this.loadAll(), { delay: 100})
+    ];
+
+    return () => disposers.forEach((dispose) => dispose());
+  }
+
+  loadAll() {
+    return this.loadItems(() => this.entities);
+  }
```
Nokel81, line 80: What is this form? Working with the current version of `ItemListLayout`?

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  addToHotbar(item: CatalogEntityItem) {
+    const hotbar = hotbarStore.getByName("default"); // FIXME
+
+    if (!hotbar) {
+      return;
+    }
+
+    hotbar.items.push({ entity: { uid: item.id }});
```
Nokel81, line 66: Why is this in catalogue? This seems like it should be part of the hotbarStore. 

Though TBH it should be something like "add to current hotbar" or even better the user should just drag the item over to the hotbar.

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  get categories() {
+    return catalogCategoryRegistry.items;
+  }
+
+  onTabChange = (tabId: string) => {
+    this.activeTab = tabId;
+
+    const activeCategory = this.categories.find((category) => category.getId() === tabId);
```
Nokel81, line 107: Shouldn't this mean that `categories` should be a map?

src/renderer/components/hotbar/hotbar-icon.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+    if (splittedName.length === 1) {
+      splittedName = splittedName[0].split("-");
+    }
+
+    if (splittedName.length === 1) {
+      splittedName = splittedName[0].split("@");
+    }
+
+    splittedName = splittedName.map((part) => part.replace(/\W/g, ""));
```
Nokel81, line 47: See #2331 but I really don't think we should be ignoring non-english characters.

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -0,0 +1,195 @@
...
+        },
+        ok: () => {
+          menuItem.onClick();
+        },
+        message: menuItem.confirm.message
+      });
+    } else {
+      menuItem.onClick();
+    }
+  }
```
Nokel81, line 98: This is duplicated with the hotbar store. Should be a free function.

src/renderer/components/menu/menu-actions.tsx
```diff
@@ -70,6 +71,10 @@ export class MenuActions extends React.Component<MenuActionsProps> {
       ...(typeof triggerIcon === "object" ? triggerIcon : {}),
     };
 
+    if (this.props.onOpen) {
+      iconProps.onClick = this.props.onOpen;
+    }
+
```
Nokel81, line 77: If this is undefined it should still be fine and `onOpen` can be destructured as on line 58.

```suggestion
      onClick: onOpen,
      ...(typeof triggerIcon === "object" ? triggerIcon : {}),
    };
```

src/renderer/protocol-handler/app-handlers.ts
```diff
@@ -16,18 +15,10 @@ export function bindProtocolAddRouteHandlers() {
...
-    })
-    .addInternalHandler("/landing/:workspaceId", ({ pathname: { workspaceId } }) => {
-      if (workspaceStore.getById(workspaceId)) {
-        workspaceStore.setActive(workspaceId);
-        navigate(landingURL());
-      } else {
-        console.log("[APP-HANDLER]: workspace with given ID does not exist", { workspaceId });
-      }
+    .addInternalHandler("/catalog", () => {
+      navigate(catalogURL());
```
Nokel81: I still think we should keep `/landing` for backward comparability purposes. Plus it is the landing page, so we don't really need `/catalog`

src/renderer/components/layout/main-layout-header.tsx
```diff
@@ -1,39 +1,17 @@
...
 
 interface Props {
   cluster: Cluster
   className?: string
 }
 
 export const MainLayoutHeader = observer(({ cluster, className }: Props) => {
-  const actions = ClusterActions(cluster);
-  const renderMenu = () => (
-    <MenuActions autoCloseOnSelect className="ClusterActionsMenu">
```
Nokel81, line 17: Why do we no longer want this, to specific to Kube Clusters?

src/renderer/components/hotbar/hotbar-menu.tsx
```diff
@@ -0,0 +1,46 @@
...
+                index={index}
+                entity={entity}
+                isActive={entity.status.active}
+                onClick={() => entity.onRun(catalogEntityRunContext)}
+              />
+            );
+          })}
+        </div>
+      </div>
+    );
```
Nokel81, line 43: I think we should have rearrange-able of these. Though that might be a follow up PR.

src/renderer/components/hotbar/hotbar-icon.tsx
```diff
@@ -0,0 +1,132 @@
...
+          toggleEvent="contextmenu"
+          position={{right: true, bottom: true }} // FIXME: position does not work
+          open={() => onOpen()}
+          close={() => this.toggleMenu()}>
+          <MenuItem key="remove-from-hotbar" onClick={() => this.removeFromHotbar(entity) }>
+            <Icon material="clear" small interactive={true} title="Remove from hotbar"/> Remove from Hotbar
+          </MenuItem>
+          { this.contextMenu && this.contextMenu.menuItems.map((menuItem) => {
+            return (
+              <MenuItem key={menuItem.title} onClick={() => this.onMenuItemClick(menuItem) }>
```
Nokel81, line 122: Given that these items are tide to the UI. We should be able to utilize mobx caching and have the context menu get the values on demand such that we can have the store be told to do "execute action number 1".

Nokel81 (dismissed):
Looks good. 

However since we might have hotbar optionally moveable, I think we should take a look at not have the top bar of the App ever have content in it.

That way when the hotbar is on the right hand side we can have the full page go all the way to the left underneath the window action buttons (from the OS).

Nokel81:
Though I didn't mean to approve the PR

src/common/catalog-entity.ts
jakolehm: Source is not same as apiVersion + kind. Example: catalog has kubernetes clusters discovered from local filesystem +  kubernetes clusters coming from extensions (remote apis).

src/renderer/components/layout/main-layout-header.tsx
jakolehm: I'm not sure if we need it anymore now that we already have menu in catalog & hotbar items. If we want to bring this back then it should be dynamic like those other menus.

src/renderer/components/hotbar/hotbar-menu.tsx
jakolehm: Yes we should allow drag&drop (even from catalog to hotbar).. but let's do those in follow-up PRs.

jakolehm:
> However since we might have hotbar optionally moveable, I think we should take a look at not have the top bar of the App ever have content in it.

@Nokel81 yes we might have but it's out of scope for this PR (and probably out of scope for 5.0.0).

src/common/catalog-entity.ts
jakolehm: This follows kubernetes api practices... meaning:
* `reason` - machine readable reason
* `message` - human readable message

src/renderer/protocol-handler/app-handlers.ts
jakolehm: You are right, I somehow missed that we also have `/` for this. I'll switch back to `/landing` for backward compat purposes.

src/common/catalog-category-registry.ts
jakolehm: Could we do it in a follow-up PR? I don't really have good ideas how to achieve async behaviour here.. mainly how long should await wait for a correct category 🤔

src/common/catalog-entities/kubernetes-cluster.ts
jakolehm: Not sure why? All catalog entities are generated from main and this function is an event handler (not a function that is called directly).

src/common/catalog-entities/kubernetes-cluster.ts
jakolehm: Yes, kind of temporary. We need to figure out how multiple sources can introduce same menu action without duplicate menu items. For example `source=gke-sync` should be able to override `delete` because it makes no sense to remove local state from Lens.

I would prefer that we solve this problem in a follow-up PR.

src/common/catalog-entity-registry.ts
jakolehm: I agree that type should be observable .. I'll investigate how to achieve that.

src/common/cluster-store.ts
jakolehm: Yes but it's not anymore related to Cluster. Let's do that in a follow-up PR (original idea was that user could also drag items from catalog to hotbar).

src/common/hotbar-store.ts
jakolehm: Maybe, I think we need to discuss what is the desired UX here (out of scope for this PR).

src/extensions/core-api/stores.ts
jakolehm: Yes, I just didn't have good idea where to put this. Maybe it should be top-level module in extension API? (like `Catalog.catalogEntities`)

src/extensions/registries/command-registry.ts
jakolehm: This is pretty much unsolved problem within this PR (see todo list). We should have a way to pass an Entity that is currently active context. How to do that? (I start to feel we should solve this in a follow-up PR so that complexity of this PR does not grow too much)

src/main/catalog-pusher.ts
jakolehm: Maybe we could add `toJS` (or similar) method to `CatalogEntity` interface?

src/main/cluster-manager.ts
jakolehm: Out of scope for this PR but yes I would like to get rid of that.

miskun (approved):
Approved. Let's continue iterations in followup PRs.

src/common/catalog-category-registry.ts
Nokel81: Sure we can do it in a follow up PR, I would say that it should wait for ever since that would be the most consistent.

src/extensions/core-api/stores.ts
Nokel81: I would prefer a new top level.

src/main/catalog-pusher.ts
Nokel81: `toJS` merely removes observables. If you look at the serializable documentation things like `functions` and class hierarchy are not preserved across IPC serialization.

PR #2421, opened 2021-03-31
Fix StorageHelper.get() returning null after clearing

Nokel81:
- Return the defaultValue

- Mark default value as required

This fixes master (which is currently failing CI)

### DIFF
--- a/src/renderer/components/+namespaces/namespace.store.ts
+++ b/src/renderer/components/+namespaces/namespace.store.ts
@@ -5,7 +5,7 @@ import { Namespace, namespacesApi } from "../../api/endpoints/namespaces.api";
 import { createPageParam } from "../../navigation";
 import { apiManager } from "../../api/api-manager";
 
-const selectedNamespaces = createStorage<string[]>("selected_namespaces", []);
+const selectedNamespaces = createStorage<string[] | undefined>("selected_namespaces", undefined);
 
 export const namespaceUrlParam = createPageParam<string[]>({
   name: "namespaces",
--- a/src/renderer/components/+namespaces/namespace.store.ts
+++ b/src/renderer/components/+namespaces/namespace.store.ts
@@ -5,7 +5,7 @@ import { Namespace, namespacesApi } from "../../api/endpoints/namespaces.api";
 import { createPageParam } from "../../navigation";
 import { apiManager } from "../../api/api-manager";
 
-const selectedNamespaces = createStorage<string[]>("selected_namespaces");
+const selectedNamespaces = createStorage<string[]>("selected_namespaces", []);
 
 export const namespaceUrlParam = createPageParam<string[]>({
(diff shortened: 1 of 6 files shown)


ixrock:
`StorageHelper.clear()` is probably redundant since there is `reset()` method which is doing what you trying to fix.
So the fix should just replace all usages to `reset()` and remove `clear()` from the class.

Nokel81:
So should we remove that too? I think we should still have this since a user can do `this.set(null)` manually.

ixrock:
> So should we remove that too? I think we should still have this since a user can do `this.set(null)` manually.

I would add a guard in `set()` to use `defaultValue` on `null`/`undefined` setting attempt.

Nokel81:
@ixrock instead or as well?

ixrock:
@Nokel81 instead. Otherwise we could just omit setting default value in constructor if it's served in `get()`?

src/renderer/hooks/useStorage.ts
```diff
@@ -2,7 +2,7 @@ import { useState } from "react";
 import { createStorage } from "../utils";
 import { CreateObservableOptions } from "mobx/lib/api/observable";
 
-export function useStorage<T>(key: string, initialValue?: T, options?: CreateObservableOptions) {
+export function useStorage<T>(key: string, initialValue: T, options?: CreateObservableOptions) {
```
jim-docker, line 5: This function doesn't appear to be used anywhere

jim-docker (dismissed):

src/renderer/hooks/useStorage.ts
Nokel81: Ah good catch. I changed this is a preemptive measure.

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -5,7 +5,7 @@ import { Namespace, namespacesApi } from "../../api/endpoints/namespaces.api";
 import { createPageParam } from "../../navigation";
 import { apiManager } from "../../api/api-manager";
 
-const selectedNamespaces = createStorage<string[]>("selected_namespaces");
+const selectedNamespaces = createStorage<string[]>("selected_namespaces", []);
```
ixrock: Default value here must be `undefined`, otherwise it breaks logic on first open and you'll see "All namespaces"

```ts
    const prevSelectedNamespaces = selectedNamespaces.get();

    // return previously saved namespaces from local-storage (if any)
    if (prevSelectedNamespaces) {
      return prevSelectedNamespaces.filter(namespace => namespaces.has(namespace));
    }
```

Nokel81: Ah thanks, will rework this.

ixrock (approved):

jim-docker (approved):

PR #2426, opened 2021-03-31
Add checks to block edits to key fields

### LINKED ISSUE #1876: Attempting to change the name of a namespace results in a (self called) app crash
**Describe the bug**
Attempting to change the name of a namespace results in a (self called) app crash.

**To Reproduce**
1. Go to Namespaces
2. Click on ⠇and select Edit
3. Change name: of namespace and click Save
The effect is that a new namespace is created, the old one is left intact, but the Lens app claims a crash.

**Expected behavior**
namespace renamed

**Screenshots**

- Lens Version: 4.0.5
- OS: Windows 10.0.19041.685
- Installation method (e.g. snap or AppImage in Linux): Lens-Setup-4.0.5.exe

**Logs:**
```
App crash at http://ade8f9f1-d507-496d-b697-df278801b50d.localhost:2021/namespaces
To help us improve the product please report bugs to Slack community or Github issues tracker.

Error stack:
 
TypeError: e.getStatus is not a function
    at customizeTableRowProps (http://ade8f9f1-d507-496d-b697-df278801b50d.localhost:2021/build/Lens.js:2:8509351)
    at r.value (http://ade8f9f1-d507-496d-b697-df278801b50d.localhost:2021/build/Lens.js:2:1360434)
    at http://ade8f9f1-d507-496d-b697-df278801b50d.localhost:2021/build/Lens.js:2:6140471
    at http://ade8f9f1-d507-496d-b697-df278801b50d.localhost:2021/build/Lens.js:2:65400
    at http://ade8f9f1-d507-496d-b697-df278801b50d.localhost:2021/build/Lens.js:2:65013
    at ce (http://ade8f9f1-d507-496d-b697-df278801b50d.localhost:2021/build/Lens.js:2:16554)
    at e.track (http://ade8f9f1-d507-496d-b697-df278801b50d.localhost:2021/build/Lens.js:2:27918)
    at http://ade8f9f1-d507-496d-b697-df278801b50d.localhost:2021/build/Lens.js:2:64989
    at http://ade8f9f1-d507-496d-b697-df278801b50d.localhost:2021/build/Lens.js:2:64826
    at M (http://ade8f9f1-d507-496d-b697-df278801b50d.localhost:2021/build/Lens.js:2:64963)

Component stack:

    in c
    in div
    in div
    in t
    in div
    in r
    in div
    in r
    in div
    in div
    in r
    in r
    in r
    in main
    in div
    in c
    in r
    in t
    in t
    in r
    in main
    in div
    in r
    in r
    in t
    in Unknown
    in r
```

Nokel81:
- Referece kubetcl:pkg/cmd/util/editor/editoptions.go:649

fixes #1876

### DIFF
--- a/src/renderer/api/kube-object.ts
+++ b/src/renderer/api/kube-object.ts
@@ -29,6 +29,7 @@ import { apiKube } from "./index";
 import type { JsonApiParams } from "./json-api";
 import { resourceApplierApi } from "./endpoints/resource-applier.api";
 import { hasOptionalProperty, hasTypedProperty, isObject, isString, bindPredicate, isTypedArray, isRecord } from "../../common/utils/type-narrowing";
+import _ from "lodash";
 
 export type IKubeObjectConstructor<T extends KubeObject = any> = (new (data: KubeJsonApiData | any) => T) & {
   kind?: string;
@@ -96,6 +97,7 @@ export class KubeObject<Metadata extends IKubeObjectMetadata = IKubeObjectMetada
   metadata: Metadata;
   status?: Status;
   spec?: Spec;
+  managedFields?: any;
 
   static create(data: KubeJsonApiData) {
     return new KubeObject(data);
@@ -178,6 +180,17 @@ export class KubeObject<Metadata extends IKubeObjectMetadata = IKubeObjectMetada
 
     return Object.entries(labels).map(([name, value]) => `${name}=${value}`);
   }
+
+  protected static readonly nonEditableFields = [
+    "apiVersion",
+    "kind",
+    "metadata.name",
+    "metadata.selfLink",
+    "metadata.resourceVersion",
+    "metadata.uid",
+    "managedFields",
+    "status",
+  ];
 
   constructor(data: KubeJsonApiData) {
     Object.assign(this, data);
@@ -267,6 +280,12 @@ export class KubeObject<Metadata extends IKubeObjectMetadata = IKubeObjectMetada
 
   // use unified resource-applier api for updating all k8s objects
   async update<T extends KubeObject>(data: Partial<T>): Promise<T> {
+    for (const field of KubeObject.nonEditableFields) {
+      if (!_.isEqual(_.get(this, field), _.get(data, field))) {
+        throw new Error(`Failed to update Kube Object: ${field} has been modified`);
+      }
+    }
+
     return resourceApplierApi.update<T>({
       ...this.toPlainObject(),
       ...data,


package.json
```diff
@@ -202,6 +203,7 @@
     "command-exists": "1.2.9",
     "conf": "^7.0.1",
     "crypto-js": "^4.0.0",
+    "deep-equal": "^2.0.5",
```
ixrock: How it differs from `lodash/isEqual`?

src/renderer/components/dock/__test__/dock-tabs.test.tsx
```diff
@@ -22,18 +22,12 @@ const getTabKinds = () => dockStore.tabs.map(tab => tab.kind);
...
-    const installChartTab: IDockTab = { id: "install", kind: TabKind.INSTALL_CHART, title: "Install chart" };
-    const logsTab: IDockTab = { id: "logs", kind: TabKind.POD_LOGS, title: "Logs" };
-
     dockStore.tabs.push(
-      terminalTab,
-      createResourceTab,
-      editResourceTab,
-      installChartTab,
-      logsTab
+      { id: "terminal1", kind: TabKind.TERMINAL, title: "Terminal" },
```
ixrock: `terminal`**1**?

src/renderer/api/kube-object.ts
```diff
@@ -65,6 +66,42 @@ export class KubeStatus {
...
+}
+
+function requiredMetadataKeyUnchanged(key: keyof KubeObject["metadata"]): EditPrecondtion {
+  return {
+    check: (original: KubeObject, edited: KubeObject): boolean => (
+      deepEqual(edited.metadata[key], original.metadata[key], { strict: true })
+    ),
+    message: `metadata.${key} has changed`,
+  };
+}
```
ixrock: This functions looks very similar.  Can we keep just one?

Also I would propose following definition for checks: `checkForErrorsBeforeUpdate(current: KubeObject, changed: Partial<KubeObject>)`. Wdyt?

And why do we need this at all? If I try to modify these fields wouldn't k8s-api respond with an error if it's not allowed?

src/renderer/components/dock/__test__/dock-tabs.test.tsx
Nokel81: I merely copied the `id` from above.

src/renderer/api/kube-object.ts
Nokel81: No the k8s-api doesn't do that unfortunately. This is actually what `kubectl` does when you try and edit something.

Do you mean instead of having it modular like this? I sort of like how this can be expanded rather easily in the future if we need to.

As for having only one, sure that isn't too hard.

package.json
Nokel81: This package uses node's deep equal algorithm, but I'll switch it out for the lodash one.

src/renderer/api/kube-object.ts
```diff
@@ -84,21 +84,27 @@ export interface EditPrecondtion {
...
-function requiredKeyUnchanged(key: keyof KubeObject): EditPrecondtion {
-  return {
-    check: (original: KubeObject, edited: KubeObject): boolean => (
-      deepEqual(edited[key], original[key], { strict: true })
-    ),
-    message: `${key} has changed`,
-  };
-}
+function requiredKeyUnchanged(key: keyof KubeObject): EditPrecondtion;
+function requiredKeyUnchanged(key: keyof KubeObject, key2?: keyof KubeObject[typeof key]): EditPrecondtion;
```
ixrock: `const key: keyof KubeObject = "getTimeDiffFromNow" // no errors, this is fine?`
Why not just a `string` and single `key` argument?

Nokel81: Okay I'll just have strings, though I need multiple args because of `metadata`.

ixrock: We can support single key and use string-path with https://lodash.com/docs/4.17.15#get
e.g. `requiredKeyUnchanged("metadata.namespace")`

package.json
ixrock: Forgotten unused package + `@types/deep-equal`

src/renderer/api/kube-object.ts
```diff
@@ -65,6 +66,33 @@ export class KubeStatus {
...
+ * This represents a precondition that must be satisfied before an edit is allowed
+ */
+export interface EditPrecondtion {
+  /**
+   * Checks to see if the edited version matches the original. Should return
+   * `true` if it passes the check.
+   * @param original The original version of the kube object
+   * @param edited The edited version of the kube object
+   */
+  check(original: any, edited: any): boolean;
```
ixrock: Why it's `any` now? 
`check(original: KubeObject, edited: Partial<KubeObject>)` doesn't fit for some reasons?

Nokel81: I changed it to `any` because it could be moved someplace else and isn't actually tied to `KubeObject`

src/renderer/api/kube-object.ts
```diff
@@ -184,6 +223,12 @@ export class KubeObject implements ItemObject {
 
   // use unified resource-applier api for updating all k8s objects
   async update<T extends KubeObject>(data: Partial<T>) {
+    for (const precondition of KubeObject.editPreconditions) {
```
ixrock: What if preconditions for some objects would be different (extended)? 
How we would overcome this in `update()`?

src/renderer/components/dock/dock.store.ts
```diff
@@ -165,18 +169,21 @@ export class DockStore implements DockStorageState {
...
       const tabNumber = this.getNewTabNumber(tab.kind);
 
       if (tabNumber > 1) tab.title += ` (${tabNumber})`;
     }
+
     this.tabs.push(tab);
-    this.selectTab(tab.id);
-    this.open();
+
+    if (open) {
```
ixrock: How this changes related to PR's scope btw?

src/renderer/api/kube-object.ts
Nokel81: We would override this method in the child classes, doing those pre-condition checks and then calling `super.update(data)`;

I could see an argument for using a registry instead (if for some reason some of these base checks are wrong for some object types). But I think doing that now is overkill.

Finally, it should be noted that `kubectl` does neither of those and merely checks all kube objects (though it doesn't block updating the `resourceVersion`, I blocked that because it fixes a bug where if you do that only the edited item is left in the store visually).

src/renderer/components/dock/dock.store.ts
Nokel81: I guess they are not, will remove.

src/renderer/api/kube-object.ts
Nokel81: Actually I have thought that is might be better to just remove them.

src/renderer/components/dock/edit-resource.store.ts
```diff
@@ -95,27 +95,24 @@ export class EditResourceStore extends DockTabStore<EditingResource> {
 export const editResourceStore = new EditResourceStore();
 
 export function editResourceTab(object: KubeObject, tabParams: Partial<IDockTab> = {}) {
-  // use existing tab if already opened
-  let tab = editResourceStore.getTabByResource(object);
+  const oldTab = editResourceStore.getTabByResource(object);
```
jakolehm: How are these tab changes related to PR topic?

Nokel81: They aren't, I thought I removed them when Roman noticed this too... sorry.

src/renderer/api/kube-object.ts
```diff
@@ -86,12 +87,20 @@ export class KubeObject implements ItemObject {
...
   }
 
+  protected static readonly nonEditableFields = [
+    "apiVersion",
+    "kind",
+    "metadata.name",
+    "metadata.selfLink",
+    "metadata.resourceVersion",
+    "metadata.uid",
+    "managedFields",
```
jakolehm, line 191: Probably also `status` should be here?

src/renderer/components/dock/edit-resource.store.ts
jakolehm: This PR still has changes that are not related to PR topic.. If you really want these changes included then please add review comments why they are here.

src/renderer/api/kube-object.ts
Nokel81: Good idea.

Nokel81:
I just found a bug from #2428 which I needed to fix to fix the conflicts. `It is src/renderer/components/kube-object/kube-object-menu.tsx:83`

Nokel81:
@jakolehm and @ixrock. I have remove the tab changes. I don't remember why I had them in here (as this PR is 2 months old) and they don't seem to be needed. PTALA.

src/renderer/api/kube-object.ts
```diff
@@ -267,6 +280,12 @@ export class KubeObject<Metadata extends IKubeObjectMetadata = IKubeObjectMetada
 
   // use unified resource-applier api for updating all k8s objects
   async update<T extends KubeObject>(data: Partial<T>): Promise<T> {
+    for (const field of KubeObject.nonEditableFields) {
+      if (!_.isEqual(_.get(this, field), _.get(data, field))) {
```
BlackHole1, line 284:
```suggestion
      if (this[field] !== data[field]) {
```

src/renderer/api/kube-object.ts
```diff
@@ -29,6 +29,7 @@ import { apiKube } from "./index";
 import type { JsonApiParams } from "./json-api";
 import { resourceApplierApi } from "./endpoints/resource-applier.api";
 import { hasOptionalProperty, hasTypedProperty, isObject, isString, bindPredicate, isTypedArray, isRecord } from "../../common/utils/type-narrowing";
+import _ from "lodash";
```
BlackHole1, line 32:
```suggestion
```

src/renderer/api/kube-object.ts
BlackHole1: There is no need to use lodash here

Nokel81: I disagree, if a user edits a deep field back to the same value we don't want to report an error.

BlackHole1: Sorry, I didn't look closely enough. You are right.

msa0311 (approved):

PR #2428, opened 2021-03-31
display close button on deleted kube details page

### LINKED ISSUE #2345: SideWindow does not close - SideWindow does not close with Pod/deployemnt info if it was open when resource was deleted via terminal
**Describe the bug**
SideWindow does not close with Pod/deployemnt info if it was open when resource was deleted via terminal

**To Reproduce**
1. Create a Pod
2. Open the information side window
3. Kill the pod/deployment from the terminal
4. Side window goes blank(expected) but then the X on the top right does not close the side window. need to disconnect from the cluster and reconnect

**Expected behavior**
Side window should close either automatically on resource deletion or close button should work

- Mac OS  Big Sur 11.1 (20C69)
- Lens: 4.1.4
Electron: 9.4.0
Chrome: 83.0.4103.122
Node: 12.14.1
Copyright 2020 Mirantis, Inc.

Nokel81:

fixes #2345

### DIFF
--- a/src/renderer/components/kube-object/kube-object-list-layout.tsx
+++ b/src/renderer/components/kube-object/kube-object-list-layout.tsx
@@ -17,6 +17,10 @@ export interface KubeObjectListLayoutProps extends ItemListLayoutProps {
 
 @observer
 export class KubeObjectListLayout extends React.Component<KubeObjectListLayoutProps> {
+  static defaultProps: Partial<KubeObjectListLayoutProps> = {
+    onDetails: (item: KubeObject) => showDetails(item.selfLink),
+  };
+
   @computed get selectedItem() {
     return this.props.store.getByPath(kubeSelectedUrlParam.get());
   }
@@ -33,14 +37,6 @@ export class KubeObjectListLayout extends React.Component<KubeObjectListLayoutPr
     ]);
   }
 
-  onDetails = (item: KubeObject) => {
-    if (this.props.onDetails) {
-      this.props.onDetails(item);
-    } else {
-      showDetails(item.selfLink);
-    }
-  };
-
   render() {
     const { className, store, items = store.contextItems, ...layoutProps } = this.props;
 
@@ -52,8 +48,7 @@ export class KubeObjectListLayout extends React.Component<KubeObjectListLayoutPr
         items={items}
         preloadStores={false} // loading handled in kubeWatchApi.subscribeStores()
         detailsItem={this.selectedItem}
-        onDetails={this.onDetails}
-        renderItemMenu={(item) => <KubeObjectMenu object={item as KubeObject} />} // safe because we are dealing with KubeObjects here
+        renderItemMenu={(item: KubeObject) => <KubeObjectMenu object={item} />} // safe because we are dealing with KubeObjects here
       />
     );
   }
--- a/src/renderer/components/kube-object/kube-object-list-layout.tsx
+++ b/src/renderer/components/kube-object/kube-object-list-layout.tsx
@@ -15,11 +15,13 @@ export interface KubeObjectListLayoutProps extends ItemListLayoutProps {
   dependentStores?: KubeObjectStore[];
 }
 
+const defaultProps: Partial<KubeObjectListLayoutProps> = {
+  onDetails: (item: KubeObject) => showDetails(item.selfLink),
+};
+
 @observer
 export class KubeObjectListLayout extends React.Component<KubeObjectListLayoutProps> {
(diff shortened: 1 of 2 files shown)


src/renderer/components/kube-object/kube-object-menu.tsx
```diff
@@ -7,13 +7,13 @@ import { hideDetails } from "./kube-object-details";
 import { apiManager } from "../../api/api-manager";
 import { kubeObjectMenuRegistry } from "../../../extensions/registries/kube-object-menu-registry";
 
-export interface KubeObjectMenuProps<T extends KubeObject = any> extends MenuActionsProps {
-  object: T;
+export interface KubeObjectMenuProps<T> extends MenuActionsProps {
+  object: T | null | undefined;
```
ixrock, line 11: Why both `null` and `undefined`?  
Can you provide an example when it could be helpful? since `null == undefined` // true

src/renderer/components/kube-object/kube-object-list-layout.tsx
```diff
@@ -53,9 +53,7 @@ export class KubeObjectListLayout extends React.Component<KubeObjectListLayoutPr
         preloadStores={false} // loading handled in kubeWatchApi.subscribeStores()
         detailsItem={this.selectedItem}
         onDetails={this.onDetails}
-        renderItemMenu={(item) => {
-          return <KubeObjectMenu object={item}/>;
-        }}
+        renderItemMenu={(item) => <KubeObjectMenu object={item as KubeObject} />} // safe because we are dealing with KubeObjects here
```
ixrock:
```suggestion
        renderItemMenu={(item: KubeObject) => {
          return <KubeObjectMenu object={item}/>;
        }}
```

ixrock (dismissed):

src/renderer/components/kube-object/kube-object-menu.tsx
Nokel81: I guess I can remove one of them, I did this to be "most correct" since we sometimes return `null` and sometimes return `undefined`. When we move to `strict` mode this will become more relevant.

src/renderer/components/kube-object/kube-object-list-layout.tsx
```diff
@@ -17,6 +17,10 @@ export interface KubeObjectListLayoutProps extends ItemListLayoutProps {
 
 @observer
 export class KubeObjectListLayout extends React.Component<KubeObjectListLayoutProps> {
+  static defaultProps: Partial<KubeObjectListLayoutProps> = {
```
ixrock: You cannot use it like this (not sure exactly why though), but it leads to that all props becomes `optional`.
For example comment out required `store={...}` in some place of use and you'll see no complains from TS.

Workaround would be as done in other places:

```ts
const defaultProps: Partial<Props> = {
  // define optionals
};

export class ComponentName extends React.Component<Props> {
```

src/renderer/components/kube-object/kube-object-menu.tsx
ixrock: Are we going to move forward `strict`? What's for if so?

Nokel81: It help catches more problems, the rules that it turns on is:

- noImplicitAny
- noImplicitThis
- strictNullChecks
- strictPropertyInitialization
- strictBindCallApply
- strictFunctionTypes

src/renderer/components/kube-object/kube-object-list-layout.tsx
Nokel81: Ah fair thanks for the catch. I think we still do this in some places but that should be a different PR.

ixrock (dismissed):

ixrock (approved):

jim-docker (approved):
Amid the cleanup and refactoring I'm not sure what actually fixes the bug. Although I couldn't reproduce the bug before this fix anyway :)

src/renderer/components/kube-object/kube-object-menu.tsx
```diff
@@ -52,34 +52,45 @@ export class KubeObjectMenu extends React.Component<KubeObjectMenuProps> {
...
+          toolbar={toolbar}
+        />
+      ));
+  }
+
   render() {
     const { remove, update, renderRemoveMessage, isEditable, isRemovable } = this;
-    const { className, object, editable, removable, toolbar, ...menuProps } = this.props;
-
-    if (!object) return null;
```
Nokel81, line 67: The moving of this check to the above is the substance of the fix

PR #2435, opened 2021-04-01
Add support for installing extensions by name

### LINKED ISSUE #1707: Install an extension from a NPM registry url

### LINKED ISSUE #1708: Allow to install an extension using npmjs.com web url
Similar to #1707 but tailored for npmjs.com web urls, for example https://www.npmjs.com/package/@mirantis/lens-extension-cc . Underneath this could be resolved to a registry api URL which returns a json object.

Nokel81:
- Optionally specify the version, defaults to latest non-prerelease version
- Usable via the protocol handlers, but requires user confirmation to actually install

fixes #1707 

FYI: I borked the previous PR while trying to fix conflicts. That is why there is a new PR.

### DIFF
--- a/src/renderer/components/+extensions/extensions.tsx
+++ b/src/renderer/components/+extensions/extensions.tsx
@@ -6,7 +6,7 @@ import { disposeOnUnmount, observer } from "mobx-react";
 import os from "os";
 import path from "path";
 import React from "react";
-import { autobind, disposer, Disposer, downloadFile, extractTar, listTarEntries, noop, readFileFromTar } from "../../../common/utils";
+import { autobind, disposer, Disposer, downloadFile, downloadJson, extractTar, listTarEntries, noop, readFileFromTar } from "../../../common/utils";
 import { docsUrl } from "../../../common/vars";
 import { extensionDiscovery, InstalledExtension, manifestFilename } from "../../../extensions/extension-discovery";
 import { extensionLoader } from "../../../extensions/extension-loader";
@@ -23,6 +23,9 @@ import { Notifications } from "../notifications";
 import { Spinner } from "../spinner/spinner";
 import { TooltipPosition } from "../tooltip";
 import { ExtensionInstallationState, ExtensionInstallationStateStore } from "./extension-install.store";
+import URLParse from "url-parse";
+import { SemVer } from "semver";
+import _ from "lodash";
 
 function getMessageFromError(error: any): string {
   if (!error || typeof error !== "object") {
@@ -46,23 +49,27 @@ function getMessageFromError(error: any): string {
   return rawMessage;
 }
 
+interface ExtensionInfo {
+  name: string;
+  version?: string;
+  requireConfirmation?: boolean;
+}
+
 interface InstallRequest {
   fileName: string;
-  filePath?: string;
-  data?: Buffer;
+  dataP: Promise<Buffer | null>;
 }
 
-interface InstallRequestPreloaded extends InstallRequest {
+interface InstallRequestValidated {
+  fileName: string;
   data: Buffer;
-}
-
-interface InstallRequestValidated extends InstallRequestPreloaded {
   id: LensExtensionId;
   manifest: LensExtensionManifest;
   tempFile: string; // temp system path to packed extension for unpacking
 }
 
-async function uninstallExtension(extensionId: LensExtensionId, manifest: LensExtensionManifest): Promise<boolean> {
+async function uninstallExtension(extensionId: LensExtensionId): Promise<boolean> {
+  const { manifest } = extensionLoader.getExtension(extensionId);
   const displayName = extensionDisplayName(manifest.name, manifest.version);
 
   try {
@@ -92,18 +99,17 @@ async function uninstallExtension(extensionId: LensExtensionId, manifest: LensEx
   }
 }
 
-function confirmUninstallExtension(extension: InstalledExtension): void {
+async function confirmUninstallExtension(extension: InstalledExtension): Promise<void> {
   const displayName = extensionDisplayName(extension.manifest.name, extension.manifest.version);
-
-  ConfirmDialog.open({
+  const confirmed = await ConfirmDialog.confirm({
     message: <p>Are you sure you want to uninstall extension <b>{displayName}</b>?</p>,
     labelOk: "Yes",
     labelCancel: "No",
-    ok: () => {
-      // Don't want the confirm dialog to stay up longer than the click
-      uninstallExtension(extension.id, extension.manifest);
-    }
   });
+
+  if (confirmed) {
+    await uninstallExtension(extension.id);
+  }
 }
 
 function getExtensionDestFolder(name: string) {
@@ -114,16 +120,10 @@ function getExtensionPackageTemp(fileName = "") {
   return path.join(os.tmpdir(), "lens-extensions", fileName);
 }
 
-async function preloadExtension({ fileName, data, filePath }: InstallRequest, { showError = true } = {}): Promise<InstallRequestPreloaded | null> {
-  if(data) {
-    return { filePath, data, fileName };
-  }
-
+async function readFileNotify(filePath: string, showError = true): Promise<Buffer | null> {
   try {
-    const data = await fse.readFile(filePath);
-
-    return { filePath, data, fileName };
-  } catch(error) {
+    return await fse.readFile(filePath);
+  } catch (error) {
     if (showError) {
       const message = getMessageFromError(error);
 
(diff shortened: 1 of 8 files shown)


src/renderer/protocol-handler/app-handlers.ts
```diff
@@ -54,5 +55,15 @@ export function bindProtocolAddRouteHandlers() {
     })
     .addInternalHandler("/extensions", () => {
       navigate(extensionsURL());
+    })
+    .addInternalHandler(`/extensions${LensProtocolRouter.ExtensionUrlSchema}`, ({ pathname, search: { version } }) => {
```
jakolehm: Should route include "install", like `/extensions/install${LensProtocolRouter.ExtensionUrlSchema}`?

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -258,20 +265,62 @@ async function unpackExtension(request: InstallRequestValidated, disposeDownload
...
+
+    if (!proceed) {
+      return disposer();
+    }
   }
 
-  const validatedRequest = await createTempFilesAndValidate(loadedRequest);
+  const url = json.versions[version].dist.tarball;
+  const fileName = path.basename(url);
+  const { promise: dataP } = downloadFile({ url, timeout: 60000 /*1m*/ });
```
jim-docker: Is 1m enough?

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -524,10 +591,10 @@ export class Extensions extends React.Component {
                 disabled={ExtensionInstallationStateStore.anyPreInstallingOrInstalling}
                 placeholder={`Path or URL to an extension package (${supportedFormats.join(", ")})`}
```
jim-docker:
```suggestion
                placeholder={`Name or path or URL to an extension package (${supportedFormats.join(", ")})`}
```

src/renderer/components/confirm-dialog/confirm-dialog.tsx
```diff
@@ -33,7 +38,17 @@ export class ConfirmDialog extends React.Component<ConfirmDialogProps> {
...
+    return new Promise(resolve => {
+      ConfirmDialog.open({
+        ok: () => resolve(true),
+        cancel: () => resolve(false),
+        ...params,
+      });
+    });
+  }
+
+  private static closeDialog() {
```
jim-docker: this is never called?

src/renderer/protocol-handler/app-handlers.ts
jim-docker: I agree, might want to reserve `/extensions/{extension-name}` for some other behaviour

jim-docker:
How should the protocol url be specified? I tried:
```
open lens://extensions@mirantis/lens-extension-cc
```
got:
```
[0] info: [PROTOCOL ROUTER]: routing lens://extensions@mirantis/lens-extension-cc
[0] error: [PROTOCOL ROUTER]: invalid host {"url":{"slashes":true,"protocol":"lens:","hash":"","query":{},"pathname":"/lens-extension-cc","auth":"extensions","host":"mirantis","port":"","hostname":"mirantis","password":"","username":"extensions","origin":"lens://mirantis","href":"lens://extensions@mirantis/lens-extension-cc"}}
```

src/renderer/protocol-handler/app-handlers.ts
jim-docker: Should it be this?

```suggestion
    .addInternalHandler(`/extensions/${LensProtocolRouter.ExtensionUrlSchema}`, ({ pathname, search: { version } }) => {
```

Nokel81: The URL Schema starts with a `/`

Nokel81:
It should be `lens://app/extension/@mirantis/lens-extension-cc` @jim-docker

src/renderer/components/+extensions/extensions.tsx
Nokel81: It is just a JSON file, I just used the same timeout as for the extensions. Now that you mention it, both of these should be longer.

src/renderer/protocol-handler/app-handlers.ts
Nokel81: Okay sounds good. FYI @stefcameron

jim-docker:
> It should be `lens://app/extension/@mirantis/lens-extension-cc` @jim-docker

I tried that too, it also didn't work, similar error

jim-docker:
> > It should be `lens://app/extension/@mirantis/lens-extension-cc` @jim-docker
> 
> I tried that too, it also didn't work, similar error

I take it back, I tried that without the `/`, will try again

jim-docker:
> It should be `lens://app/extension/@mirantis/lens-extension-cc` @jim-docker

Are you changing it to `lens://app/extensions/install...`?

Nokel81:
@jim-docker Yes I am planning to

src/renderer/components/+extensions/extensions.tsx
Nokel81: I'll bump them up to 10 minutes. But I think giving the user control over this would be better in the future.

stefcameron:
OK, so it's going to be `lens://app/extension/install/@mirantis/lens-extension-cc` ?

Nokel81:
@stefcameron yes

jakolehm (approved):

jim-docker:
> OK, so it's going to be `lens://app/extension/install/@mirantis/lens-extension-cc` ?

It is `extensions`:
```
    .addInternalHandler(`/extensions/install${LensProtocolRouter.ExtensionUrlSchema}`, ({ pathname, search: { version } }) => {
```

stefcameron:
> > OK, so it's going to be `lens://app/extension/install/@mirantis/lens-extension-cc` ?
> 
> It is `extensions`:
> 
> ```
>     .addInternalHandler(`/extensions/install${LensProtocolRouter.ExtensionUrlSchema}`, ({ pathname, search: { version } }) => {
> ```

Ah, plural, thank you, @jim-docker !

jim-docker:
I'm seeing this:
Screen Shot 2021-04-08 at 3 20 42 PM
when I run this:
```
open lens://app/extensions/install/@mirantis/lens-extension-cc
```
(the extension was not already installed)

Nokel81:
Thanks for the check, will investigate after I review the catalog PR.

jim-docker:
> (the extension was not already installed)

Actually, doesn't matter, the dialog is blank in any case. Installs ok...

Nokel81:
@jim-docker Thanks, I forgot to add the message to that dialog.

jim-docker:
minor points
- if I try to install a non-existent extension via protocol handler Lens navigates to the extensions page but nothing happens. Maybe it could show a dialog stating extension not found.
- if the extension is already installed at the same version should it really uninstall and then reinstall it? Could just show a dialog stating it is already installed. If the user really wants to install the same version again they could manually uninstall it and then install it again.

Nokel81:
@jim-docker 

1. Ah good point
2. Do you mean from the protocol handler? Or do you mean at all?

jim-docker:
> 2. Do you mean from the protocol handler? Or do you mean at all?

I think for both protocol handler and from extensions page. But this is just my opinion. I think it's a bit odd to see in the notification that it is going to install v2.1.2 but first it will remove v2.1.2. It's unnecessary.

Nokel81:
I agree it is unnecessary, however, this was the original functionality and it sort of seems reasonable. Since it makes the "I want to reinstall this" faster.

jim-docker:
> * if I try to install a non-existent extension via protocol handler Lens navigates to the extensions page but nothing happens. Maybe it could show a dialog stating extension not found.

actually, it looks like the extensions page gets stuck in "install" mode (animation of Install button, can't specify/pick an extension name. Have to reload to clear

Nokel81:
Ah that is worse, thanks for finding that bug

stefcameron:
> I agree it is unnecessary, however, this was the original functionality and it sort of seems reasonable. Since it makes the "I want to reinstall this" faster.

Along this train of thought, for a future release, you might consider adding an option to have Lens clear the extension's data store as part of the uninstallation since that's not done today -- which is good in some ways, but might not be desired in others. If an extension got hosed in some way for some reason because of bad state data in its store, right now, uninstalling and reinstalling wouldn't fix the issue.

Nokel81:
@stefcameron Agreed, we probably should do that by default...

Nokel81:
@jim-docker Fixed

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -269,14 +269,12 @@ export async function attemptInstallByInfo({ name, version, requireConfirmation
   const disposer = ExtensionInstallationStateStore.startPreInstall();
   const registryUrl = new URLParse("https://registry.npmjs.com").set("pathname", name).toString();
   const { promise } = downloadJson({ url: registryUrl });
+  const json = await promise.catch(console.error);
 
-  let json;
+  if (!json || json.error || typeof json.values !== "object" || !json.values) {
```
jim-docker: this seems to be blocking valid extensions now too:

Screen Shot 2021-04-09 at 3 57 45 PM

(same for protocol handler)

jim-docker (approved):

jakolehm (approved):

PR #2443, opened 2021-04-02
Ingress and PersistentVolumeClaim metrics need to also be filtered and grouped by namespace

iciclespider:
Ingress and PersistentVolumeClaim metrics need to also be filtered and grouped by namespace, in addition to ingress or persistentvolumeclaim.

I have a cluster where there is a `testing` and a `production` namespace with identical Kubernetes resources, including identically named Ingresses. Lens was showing the same metrics for these Ingresses, which was the sum of both Ingresses.

### DIFF
--- a/src/main/prometheus/lens.ts
+++ b/src/main/prometheus/lens.ts
@@ -67,18 +67,18 @@ export class PrometheusLens implements PrometheusProvider {
         };
       case "pvc":
         return {
-          diskUsage: `sum(kubelet_volume_stats_used_bytes{persistentvolumeclaim="${opts.pvc}"}) by (persistentvolumeclaim, namespace)`,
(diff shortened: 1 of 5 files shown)


Nokel81 (approved):
Does this fix any reported issues (on the issue tracker)?

iciclespider:
> Does this fix any reported issues (on the issue tracker)?

Not that I am aware. The above description does explain how this issue occurs. I could create an issue from that description if that helps.

Nokel81:
Nah that is good, no need now that there is a PR. Thanks for this PR.

iciclespider:
I noticed this was tagged as an `enhancement`. I would call it a `bug`, because this is an existing feature that is reporting the wrong numbers.

Nokel81:
@iciclespider fair, will change

jakolehm (approved):

PR #2454, opened 2021-04-06
Menu: more ways to position menu

ixrock:
Allows more fine-grained positioning for `<Menu usePortal={false}>` with following `props.position={{...}}` interface:

```ts
export type MenuPositionSide = "inside" | "outside"; // of parent's box area

export interface MenuPosition {
  left?: MenuPositionSide;
  top?: MenuPositionSide;
  right?: MenuPositionSide;
  bottom?: MenuPositionSide;
}
```

_As an example, workspace-menu moved right to the icon (previously at the bottom):_

Screenshot 2021-04-06 at 14 59 51

### DIFF
--- a/src/renderer/components/menu/menu.tsx
+++ b/src/renderer/components/menu/menu.tsx
@@ -9,7 +9,13 @@ import debounce from "lodash/debounce";
 
 export const MenuContext = React.createContext<MenuContextValue>(null);
 export type MenuContextValue = Menu;
-export type MenuPositionSide = "inside" | "outside";
+
+/**
+ * Positioning menu relative to parent's element box-area.
+ * In that case menu is rendered in parent's element DOM-tree.
+ * Applicable only when usePortal={false} (default)
+ */
+export type MenuPositionSide = "inside" | "outside"; // of parent element boundaries
 
 export interface MenuPosition {
   left?: MenuPositionSide;
@@ -36,6 +42,7 @@ export interface MenuProps {
 }
 
 interface State {
+  // Auto-positioning menu in <body> in case of `usePortal={true}`
   position?: {
     left?: boolean;
     top?: boolean;
--- a/src/renderer/components/menu/menu.tsx
+++ b/src/renderer/components/menu/menu.tsx
@@ -9,12 +9,13 @@ import debounce from "lodash/debounce";
 
 export const MenuContext = React.createContext<MenuContextValue>(null);
 export type MenuContextValue = Menu;
+export type MenuPositionSide = "inside" | "outside";
 
 export interface MenuPosition {
-  left?: boolean;
-  top?: boolean;
-  right?: boolean;
-  bottom?: boolean;
+  left?: MenuPositionSide;
+  top?: MenuPositionSide;
+  right?: MenuPositionSide;
+  bottom?: MenuPositionSide;
 }
 
 export interface MenuProps {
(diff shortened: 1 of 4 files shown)


src/renderer/components/menu/menu.tsx
```diff
@@ -9,12 +9,13 @@ import debounce from "lodash/debounce";
 
 export const MenuContext = React.createContext<MenuContextValue>(null);
 export type MenuContextValue = Menu;
+export type MenuPositionSide = "inside" | "outside";
```
Nokel81: I would like to see a comment explaining exactly what each of these means. I am having a really hard time trying to figure out what `{right: "outside", top: "inside"}` means.

src/renderer/components/menu/menu.tsx
```diff
@@ -35,11 +36,16 @@ export interface MenuProps {
 }
 
 interface State {
-  position?: MenuPosition;
+  position?: {
+    left?: boolean;
+    top?: boolean;
+    right?: boolean;
+    bottom?: boolean;
+  };
```
Nokel81, line 51: Why are these booleans? Furthermore, it doesn't look like this is modifiable so it isn't really state.

src/renderer/components/menu/menu.tsx
ixrock: Imaging a box (parent element with borders). When you position a child element inside you can put it "in" or "out" of parent's boundaries. (if it wouldn't help I'll illustrate next time :)

src/renderer/components/menu/menu.tsx
ixrock: It's different type of positioning (currently auto-positioning only). Applicable when `usePortal={true}`

src/renderer/components/menu/menu.tsx
Nokel81: That sounds good, would you mind adding it as a comment in the code?

ixrock: Sure.

aleksfront:
`export type MenuPositionSide = "inside" | "outside";` seems like advanced way of positioning which can be hard to understand for the developer. Can we keep this like a low-level structure but provide easier keywords for menu position, eg
```
"top" | "topRight" | "right" | "bottomRight" | "bottom" | "bottomLeft" | "left" | "topLeft"
```
so it can de defined like this
```
<Menu
  usePortal={false}
  placement="right"
  isOpen={this.menuOpen}
  ...
>
</Menu>
```
So `top` became equal to `position={{left: "inside", top: "outside"}}`
`bottomRight` equals to `position={{right: "outside", top: "inside"}}` and so on.

ixrock:
> So `top` became equal to `position={{left: "inside", top: "outside"}}`
> `bottomRight` equals to `position={{right: "outside", top: "inside"}}` and so on.

@aleksfront if there is replacement for all possible options with "inside"/"outside" I don't mind.

How to make `bottomRight` to be "outside" for both right & top in your example?
Can you provide replacement table for all options?

aleksfront:
The original issue was to fix Menu positions with `usePortal` enabled. It may be implemented first.
https://github.com/lensapp/lens/blob/master/src/renderer/components/hotbar/hotbar-icon.tsx#L107

Closing this because the solution doesn't satisfy all parties and getting outdated. We can go back to this question while refactoring `<Menu/>` component in further PRs.

PR #2457, opened 2021-04-06
switch to fuzzing pod table tests

Nokel81:

This way we can still test a bunch of different configurations and we don't do so many tests.

### DIFF
--- a/src/renderer/api/__tests__/pods.test.ts
+++ b/src/renderer/api/__tests__/pods.test.ts
@@ -19,6 +19,7 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
+import { random } from "lodash";
 import { Pod } from "../endpoints";
 
 interface GetDummyPodOptions {
@@ -28,16 +29,7 @@ interface GetDummyPodOptions {
   initDead?: number;
 }
 
-function getDummyPodDefaultOptions(): Required<GetDummyPodOptions> {
-  return {
-    running: 0,
-    dead: 0,
-    initDead: 0,
-    initRunning: 0,
-  };
-}
-
-function getDummyPod(opts: GetDummyPodOptions = getDummyPodDefaultOptions()): Pod {
+function getDummyPod({ running = 0, dead = 0, initDead = 0, initRunning = 0 }: GetDummyPodOptions = {}): Pod {
   const pod = new Pod({
     apiVersion: "v1",
     kind: "Pod",
@@ -66,7 +58,7 @@ function getDummyPod(opts: GetDummyPodOptions = getDummyPodDefaultOptions()): Po
     initContainerStatuses: [],
   };
 
-  for (let i = 0; i < opts.running; i += 1) {
+  for (let i = 0; i < running; i += 1) {
     const name = `container_r_${i}`;
 
     pod.spec.containers.push({
@@ -88,7 +80,7 @@ function getDummyPod(opts: GetDummyPodOptions = getDummyPodDefaultOptions()): Po
     });
   }
 
-  for (let i = 0; i < opts.dead; i += 1) {
+  for (let i = 0; i < dead; i += 1) {
     const name = `container_d_${i}`;
 
     pod.spec.containers.push({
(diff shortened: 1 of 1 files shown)


nevalla:
Should this go to `master` instead of `v4.2`. Or is this fixing any issue or flaky test?

jim-docker (commented):
This has the makings of a flaky test, for a bug that appears only for a specific set of parameters. Is this test set currently a bottleneck?

Nokel81:
It is currently a bottleneck at least there are currently 40,000 tests that these are replacing. I probably don't know a lot about test fuzzing so maybe this isn't the best way of doing it.

jim-docker:
40,000 is not a problem if they run quickly. Are our tests taking too long? I don't know much about fuzzing either, but if you run a different random set every time then you don't know if or when you'll catch a bug.  One thing we used to do was run a different 1/7th of the tests on each day of the week. So we'd reproducibly cover the whole set in a week.

Nokel81:
@jakolehm Removed switched it from `10^4` to `3^4` on master. So he thinks that those are too many tests.

I guess it would be a good idea to somehow "remember" which versions failed and then run those again.

jakolehm:
> @jakolehm Removed switched it from `10^4` to `3^4` on master. So he thinks that those are too many tests.
> 
> I guess it would be a good idea to somehow "remember" which versions failed and then run those again.

40k unit tests is too much.. it also makes running unit tests way too slow for my taste. But the biggest issue there is that we get miles of test logs -> impossible to find anything from the logs. If we really need to run that many iterations why not just do them in a single test?

Nokel81:
@jakolehm I don't see what you mean about the duration issue. Since those tests take almost no time at all to run (maybe all of the 40k of them takes maybe 10 second). 

As to running all of them in one test, if one of them failed it would be nice to be told immediately which case failed. I tend to run "Find" when searching through the logs, especially since `jest` outputs the failed tests at the end of the log.

Furthermore, we don't need to run all 40k of them. That is why I am in favour of fuzzing them.

Nachasic (dismissed):
Honestly, I don't know a lot about fuzzing tests, though this looks good to me.

src/renderer/api/__tests__/pods.test.ts
```diff
@@ -166,14 +158,13 @@ function getDummyPod(opts: GetDummyPodOptions = getDummyPodDefaultOptions()): Po
...
 
-  for (let r = 0; r < 3; r += 1) {
-    for (let d = 0; d < 3; d += 1) {
-      for (let ir = 0; ir < 3; ir += 1) {
-        for (let id = 0; id < 3; id += 1) {
-          podTests.push([r, d, ir, id]);
-        }
-      }
-    }
+  for (let r = 0; r < 100; r += 1) {
```
jim-docker, line 161: I think Jari changed this test to range over 3 instead of 10 per for loop, so this PR is now actually adding more tests :)

Nokel81: It is, but not nearly to the same extent.

Nokel81: Plus 3^4 is pretty close (from an order of magnitude POV) to 10^2

jim-docker: I'm confused. Is sparse coverage over a higher value set (0->50)  a better test set than full coverage over  (0->3)? The functionality being tested isn't really dependent on how big n is in (0->n), is it?

Nokel81: For certain types of tests I believe that is that case. This way you can have a greater coverage area.

I acknowledge that it doesn't prevent people from just rerunning the tests until they pass though.

jim-docker (commented):
If I understand the lodash random() function correctly this test will get a different sequence of random numbers each time it is run. Good for getting broad coverage of test parameters over time but problematic in that it may fail on some runs, making it flaky. Since the test will output the specific parameters that cause it to fail then the bug should at least be reproducible. But flaky tests mean that you may have to fix some unrelated bug to get your PR to pass CI. Or, ugh, keep running the CI build till you get a pass. I'm not sure it's a good idea to allow potentially flaky tests. We should get more opinions.

If we want to avoid flaky tests we could employ a different strategy in the future. We could run more comprehensive test suites on a nightly (or weekly) basis. These tests wouldn't block anything (or maybe just significant releases)

Nokel81:
I like the idea of running weekly tests like this, which then get reported to Slack or something.

Nachasic (approved):

PR #2469, opened 2021-04-07
cleanup shell sessions

Nokel81:

This is targeting a branch since that should be merged before this.

I have tested that this works still with node (remote) shells and local shells.


jakolehm (commented):
Could you describe where the fix actually is? It's quite hard to spot it now that files were renamed/relocated.

Nokel81:
@jakolehm Now that #2468 is in master, this isn't really a fix in the sense of a bug but more in the conceptual sense.

I meant for this PR to clean up the design of these two classes so that during future reviews it is easier to catch the problems that were accidentally added in #2372.

jakolehm (approved):

nevalla (approved):

Issue #2486, opened 2021-04-09
Improve Catalog interface types


jakolehm:
See following review comments for details:

- https://github.com/lensapp/lens/pull/2418/files#r610094234
- https://github.com/lensapp/lens/pull/2418#discussion_r610064259
- https://github.com/lensapp/lens/pull/2418#discussion_r610088488
- https://github.com/lensapp/lens/pull/2418/files#r610090925

PR #2495, opened 2021-04-09
Improve CatalogEntityRegistry types & add tests

jakolehm:
Related to #2486


aleksfront (approved):

src/common/catalog-entities/web-link.ts
```diff
@@ -17,6 +17,12 @@ export class WebLink implements CatalogEntity {
   @observable public status: WebLinkStatus;
   @observable public spec: WebLinkSpec;
 
+  constructor(data: CatalogEntityData) {
+    this.metadata = data.metadata;
+    this.status = data.status as WebLinkStatus;
+    this.spec = data.spec as WebLinkSpec;
```
Nokel81, line 23: I think we should verify these here instead of just casting them.

nevalla (approved):

PR #2497, opened 2021-04-09
Allow register actions to "add to catalog" button

### LINKED ISSUE #2488: Catalog / Entity api for adding actions to "Add X to Catalog" button
This was not implemented in #2418 . We need an event handler that allows to register menu items to "+" button in Catalog view.

jakolehm:
image

Fixes #2488


src/renderer/components/+catalog/catalog-add-button.tsx
```diff
@@ -0,0 +1,77 @@
...
+@observer
+export class CatalogAddButton extends React.Component<CatalogAddButtonProps> {
+  @observable protected isOpen = false;
+  protected menuItems = observable.array<CatalogEntityContextMenu>([]);
+
+  componentDidMount() {
+    const { category } = this.props;
+
+    disposeOnUnmount(this, [
+      reaction(() => category, (category) => {
```
jakolehm: For some reason this does not react at all when category changes.

jakolehm:
Of course this broke integration tests... will fix asap.

src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -97,6 +97,20 @@ export class KubernetesClusterCategory extends EventEmitter implements CatalogCa
...
+    this.on("onCatalogAddMenu", (ctx: CatalogEntityAddMenuContext) => {
+      ctx.menuItems.push({
+        icon: "text_snippet",
+        title: "Add from kubeconfig",
+        onClick: async () => {
+          ctx.navigate("/add-cluster");
+        }
+      });
+    });
+  }
```
Nokel81, line 112: I think it would be a better design (and would prevent malicious extensions) if this function had to return the items it wished to add, and we do the combining ourselves.

src/renderer/components/+catalog/catalog-add-button.tsx
```diff
@@ -0,0 +1,76 @@
...
+        onClose={this.onClose}
+        icon={<Icon material="add" />}
+        direction="up"
+      >
+        { this.menuItems.map((menuItem, index) => {
+          return <SpeedDialAction
+            key={index}
+            icon={<Icon material="text_snippet" />}
+            tooltipTitle="Import Kubeconfig"
+            onClick={() => menuItem.onClick()}
```
Nokel81, line 68: Given that menu items can ask for confirmation, shouldn't we support that here?

src/renderer/components/+catalog/catalog-add-button.tsx
aleksfront: Just use `this.props.category` directly here without upper line
```
const { category } = this.props;
```

src/renderer/components/+catalog/catalog-add-button.tsx
aleksfront: Can you elaborate please? What confirmation should be supported?

Nokel81: See what is done on `src/renderer/components/+catalog/catalog.tsx:83` and `src/renderer/components/hotbar/hotbar-icon.tsx:72`.

I think that this is a good reason for having a registry that maps from `CatalogEntityContextMenu` to `ResolvedCatalogEntityContextMenu` so that the code here works as intended.

src/common/catalog-entities/kubernetes-cluster.ts
jakolehm: I think the current model already supports "sandboxing" of menuItems (if we want to implement it). Basically we could give an empty menuItems to each callback and merge them together.

src/renderer/components/+catalog/catalog-add-button.tsx
jakolehm: IMHO a bit theoretical use case, we can open an issue about confirmation and implement it later (if there is demand).

jakolehm:
Integration tests should now pass 🤞

nevalla:
> Integration tests should now pass 🤞

They did not 😿

src/common/catalog-entities/kubernetes-cluster.ts
Nokel81: True but from an API perspective (imo) it feels more intuitive to return what I want to add instead of modifying my own input.

jakolehm: I would like to get more feedback on this. How about if we keep current implementation and change it based on feedback (if needed, in follow-up PR)?

jakolehm:
> > Integration tests should now pass 🤞
> 
> They did not 😿

Integration tests should now pass 🤞 🤞

nevalla (approved):
LGTM

aleksfront (approved):

PR #2501, opened 2021-04-09
Broadcast failure to list namespaces less frequently

### LINKED ISSUE #2436: Add Accessible Namespaces notification shown too often
Occasionally the "Add Accessible Namespaces" notification appears unexpectedly (no RBAC set, user has namespace get/list permissions). This alone may be acceptable, under some circumstances the namespace query may fail(?) But the notification should not continually pop up after being dismissed. The namespace query should be checked again and the notification redisplayed only if it fails again.

Screen Shot 2021-04-01 at 7 07 33 PM

Lens 4.2.0, any OS

Nokel81:
- Only of three successive (non-overlapping) list failures

fixes #2436

### DIFF
--- a/src/main/cluster.ts
+++ b/src/main/cluster.ts
@@ -677,6 +677,8 @@ export class Cluster implements ClusterModel, ClusterState {
     };
   }
 
+  protected getAllowedNamespacesErrorCount = 0;
+
   protected async getAllowedNamespaces() {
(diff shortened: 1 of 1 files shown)


nevalla (approved):
LGTM

jakolehm (approved):

PR #2509, opened 2021-04-12
Test using GH Actions

### LINKED ISSUE #2508: Move test automation from Azure to Github Actions

jakolehm:
Fixes #2508

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,71 @@
+name: Test
+on:
+  - pull_request
+jobs:
+  build:
+    name: Test
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-16.04, macos-10.15, windows-2019]
+        node-version: [12.x]
+    steps:
+      - name: Checkout Release from lens
+        uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,71 @@
...
+          curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64
+          sudo install minikube-linux-amd64 /usr/local/bin/minikube
+          sudo minikube start --driver=none
+          # Although the kube and minikube config files are in placed $HOME they are owned by root
+          sudo chown -R $USER $HOME/.kube $HOME/.minikube
+        name: Install integration test dependencies
+        if: runner.os == 'Linux'
+      - run: |
+          set -e
+          rm -rf extensions/telemetry
```
nevalla, line 50: just thinking could `rm -rf extensions/telemetry` and `git checkout extensions/telemetry` be separate steps so we don't need to duplicate those on every environment?

jakolehm: They could but I tried to keep changes minimal compared to Azure. Anyway we probably will move this extension to a separate repository.

nevalla (approved):

Nokel81 (approved):

PR #2515, opened 2021-04-12
Fix releases not reloading when selecting namespaces 

### LINKED ISSUE #2239: Regression 4.0.8 -> 4.1.3 Cannot list helm releases on OpenShift 3.11
**Describe the bug**
After the upgrade from 4.0.8 to 4.1.3 i cannot get information about my Helm Releases.
The spinner spins forever.

**To Reproduce**
1. Click on Apps
2. Click on Releases
3. Spinner spins forever

**Expected behavior**
List of my Helm Releases

- Lens Version: 4.1.3
- OS: Windows 10
exe

**Logs:**
```
Your logs go here...
```
Any hints how i can get log output for this issue?

Nokel81:

fixes #2239 

As for UI changes, the Header of `ItemListLayout` is not always displayed even while loading. Only the rows of data are conditionally rendered (with a spinner other wise).

### DIFF
--- a/src/renderer/components/+apps-releases/release.store.ts
+++ b/src/renderer/components/+apps-releases/release.store.ts
@@ -66,7 +66,9 @@ export class ReleaseStore extends ItemStore<HelmRelease> {
 
       this.items.replace(this.sortItems(items));
       this.isLoaded = true;
+      this.failedLoading = false;
     } catch (error) {
+      this.failedLoading = true;
       console.error("Loading Helm Chart releases has failed", error);
 
       if (error.error) {
--- a/src/renderer/components/+apps-releases/release.store.ts
+++ b/src/renderer/components/+apps-releases/release.store.ts
@@ -57,7 +57,6 @@ export class ReleaseStore extends ItemStore<HelmRelease> {
 
   @action
   async loadAll(namespaces: string[]) {
-    console.log("reloading releases", namespaces);
     this.isLoading = true;
     this.isLoaded = false;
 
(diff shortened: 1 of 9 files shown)


jim-docker:
If I don't have access to secrets, Lens correctly reports that it can't list secrets, but the spinner spins forever instead of saying "item list is empty".

Screen Shot 2021-04-14 at 12 43 17 PM

Nokel81:
@jim-docker So you think it would be better to say "0 items in list"? How about, "failed to list items"?

jim-docker:
> @jim-docker So you think it would be better to say "0 items in list"? How about, "failed to list items"?

I think most important is that the spinner goes away. "item list is empty" is what is shown when there really are no items. But "failed to list items" would be more precise. Theoretically someone could have RBAC configured so that the user can see some secrets but not all? The releases list might be partial as a result? We could say "failed to list some items"?

Nokel81:
@jim-docker I think you can restrict listing of secrets by namespace. So I think your alternative option is best.

jim-docker (approved):

src/renderer/components/+apps-releases/release.store.ts
```diff
@@ -10,64 +10,65 @@ import { Notifications } from "../notifications";
...
-
-    return secretsStore.getByLabel(labels)
-      .filter(secret => secret.getNs() == release.getNs())[0];
+    })
+      .find(secret => secret.getNs() == release.getNs());
   }
 
   @action
   async loadAll(namespaces: string[]) {
+    console.log("reloading releases", namespaces);
```
aleksfront: Forgotten `console.log()`

aleksfront (commented):
Minor thing but noticed title blinking in releases page:
items blinking

aleksfront (approved):

PR #2520, opened 2021-04-13
MobX 6.2.0

### LINKED ISSUE #1536: Update MobX to 6.x
This will also enable previousValue for `reaction`, which can be handy.

jakolehm:
Attempt to upgrade from 5.x to 6.x. 

**ToDo:**
- [ ] fix random bugs
- [ ] fix unit tests

**Main pitfalls:**
- when using `mobx` decorators (e.g `@observable`) class constructor must include `makeObservable(this)`
-  don't use`mobx.toJS()` with non-observable data structure (use `common/utils/toJS` wrapper instead)
- don't apply `@action` for overriding class methods or use extra `@override` from `mobx` package
- don't use `@observable static` for classes (there is no way to wrap it with `makeObservable`)
-  don't use `ObservableValue.toJS()` and `{recursiveEverything: true}` (removed in `mobx@6`)

**More info:**
https://mobx.js.org/migrating-from-4-or-5.html

Fixes #1536


Nokel81:
Exciting

src/common/utils/autobind.ts
```diff
@@ -2,6 +2,7 @@
 // Can be applied to class or single method as @autobind()
 type Constructor<T = {}> = new (...args: any[]) => T;
 
+// FIXME-or-REMOVE: doesn't work as class-decorator with mobx-6 decorators, e.g. @action, @computed, etc.
```
jakolehm: My vote goes to remove.

src/extensions/extension-discovery.ts
```diff
@@ -341,7 +346,7 @@ export class ExtensionDiscovery extends Singleton {
     const userExtensions = await this.loadFromFolder(this.localFolderPath, bundledExtensions.map((extension) => extension.manifest.name));
 
     for (const extension of userExtensions) {
-      if (await fse.pathExists(extension.manifestPath) === false) {
+      if ((await fse.pathExists(extension.manifestPath)) === false) {
```
jakolehm, line 358: How did you spot this? Is this a bug in master branch?

src/renderer/api/endpoints/cron-job.api.ts
```diff
@@ -43,22 +43,6 @@ export class CronJob extends KubeObject {
   static namespaced = true;
   static apiBase = "/apis/batch/v1beta1/cronjobs";
 
-  kind: string;
```
jakolehm, line 46: Removed because dead-code?

src/renderer/api/endpoints/daemon-set.api.ts
```diff
@@ -10,7 +10,7 @@ export class DaemonSet extends WorkloadKubeObject {
   static namespaced = true;
   static apiBase = "/apis/apps/v1/daemonsets";
 
-  spec: {
+  declare spec: {
```
jakolehm, line 11: Why `declare`?

src/renderer/api/kube-object.ts
```diff
@@ -144,7 +144,6 @@ export class KubeObject implements ItemObject {
   apiVersion: string;
   kind: string;
   metadata: IKubeObjectMetadata;
-  status?: any; // todo: type-safety support
```
jakolehm, line 147: Should we keep this?

src/renderer/api/kube-watch-api.ts
```diff
@@ -29,14 +27,13 @@ export interface IKubeWatchLog {
   cssStyle?: string;
 }
 
-@autobind()
 export class KubeWatchApi {
   @observable context: ClusterContext = null;
 
   contextReady = when(() => Boolean(this.context));
 
-  isAllowedApi(api: KubeApi): boolean {
```
jakolehm, line 38: Dead code?

src/renderer/bootstrap.tsx
```diff
@@ -1,17 +1,17 @@
 import "./components/app.scss";
-
+import "../common/libs-config";
```
jakolehm: Just a note that import with side-effects (configure) might be a bad idea.

src/renderer/components/+preferences/add-helm-repo-dialog.tsx
```diff
@@ -27,22 +27,31 @@ enum FileType {
 
 @observer
 export class AddHelmRepoDialog extends React.Component<Props> {
-  private emptyRepo = {name: "", url: "", username: "", password: "", insecureSkipTlsVerify: false, caFile:"", keyFile: "", certFile: ""};
+  static metadata = observable({
+    isOpen: false,
```
jakolehm, line 31: Just wondering why we have to do these like this?

src/renderer/components/dock/dock.store.ts
```diff
@@ -125,18 +127,18 @@ export class DockStore implements DockStorageState {
     }
   }
 
-  @action
+  @action.bound
```
jakolehm: Why `bound`?

src/extensions/extension-discovery.ts
ixrock: I guess it was auto-correction from IDE :) And it seems like `()`-wrapping with `await` is redundant.

src/common/utils/autobind.ts
ixrock: I'm also for remove.

src/renderer/api/endpoints/cron-job.api.ts
ixrock: Because defined in inherited class (KubeObject)

src/renderer/api/endpoints/daemon-set.api.ts
ixrock: Because required by tsconfig.json `useDefineForClassFields: true` when overrides a field definition.

src/renderer/api/kube-watch-api.ts
ixrock: Yes

src/extensions/extension-discovery.ts
jakolehm: Right, somehow I read it wrong... it's just syntax (no functional difference).

src/renderer/bootstrap.tsx
ixrock: Do you any have suggestions how to configure packages  instead?
Currently `immer`, for example, configured inside `storageHelper.ts` which is even more bad idea since it's global configuration.

src/renderer/components/+preferences/add-helm-repo-dialog.tsx
ixrock: What exactly? Keeping observable state for single dialog instance? 
Not sure why about current dialog, but for example for `ConfirmDialog` observable state is required (it's rendered once in `[lens-]app.tsx`)

src/renderer/components/dock/dock.store.ts
ixrock: It's just `@action` + autobind. 
Useful when destructing methods from object with keeping context of `this`:

```ts
const {toggleFillSize} = dockStore;
toggleFillSize(); // context of this == dockStore
```

Probably this should be avoided and called always directly, e.g. `dockStore.toggle()`

src/renderer/components/+preferences/add-helm-repo-dialog.tsx
jakolehm: I meant that previously we had `@observable isOpen = false`, now we wrap these in `metadata`.

src/renderer/api/kube-object.ts
ixrock: Every `KubeObject` has this field? Could  keep, but then we need to add `declare status` in each child class with overriding `status`.

src/renderer/components/+preferences/add-helm-repo-dialog.tsx
ixrock: Because of `static` field. It's not possible to use it with `@observable` decorator.

Other option: `static isOpen = observable.box(false)`, but for consistency with other dialogs it's just an observable object with some metadata.

P.S. this is mentioned in PR's description (main pitfalls:)

src/renderer/bootstrap.tsx
ixrock: Or did you mean to extract code to function and clearly call it instead of "direct import evaluation"?

src/renderer/api/endpoints/daemon-set.api.ts
Nokel81: Wouldn't it be better to have `KubeApi` be generic over `Spec` and `Metadata`? I have started that work and makes things quite clean. Probably out of scope for this PR though.

src/renderer/api/kube-object.ts
Nokel81: See my above comment. `Status` would be a good third type to be generic over.

docs/extensions/guides/stores.md
```diff
@@ -42,18 +42,17 @@ export class ExamplePreferencesStore extends Store.ExtensionStore<ExamplePrefere
         enabled: false
       }
     });
+    makeObservable(this);
```
Nokel81, line 45: Shouldn't we stop using the annotation then?

src/common/utils/autobind.ts
Nokel81: Same here, I think we should remove all together and use the field initializer of an arrow function instead (we already do this in some places).

src/common/utils/cloneJson.ts
```diff
@@ -1,5 +1,5 @@
 // Clone json-serializable object
 
-export function cloneJsonObject<T = object>(obj: T): T {
-  return JSON.parse(JSON.stringify(obj));
+export function cloneJson<T>(data: T): T {
+  return JSON.parse(JSON.stringify(data));
```
Nokel81, line 4: This is wrong, the output should be something akin to `Jsonify<T>` (though without symbol keys as those are ignored):

Shouldn't we just use `lodash.clone` instead?

```typescript
// This converts an object type to a union of {key, value} types.
// (Thanks, Titian! https://stackoverflow.com/a/46101222/388951)
type AllValues<T extends object> = {
  [P in keyof T]: {key: P; value: T[P]};
}[keyof T];

// This filters an object type down to just those properties
// whose corresponding value types are not assignable to V.
type FilterValues<T extends object, V> = Pick<
  T,
  Exclude<AllValues<T>, { value: V }>['key']
>;

/** If x is T, then JSON.parse(JSON.stringify(x)) is Jsonify<T> */
type Jsonify<T> = T extends { toJSON(): infer U }
  ? Jsonify<U>
  : (
    T extends bigint
      ? never
      : (
        T extends Array<infer U>
          ? Jsonify<U>[]
          : (
            T extends object
              ? {
                  [k in keyof FilterValues<T, Function>]: Jsonify<T[k]>;
                }
              : T
          )
      )
  );
```

src/common/utils/toJS.ts
```diff
@@ -0,0 +1,13 @@
...
+import * as mobx from "mobx";
+import { isObservable, observable } from "mobx";
+
+export function toJS<T>(data: T): T {
+  if (isObservable(data)) {
+    return mobx.toJS(data);
+  }
+
+  // make data observable for recursive toJS()-call
+  return mobx.toJS(observable.box(data).get());
```
Nokel81, line 12: This works for deep values (more than one layer deep)?

src/extensions/extension-discovery.ts
```diff
@@ -66,7 +65,13 @@ export class ExtensionDiscovery extends Singleton {
   // IPC channel to broadcast changes to extension-discovery from main
   protected static readonly extensionDiscoveryChannel = "extension-discovery:main";
 
-  public events = new EventEmitter();
+  public events: EventEmitter;
```
Nokel81, line 72: What was wrong with assigning this here?

src/extensions/extension-loader.ts
```diff
@@ -311,10 +317,7 @@ export class ExtensionLoader extends Singleton {
   }
 
   toJSON(): Map<LensExtensionId, InstalledExtension> {
-    return toJS(this.extensions, {
-      exportMapsAsObjects: false,
-      recurseEverything: true,
-    });
+    return toJS(this.extensions);
```
Nokel81, line 320: Isn't the type of this wrong? I though `mobx6` returned arrays from `toJS` calls on `observable.map`'s.

src/extensions/registries/page-menu-registry.ts
```diff
@@ -22,7 +21,6 @@ export interface PageMenuComponents {
 }
 
 export class PageMenuRegistry<T extends PageMenuRegistration> extends BaseRegistry<T> {
-  @action
```
Nokel81, line 25: Would I be able to get an explanation as to why this needed to be removed? Is it because `BaseRegistry` already marks `add` as an action?

src/main/cluster-manager.ts
```diff
@@ -139,7 +145,9 @@ export class ClusterManager extends Singleton {
         message: "",
         active: !cluster.disconnected
       }
-    }));
+    });
+
+    return new KubernetesCluster(data as any);
```
Nokel81: why the cast?

src/main/extension-filesystem.ts
```diff
@@ -50,9 +52,7 @@ export class FilesystemProvisionerStore extends BaseStore<FSProvisionModel> {
 
   toJSON(): FSProvisionModel {
     return toJS({
-      extensions: this.registeredExtensions.toJSON(),
-    }, {
-      recurseEverything: true
+      extensions: Object.fromEntries(this.registeredExtensions),
```
Nokel81, line 55: I guess we might want to transition to using arrays here in the future (after 5.0)

src/renderer/bootstrap.tsx
Nokel81: I think it would be better to call it explicitly

src/renderer/components/+apps-releases/release-rollback-dialog.tsx
```diff
@@ -16,24 +16,31 @@ interface Props extends DialogProps {
 
 @observer
 export class ReleaseRollbackDialog extends React.Component<Props> {
-  @observable static isOpen = false;
-  @observable.ref static release: HelmRelease = null;
+  static metadata = observable({
+    isOpen: false,
+    release: null as HelmRelease,
+  });
```
Nokel81, line 22: Isn't `observable` generic over its input? Then this could be:


```suggestion
interface Metadata {
  isOpen: boolean;
  release: HelmRelease | null;
}

...

  static metadata = observable<Metadata>({
    isOpen: false,
    release: null,
  });
```

src/renderer/components/+catalog/catalog-add-button.tsx
```diff
@@ -64,14 +69,14 @@ export class CatalogAddButton extends React.Component<CatalogAddButtonProps> {
         open={this.isOpen}
         onOpen={this.onOpen}
         onClose={this.onClose}
-        icon={<Icon material="add" />}
```
Nokel81, line 67: I think we need to set a `eslint` rule for this. Because they keep getting change back and forth. I would argue for keeping the space because we tend to do that on object literals.

src/renderer/components/+cluster/cluster-overview.store.ts
```diff
@@ -21,17 +21,24 @@ export interface ClusterOverviewStorageState {
...
-    metricType: MetricType.CPU, // setup defaults
-    metricNodeRole: MetricNodeRole.WORKER,
-  });
+  constructor() {
+    super();
+
+    this.storage = createStorage<ClusterOverviewStorageState>("cluster_overview", {
+      metricType: MetricType.CPU, // setup defaults
+      metricNodeRole: MetricNodeRole.WORKER,
+    });
```
Nokel81, line 38: what is wrong with assigning to the field directly?

src/renderer/components/+config-maps/config-map-details.tsx
```diff
@@ -20,26 +20,21 @@ interface Props extends KubeObjectDetailsProps<ConfigMap> {
...
-
-        if (configMap) {
-          this.data.replace(configMap.data); // refresh
-        }
-      })
-    ]);
-  }
+  @disposeOnUnmount
+  autoCopyData = reaction(() => this.props.object, configMap => {
+    this.data = configMap?.data ?? {}; // copy-or-update config-map's data for editing
```
Nokel81, line 27: Why switch from a map? Also is the `observable` annotation deep by default even with assignments?

src/renderer/components/+workloads-statefulsets/statefulset-scale-dialog.tsx
```diff
@@ -16,24 +16,31 @@ interface Props extends Partial<DialogProps> {
...
+  constructor(props: Props) {
+    super(props);
+    makeObservable(this);
+  }
+
   static open(statefulSet: StatefulSet) {
-    StatefulSetScaleDialog.isOpen = true;
-    StatefulSetScaleDialog.data = statefulSet;
+    StatefulSetScaleDialog.metadata.isOpen = true;
+    StatefulSetScaleDialog.metadata.data = statefulSet;
```
Nokel81, line 35: FYI you could have kept these like this with static getters too

src/renderer/components/dock/dock.store.ts
```diff
@@ -28,17 +28,22 @@ export interface DockStorageState {
...
+  }
 
-  private storage = createStorage<DockStorageState>("dock", {
-    height: 300,
-    tabs: [
-      { id: "terminal", kind: TabKind.TERMINAL, title: "Terminal" },
-    ],
-  });
+  private storage: StorageHelper<DockStorageState>;
+  public readonly minHeight = 100;
```
Nokel81: should this be static?

src/renderer/components/hotbar/hotbar-icon.tsx
```diff
@@ -53,10 +58,12 @@ export class HotbarIcon extends React.Component<Props> {
   }
 
   @computed get iconString() {
-    const [rawFirst, rawSecond, rawThird] = getNameParts(this.props.entity.metadata.name);
+    const { name, labels } = this.props.entity.metadata;
```
Nokel81, line 61: Isn't `name` a required field on metadata?

src/renderer/item.store.ts
```diff
@@ -1,13 +1,11 @@
-import orderBy from "lodash/orderBy";
-import { autobind, noop } from "./utils";
-import { action, computed, observable, when } from "mobx";
+import { action, computed, makeObservable, observable, toJS, when } from "mobx";
+import { noop, orderBy } from "lodash";
```
Nokel81, line 2: Why are you using `lodash`'s `noop`?

src/common/utils/toJS.ts
ixrock: Yes. (`opts = {recursiveEverything: true}` is removed, but value must be observable, otherwise return as-is)

src/extensions/extension-discovery.ts
ixrock: Don't remember for sure. Most probably came from multiple `master`-merges with conflicts.

src/renderer/components/+workloads-statefulsets/statefulset-scale-dialog.tsx
ixrock: And keep the state in some `static data = observable.box({})`?  It would be more cumbersome imho..
Or did you mean something different?

src/renderer/item.store.ts
ixrock: Because of `lodash/orderBy` is used.🤔 (or it was auto-import from IDE, not sure)

src/renderer/components/+workloads-statefulsets/statefulset-scale-dialog.tsx
Nokel81: I mean the following:

```ts
static metadata = observable({
	  isOpen: false,
	  data: null as StatefulSet,
});

get isOpen() {
	return StatefulSetScaleDialog.metadata.isOpen;
}

set isOpen(val) {
	StatefulSetScaleDialog.metadata.isOpen = val;
}

get data() {
	return StatefulSetScaleDialog.metadata.data;
}

set data(val) {
	StatefulSetScaleDialog.metadata.data = val;
}
```

This way you change this top thing and not have all the other changes sprinkled throughout the file. Just a thought

src/renderer/item.store.ts
Nokel81: I think you removed all of the old line 2 instead of just the `autobind` one

ixrock:
rebased to https://github.com/lensapp/lens/pull/2718

Issue #2412, opened 2021-03-30
Multiple log whitespace is being truncated in the terminal output


firefixmaarten:
**Describe the bug**
Logs whitespace is being scrambled in terminal output if I force color output using logrus (https://github.com/sirupsen/logrus)

**To Reproduce**
1. use a logrus configuration in the style of:
``` 
// Make sure to log the method
logrus.SetReportCaller(true)

// Some format options
logrus.SetFormatter(&log.TextFormatter{
  ForceColors:            true,
  FullTimestamp:          true,
  DisableLevelTruncation: true,
  PadLevelText:           true,
})
```
2. see the output using kubectl in the terminal
![image](https://user-images.githubusercontent.com/2439070/112963989-30f57e80-9148-11eb-9a6d-a81bc6e2cbe4.png)
3. Compare with the lens buildin terminal
![image](https://user-images.githubusercontent.com/2439070/112964184-613d1d00-9148-11eb-9d77-b7aa5b16a886.png)

**Expected behavior**
The output to retain the original white spacing like in step 2.
Seems like some non asci bytes might cause the issue (this is in vscode after downloading the log), but at least the space chars are all there?
![image](https://user-images.githubusercontent.com/2439070/112965846-f0970000-9149-11eb-948d-8736bbac68d9.png)

**Environment (please complete the following information):**
- Lens Version: 4.1.4
- OS: Fedora 33
- Installation method (e.g. snap or AppImage in Linux): snap

Nokel81:
Question, are the all "space" characters, or could they also be `\t`? I have a sneaking suspicion that logrus (when I used it last) uses both. 

If that is the case, it would explain why INFO lines seem to loose more whitespace.

firefixmaarten:
The downloaded file seems to indicated these are just pure spaces. (ignore the colored green text, it just a style thing)
![image](https://user-images.githubusercontent.com/2439070/113822207-6e2bc300-977d-11eb-8249-a1b0625395aa.png)

As I am now using logfmt style to do processing, I stopped using colors anyway, but I guess the bug could still be fixed for others.

firefixmaarten:
Also, it is also the case in the parts not shown here (without color) after the file etc, multiple spaces are reduced to a single space somehow.
No picture of that anymore, sorry.

Nokel81:
Thanks for the info.

Nokel81:
Fixed by #2655

PR #2523, opened 2021-04-13
Add ability to configure the locale timezone

### LINKED ISSUE #569: Support timezone, especially local timezone default

Support timezone, especially local timezone default

**Why is this needed**:

All time is formatted as **RFC 3339** and **+0** timezone, such as `Created` and `Event Time` field. So it's very inconvenient
to check time.

Lens app shows as follow:

`Created        5d 22h 53m ago (2020-07-04T04:48:04Z)`

executing the command 'kubectl describe pod' shows as follow:

`Start Time:   Sat, 04 Jul 2020 12:48:04 +0800`

- Kubernetes distribution: K8s 1.17.5 with kubespray
- Desktop OS: macOS Mojave 10.14.6
- Lens Version: 3.5.1

### LINKED ISSUE #2397: how to adjust lens timezone?
how to adjust lens timezone?
add some settings in cluster configuration about timezone

**Why is this needed**:
k8s log time and the host time default using utc , i changed host timezone ,but change noting in lens.

- Kubernetes distribution: [kubernetes 1.20.5]
- Desktop OS: <Windows>

ArthurKnoep:
This PR adds the possibility to configure the locale timezone of the user in the preferences of Lens. Then, all the date displayed in the application (ressources creation, events date, logs, ...) should follow this timezone.
By default, it uses the user's timezone or fallback to UTC if it cannot be determined.

Closes #2397 
Closes #569

As a video is worth a thousand words:

https://user-images.githubusercontent.com/27413532/114598157-dc96e680-9c91-11eb-8155-e930287d4618.mp4

### DIFF
--- a/src/renderer/components/dock/log-list.tsx
+++ b/src/renderer/components/dock/log-list.tsx
@@ -87,9 +87,9 @@ export class LogList extends React.Component<Props> {
       return logStore.logsWithoutTimestamps;
     }
 
-    return this.props.logs.map(log => (
-      `${moment.tz(logStore.getTimestamps(log)[0] || "", preferences.localeTimezone).format()} ${logStore.removeTimestamps(log)}`
-    ));
+    return this.props.logs
+      .map(log => logStore.splitOutTimestamp(log))
+      .map(([logTimestamp, log]) => (`${moment.tz(logTimestamp, preferences.localeTimezone).format()} ${log}`));
   }
 
   /**
--- a/src/renderer/components/dock/log-list.tsx
+++ b/src/renderer/components/dock/log-list.tsx
@@ -6,9 +6,11 @@ import DOMPurify from "dompurify";
 import debounce from "lodash/debounce";
 import { action, computed, observable } from "mobx";
 import { observer } from "mobx-react";
+import moment from "moment-timezone";
 import { Align, ListOnScrollProps } from "react-window";
 
 import { SearchStore, searchStore } from "../../../common/search-store";
+import { userStore } from "../../../common/user-store";
 import { cssNames } from "../../utils";
 import { Button } from "../button";
 import { Icon } from "../icon";
@@ -79,12 +81,15 @@ export class LogList extends React.Component<Props> {
   @computed
   get logs() {
     const showTimestamps = logTabStore.getData(this.props.id).showTimestamps;
+    const { preferences } = userStore;
 
     if (!showTimestamps) {
       return logStore.logsWithoutTimestamps;
     }
 
-    return this.props.logs;
+    return this.props.logs.map(log => (
+      `${moment.tz(logStore.getTimestamps(log)[0] || "", preferences.localeTimezone).format()} ${logStore.removeTimestamps(log)}`
+    ));
   }
 
   /**
--- a/src/renderer/components/dock/log-list.tsx
+++ b/src/renderer/components/dock/log-list.tsx
@@ -89,7 +89,7 @@ export class LogList extends React.Component<Props> {
 
     return this.props.logs
       .map(log => logStore.splitOutTimestamp(log))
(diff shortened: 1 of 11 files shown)


Nokel81:
@ArthurKnoep please sign all of your commits to pass DCO

ArthurKnoep:
> @ArthurKnoep please sign all of your commits to pass DCO

Yep yep, it's planned, I just went to eat 😄

src/renderer/components/dock/log-list.tsx
```diff
@@ -79,12 +81,15 @@ export class LogList extends React.Component<Props> {
...
+    const { preferences } = userStore;
 
     if (!showTimestamps) {
       return logStore.logsWithoutTimestamps;
     }
 
-    return this.props.logs;
+    return this.props.logs.map(log => (
+      `${moment.tz(logStore.getTimestamps(log)[0] || "", preferences.localeTimezone).format()} ${logStore.removeTimestamps(log)}`
+    ));
```
Nokel81: I think that logStore should have a function "splitOutTimestamp" which returns the "timestamp" string and the "rest" string. Would you mind adding that?

Nokel81 (commented):
Looks great, nice a clean. I guess if you have missed any it will be very trivial to fix in the future.

Thanks for doing this.

I have just one comment.

src/renderer/components/dock/log-list.tsx
ArthurKnoep: done

src/renderer/components/dock/log.store.ts
```diff
@@ -146,6 +146,16 @@ export class LogStore {
     return stamp.toISOString();
   }
 
+  splitOutTimestamp(logs: string): [string, string] {
+    const extraction = /^(\d+\S+) (.*)/gm.exec(logs);
+
+    if (extraction.length < 3) {
+      return ["", ""];
+    }
```
Nokel81: exec can return null
```suggestion
    if (!extraction || extraction.length < 3) {
      return ["", ""];
    }
```

src/renderer/components/dock/log.store.ts
```diff
@@ -146,6 +146,16 @@ export class LogStore {
     return stamp.toISOString();
   }
 
+  splitOutTimestamp(logs: string): [string, string] {
+    const extraction = /^(\d+\S+) (.*)/gm.exec(logs);
```
Nokel81: I don't think this needs to be global. Also wouldn't I don't think that space should be there between the two groups.

ArthurKnoep: But if I don't match the space between the two groups it will end up in my second capture group, which would require me to trim the string afterward

Nokel81: Wouldn't that be a better solution? Since their might be spaces before hand? Though TBH #2412 implies that we shouldn't be trimming at all.

ArthurKnoep: Ok, so I've removed the space in the regex and in consequence of that I've also removed the space in the log formatter (https://github.com/lensapp/lens/pull/2523/commits/dbc44baea28d982d96ba378409ae62a76b0b15ca#diff-92bc22dec3f8bbca956be35e96f4152145e3b798349bf5ae04d5cf38e94cc854R92) so that I won't get an extra space.
However I'm not sure if that will fix #2412 as I'm just displaying what the kube API returns me

Nokel81 (dismissed):

src/renderer/components/dock/log.store.ts
Nokel81: Thanks, yeah I don't know if that will fix that entirely, but it might.

Nokel81 (approved):

jakolehm (approved):
LGTM

PR #2525, opened 2021-04-14
Refactor cluster settings to catalog entity settings

jakolehm:
Refactors old cluster settings logic to more generic catalog entity settings. All items within settings menu are registered through API that is also exposed to extensions.

Also changes page layout to look more like Preferences.

image

image


src/renderer/components/layout/page-layout.tsx
```diff
@@ -8,8 +8,6 @@ import { Icon } from "../icon";
 
 export interface PageLayoutProps extends React.DOMAttributes<any> {
   className?: IClassName;
-  header?: React.ReactNode;
```
jakolehm, line 11: These aren't used anywhere.

src/renderer/components/+cluster-settings/cluster-settings.tsx
```diff
@@ -49,20 +61,94 @@ export class ClusterSettings extends React.Component<Props> {
...
-      <PageLayout className="ClusterSettings" header={header} showOnTop={true}>
-        <Status cluster={cluster}></Status>
-        <General cluster={cluster}></General>
+      <PageLayout
+        className="ClusterSettings"
+        navigation={this.renderNavigation()}
+        showOnTop={true}
+        contentGaps={false}
+      >
+        {this.activeTab == "General" && (
```
Nokel81: This doesn't seem very extendable. Would it be better to render all of them and then switch one one is "active" using `css-grid` or something?

jakolehm: I was mainly following how "Preferences" is implemented. I will open more issues related to this page once we have the layout in place (this PR merged) ... extensions should be able to hook into this page with new tabs.

Nokel81: So we will be moving to that sort of design later?

jakolehm: I was talking about dynamic "tabs" on PageLayout.. currently PageLayout is quite static but follow-up PR should add extension api that allows to add items into "catalog entity settings page". If that requires changes to PageLayout then we have to move... but I'm not sure yet.

Nokel81: I meant more how we are rendering cluster settings, since I think that it is possible for extensions to add them but we are no longer rendering them.

jakolehm: Ok, that will change. In the past extensions could only add "cluster features" but now they should be able to add almost anything into this page. ("cluster features" were removed already in #2418 ).

src/renderer/components/+cluster-settings/components/cluster-kubeconfig.tsx
```diff
@@ -0,0 +1,34 @@
...
+    shell.showItemInFolder(cluster.kubeConfigPath);
+  }
+
+  render() {
+    return (
+      <>
+        <SubTitle title="Kubeconfig" />
+
+        <span>
+          <a className="link value" onClick={this.openKubeconfig}>{this.props.cluster.kubeConfigPath}</a>
```
jim-docker: Seems a bit odd that clicking on a link brings up a file select dialog. So user can view it?

jim-docker (dismissed):
Are we ditching most of the status info? Or moving it?

src/renderer/components/+cluster-settings/components/cluster-kubeconfig.tsx
jakolehm: It has been link also in the past. Link allows user to find/edit kubeconfig easily.

image

src/renderer/components/+cluster-settings/cluster-settings.tsx
jakolehm: Refactored to "catalog entity settings" -> pluggable.

src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -49,8 +49,8 @@ export class KubernetesCluster implements CatalogEntity {
       {
         icon: "settings",
         title: "Settings",
-        onlyVisibleForSource: "local",
-        onClick: async () => context.navigate(`/cluster/${this.metadata.uid}/settings`)
+        onClick: async () => context.navigate(`/entity/${this.metadata.uid}/settings`),
+        onlyVisibleForSource: "local"
```
Nokel81: Can you revert this order please?

src/extensions/registries/entity-setting-registry.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export interface RegisteredEntitySetting extends EntitySettingRegistration {
+  id: string;
+}
+
+export class EntitySettingRegistry extends BaseRegistry<EntitySettingRegistration, RegisteredEntitySetting> {
+  getRegisteredItem(item: EntitySettingRegistration): RegisteredEntitySetting {
+    return {
+      id: item.id || item.title.toLowerCase().replace(/[^0-9a-zA-Z]+/g, "-"),
```
Nokel81: Why replace the title's beginning?

src/extensions/registries/entity-setting-registry.ts
```diff
@@ -0,0 +1,49 @@
...
+      ...item,
+    };
+  }
+
+  getItemsForKind(kind: string, apiVersion: string, source?: string) {
+    const items = this.getItems().filter((item) => {
+      return item.kind === kind && item.apiVersions.includes(apiVersion);
+    });
+
+    if (source && source !== "") {
```
Nokel81: This is the same and is clearer imo

```suggestion
    if (source) {
```

src/extensions/registries/entity-setting-registry.ts
```diff
@@ -0,0 +1,49 @@
...
+      return items.filter((item) => {
+        return !item.source || item.source === source;
```
Nokel81, line 41: Shouldn't source be a required field?

src/main/menu.ts
```diff
@@ -112,19 +101,6 @@ export function buildMenu(windowManager: WindowManager) {
...
-          accelerator: "CmdOrCtrl+Shift+S",
-          click() {
-            navigate(clusterSettingsURL({
-              params: {
-                clusterId: windowManager.activeClusterId
-              }
-            }));
-          }
-        }
-      ]),
```
Nokel81, line 127: We should have entity settings with the same accelerator

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -44,6 +44,10 @@ export class CatalogEntityRegistry {
     return this._items;
   }
 
+  getById(id: string) {
+    return this._items.find((entity) => entity.metadata.uid === id);
+  }
```
Nokel81, line 49: The fact that this code is being duplicated seems to imply that there is a better design for these registries.

src/renderer/components/+entity-settings/entity-settings.route.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export interface EntitySettingsRouteParams {
+  entityId: string;
+}
+
+export const entitySettingsRoute: RouteProps = {
+  path: `/entity/:entityId/settings`,
+};
+
+export const entitySettingsURL = buildURL<EntitySettingsRouteParams>(entitySettingsRoute.path);
```
Nokel81, line 12: Something that I have discovered while working on the kube sync feature is that these really should be in `common/routes` (a new folder I made) so that all code and include it and we don't have "magic strings" of these paths littered around the code.

It has to be in `common` because `webpack` will fail to compile if `main` includes this directly from `renderer`

src/renderer/components/+entity-settings/entity-settings.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  renderNavigation() {
+    return (
+      <>
+        <h2>{this.entity.metadata.name}</h2>
+        <Tabs className="flex column" scrollable={false} onChange={this.onTabChange} value={this.activeTab}>
+          <div className="header">Settings</div>
+          { this.menuItems.map((setting) => {
+            return <Tab key={setting.id} value={setting.id} label={setting.title} data-testid={`${setting.id}-tab`} />;
+          })}
```
Nokel81:
```suggestion
          {
          	this.menuItems.map(setting => (
          	  <Tab 
          	    key={setting.id} 
          	    value={setting.id} 
          	    label={setting.title} 
          	    data-testid={`${setting.id}-tab`} 
          	  />
          	))
          }
```

src/renderer/components/+entity-settings/entity-settings.tsx
```diff
@@ -0,0 +1,88 @@
...
+        </Tabs>
+      </>
+    );
+  }
+
+  render() {
+    if (!this.entity) return null;
```
Nokel81: If we have entered this state, should we not log something to the console?

src/renderer/components/+entity-settings/entity-settings.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+    const activeSetting = this.menuItems.find((setting) => setting.id === this.activeTab);
+
+    if (!activeSetting) {
+      return null;
+    }
```
Nokel81: We should default to the first setting page.

src/renderer/components/cluster-settings/cluster-settings.tsx
```diff
@@ -0,0 +1,130 @@
...
+    kind: "KubernetesCluster",
+    source: "local",
+    title: "General",
+    components: {
+      View: (props: { entity: CatalogEntity }) => {
+        const cluster = clusterStore.enabledClustersList.find((cluster) => cluster.id === props.entity.metadata.uid);
+
+        if (!cluster) {
+          return null;
+        }
```
Nokel81: I think it would be better to do the filter on enabled after finding by ID. Easier to refactor later (when we remove enabled) and more performant in the meantime.
```suggestion
        const cluster = clusterStore.getById(props.entity.metadata.uid);

        if (!cluster?.isEnabled) {
          return null;
        }
```

src/renderer/components/cluster-settings/components/cluster-home-dir-setting.tsx
```diff
@@ -33,7 +33,6 @@ export class ClusterHomeDirSetting extends React.Component<Props> {
     return (
       <>
         <SubTitle title="Working Directory"/>
-        <p>Terminal working directory.</p>
```
Nokel81, line 36: I think you should rename the subtitle then since working directory is an ambiguous phrase here.

src/renderer/components/cluster-settings/components/cluster-name-setting.tsx
```diff
@@ -34,7 +34,6 @@ export class ClusterNameSetting extends React.Component<Props> {
     return (
       <>
         <SubTitle title="Cluster Name" />
-        <p>Define cluster name.</p>
```
Nokel81, line 37: Same here as above, maybe "Cluster Display Name"?

src/renderer/components/layout/page-layout.scss
```diff
@@ -42,6 +42,16 @@
       width: 218px;
       padding: 60px 10px 60px 20px;
 
+      h2 {
+        margin-bottom: 10px;
+        font-size: 18px;
+        padding: 6px 10px;
+        overflow-wrap: anywhere;
+        color: var(--textColorAccent);
+        text-transform: uppercase;
```
Nokel81: I think that `capitalize` would be a better text-transform.

src/extensions/registries/entity-setting-registry.ts
jakolehm: TBH this was copied from AppPreferences ... most likely we can remove `replace`.

src/extensions/registries/entity-setting-registry.ts
jakolehm: Currently it's not (this applies also to other interfaces)... we could replace empty source with `*` if we want (basically empty source means "match all").

src/main/menu.ts
jakolehm: Active entity tracking is "wip" and out of scope for this PR. But I agree that we should add this back once it works.

src/extensions/registries/entity-setting-registry.ts
Nokel81: I think that would be better, and have the source being `/^[\w-]+$/` when provided would be allow us to have this sort of thing.

src/renderer/components/+entity-settings/entity-settings.route.ts
jakolehm: Makes sense, we probably should refactor this in a separate PR?

src/renderer/components/+entity-settings/entity-settings.tsx
jakolehm: I think we already are (in `componentDidMount`, line 40). This is just to protect render.

src/renderer/components/layout/page-layout.scss
jakolehm: You might be right, will test.

src/main/menu.ts
Nokel81: If we don't have a ticket for that we should.

src/renderer/components/+entity-settings/entity-settings.route.ts
Nokel81: Could work, we could do all of it at once in a separate PR.

Nokel81: Will make a ticket for that.

src/renderer/components/+entity-settings/entity-settings.tsx
Nokel81: Okay sounds good. I thought that if it happens again then we should reset to first again.

src/common/catalog-entities/kubernetes-cluster.ts
jakolehm: Done.

src/extensions/registries/entity-setting-registry.ts
jakolehm: Fixed.

src/extensions/registries/entity-setting-registry.ts
jakolehm: Fixed.

src/extensions/registries/entity-setting-registry.ts
jakolehm: Let's do that in a separate PR because it will affect interfaces that are out-of-scope for this PR.

src/renderer/api/catalog-entity-registry.ts
jakolehm: Removed duplicate in common because it wasn't actually used. But yes, I agree that we should find better design for this.

src/renderer/components/+entity-settings/entity-settings.tsx
jakolehm: Changed.

src/renderer/components/cluster-settings/cluster-settings.tsx
jakolehm: Refactored.

src/renderer/components/cluster-settings/components/cluster-home-dir-setting.tsx
jakolehm: I think the context is clear without repetition, see: 
image

src/renderer/components/cluster-settings/components/cluster-name-setting.tsx
jakolehm: IMHO "Cluster Name" is enough.

src/renderer/components/layout/page-layout.scss
jakolehm: Ended up to remove whole text-transform, does not make sense here.

src/extensions/registries/entity-setting-registry.ts
Nokel81: okay sounds good, please make a tracking issue.

src/renderer/components/cluster-settings/components/cluster-home-dir-setting.tsx
Nokel81: Yup, it is good.

src/renderer/components/cluster-settings/components/cluster-name-setting.tsx
Nokel81: okay

src/renderer/components/layout/page-layout.scss
Nokel81: sounds good.

Nokel81 (approved):

jakolehm:
@jim-docker @aleksfront ptal

src/extensions/extension-loader.ts
```diff
@@ -211,8 +211,8 @@ export class ExtensionLoader {
     this.autoInitExtensions(async (extension: LensRendererExtension) => {
       const removeItems = [
         registries.globalPageRegistry.add(extension.globalPages, extension),
-        registries.globalPageMenuRegistry.add(extension.globalPageMenus, extension),
```
jim-docker, line 214: Are we looking for a new home for these? hot bar items?

src/extensions/registries/entity-setting-registry.ts
```diff
@@ -0,0 +1,49 @@
...
+    const items = this.getItems().filter((item) => {
+      return item.kind === kind && item.apiVersions.includes(apiVersion);
+    });
+
+    if (source) {
+      return items.filter((item) => {
+        return !item.source || item.source === source;
+      });
+    } else {
+      return items;
```
jim-docker, line 44: I guess you don't need this in an `else` block

src/renderer/components/cluster-settings/cluster-settings.tsx
```diff
@@ -0,0 +1,141 @@
...
+    components: {
+      View: (props: { entity: CatalogEntity }) => {
+        const cluster = getClusterForEntity(props.entity);
+
+        if (!cluster) {
+          return null;
+        }
+
+        return (
+          <section>
```
jim-docker, line 40: This outer `<section>` is unnecessary because the `View` already gets rendered in a `<section>` in `EntitySettings`'s `render()`?

src/renderer/protocol-handler/app-handlers.ts
```diff
@@ -23,6 +24,19 @@ export function bindProtocolAddRouteHandlers() {
...
+        navigate(entitySettingsURL({ params: { entityId } }));
+      } else {
+        console.log("[APP-HANDLER]: catalog entity with given ID does not exist", { entityId });
+      }
+    })
+    .addInternalHandler("/extensions", () => {
+      navigate(extensionsURL());
+    })
+    // Handlers below are deprecated and only kept for backward compat purposes
     .addInternalHandler("/cluster/:clusterId", ({ pathname: { clusterId } }) => {
```
jim-docker, line 40: Will there be a comparable :
```
    .addInternalHandler("/entity/:entityId", ({ pathname: { entityId } }) => ...
```
?

src/renderer/protocol-handler/app-handlers.ts
```diff
@@ -23,6 +24,19 @@ export function bindProtocolAddRouteHandlers() {
     .addInternalHandler("/cluster", () => {
```
jim-docker, line 24: this would be deprecated too, in favour of something like "/entity/:kind"?

jim-docker (approved):

src/extensions/extension-loader.ts
jakolehm: You are right, these are basically replaced by hotbar items (no need for separate api).

src/renderer/protocol-handler/app-handlers.ts
jakolehm: Yes, we should add route/api for an entity (out of scope for this PR). While doing it we should probably refactor current cluster view to use it.

src/renderer/protocol-handler/app-handlers.ts
jakolehm: Not sure if we need replacement for this deprecated handler.

PR #2551, opened 2021-04-19
Fix initial hotbar not showing

jakolehm:
Initial hotbar was not initialized properly -> hotbar was completely missing from UI if no prior clusters existed (for migration).

### DIFF
--- a/src/common/hotbar-store.ts
+++ b/src/common/hotbar-store.ts
@@ -35,10 +35,14 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
   }
 
   @action protected async fromStore(data: Partial<HotbarStoreModel> = {}) {
-    this.hotbars = data.hotbars || [{
-      name: "default",
-      items: []
-    }];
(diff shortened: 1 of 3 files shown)


src/common/hotbar-store.ts
```diff
@@ -35,10 +35,14 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
   }
 
   @action protected async fromStore(data: Partial<HotbarStoreModel> = {}) {
-    this.hotbars = data.hotbars || [{
-      name: "default",
-      items: []
-    }];
+    if (data.hotbars?.length === 0) {
+      this.hotbars = [{
+        name: "default",
```
Nokel81, line 40: Is the name the same as ID? Are we allowed to delete the default hotbar? Or is that for future considerations?

jakolehm: See #2552 (which is follow-up for this PR).

Nokel81 (approved):

nevalla (approved):

PR #2552, opened 2021-04-19
Hotbar command palette + switching

jakolehm:
https://user-images.githubusercontent.com/1446224/115190789-3adc2300-a0f1-11eb-9513-6bab2ab626f4.mov

Requires #2551


jakolehm:
Added clickable index to prev/next switcher (see bottom left):

image

Click opens hotbar switch command palette.

src/common/hotbar-store.ts
```diff
@@ -34,32 +38,78 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
...
+    return this.getByid(this.activeHotbarId);
   }
 
   getByName(name: string) {
     return this.hotbars.find((hotbar) => hotbar.name === name);
   }
 
-  add(hotbar: Hotbar) {
-    this.hotbars.push(hotbar);
+  getByid(id: string) {
```
aleksfront: `getByid`-> `getById`

src/renderer/components/hotbar/hotbar-menu.tsx
```diff
@@ -25,8 +43,33 @@ export class HotbarMenu extends React.Component<Props> {
     return hotbar.items.map((item) => catalogEntityRegistry.items.find((entity) => entity.metadata.uid === item.entity.uid)).filter(Boolean);
   }
 
+  previous() {
```
aleksfront, line 32: Maybe better to move logic from `previous()` and `next()` methods into store itself. To keep clean `HotbarMenu` from store internals.

aleksfront (commented):
While testing hotbar switcher found some issues:
1. Clicking active hotbar number opens a command pallete with few strange `[object Object]` items.
hotbar object
And there's no "initial" or "default" hotbar in the list

2. Probably related to num 1 but `prev` & `next` buttons isn't working right
next prev buttons

3. Add to hotbar option not working! ⚠️ 
adding to hotbar broken

4. `Remove from hotbar` menu item should be disabled for items not listed inside hotbar.

jakolehm:
> While testing hotbar switcher found some issues:

@aleksfront I think most of these issues are because of hotbar-store didn't migrate properly. I'll tweak this PR to have a separate migration step that fixes existing hotbar stores... meanwhile you can just remove the json file before starting the app.

Nokel81:
Shouldn't the "fromStore" function do validation to make sure it is in the correct state?

jakolehm:
> Shouldn't the "fromStore" function do validation to make sure it is in the correct state?

Maybe but not sure if it's good idea to generate uuids on-the-fly if those are missing.. I think it would make more sense to create proper migration that fixes the persistent data here.

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "kontena-lens",
   "productName": "Lens",
   "description": "Lens - The Kubernetes IDE",
-  "version": "5.0.0-alpha.1",
```
Nokel81, line 5: Why is this changed in this PR?

src/common/hotbar-store.ts
```diff
@@ -1,6 +1,7 @@
 import { action, comparer, observable, toJS } from "mobx";
 import { BaseStore } from "./base-store";
 import migrations from "../migrations/hotbar-store";
+import { v4 as uuid } from "uuid";
```
Nokel81: nit: I prefer doing `import "uuid" from "uuid";` and then using it as `uuid.v4()` since that means that when looking at the uses I don't have to come back and checker which version we are importing.

src/migrations/hotbar-store/5.0.0-alpha.2.ts
```diff
@@ -0,0 +1,19 @@
...
+  run(store) {
+    const hotbars = (store.get("hotbars") || []) as Hotbar[];
+
+    hotbars.forEach((hotbar) => {
+      if (!hotbar.id) {
+        hotbar.id = uuid();
+      }
+    });
+
+    store.set("hotbars", hotbars);
```
Nokel81: nit: I find assigning to the input confusing when reading code.
```suggestion
    const rawHotbars = (store.get("hotbars") || []) as Hotbar[];

    store.set("hotbars", rawHotbars.map(({ id = uuid.v4(), ...rest }) => ({
      id,
      ...rest,
    }));
```

src/renderer/components/hotbar/hotbar-add-command.tsx
```diff
@@ -0,0 +1,48 @@
+import React from "react";
+import { observer } from "mobx-react";
+import { hotbarStore } from "../../../common/hotbar-store";
+import { CommandOverlay } from "../command-palette";
+import { Input, InputValidator } from "../input";
+
+const uniqueWorkspaceName: InputValidator = {
```
Nokel81: I don't think this should be called ...workspace...

src/renderer/components/hotbar/hotbar-menu.tsx
```diff
@@ -15,8 +22,19 @@ interface Props {
...
+  componentDidMount() {
+    disposeOnUnmount(this, [
+      reaction(() => hotbarStore.activeHotbarId, () => {
+        Notifications.info(`Hotbar "${hotbarStore.getActive().name}" is now active.`, {
+          id: "active-hotbar",
+          timeout: 5_000
+        });
+      })
+    ]);
+  }
```
Nokel81: I too think that this is a bit obnoxious. Would it be possible to do a sliding animation on the hotbar instead?

package.json
jakolehm: Migration is not executed without bumping version. Is there better way to achieve this?

Nokel81: ah I see, makes sense. I was just confused because this wasn't a release PR

src/common/hotbar-store.ts
jakolehm: Changed (although other parts of codebase are using "v4 as uuid").

src/migrations/hotbar-store/5.0.0-alpha.2.ts
jakolehm: Refactored a bit, your suggestion was a bit confusing to my brain so the code is now somewhere in the middle.

src/renderer/components/hotbar/hotbar-add-command.tsx
jakolehm: Good catch, fixed.

src/renderer/components/hotbar/hotbar-menu.tsx
jakolehm: Not sure how to do that without hacks. Maybe I'll just remove notifications from this PR?

src/renderer/components/hotbar/hotbar-menu.tsx
jakolehm: Refactored.

src/common/hotbar-store.ts
jakolehm: Good catch, fixed.

src/common/hotbar-store.ts
Nokel81: thanks, I know. For my code I tend to use it.

src/migrations/hotbar-store/5.0.0-alpha.2.ts
Nokel81: okay

src/renderer/components/hotbar/hotbar-menu.tsx
Nokel81: Sounds good probably a good idea to have it in a separate PR.

I would do it by rendering all of them but explicitly set the bounding box sizes and the `overflow-display: none` and then use JS powered css vars to change the animation placement. 

At least that is how I would start the impl attempt.

src/common/hotbar-store.ts
```diff
@@ -34,32 +38,98 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
...
+    hotbar.id = uuid.v4();
+
+    if (!hotbar.items) {
+      hotbar.items = [];
+    }
+
+    this.hotbars.push(hotbar as Hotbar);
+
+    return hotbar as Hotbar;
   }
```
Nokel81: You don't need the casts if you create the entry after deconstruction.

Also, except for `render*()` functions I think that the return value should be explicitly written out so that function definitions are fixed points. This also should help when changing the implementation of a function because typescript will fail to compile if you don't satisfy the contract.

Plus it seems like you want the `name` to still be required:

```suggestion
interface HotbarCreateOptions extends Hotbar {
  id?: string;
  items?: HotbarItem[];
}

  add(rawHotbar: HotbarCreateOptions): Hotbar {
    const {
      id = uuid.v4(), // initialize to a new ID if not provided
      items = [], // initialize to empty array if not provided
      name,
    } = rawHotbar;
    const hotbar = { id, items, name };

    this.hotbars.push(hotbar);

    return hotbar;
  }
```

src/common/hotbar-store.ts
```diff
@@ -34,32 +38,98 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
...
 
   remove(hotbar: Hotbar) {
```
Nokel81, line 110: this should be an action

src/renderer/components/hotbar/hotbar-add-command.tsx
```diff
@@ -0,0 +1,48 @@
+import React from "react";
+import { observer } from "mobx-react";
+import { hotbarStore } from "../../../common/hotbar-store";
+import { CommandOverlay } from "../command-palette";
+import { Input, InputValidator } from "../input";
+
+const uniqueHotbarName: InputValidator = {
+  condition: ({ required }) => required,
+  message: () => `Hotbar with this name already exists`,
```
Nokel81: nit: doesn't need to be ticks.

src/renderer/components/hotbar/hotbar-menu.tsx
Nokel81: Probably don't need them to be part of this component then.

aleksfront:
Still can't add anything to newly created hotbar. In my case, items can be added to hotbar num 2.
add to hotbar broken

jakolehm:
> Still can't add anything to newly created hotbar. In my case, items can be added to hotbar num 2.

@aleksfront did you remove the hotbar-store json file? It might be corrupted because initially this PR didn't have migration in place...

src/common/hotbar-store.ts
```diff
@@ -86,18 +92,21 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
...
 
-    if (!hotbar.items) {
-      hotbar.items = [];
-    }
+    const hotbar = { id, name, items };
 
     this.hotbars.push(hotbar as Hotbar);
 
     return hotbar as Hotbar;
   }
```
Nokel81: Don't need these casts

aleksfront:
@jakolehm Removed hotbar store file and still can't add anything into any of hotbars. 🐛

Nokel81:
Also I just remembered that in the catalogue, the "Remove item from hotbar" option exists. IMO this is rather confusing (since it would be better to have that on the hotbar item instead). But if we do want it, then it should only be present when the entity actually has a hotbar item.

Further complicating matters it should be possible to have the same entity have multiple hotbar items so would such a button remove all of them?

jakolehm:
> @jakolehm Removed hotbar store file and still can't add anything into any of hotbars. 🐛

@aleksfront you were right, I didn't push all the fixes I had locally 🤦 Should work now.

src/common/hotbar-store.ts
```diff
@@ -34,32 +44,101 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
...
+
+    hotbarStore.activeHotbarId = hotbarStore.hotbars[index].id;
+  }
+
+  switchToNext() {
+    let index = hotbarStore.activeHotbarIndex + 1;
+
+    if (index >= hotbarStore.hotbars.length) {
+      index = 0;
+    }
```
Nokel81, line 135: nit: 
```suggestion
	const total = hotbarStore.hotbars.length;
	const current = hotbarStore.activeHotbarIndex;
    const next = (current + 1) % total;
```

src/common/hotbar-store.ts
```diff
@@ -34,32 +44,101 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
...
+      this.activeHotbarId = this.hotbars[0].id;
+    }
+  }
+
+  switchToPrevious() {
+    let index = hotbarStore.activeHotbarIndex - 1;
+
+    if (index < 0) {
+      index = hotbarStore.hotbars.length - 1;
+    }
```
Nokel81, line 124: nit:
```suggestion
	const total = hotbarStore.hotbars.length;
	const current = hotbarStore.activeHotbarIndex
    const prev = (current + total - 1) % total;
```

src/migrations/hotbar-store/5.0.0-alpha.0.ts
```diff
@@ -2,18 +2,22 @@
...
+import { v4 as uuid } from "uuid";
 
 export default migration({
   version: "5.0.0-alpha.0",
   run(store) {
     const hotbars: Hotbar[] = [];
 
     clusterStore.enabledClustersList.forEach((cluster: any) => {
-      const name = cluster.workspace || "default";
+      const name = cluster.workspace;
```
Nokel81, line 13: We should remove the workspace from the cluster after this migration.

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -57,7 +57,7 @@ export class Catalog extends React.Component {
   }
 
   addToHotbar(item: CatalogEntityItem) {
-    const hotbar = hotbarStore.getByName("default"); // FIXME
+    const hotbar = hotbarStore.getActive();
 
     if (!hotbar) {
       return;
```
Nokel81, line 63: If there is no hotbar we should tell the user to create one. (Or will this function only be displayed to the user if there are hotbars)

src/migrations/hotbar-store/5.0.0-alpha.0.ts
jakolehm: Let's keep it here for a while.... just because someone might still want to go back to 4.x series after testing 5.x.

src/renderer/components/+catalog/catalog.tsx
jakolehm: Story for another PR... currently Lens will create a hotbar if none exists (and does not allow to remove last one).

src/migrations/hotbar-store/5.0.0-alpha.0.ts
Nokel81: okay sounds good. Maybe once we release 5.0?

src/renderer/components/+catalog/catalog.tsx
Nokel81: okay sounds good.

jakolehm:
@aleksfront @nevalla PTAL

nevalla (approved):
LGTM

How user will know what is currently active hotbar, do we show the name anywhere? For example when switching hotbar, the active hotbar could be highlighted or having `(active)`? WDYT?

In future PR we could also add keyboard shortcuts to easily activate next and previous buttons.

jakolehm:
> LGTM
> 
> How user will know what is currently active hotbar, do we show the name anywhere? For example when switching hotbar, the active hotbar could be highlighted or having `(active)`? WDYT?
> 
> In future PR we could also add keyboard shortcuts to easily activate next and previous buttons.

Good points, let's iterate on follow-up PRs 👍

aleksfront:
Do we want to allow adding same cluster to hotbar many times?
same hotbar items

jakolehm:
> Do we want to allow adding same cluster to hotbar many times?
> same hotbar items

I'm not sure, currently we allow. Anyway that decision/change is out-of-scope for this PR 🙂

aleksfront (approved):

PR #2554, opened 2021-04-19
Reserve cluster proxy port just in time

jakolehm:
KubeconfigManager had a bad side-effect whenever `KubeconfigManager.create` was called -> it triggered a port reserve which is is heavy operation. This is bad because it's called for every enabled cluster on boot -> app boot is slow for no reason.

This PR fixes it by calling `KubeconfigManager.init` on-demand -> much faster app boot.


src/main/kubeconfig-manager.ts
```diff
@@ -11,26 +11,13 @@ export class KubeconfigManager {
...
-
-  protected async init() {
-    try {
-      await this.contextHandler.ensurePort();
-      this.tempFile = await this.createProxyKubeconfig();
-    } catch (err) {
-      logger.error(`Failed to created temp config for auth-proxy`, { err });
+  async getPath() {
+    if (!this.tempFile) {
+      await this.init();
```
jakolehm, line 22: This is the actual fix, rest of this PR is mostly refactoring because static `create` is not needed anymore.

nevalla (dismissed):
LGTM

src/main/kubeconfig-manager.ts
```diff
@@ -9,28 +9,15 @@ import logger from "./logger";
...
-  }
-
-  protected async init() {
-    try {
-      await this.contextHandler.ensurePort();
-      this.tempFile = await this.createProxyKubeconfig();
-    } catch (err) {
-      logger.error(`Failed to created temp config for auth-proxy`, { err });
+  async getPath() {
+    if (!this.tempFile && this.tempFile !== undefined) {
```
jim-docker: Doesn't this cover it?
```suggestion
    if (this.tempFile === null) {
```

jim-docker (dismissed):

src/main/kubeconfig-manager.ts
Nokel81: No because we are not being very consistent with the types, on line 40 bellow we assign `this.tempFile = undefined`. Though I don't see why `if (!this.tempFile)` wouldn't be good enough.

src/main/kubeconfig-manager.ts
```diff
@@ -9,28 +9,15 @@ import logger from "./logger";
...
+      await this.init();
     }
-  }
 
-  async getPath() {
     // create proxy kubeconfig if it is removed
     if (this.tempFile !== undefined && !(await fs.pathExists(this.tempFile))) {
```
Nokel81: Isn't this case covered by `init()`?

src/main/kubeconfig-manager.ts
```diff
@@ -41,7 +28,25 @@ export class KubeconfigManager {
...
+  }
+
+  async unlink() {
+    if (!this.tempFile) {
+      return;
+    }
 
+    logger.info(`Deleting temporary kubeconfig: ${this.tempFile}`);
+    await fs.unlink(this.tempFile);
+    this.tempFile = undefined;
```
Nokel81, line 44: Should be `null` to be consistent.

src/main/kubeconfig-manager.ts
```diff
@@ -41,7 +28,25 @@ export class KubeconfigManager {
...
     return this.tempFile;
+  }
+
+  async unlink() {
+    if (!this.tempFile) {
+      return;
+    }
 
+    logger.info(`Deleting temporary kubeconfig: ${this.tempFile}`);
+    await fs.unlink(this.tempFile);
```
Nokel81, line 43: Catch this?

src/main/kubeconfig-manager.ts
jakolehm: I didn't change this code, only moved it's place.

src/main/kubeconfig-manager.ts
jakolehm: Same here, I didn't change this code.

src/main/kubeconfig-manager.ts
jim-docker: I'm confused. What you say is true if that was `||` not `&&`. But as written the condition is "when `this.tempFile` is falsey, except if it is `undefined`", right? What exactly are the conditions when `init()` should be called here?

Nokel81: When @nevalla  was last touching this file he didn't want to remove that `undefined` since I think he wanted the path not to get recreated after `unlink()`. I don't know why that should be case.

jakolehm: @jim-docker you are probably right (with `this.tempFile === null`). I was just following the existing logic here which ended up to be confusing. Basically we want to call init only if tempFile is not set and unlink has not been called (unlink sets this `undefined`).

Nokel81: Shouldn't we throw if `this.tempFile === undefined` then? Since that would mean that we would otherwise be returning `undefined`? Where an error would be easier to track down?

jakolehm: Refactored.

Nokel81 (approved):

jim-docker (approved):
Not sure if this is related but there seems to be an empty (no progress indicator) dashboard shown for a brief time while connecting:

https://user-images.githubusercontent.com/40840436/115598707-b2a07e00-a2a8-11eb-9b18-0678acca8ac8.mov

jim-docker:
> Not sure if this is related but there seems to be an empty (no progress indicator) dashboard shown for a brief time while connecting:
> 

not related to this PR because I see it on master too

jakolehm:
> > Not sure if this is related but there seems to be an empty (no progress indicator) dashboard shown for a brief time while connecting:
> 
> not related to this PR because I see it on master too

I was wondering the same (on master). It seems that `helmRepoManager.init(),` blocks in `bootstrap.tsx` which causes blank screen for a while.

PR #2567, opened 2021-04-19
Add the ability to sync kube config files

### LINKED ISSUE #2507: Syncing kubeconfig files in real time

Discovery mechanism for local kubeconfig files. Discovered clusters should be added to Catalog automatically. 

**Why is this needed**:

Manually adding clusters is annoying.

Nokel81:
- Will update when the files changes

To Dos:

- [x] Rework picking UI
- [x] ~Support proxy information?~
- [x] Add removal UI
- [x] Add label of the source file for filtering purposes
- [x] Support globs or watching folders?
- [x] Default watching of the default kubeconfig file

fixes: #2507 

UI
---
Screen Shot 2021-04-23 at 9 35 37 AM


Nokel81:
@jakolehm Some design questions:

1. Should the removal UI be "part" of the catalogue. What I am thinking is that I introduce a new type of catalogue entry of the form "KubeconfigSync".
2. Should the picking UI be its own page like the cluster add is currently? Or should it be a modal?

jakolehm:
> @jakolehm Some design questions:
> 
> 1. Should the removal UI be "part" of the catalogue. What I am thinking is that I introduce a new type of catalogue entry of the form "KubeconfigSync".

We should probably leave removal of these discovered clusters to a separate issue/pr. Discovered catalog entities should be `KubernetesCluster` with `source=local`.

> 2. Should the picking UI be its own page like the cluster add is currently? Or should it be a modal?

Configuration of discovered folders should be in "Preferences". Manual cluster add should only have "c&p kubeconfig" which actually just writes kubeconfig to preconfigured folder (Lens internal folder for kubeconfigs).

Nokel81:
@jakolehm Sorry for the confusion, let me try and reword my questions:

1. Should this PR have the UI for the removal of the syncs itself (not the individual clusters, those probably shouldn't be able to be removed)
2. I currently have the adding of the syncs gotten to from the "quick dial" menu on the catalogue main page since that is where the button for manually adding clusters is.

src/common/catalog-entity-registry.ts
```diff
@@ -1,10 +1,14 @@
...
 
 export class CatalogEntityRegistry {
-  protected sources = observable.map<string, IObservableArray<CatalogEntity>>([], { deep: true });
+  protected sources = observable.map<string, IComputedValue<CatalogEntity[]>>([], { deep: true });
 
   @action addSource(id: string, source: IObservableArray<CatalogEntity>) {
+    this.sources.set(id, computed(() => source));
+  }
+
+  @action addComputedSource(id: string, source: IComputedValue<CatalogEntity[]>) {
```
jakolehm: I think we should add computed support to `addSource` or rename `addSource` -> `addObservableSource` just to match naming.

src/main/catalog-sources/kubeconfig-sync.source.ts
```diff
@@ -0,0 +1,223 @@
...
+import { Singleton } from "../../common/utils";
+import logger from "../logger";
+import { KubeConfig } from "@kubernetes/client-node";
+import { loadConfigFromString, splitConfig, validateKubeConfig } from "../../common/kube-helpers";
+import { Cluster } from "../cluster";
+import { catalogEntityFromCluster } from "../cluster-manager";
+
+type Disposer = () => void;
+
+export class KubeconfigSyncManager extends Singleton {
```
jakolehm: Maybe add tests for this class?

Nokel81:
@jakolehm Will add tests tomorrow, other than that this should be ready for review.

jakolehm:
@Nokel81 please post screenshots (or video) related to UI changes.

Nokel81:
@jakolehm Added a screenshot to the top post

src/renderer/bootstrap.tsx
```diff
@@ -90,9 +91,41 @@ export async function bootstrap(App: AppComponent) {
   });
   render(<>
     {isMac && <div id="draggable-top" />}
-    <App />
+    {DefaultProps(App)}
   </>, rootElem);
 }
 
+const defaultTheme = createMuiTheme({
```
jakolehm: What is this and why styling is done via "MuiTheme"?

Nokel81: This is how we change the defaults for material ui. This basically adds a new `<style>` to our DOM at the bottom of the `<head>` so it has the highest priority.

jakolehm: I think we have done this differently in the past (through scss). @aleksfront @ixrock  wdyt about this?

src/main/cluster-manager.ts
```diff
@@ -59,17 +59,17 @@ export class ClusterManager extends Singleton {
   }
 
   @action protected updateCatalogSource(clusters: Cluster[]) {
```
jakolehm, line 61: Should we remove whole `updateCatalogSource`  because all local clusters in the catalog should come from "discovery"? We could migrate existing clusters by adding all kubeconfig files to watch list, right?

Nokel81: Yes I think so. Yes we could do that, though since we are already watching `ClusterStore.storedKubeConfigFolder` and `~/.kube/` we probably will get most of them anyway...

Nokel81: Will do is separate PR as discussed.

package.json
```diff
@@ -171,6 +171,7 @@
     "@hapi/call": "^8.0.0",
     "@hapi/subtext": "^7.0.3",
     "@kubernetes/client-node": "^0.12.0",
+    "@material-ui/icons": "^4.11.2",
```
jakolehm: Should be in `devDependencies`.

src/common/catalog/catalog-entity-registry.ts
```diff
@@ -1,10 +1,15 @@
-import { action, computed, observable, IObservableArray } from "mobx";
+import { action, computed, observable, IComputedValue, IObservableArray } from "mobx";
 import { CatalogEntity } from "./catalog-entity";
+import { iter } from "../utils";
 
 export class CatalogEntityRegistry {
-  protected sources = observable.map<string, IObservableArray<CatalogEntity>>([], { deep: true });
+  protected sources = observable.map<string, IComputedValue<CatalogEntity[]>>([], { deep: true });
```
jakolehm, line 6: Does `IComputedValue` cover both computed and observable?

src/common/user-store.ts
```diff
@@ -71,14 +76,32 @@ export class UserStore extends BaseStore<UserStoreModel> {
...
+     * This has to be here before the call to `new Config` in `super.load()`
+     * as we have to make sure that file is in the expected place for that call
+     */
+    await fileNameMigration();
+    await super.load();
 
     // refresh new contexts
-    this.refreshNewContexts();
-    reaction(() => this.kubeConfigPath, this.refreshNewContexts);
+    await this.refreshNewContexts();
```
jakolehm, line 103: Just a note that we probably should remove `refreshNewContexts` because that information is not used anymore.

src/main/catalog-sources/index.ts
```diff
@@ -0,0 +1 @@
+export { KubeconfigSyncManager } from "./kubeconfig-sync.source";
```
jakolehm: I think we don't need `.source` in the filename. Folder name is `catalog-sources`.

src/renderer/bootstrap.tsx
jakolehm: We didn't get any feedback on this... at least move this styling to a separate file and import it here (just to keep things cleaner).

src/renderer/components/app.scss
```diff
@@ -55,7 +55,7 @@
 }
 
 html {
-  font-size: 62.5%; // 1 rem == 10px
+  font-size: var(--font-size);
```
jakolehm: Out of scope for this PR?

src/common/__tests__/kube-helpers.test.ts
```diff
@@ -72,55 +72,63 @@ describe("kube helpers", () => {
     });
     describe("with default validation options", () => {
       describe("with valid kubeconfig", () => {
-        it("does not raise exceptions", () => {
-          expect(() => { validateKubeConfig(kc, "valid");}).not.toThrow();
+        it("does not return an error", () => {
```
Nachasic, line 75: Appreciate the phrasing.

package.json
```diff
@@ -249,7 +250,7 @@
     "@types/hoist-non-react-statics": "^3.3.1",
     "@types/html-webpack-plugin": "^3.2.3",
     "@types/http-proxy": "^1.17.5",
-    "@types/jest": "^25.2.3",
+    "@types/jest": "^26.0.22",
```
Nachasic, line 254: For some reason `yarn` keeps complaining about the integrity of jest types (from `make dev`):
```
yarn check --verify-tree --integrity
yarn check v1.22.10
error "@types/jest#pretty-format#@jest/types#@types/istanbul-reports" is wrong version: expected "^3.0.0", got "1.1.2"
error Found 1 errors.
```
I only was able to build this branch by updating dependency on `ts-jest` from `26.0.1` to `26.3.0`

src/common/kube-helpers.ts
```diff
@@ -28,28 +28,41 @@ function readResolvedPathSync(filePath: string): string {
   return fse.readFileSync(path.resolve(resolveTilde(filePath)), "utf8");
 }
 
+function checkRawCluster(rawCluster: any): boolean {
+  return rawCluster?.name && rawCluster?.cluster?.server;
+}
+
+function checkRawUser(rawUser: any): boolean {
+  return rawUser?.name;
```
Nachasic: Personally, I'm in favor of explicit type conversion here.
```suggestion
  return Boolean(rawUser?.name);
```

src/common/kube-helpers.ts
```diff
@@ -182,7 +207,7 @@ export function getNodeWarningConditions(node: V1Node) {
 /**
  * Checks if `config` has valid `Context`, `User`, `Cluster`, and `exec` fields (if present when required)
```
Nachasic, line 208: Perhaps there should be an annotation about this function returning exceptions instead of throwing them.

src/common/catalog/catalog-entity-registry.ts
Nokel81: Yes because an observable array can be trivially turned into a computed, see below

package.json
Nokel81: Okay

src/common/user-store.ts
Nokel81: Will do in follow up PR

src/main/catalog-sources/index.ts
Nokel81: sounds good

src/renderer/bootstrap.tsx
Nokel81: okay

src/renderer/components/app.scss
Nokel81: I needed this to do the styling correctly for icons. This shouldn't break anything of ours since we don't use `rem`.

src/common/__tests__/kube-helpers.test.ts
```diff
@@ -72,55 +72,63 @@ describe("kube helpers", () => {
...
+          expect(validateKubeConfig(kc, "valid")).toBeUndefined();
         });
       });
       describe("with invalid context object", () => {
-        it("it raises exception", () => {
-          expect(() => { validateKubeConfig(kc, "invalid");}).toThrow("No valid context object provided in kubeconfig for context 'invalid'");
+        it("it returns an error", () => {
```
jim-docker: While you're at it you could remove the extraneous "it" from the following test descriptions
```suggestion
        it("returns an error", () => {
```

src/common/catalog/catalog-entity-registry.ts
```diff
@@ -1,10 +1,15 @@
...
 
-  @action addSource(id: string, source: IObservableArray<CatalogEntity>) {
+  @action addObservableSource(id: string, source: IObservableArray<CatalogEntity>) {
+    this.sources.set(id, computed(() => source));
+  }
```
jim-docker, line 10: Not really relevant to this PR but I bring this up here because you are changing this code. I find the term source too vague. It's an "independent" or sourced list of entities, right? This makes it a little clearer, and we could do the same for the exposed API?
```suggestion
  @action addObservableSource(id: string, entities: IObservableArray<CatalogEntity>) {
    this.sources.set(id, computed(() => entities));
  }
```
or call it `addObservableEntitySource`?

src/common/catalog/catalog-entity-registry.ts
```diff
@@ -1,10 +1,15 @@
...
+
+  @action addComputedSource(id: string, source: IComputedValue<CatalogEntity[]>) {
     this.sources.set(id, source);
   }
```
jim-docker, line 14:
```suggestion
  @action addComputedSource(id: string, entities: IComputedValue<CatalogEntity[]>) {
    this.sources.set(id, entities);
  }
```

src/common/catalog/catalog-entity-registry.ts
```diff
@@ -13,7 +18,7 @@ export class CatalogEntityRegistry {
...
 
   @computed get items(): CatalogEntity[] {
-    return Array.from(this.sources.values()).flat();
+    return Array.from(iter.flatMap(this.sources.values(), source => source.get()));
```
jim-docker, line 21:
```suggestion
    return Array.from(iter.flatMap(this.sources.values(), entity => entity.get()));
```

src/common/kube-helpers.ts
```diff
@@ -181,8 +206,10 @@ export function getNodeWarningConditions(node: V1Node) {
 
 /**
  * Checks if `config` has valid `Context`, `User`, `Cluster`, and `exec` fields (if present when required)
+ *
+ * Note: This function returns an error instead of throwing it, returning `undefined` if the validation passes
```
jim-docker, line 210: Should you `try`/`catch` then in here, in case any of the called functions throw?

src/common/user-store.ts
```diff
@@ -230,32 +243,44 @@ export class UserStore extends BaseStore<UserStoreModel> {
     for (const [tableId, columnIds] of preferences.hiddenTableColumns ?? []) {
       this.hiddenTableColumns.set(tableId, new ObservableToggleSet(columnIds));
     }
+
+    if (preferences.syncKubeconfigEntries) {
+      this.syncKubeconfigEntries.replace(preferences.syncKubeconfigEntries.map(({ filePath }) => filePath));
+    } else {
+      this.syncKubeconfigEntries.clear();
```
jim-docker: Don't we want it to always have ".kube" and `ClusterStore.storedKubeConfigFolder`?

jim-docker (commented):
about halfway through...

src/common/user-store.ts
Nokel81: Well if a user removes them we don't.

src/common/catalog/catalog-entity-registry.ts
Nokel81: So these are used `catalogEntityRegistry.addObservableSource(...)`. Do we need to have the word "Entity" twice, namely: `catalogEntityRegistry.addObservableEntitySource(...)` ?

src/common/catalog/catalog-entity-registry.ts
Nokel81: I don't think this specific name change is very good, because we are flat mapping over `sources.values()`. Also `sources` shouldn't really be called `entities` because it is a map between IDs and `IComputedValue<CatalogEntity[]>`.

src/common/kube-helpers.ts
Nokel81: Probably

src/common/catalog/catalog-entity-registry.ts
jim-docker: I guess conceptually it is a source but it is confusing for me because there are two different data types you are calling "source". You're not iterating over the `this.sources` map, but rather `this.sources.values()`, which gives you an iterable over just the entities, free of the IDs, right? I still like `entities` (not `entity`, my mistake) better here than `source`. Anyway, it's fine

Nokel81: Yes I am iterating over `this.sources.values()` but that is of the type `Iterable<CatalogEntities[]>` which is why I do a `iter.flatMap`

jim-docker:
If I sync a folder that has multiple configs, some with duplicate contexts, then when I remove the sync not all of the clusters are removed from the catalog.


https://user-images.githubusercontent.com/40840436/116318012-226ba880-a782-11eb-983b-247be2b3d373.mov

src/common/user-store.ts
```diff
@@ -71,14 +76,32 @@ export class UserStore extends BaseStore<UserStoreModel> {
...
 
-  protected async handleOnLoad() {
-    await this.whenLoaded;
+  /**
+   * The set of file/folder paths to be synced
+   */
+  syncKubeconfigEntries = observable.set([
+    path.join(os.homedir(), ".kube"),
+    ClusterStore.storedKubeConfigFolder,
+  ]);
```
jim-docker: Is this useful, since this set is cleared the first time `fromStore()` is called (assuming the store does not initially have `preferences.syncKubeconfigEntries`)?

src/migrations/user-store/5.0.0-alpha.2.ts
```diff
@@ -0,0 +1,18 @@
...
+    const preferences = store.get("preferences");
+    const oldHiddenTableColumns: Record<string, string[]> = preferences?.hiddenTableColumns;
+
+    if (!oldHiddenTableColumns) {
+      return;
+    }
+
+    preferences.hiddenTableColumns = Object.entries(oldHiddenTableColumns);
+
+    store.get("preferences", preferences);
```
jim-docker: `store.set`?  (deja vu)

src/common/catalog/catalog-entity-registry.ts
jakolehm: > Do we need to have the word "Entity" twice

No 🙂

src/common/user-store.ts
jakolehm: I think we want to keep at least `ClusterStore.storedKubeConfigFolder` always. Otherwise "add cluster from c&p" will be seriously broken.

src/migrations/user-store/5.0.0-alpha.2.ts
jakolehm: This needs to be fixed.

src/migrations/user-store/5.0.0-alpha.2.ts
```diff
@@ -0,0 +1,18 @@
+// Switch representation of hiddenTableColumns in store
+import { migration } from "../migration-wrapper";
+
+export default migration({
+  version: "5.0.0-alpha.2",
```
jakolehm: We should bump this version, right?

src/renderer/components/app.scss
jakolehm: Does not sound right, could you revert this and add more specific styling for icons?

src/migrations/user-store/5.0.0-alpha.2.ts
Nokel81: Agreed, I am sorry that this slipped in. It almost certainly was the result of rebasing against the wrong branch target.

src/common/user-store.ts
Nokel81: Okay, understood

src/common/user-store.ts
jim-docker: I stand corrected, at least when you start with an empty app data folder then these syncs are added. Another question though, should the `ClusterStore.storedKubeConfigFolder` even be visible to the user on the Preferences page? Shouldn't we always, and invisibly, sync this folder?

src/renderer/components/app.scss
Nokel81: Sure, do you know why we want the rem to be smaller?

src/migrations/user-store/5.0.0-alpha.2.ts
Nokel81: This really shouldn't be here at all, will delete these two files.

src/common/user-store.ts
```diff
@@ -23,10 +22,12 @@ export interface UserStoreModel {
   preferences: UserPreferencesModel;
 }
 
-export interface KubeconfigSyncEntry {
+export interface KubeconfigSyncEntry extends KubeconfigSyncValue {
   filePath: string;
 }
 
+export interface KubeconfigSyncValue {}
```
jim-docker, line 29: for future use?

Nokel81: Yes, namely `httpProxy` settings

src/renderer/components/+preferences/kubeconfig-syncs.tsx
```diff
@@ -0,0 +1,178 @@
...
+        <ListItem>
+          <ListItemAvatar>
+            <Avatar>
+              {this.renderEntryIcon(entry)}
+            </Avatar>
+          </ListItemAvatar>
+          <ListItemText
+            primary={entry.filePath}
+            className="description"
+          />
```
jim-docker, line 125: I think the filePath text is too small
Screen Shot 2021-04-28 at 7 25 15 PM

jim-docker (commented):
- if I sync both a folder and a file inside the folder the clusters get added twice. Though this is  exactly what was asked for, it's not a great UX. Should we remove file syncs already covered by a folder sync?
- got this when adding a sync in Lens's app data:
Screen Shot 2021-04-28 at 7 24 45 PM
Of course I shouldn't be doing that...

src/renderer/components/+preferences/kubeconfig-syncs.tsx
Nokel81: I agree, will fix this.

jim-docker: Could the folder icon be larger too, to fill more of the circular background?

Nokel81:
New sizes: 
Screen Shot 2021-04-29 at 10 12 11 AM

jim-docker:
Looking good. 

If I set a sync on a file (or folder), then remove the file (or a file from the folder), all the clusters from that file remain in the catalog.

src/common/utils/iter.ts
```diff
@@ -23,3 +23,67 @@ export function* take<T>(src: Iterable<T>, n: number): Iterable<T> {
     break outer;
   }
 }
+
```
jakolehm, line 26: Just a note: would make sense to add tests for these utils if they are used in many places.

src/renderer/api/__tests__/catalog-entity-registry.test.ts
```diff
@@ -43,27 +43,6 @@ describe("CatalogEntityRegistry", () => {
       expect(catalog.items.length).toEqual(2);
     });
 
-    it("ignores unknown items", () => {
```
jakolehm, line 46: Why remove? Makes no sense anymore?

Nokel81: Because I removed the merge logic, so the `renderer` just takes the whole blast from `main` and uses that.

src/common/utils/iter.ts
Nokel81: I can do that in a separate PR, but good idea.

src/main/catalog-sources/kubeconfig-sync.ts
```diff
@@ -0,0 +1,249 @@
...
+
+          if (!cluster.apiUrl) {
+            throw new Error("Cluster constructor failed, see above error");
+          }
+
+          cluster.init(port);
+
+          const entity = catalogEntityFromCluster(cluster);
+
+          entity.metadata.labels.KUBECONFIG_SYNC = filePath;
```
jakolehm: I think we should not add this label (unless we really have to). It adds too much noise to catalog.

jakolehm (commented):
For unknown reason I did not get the default `~/.kube` folder synced when I started the app.

jakolehm (approved):
LGTM

jim-docker (approved):

PR #2569, opened 2021-04-19
Fix how integration tests activate adding a cluster via SpeedDial

### LINKED ISSUE #2564: Integration tests are flaky under github actions
started happening at the same time as the switch to github actions, though could be a coincidence

jim-docker:
In adding minikube cluster, simulate hover on `SpeedDial` to bring up the "Add from kubeconfig" `SpeedDialAction`

Before this fix, the integration tests run on Ubuntu by CI and locally fail most of the time to add a cluster via the `SpeedDial` element.

This fix involves using spectron client's `moveToObject()` method (which appears to rely on the `moveTo()` method, which is deprecated). 

tries to fix #2564

### DIFF
--- a/integration/helpers/minikube.ts
+++ b/integration/helpers/minikube.ts
@@ -40,7 +40,7 @@ export function minikubeReady(testNamespace: string): boolean {
 
 export async function addMinikubeCluster(app: Application) {
   await app.client.waitForVisible("button.MuiSpeedDial-fab");
-  await app.client.click("button.MuiSpeedDial-fab");
+  await app.client.moveToObject("button.MuiSpeedDial-fab");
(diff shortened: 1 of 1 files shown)


Nokel81:
@jim-docker Is it deprecated in favour of something else?

Nokel81 (approved):

jim-docker:
> @jim-docker Is it deprecated in favour of something else?

Yes, the actions command, which I've tried to use before to fix other issues. IIRC they aren't supported in the version of Spectron we're stuck on. (I could not get them to work in any case)

Nokel81:
Ah okay thanks for the info. So this will be something that we will need to do when we upgrade to the latest version of electron (which is needed for spectron).

jakolehm (approved):
🏆

PR #2571, opened 2021-04-19
Add ability to add accessible namespaces when adding clusters 

### LINKED ISSUE #2528: Enter accessible namespaces when adding a cluster
**What would you like to be added**: When adding a cluster, users should be able to specify the accessible namespaces they have access to and be able to select the namespace they want as the default.

**Why is this needed**: Users can have limited permissions via RBAC, ie they are not granted list all namespaces and can only access the resources in a specific namespace. It can be very confusing for a user to add a cluster and not see anything under workloads because it defaults to the `default` namespace where they might not have any permissions to it and then they have to add their namespaces after adding the cluster.

- Kubernetes distribution: On Prem
- Desktop OS: MacOS

Nokel81:
- Switched to using Material-UI tabs as our tabs don't work with scrolling for some reason
- Built upon the branch that this is targetting.

Fixes: #2528 

New
---
Screen Shot 2021-04-19 at 5 19 29 PM

### DIFF
--- a/src/renderer/components/+add-cluster/add-cluster.scss
+++ b/src/renderer/components/+add-cluster/add-cluster.scss
@@ -17,6 +17,14 @@
     }
   }
 
+  .content {
+    overflow-y: scroll;
+  }
+
+  .MuiTabs-indicator {
+    background-color: var(--primary);
+  }
+
   code {
     color: $pink-400;
   }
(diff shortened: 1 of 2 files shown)


nevalla:
How is this working when user is adding multiple contexts?  Are we adding accessible namespaces to every cluster then?

Nokel81:
@nevalla yes just like `HTTP Proxy` settings when adding multiple clusters. There is a hint on the screen about that.

Nokel81:
Closing as scope has now changed. Will open a new PR with new scope.

PR #2575, opened 2021-04-20
Display a pod as terminating even if it isn't running

### LINKED ISSUE #1893: Pod status is wrong when pod is being terminated
**Describe the bug**

Pod shows "Container creating" status when pod is being terminated.

**To Reproduce**

Scale down deployment via terminal and watch pod list view/status column.

Nokel81:

fixes #1893

### DIFF
--- a/src/renderer/api/endpoints/pods.api.ts
+++ b/src/renderer/api/endpoints/pods.api.ts
@@ -342,7 +342,7 @@ export class Pod extends WorkloadKubeObject {
   // Returns pod phase or container error if occurred
   getStatusMessage() {
     if (this.getReason() === PodStatus.EVICTED) return "Evicted";
-    if (this.getStatus() === PodStatus.RUNNING && this.metadata.deletionTimestamp) return "Terminating";
+    if (this.metadata.deletionTimestamp) return "Terminating";
 
     const statuses = this.getContainerStatuses(false); // not including initContainers
 


src/renderer/api/endpoints/pods.api.ts
```diff
@@ -342,7 +342,7 @@ export class Pod extends WorkloadKubeObject {
   // Returns pod phase or container error if occurred
   getStatusMessage() {
     if (this.getReason() === PodStatus.EVICTED) return "Evicted";
-    if (this.getStatus() === PodStatus.RUNNING && this.metadata.deletionTimestamp) return "Terminating";
+    if (this.metadata.deletionTimestamp) return "Terminating";
```
jim-docker, line 345: with this it will probably never show "Terminated", which is checked for below.

Nokel81: Even though we use the same root word ("terminate"), these two are actually different. This means that the pod has been chosen to be stopped by its parent deployment whereas the `state.terminated` means that the pod has run and then stopped. Which is why there is an `exitCode` field under `IPodContainerStatus.state.terminated`

jim-docker: Ah, thanks. So when stopped by the parent deployment we just show "Terminating" until the pod disappears from the list?

jim-docker (approved):

src/renderer/api/endpoints/pods.api.ts
Nokel81: Yes, because that is what `kubectl` does.

jakolehm (approved):

PR #2577, opened 2021-04-20
Update in-tree extensions package files

Nokel81:

Caused by the increase of `npm` version.


jakolehm (approved):

src/extensions/npm/extensions/package.json
```diff
@@ -1,25 +0,0 @@
-{
```
jakolehm: We should not remove these.

jakolehm (changes requested):

src/extensions/npm/extensions/package.json
Nokel81: Yes that is true, I am finding that `make dev -B` delete these once webpack starts to be run. Are you seeing this?

jakolehm (approved):

jakolehm:
Not needed because npm upgrade was reverted?

Nokel81:
Fair, we will probably need this once we do that again

PR #2580, opened 2021-04-21
Use productName from package.json

jakolehm:
- uses `productName` where possible
- shows auto-update in tray only if package.json has publish configured

### DIFF
--- a/src/renderer/template.html
+++ b/src/renderer/template.html
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
-  <title>Lens - The Kubernetes IDE</title>
+  <title>OpenLens - Open Source Kubernetes IDE</title>
 </head>
 <body>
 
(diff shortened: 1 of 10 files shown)


src/renderer/template.html
```diff
@@ -2,11 +2,11 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
-  <title>Lens - The Kubernetes IDE</title>
+  <title>OpenLens - Open Source Kubernetes IDE</title>
```
Nokel81, line 5: Is it possible to have this templated?

Nokel81 (approved):

jim-docker (approved):
There are a few more occurrences of "Lens" hardcoded in `src/main/index.ts`, and `catalog.tsx` (the Welcome! notification), `extensions.tsx`, and `preferences.tsx`, which show up in the UI, that could be templated?

PR #2466, opened 2021-04-07
Increase npm version to 7.9

Nokel81:
Signed-off-by: Sebastian Malton <sebastian@malton.name>

fixes #2465

PR #2583, opened 2021-04-21
Revert "Increase npm version to 7.9"

jakolehm:
Reverts lensapp/lens#2466 because of https://github.com/npm/cli/issues/2380


Nokel81 (approved):
We should add our case to that issue.

jim-docker (approved):

Nokel81:
@jakolehm Do you remember where we use `npm link` at all?

PR #2587, opened 2021-04-21
Flatten user preferences in user store

Nokel81:

This is most done as a precursor to #2567 as I would like to rely on the observable nature of these fields (and new ones in the future) in that PR.

This also helps break the bond between the serialization format and the in-memory format so that we can focus on changing one without affecting the other.

### DIFF
--- a/src/common/user-store.ts
+++ b/src/common/user-store.ts
@@ -12,15 +12,16 @@ import { appEventBus } from "./event-bus";
 import logger from "../main/logger";
 import path from "path";
 import { fileNameMigration } from "../migrations/user-store";
+import { ObservableToggleSet } from "../renderer/utils";
 
 export interface UserStoreModel {
   kubeConfigPath: string;
   lastSeenAppVersion: string;
   seenContexts: string[];
-  preferences: UserPreferences;
+  preferences: UserPreferencesModel;
 }
 
-export interface UserPreferences {
+export interface UserPreferencesModel {
   httpsProxy?: string;
   shell?: string;
   colorTheme?: string;
@@ -32,7 +33,7 @@ export interface UserPreferences {
   downloadBinariesPath?: string;
   kubectlBinariesPath?: string;
   openAtLogin?: boolean;
-  hiddenTableColumns?: Record<string, string[]>;
+  hiddenTableColumns?: [string, string[]][];
 }
 
 export class UserStore extends BaseStore<UserStoreModel> {
@@ -48,20 +49,29 @@ export class UserStore extends BaseStore<UserStoreModel> {
   }
 
   @observable lastSeenAppVersion = "0.0.0";
-  @observable kubeConfigPath = kubeConfigDefaultPath; // used in add-cluster page for providing context
+
+  /**
+   * used in add-cluster page for providing context
+   */
+  @observable kubeConfigPath = kubeConfigDefaultPath;
   @observable seenContexts = observable.set<string>();
   @observable newContexts = observable.set<string>();
+  @observable allowTelemetry = true;
+  @observable allowUntrustedCAs = false;
+  @observable colorTheme = UserStore.defaultTheme;
+  @observable localeTimezone = moment.tz.guess(true) || "UTC";
+  @observable downloadMirror = "default";
+  @observable httpsProxy?: string;
+  @observable shell?: string;
+  @observable downloadBinariesPath?: string;
+  @observable kubectlBinariesPath?: string;
 
-  @observable preferences: UserPreferences = {
-    allowTelemetry: true,
-    allowUntrustedCAs: false,
-    colorTheme: UserStore.defaultTheme,
-    localeTimezone: moment.tz.guess(true) || "UTC",
-    downloadMirror: "default",
-    downloadKubectlBinaries: true,  // Download kubectl binaries matching cluster version
-    openAtLogin: false,
-    hiddenTableColumns: {},
-  };
+  /**
+   * Download kubectl binaries matching cluster version
+   */
+  @observable downloadKubectlBinaries = true;
+  @observable openAtLogin = false;
+  hiddenTableColumns = observable.map<string, ObservableToggleSet<string>>();
 
   protected async handleOnLoad() {
     await this.whenLoaded;
@@ -72,12 +82,12 @@ export class UserStore extends BaseStore<UserStoreModel> {
 
     if (app) {
       // track telemetry availability
-      reaction(() => this.preferences.allowTelemetry, allowed => {
+      reaction(() => this.allowTelemetry, allowed => {
         appEventBus.emit({ name: "telemetry", action: allowed ? "enabled" : "disabled" });
       });
 
       // open at system start-up
-      reaction(() => this.preferences.openAtLogin, openAtLogin => {
+      reaction(() => this.openAtLogin, openAtLogin => {
         app.setLoginItemSettings({
           openAtLogin,
           openAsHidden: true,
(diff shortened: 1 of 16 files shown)


src/common/utils/toggle-set.ts
```diff
@@ -0,0 +1,24 @@
...
+export class ObservableToggleSet<T> extends ObservableSet<T> {
+  @action
+  public toggle(value: T): void {
+    if (!this.delete(value)) {
+      // Set.prototype.delete returns false if `value` was not in the set
+      this.add(value);
+    }
+  }
+
+  toJS(): ToggleSet<T> {
```
ixrock: `mobx6` returns an array of values instead of `Set`. Maybe better to return `mobx.toJS(this)`?

src/renderer/components/+preferences/kubectl-binaries.tsx
```diff
@@ -29,8 +29,8 @@ export const KubectlBinaries = observer(({ preferences }: { preferences: UserPre
         <FormSwitch
           control={
             <Switcher
-              checked={preferences.downloadKubectlBinaries}
-              onChange={v => preferences.downloadKubectlBinaries = v.target.checked}
+              checked={UserStore.getInstance().downloadKubectlBinaries}
```
ixrock: Maybe better to extract to `const` within current  function scope and re-use below?

src/renderer/utils/createStorage.ts
```diff
@@ -33,7 +33,7 @@ export function createStorage<T>(key: string, defaultValue: T, observableOptions
 
     // remove json-file when cluster deleted
     if (clusterId !== undefined) {
-      when(() => ClusterStore.getInstance().removedClusters.has(clusterId)).then(removeFile);
+      when(() => ClusterStore.getInstance(false)?.removedClusters.has(clusterId)).then(removeFile);
```
ixrock: Why `Singleton.getInstance(strict?: boolean)` has such definition?
Isn't it simpler to always return instance or `undefined` and use this info at place of usage (throw or whatever)?

src/common/user-store.ts
```diff
@@ -175,15 +208,59 @@ export class UserStore extends BaseStore<UserStoreModel> {
...
+        shell: toJS(this.shell),
+        colorTheme: toJS(this.colorTheme),
+        localeTimezone: toJS(this.localeTimezone),
+        allowUntrustedCAs: toJS(this.allowUntrustedCAs),
+        allowTelemetry: toJS(this.allowTelemetry),
+        downloadMirror: toJS(this.downloadMirror),
+        downloadKubectlBinaries: toJS(this.downloadKubectlBinaries),
+        downloadBinariesPath: toJS(this.downloadBinariesPath),
+        kubectlBinariesPath: toJS(this.kubectlBinariesPath),
+        openAtLogin: toJS(this.openAtLogin),
```
ixrock: Why there  are `toJS` calls? Below `toJS(model)` should do the job.

src/common/utils/toggle-set.ts
Nokel81: Yes good point. Thanks

src/renderer/components/+preferences/kubectl-binaries.tsx
Nokel81: Sounds good

src/renderer/utils/createStorage.ts
Nokel81: Basically for better error messages. Basically all users of `.getInstance()` won't work without an instance. So this doubly helps to call out where they can handle it.

src/common/user-store.ts
Nokel81: Oh I missed it bellow, will remove

ixrock (dismissed):

src/migrations/user-store/5.0.0-alpha.2.ts
```diff
@@ -0,0 +1,18 @@
+// Switch representation of hiddenTableColumns in store
+import { migration } from "../migration-wrapper";
+
+export default migration({
+  version: "5.0.0-alpha.2",
```
jakolehm:
```suggestion
  version: "5.0.0-alpha.3",
```

src/migrations/user-store/index.ts
```diff
@@ -1,6 +1,7 @@
 // User store migrations
 
 import version210Beta4 from "./2.1.0-beta.4";
+import version500Alpha2 from "./5.0.0-alpha.2";
```
jakolehm:
```suggestion
import version500Alpha3 from "./5.0.0-alpha.3";
```

jakolehm (dismissed):

src/common/user-store.ts
```diff
@@ -99,17 +109,40 @@ export class UserStore extends BaseStore<UserStoreModel> {
...
+    return columnIds.some(columnId => config.has(columnId));
   }
 
-  getHiddenTableColumns(tableId: string): Set<string> {
-    return new Set(this.preferences.hiddenTableColumns[tableId]);
+  @action
+  /**
+   * Toggles the hidden configuration of a table's column
+   */
+  toggleTableColumnVisility(tableId: string, columnId: string) {
```
jim-docker:
```suggestion
  toggleTableColumnVisibility(tableId: string, columnId: string) {
```

src/common/user-store.ts
```diff
@@ -175,15 +208,59 @@ export class UserStore extends BaseStore<UserStoreModel> {
...
+    this.kubectlBinariesPath = preferences.kubectlBinariesPath;
+    this.openAtLogin = preferences.openAtLogin;
+
+    if (preferences.hiddenTableColumns) {
+      for (const [tableId, columnIds] of preferences.hiddenTableColumns) {
+        if (this.hiddenTableColumns.has(tableId)) {
+          this.hiddenTableColumns.get(tableId).replace(columnIds);
+        } else {
+          this.hiddenTableColumns.set(tableId, new ObservableToggleSet(columnIds));
+        }
```
jim-docker: Should a `tableId` in `this.hiddenTableColumns` that is not in `preferences.hiddenTableColumns` be removed from `this.hiddenTableColumns`?

src/main/shell-session/shell-session.ts
```diff
@@ -119,13 +119,12 @@ export abstract class ShellSession {
   protected async getShellEnv() {
     const env = clearKubeconfigEnvVars(JSON.parse(JSON.stringify(await shellEnv())));
     const pathStr = [...this.getPathEntries(), await this.kubectlBinDirP, process.env.PATH].join(path.delimiter);
-    const shell = UserStore.getInstance().preferences.shell || process.env.SHELL || process.env.PTYSHELL;
```
jim-docker, line 122: Or 
```
const shell = UserStore.getInstance().resolvedShell;
```

src/migrations/user-store/5.0.0-alpha.3.ts
```diff
@@ -0,0 +1,18 @@
...
+    const preferences = store.get("preferences");
+    const oldHiddenTableColumns: Record<string, string[]> = preferences?.hiddenTableColumns;
+
+    if (!oldHiddenTableColumns) {
+      return;
+    }
+
+    preferences.hiddenTableColumns = Object.entries(oldHiddenTableColumns);
+
+    store.get("preferences", preferences);
```
jim-docker: `store.set`?

Nokel81: yes thanks

src/main/shell-session/shell-session.ts
Nokel81: fair enough

src/common/user-store.ts
Nokel81: Yes

jim-docker (dismissed):

jakolehm (dismissed):

nevalla (dismissed):

jim-docker (approved):

jakolehm (approved):
LGTM

PR #2591, opened 2021-04-22
Initialize HelmRepoManager on demand

### LINKED ISSUE #2594: Loading of screens seems to take more than 5s
**Describe the bug**
When clicking on a cluster from the catalog screen the green notification box says the cluster is connected but the screen is empty for approx 5s before the cluster view is displayed.

**To Reproduce**

**Expected behavior**

**Screenshots**

- Lens Version: 5.0.0-alpha.2
- OS: [e.g. OSX] OSX

**Logs:**
```
Your logs go here...
```

**Kubeconfig:**
```
your kubeconfig here
```

**Additional context**

Nokel81:
- Make most functions static as they don't actually reference the parsed
  helm ENV

- Fix HelmRepoManager not being made on main

Fixes #2594

### DIFF
--- a/src/main/helm/helm-repo-manager.ts
+++ b/src/main/helm/helm-repo-manager.ts
@@ -35,7 +35,7 @@ export class HelmRepoManager extends Singleton {
   protected helmEnv: HelmEnv;
   protected initialized: boolean;
 
-  async loadAvailableRepos(): Promise<HelmRepo[]> {
+  public static async loadAvailableRepos(): Promise<HelmRepo[]> {
     const res = await customRequestPromise({
       uri: "https://github.com/lensapp/artifact-hub-repositories/releases/download/latest/repositories.json",
       json: true,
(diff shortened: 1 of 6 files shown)


jakolehm (approved):

src/main/helm/helm-repo-manager.ts
```diff
@@ -102,13 +106,7 @@ export class HelmRepoManager extends Singleton {
     }
   }
 
-  public async repository(name: string) {
-    const repositories = await this.repositories();
-
-    return repositories.find(repo => repo.name == name);
-  }
```
jim-docker, line 109: Why eliminate this method in favour of duplicating code (calling `repositories()` and the `find()` call everywhere we used to just call `repository()`)?

jim-docker (approved):

src/main/helm/helm-repo-manager.ts
Nokel81: Mostly to call out that have to get all `repositories` and that there isn't a way to get a single repo.

PR #2596, opened 2021-04-22
Allow to define entity setting priority

jakolehm:
This is needed so that extension developer can control order of tabs/menus.

### DIFF
--- a/src/extensions/registries/entity-setting-registry.ts
+++ b/src/extensions/registries/entity-setting-registry.ts
@@ -15,8 +15,9 @@ export interface EntitySettingRegistration {
   kind: string;
   apiVersions: string[];
   source?: string;
-  id?: string;
   components: EntitySettingComponents;
+  id?: string;
+  priority?: number;
 }
 
 export interface RegisteredEntitySetting extends EntitySettingRegistration {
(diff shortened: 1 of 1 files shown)


Nokel81 (commented):
Wouldn't it be better to do what `Menu` does (from electron) and define both a "beforeEntry" and "afterEntry"?

jim-docker:
> Wouldn't it be better to do what `Menu` does (from electron) and define both a "beforeEntry" and "afterEntry"?

But we're already using a similar `priority` field for `KubeObjectDetailRegistration`...

Nokel81:
Fair, I guess. It was merely a suggestion. I would change the name though, `priority` is a strange name for this sort of field, maybe `position`.

I would say that before/after shows more intent but I am not set on it.

jim-docker:
I think `priority` is ok. To change it we'd want to be consistent and change `KubeObjectDetailRegistration` too, which means a breaking change, not worth it, I think.

jakolehm:
Yes, `priority` follows the existing logic.

nevalla (approved):

Nokel81 (approved):

PR #2597, opened 2021-04-22
Wait for LensApp to report that the view has loaded before showing main window

Nokel81:
This prevents showing an empty main window for a few seconds. From a UX perspective it feels more responsive because the user is not greeted with a blank page after the loading window goes away.

### DIFF
--- a/src/main/window-manager.ts
+++ b/src/main/window-manager.ts
@@ -21,18 +21,22 @@
 
 import type { ClusterId } from "../common/cluster-store";
 import { makeObservable, observable } from "mobx";
-import { app, BrowserWindow, dialog, shell, webContents } from "electron";
+import { app, BrowserWindow, dialog, ipcMain, shell, webContents } from "electron";
 import windowStateKeeper from "electron-window-state";
 import { appEventBus } from "../common/event-bus";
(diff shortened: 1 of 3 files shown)


src/main/window-manager.ts
```diff
@@ -181,8 +199,8 @@ export class WindowManager extends Singleton {
   }
 
   hide() {
-    if (this.mainWindow && !this.mainWindow.isDestroyed()) this.mainWindow.hide();
-    if (this.splashWindow && !this.splashWindow.isDestroyed()) this.splashWindow.hide();
+    if (!this.mainWindow?.isDestroyed()) this.mainWindow.hide();
+    if (!this.splashWindow?.isDestroyed()) this.splashWindow.hide();
```
jim-docker: I think this should be left as it was or:
```suggestion
    if (!this.mainWindow?.isDestroyed()) this.mainWindow?.hide();
    if (!this.splashWindow?.isDestroyed()) this.splashWindow?.hide();
```
Screen Shot 2021-04-26 at 12 31 43 PM

Nokel81: Ah yes, you are right

nevalla (approved):

PR #2604, opened 2021-04-23
Add tooltip to hotbar index

jakolehm:
Tooltip helps to identify which hotbar is currently active.

### DIFF
--- a/src/renderer/components/hotbar/hotbar-menu.tsx
+++ b/src/renderer/components/hotbar/hotbar-menu.tsx
@@ -12,6 +12,7 @@ import { Icon } from "../icon";
 import { Badge } from "../badge";
 import { CommandOverlay } from "../command-palette";
 import { HotbarSwitchCommand } from "./hotbar-switch-command";
+import { Tooltip, TooltipPosition } from "../tooltip";
 
 interface Props {
   className?: IClassName;
@@ -43,6 +44,7 @@ export class HotbarMenu extends React.Component<Props> {
 
   render() {
     const { className } = this.props;
+    const hotbar = HotbarStore.getInstance();
     const hotbarIndex = HotbarStore.getInstance().activeHotbarIndex + 1;
 
     return (
@@ -63,7 +65,13 @@ export class HotbarMenu extends React.Component<Props> {
         <div className="HotbarSelector flex gaps auto">
           <Icon material="chevron_left" className="previous box" onClick={() => this.previous()} />
           <div className="box">
-            <Badge small label={hotbarIndex} onClick={() => this.openSelector()} />
+            <Badge id="hotbarIndex" small label={hotbarIndex} onClick={() => this.openSelector()} />
+            <Tooltip
+              targetId="hotbarIndex"
+              preferredPositions={TooltipPosition.TOP}
+            >
+              {hotbar.name}
+            </Tooltip>
           </div>
           <Icon material="chevron_right" className="next box" onClick={() => this.next()} />
         </div>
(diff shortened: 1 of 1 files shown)


nevalla (dismissed):

jim-docker (commented):
huh?
Screen Shot 2021-04-23 at 8 48 21 AM

src/renderer/components/hotbar/hotbar-menu.tsx
```diff
@@ -63,7 +65,13 @@ export class HotbarMenu extends React.Component<Props> {
         <div className="HotbarSelector flex gaps auto">
           <Icon material="chevron_left" className="previous box" onClick={() => this.previous()} />
           <div className="box">
-            <Badge small label={hotbarIndex} onClick={() => this.openSelector()} />
+            <Badge id="hotbarIndex" small label={hotbarIndex} onClick={() => this.openSelector()} />
+            <Tooltip
+              targetId="hotbarIndex"
+              preferredPositions={TooltipPosition.TOP}
+            >
+              {hotbar.name}
```
jim-docker, line 74: This is the name of the hotbar store?

src/renderer/components/hotbar/hotbar-menu.tsx
```diff
@@ -43,6 +44,7 @@ export class HotbarMenu extends React.Component<Props> {
 
   render() {
     const { className } = this.props;
+    const hotbar = HotbarStore.getInstance();
     const hotbarIndex = HotbarStore.getInstance().activeHotbarIndex + 1;
```
jim-docker: you don't need to get the instance again

src/renderer/components/hotbar/hotbar-menu.tsx
jakolehm: Sounds like migration gone south?

src/renderer/components/hotbar/hotbar-menu.tsx
jakolehm: Oh, that's typo (from auto-complete). Thanks for catching 👍

src/renderer/components/hotbar/hotbar-menu.tsx
jakolehm: It was the typo...

src/renderer/components/hotbar/hotbar-menu.tsx
```diff
@@ -43,7 +44,9 @@ export class HotbarMenu extends React.Component<Props> {
 
   render() {
     const { className } = this.props;
-    const hotbarIndex = HotbarStore.getInstance().activeHotbarIndex + 1;
+    const hotbarStore = HotbarStore.getInstance();
+    const hotbar = hotbarStore.getActive();
+    const hotbarIndex = hotbarStore.activeHotbarIndex + 1;
```
Nokel81: This line is confusing, maybe rename to `activeIndexDisplay`?

jakolehm: Done.

Nokel81 (approved):

jim-docker (approved):
Out of scope for this PR but it still feels like this isn't enough. I have to hover over the index to see what hotbar I'm using. Can the hotbar name be a title over the catalog list? More like this is a "View", and the hotbar is a UI element belonging to the View, and the View title and hotbar name are the same. Or instead of View we can call it Workspace? :)

Nokel81:
Maybe the name could be displayed just above the `< Index >` line.

PR #2606, opened 2021-04-23
Add icon to catalog category

### LINKED ISSUE #2605: CatalogCategory should have icon
Icon should support both material and custom svg icons.

UI mockup:
image

jakolehm:
Implements initial support for `CatalogCategory` icons. Currently icon is only shown in hotbar but we should show this also in Catalog (out-of-scope for this PR).

Also improves `Icon` component to support custom svgs.

image

Fixes #2605


Nachasic (approved):
Looks good

src/renderer/components/icon/icon.tsx
```diff
@@ -94,7 +94,7 @@ export class Icon extends React.PureComponent<IconProps> {
 
     // render as inline svg-icon
     if (svg) {
-      const svgIconText = require(`!!raw-loader!./${svg}.svg`).default;
+      const svgIconText = svg.includes("<svg") ? svg : require(`!!raw-loader!./${svg}.svg`).default;
```
ixrock, line 97: How does it work / what's fixing? 
There are some working usages of `svg` prop in codebase:

image

src/renderer/components/hotbar/hotbar-icon.tsx
```diff
@@ -55,6 +56,21 @@ export class HotbarIcon extends React.Component<Props> {
...
+    if (!category) {
+      return <Icon material="bug_report" className={className} />;
+    }
+
+    if (category.metadata.icon.includes("<svg")) {
+      return <Icon svg={category.metadata.icon} className={className} />;
+    } else {
+      return <Icon material={category.metadata.icon} className={className} />;
+    }
+  }
```
ixrock, line 72: **nit**: I'd better to use `const iconProps: IconProps = {};` with `if-else` updates and `return <Icon {...iconProps}/>`.

ixrock (approved):

src/renderer/components/icon/icon.tsx
jakolehm: Those usages expect that `svg` resides within `components/icon` folder. This PR allows to pass raw svg string to `Icon` component.

PR #2610, opened 2021-04-23
Support non-ascii characters in entity names

Nokel81:


jakolehm (approved):
LGTM

jim-docker:
Where can these non-ascii characters come from?

Nokel81:
@jim-docker JS accepts unicode strings.... so the user can specify a display name for a cluster.

nevalla (approved):

PR #2611, opened 2021-04-23
Fix: don't assign to priority in getItemsForKind

Nokel81:
Otherwise `mobx` throws a modification error (if getItemsForKind is within an observable)

### DIFF
--- a/src/extensions/registries/entity-setting-registry.ts
+++ b/src/extensions/registries/entity-setting-registry.ts
@@ -35,10 +35,6 @@ export class EntitySettingRegistry extends BaseRegistry<EntitySettingRegistratio
   getItemsForKind(kind: string, apiVersion: string, source?: string) {
     let items = this.getItems().filter((item) => {
       return item.kind === kind && item.apiVersions.includes(apiVersion);
-    }).map((item) => {
-      item.priority = item.priority ?? 50;
-
-      return item;
     });
 
     if (source) {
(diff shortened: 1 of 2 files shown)


src/extensions/registries/entity-setting-registry.ts
```diff
@@ -47,7 +43,7 @@ export class EntitySettingRegistry extends BaseRegistry<EntitySettingRegistratio
       });
     }
 
-    return items.sort((a, b) => b.priority - a.priority);
+    return items.sort((a, b) => (b.priority ?? 50) - (a.priority ?? 50));
```
jakolehm, line 46: We probably need to fix this also in other places.

Nokel81:
@jakolehm done

jakolehm (approved):

src/extensions/registries/entity-setting-registry.ts
jim-docker, line 46: I guess there's no initializer where the default value can be set, instead of doing this check during sorting? Gone are the days of optimizing code :)

jim-docker (approved):

src/extensions/registries/entity-setting-registry.ts
Nokel81: Oh we could do that instead. Might be better. I would consider that a chore. Will submit a PR (but I think this should be in the next alpha).

PR #2629, opened 2021-04-26
Add browse all tab to catalog page

### LINKED ISSUE #2601: Catalog menu "browse all"
- catalog menu should have top item called "Browse" which lists items from all categories
- catalog categories should be under "Categories"

Mockup:
image

Nokel81:

fixes #2601

### DIFF
--- a/src/renderer/components/+catalog/catalog.tsx
+++ b/src/renderer/components/+catalog/catalog.tsx
@@ -27,7 +27,7 @@ enum sortBy {
 export class Catalog extends React.Component {
   @observable private catalogEntityStore?: CatalogEntityStore;
   @observable.deep private contextMenu: CatalogEntityContextMenuContext;
-  @observable activeTab: string;
+  @observable activeTab?: string;
 
   async componentDidMount() {
     this.contextMenu = {
@@ -89,7 +89,7 @@ export class Catalog extends React.Component {
         <div className="sidebarHeader">Catalog</div>
         <div className="sidebarTabs">
           <Tab
-            value={null}
+            value={undefined}
             key="*"
             label="Browse"
             data-testid="*-tab"
--- a/src/renderer/components/+catalog/catalog.tsx
+++ b/src/renderer/components/+catalog/catalog.tsx
@@ -76,14 +76,11 @@ export class Catalog extends React.Component {
   }
 
   @action
-  onTabChange = (tabId: string) => {
-    this.activeTab = tabId;
+  onTabChange = (tabId: string | null) => {
+    const activeCategory = this.categories.find(category => category.getId() === tabId);
 
-    const activeCategory = this.categories.find((category) => category.getId() === tabId);
-
-    if (activeCategory) {
-      this.catalogEntityStore.activeCategory = activeCategory;
-    }
+    this.catalogEntityStore.activeCategory = activeCategory;
+    this.activeTab = activeCategory?.getId();
   };
 
   renderNavigation() {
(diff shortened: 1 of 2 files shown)


src/renderer/components/+catalog/catalog.tsx
```diff
@@ -92,24 +92,34 @@ export class Catalog extends React.Component {
   }
 
   @action
-  onTabChange = (tabId: string) => {
-    this.activeTab = tabId;
+  onTabChange = (tabId: string | null) => {
```
Nachasic, line 79: Why `tabId` is allowed to be `null`?
It looks like if it is null, `this.activeTab` is going to get reassigned to `undefined` - this looks like potential UB.

Nokel81: UB doesn't exist in JS... I did this because the ID of a tab normally has to be the filtering type, and I chose `null` since that is not a string. I guess I could have chosen `true` or some other marker value.

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -92,24 +92,34 @@ export class Catalog extends React.Component {
...
       <Tabs className="flex column" scrollable={false} onChange={this.onTabChange} value={this.activeTab}>
         <div className="sidebarHeader">Catalog</div>
         <div className="sidebarTabs">
-          { this.categories.map((category, index) => {
-            return <Tab value={category.getId()} key={index} label={category.metadata.name} data-testid={`${category.getId()}-tab`} />;
-          })}
+          <Tab
+            value={null}
+            key="*"
+            label="Browse"
```
jim-docker, line 94: Or
```suggestion
            label="All Categories"
```

Can this Tab be visually differentiated from the actual Category tabs? Also, it should highlight like the individual categories do.

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -151,7 +157,7 @@ export class Catalog extends React.Component {
         provideBackButtonNavigation={false}
         contentGaps={false}>
         <ItemListLayout
-          renderHeaderTitle={this.catalogEntityStore.activeCategory?.metadata.name}
+          renderHeaderTitle={this.catalogEntityStore.activeCategory?.metadata.name ?? "Browse All"}
```
jim-docker, line 144: I would make this the same as the Tab label, like is done for the actual categories
```suggestion
          renderHeaderTitle={this.catalogEntityStore.activeCategory?.metadata.name ?? "All Categories"}
```

src/renderer/components/+catalog/catalog.tsx
jim-docker: I've seen the mockup, it has differentiation from the categories, but `"Browse"` still seems odd to me. If there's a nested list under CATEGORIES as in the mockup then we might as well support multi-select of the categories, and selecting CATEGORIES acts as select-all? Or checkboxes, with a select-all checkbox next to CATEGORIES?

src/renderer/components/+catalog/catalog.tsx
Nachasic: I guess I used the term incorrectly, sorry. What I meant is that one of potential results of this function now is assigning `undefined` to a value, the type definition of which doesn't allow `undefined`. You'll see what I mean if you add `"strictNullChecks": true,` to `tsconfig.json`:
Screen Shot 2021-04-30 at 1 32 36 PM

If this is intended, it should be reflected in the type:
```ts
// line 31
@observable activeTab?: string; 
```
which would make sense, because health checks for this value are already present in some places, e.g. on line 42:
```
if (!this.activeTab && items.length > 0) {
```
Right now it seems a bit confusing to me.

Nokel81: Oh I guess I should use `string | undefined` instead of `string | null`. Will do

jim-docker (approved):

Nachasic (approved):

PR #2654, opened 2021-04-28
Refactor / fix cluster view visibility

### LINKED ISSUE #2653: Hotbar click does not always open a cluster dashboard
**Describe the bug**
Sometimes hotbar cluster icon click opens a catalog instead of a cluster dashboard.

**To Reproduce**
1. Add cluster A & B to hotbar
2. Click cluster A in hotbar
3. Click cluster B in hotbar
4. In some cases Lens shows Catalog instead of cluster B dashboard

**Expected behavior**
Lens should show clicked cluster dashboard always.

jakolehm:
Fixes #2653

### DIFF
--- a/src/renderer/api/catalog-entity-registry.ts
+++ b/src/renderer/api/catalog-entity-registry.ts
@@ -5,6 +5,7 @@ import "../../common/catalog-entities";
 
 export class CatalogEntityRegistry {
   @observable protected _items: CatalogEntity[] = observable.array([], { deep: true });
+  @observable protected _activeEntity: CatalogEntity;
 
   constructor(private categoryRegistry: CatalogCategoryRegistry) {}
 
@@ -37,6 +38,14 @@ export class CatalogEntityRegistry {
         this._items.splice(index, 1, item);
       }
     });
+  }
+
+  set activeEntity(entity: CatalogEntity) {
+    this._activeEntity = entity;
+  }
+
+  get activeEntity() {
+    return this._activeEntity;
   }
 
   get items() {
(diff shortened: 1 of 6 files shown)


src/renderer/components/hotbar/hotbar-menu.tsx
```diff
@@ -26,7 +25,7 @@ export class HotbarMenu extends React.Component<Props> {
   }
 
   isActive(item: CatalogEntity) {
-    return ClusterStore.getInstance().activeClusterId == item.getId();
+    return catalogEntityRegistry.activeEntity?.metadata?.uid == item.getId();
```
jakolehm, line 28: Changed because hotbar should only know about CatalogEntities.

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -39,6 +40,14 @@ export class CatalogEntityRegistry {
     });
   }
 
+  set activeEntity(entity: CatalogEntity) {
```
jakolehm, line 43: Set/get because we might need to do some additional operations on `set` soon.

aleksfront (approved):

src/renderer/components/cluster-manager/cluster-view.tsx
```diff
@@ -26,15 +27,37 @@ export class ClusterView extends React.Component<Props> {
...
+
+  showCluster(clusterId: string) {
+    initView(clusterId);
+    requestMain(clusterActivateHandler, this.clusterId, false);
+
+    const entity = catalogEntityRegistry.getById(this.clusterId);
+
+    if (entity) {
+      catalogEntityRegistry.activeEntity = entity;
+    }
```
Nokel81, line 50: Why would we not revert to the catalog in the invalid ID state?

Nokel81 (approved):

src/renderer/components/cluster-manager/cluster-view.tsx
jakolehm: Good question, we probably should. I'll this (and few other things) in a follow-up PR.

PR #2655, opened 2021-04-28
fix(renderer-log): multiple spaces merge into one

BlackHole1:
before:
图片

after:
图片

### DIFF
--- a/src/renderer/components/dock/log-list.scss
+++ b/src/renderer/components/dock/log-list.scss
@@ -30,6 +30,7 @@
 
         span {
           -webkit-font-smoothing: auto; // Better readability on non-retina screens
+          white-space: pre;
         }
 
         span.overlay {


Nokel81:
does this fix #2412?

Nokel81 (approved):

BlackHole1:
> does this fix #2412?

I took a look and found that this PR cannot solve this problem 100%, but the situation will improve.

Nokel81:
Thanks for checking.

BlackHole1:
I will take a closer look at the problem tomorrow. Personally, I think it should be easier to fix. If I find a solution, I will submit a new PR.

jakolehm (approved):

BlackHole1:
hi @Nokel81 , I just wrote a test demo and found that this PR can fix #2412 .

before:
图片

after:
图片

my test code:

```golang
package main

import (
	log "github.com/sirupsen/logrus"
	"os"
)

func main() {
	log.SetFormatter(&log.TextFormatter{
		ForceColors:            true,
		FullTimestamp:          true,
		DisableLevelTruncation: true,
		PadLevelText:           true,
	})
	log.SetOutput(os.Stdout)
	log.SetLevel(log.TraceLevel)

	log.Error("this is error log")
	log.Warning("this is warning log")
	log.Info("this is info log")
	log.Debug("this is debug log")
	log.Trace("this is trace log")
}

```

Nokel81:
Excellent thank you.

PR #2661, opened 2021-04-28
Ignore failures due to files in bundled folder

Nokel81:

### DIFF
--- a/src/extensions/extension-discovery.ts
+++ b/src/extensions/extension-discovery.ts
@@ -43,6 +43,10 @@ interface ExtensionDiscoveryChannelMessage {
  */
 const isDirectoryLike = (lstat: fse.Stats) => lstat.isDirectory() || lstat.isSymbolicLink();
 
+interface LoadFromFolderOptions {
+  isBundled?: boolean;
+}
+
 /**
  * Discovers installed bundled and local extensions from the filesystem.
  * Also watches for added and removed local extensions by watching the directory.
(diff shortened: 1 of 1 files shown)


nevalla (approved):

src/extensions/extension-discovery.ts
```diff
@@ -327,7 +331,12 @@ export class ExtensionDiscovery extends Singleton {
         isEnabled
       };
     } catch (error) {
-      logger.error(`${logModule}: can't load extension manifest at ${manifestPath}: ${error}`);
+      if (error.code === "ENOTDIR") {
+        // ignore this error, probably from .DS_Store file
+        logger.debug(`${logModule}: failed to load extension manifest through a not-dir-like at ${manifestPath}`);
```
jim-docker, line 336: Does this prevent the popup dialog error message on startup?

Nokel81: No because what ever is causing that couldn't have come from here. The error was always being logged. I just now change what log level it is.

jim-docker (approved):

src/extensions/extension-discovery.ts
jim-docker: Does this PR address an issue?

Nokel81: No just a mild annoyance ever since bundled extensions are no longer being filtered.

PR #2691, opened 2021-05-03
Hotbar inner drag-n-drop

aleksfront:
It allows to restack items within a hotbar.

https://user-images.githubusercontent.com/9607060/116844705-83650780-abec-11eb-9a66-ba551393dc48.mp4


src/renderer/components/hotbar/hotbar-menu.tsx
```diff
@@ -38,36 +35,67 @@ export class HotbarMenu extends React.Component<Props> {
...
+              key={entity ? entity.getId() : `cell${index}`}
+              innerRef={provided.innerRef}
+              className={cssNames({ isDraggingOver: snapshot.isDraggingOver })}
+              {...provided.droppableProps}
+            >
+              {entity && (
+                <Draggable draggableId={item.entity.uid} key={item.entity.uid} index={0}>
+                  {(provided, snapshot) => {
+                    const style = {
+                      zIndex: defaultHotbarCells - index,
```
aleksfront, line 69: Needed for moving cell effect - old cell must cover an empty one.
cell zindex

src/renderer/components/hotbar/hotbar-selector.scss
```diff
@@ -0,0 +1,36 @@
...
+    &:hover {
+      box-shadow: none;
+      background-color: transparent;
+    }
+
+    &.previous {
+      transform: rotateY(180deg);
+    }
+  }
+}
```
Nachasic, line 36: Please add newline at the EOF

Nachasic (approved):
LGTM, beautiful!

src/common/__tests__/hotbar-store.test.ts
```diff
@@ -31,4 +95,137 @@ describe("HotbarStore", () => {
...
+      const items = hotbarStore.getActive().items.map(item => item?.entity.uid || null);
+
+      expect(items.slice(0, 4)).toEqual(["aws", "test", "minikube", null]);
+    });
+
+    it("moves items up", () => {
+      const hotbarStore = HotbarStore.createInstance();
+      const test = new CatalogEntityItem(testCluster);
+      const minikube = new CatalogEntityItem(minikubeCluster);
+      const aws = new CatalogEntityItem(awsCluster);
```
ixrock, line 184: Why do we need such tests? 
Wouldn't be better to check behaviour of `HotbarStore.restackItems()`in dedicated test?

aleksfront: What you mean by dedicated tests? Introduce separate file for dragndrop behaviour or test it within ui tests?

ixrock: I meant instead of having tests like "_moves items **up**/**down/etc**_" better to create separated test-case:

```ts
test(("Hotbar.restackItems() allows to reorganize hotbar items") => {
 // test method logic and just check items-list in the end
})
```

Because in the end you're checking logic for `restackItems()` after some manipulations with involving other methods like `addToHotbar()` (which is against unit-testing nature). Wdyt?

Nokel81:
Looks good but one part of the video was weird (starting at 14s) it seems strange that it moved up items.

jim-docker:
> Looks good but one part of the video was weird (starting at 14s) it seems strange that it moved up items.

I think it's right, when you drag an item down the displaced item, and items above it, should move up. However perhaps the animation of the one being displaced while the user drags should be upward instead of downward? (the second video shows animation downward)

When dragging an item down, animate the displaced item up; when dragging up, animate the displaced item down?

jakolehm (approved):
LGTM

Issue #2663, opened 2021-04-28
Fix Bugs in Kubeconfig Sync


Nokel81:
**Describe the bug**
Once #2567 lands there will be the following items that need to fixed:

- [x] Remove notifcations for invalid contexts.
- [x] Fix catalog being empty after closing and opening window (#2789)
- [x] Change all delete clusters to run `kubectl config delete context --kubeconfig <path>` (#2923)
- [ ] #3439
- [ ] #3438
- [ ] #3437
- [ ] #3436
- [ ] ~~refactor cluster to store the copy (in memory) of the KubeConfig itself so that we both don’t read from the FS as often and so we can do diffing~~
- [x] Find some way to produce consistent IDs so that settings work
- [x] Remove kube config file picker
- [ ] #3435
- [x] Investigate where AWS, GKE, store their configs. So that we can "auto-detect" those clusters. (These seem to be just added to the default location)

jakolehm:
Couple more:

- ~~sync does not watch internal (c&p) folder at all (imho **blocker**)~~
- it's possible to remove default kubeconfig path, bug or feature?
- Preferences say "Sync an individual file or all files in a folder (non-recursive)." although scan _IS_ recursive.

Nokel81:
- Second was feature, see slack discussion.
- If that is the case then it is a bug, I though I checked to make sure that it wasn't recursive.

Nokel81:
Closing in favour of the subtickets.

PR #2692, opened 2021-05-03
Fix kubeconfig-sync issues

jakolehm:
Fixes regressions from #2567

- `CatalogPusher` was not syncing changes to item properties
- `KubeconfigSync` 
  - always created a new `Cluster` object
  - did not create stable `id` for a cluster
  - called `cluster.init()` (for no reason?)
- `ClusterManager` logic was completely broken


src/main/catalog-pusher.ts
```diff
@@ -14,8 +14,8 @@ export class CatalogPusher {
   init() {
     const disposers: Disposer[] = [];
 
-    disposers.push(reaction(() => this.catalog.items, (items) => {
-      broadcastMessage("catalog:items", toJS(items, { recurseEverything: true }));
+    disposers.push(reaction(() => toJS(this.catalog.items, { recurseEverything: true }), (items) => {
```
jakolehm, line 17: Otherwise changes to catalog items won't trigger reaction.

src/main/catalog-sources/kubeconfig-sync.ts
```diff
@@ -24,7 +24,7 @@ export class KubeconfigSyncManager extends Singleton {
   protected static readonly syncName = "lens:kube-sync";
 
   @action
-  startSync(port: number): void {
+  startSync(): void {
```
jakolehm, line 27: Passing port multiple layers down to just for `cluster.init()` felt wrong. Init is anyway called from `ClusterManager`.

src/main/catalog-sources/kubeconfig-sync.ts
```diff
@@ -156,14 +156,13 @@ export function computeDiff(contents: string, source: RootSource, port: number,
...
         try {
-          const cluster = new Cluster({ ...model, id: uuid.v4() });
+          const clusterId = createHash("md5").update(`${filePath}:${contextName}`).digest("hex");
+          const cluster = ClusterStore.getInstance().getById(clusterId) || new Cluster({ ...model, id: clusterId});
 
           if (!cluster.apiUrl) {
             throw new Error("Cluster constructor failed, see above error");
           }
 
-          cluster.init(port);
```
jakolehm, line 165: This is done via `ClusterManager` (for now).

src/main/catalog-sources/kubeconfig-sync.ts
```diff
@@ -156,14 +156,13 @@ export function computeDiff(contents: string, source: RootSource, port: number,
       for (const [contextName, model] of models) {
         // add new clusters to the source
         try {
-          const cluster = new Cluster({ ...model, id: uuid.v4() });
+          const clusterId = createHash("md5").update(`${filePath}:${contextName}`).digest("hex");
```
jakolehm, line 159: We need stable id for catalog<->cluster relation.

src/main/catalog-sources/kubeconfig-sync.ts
```diff
@@ -156,14 +156,13 @@ export function computeDiff(contents: string, source: RootSource, port: number,
...
+          const cluster = ClusterStore.getInstance().getById(clusterId) || new Cluster({ ...model, id: clusterId});
```
jakolehm, line 160: Uses saved cluster for base if it exists. Otherwise we don't see any saved changes on catalog items.

src/main/cluster-manager.ts
```diff
@@ -1,24 +1,21 @@
...
 
 const clusterOwnerRef = "ClusterManager";
 
 export class ClusterManager extends Singleton {
-  catalogSource = observable.array<CatalogEntity>([]);
-
   constructor(public readonly port: number) {
     super();
 
-    catalogEntityRegistry.addObservableSource("lens:kubernetes-clusters", this.catalogSource);
```
jakolehm, line 21: Was probably forgotten during the original PR. We don't need this now that the source is `KubeconfigSync`.

src/main/catalog-pusher.ts
Nokel81: Okay... this breaks my mental model of how reactions work.

src/main/catalog-sources/kubeconfig-sync.ts
Nokel81: So I have noticed that all of `Cluster.init()` can be hoisted into `Cluster.constructor`.

Nokel81: Also I specifically didn't want to rely on `ClusterManager`. Wouldn't it be possible to query this for a cluster ID instead?

src/main/catalog-sources/kubeconfig-sync.ts
Nokel81: Sounds good, I already knew we would need this

src/main/catalog-sources/kubeconfig-sync.ts
Nokel81: If `clusterId` already exists we shouldn't even be in this code path...

src/main/cluster-manager.ts
Nokel81: See #2663, there is a lot that I knew I didn't to do. Was planning to do this in the future.

Nokel81 (approved):

jim-docker (approved):

Nokel81:
Merging as this fixes the problem with not being able to connect. I will continue investigating why installing lens prometheus doesn't seem to work.

PR #2698, opened 2021-05-03
Cleanup Cluster, ClusterManger, et al.

Nokel81:
- Removed `getFreePort` as its use is always a race condition. Change all uses of it to retrive the port after listening

- Remove `Cluster.ownerRef` as it is outdated and no longer needed for 5.0

- Remove `Cluster.enabled`, no longer needed because of above

- Removed `Cluster.init`, moved its contents into `Cluster.constructor` as nothing in that function is asyncronous. Currently only being run on `main` as a stop gap until `renderer` gets its own version of `Cluster`

- Refactored `LensProxy` so as to prevent `pty.node` (a NodeJS extension) being included in `webpack.extension.ts`'s run

- Removed the passing around of the proxy port as that can now be accessed from an instance of `LensProxy`

These are the fixes required to make #2692 more bullet proof. As well as some refactors that we discussed.


src/main/routes/port-forward-route.ts
```diff
@@ -2,48 +2,58 @@ import { LensApiRequest } from "../router";
...
+
+interface PortForwardArgs {
+  clusterId: string;
+  kind: string;
+  namespace: string;
+  name: string;
+  port: string;
+}
+
+const internalPortRegex = /^forwarding from (?<address>.+) ->/i;
```
jim-docker, line 18: relying on stdout of `kubectl port-forward` to contain this string could be fragile?

src/main/routes/port-forward-route.ts
```diff
@@ -58,8 +68,12 @@ class PortForward {
       }
     });
 
+    this.internalPort = await getPortFrom(this.process.stdout, {
```
jim-docker, line 71: Should this be in the try/catch?

jim-docker (approved):

src/main/routes/port-forward-route.ts
Nokel81: Probably, but so was the old mechanism. I will do some investigation tomorrow into seeing if I can ask the OS what ports a certain process is listening on. I wish it was possible to pass in another stream or maybe a socket for this sort of information instead.

Anther option is figuring out if `client-javascript` has this sort of functionality instead.

src/main/routes/port-forward-route.ts
Nokel81: I sort of want the error to be thrown. That way it gets passed to the renderer.

src/main/routes/port-forward-route.ts
jakolehm: I think regex is fine here.

> Anther option is figuring out if client-javascript has this sort of functionality instead.

We cannot use client-javascript as kubectl replacement... that's how it worked in the early days and it was not working well (many auth configurations did not work at all).

src/main/utils/get-port.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+/**
+ * Parse lines from `stream` (assumes data comes in lines) to find the port
+ * which the source of the stream is watching on.
+ * @param stream A readable stream to match lines against
+ * @param args The args concerning the stream
+ * @returns A Promise for port number
+ */
+export function getPortFrom(stream: Readable, args: GetPortArgs): Promise<number> {
```
jakolehm, line 28: Would make sense to write tests for this.

jakolehm (approved):

jakolehm:
It seems that GH actions did not run tests .... they are failing.

src/main/utils/get-port.ts
Nokel81: Very much so

PR #2706, opened 2021-05-05
Fix ClusterScoped kube APIs not subscribing

Nokel81:

fixes regression from #2266

### DIFF
--- a/src/renderer/kube-object.store.ts
+++ b/src/renderer/kube-object.store.ts
@@ -1,7 +1,7 @@
 import type { ClusterContext } from "./components/context";
 
 import { action, computed, observable, reaction, when } from "mobx";
-import { autobind, noop, rejectPromiseBy } from "./utils";
+import { autobind, bifurcateArray, noop, rejectPromiseBy } from "./utils";
 import { KubeObject, KubeStatus } from "./api/kube-object";
 import { IKubeWatchEvent } from "./api/kube-watch-api";
 import { ItemStore } from "./item.store";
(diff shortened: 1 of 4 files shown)


src/renderer/kube-object.store.ts
```diff
@@ -281,21 +281,38 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
-        } else {
-          apis.forEach(api => {
-            this.loadedNamespaces.forEach((namespace) => {
-              this.watchNamespace(api, namespace, abortController);
-            });
-          });
-        }
-      })
-      .catch(noop); // ignore DOMExceptions
+    console.log([clusterScopedApis, namespaceScopedApis]);
```
jim-docker: debug?

src/renderer/kube-object.store.ts
```diff
@@ -281,21 +281,38 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
...
+
+    for (const api of namespaceScopedApis) {
+      const store = apiManager.getStore(api);
+
+      // This waits for the context and namespaces to be ready or fails fast if the disposer is called
+      Promise.race([rejectPromiseBy(abortController.signal), Promise.all([store.contextReady, store.namespacesReady])])
```
jim-docker, line 290: It was also a bug that this race was performed only for the `this` store?

jim-docker (dismissed):

src/renderer/kube-object.store.ts
Nokel81: Yes I think so. Though I now that I think about it. Subscribing really should be reactive on the change set of selected namespaces (adding and removing the watches as the namespaces get selected and deselected).

But I think that is out of scope of this fix.

jakolehm (approved):

nevalla (approved):
LGTM. Having integration test addressing this would be nice too!

Nokel81:
> LGTM. Having integration test addressing this would be nice too!

I'll make an issue about that since I agree.

PR #2707, opened 2021-05-05
Refactor PrometheusProviders

Nokel81:
~~Don't merge until #2712 has been merged to master~~

- Remove implementation leak by passing in query name

- Remove renderer assuming static list by providing information from main

- Fix up cluster features (a bit) so that they properly track an individual cluster

- Catch errors from installing lens metrics, and propagate the errors from main to renderer

### DIFF
--- a/src/main/__test__/context-handler.test.ts
+++ b/src/main/__test__/context-handler.test.ts
@@ -1,0 +1,210 @@
+/**
+ * Copyright (c) 2021 OpenLens Authors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import { UserStore } from "../../common/user-store";
+import { ContextHandler } from "../context-handler";
+import { PrometheusProvider, PrometheusProviderRegistry, PrometheusService } from "../prometheus";
+
+enum ServiceResult {
+  Success,
+  Failure,
+  Undefined,
+}
+
+class TestProvider extends PrometheusProvider {
+  name = "TestProvider1";
+  rateAccuracy = "1h";
+  isConfigurable = false;
+
+  constructor(public id: string, public alwaysFail: ServiceResult) {
+    super();
+  }
+
+  getQuery(): string {
+    throw new Error("getQuery is not implemented.");
+  }
+
+  async getPrometheusService(): Promise<PrometheusService> {
+    switch (this.alwaysFail) {
+      case ServiceResult.Success:
+        return {
+          id: this.id,
+          namespace: "default",
+          port: 7000,
+          service: "",
+        };
+      case ServiceResult.Failure:
+        throw new Error("does fail");
+      case ServiceResult.Undefined:
+        return undefined;
+    }
+  }
+}
+
+function getHandler() {
+  return new ContextHandler(({
+    getProxyKubeconfig: (): any => ({
+      makeApiClient: (): any => undefined,
+    }),
+    apiUrl: "http://localhost:81",
+  }) as any);
+}
+
+describe("ContextHandler", () => {
+  beforeEach(() => {
+    PrometheusProviderRegistry.createInstance();
+    UserStore.createInstance();
+  });
+
+  afterEach(() => {
+    PrometheusProviderRegistry.resetInstance();
+    UserStore.resetInstance();
+  });
+
+  describe("getPrometheusService", () => {
+    it.each([
+      [0, 0],
+      [0, 1],
+      [0, 2],
+      [0, 3],
+    ])("should return undefined from %d success(es) after %d failure(s)", async (successes, failures) => {
+      const reg = PrometheusProviderRegistry.getInstance();
+      let count = 0;
+
+      for (let i = 0; i < failures; i += 1) {
+        const serviceResult = i % 2 === 0 ? ServiceResult.Failure : ServiceResult.Undefined;
+
+        reg.registerProvider(new TestProvider(`id_${count++}`, serviceResult));
+      }
+
+      for (let i = 0; i < successes; i += 1) {
+        reg.registerProvider(new TestProvider(`id_${count++}`, ServiceResult.Success));
+      }
+
+      const service = await getHandler().getPrometheusService();
+
+      expect(service).toBeUndefined();
+    });
+
+    it.each([
+      [1, 0],
+      [1, 1],
+      [1, 2],
+      [1, 3],
+      [2, 0],
+      [2, 1],
+      [2, 2],
+      [2, 3],
+    ])("should pick the first provider of %d success(es) after %d failure(s)", async (successes, failures) => {
+      const reg = PrometheusProviderRegistry.getInstance();
+      let count = 0;
+
+      for (let i = 0; i < failures; i += 1) {
+        const serviceResult = i % 2 === 0 ? ServiceResult.Failure : ServiceResult.Undefined;
+
+        reg.registerProvider(new TestProvider(`id_${count++}`, serviceResult));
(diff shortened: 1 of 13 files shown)


src/extensions/cluster-feature.ts
```diff
@@ -39,29 +39,31 @@ export abstract class ClusterFeature {
     canUpgrade: false
   };
 
+  constructor(protected cluster: KubernetesCluster){}
+
   /**
    * to be implemented in the derived class, this method is typically called by Lens when a user has indicated that this feature is to be installed. The implementation
    * of this method should install kubernetes resources using the applyResources() method, or by directly accessing the kubernetes api (K8sApi)
    *
    * @param cluster the cluster that the feature is to be installed on
```
jim-docker: remove this (and similarly below)

src/main/router.ts
```diff
@@ -5,7 +5,7 @@ import path from "path";
 import { readFile } from "fs-extra";
 import { Cluster } from "./cluster";
 import { apiPrefix, appName, publicPath, isDevelopment, webpackDevServerPort } from "../common/vars";
-import { helmRoute, kubeconfigRoute, metricsRoute, portForwardRoute, resourceApplierRoute, versionRoute } from "./routes";
+import { HelmApiRoute, KubeconfigRoute, MetricsRoute, PortForwardRoute, ResourceApplierApiRoute, VersionRoute } from "./routes";
```
jim-docker: This PR is pretty big, can these router changes be done in another PR?

src/renderer/components/cluster-settings/components/cluster-prometheus-setting.tsx
```diff
@@ -1,16 +1,18 @@
...
-import { observable, computed, autorun } from "mobx";
+import { observable, autorun, computed } from "mobx";
+import { metricsApi } from "../../../api/endpoints/metrics.api";
+import { Spinner } from "../../spinner";
 
-const options: SelectOption<string>[] = [
-  { value: "", label: "Auto detect" },
-  ...prometheusProviders.map(pp => ({value: pp.id, label: pp.name}))
-];
+interface MetricProviderInfo {
```
jim-docker: This is already defined in `metrics.api.ts`

extensions/metrics-cluster-feature/renderer.tsx
```diff
@@ -12,42 +24,73 @@ export default class ClusterMetricsFeatureExtension extends LensRendererExtensio
...
         onClick: async () => {
-          metricsFeature.install(cluster);
+          try {
+            Component.Notifications.info(`Lens Metrics is being installed to ${cluster.metadata.name}`, { timeout: 7_500 });
 
-          Component.Notifications.info(`Lens Metrics is now installed to ${cluster.metadata.name}`, { timeout: 10_000 });
+            await metricsFeature.install();
+            await poll(() => metricsFeature.updateStatus(), ({ installed }) => installed, 5_000);
+
+            Component.Notifications.info(`Lens Metrics has been installed to ${cluster.metadata.name}`, { timeout: 7_500 });
```
jim-docker: Minor UX thing: Is there a way to dismiss the first notification ("Lens Metrics is being installed...") when posting this one? It looks a little odd/cluttered to see both at the same time (when it takes less than 7.5 seconds to install)

src/main/router.ts
Nokel81: Good idea, will do

extensions/metrics-cluster-feature/renderer.tsx
Nokel81: Yes we can

nevalla:
Just curious were there some issue with `LensApi` where `respondJson` and `respondText` are now exported as function?

Nokel81:
> Just curious were there some issue with `LensApi` where `respondJson` and `respondText` are now exported as function?

Do you mean in the past? Because to me there are just helper functions are they work entirely on their inputs

jim-docker (dismissed):

src/extensions/cluster-feature.ts
```diff
@@ -39,96 +39,96 @@ export abstract class ClusterFeature {
     canUpgrade: false
   };
 
+  constructor(protected cluster: KubernetesCluster){}
```
jim-docker: Are we tracking breaking changes?

Nokel81: We should be, @jakolehm is going to be change this even more (I think eventually removing this class all together). We should have a doc somewhere for keeping track of all the breaking changes.

nevalla:
Good job! I think after these changes it would be relatively easy to expose provider registry to extension api?

Makefile
```diff
@@ -98,6 +98,7 @@ copy-extension-themes:
 src/extensions/npm/extensions/__mocks__:
 	cp -r __mocks__ src/extensions/npm/extensions/
 
+.PHONY: src/extensions/npm/extensions/dist
```
jakolehm: Could you revert this change?

Nokel81:
@nevalla Yes it should be, since anything that behaves like these black boxes should(tm) work

Nokel81:
@jakolehm PTAL

jakolehm (commented):
LGTM (expect Makefile change)

Makefile
Nokel81: Sure, though I always find that this correct, namely that the extensions should almost always be recreated.

Nokel81: reverted

src/main/context-handler.ts
```diff
@@ -72,18 +72,34 @@ export class ContextHandler {
...
+    }
+
+    return Array.from(registry.providers.values());
+  }
+
+  async getPrometheusService(): Promise<PrometheusService | undefined> {
+    const providers = this.listPotentialProviders();
+    const proxyConfig = await this.cluster.getProxyKubeconfig();
+    const apiClient = proxyConfig.makeApiClient(CoreV1Api);
+    const potentialServices = await Promise.allSettled(
```
jakolehm, line 92: Does this alter priority of potential services? (test that confirms would be nice)

jakolehm (dismissed):

src/main/context-handler.ts
Nokel81: It does not, it just means that if one of the `provider.getPrometheusService()` calls rejects, we can still proceed (namely it is skipped).

Nokel81: Test added.

nevalla (dismissed):
Tested and seems to work

src/main/__test__/context-handler.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+      const service = await getHandler().getPrometheusService();
+
+      expect(service.id === "id_foobar");
+    });
+
+    it("should pick the first provider of 1 success, after 2 fails", async () => {
+      PrometheusProviderRegistry.getInstance()
+        .registerProvider(new TestProvider("id_foobar_1", AlwaysFail.Throw))
+        .registerProvider(new TestProvider("id_foobar_2", AlwaysFail.No))
```
jim-docker: Did you mean
```suggestion
        .registerProvider(new TestProvider("id_foobar_2", AlwaysFail.Return))
```

src/main/__test__/context-handler.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      const service = await getHandler().getPrometheusService();
+
+      expect(service.id === "id_foobar_3");
+    });
+
+    it("should pick the first provider of 1 success, between 2 fails", async () => {
+      PrometheusProviderRegistry.getInstance()
+        .registerProvider(new TestProvider("id_foobar_1", AlwaysFail.Throw))
+        .registerProvider(new TestProvider("id_foobar_2"))
+        .registerProvider(new TestProvider("id_foobar_3", AlwaysFail.No));
```
jim-docker: same here

src/main/__test__/context-handler.test.ts
Nokel81: Yes I did.

src/main/__test__/context-handler.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+      const service = await getHandler().getPrometheusService();
+
+      expect(service.id === "id_foobar_2");
```
jim-docker: this should've failed then?

Nokel81: No this is still correct because `AlwaysFail.No` is the default, and by design we pick the first one that succeeds.

jim-docker: Then the above test should've failed?

Nokel81: Maybe? will investigate.

nevalla (dismissed):

jim-docker (dismissed):
Did you figure out why that test (`should pick the first provider of 1 success, after 2 fails`) passed before you corrected it? Was it not run? Otherwise it's not a good test, right?

Nokel81:
@jim-docker That test was correct in passing because it picked the first one that succeeded, the fact that the third one also succeeded shouldn't change that.

src/main/__test__/context-handler.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    it("should pick the first provider of 1 success, after 2 fails", async () => {
+      PrometheusProviderRegistry.getInstance()
+        .registerProvider(new TestProvider("id_foobar_1", ServiceResult.Failure))
+        .registerProvider(new TestProvider("id_foobar_2", ServiceResult.Undefined))
+        .registerProvider(new TestProvider("id_foobar_3", ServiceResult.Success));
+
+      const service = await getHandler().getPrometheusService();
+
+      expect(service.id === "id_foobar_3");
```
jim-docker: Before you corrected this test the first one that succeeded would've been "id_foobar_2", so this line should've failed, right?

Nokel81: Oh right, I thought you were talking about the third test. Will investigate.

Nokel81: I don't know why this test previously was passing. But I have added many more tests, and even one to validate that that specific case doesn't happen.

jim-docker (approved):

nevalla (approved):

PR #2708, opened 2021-05-05
Formalize the to/from Store rules for UserStore

Nokel81:
- Explicitly don't save to disk if the value in the store is the same as
  the default. That way we can change the defaults between versions

- Calculate the PreferencesModel on demand

- In the future, when we turn on strict null checking, we can enforce
  that the UserStore expects the correct types too

### DIFF
--- a/src/common/user-store/preferences-helpers.ts
+++ b/src/common/user-store/preferences-helpers.ts
@@ -1,0 +1,237 @@
+/**
+ * Copyright (c) 2021 OpenLens Authors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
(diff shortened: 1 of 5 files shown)


src/common/user-store/preferences-helpers.ts
```diff
@@ -0,0 +1,216 @@
...
+    return val;
+  },
+};
+
+const localeTimezone: PreferenceDescription<string> = {
+  fromStore(val) {
+    return val || moment.tz.guess(true) || "UTC";
+  },
+  toStore(val) {
+    if (!val || val === (moment.tz.guess(true) || "UTC")) {
```
Nachasic: `|| ""UTC` predicate is always true due to type conversion, did you mean this?

```suggestion
    if (!val || val === (moment.tz.guess(true) || val === "UTC")) {
```

Nachasic (changes requested):

src/common/user-store/preferences-helpers.ts
Nokel81: I will change this because it confusing to read. But your suggested fix isn't correct either.

Notice the brackets around `moment.tz.guess(true)` and `"UTC"`. Those mean that `val` is compared strict equals to either the guess or `"UTC"` if it couldn't guess.

Nachasic: Oh, I misread that, sorry. Thanks for clarification

Nachasic (dismissed):
LGTM

Nachasic (dismissed):

src/common/user-store/preferences-helpers.ts
```diff
@@ -0,0 +1,237 @@
...
+import { ObservableToggleSet } from "../utils";
+
+export interface KubeconfigSyncEntry extends KubeconfigSyncValue {
+  filePath: string;
+}
+
+export interface KubeconfigSyncValue { }
+
+interface PreferenceDescription<T, R = T> {
+  fromStore(val: T | undefined): R;
```
Nachasic, line 35: Doesn't "serialize/deserialize" sound more fitting here?

Nachasic (dismissed):

src/common/user-store/preferences-helpers.ts
Nokel81: Serialization & Deserialization are very specific things. They are the transformation from one format to another. Specifically with a parsing step.

If we are not parsing, and if a function called `serialize()` doesn't produce either a `string` or a `U8intBuffer` or similar I would be against such a name.

Nachasic (approved):

PR #2710, opened 2021-05-06
Keep disabled Hotbar entries visible

### LINKED ISSUE #2650: Adding a cluster to the hotbar should override "empty" entries
**Describe the bug**
Now that the hotbar has "empty" entries by default (see #2638) (though I don't know why they actually exist instead of just being placeholders). If you add a cluster they should override them.

This is what happens if you add a cluster currently:
Screen Shot 2021-04-27 at 3 52 43 PM

@aleksfront ?

aleksfront:
When items get deleted from catalog (technically checking their existence inside in `catalogEntityRegistry.items`) - associated hotbar items became `disabled`.

Disabled state means it can't be clicked, has dimmed styles and has one menu option `Remove from Hotbar`.

https://user-images.githubusercontent.com/9607060/117285386-a273cb80-ae70-11eb-9449-25f2c9b910e4.mp4

Now, instead of "reserver" empty cells user will see disabled ones.

Fixes #2650 

In addition, added some more tooltip styling.

### DIFF
--- a/src/common/hotbar-store.ts
+++ b/src/common/hotbar-store.ts
@@ -8,8 +8,8 @@ import isNull from "lodash/isNull";
 export interface HotbarItem {
   entity: {
     uid: string;
-    name: string;
-    source: string;
+    name?: string;
+    source?: string;
   };
   params?: {
     [key: string]: string;
--- a/src/common/hotbar-store.ts
+++ b/src/common/hotbar-store.ts
@@ -8,6 +8,8 @@ import isNull from "lodash/isNull";
 export interface HotbarItem {
   entity: {
     uid: string;
+    name: string;
+    source: string;
   };
   params?: {
     [key: string]: string;
@@ -123,9 +125,14 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
     }
   }
 
+  @action
   addToHotbar(item: CatalogEntityItem, cellIndex = -1) {
     const hotbar = this.getActive();
-    const newItem = { entity: { uid: item.id }};
+    const newItem = { entity: {
+      uid: item.id,
+      name: item.name,
+      source: item.source
+    }};
 
     if (hotbar.items.find(i => i?.entity.uid === item.id)) {
       return;
@@ -146,6 +153,7 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
     }
   }
 
+  @action
   removeFromHotbar(uid: string) {
     const hotbar = this.getActive();
     const index = hotbar.items.findIndex((i) => i?.entity.uid === uid);
--- a/src/migrations/hotbar-store/5.0.0-alpha.0.ts
+++ b/src/migrations/hotbar-store/5.0.0-alpha.0.ts
@@ -22,11 +22,7 @@ export default migration({
       }
 
       hotbar.items.push({
-        entity: {
-          uid: cluster.id,
-          name: cluster.name,
-          source: cluster.source
-        },
+        entity: { uid: cluster.id },
         params: {}
       });
     });
--- a/src/migrations/hotbar-store/5.0.0-alpha.0.ts
+++ b/src/migrations/hotbar-store/5.0.0-alpha.0.ts
@@ -22,7 +22,11 @@ export default migration({
       }
 
       hotbar.items.push({
-        entity: { uid: cluster.id },
+        entity: {
+          uid: cluster.id,
+          name: cluster.name,
+          source: cluster.source
+        },
         params: {}
       });
     });
--- a/src/migrations/hotbar-store/5.0.0-beta.4.ts
+++ b/src/migrations/hotbar-store/5.0.0-beta.4.ts
@@ -1,0 +1,30 @@
+import { Hotbar } from "../../common/hotbar-store";
+import { migration } from "../migration-wrapper";
+import { catalogEntityRegistry } from "../../renderer/api/catalog-entity-registry";
+
+export default migration({
+  version: "5.0.0-beta.4",
+  run(store) {
+    const hotbars: Hotbar[] = store.get("hotbars");
+
+    hotbars.forEach((hotbar, hotbarIndex) => {
+      hotbar.items.forEach((item, itemIndex) => {
+        const entity = catalogEntityRegistry.items.find((entity) => entity.metadata.uid === item?.entity.uid);
+
+        if (!entity) {
+          // Clear disabled item
+          hotbars[hotbarIndex].items[itemIndex] = null;
+        } else {
+          // Save additional data
+          hotbars[hotbarIndex].items[itemIndex].entity = {
+            ...item.entity,
+            name: entity.metadata.name,
+            source: entity.metadata.source
+          };
+        }
+      });
+    });
+
+    store.set("hotbars", hotbars);
+  }
+});
--- a/src/migrations/hotbar-store/5.0.0-beta.4.ts
+++ b/src/migrations/hotbar-store/5.0.0-beta.4.ts
@@ -1,4 +1,4 @@
-import { Hotbar } from "../../common/hotbar-store";
+import type { Hotbar } from "../../common/hotbar-store";
 import { migration } from "../migration-wrapper";
 import { catalogEntityRegistry } from "../../renderer/api/catalog-entity-registry";
 
--- a/src/migrations/hotbar-store/index.ts
+++ b/src/migrations/hotbar-store/index.ts
@@ -7,5 +7,5 @@ import version500beta4 from "./5.0.0-beta.4";
 export default {
   ...version500alpha0,
   ...version500alpha2,
-  ...version500beta4
+  ...version500beta4,
 };
--- a/src/migrations/hotbar-store/index.ts
+++ b/src/migrations/hotbar-store/index.ts
@@ -2,8 +2,10 @@
 
 import version500alpha0 from "./5.0.0-alpha.0";
 import version500alpha2 from "./5.0.0-alpha.2";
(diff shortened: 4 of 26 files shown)


src/common/hotbar-store.ts
```diff
@@ -8,6 +8,8 @@ import isNull from "lodash/isNull";
 export interface HotbarItem {
   entity: {
     uid: string;
+    name: string;
+    source: string;
```
aleksfront: Those extra data needed for disabled HotbarIcon, since entity might be disappeared, but we still need to show name and background color for `disabled` item.

src/renderer/components/+catalog/material-tooltip/material-tooltip.tsx
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import { makeStyles, Tooltip, TooltipProps } from "@material-ui/core";
+
+const useStyles = makeStyles(() => ({
+  arrow: {
+    color: "#18191c",
```
aleksfront: Light theme coloring would come with another PR

src/renderer/components/hotbar/hotbar-entity-icon.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+interface Props extends DOMAttributes<HTMLElement> {
+  entity: CatalogEntity;
+  className?: IClassName;
+  errorClass?: IClassName;
+  remove: (uid: string) => void;
+}
+
+@observer
+export class HotbarEntityIcon extends React.Component<Props> {
```
aleksfront, line 25: I've separated `HotbarEntityIcon` out of `HotbarIcon` to create another layer of abstraction and make `HotbarIcon` more generic regardless of entity availability.

src/renderer/components/+catalog/material-tooltip/material-tooltip.tsx
Nokel81: Wouldn't it make sense to have it in this PR though?

src/common/hotbar-store.ts
Nokel81: Have you tested the flow of changing an entity's name? Does the hotbar icon update accordingly?

Nokel81 (dismissed):

aleksfront:
> Have you tested the flow of changing an entity's name? Does the hotbar icon update accordingly?

Yes, if cluster name is changes via settings, the icon changes accordingly.

super-minikube

src/migrations/hotbar-store/5.0.0-alpha.0.ts
```diff
@@ -22,7 +22,11 @@ export default migration({
       }
 
       hotbar.items.push({
-        entity: { uid: cluster.id },
+        entity: {
```
jakolehm: We need to create a new migration because this will not be applied to existing beta users.

aleksfront: Created new migration in `src/migrations/hotbar-store/5.0.0-beta.4.ts`

src/renderer/components/+catalog/material-tooltip/material-tooltip.tsx
aleksfront: Yes, this changes took a while for some cache issue which I can't solve right now quickly. So don't want to freeze current PR because of that.

src/common/hotbar-store.ts
```diff
@@ -123,9 +125,14 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
     }
   }
 
+  @action
   addToHotbar(item: CatalogEntityItem, cellIndex = -1) {
     const hotbar = this.getActive();
-    const newItem = { entity: { uid: item.id }};
+    const newItem = { entity: {
+      uid: item.id,
+      name: item.name,
```
Nachasic, line 133: This change seems to have broken a migration in `src/migrations/hotbar-store/5.0.0-alpha.0.ts(25,9)`:
```
TS2739: Type '{ uid: any; }' is missing the following properties from type '{ uid: string; name: string; source: string; }': name, source
```

Nachasic (changes requested):

aleksfront:
@jakolehm @Nachasic Please take another look. I've added new migration along with related tests.

src/migrations/hotbar-store/5.0.0-beta.4.ts
```diff
@@ -0,0 +1,30 @@
+import { Hotbar } from "../../common/hotbar-store";
```
Nokel81: This prevents a circular dependency.

```suggestion
import type { Hotbar } from "../../common/hotbar-store";
```

src/migrations/hotbar-store/index.ts
```diff
@@ -2,8 +2,10 @@
 
 import version500alpha0 from "./5.0.0-alpha.0";
 import version500alpha2 from "./5.0.0-alpha.2";
+import version500beta4 from "./5.0.0-beta.4";
 
 export default {
   ...version500alpha0,
-  ...version500alpha2
+  ...version500alpha2,
+  ...version500beta4
```
Nokel81: Please add a comma here so that future diffs are smaller. (I am thinking of adding a lint for this)

src/renderer/components/+catalog/material-tooltip/material-tooltip.tsx
Nokel81: Cache issue? What is wrong with just putting these into the default MuiProvider which is called in `bootstrap.tsx`? That way we don't even need this special component.

src/renderer/components/hotbar/hotbar-cell.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export function HotbarCell({ innerRef, children, className, ...rest }: Props) {
+  const [animating, setAnimating] = useState(false);
+  const onAnimationEnd = () => { setAnimating(false); };
+  const onClick = () => {
+    if (className.includes("isDraggingOver")) {
+      return;
+    }
+
+    setAnimating(true);
```
Nokel81: I think the following would have the same effect.

```suggestion
    setAnimating(!className.includes("isDraggingOver"));
```

src/renderer/components/+catalog/material-tooltip/material-tooltip.tsx
aleksfront: Recently we removed global styles pointing MuiProvider to allow redesign tooltips from different places. Meanwhile, the cache bug is fixed with `make dev`, so css variables are set here. So this issue is fixed and no longer relevant.

Nachasic (dismissed):
LGTM

src/migrations/hotbar-store/5.0.0-beta.4.ts
```diff
@@ -0,0 +1,30 @@
+import type { Hotbar } from "../../common/hotbar-store";
+import { migration } from "../migration-wrapper";
+import { catalogEntityRegistry } from "../../renderer/api/catalog-entity-registry";
+
+export default migration({
+  version: "5.0.0-beta.4",
```
jakolehm: I think this needs to be bumped to `beta.5` because `beta.4` has already been released.

src/renderer/components/hotbar/hotbar-selector.tsx
```diff
@@ -2,43 +2,31 @@ import "./hotbar-selector.scss";
...
   const store = HotbarStore.getInstance();
   const activeIndexDisplay = store.activeHotbarIndex + 1;
-  const classes = useStyles();
 
   return (
     <div className="HotbarSelector flex align-center">
       <Icon material="play_arrow" className="previous box" onClick={() => store.switchToPrevious()} />
       <div className="box grow flex align-center">
-        <Tooltip arrow title={hotbar.name} classes={classes}>
+        <MaterialTooltip arrow title={hotbar.name}>
```
Nokel81, line 22: Nit: is there a reason why we don't just override the default CSS. That way we don't need to write out own component (if all we are doing is changing the styles).

Or is that not a good direction to take this is?

aleksfront: Yes, `MaterialTooltip` used as a styled wrapper for Tooltip component. It encapsulates styles inside and allows to use default (or any other) styling for material tooltip from other places.

jakolehm: Is `MaterialTooltip` specific to just Catalog? If yes: why catalog tooltips are different?

Nachasic (approved):

src/renderer/components/hotbar/hotbar-selector.tsx
aleksfront: It's replacement for our native `Tooltip`. Used in couple of places so far.

Nokel81: I still don't understand why we don't just set default values for the `Tooltip` styling and then not need a new component.

jakolehm: > It's replacement for our native `Tooltip`. Used in couple of places so far.

I think the location of that component is not right if it's a reusable component. Is there some reason not to replace all uses of the old `Tooltip`?

jakolehm (approved):
LGTM. We should move `MaterialTooltip` under `components` in follow-up PR (or merge with `Tooltip`).

Nokel81:
I agree with @jakolehm I think that all our uses of `<Tooltip>` should be replaced with the one from Material UI.

Futhermore, I think that this should be done by setting defaults and importing `Tooltip` from `@material-ui/core` directly.

PR #2712, opened 2021-05-06
Switch main's route handlers to be static functions

Nokel81:


src/main/lens-api.ts
```diff
@@ -1,17 +1,15 @@
 import http from "http";
```
jim-docker: guess this file could be now named `src/main/utils/http-respond.ts` or similar?

jim-docker (dismissed):
Why do we need this? Cleanup/modernize?

src/main/lens-api.ts
Nokel81: Probably would be a better name.

Nokel81:
> Why do we need this? Cleanup/modernize?

Mostly cleanup. These don't need to be instance methods. The fact that they are instance methods is slightly confusing because (one would hope) that all of them don't store state (even though I know that this change doesn't do anything to enforce this).

jim-docker (approved):

nevalla (approved):
LGTM

PR #2714, opened 2021-05-06
Show lens-metrics on cluster settings

jakolehm:
## Breaking changes

- removes `ClusterFeature` (replaced by `K8sApi.ResourceStack`)
- removes icon from entity context menus (makes UI more calm)

## Global CSS changes

- improved `Menu` component styles (thanks @miskun for help)

## Other

- allow to specify cluster settings group via api
- allow to enable each metrics component separately
- show warning if other metrics provider is used or if cluster is offline
- update Prometheus/node-exporter/kube-state-metrics to latest stable version

image
image
image
image


extensions/metrics-cluster-feature/src/metrics-settings.tsx
```diff
@@ -0,0 +1,220 @@
...
+          <Component.FormSwitch
+            control={
+              <Component.Switcher
+                disabled={this.featureStates.kubeStateMetrics === undefined || !this.props.cluster.status.active}
+                checked={!!this.featureStates.prometheus && this.props.cluster.status.active}
+                onChange={v => this.togglePrometheus(v.target.checked)}
+                name="prometheus"
+              />
+            }
+            label="Install built-in Prometheus metrics stack"
```
jim-docker:
```suggestion
            label="Install bundled Prometheus metrics stack"
```

extensions/metrics-cluster-feature/src/metrics-settings.tsx
```diff
@@ -0,0 +1,220 @@
...
+          <Component.FormSwitch
+            control={
+              <Component.Switcher
+                disabled={this.featureStates.kubeStateMetrics === undefined || !this.props.cluster.status.active}
+                checked={!!this.featureStates.kubeStateMetrics && this.props.cluster.status.active}
+                onChange={v => this.toggleKubeStateMetrics(v.target.checked)}
+                name="node-exporter"
+              />
+            }
+            label="Install built-in kube-state-metrics stack"
```
jim-docker:
```suggestion
            label="Install bundled kube-state-metrics stack"
```

extensions/metrics-cluster-feature/src/metrics-settings.tsx
```diff
@@ -0,0 +1,220 @@
...
+          <Component.FormSwitch
+            control={
+              <Component.Switcher
+                disabled={this.featureStates.nodeExporter === undefined || !this.props.cluster.status.active}
+                checked={!!this.featureStates.nodeExporter && this.props.cluster.status.active}
+                onChange={v => this.toggleNodeExporter(v.target.checked)}
+                name="node-exporter"
+              />
+            }
+            label="Install built-in node-exporter stack"
```
jim-docker:
```suggestion
            label="Install bundled node-exporter stack"
```

extensions/metrics-cluster-feature/src/metrics-settings.tsx
```diff
@@ -0,0 +1,220 @@
...
+      }
+    }
+  }
+
+  async save() {
+    this.config.prometheus.enabled = !!this.featureStates.prometheus;
+    this.config.kubeStateMetrics.enabled = !!this.featureStates.kubeStateMetrics;
+    this.config.nodeExporter.enabled = !!this.featureStates.nodeExporter;
+
+    if (!this.config.prometheus.enabled && !this.config.kubeStateMetrics.enabled && !this.config.nodeExporter.enabled) {
```
jim-docker: this means I can't delete/disable just one of them? It's a misleading UX. I think it works better if there's an install button. The user picks which of the three components they want, then clicks install. After this the toggles aren't selectable. If the user wants to delete any components user must click the uninstall button (which was the install button) and then try again.

src/main/resource-applier.ts
```diff
@@ -64,19 +73,24 @@ export class ResourceApplier {
...
 
-      console.log("shooting manifests with:", cmd);
-      exec(cmd, (error, stdout, stderr) => {
+      logger.info(`[RESOURCE-APPLIER] running cmd ${cmd}`);
+      exec(cmd, (error, stdout) => {
         if (error) {
-          reject(`Error applying manifests:${error}`);
-        }
+          logger.error(`[RESOURCE-APPLIER] cmd errored: ${error}`);
+          const splittedError = error.toString().split(`.yaml": `);
```
jim-docker:
```suggestion
          const splitError = error.toString().split(`.yaml": `);
```
past tense of split is ... split  :)

jim-docker (commented):
I still find the metrics settings confusing. We have the "Metrics" tab and the "Lens Metrics" tab, but "Lens Metrics" is not really relevant unless "Lens" is chosen for the "Prometheus installation method" (or "Auto detect", if "Lens Metrics" have already been installed)

Maybe "Metrics Stack" instead of "Metrics". 
Maybe "Bundled Metrics" instead of "Lens Metrics"
And maybe on the "Lens/Bundled Metrics" page there should be some text indicating that this is configuration that is relevant when the "Prometheus installation method" is set to "Lens/Bundled"
Does Auto detect pick Lens metrics over any other stack that might be installed?

jakolehm:
> I still find the metrics settings confusing. We have the "Metrics" tab and the "Lens Metrics" tab, but "Lens Metrics" is not really relevant unless "Lens" is chosen for the "Prometheus installation method" (or "Auto detect", if "Lens Metrics" have already been installed)
> 

I completely agree. I just refactored the UI to make this more clear + added warning if metrics settings are pointing to a different provider (see PR description for updated screenshots).

> Does Auto detect pick Lens metrics over any other stack that might be installed?

Yes, I think so.

extensions/metrics-cluster-feature/src/metrics-settings.tsx
jakolehm: User can delete/disable any of them. Problem with dedicated "install/delete" is the fact that "delete" would wipe all metrics history away. With toggles it's only happening if user chooses to disable "Prometheus" component.

Reason I chose to give freedom here: many times clusters might already have working metrics stack but it's configured so that Lens cannot query it properly. With this PR it's possible to just enable "Prometheus" which adds "Lens compatible" Prometheus service without the need to deploy duplicate node-exporter/kube-state-metrics (which would lead to duplicate metrics (2x cpu etc..))

extensions/metrics-cluster-feature/src/metrics-settings.tsx
jakolehm: Changed 👍

extensions/metrics-cluster-feature/src/metrics-settings.tsx
```diff
@@ -164,13 +187,13 @@ export class MetricsSettings extends React.Component<Props> {
...
               <Component.Switcher
-                disabled={this.featureStates.kubeStateMetrics === undefined || !this.props.cluster.status.active}
+                disabled={this.featureStates.kubeStateMetrics === undefined || !this.isTogglable}
                 checked={!!this.featureStates.prometheus && this.props.cluster.status.active}
                 onChange={v => this.togglePrometheus(v.target.checked)}
                 name="prometheus"
               />
             }
-            label="Install built-in Prometheus metrics stack"
+            label="Install bundled Prometheus metrics stack"
```
jim-docker: Since you changed "installed" to "enabled" above should this be
```suggestion
            label="Enable bundled Prometheus metrics stack"
```
and similarly below?

Or just
```suggestion
            label="Bundled Prometheus metrics stack"
```

extensions/metrics-cluster-feature/renderer.tsx
```diff
@@ -1,55 +1,27 @@
...
-          Component.Notifications.info(`Lens Metrics has been removed from ${cluster.metadata.name}`, { timeout: 10_000 });
+  entitySettings = [
+    {
+      apiVersions: ["entity.k8slens.dev/v1alpha1"],
+      kind: "KubernetesCluster",
+      title: "Lens Metrics",
+      priority: 5,
+      components: {
+        View: (props: {entity: Catalog.KubernetesCluster}) => { // eslint-disable-line
+          const cluster = props.entity; // eslint-disable-line
```
Nokel81: If these disables are from what I think they are, I think we should just turn off that lint rule.

src/extensions/registries/entity-setting-registry.ts
```diff
@@ -11,13 +11,14 @@ export interface EntitySettingComponents {
...
-  kind: string;
   apiVersions: string[];
-  source?: string;
+  kind: string;
+  title: string;
   components: EntitySettingComponents;
+  source?: string;
   id?: string;
   priority?: number;
+  group?: string;
```
jim-docker, line 42: Is this exported in the extension api? (It shouldn't be, right?)

jim-docker (commented):
That's better :)

I still think it's a weird UX with the toggles (not exactly sure how I got into this state, apart from mashing the toggles):

https://user-images.githubusercontent.com/40840436/118046733-ba9f8b00-b347-11eb-8545-7358aa8db3f9.mov

https://user-images.githubusercontent.com/40840436/118046750-c25f2f80-b347-11eb-8994-b71d07b84e54.mov

It eventually settles into the state I last selected, though the toggles on the settings screen are sometimes incorrect, until I close and open the settings page (or maybe I'm not waiting long enough)

After toggling, can the toggles be disabled till the component enabling/disabling has completed?

jakolehm:
Now with dedicated button to "apply" changes:

https://user-images.githubusercontent.com/1446224/118502923-d386b280-b732-11eb-8dd1-084d8e38e81c.mov

src/extensions/registries/entity-setting-registry.ts
jakolehm: It is exported, why not?

extensions/metrics-cluster-feature/renderer.tsx
jakolehm: It feels bad, yes. Not sure what's the proper way to fix this.

Nokel81: The way I did it was to make it a free function:

```ts
interface EntitySettingViewProps {
  entity: Catalog.KubernetesCluster;
}

function Metrics({ entity }: EntitySettingViewProps) {
  return (
    <section>
      <section>
        <MetricsSettings cluster={entity} />
      </section>
    </section>
  );
}

export default class ClusterMetricsFeatureExtension extends LensRendererExtension {
  entitySettings = [
    {
      apiVersions: ["entity.k8slens.dev/v1alpha1"],
      kind: "KubernetesCluster",
      title: "Lens Metrics",
      priority: 5,
      components: {
        View: Metrics,
      },
    },
  ];
}
```

extensions/metrics-cluster-feature/src/metrics-settings.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+  get isActiveMetricsProvider() {
+    return (!this.metricsProvider || this.metricsProvider === "lens");
+  }
+
+  async componentDidMount() {
+    this.feature = new MetricsFeature(this.props.cluster);
+
+    await this.updateFeatureStates();
+  }
```
Nokel81, line 89: This `await` doesn't really do anything because `componentDidMount` is treated as if it returns `void`.

src/common/k8s/resource-stack.ts
```diff
@@ -0,0 +1,150 @@
...
+    kubectlArgs = this.appendKubectlArgs(kubectlArgs);
+
+    if (app) {
+      return await new ResourceApplier(clusterModel).kubectlApplyAll(resources, kubectlArgs);
+    } else {
+      const response = await requestMain(clusterKubectlApplyAllHandler, this.cluster.metadata.uid, resources, kubectlArgs);
+
+      if (response.stderr) {
+        throw new Error(response.stderr);
+      }
```
Nokel81, line 77: Why not just have the `handler` in `main` throw `stderr`? You wouldn't need this step and it would function the exact same.

extensions/metrics-cluster-feature/renderer.tsx
jakolehm: Should be now ok.

src/common/k8s/resource-stack.ts
```diff
@@ -0,0 +1,150 @@
...
+
+      const resourceArray = yaml.safeLoadAll(resourceData.toString());
+
+      resourceArray.forEach((resource) => {
+        if (resource?.metadata) {
+          resource.metadata.labels ||= {};
+          resource.metadata.labels["app.kubernetes.io/name"] = this.name;
+          resource.metadata.labels["app.kubernetes.io/managed-by"] = productName;
+          resource.metadata.labels["app.kubernetes.io/created-by"] = "resource-stack";
+        }
```
Nokel81, line 144: So we don't want to add these metadata labels if metadata isn't defined?

src/common/k8s/resource-stack.ts
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    return kubectlArgs;
+  }
+
+  protected renderTemplates(folderPath: string, templateContext: any): string[] {
+    const resources: string[] = [];
+
+    logger.info(`[RESOURCE-STACK]: render templates from ${folderPath}`);
+    fs.readdirSync(folderPath).forEach(filename => {
```
Nokel81: I don't think this should be sync.

src/main/resource-applier.ts
```diff
@@ -73,10 +73,19 @@ export class ResourceApplier {
...
-  public async kubectlApplyAll(resources: string[]): Promise<string> {
+  public async kubectlApplyAll(resources: string[], extraArgs = ["-o", "json"]): Promise<string> {
+    return this.kubectlCmdAll("apply", resources, extraArgs);
+  }
+
+  public async kubectlDeleteAll(resources: string[], extraArgs?: string[]): Promise<string> {
+    return this.kubectlCmdAll("delete", resources, extraArgs);
+  }
+
+  protected async kubectlCmdAll(subCmd: string, resources: string[], args?: string[]): Promise<string> {
```
Nokel81:
```suggestion
  protected async kubectlCmdAll(subCmd: string, resources: string[], args: string[] = []): Promise<string> {
```

src/main/resource-applier.ts
```diff
@@ -73,10 +73,19 @@ export class ResourceApplier {
...
     const { kubeCtl } = this.cluster;
     const kubectlPath = await kubeCtl.getPath();
     const proxyKubeconfigPath =  await this.cluster.getProxyKubeconfigPath();
+    let kubectlArgs = args || [];
```
Nokel81: no longer need these.

src/main/resource-applier.ts
```diff
@@ -85,19 +94,24 @@ export class ResourceApplier {
       resources.forEach((resource, index) => {
         fs.writeFileSync(path.join(tmpDir, `${index}.yaml`), resource);
       });
-      const cmd = `"${kubectlPath}" apply --kubeconfig "${proxyKubeconfigPath}" -o json -f "${tmpDir}"`;
+      kubectlArgs = kubectlArgs.concat(["-f", `"${tmpDir}"`]);
```
Nokel81:
```suggestion
      args.push("-f", `"${tmpDir}"`);
```

src/common/k8s/resource-stack.ts
jakolehm: Yes, it's just additional protection. All relevant kube resources should have metadata.

Nokel81: Okay. It just looked weird not to do something like:

```ts
if (resource) {
  resource.metadata ??= {};
  resource.metadata.labels ??= {};
  
  ...
}
```

src/common/k8s/resource-stack.ts
jakolehm: I just moved the code from `ClusterFeature` here... so it has been like this forever. I agree that async would be better.

jakolehm: Refactored to async.

src/main/resource-applier.ts
jakolehm: Done.

Nokel81 (dismissed):
This looks good, other internal improvements can be done at a future date.

Nokel81:
Actually I just found some issues, while trying it out on my local `minikube`:

1. If I deselect all options an then press "Uninstall", it should read "Uninstalling..."
2. It never stops the loading animation while uninstalling
3. I don't think we should be printing 404 errors to the console while viewing these settings.

src/main/resource-applier.ts
```diff
@@ -85,19 +93,24 @@ export class ResourceApplier {
       resources.forEach((resource, index) => {
```
jim-docker, line 93: I'm encountering `resources` not defined here

jim-docker (commented):
I'm seeing this now:


https://user-images.githubusercontent.com/40840436/118554665-bd2b2780-b72f-11eb-8396-f7916de6359f.mov

src/main/resource-applier.ts
jakolehm: I see the same now, maybe async file operations did break something (it worked before).

jakolehm: Fixed.

nevalla (approved):
LGTM

jakolehm:
@Nokel81 @jim-docker ptal

src/common/k8s/resource-stack.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  protected async renderTemplates(folderPath: string, templateContext: any): Promise<string[]> {
+    const resources: string[] = [];
+
+    logger.info(`[RESOURCE-STACK]: render templates from ${folderPath}`);
+    const files = await fse.readdir(folderPath);
+
+    for(const filename of files) {
+      const file = path.join(folderPath, filename);
+      const raw = await fse.readFile(file);
```
Nokel81, line 123: nit: since we always `.toString()` this, we might as well declare the expected encoding and get `readFile()` to do it for us.
```suggestion
      const raw = await fse.readFile(file, "utf-8");
```

src/common/k8s/resource-stack.ts
```diff
@@ -0,0 +1,152 @@
...
+
+      if (filename.endsWith(".hb")) {
+        const template = hb.compile(raw.toString());
+
+        resourceData = template(templateContext);
+      } else {
+        resourceData = raw.toString();
+      }
+
+      if (!resourceData.trim()) continue;
```
Nokel81, line 134: Which should allow us to simplify this:

```suggestion
      const resourceData = (
        filename.endsWith(".hb")
          ? hb.compile(raw)(templateContext)
          : raw
      ).trim();
      
      if (!resourceData) {
        continue;
      }
```

src/renderer/components/+entity-settings/entity-settings.tsx
```diff
@@ -57,28 +58,40 @@ export class EntitySettings extends React.Component<Props> {
   async componentDidMount() {
     const { hash } = navigation.location;
 
-    this.ensureActiveTab();
+    if (hash) {
+      const item = this.menuItems.find((item) => item.title === hash.slice(1));
+
+      if (item) {
```
Nokel81, line 64: Something that I have thought about from conversations with @leenamba. I think that the user should probably be warned (via a Notification or something) if an item is not found. 

That should help with spelling mistakes on `lens://` query parameters.

src/renderer/components/+entity-settings/entity-settings.tsx
```diff
@@ -57,28 +58,40 @@ export class EntitySettings extends React.Component<Props> {
   async componentDidMount() {
     const { hash } = navigation.location;
 
-    this.ensureActiveTab();
+    if (hash) {
+      const item = this.menuItems.find((item) => item.title === hash.slice(1));
```
Nokel81, line 62: I don't think we should be calling `.slice(1)` for every item in `this.menuItems`.

I also don't think that we should do this if `hash === "#"` which is very possible.

src/renderer/components/+entity-settings/entity-settings.tsx
```diff
@@ -57,28 +58,40 @@ export class EntitySettings extends React.Component<Props> {
...
-            <Tab
-              key={setting.id}
-              value={setting.id}
-              label={setting.title}
-              data-testid={`${setting.id}-tab`}
-            />
+          { groups.map((group) => (
+            <>
+              <div className="header">{group[0]}</div>
+              { group[1].map((setting, index) => (
```
Nokel81, line 86:
```suggestion
          { groups.map(([name, entries]) => (
            <>
              <div className="header">{name}</div>
              { entries.map((setting, index) => (
```

Nokel81 (approved):

jim-docker (approved):
minor UX thing, if I click a toggle then the Apply button goes from disabled to enabled. If I click the toggle again then the Apply button should go back to disabled (i.e. if the current state of the toggles matches the state of the installation then the Apply button should be disabled).

PR #2716, opened 2021-05-06
Simplify add cluster view

### LINKED ISSUE #2568: Improve Add Cluster UI
Switch the selection of context to a list of labelled toggles. In conjunction it should display contexts that are invalid with errors in line (and the toggle itself being disabled).

**Why is this needed**:

Currently, when selecting a kubeconfig the contexts are detected and put into a dropdown. However, this makes for less than ideal UX since it requires 2 addtional clicks (one to open the drop down and one to close it) whenever selecting context(s).

Nokel81:
- Only shows editor, and will save all text to file

- Reports some validation errors, but doesn't block the add cluster button

- Cleanup kube helpers to facilitate clearer FS interactions


src/common/kube-helpers.ts
```diff
@@ -71,67 +131,44 @@ export function loadFromOptions(options: KubeConfigOptions): KubeConfig {
   return kc;
 }
 
-export function loadConfig(pathOrContent?: string): KubeConfig {
-  return loadConfigFromString(
-    fse.pathExistsSync(pathOrContent)
-      ? readResolvedPathSync(pathOrContent)
-      : pathOrContent
-  );
+export interface ConfigResult {
```
Nachasic, line 155: Sounds like a good precedent to adopt monadic return types (e.g. https://github.com/hqoss/monads) in the future? This particular case looks like it would be a good use of `Result<KubeConfig, Joi.ValidationError>`

src/common/kube-helpers.ts
```diff
@@ -71,67 +131,44 @@ export function loadFromOptions(options: KubeConfigOptions): KubeConfig {
...
+  const { options, error } = loadToOptions(content);
 
-  if (!config.clusters || config.clusters.length == 0) {
-    throw new Error("No clusters provided in config");
-  }
-
-  if (!config.contexts || config.contexts.length == 0) {
-    throw new Error("No contexts provided in config");
-  }
+  return {
```
Nachasic, line 163: Monadic types (referring to my previous comment) could be mapped:
```ts
// assuming `loadToOptions()` returns `Result<Options, Error>`
return loadToOptions(content).map(options => loadFromOptions(options)) // Result<Options, Error>
```

src/main/catalog-sources/kubeconfig-sync.ts
```diff
@@ -127,7 +125,13 @@ type RootSource = ObservableMap<string, RootSourceValue>;
 export function computeDiff(contents: string, source: RootSource, filePath: string): void {
   runInAction(() => {
     try {
-      const rawModels = configToModels(loadConfigFromString(contents), filePath);
+      const { config, error } = loadConfigFromString(contents);
```
Nachasic, line 157: Again, with monadic types this could be simplified to 
```ts
const config = loadConfigFromString(contents);

if (config.isErr()) {
    logger.warn();
}
const rawModels = configToModels(config.unwrapOr(fallbackConfig), filePath);
```

Nachasic (dismissed):
Posted some comments for consideration, LGTM

src/common/kube-helpers.ts
Nokel81: While I really like monadic return types (especially coming from my liking of rust). I don't think this place would be good. Because it really should be (if I was in rust) `(KubeConfig, Option<Joi.ValidationError>)`, which isn't that more ergonomic than what I have now.

Nachasic: Makes sense.

aleksfront:
I have couple of question related comparing before/after views:

Before
current view

After
simplified view

1. Is it simplifies view or some interactions in it? Not much elements changed so it's still "simple" (maybe not simple for the user, however)
2. Should we remove description which is currently presented ("Paste kubeconfig as a text...")?
3. New message `Clusters added here are **not** merged...` doesn't seem right. *Not* word can confuse user. For example, first thoughts for me was "why new clusters should be merged?"

From the UI point of view I suggest whole look and UX should be refactored here. But we can do this in separate PR though.

Nokel81:
@aleksfront 
1. It simplifies it because there was initially two PRs (one for a fix like this and one to add the ability to add AccessibleNamespaces). But now there is just this one. I don't just simplify the code and the top text, but it also removes the "Proxy Settings"
2. I have removed that description, because we don't support that anymore.
3. I have that message because when you use `kubectl config set-context` (and siblings) it does modify the default kubeconfig file (unless you specify the `--config` option).

Nachasic (dismissed):

package.json
```diff
@@ -181,6 +181,7 @@
     "@hapi/call": "^8.0.0",
     "@hapi/subtext": "^7.0.3",
     "@kubernetes/client-node": "^0.12.0",
+    "@material-ui/icons": "^4.11.2",
```
jakolehm: Is this actually used?

src/common/utils/index.ts
```diff
@@ -21,7 +21,9 @@
 
 // Common utils (main OR renderer)
 
-export const noop: any = () => { /* empty */ };
+export function noop<T extends any[]>(...args: T): void {
```
jakolehm, line 24: Looks like unnecessary change? Is it?

src/renderer/bootstrap.tsx
```diff
@@ -44,11 +44,11 @@ import { HelmRepoManager } from "../main/helm/helm-repo-manager";
 import { ExtensionInstallationStateStore } from "./components/+extensions/extension-install.store";
 import { DefaultProps } from "./mui-base-theme";
 
-/**
```
jakolehm: Any reason for removing this?

package.json
Nokel81: Ah this is actually a duplicate of the item in `devDependencies`

src/common/utils/index.ts
Nokel81: It is a minor improvement in the typing. That is all. I forget why I did this in the original PRs.

src/renderer/bootstrap.tsx
Nokel81: Nope. Didn't remember that I did this.

Nachasic (dismissed):

Nachasic (approved):

PR #2718, opened 2021-05-06
Mobx-6 migration

### LINKED ISSUE #1536: Update MobX to 6.x

ixrock:
Attempt to upgrade from 5.x to 6.x. 

(re)based on https://github.com/lensapp/lens/pull/2520

**Main pitfalls:**
- when using `mobx` decorators (e.g `@observable`) class constructor must include `makeObservable(this)`
-  don't use`mobx.toJS()` with partial observable data, e.g. `toJS({x: 1, y: observable([])})` still returns `result.y` as `observable` => use `common/utils/toJS` to handle such cases.
- don't apply `@action` for overriding class methods or use extra `@override` from `mobx` package
- don't use `@observable static` for classes (there is no way to wrap it with `makeObservable`)
- second argument in `mobx.toJS` has removed, e.g. `toJS(data, {recursiveEverything: true})` will throw an error.
- don't cache `when(() => this.someObservableBoolean)` in class fields (might not work as you expected)

**Try it locally:**
- `git checkout mobx6-migration && yarn`
- `make clean && make dev`

**More info:**
https://mobx.js.org/migrating-from-4-or-5.html

Fixes #1536


docs/extensions/guides/stores.md
```diff
@@ -42,18 +42,17 @@ export class ExamplePreferencesStore extends Store.ExtensionStore<ExamplePrefere
...
     });
+    makeObservable(this);
   }
 
   protected fromStore({ enabled }: ExamplePreferencesModel): void {
     this.enabled = enabled;
   }
 
   toJSON(): ExamplePreferencesModel {
-    return toJS({
```
ixrock, line 52: Wrapping with `toJS()` probably not necessary since:

- properly handled in `BaseStore.enableSync()` for reacting to every observable from `BaseStore.toJSON()`
- sanitizing payload for IPC-messaging (removing all possible observables for `handleRequest`, `requestMain` and `broadcastMessage` in `ipc.tsx`)

src/common/base-store.ts
```diff
@@ -20,13 +19,18 @@ export interface BaseStoreParams<T = any> extends ConfOptions<T> {
...
 export abstract class BaseStore<T = any> extends Singleton {
   protected storeConfig?: Config<T>;
-  protected syncDisposers: Function[] = [];
+  protected syncDisposers: Disposer[] = [];
 
-  whenLoaded = when(() => this.isLoaded);
   @observable isLoaded = false;
 
+  get whenLoaded() {
+    return when(() => this.isLoaded);
```
ixrock, line 48: Seems like `when` evaluates callback to detect observables before `makeObservable(this)` is done in `constructor`.

src/common/base-store.ts
```diff
@@ -93,7 +97,11 @@ export abstract class BaseStore<T = any> extends Singleton {
 
   enableSync() {
     this.syncDisposers.push(
-      reaction(() => this.toJSON(), model => this.onModelChange(model), this.params.syncOptions),
+      reaction(
+        () => toJS(this.toJSON()), // unwrap possible observables and react to everything
```
ixrock, line 122: This probably should have better compatibility with existing extensions after migration.

src/extensions/extensions-store.ts
```diff
@@ -47,9 +49,7 @@ export class ExtensionsStore extends BaseStore<LensExtensionsStoreModel> {
 
   toJSON(): LensExtensionsStoreModel {
     return toJS({
-      extensions: this.state.toJSON(),
-    }, {
-      recurseEverything: true
+      extensions: Object.fromEntries(this.state),
```
ixrock, line 73: FYI: `Object.fromEntries` might accept `Map` or `ObservableMap`

Nokel81: Might accept? I think it does, since what it wants is an `Iterable<[string, T]>`

src/common/base-store.ts
Nokel81: Why do you think this? That getter shouldn't be called until after that `makeObservable(this)` executes.

src/main/cluster-manager.ts
```diff
@@ -137,9 +139,7 @@ export class ClusterManager extends Singleton {
 }
 
 export function catalogEntityFromCluster(cluster: Cluster) {
-  return new KubernetesCluster(toJS({
-    apiVersion: "entity.k8slens.dev/v1alpha1",
-    kind: "KubernetesCluster",
```
ixrock, line 174: Those are already defined as `public readonly` in `KubernetesCluster`

src/renderer/api/endpoints/cluster.api.ts
```diff
@@ -54,7 +54,7 @@ export class Cluster extends KubeObject {
   static kind = "Cluster";
   static apiBase = "/apis/cluster.k8s.io/v1alpha1/clusters";
 
-  spec: {
+  declare spec: {
```
ixrock: This is required with `"useDefineForClassFields": true` in `tsconfig.ts` when field already defined at parent's class.

src/renderer/components/+apps/apps.tsx
```diff
@@ -3,12 +3,10 @@ import { observer } from "mobx-react";
 import { TabLayout, TabLayoutRoute } from "../layout/tab-layout";
 import { HelmCharts, helmChartsRoute, helmChartsURL } from "../+apps-helm-charts";
 import { HelmReleases, releaseRoute, releaseURL } from "../+apps-releases";
-import { namespaceUrlParam } from "../+namespaces/namespace.store";
 
 @observer
 export class Apps extends React.Component {
   static get tabRoutes(): TabLayoutRoute[] {
-    const query = namespaceUrlParam.toObjectParam();
```
ixrock, line 32: Keeping selected namespaces in URL not necessary anymore since `lens-local-storage` persists state in external json file.

src/renderer/components/+config-maps/config-map-details.tsx
```diff
@@ -39,7 +44,10 @@ export class ConfigMapDetails extends React.Component<Props> {
 
     try {
       this.isSaving = true;
-      await configMapsStore.update(configMap, { ...configMap, data: this.data.toJSON() });
```
ixrock, line 63: N.B. In _mobx@6_ `ObservableMap.toJSON()` returns array of entries, instead of plain-object.

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -130,41 +100,50 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
   }
 
   @action
-  setContext(namespace: string | string[]) {
```
ixrock, line 154: There are some breaking changes with names of class methods. 
If this is critical we could add wrappers with old names and mark as `@deprecated`. Wdyt?

src/renderer/components/+apps/apps.tsx
Nokel81: Thanks for finding this. I was thinking that was the case too.

src/renderer/components/+namespaces/namespace.store.ts
Nokel81: I don't think you should bother. I have been working (slowly and around other things) to do an almost complete refactor of how we deal with namespaces. So I don't think we should bother trying to support the older (and frankly very complex) form.

src/renderer/components/cluster-manager/cluster-view.tsx
```diff
@@ -1,81 +1,70 @@
...
-      }, { fireImmediately: true}
-      ),
-      reaction(() => this.cluster?.ready, (ready) => {
-        const clusterView = lensViews.get(this.clusterId);
-
-        if (clusterView && clusterView.isLoaded && !ready) {
-          navigate(catalogURL());
-        }
-      })
-    ]);
```
ixrock: Not sure if this is needed anymore. Wdyt? cc @jakolehm

Nokel81: Doesn't that redirect you to the catalog if you disconnect from the currently viewed cluster?

src/renderer/utils/storageHelper.ts
```diff
@@ -1,14 +1,10 @@
...
-import type { CreateObservableOptions } from "mobx/lib/api/observable";
-import { action, comparer, observable, toJS, when, IObservableValue } from "mobx";
-import produce, { Draft, enableMapSet, setAutoFreeze } from "immer";
+import { action, comparer, makeObservable, observable, toJS, when, } from "mobx";
+import produce, { Draft } from "immer";
 import { isEqual, isFunction, isPlainObject } from "lodash";
 import logger from "../../main/logger";
 
-setAutoFreeze(false); // allow to merge observables
-enableMapSet(); // allow merging maps and sets
```
ixrock, line 31: Moved to `src/common/configure-packages.ts` + mobx configurations.

Nokel81:
I prefer opening devtools in a separate window. Maybe instead of always opening it in dev mode. We have a ENV var (maybe `LENS_DEV_TOOLS`) which has the value of `"right" | "bottom" | "undocked" | "detach"`. And only open it in development mode if one of those is set.

src/common/base-store.ts
ixrock: I was talking about how it was done before.

Nokel81:
First bug: if I close the window and reopen it using the tray icon. The catalog is empty.

src/common/base-store.ts
Nokel81: Oh okay makes sense.

Nokel81:
Second bug: Clicking on a node results in the following crash:
Screen Shot 2021-05-11 at 9 22 23 AM

Scratch that, clicking on any item in an `ItemListLayout` results in that crash.

src/renderer/components/cluster-manager/cluster-view.tsx
ixrock: I have to check this more.. I wasn't able to disconnect from cluster:
Screenshot 2021-05-11 at 16 24 55

Maybe because of this weirdness (went to catalog from left-bottom icon). It show stale state for some reasons?
Screenshot 2021-05-11 at 16 25 11

Nokel81: That is weird because that should be an option.

src/common/search-store.ts
```diff
@@ -107,12 +108,12 @@ export class SearchStore {
     return prev;
   }
 
-  @autobind()
+  @boundMethod
```
aleksfront, line 132: Since it will be located in many different places is it possible to shorten name to just `@bound`?

src/main/window-manager.ts
```diff
@@ -84,6 +85,7 @@ export class WindowManager extends Singleton {
           shell.openExternal(url);
         })
         .on("dom-ready", () => {
+          this.mainWindow.webContents.openDevTools({ mode: "right", activate: isDevelopment });
```
aleksfront: Very convenient!

aleksfront:
🐞  Found a lot of app crashes in different places. Most of them happened in `Workloads` section and when opening details of various resources. @ixrock Posted all details and screenshots in slack direct message. 🐞

src/common/search-store.ts
Nokel81: Is there a good reason to not just use the `field = () => { ... };`

ixrock: @Nokel81 for example for when overriding a method in sub-class + that way it keeps single method definition in prototype (less memory consumption in some cases)

ixrock: @aleksfront not sure if a short name is better (I hope you use auto-import / auto-complete in your IDE:)

ixrock:
> Second bug: Clicking on a node results in the following crash:
> Screen Shot 2021-05-11 at 9 22 23 AM
> 
> Scratch that, clicking on any item in an `ItemListLayout` results in that crash.

This is fixed as well as various app-crashes found by @aleksfront (all pages affected by kube-event rendering).
Surprisingly all these crashes where related to `useDefineForClassFields: true` in `tsconfig.ts` (some explanation in fixing commit above).

src/main/window-manager.ts
ixrock: Probably better to move this to separated PR.

Nokel81:
@ixrock Thanks, will take a look again. I think that is something that we want to do in the future. But it will require changing how we specify things on `KubeObject`. Good to have that as a follow up PR anyway.

Nokel81:
@ixrock with regards to your last commit. I think it would have been better to change `class KubeObject` to be `class KubeObject<Metadata extends IKubeObjectMetadata = IKubeObjectMetadata, Status>`

Nokel81:
Oh there is also `Spec` too

src/extensions/extension-loader.ts
```diff
@@ -332,13 +341,10 @@ export class ExtensionLoader extends Singleton {
...
-    return toJS(this.extensions, {
-      exportMapsAsObjects: false,
-      recurseEverything: true,
-    });
+    return toJS(this.extensions);
   }
 
-  broadcastExtensions(main = true) {
-    broadcastMessage(main ? ExtensionLoader.extensionsMainChannel : ExtensionLoader.extensionsRendererChannel, Array.from(this.toJSON()));
+  broadcastExtensions(channel = ExtensionLoader.extensionsMainChannel) {
```
Nokel81: This shouldn't be an optional parameter.

src/common/catalog/catalog-entity-registry.ts
```diff
@@ -19,27 +19,33 @@
...
 
   @action removeSource(id: string) {
     this.sources.delete(id);
   }
 
   @computed get items(): CatalogEntity[] {
-    return Array.from(iter.flatMap(this.sources.values(), source => source.get()));
+    return Array.from(this.sources.values())
+      .map(source => source.get())
+      .flat();
```
Nokel81: Please revert things like this. This just clutters up the diff making it hard to review.

src/common/catalog/catalog-entity-registry.ts
```diff
@@ -19,27 +19,33 @@
...
-  @action addObservableSource(id: string, source: IObservableArray<CatalogEntity>) {
-    this.sources.set(id, computed(() => source));
+  constructor() {
+    makeObservable(this);
   }
 
-  @action addComputedSource(id: string, source: IComputedValue<CatalogEntity[]>) {
-    this.sources.set(id, source);
+  @action
+  addObservableSource(id: string, source: IObservableArray<CatalogEntity> | IComputedValue<CatalogEntity[]>) {
```
Nokel81: Again, please revert changes like this. This is not at all related to changing to `mobx6`.

src/extensions/registries/page-registry.ts
```diff
@@ -34,21 +34,18 @@ export interface PageRegistration {
    * When not provided, first registered page without "id" would be used for page-menus without target.pageId for same extension
    */
   id?: string;
-  params?: PageParams<string | ExtensionPageParamInit>;
+  params?: PageParams<string | Omit<PageParamInit<any>, "name" | "prefix">>;
```
Nokel81, line 36: These changes are not related to `mobx6`. Please move them to a different PR.

src/main/catalog-sources/kubeconfig-sync.ts
```diff
@@ -68,7 +74,7 @@ export class KubeconfigSyncManager extends Singleton {
       this.startNewSync(filePath);
     }
 
-    this.syncListDisposer = UserStore.getInstance().syncKubeconfigEntries.observe(change => {
+    this.syncListDisposer = UserStore.getInstance().syncKubeconfigEntries.observe_(change => {
```
Nokel81: Why the `_`?

src/main/catalog-sources/kubeconfig-sync.ts
```diff
@@ -248,17 +255,25 @@ async function watchFileChanges(filePath: string): Promise<[IComputedValue<Catal
     depth: stat.isDirectory() ? 0 : 1, // DIRs works with 0 but files need 1 (bug: https://github.com/paulmillr/chokidar/issues/1095)
     disableGlobbing: true,
   });
-  const rootSource = new ExtendedObservableMap<string, ObservableMap<string, RootSourceValue>>(observable.map);
+  const rootSource = new ObservableMap<string, ObservableMap<string, RootSourceValue>>();
```
Nokel81: Please revert unless my class doesn't work for some reason. This change is extraneous.

src/renderer/api/endpoints/helm-releases.api.ts
```diff
@@ -155,24 +155,26 @@ export async function rollbackRelease(name: string, namespace: string, revision:
...
-@autobind()
+export interface HelmRelease {
+  appVersion: string;
+  name: string;
+  namespace: string;
+  chart: string;
+  status: string;
+  updated: string;
+  revision: string;
+}
```
Nokel81, line 166: This seems strange, I know that it is legal but what happens if you import "HelmRelease" somewhere else, does it mean both?

Should we change this to a different name then?

src/renderer/api/endpoints/ingress.api.ts
```diff
@@ -117,6 +113,17 @@ export class Ingress extends KubeObject {
       ingress: ILoadBalancerIngress[];
     };
   };
+}
+
+export class Ingress extends KubeObject {
```
Nokel81, line 118: What I meant is that this should be written as:

```typescript
export interface IngressSpec {
}

export interface IngressMetadata extends IKubeMetadata {
  tls: {
    secretName: string;
  }[];
  rules?: {
    host?: string;
    http: {
      paths: {
        path?: string;
        backend: IIngressBackend;
      }[];
    };
  }[];
  // extensions/v1beta1
  backend?: IExtensionsBackend;
  // networking.k8s.io/v1
  defaultBackend?: INetworkingBackend & {
    resource: {
      apiGroup: string;
      kind: string;
      name: string;
    }
  }
}

export interface IngressStatus {
  loadBalancer: {
    ingress: ILoadBalancerIngress[];
  };
}

export class Ingress extends KubeObject<IngressMetadata, IngressStatus, IngressSpec> {
   ...
}
```

Nokel81 (changes requested):

src/main/catalog-sources/kubeconfig-sync.ts
ixrock: There is no `.observe()` anymore in `mobx-6`

Nokel81: Yes it does: https://mobx.js.org/intercept-and-observe.html

Just is a free function instead.

src/common/catalog/catalog-entity-registry.ts
ixrock: I wasn't sure previous code was working with mobx-6. So, I'll prefer to keep it as is for now.
Moreover I think form is more readable than previous and with less extra apis/helpers involved).

Nokel81: I completely disagree, if you can show that it wasn't working then sure go ahead. But from all the docs of how JS's `Array.from()` works. The previous code was correct.

src/renderer/api/endpoints/helm-releases.api.ts
ixrock: Nope, don't we think we should. Class can be merged with interfaces with same name within a module.
I guess it could be considered it as class type extension, but without fields definition in case of `{useDefineForClassFields: true}`

Nokel81: What is the benefit of moving this out then? There is `import type { HelmRelease }` already in typescript (see #2777)

src/main/catalog-sources/kubeconfig-sync.ts
ixrock: Replaced.

src/extensions/extension-loader.ts
ixrock: Fixed

src/renderer/api/endpoints/helm-releases.api.ts
ixrock: Consider an example:

```ts
class Test {
  constructor(data: any){
    Object.assign(this, data);
  }
}

class Test2 extends Test {
  public notGood: any;
  
  constructor(data: any){
    super(data);
    console.log(this.notGood); // prints "undefined" when compiled with {useDefineForClassFields: true}
  }
}

console.log(
  new Test2({notGood: true}),
);
```

If we move definition from `Test2` to external interface within class module, then all good and console prints `true`:

```ts
interface Test2 {
 notGood: any;
}
```

src/common/catalog/catalog-entity-registry.ts
ixrock: Yeah, maybe your version is correct. But I prefer mine for the sake of simplicity.

src/renderer/api/endpoints/helm-releases.api.ts
Nokel81: Wouldn't just not using `Object.assign` also fix this problem?

src/common/catalog/catalog-entity-registry.ts
Nokel81: How is it more simple? Also a migration PR is not the place for this sort of thing.

src/main/catalog-sources/kubeconfig-sync.ts
ixrock: @Nokel81 what the purpose for that extendable map over normal? @jakolehm wdyt?

src/extensions/registries/page-registry.ts
ixrock: It's partially related because `PageParam` helper depends on `mobx-observable-history` and it was refactored to align with new version.

src/common/catalog/catalog-entity-registry.ts
ixrock: It simple because:
- uses only native JS apis
- don't use generators/iterators (less complexity imho)

Let's ask from @lensapp/lens-maintainers which form they think more simple.
Anyway, I can revert these just to avoid tedious arguing with you.

ixrock: Reverted.

src/main/catalog-sources/kubeconfig-sync.ts
ixrock: Reverted

src/renderer/api/endpoints/helm-releases.api.ts
ixrock: Don't use `Object.assign` at all and remember this choice for every inheriting class? Sounds weird.

src/renderer/api/endpoints/ingress.api.ts
ixrock: You can do that in follow-up PRs if you like.

extensions/kube-object-event-status/package.json
```diff
@@ -18,6 +18,9 @@
   "dependencies": {},
   "devDependencies": {
     "@k8slens/extensions": "file:../../src/extensions/npm/extensions",
+    "react": "latest",
```
jakolehm: We can probably remove these.

src/common/catalog/catalog-entity.ts
```diff
@@ -56,8 +56,13 @@ export abstract class CatalogCategory extends EventEmitter {
...
   abstract spec: CatalogCategorySpec;
 
+  // Should be URL-compatible, otherwise "Uncaught TypeError: Expected "categoryId" to match "[^\/#\?]+?""
+  static getId(group: string, kind: string): string {
+    return `${group}--${kind}`;
+  }
+
   public getId(): string {
-    return `${this.spec.group}/${this.spec.names.kind}`;
+    return `${this.spec.group}--${this.spec.names.kind}`;
```
Nokel81: Why does this need to be URL compatible?

src/renderer/components/+catalog/catalog.route.ts
```diff
@@ -23,11 +23,10 @@ import type { RouteProps } from "react-router";
 import { buildURL } from "../../../common/utils/buildUrl";
 
 export interface ICatalogViewRouteParam {
-  group?: string;
-  kind?: string;
+  categoryId?: string;
 }
 export const catalogRoute: RouteProps = {
-  path: "/catalog/:group?/:kind?"
+  path: "/catalog/:categoryId?" // `categoryId` == `${group}--${kind}`
```
Nokel81: What is wrong with the old way?

src/renderer/api/endpoints/helm-releases.api.ts
Nokel81: Exactly, `Object.assign` is weird and is contrary to how Typescript would like you do it. It prevents detecting things like guaranteed assignment in constructors.

src/common/catalog/catalog-entity.ts
ixrock: To navigate to particular category with `catalogURL()`, see: `catalog.route.ts`

ixrock: For example `/` cannot be used, otherwise see exception from the comment above.

extensions/kube-object-event-status/package.json
```diff
@@ -18,6 +18,9 @@
...
+    "mobx": "latest",
+    "mobx-react": "latest",
```
Nokel81: How were these being supported before?

extensions/kube-object-event-status/tsconfig.json
```diff
@@ -13,6 +13,7 @@
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
     "experimentalDecorators": true,
+    "useDefineForClassFields": true,
```
Nokel81, line 16: I thought you were getting rid of this change.

src/common/catalog/catalog-entity.ts
Nokel81: But it was working before in Alex's PR.

Nokel81 (dismissed):
This PR is pretty big by itself without all the changes related to turning on `useDefineForClassFields`. 

I think that should be split out into its own PR since it is not part of `mobx-6` and we can discuss how to go about it there.

src/renderer/components/+catalog/catalog.route.ts
ixrock: Yeah, probably having splitted url-params for the route is better. Fixed.

extensions/kube-object-event-status/package.json
ixrock: Some leftovers. Removed.

extensions/kube-object-event-status/tsconfig.json
ixrock: Nope. It was returned, see commits history. 
This is actually a part of official mobx-6 migration guide.

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -86,17 +88,6 @@ export class Catalog extends React.Component<Props> {
       reaction(() => this.entities, items => this.catalogEntityStore.loadItems(items), {
         fireImmediately: true,
       }),
-
-      // select initial category if nothing yet selected (via url-params)
-      when(() => this.categories.length > 0, () => {
```
ixrock: Default category for the catalog is `Browse` if no additional params provided via URL. Wdyt? Good/bad/ugly? ©

src/common/catalog/catalog-entity.ts
```diff
@@ -56,13 +56,14 @@ export abstract class CatalogCategory extends EventEmitter {
...
   abstract spec: CatalogCategorySpec;
 
-  // Should be URL-compatible, otherwise "Uncaught TypeError: Expected "categoryId" to match "[^\/#\?]+?""
-  static getId(group: string, kind: string): string {
-    return `${group}--${kind}`;
+  static parseId(id = ""): { group?: string, kind?: string } {
+    const [group, kind] = id.split("/") ?? [];
+
+    return { group, kind };
   }
```
Nokel81: This could be instead:

```ts
public getRouteParams(): ICatalogViewRouteParam {
  return {
	group: this.spec.group,
    kind: this.spec.names.kind,
  }
}
```

src/renderer/components/+catalog/catalog.tsx
Nokel81: I think that is a good idea.

extensions/kube-object-event-status/tsconfig.json
Nokel81: Then I think we should do that in a separate PR before this one gets merged. It is a pretty big change in and of itself.

src/common/catalog/catalog-entity.ts
ixrock: It's a`static` method. It's done to match with selected entity: 
See `catalog.tsx`:

```tsx
  onTabChange = (categoryId: string) => {
    const { group, kind } = CatalogCategory.parseId(categoryId);

    if (group && kind) {
      navigate(catalogURL({ params: { group, kind } }));
    } else {
      navigate(catalogURL());
    }
  };
```

Nokel81: It could also be:

```ts
  onTabChange = (categoryId: string) => {
    const activeCategory = this.categories.find(category => category.getId() === tabId);

    navigate(catalogURL({ params: activeCategory?.getRouteParams() }));
  };
```

stefcameron:
@ixrock Since you point out that "second argument in mobx.toJS has removed, e.g. toJS(data, {recursiveEverything: true}) will throw an error.", and since this is so far part of [how extensions are to create their own stores](https://docs.k8slens.dev/v4.2.4/extensions/guides/stores/), it means that if Lens upgrades to mobx 6, a lot of third-party extension will crash, if they have followed the instructions on that docs page... (which is the case with the [CC extension](https://github.com/Mirantis/lens-extension-cc/blob/master/src/cc/store/PreferencesStore.js#L144)).

Will this PR be marked as a breaking change in whatever release it gets included in?

Or will extensions somehow be isolated from this change (though you did mention that all extensions should be disabled when trying this PR locally, lest these extensions cause crashes...)?

stefcameron:
Are the [Stores docs](https://docs.k8slens.dev/v4.2.4/extensions/guides/stores/) being updated as a result of this PR, to at least remove the second argument to that `mobx.toJS()` example? Not sure if your code generator also does this? FYI @chenhunghan

chenhunghan:
> Are the [Stores docs](https://docs.k8slens.dev/v4.2.4/extensions/guides/stores/) being updated as a result of this PR, to at least remove the second argument to that `mobx.toJS()` example? Not sure if your code generator also does this? FYI @chenhunghan

Thanks, the generator doesn't have store examples.

jakolehm:
> @ixrock Since you point out that "second argument in mobx.toJS has removed, e.g. toJS(data, {recursiveEverything: true}) will throw an error.", and since this is so far part of how extensions are to create their own stores, it means that if Lens upgrades to mobx 6, a lot of third-party extension will crash, if they have followed the instructions on that docs page... (which is the case with the CC extension).

Mobx6 is a breaking change in many ways, that's why we are trying to push it for Lens 5.0 (major release, it's possible to break things in a big way).

> Will this PR be marked as a breaking change in whatever release it gets included in?

Yes.

> Or will extensions somehow be isolated from this change (though you did mention that all extensions should be disabled when trying this PR locally, lest these extensions cause crashes...)?

No. It's not possible to isolate extension from this because extensions are sharing the same mobx runtime.

stefcameron:
Are the [Stores docs](https://docs.k8slens.dev/v4.2.4/extensions/guides/stores/) being updated as a result of this PR, to at least remove the second argument to that `mobx.toJS()` example?

docs/extensions/guides/stores.md
```diff
@@ -42,18 +42,17 @@ export class ExamplePreferencesStore extends Store.ExtensionStore<ExamplePrefere
...
+    return {
       enabled: this.enabled
-    }, {
-      recurseEverything: true
```
ixrock, line 55: @stefcameron it's removed from current docs. Is that what you asked? ^

stefcameron: If by "current" you mean the docs associated with Lens 5 changes, then yes. If by "current" you mean Lens 4.2.4 docs, then it's still there (and should be, I guess, since 4.2.4 still users `mobx` < 6).

ixrock: By "current" I mean a version mentioned in project's `package.json` or above whenever it's gonna be released.
Otherwise in current branch all mentioning about `recurseEverything:true` was removed.

stefcameron: Ok, thank you! 👍

src/renderer/api/endpoints/helm-releases.api.ts
jakolehm: Btw why did we enable `useDefineForClassFields` in this PR? Does Mobx6 require it?

extensions/kube-object-event-status/tsconfig.json
aleksfront: I guess we should consider to keep this into current PR. The reasons:
1. It's part of migration guide for TS users https://mobx.js.org/migrating-from-4-or-5.html. And since we want to use decorators, we should include this.
2. It's impossible to satisfy TS compiler in current PR without `useDefineForClassFields`, so after merge `master` would be broken?

aleksfront:
@ixrock Please check `catalog.tsx` and corresponding items in actions menu. We don't have right menu items in there.
Maybe revert refactoring in here if they're not related to migration?

In `mobx6-migration` branch:
broken menu items

In `master`:
actions menu items

extensions/kube-object-event-status/tsconfig.json
ixrock: @aleksfront what do you mean by after merge `master` would be broken? 
`useDefineForClassFields: true` required for class fields definitions with decorators without initial value, e.g.

```ts
class Test {
 @observable someVar: string; // no initial value, doesn't work with mobx-6
}
```

ixrock:
> @ixrock Please check `catalog.tsx` and corresponding items in actions menu. We don't have right menu items in there.
> Maybe revert refactoring in here if they're not related to migration?
> In `master`:
> actions menu items

Good catch! Fixed (reverted `catalog.tsx`).

aleksfront (approved):

ixrock:
@lensapp/lens-maintainers PTAL again [second approve required]
@jakolehm would be nice to have some final review from you asap & finally get this merged before next `master`-conflict? Thanks! :)

jakolehm:
> @lensapp/lens-maintainers PTAL again [second approve required]
> @jakolehm would be nice to have some final review from you asap & finally get this merged before next `master`-conflict? Thanks! :)

I have read through 50% of changed files 😅

jakolehm (approved):
Awesome job 🏆

ixrock:
> Btw why did we enable useDefineForClassFields in this PR? Does Mobx6 require it?

@jakolehm  Yes, it's mentioned in [mobx-6 migration](https://mobx.js.org/migrating-from-4-or-5.html) guide. 

I figured out that it is required in case when class field defined without initial value, e.g `@observable data: any`
In this case `makeObservable(this)`  in constructor seems like doesn't work at all and the field remains un-observable (`useDefineForClassFields: true` creates initial `undefined` values for them and `makeObservable` makes those as observables).
Only possible "safe" option to keep `useDefineForClassFields: false` (if we decide) and `mobx-6` would be avoiding decorators at all.

extensions/kube-object-event-status/package.json
ixrock: Yes. I'll create quick follow-up once this merged (since 2 approves already done)

ixrock:
@Nokel81 please dismiss your requested changes since 2 approves already done (there is no need to postpone this branch). 
See my explanation about `useDefineForClassFields: true` from above.

Nokel81:
@ixrock 

> I figured out that it is required in case when class field defined without initial value, e.g @observable data: any
In this case makeObservable(this) in constructor seems like doesn't work at all and the field remains un-observable (useDefineForClassFields: true creates initial undefined values for them and makeObservable makes those as observables).
Only possible "safe" option to keep useDefineForClassFields: false (if we decide) and mobx-6 would be avoiding decorators at all.

Given that to turn on `strict: true` those fields would have to be definitely initialized (either in a constructor or in the field declaration itself) I would lean towards giving default values anyway.

Nokel81:
But I can dismiss my review since that work can still get done before 5.0.0 is released. Along with the better `KubeObject` typings.

PR #2721, opened 2021-05-07
Fix auto-update exiting too fast

### LINKED ISSUE #2715: beta update notification quits and does nothing
after I click "Yes"
image

lens just quits and then when I re-open the beta it prompts this again.

jakolehm:
Fixes #2715


msa0311 (approved):

src/main/app-updater.ts
```diff
@@ -17,7 +17,6 @@ function handleAutoUpdateBackChannel(event: Electron.IpcMainEvent, ...[arg]: Upd
     if (arg.now) {
       logger.info(`${AutoUpdateLogPrefix}: User chose to update now`);
       autoUpdater.quitAndInstall(true, true);
-      app.exit(); // this is needed for the installer not to fail on windows.
```
nevalla, line 20: Is the whole app exiting without this. I mean that the tray service also is closed?

nevalla (approved):
Tested on Mac and Windows and update process worked as expected.

PR #2728, opened 2021-05-07
remove removedClusters from ClusterStore

Nokel81:

This is a preamble to doing this sort of thing at the sources instead.


nevalla:
The change itself looks good. Just thinking is catalog already handling these? Like if cluster is removed from the catalog, does it stop auth proxies etc?

Nokel81:
It will. Maybe I should close this and do that all together. 

I think that is what I'll do

PR #2730, opened 2021-05-07
Refactor <Input> to be clearer. Add namespace name validator

### LINKED ISSUE #2727: Feature request: Clean input for accesible namespaces
Hi there!

This morning added a new cluster to Lens, but it wasn't showing anything.
I copied the namespace-names from Rancher into lens, but turns out, I accidentally copied it with a space in front: `<space>namespace-name`
Would be nice if characters that can't be in a namespace-name anyway, would be cleaned on input.

Though my colleagues and I are very happy with Lens, thanks!

Kind regards

Nokel81:
- Changes support of async validators to only run on submit
- Debounces validation to only run 500ms after the last edit.

fixes #2727 

Videos:

https://user-images.githubusercontent.com/8225332/117486490-b7fff880-af37-11eb-802c-a3f4c30d7257.mov

Note: that validator at the end is purely for the demo and is not in this PR.

### DIFF
--- a/src/renderer/components/input/input.tsx
+++ b/src/renderer/components/input/input.tsx
@@ -1,15 +1,15 @@
 import "./input.scss";
 
 import React, { DOMAttributes, InputHTMLAttributes, TextareaHTMLAttributes } from "react";
-import { autobind, cssNames, debouncePromise, getRandId } from "../../utils";
+import { autobind, cssNames, getRandId } from "../../utils";
 import { Icon } from "../icon";
 import { Tooltip, TooltipProps } from "../tooltip";
 import * as Validators from "./input_validators";
 import { InputValidator } from "./input_validators";
 import isString from "lodash/isString";
-import isFunction from "lodash/isFunction";
-import isBoolean from "lodash/isBoolean";
-import uniqueId from "lodash/uniqueId";
+import { action, computed, observable } from "mobx";
+import { debounce } from "lodash";
+import { observer } from "mobx-react";
 
 const { conditionalValidators, ...InputValidators } = Validators;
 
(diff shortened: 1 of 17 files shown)


src/renderer/components/input/input.tsx
```diff
@@ -106,125 +117,105 @@ export class Input extends React.Component<InputProps, State> {
...
-      // debounce async validators
-      .map(({ debounce, ...validator }) => {
-        if (debounce) validator.validate = debouncePromise(validator.validate, debounce);
-
-        return validator;
-      });
-    // run validation
-    this.validate();
-  }
+  runValidators = debounce(() => this.runValidatorsRaw(this.getValue()), 500, {
```
jim-docker, line 183: This is tricky. Debouncing helps when entering a url, you want to wait till the user is done typing to validate the input. But in the case of an invalid character I would prefer to be told I entered an invalid character immediately. Since you added the `asyncValidator`, which is run only before submitting (and could be used to check for a well-formed url, etc), then you don't really need the debouncing?

jim-docker (commented):
Doesn't seem quite right to leave the error message when the input field is returned to its initial, empty state (for which there was, correctly, no error message). Though I can see why you might leave it there.
 
https://user-images.githubusercontent.com/40840436/117495316-75441d80-af43-11eb-8b1d-de134d1fa51f.mov

src/renderer/components/input/input.tsx
Nokel81: I think some form of denouncing is a good idea. That is because what is considered an "invalid character" depends on the final version.

For instance, a namespace must be a DNS label. Which cannot end in a `-`. I as a user would find it rather annoying to be told that while I am typing unless I pause. Because if I am typing slowly (but within that denounce window) I shouldn't see an error between when I type the `-` and when I type an `a` for instance.

Nokel81:
> Doesn't seem quite right to leave the error message when the input field is returned to its initial, empty state (for which there was, correctly, no error message). Though I can see why you might leave it there.

I think you are right, "required" should be special cased for `onSubmit` only. And maybe a red `*` before the placeholder.

src/renderer/components/input/input.tsx
jim-docker: Still sounds to me like, for such a type, you should have 2 validators, one to check for invalid characters in real-time while the user is typing (no debouncing), and the other on-submit to check for well-formed input, such as no trailing `-`

Nokel81: I guess we could do that too. I was trying to keep it as consistent with the old design, while also acknowledging that `async` validators (currently) don't generally support aborting.

Nokel81:
@aleksfront I have found this this change does not work well with `<Wizard>` (which is why the integration tests are failing). 

Is there a reason (besides just we haven't) as to why we are not using https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/HTML5/Constraint_validation?

I think I will create a new PR with just the fix and have this PR open until we decide the way forward. I think it would be better if we standardized around how we wish to do this. I think moving to using more `<form>`'s would be a good idea. WDYT?

Nokel81:
Closing in favour of #2754. I think we should look into this after 5.0

PR #2743, opened 2021-05-10
Remove `isClusterScoped` from `ItemListLayout`

Nokel81:

Part of my work reworking the `NamespaceStore`

There are no longer UI changes with this PR.

### DIFF
--- a/src/renderer/components/+custom-resources/crd-details.tsx
+++ b/src/renderer/components/+custom-resources/crd-details.tsx
@@ -145,7 +145,7 @@ export class CRDDetails extends React.Component<Props> {
         <>
           <DrawerTitle title="Validation"/>
           <AceEditor
-            mode="json"
+            mode="yaml"
             className="validation"
             value={validation}
             readOnly
--- a/src/renderer/components/item-object-list/item-list-layout.tsx
+++ b/src/renderer/components/item-object-list/item-list-layout.tsx
@@ -24,7 +24,7 @@ import groupBy from "lodash/groupBy";
 
 import React, { ReactNode } from "react";
 import { computed, makeObservable } from "mobx";
-import { disposeOnUnmount, observer } from "mobx-react";
+import { observer } from "mobx-react";
 import { ConfirmDialog, ConfirmDialogParams } from "../confirm-dialog";
 import { Table, TableCell, TableCellProps, TableHead, TableProps, TableRow, TableRowProps, TableSortCallback } from "../table";
 import { boundMethod, createStorage, cssNames, IClassName, isReactNode, noop, ObservableToggleSet, prevDefault, stopPropagation } from "../../utils";
@@ -43,6 +43,7 @@ import { MenuItem } from "../menu";
 import { Checkbox } from "../checkbox";
 import { UserStore } from "../../../common/user-store";
 import { namespaceStore } from "../+namespaces/namespace.store";
+import { KubeObjectStore } from "../../kube-object.store";
 
 // todo: refactor, split to small re-usable components
 
@@ -63,7 +64,6 @@ export interface ItemListLayoutProps<T extends ItemObject = ItemObject> {
   store: ItemStore<T>;
   dependentStores?: ItemStore[];
   preloadStores?: boolean;
-  isClusterScoped?: boolean;
   hideFilters?: boolean;
   searchFilters?: SearchFilter<T>[];
   /** @deprecated */
@@ -137,7 +137,7 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
   }
 
   async componentDidMount() {
-    const { isClusterScoped, isConfigurable, tableId, preloadStores } = this.props;
+    const { isConfigurable, tableId, preloadStores } = this.props;
 
     if (isConfigurable && !tableId) {
       throw new Error("[ItemListLayout]: configurable list require props.tableId to be specified");
@@ -149,12 +149,6 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
 
     if (preloadStores) {
       this.loadStores();
-
-      if (!isClusterScoped) {
-        disposeOnUnmount(this, [
-          namespaceStore.onContextChange(() => this.loadStores())
-        ]);
-      }
     }
   }
 
@@ -162,7 +156,6 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
     const { store, dependentStores } = this.props;
     const stores = Array.from(new Set([store, ...dependentStores]));
 
-    // load context namespaces by default (see also: `<NamespaceSelectFilter/>`)
     stores.forEach(store => store.loadAll(namespaceStore.contextNamespaces));
   }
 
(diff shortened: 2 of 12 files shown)


src/renderer/components/+custom-resources/crd-details.tsx
```diff
@@ -124,7 +124,7 @@ export class CRDDetails extends React.Component<Props> {
         <>
           <DrawerTitle title="Validation"/>
           <AceEditor
-            mode="json"
+            mode="yaml"
```
Nokel81, line 148: This caused a crash because we are not loading any `json` mode for the `AceEditor`. This was the only place that `yaml` wasn't used and `yaml` is a superclass of `json` so this should be fine (it looked fine to me).

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -123,12 +121,6 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
 
     if (preloadStores) {
       this.loadStores();
-
-      if (!isClusterScoped) {
-        disposeOnUnmount(this, [
-          namespaceStore.onContextChange(() => this.loadStores())
-        ]);
-      }
```
Nokel81, line 157: This is no longer needed `NamespaceStore` as loading stores is already responsive to this.

src/renderer/components/layout/main-layout-header.tsx
```diff
@@ -12,6 +13,7 @@ export const MainLayoutHeader = observer(({ cluster, className }: Props) => {
   return (
     <header className={cssNames("flex gaps align-center justify-space-between", className)}>
       <span className="cluster">{cluster.name}</span>
+      <NamespaceSelectFilter />
```
Nokel81: I moved this up here because it shows better (imo) that this component is used by the whole app and is not per view.

src/renderer/components/+custom-resources/crd-details.tsx
Nachasic: Should we make a separate issue to investigate this further?

Nachasic (dismissed):
LGTM

src/renderer/components/+custom-resources/crd-details.tsx
Nokel81: I don't think so, I am just surprised that we haven't seen this crash before. The thing is. Once you clicked away from the "X" the app would work fine anyway. So maybe that is what people did.

ixrock: Add `require("ace-builds/src-noconflict/mode-json");` in constructor in `ace-editor.tsx` to fix this.

Nokel81: Sure, but crd details probably could be `yaml` too. So I think this is the better fix.

Nachasic (dismissed):

Nachasic (dismissed):

src/renderer/components/layout/main-layout-header.tsx
jim-docker: I like this a lot but it has to get past the UX police :) . Maybe it could be disabled on pages where it clearly doesn't apply (like Namespaces). And the Nodes page, where if you view the node details page it currently lists all of the pods from all namespaces on the node. Although that could still cause performance issues for a cluster with a large number of namespaces and pods? May be better to be able to filter by namespaces even on the node details page? And it would be more obvious with the namespaces filter here.

jim-docker (dismissed):

Nokel81:
@jakolehm care to comment?

src/renderer/components/layout/main-layout-header.tsx
jakolehm: Unfortunately this is in conflict with the new (upcoming) TopBar logic. I'll block merging until we figure out how to proceed with this.

jakolehm (changes requested):

Nokel81:
@jakolehm Shall we discuss that at a meeting on Monday?

jakolehm (dismissed):

jim-docker (commented):
Compared to before the Nodes page now has the namespace selector filter,
the Role Bindings and the Roles pages no longer have the namespace select filter.

Nokel81:
> Compared to before the Nodes page now has the namespace selector filter,
the Role Bindings and the Roles pages no longer have the namespace select filter.

Nodes page has it because `podsStore` is listed as a `"dependentStore"`. The `Roles` and `Role Bindings` pages don't because their stores list the `clusterRolesApi` and `clusterRoleBindingsApi` as their `api`'s. This will be fixed by #2755. 

I will make a follow up PR to remove `podsStore` from `nodes.tsx` since the details page does `podsStore.reloadAll()` in `componentDidMount`

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -198,6 +191,21 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
     },
   };
 
+  @computed get isNamespaced(): boolean {
+    const { store, dependentStores = [] } = this.props;
```
jakolehm: I think we should not check `dependentStores` here (noticed this when you opened #2950).

Nokel81: Okay will change this then and close #2950. I will leave the other bug to be fixed by #2755.

src/renderer/components/kube-object/kube-object-list-layout.tsx
```diff
@@ -76,6 +77,15 @@ export class KubeObjectListLayout extends React.Component<KubeObjectListLayoutPr
         items={items}
         preloadStores={false} // loading handled in kubeWatchApi.subscribeStores()
         detailsItem={this.selectedItem}
+        customizeHeader={({ filters, ...placeholders }) => ({
+          filters: (
+            <>
+              {store.api.isNamespaced && <NamespaceSelectFilter />}
```
jim-docker: at least it is more appropriate here than in ItemListLayout

jim-docker (commented):
The Groups filter is missing now for Custom Resources -> Definitions?

Screen Shot 2021-06-03 at 3 36 09 PM

src/renderer/components/kube-object/kube-object-list-layout.tsx
Nokel81: I agree but moving it here breaks other things and fixing those would probably be out of scope. So I am going to revert to a simpler fix, and then make a new PR to improve the customization of the header.

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -397,20 +390,26 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
-      filters: <>
-        {!isClusterScoped && <NamespaceSelectFilter/>}
-        <PageFiltersSelect allowEmpty disableFilters={{
-          [FilterType.NAMESPACE]: true, // namespace-select used instead
-        }}/>
-      </>,
-      search: <SearchInputUrl/>,
+      filters: (
+        <>
+          {showNamespaceSelectFilter && <NamespaceSelectFilter />}
```
jakolehm, line 406: We might want to get rid of the default (namespace) filter in future. ItemListLayout should be "item agnostic" and Kube object list can set this easily.

jakolehm (approved):
LGTM

nevalla (approved):

src/renderer/components/item-object-list/item-list-layout.tsx
Nokel81: That is what the follow up PRs are doing.

PR #2751, opened 2021-05-11
Compute the gradient color from the DOM for NamespaceSelect

Nokel81:
- Once on initial load, and the subsequently only when theme changes
  have applied

Before: (This is on the beta for v5)
image

### DIFF
--- a/src/renderer/utils/color.ts
+++ b/src/renderer/utils/color.ts
@@ -1,0 +1,46 @@
+import Color from "color";
+
+/**
+ * Some notes:
+ * - CSS colours use straight alpha channels:
+ *    - https://en.wikipedia.org/wiki/Alpha_compositing
+ *    - https://stackoverflow.com/a/45526785/5615967
+ * - The `getComputedStyles` function does not do this already
+ */
+
+/**
+ * Compute the "actual" color at an element by walking up the tree while
+ * @param elem The root element to up the DOM from
+ * @param field Which computedStyle field to work against
+ */
+export function computeStackingColor(elem: HTMLElement | undefined, field: "color" | "backgroundColor"): Color {
+  if (!elem) {
+    return Color.rgb(0, 0, 0).alpha(0);
+  }
+
+  const curColor = Color(window.getComputedStyle(elem)[field]);
+
+  if (curColor.alpha() === 1) {
+    return curColor;
+  }
+
+  return blend(computeStackingColor(elem.parentElement, field), curColor);
+}
+
+/**
+ * Blends the two colors where the parent color is place beneath the child color.
+ * And the returned color is what the browser would render.
+ *
+ * The parent color is placed beneath because child elements are rendered "above"
+ * their DOM parents.
+ * @param parent The color that is from the parent element
+ * @param child The color that is from the child element
+ */
+export function blend(parent: Color, child: Color): Color {
+  const alpha = (1 - child.alpha()) * parent.alpha() + child.alpha();
+  const red = ((1 - child.alpha()) * parent.alpha() * parent.red() + child.alpha() * child.red()) / alpha;
+  const green = ((1 - child.alpha()) * parent.alpha() * parent.green() + child.alpha() * child.green()) / alpha;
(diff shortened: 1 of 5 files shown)


src/renderer/utils/color.ts
```diff
@@ -0,0 +1,50 @@
...
+ * @param field Which computedStyle field to work against
+ */
+export function computeStackingColor(elem: HTMLElement | undefined, field: "color" | "backgroundColor"): Color {
+  if (!elem) {
+    return Color.rgb(0, 0, 0).alpha(0);
+  }
+
+  const curColor = Color(window.getComputedStyle(elem)[field]);
+
+  // console.log(elem);
```
Nachasic: These probably don't need to be here anymore.

src/renderer/utils/color.ts
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Blends the two colors where the parent color is place beneath the child color.
+ * And the returned color is what the browser would render.
+ *
+ * The parent color is placed beneath because child elements are rendered "above"
+ * their DOM parents.
+ * @param parent The color that is from the parent element
+ * @param child The color that is from the child element
+ */
+export function blend(parent: Color, child: Color): Color {
```
Nachasic, line 43: Can you provide a comment on what blending mode is implemented here? It's a bit hard to parse the math on it's own.

Nachasic (commented):
This looks like a pretty complicated problem to nail down. At the same time, when I inspect the `div.GradientValueContainer` (both under light and dark themes), all I can see in gradient styles is:
```css
element.style {
    right: 0px;
    background: linear-gradient(to left, rgb(255, 255, 255) 0px, transparent);
}
```
which ultimately amounts to no visible gradient at all, since the background color of parent element is white.

aleksfront (changes requested):
We shouldn't overcomplicate things here. I've already had fix, but then it was reverted (since it was part of broken PR).

CSS-only decision must be used for this.

Please use styling provided here
https://github.com/lensapp/lens/pull/2365/commits/4378a58b81be5f1bb1e2312ccc1cdac0cf2da2b5

Nokel81:
@aleksfront should I just use your diff? The problem I see with your fix is that it doesn't work for use in the Wizard.

aleksfront:
> @aleksfront should I just use your diff? The problem I see with your fix is that it doesn't work for use in the Wizard.

I will prepare a separate PR with those fixes asap.

Nokel81:
> This looks like a pretty complicated problem to nail down. At the same time, when I inspect the `div.GradientValueContainer` (both under light and dark themes), all I can see in gradient styles is:
> 
> ```css
> element.style {
>     right: 0px;
>     background: linear-gradient(to left, rgb(255, 255, 255) 0px, transparent);
> }
> ```
> 
> which ultimately amounts to no visible gradient at all, since the background color of parent element is white.

That would be it when used in a `<Wizard>`. But for instance on "Pods" view I see the following:
Screen Shot 2021-05-12 at 8 32 03 AM

aleksfront:
Alternatie fix https://github.com/lensapp/lens/pull/2761

Nachasic:
This is a complicated piece of logic, but it has uses elsewhere and is documented very well, so LGTM

Nachasic (approved):

src/renderer/utils/color.ts
```diff
@@ -0,0 +1,46 @@
...
+  const alpha = (1 - child.alpha()) * parent.alpha() + child.alpha();
+  const red = ((1 - child.alpha()) * parent.alpha() * parent.red() + child.alpha() * child.red()) / alpha;
+  const green = ((1 - child.alpha()) * parent.alpha() * parent.green() + child.alpha() * child.green()) / alpha;
+  const blue = ((1 - child.alpha()) * parent.alpha() * parent.blue() + child.alpha() * child.blue()) / alpha;
```
jim-docker, line 43: what if alpha is 0?

Nokel81: If alpha is 0 then this algorithm will not take any RGB "value" from that entry because the browser wouldn't either.

Nokel81:
Closing because BG of transparent is now considered a bug, with that being static (wrt the theme) this will no longer be needed.

PR #2752, opened 2021-05-11
Allow dynamic menu titles in 'WelcomeMenuRegistration'

Nachasic:
Present an option to describe a welcome menu item title as a function that returns a string, to allow for dynamic captions.

### DIFF
--- a/src/extensions/registries/welcome-menu-registry.ts
+++ b/src/extensions/registries/welcome-menu-registry.ts
@@ -1,7 +1,7 @@
 import { BaseRegistry } from "./base-registry";
 
 export interface WelcomeMenuRegistration {
-  title: string;
+  title: string | (() => string);
   icon: string;
   click: () => void | Promise<void>;
 }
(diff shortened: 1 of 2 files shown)


Nokel81 (approved):

jakolehm (approved):

src/extensions/registries/welcome-menu-registry.ts
```diff
@@ -1,7 +1,7 @@
 import { BaseRegistry } from "./base-registry";
 
 export interface WelcomeMenuRegistration {
-  title: string;
```
Nokel81, line 4: Actually, given that this is just used a value. Shouldn't this be `React.ComponentLike`? @jakolehm ?

Nachasic: I've been thinking about it as well. Decided agains it for the sake of simplicity and consistency in UI

ixrock: Maybe `React.ComponentType` or `React.ReactNode` (includes `string`)?

Nachasic: Since this PR got approved, I will merge the changes as-is. We can return to this conversation later.

PR #2754, opened 2021-05-11
Add SystemName validation to AccessibleNamespaces

### LINKED ISSUE #2727: Feature request: Clean input for accesible namespaces

Nokel81:

fixes #2727 

supersedes #2730

### DIFF
--- a/src/renderer/components/cluster-settings/components/cluster-accessible-namespaces.tsx
+++ b/src/renderer/components/cluster-settings/components/cluster-accessible-namespaces.tsx
@@ -25,6 +25,7 @@ import type { Cluster } from "../../../../main/cluster";
 import { SubTitle } from "../../layout/sub-title";
 import { EditableList } from "../../editable-list";
 import { observable, makeObservable } from "mobx";
+import { systemName } from "../../input/input_validators";
 
 interface Props {
   cluster: Cluster;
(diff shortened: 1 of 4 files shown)


jim-docker (commented):
The debounce appears to only work once. If I erase the entry and repeat, I am informed of invalid space or dash immediately, even though I'm typing fast.

jim-docker:
probably not related, but I noticed that if you specify a very long name and submit, the trash icon is missing. How to delete the namespace?

Screen Shot 2021-05-27 at 8 40 31 PM

Nokel81:
Thanks for finding those two issues, will investigate tomorrow.

Nokel81:
@jim-docker I'll fix the horizontal scroll issue in this PR, but the other issue is something that is out of scope of this PR, since it would require changing a lot of `<Input>`.

jim-docker (approved):
It's fine to check for invalid characters on-the-fly but I'm not a big fan of on-the-fly validation of character combinations (like `-` is ok as long as it's not the first or last character). I think that kind of thing should be checked on submission.

Nokel81:
@jim-docker I agree. Though that requires debouncing the validation or having two different validators (which I think is worse). Something to do in a different PR.

PR #2755, opened 2021-05-11
Split UserManagement into seperate Cluster and Namespace Roles and RoleBindings

### LINKED ISSUE #2199: Separate Cluster Roles and Cluster Role Bindings

Separate Cluster Roles and Cluster Role Bindings from Roles and Role Bindings under "Access Control".

**Why is this needed**:

For consistency. It is counterintuitive that these are combined today, since they are technically separate resource types.

- Kubernetes distribution: EKS 1.16
- Desktop OS: macOS 10.15.7

Nokel81:
- EditableList blocks event propogation when adding an item

- Clarifications on which values should be rendered between cluster and
  non-cluster RB and R

Fixes https://github.com/lensapp/lens/issues/2199

Screenshots:
Screen Shot 2021-05-31 at 8 57 54 AM
Screen Shot 2021-05-31 at 8 57 47 AM
Screen Shot 2021-05-31 at 8 57 40 AM

Needed work:

- [x] ~~Add back name editing.~~ (This shouldn't have been possible, `kubectl` blocks name editing).
- [x] Remove multi-api from `KubeApi` and subscribing.


Nachasic (dismissed):
That's a hefty amount of changes. Overall LGTM

nevalla (changes requested):
Got app crash when tried to create new RoleBinding:

image

Can't open ClusterRoleBinding details:
image

When trying to create new ClusterRoleBinding nothing is happening when clicking + button.

Rules are not shown in ClusterRole details.

Also seeing some mobx errors
```
mobx] Encountered an uncaught exception that was thrown by a reaction or observer component, in: 'Reaction[Reaction@691]' Error: Unable to find node on an unmounted component.
    at findCurrentFiberUsingSlowPath (react-dom.development.js:4303)
    at findCurrentHostFiber (react-dom.development.js:4465)
    at findHostInstanceWithWarning (react-dom.development.js:25389)
    at findDOMNode (react-dom.development.js:26067)
    at Dialog.get elem [as elem] (dialog.tsx:74)
    at Dialog.value [as onClose] (dialog.tsx:128)
    at Dialog.close (dialog.tsx:109)
    at dialog.tsx:66
    at executeAction (mobx.esm.js:1128)
    at Reaction@691 (mobx.esm.js:1112)
    at reactionRunner (mobx.esm.js:2679)
    at Reaction.onInvalidate_ (mobx.esm.js:2660)
    at Reaction.runReaction_ (mobx.esm.js:2289)
    at runReactionsHelper (mobx.esm.js:2455)
    at reactionScheduler (mobx.esm.js:2430)
    at mobx.esm.js:2468
```

Nokel81:
Thanks for finding this, must have slipped in recently. But will take a look.

Nokel81:
@nevalla I have fixed that bug, it was due to `useDefineForClassFields` being turned on and my not updating this PR correctly.

Nachasic (dismissed):

nevalla:
Still seeing some crashes:

Trying to create new role binding:
```
Component stack:


    at EditableList (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:921266:9)
    at div
    at form
    at WizardStep (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:943630:9)
    at div
    at Wizard (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:943546:9)
    at div
    at div
    at Animate (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:903160:9)
    at Dialog (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:911522:9)
    at AddRoleBindingDialog (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:883838:9)
    at RoleBindings (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:884971:20)
    at Route (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:696420:29)
    at Switch (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:696626:29)
    at ErrorBoundary (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:921650:9)
    at main
    at div
    at http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:556143:80
    at UserManagement (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:888185:41)
    at Route (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:696420:29)
    at Switch (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:696626:29)
    at ErrorBoundary (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:921650:9)
    at main
    at div
    at MainLayout (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:930255:9)
    at ErrorBoundary (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:921650:9)
    at Router (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:696049:30)
    at App (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:903705:9)
    at ThemeProvider (http://08670916d2e0a2db6d528cd4cea1261b.localhost:62158/build/OpenLensDev.js:325564:24)
```

When trying to create new ClusterRoleBinding or ClusterRole nothing is happening when clicking + button.

nevalla:
Also to be noted that user should be able to select also cluster roles also when creating RoleBindings

Nokel81:
@nevalla I have fixed all those issues and it seems very stable now.

I plan to do a follow-up PR after #2892 is merged because then we have a UI for editing the rules of roles and clusterroles.

nevalla:
Is adding new `ClusterRole` working on your machine? For some reason it still does not work for me.

Also now when playing a bit with these changes, I would like to re-order these so that the order would be
Service Accounts, Cluster Roles, Roles, Cluster Role Bindings, Role Bindings, Pod Security Policies. WDYT?

Nokel81:
@nevalla I thought it was working, but if it isn't working for you then I should fix that.

As for the order, I'll just change it. I don't have a lot of stake either way.

nevalla:
Still some issues with ClusterRole dialog. I can't close it with the `Cancel` button.

src/common/utils/hash-set.ts
```diff
@@ -0,0 +1,147 @@
...
+
+export * from "./hashers";
+
+export function makeIterable<T>(iterator: Iterator<T>): IterableIterator<T> {
+  (iterator as IterableIterator<T>)[Symbol.iterator] = () => iterator as IterableIterator<T>;
+
+  return iterator as IterableIterator<T>;
+}
+
+export class ObservableHashSet<T> implements Set<T>, IInterceptable<ISetWillChange>, IListenable {
```
jakolehm, line 141: Do we really need to invent these new types? What's wrong with the existing ones? If we really need these please write unit tests.

src/common/utils/hashers.ts
```diff
@@ -0,0 +1,37 @@
...
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import { MD5 } from "crypto-js";
+import type { RoleBindingSubject } from "../../renderer/api/endpoints";
+import type { KubeObject } from "../../renderer/api/kube-object";
+
+export function hashKubeObject(k: KubeObject): string {
```
jakolehm: This is used only in single place. Do we really need it as an util? (are we going to use this somewhere else)

src/common/utils/hashers.ts
```diff
@@ -0,0 +1,37 @@
...
+  return k.metadata.uid;
+}
+
+export function hashRoleBindingSubject(subject: RoleBindingSubject): string {
```
jakolehm: Same here, used only in one place... does it need to be an util?

src/common/utils/n-fircate.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+/**
+ * Split an iterable into several arrays with matching fields
+ * @param from The iterable of items to split up
+ * @param field The field of each item to split over
+ * @param parts What each array will be filtered to
+ * @returns A `parts.length` tuple of `T[]` where each array has matching `field` values
+ */
+export function nFircate<T>(from: Iterable<T>, field: keyof T, parts: [T[typeof field]]): [T[]];
```
jakolehm, line 30: Extremely hard to understand what's happening here. Needs unit tests if we really need this kind of util.

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -97,13 +97,16 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
     return this.selectedNamespaces;
   }
 
-  getSubscribeApis() {
-    // if user has given static list of namespaces let's not start watches because watch adds stuff that's not wanted
+  subscribe() {
```
jakolehm, line 100: Why did we change this?

src/renderer/components/+user-management/+roles-bindings/details.tsx
```diff
@@ -19,35 +19,32 @@
...
 interface Props extends KubeObjectDetailsProps<RoleBinding> {
 }
 
 @observer
 export class RoleBindingDetails extends React.Component<Props> {
-  @observable selectedSubjects = observable.array<IRoleBindingSubject>([], { deep: false });
-
-  constructor(props: Props) {
-    super(props);
-    makeObservable(this);
```
jakolehm: We don't need `makeObservable` because no `@observable`?

src/renderer/components/+user-management/+service-accounts/details.tsx
```diff
@@ -19,27 +19,27 @@
...
+import { Icon } from "../../icon";
+import { getDetailsUrl, KubeObjectDetailsProps } from "../../kube-object";
+import { KubeObjectMeta } from "../../kube-object/kube-object-meta";
+import { Spinner } from "../../spinner";
+import { ServiceAccountsSecret } from "./secret";
 
 interface Props extends KubeObjectDetailsProps<ServiceAccount> {
 }
 
-@observer
```
jakolehm: Why observer removed?

src/renderer/components/app.scss
```diff
@@ -29,7 +29,7 @@
   --unit: 8px;
   --padding: var(--unit);
   --margin: var(--unit);
-  --border-radius: 3px;
+  --border-radius: 5px;
```
jakolehm: This probably should be reverted, out-of-scope for this PR.

src/renderer/components/app.tsx
```diff
@@ -20,57 +20,57 @@
  */
 import React from "react";
 import { observable, makeObservable } from "mobx";
+import { webFrame } from "electron";
```
jakolehm: Did you shuffle order of imports by accident? Quite hard to say what's going on with this file.

src/renderer/components/editable-list/editable-list.scss
```diff
@@ -23,7 +23,7 @@
   .el-contents {
     display: flex;
     flex-direction: column;
-    margin-top: $padding * 2;
+    margin: $padding 0px;
```
jakolehm, line 26: Any explanation why do we want to change EditableList styles?

src/renderer/components/editable-list/editable-list.tsx
```diff
@@ -71,17 +65,15 @@ export class EditableList<T> extends React.Component<Props<T>> {
...
             onSubmit={this.onSubmit}
             placeholder={placeholder}
-            onChange={val => this.currentNewItem = val}
           />
         </div>
         <div className="el-contents">
           {
             items.map((item, index) => (
-              <div key={`${item}${index}`} className="el-item Badge">
+              <div key={`${item}${index}`} className="el-item">
```
jakolehm, line 76: Why Badge?

src/renderer/components/editable-list/editable-list.tsx
```diff
@@ -71,17 +65,15 @@ export class EditableList<T> extends React.Component<Props<T>> {
       <div className="EditableList">
         <div className="el-header">
           <Input
-            theme="round-black"
-            value={this.currentNewItem}
+            theme="round"
```
jakolehm, line 68: Why different theme?

src/renderer/components/input/input.scss
```diff
@@ -48,7 +48,7 @@
     --flex-gap: #{$padding / 1.5};
 
     position: relative;
-    padding: $padding /4 * 3 0;
+    padding: $padding / 4 * 3 0;
```
jakolehm, line 51: These are global changes to Input are out-of-scope for this PR. Please describe why these are needed (if they are needed).

src/renderer/components/input/input.tsx
```diff
@@ -41,7 +41,7 @@ type InputElement = HTMLInputElement | HTMLTextAreaElement;
 type InputElementProps = InputHTMLAttributes<InputElement> & TextareaHTMLAttributes<InputElement> & DOMAttributes<InputElement>;
 
 export type InputProps<T = string> = Omit<InputElementProps, "onChange" | "onSubmit"> & {
-  theme?: "round-black";
+  theme?: "round-black" | "round";
```
jakolehm, line 44: Did we invent new theme, `round` here?

src/renderer/components/input/input.tsx
```diff
@@ -90,7 +90,7 @@ export class Input extends React.Component<InputProps, State> {
     return this.state.valid;
   }
 
-  setValue(value: string) {
+  setValue(value = "") {
```
jakolehm, line 93: Why do we change logic of `Input`?

src/renderer/components/kube-object/kube-object-details.tsx
```diff
@@ -99,11 +100,7 @@ export class KubeObjectDetails extends React.Component {
   }
 
   @computed get object() {
-    const store = apiManager.getStore(this.path);
-
-    if (store) {
-      return store.getByPath(this.path);
-    }
+    return apiManager.getStore(this.path)?.getByPath(this.path);
```
jakolehm: Idk how changes within this file are related to PR. Looks better but are these added here by accident?

src/renderer/components/select/select.scss
```diff
@@ -46,16 +46,15 @@ html {
     }
 
     &__control {
-      border: none;
+      border: solid 1px $halfGray;
```
jakolehm: And again, we are modifying global component styles without any explanation. Probably out-of-scope for this PR?

src/renderer/components/select/select.tsx
```diff
@@ -65,8 +66,10 @@ export class Select extends React.Component<SelectProps> {
     makeObservable(this);
   }
 
-  @computed get theme() {
-    return this.props.themeName || ThemeStore.getInstance().activeTheme.type;
+  @computed get themeClass() {
```
jakolehm, line 69: Why did we change `Select`? Out of scope for this PR, right?

src/renderer/kube-object.store.ts
```diff
@@ -306,51 +306,31 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
     });
   }
 
-  getSubscribeApis(): KubeApi[] {
-    return [this.api];
-  }
-
-  subscribe(apis = this.getSubscribeApis()) {
```
jakolehm, line 316: So we decided to change KubeObjectStore subscibe api in this PR? Why?

src/common/utils/hashers.ts
Nokel81: This is used in two places, `src/renderer/components/+user-management/+cluster-roles-bindings/dialog.tsx` and `src/renderer/components/+user-management/+roles-bindings/dialog.tsx` but I can move it if you would like.

src/renderer/components/+namespaces/namespace.store.ts
Nokel81: Removing `getSubscribeApis()` very much simplifies the `subscribe()` logic because it is not fixed at always 1 API.

src/renderer/components/+user-management/+roles-bindings/details.tsx
Nokel81: That is correct.

src/renderer/components/editable-list/editable-list.scss
Nokel81: The `*/details.tsx` files use `<EditableList>` so these changes we to make it compatible with several of them in a column.

src/renderer/components/app.tsx
Nokel81: Yes, will revert.

src/renderer/components/editable-list/editable-list.tsx
Nokel81: Because when used within a `<Wizard>` we don't want it to be dark, but I do want the round outlined version.

src/renderer/components/editable-list/editable-list.tsx
Nokel81: Do you mean, why am I removing that `className`? Because I am now doing the css for these items directly.

src/renderer/components/input/input.scss
Nokel81: They are needed to facilitate the usability of `<EditableList>` within `<Wizard>`. I have tried to make sure that there are no visual changes to the rest of the application.

src/renderer/components/input/input.tsx
Nokel81: Yes, just like `round-black` isn't a theme that is selectable. (From the `ThemeStore`'s POV).

src/renderer/components/input/input.tsx
Nokel81: Because `<Wizard>` wraps its contents in a `<form>` and `<EditableList>` also uses the equivalent of `onEnter`, these changes are needed to make sure that the `<form>.onSubmit` isn't always called.

src/renderer/components/select/select.tsx
Nokel81: There is one more bug with `<Select>` still (which I found while writing this PR). That when used within `<Wizard>` the background colour for "Selected and hover" is still the global theme.

src/renderer/kube-object.store.ts
Nokel81: Because all stores now have only one API that they know about/need to know about. Since there are now separate stores for the `ClusterRoles` and `ClusterRoleBindings`.

If you really want I can move that to a separate PR, but I felt that it made sense to do it in the same PR that allowed for it to happen.

src/renderer/components/kube-object/kube-object-details.tsx
Nokel81: Yes by accident. I'll revert into a separate PR.

src/renderer/components/select/select.scss
Nokel81: I don't remember about this one. The other style changes I remember being needed for `<Wizard>` but I don't remember if this one way, will revert and check.

src/common/utils/n-fircate.ts
Nokel81: Will write some tests.

src/common/utils/hash-set.ts
Nokel81: Will write some tests. I believe that this is a useful type since it acts like a `Set<T>` with custom hash function (where hashing is 1-to-1).

jim-docker:
The Create ClusterRole dialog does not accept any key presses when entering the name

src/common/utils/hash-set.ts
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  @action
+  replace(other: ObservableHashSet<T> | ObservableSet<T> | Set<T> | readonly T[]): this {
+    if (Array.isArray(other)) {
+      this.clear();
+      other.forEach(value => this.add(value));
+    } else if (other instanceof Set || other instanceof ObservableHashSet || other instanceof ObservableSet) {
+      this.clear();
+      other.forEach(value => this.add(value));
```
jim-docker: Doing the same thing in both cases?

Nokel81:
> The Create ClusterRole dialog does not accept any key presses when entering the name

Cannot reproduce, can you please try again

src/renderer/components/+user-management/+cluster-role-bindings/details.tsx
```diff
@@ -0,0 +1,167 @@
...
+            <TableCell>{roleRef.apiGroup}</TableCell>
+          </TableRow>
+        </Table>
+
+        <DrawerTitle title="Bindings" />
+        {subjects.length > 0 && (
+          <Table selectable className="bindings box grow">
+            <TableHead>
+              <TableCell checkbox />
+              <TableCell className="binding">Binding</TableCell>
```
jim-docker: The title of this drawer is Bindings so this cell header could be Name instead of Binding?

src/renderer/components/+user-management/+cluster-role-bindings/details.tsx
```diff
@@ -0,0 +1,167 @@
...
+            {
+              subjects.map((subject, i) => {
+                const { kind, name } = subject;
+                const isSelected = selectedSubjects.includes(subject);
+
+                return (
+                  <TableRow
+                    key={i}
+                    selected={isSelected}
+                    onClick={prevDefault(() => this.selectSubject(subject))}
```
jim-docker: seems a bit inconsistent that clicking on the whole item selects it (instead of just the checkbox). Or is there a precedent for selection on a details page?

jim-docker (commented):
tried to add a new binding from the role bindings details page, it added the binding but also duplicated the existing binding (user1). Tried to delete the binding but it stayed. Then couldn't open the details page (nor the vertical dots menu). This video shows some of the weirdness:

https://user-images.githubusercontent.com/40840436/120718027-9b9f9f00-c496-11eb-83c8-5aac374e73ae.mov

jim-docker:
> > The Create ClusterRole dialog does not accept any key presses when entering the name
> 
> Cannot reproduce, can you please try again

Happens all the time for me. 


https://user-images.githubusercontent.com/40840436/120722397-a90c5780-c49d-11eb-8fb8-44f9fe626129.mov

src/renderer/components/+user-management/+cluster-role-bindings/details.tsx
Nokel81: How that is normal (iirc).

jim-docker:
can't seem to be able to select service accounts on the Cluster Role Binding dialog:


https://user-images.githubusercontent.com/40840436/120861932-11b70b00-c556-11eb-893c-31eeba0793fa.mov

(If I can't actually pick any of these they shouldn't be in the dropdown list)

Nokel81:
@jim-docker I have now fixed that issue. The previously selected service accounts should be also present when opening the dialogs to edit a binding.

jim-docker (commented):
Is this a bug?

https://user-images.githubusercontent.com/40840436/121082017-7ae38c00-c7ab-11eb-861b-af37b7768596.mov

Nokel81:
Yes

Nokel81:
@jim-docker I have fixed that issue, and another one I found while testing concerning using `ServiceAccount`'s as binding targets.

src/renderer/api/endpoints/role-binding.api.ts
```diff
@@ -24,15 +24,17 @@ import { KubeObject } from "../kube-object";
...
 
-export interface IRoleBindingSubject {
-  kind: string;
+export type RoleBindingSubjectKind = "Group" | "ServiceAccount" | "User";
+
+export interface RoleBindingSubject {
+  kind: RoleBindingSubjectKind;
   name: string;
-  namespace?: string;
+  namespace: string;
```
nevalla: does this mean that `Group` or `User` must always be namespaced?

src/renderer/components/+user-management/+cluster-role-bindings/dialog.tsx
```diff
@@ -0,0 +1,283 @@
...
+        <SubTitle title="Cluster Role" />
+        <Select
+          themeName="light"
+          placeholder="Select cluster role ..."
+          isDisabled={this.isEditing}
+          options={this.clusterRoleRefoptions}
+          value={this.selectedRoleRef}
+          onChange={({ value }) => {
+            this.selectedRoleRef = value;
+            this.bindingName = this.selectedRoleRef.getName();
```
jim-docker: it shouldn't do this if `this.bindingName` has already been set (i.e.when the user has already entered a name in the "Binding Name" field). Hmm, I guess this would lead to problems if the user changes the cluster role. Maybe don't set the binding name to the role name till create is invoked (and it wasn't already set by the user)?

src/renderer/kube-object.store.ts
nevalla: should we also modify signature of `ItemStore.subscribe()` as well? (separate PR is fine if needed)

nevalla (dismissed):
Tested and seems to work as expected! In general LGTM, added one question that we need to verify before merging.

src/renderer/components/+user-management/+role-bindings/dialog.tsx
```diff
@@ -0,0 +1,305 @@
...
+          options={this.roleRefOptions}
+          value={this.selectedRoleRef}
+          onChange={({ value }) => {
+            this.selectedRoleRef = value;
+
+            if (this.selectedRoleRef.kind === "Role") {
+              this.bindingNamespace = this.selectedRoleRef.getNs();
+            }
+            
+            this.bindingName = this.selectedRoleRef.getName();
```
jim-docker: same here

src/renderer/components/+user-management/+role-bindings/view.tsx
```diff
@@ -0,0 +1,91 @@
...
+            { title: "Name", className: "name", sortBy: columnId.name, id: columnId.name },
+            { className: "warning", showWithColumn: columnId.name },
+            { title: "Namespace", className: "namespace", sortBy: columnId.namespace, id: columnId.namespace },
+            { title: "Bindings", className: "bindings", sortBy: columnId.bindings, id: columnId.bindings },
+            { title: "Age", className: "age", sortBy: columnId.age, id: columnId.age },
+          ]}
+          renderTableContents={(binding: RoleBinding) => [
+            binding.getName(),
+            <KubeObjectStatusIcon key="icon" object={binding} />,
+            binding.getNs() || "-",
```
jim-docker: role bindings must be associated with a namespace so it'd never be `"-"`?

src/renderer/components/+user-management/+role-bindings/dialog.tsx
```diff
@@ -0,0 +1,305 @@
...
+          maxMenuHeight={200}
+        />
+      </>
+    );
+  }
+
+  render() {
+    const { ...dialogProps } = this.props;
+    const [action, nextLabel] = this.isEditing ? ["Edit", "Update"] : ["Add", "Create"];
+    const disableNext = !this.selectedRoleRef || !this.selectedBindings.length;
```
jim-docker: should it test for namespace selected too? (I got a server error when I tried to create a role binding without selecting a namespace)

src/renderer/components/+user-management/+roles/details.scss
```diff
@@ -0,0 +1,21 @@
+.ClusterRoleDetails {
```
jim-docker: Should this be
```suggestion
.RoleDetails {
```
?

src/renderer/kube-object.store.ts
Nokel81: So the signature of `ItemStore.subscribe()` is `subscribe(...args: any[])`. Which is rather too JS-like for my liking. I think other changes will have to be done before `ItemStore.subscribe()` can be changed, though TBH I think only `KubeObjectStore` overrides `subscribe()` so that might be a possible solution (for a follow up PR).

src/renderer/components/+user-management/+cluster-role-bindings/dialog.tsx
Nokel81: How about detecting if the name has been changed and only overriding it if it hasn't been?

src/renderer/api/endpoints/role-binding.api.ts
Nokel81: For `RoleBinding` subjects yes, they need to be the same as the namespace that the `RoleBinding` is in.

src/renderer/components/+user-management/+cluster-role-bindings/dialog.tsx
jim-docker: But if I pick roleRef foo and then pick roleRef bar the name would stay foo? 

Personally I think the name should be the first input field and be required to be specified by the user. No autofill based on the role.

Nokel81: So it is common for the name of the binding to be the same as the role/clusterrole.

No, if you pick roleRef `foo` and then pick roleRef `bar`. Then name would change to `bar` unless you had modified it.

src/renderer/components/+user-management/+role-bindings/dialog.tsx
Nokel81: Sounds good

src/renderer/api/endpoints/role-binding.api.ts
nevalla: > For `RoleBinding` subjects yes, they need to be the same as the namespace that the `RoleBinding` is in.

That's not exactly true, the namespace just indicates in what namespace the service account does exist. It does not need to be the same namespace where RoleBinding is created.

`user` and `group` does not need namespace, like this [example](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-example):
```
apiVersion: rbac.authorization.k8s.io/v1
# This role binding allows "jane" to read pods in the "default" namespace.
# You need to already have a Role named "pod-reader" in that namespace.
kind: RoleBinding
metadata:
  name: read-pods
  namespace: default
subjects:
# You can specify more than one "subject"
- kind: User
  name: jane # "name" is case sensitive
  apiGroup: rbac.authorization.k8s.io
roleRef:
  # "roleRef" specifies the binding to a Role / ClusterRole
  kind: Role #this must be Role or ClusterRole
  name: pod-reader # this must match the name of the Role or ClusterRole you wish to bind to
  apiGroup: rbac.authorization.k8s.io

Nokel81: Ah okay, will revert that. From my testing it seemed like it was always there.

jim-docker (approved):
This could be addressed in a followup PR, if necessary:

https://user-images.githubusercontent.com/40840436/121401303-8effc880-c926-11eb-9bbe-3566269b0f32.mov

For a binding created via the create dialog, then updated via the update dialog, this error message appears. Though the resource is probably fine, the dialog stays open, and pressing "Update" again (which is probably not necessary?) works without error.

Nokel81:
Yes I'll fix that with a follow up PR.

PR #2758, opened 2021-05-12
Add connect/disconnect methods to KubernetesCluster kind

jakolehm:


aleksfront (dismissed):

Nokel81 (commented):
Why is this needed?

src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -19,6 +20,38 @@ export class KubernetesCluster extends CatalogEntity<CatalogEntityMetadata, Kube
   public readonly apiVersion = "entity.k8slens.dev/v1alpha1";
   public readonly kind = "KubernetesCluster";
 
+  async connect(): Promise<boolean> {
```
Nokel81: I don't think these need to return anything. Especially since it is currently not very useful of a return value.

src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -19,6 +20,38 @@ export class KubernetesCluster extends CatalogEntity<CatalogEntityMetadata, Kube
...
+    if (app) {
```
Nokel81, line 45: Is there a use case for this being run on `main` at all? I guess that should be discussed in the future. But it seems very strange indeed for any `CatalogEntity` on `main` providing a function something like `onContextMenuOpen` or even `connect` for that matter.

src/common/catalog-entities/kubernetes-cluster.ts
jakolehm: Changed to `Promise<void>`.

src/common/catalog-entities/kubernetes-cluster.ts
jakolehm: Use case for `main`, currently no. It just feels old cluster class all over again if it works partially on the other side.

Nokel81: I don't understand:

> It just feels old cluster class all over again

but I think this is a better version.

Nokel81 (approved):

nevalla (approved):
LGTM

src/common/catalog-entities/kubernetes-cluster.ts
jakolehm: Lens 4.x `Cluster` object had public methods that worked only in `main`.

PR #2761, opened 2021-05-12
Fix namespace selector gradients

aleksfront:
CSS-only alternative to https://github.com/lensapp/lens/pull/2751

white theme
dark theme
roles dialog


Nokel81 (dismissed):
This is not a general fix as it claims to be. If `<NamespaceSelectFilter>` (or more specifically the gradients by using the `className="NamespaceSelectFilter"` option) is desired to be used within a `<Wizard>` then this is shown:

Screen Shot 2021-05-12 at 9 08 50 AM

This code will have to be changed every single time a location of `<NamespaceSelectFilter>` is moved.

aleksfront:
>This is not a general fix as it claims to be.

Fixed.
Screenshot 2021-05-12 at 16 20 08

Nokel81:
What about #2743 where I have moved the component? Yes in that PR I change the default gradient, but that was a temp fix until a more general solution is implemented.

aleksfront:
> What about #2743 where I have moved the component? Yes in that PR I change the default gradient, but that was a temp fix until a more general solution is implemented.

In example you provided the gradients will work as expected since colors will be given from the active theme. Or maybe I misunderstood what general problem left.

Nokel81:
What I mean is that I had to change the "default" gradient colour from `var(--contentColor)` to `var(--layoutBackground)`. It is that sort of special casing which my solution would do away with.

Namely, currently (on master) we really only support gradients in one place. With this PR there is now 2 (if the user opts in with specifying `theme: "light"`). However, with my PR it works in all locations on first attempt.

aleksfront:
Why we need to set another colors on namespace selector gradients? It just shows user that he's able to scroll things left or right.

Nokel81:
Exactly, so I don't think that a consumer of the `<NamespaceSelectFilter>` should need to care about that sort of thing. It is an implementation detail. Which is why I don't think we should have to pass anything in (like `theme: "light"`). All that should be received from the rendered context. 

And since the background of `Select` is transparent we need to do some calculation in the general case (for a use anyway component).

src/renderer/components/+namespaces/namespace-select-filter.scss
```diff
@@ -1,5 +1,11 @@
 .NamespaceSelectFilter {
+  --gradientColor: var(--contentColor);
+
   .Select {
+    &.theme-light {
+      --gradientColor: #fff;
```
jim-docker: why does it have to be hard-coded? Isn't there a light theme-appropriate variable for this?

jim-docker (dismissed):
LGTM. It looks like a general solution to me...

src/renderer/components/+namespaces/namespace-select-filter.scss
aleksfront: Well, we specifying `#fff` for a theme which is always will be light. If we would set some variable in theme files, we have to write something like `"themeLightNamespaceGradient": #fff` in both `lens-dark.json` and `lens-light.json` files.

jim-docker: Yeah, it just seems odd that there is a variable that can be used when it is not light themed, but not when it is light themed. And what if one day a pixel-perfect-minded designer decides the light theme should be more off-white, and this value should be `#efe`? :)

Nokel81: @jim-docker That is why I don't particularly think a pure CSS solution is a good idea. Because it basically requires future tweaking. My PR shouldn't need such tweaking in the future.

jim-docker (dismissed):

src/renderer/components/+namespaces/namespace-select-filter.scss
jim-docker: But it should be possible to do it in CSS without requiring future tweaking, right? Variables for foreground and background colours that are a function of the theme? Probably beyond the scope of this PR.

Nokel81: Not quite. As I expounded in my PR, the background of this component is desirable to be the same as background colour around it (no matter where it is). 

And because our background colours change depending on context (even within a single theme) it is not possible do this in general by CSS variables because even if we had some `--currentBackgroundColour` that we set in a few places it wouldn't ever be correct in general. Especially since in many places we set the background colour to be `rgba(0, 0, 0, 0)` (or `transparent`). 

This causes a problem because this gradient is going from `transparent` to the "actual" background colour.

jim-docker: Did we consider instead using an ellipsis at front and back of the displayed text as needed?

Nokel81: Do you mean by that the css option `text-overflow: ellipsis`? Sure that would be another possible solution. But one of the nice things about having `text-overflow: scroll` and these gradients is it allows the user to view which are selected without having to open up the component first.

jim-docker: I meant use ellipsis as indicators but still bet able to scroll. Does `text-overflow: ellipsis` do that?

Nokel81: No, it truncates output you would have to do the truncation yourself. Which is even more difficult because then you have to care about font rendering.

jim-docker: little arrowheads (like &#9658;)? Then you can click on them to scroll

ixrock (dismissed):
LGTM

aleksfront:
@Nokel81 I'm gonna merge this decision until more complex color problems will show up. Can you dismiss your requested changes?

aleksfront:
Resolved a conflict. Please do re-review.

ixrock (approved):

Nokel81 (approved):

PR #2770, opened 2021-05-13
Include the hotbar index in the label displayed for a hotbar

### LINKED ISSUE #2681: Hotbar chooser should show index
Now it shows just the name.. should be something like `<index>: <name>` (for example: `2: dev`)

jim-docker:
made a function to generate the display label of a hotbar given the hotbar id. This function is used in generating the switch hotbar list, remove hotbar list, and the hotbar selector.

fixes #2681

### DIFF
--- a/src/renderer/components/hotbar/hotbar-display-label.ts
+++ b/src/renderer/components/hotbar/hotbar-display-label.ts
@@ -21,14 +21,14 @@
 
 import { HotbarStore } from "../../../common/hotbar-store";
 
-export function hotbarDisplayLabel(id: string, withName = false) : string {
+export function hotbarDisplayLabel(id: string, withName = true) : string {
   const hotbarStore = HotbarStore.getInstance();
   const index = hotbarStore.hotbarIndex(id) + 1;
 
   if (withName) {
     const hotbar = hotbarStore.getById(id);
 
-    return [ index, ": ", hotbar.name ].join("");
+    return `${index}: ${hotbar.name}`;
   }
 
   return index.toString();
--- a/src/renderer/components/hotbar/hotbar-display-label.ts
+++ b/src/renderer/components/hotbar/hotbar-display-label.ts
@@ -21,15 +21,16 @@
 
 import { HotbarStore } from "../../../common/hotbar-store";
 
-export function hotbarDisplayLabel(id: string, withName = true) : string {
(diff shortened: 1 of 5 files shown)


src/renderer/components/hotbar/hotbar-display-label.ts
```diff
@@ -0,0 +1,35 @@
...
+import { HotbarStore } from "../../../common/hotbar-store";
+
+export function hotbarDisplayLabel(id: string, withName = false) : string {
+  const hotbarStore = HotbarStore.getInstance();
+  const index = hotbarStore.hotbarIndex(id) + 1;
+
+  if (withName) {
+    const hotbar = hotbarStore.getById(id);
+
+    return [ index, ": ", hotbar.name ].join("");
```
Nokel81: I think formatting should use the backtick formatting style.

```suggestion
    return `${index}: ${hotbar.name}`;
```

src/renderer/components/hotbar/hotbar-display-label.ts
```diff
@@ -0,0 +1,35 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import { HotbarStore } from "../../../common/hotbar-store";
+
+export function hotbarDisplayLabel(id: string, withName = false) : string {
```
Nokel81: It seems strange to have `false` as the default value. Especially since there are more uses when this is `true`.

I would suggest just making it `withName: boolean`.

Nokel81 (commented):
Looks good

Nokel81 (dismissed):

Nokel81 (approved):

jakolehm (approved):

PR #2774, opened 2021-05-14
Add `esbuild-loader` (experimental)

chenhunghan:
PR for `esbuild` experiments, add `esbuild-loader` for non-production environments for main/renderer, and extensions, 

I am surprise that `make dev` works without any issue, (and it's much faster then `ts-loader`), please test on your machine.


chenhunghan:
Change to draft, sorry for the noises!

PR #2775, opened 2021-05-14
Add IPC capabilities for Extensions

Nokel81:


docs/extensions/guides/ipc.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+If you want to register a "handler" you would call `this.handleIpc(...)` instead.
+
+The `LensExtension.prototype.disopsers` is a list of `() => void`'s (or `Utils.Disposer`'s).
+Those functions are sort of like destructors in that they should clean up some state.
+Some functions return them to indicate that it will clean up the state which was just registered.
+
+As an extension developer you do not need to run them yourself.
+Lens will run them when your extension gets deactivated or uninstalled.
```
jim-docker: Can't `listenIpc()` be written to do the push() itself, so the extension dev doesn't need to know anything about these disposers?

Nokel81: It could, that might be a better design.

docs/extensions/guides/ipc.md
```diff
@@ -0,0 +1,92 @@
...
+Both `LensMainExtension` and `LensRendererExtension` can do this sort of IPC.
+
+### Request Based IPC
+
+This is more like a Remote Procedure Call (RPC).
+With this sort of IPC the caller waits for the result from the other side.
+This is accomplished by returning a `Promise<T>` which needs to be `await`-ed.
+
+This is a unidirectional form of communication.
+Only `LensRendererExtension` can initiate this kind of request, and only `LensMainExtension` and respond this this kind of request.
```
stefcameron: "and only `LensMainExtension` and respond this this kind of request." -> (typos) "and only `LensMainExtension` can respond to this kind of request."

docs/extensions/guides/ipc.md
```diff
@@ -0,0 +1,92 @@
...
+    return args.length === 1 && typeof args[0] === "string";
+  }
+}
+```
+
+If you want to register a "handler" you would call `this.handleIpc(...)` instead.
+
+The `LensExtension.prototype.disopsers` is a list of `() => void`'s (or `Utils.Disposer`'s).
+Those functions are sort of like destructors in that they should clean up some state.
+Some functions return them to indicate that it will clean up the state which was just registered.
```
stefcameron: This isn't clear. Is there an optional 4th property in `this.listenIpc()` or `this.handleIpc()` called `dispose` that takes a `() => void` function that will be called on extension uninstall/disable if I provide one, similar to how one uses the `useEffect()` hook in React, whereby I can optionally return a `() => void` function that React will call if/when my component gets disposed so I can undo some state, like remove DOM node event listeners I may have added?

docs/extensions/guides/ipc.md
```diff
@@ -0,0 +1,92 @@
...
+
+  initializeListener = (event: Types.IpcMainEvent, uid: string) => {
+    console.log(`starting to initialize: ${uid}`);
+  };
+
+  initializeVerifier = (args: unknown[]): args is [uid: string] => {
+    return args.length === 1 && typeof args[0] === "string";
+  }
+}
+```
```
stefcameron: Do you do this very same thing from `LensRendererExtension` and then from `main` call `this.sendIpc("channel", ...args)`? Does it matter if the channel names are different? Or listeners are a way for `main` to push to `renderer` as well as, optionally, `renderer` to push to `main`?

docs/extensions/guides/ipc.md
```diff
@@ -0,0 +1,92 @@
...
+
+The verification function should do some cursory validation on the values send along the channel.
+Your handler or listener will not be called if it fails this validation.
+Instead an error or log message will occur.
+This should help with debugging because you are notified immediately that there is a mismatch between what you are expecting and what was sent.
+
+## Using IPC
+
+Calling IPC is very simple.
+If you are meaning to do an event based call, merely call `this.sendIpc(<channel>, ...<args>)` from within your extension.
```
stefcameron: Is there a restriction to the data types that can be included in `args`? Do they all have to be JSON-serializable types, or can they be things like ES6 `Map` and `Set`, or even `Function` (things that are beyond the scope of JSON, but that I can freely pass around between functions within a normal JS program)?

stefcameron (commented):
Thanks for writing this POC!

docs/extensions/guides/ipc.md
Nokel81: No `this.listenIpc()` and `this.handleIpc()` currently return `Disoper`. But I am starting to like Jim's suggestion so I'll probably remove this section.

docs/extensions/guides/ipc.md
Nokel81: The channel names have to be the same and you have to register a "listener" on `renderer` for events sent from `main` to be known about.

docs/extensions/guides/ipc.md
Nokel81: That is probably a good thing to call out. It can be more than JSON-serializable. It has to follow the https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm.

So things like `Map` and `Set` are good but `Function` will throw an error

docs/extensions/guides/ipc.md
stefcameron: I think that needs to be clarified in the docs, then. The docs are only telling one side of the story.

docs/extensions/guides/ipc.md
stefcameron: I think so, yes. That should definitely be called out.

docs/extensions/guides/ipc.md
```diff
@@ -40,38 +40,46 @@ The general terminology is as follows:
...
+  initializeVerifier = (args: unknown[]): args is [id: string] => {
     return args.length === 1 && typeof args[0] === "string";
   }
 }
 ```
 
-If you want to register a "handler" you would call `this.handleIpc(...)` instead.
+File2:
+```typescript
+import { LensMainExtension, Interface, Types } from "@k8slens/extensions";
```
stefcameron: `LensRendererExtension` here instead of the Main import?

docs/extensions/guides/ipc.md
```diff
@@ -40,38 +40,46 @@ The general terminology is as follows:
 
 To register either a handler or a listener, you should do something like the following:
 
+File1:
```
stefcameron: Can we call this `main.ts` instead of "File1" so it's clear this is code you could put on the `main` thread?

docs/extensions/guides/ipc.md
```diff
@@ -40,38 +40,46 @@ The general terminology is as follows:
...
 
-  initializeVerifier = (args: unknown[]): args is [uid: string] => {
+  initializeVerifier = (args: unknown[]): args is [id: string] => {
     return args.length === 1 && typeof args[0] === "string";
   }
 }
 ```
 
-If you want to register a "handler" you would call `this.handleIpc(...)` instead.
+File2:
```
stefcameron: Can we call this `renderer.ts` instead of "File1" so it's clear this is code you could put on the `renderer` thread?

docs/extensions/guides/ipc.md
```diff
@@ -40,38 +40,46 @@ The general terminology is as follows:
...
+export class ExampleExtensionRenderer extends LensRendererExtension {
+  onActivate() {
+    setTimeout(() => this.sendIpc("initialize", "an-id"), 5000);
+  }
+}
+```
+
+As this example shows: the channel names *must* be the same.
+It should also be noted that "listeners" and "handlers" are specific to either `LensRendererExtension` and `LensMainExtension`.
+There is no behind the scenes transfer of these functions.
```
stefcameron, line 113: So from this, my understanding is that a listener is not a two-way channel or a two-way pipe. It's for one side or the other to declare they're just listening, and you use this when, in this case, as `main`, you just want to be notified of some event, and the `renderer` here notifies without waiting for a response.

And if I wanted the opposite to happen, where `main` notifies `renderer` of something, then in `LensRendererExtension`, I would declare this:

```javascript
this.listenIpc({
  channel: "initialize",
  listener: this.initializeListener,
  verifier: this.initializeVerifier,
});
```

Now would this cause problems if I use the same channel name?

Nokel81: Yes that is correct, it is not a two way channel.

If you want `renderer` to wait for a response from `main`. Then on `main` you should use `this.handleIpc()` and on `renderer` you should use `await this.invokeIpc()`. (I think that is in the docs).

I guess I should probably change `this.sendIpc()` to `this.broadcastIpc()` because that is what it is doing. It is sending to all frames and `main`.

So I would recommend using a different name if you want to have `main` be the source.

Will add some items to the docs about this.

docs/extensions/guides/ipc.md
```diff
@@ -0,0 +1,105 @@
...
+
+If you want to register a "handler" you would call `this.handleIpc(...)` instead.
+The cleanup of these handlers is handled by Lens itself.
+
+### Note about verification
+
+We require that extension developers provide a verification function when registering a listener or handler.
+This is done as a preventative measure to help separate issues that can happen at runtime.
+While it is possible to use the unary truth function, this is highly discouraged.
```
stefcameron: I'm a big proponent of runtime verifications. I wrote [rtv.js](https://rtvjs.stefcameron.com/) after all, and we use it in our MDS Dev builds for that very purpose. So I get the logic of enforcing the verifier, but as much as I like runtime verification and its benefits, I don't think it should be enforced. I see this as a nice to have for that extra step of quality. But because there's nothing you can do to prevent an extension developer from just providing you with `() => true`, I think it'll be seen as strange to be required.

But if you're going to require it, then I think it would be nice to also support throwing an error to fail the validation, and then Lens can log the error as a result.

There are a number of runtime validation libraries people will likely want to use, like [yup](https://github.com/jquense/yup), [joi](https://github.com/sideway/joi), or maybe even [rtv.js](https://rtvjs.stefcameron.com/) 🙂 

If I were using `rtv.js` here, I'd find it handy to write this:

```javascript
  initializeVerifier = (args) => {
    rtv.verify(args, [rtv.ARRAY, { $: [rtv.STRING], length: 1 }]);
  }
```

And then let that blow-up if `args` is `[]` or `['']` or `['asdf', 3]` or not an array.

docs/extensions/guides/ipc.md
stefcameron: Sounds good.

docs/extensions/guides/ipc.md
Nokel81: So for some context this will work too as it stands since we currently only progress if the return value is truthy. We even already have a `try/catch` block and do log the error.

The concept of returning `true` here comes from Typescript's https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

jakolehm: I also think this should not be enforced. I get this kind of verification if listener/handler would have external consumers but now the consumer is the extension itself.

docs/extensions/guides/ipc.md
jakolehm: > I guess I should probably change this.sendIpc() to this.broadcastIpc() because that is what it is doing. It is sending to all frames and main.

SGTM.

src/extensions/lens-renderer-extension.ts
```diff
@@ -60,4 +62,27 @@ export class LensRendererExtension extends LensExtension {
   async isEnabledForCluster(cluster: Cluster): Promise<Boolean> {
     return (void cluster) || true;
   }
+
+  listenIpc<
```
jakolehm: Could we move all extension "ipc" methods to separate (singleton?) class? If we bundle these methods to extension class itself it means that extension needs to pass extension instance around just to be able to communicate via ipc.. which feels a bit clumsy.

docs/extensions/guides/ipc.md
```diff
@@ -40,57 +40,69 @@ The general terminology is as follows:
 
 To register either a handler or a listener, you should do something like the following:
 
-File1:
+`main/extension.ts`:
```
stefcameron: Isn't the required file name `main.ts`? It doesn't necessarily need to be, if you re-configure your Webpack build to have an entry other than `main.ts` and then output to `main.ts` when building, but I think your generator creates a file named `main.ts` where we find the class that extends `LensMainExtension`, so I think readers will be more familiar with things if we stick with `main.ts` here instead of `main/extension.ts`...

docs/extensions/guides/ipc.md
```diff
@@ -40,57 +40,69 @@ The general terminology is as follows:
...
 }
 ```
 
-File2:
+In other files, it is not necessary to pass around any instances.
+It should be able to just call `getInstance()` everywhere in your extension as needed.
+
+---
+
+`renderer/extension.ts`:
```
stefcameron: Same here, but `renderer.ts` instead...

docs/extensions/guides/ipc.md
```diff
@@ -40,57 +40,69 @@ The general terminology is as follows:
...
-We require that extension developers provide a verification function when registering a listener or handler.
-This is done as a preventative measure to help separate issues that can happen at runtime.
-While it is possible to use the unary truth function, this is highly discouraged.
-
-The verification function should do some cursory validation on the values send along the channel.
-Your handler or listener will not be called if it fails this validation.
-Instead an error or log message will occur.
-This should help with debugging because you are notified immediately that there is a mismatch between what you are expecting and what was sent.
+`Store.RendererIpcStore.broadcastIpc(...)` and `Store.MainIpcStore.broadcastIpc(...)` sends an event to all renderer frames and to main.
+Because of this, if you have the same channel name in both `main` and `renderer` both will receive the events.
```
stefcameron: Do you mean that if I declare `listenIpc('initialize', ...)` in `main`, and from `main`, I call `broadcastIpc('initialize', ...)`, my `main` listener will get called?

Nokel81: Yes

Nachasic (dismissed):

docs/extensions/guides/ipc.md
Nokel81: Okay will do.

docs/extensions/guides/ipc.md
stefcameron: Then maybe clarify this by changing the sentence to something like, "Because of this, no matter where you broadcast from, all listeners in `main` and `renderer` will be notified."? That seems clearer to me.

Nokel81: Sure sounds good.

src/extensions/ipc-main-store.ts
```diff
@@ -0,0 +1,25 @@
+import { ipcMain } from "electron";
+import { IpcPrefix, IpcStore } from "./ipc-store";
+import { Disposers } from "./lens-extension";
+import { LensMainExtension } from "./lens-main-extension";
+
+export class MainIpcStore extends IpcStore {
```
jakolehm: This and `RendererIpcStore` needs to be abstract classes.

jakolehm: Also file name should match with class name 🙂

docs/extensions/guides/ipc.md
```diff
@@ -0,0 +1,131 @@
...
+## Registering IPC Handlers and Listeners
+
+The general terminology is as follows:
+
+- A "handler" is the function that responds to a "Request Based IPC" event.
+- A "listener" is the function that is called when a "Event Based IPC" event is emitted.
+
+To register either a handler or a listener, you should do something like the following:
+
+`main/extension.ts`:
```
stefcameron: Just `main.ts`, right...? 😄

Nokel81: What? I thought I changed that, sorry. Yes.

docs/extensions/guides/ipc.md
```diff
@@ -0,0 +1,131 @@
...
+```typescript
+import { Store } from "@k8slens/extensions";
+
+export class IpcMain extends Store.MainIpcStore {}
+
+function onInitialize(event: Types.IpcMainEvent, id: string) {
+  console.log(`starting to initialize: ${id}`);
+}
+
+export function registerListeners() {
```
jakolehm: These functions could be also within `IpcMain`, not sure which one is more fluent for a reader.

stefcameron (dismissed):
This is looking good to me. Thank you! As soon as it's available in the next `alpha`, I'll start playing with it.

jakolehm (dismissed):
LGTM. I think this will help extension builders a lot 👍

docs/extensions/guides/ipc.md
Nokel81: Oh you mean something like the following?

```ts
export class IpcMain extends Store.MainIpcStore {
  constructor() {
    super();

    this.listenIpc("initialize", onInitialize);
  }
}
```

stefcameron:
@jakolehm 

> LGTM. I think this will help extension builders a lot 👍

Yes, and I think you also ought to add a new docs page dedicated to explaining this concept of `main` vs `renderer`.

When @Nokel81 explained it to me using "backend" and "frontend" analogies, it made a lot of sense. Now I think I have a much clearer mental model of how Lens expects extensions to be architected.

Because Lens is a native app, and all code is executed within the context of Lens, I don't think this is a concept that extension devs will naturally figure out just from the nature of the extensions API as it is -- especially since there are a number of singleton classes that marshal things behind the scenes between `main` and `renderer`, which means as an extension dev, I'm unaware of this separation.

Yet from Lens' perspective, there's always been a clear separation, and now that the Catalog API is _only_ available on `main`, that separation has suddenly become more apparent in the API when it shouldn't have been "sudden"...

docs/extensions/guides/ipc.md
jakolehm: Yes, something like that.

Nokel81: I think that is indeed better.

jakolehm:
> @jakolehm
> 
> > LGTM. I think this will help extension builders a lot 👍
> 
> Yes, and I think you also ought to add a new docs page dedicated to explaining this concept of `main` vs `renderer`.
> 

@stefcameron I agree, Lens docs does not currently explain the process model Lens uses (which comes from Electron). We need to improve docs also on this regard.

docs/extensions/guides/ipc.md
stefcameron: Seems more natural.

Nokel81: Changed.

jakolehm (approved):

Nachasic (approved):

src/common/ipc/type-enforced-ipc.ts
```diff
@@ -68,25 +70,53 @@ export function onceCorrect<
...
+  Handler extends (event: Electron.IpcMainInvokeEvent, ...args: any[]) => any,
+>({
+  channel,
+  handler,
+  verifier,
+}: {
+  channel: string,
+  handler: Handler,
+  verifier: ListVerifier<Rest<Parameters<Handler>>>,
+}): Disposer {
```
nevalla, line 110: this part is a bit hard to follow. Could this be defined type instead?
```
{
  channel: string,
  handler: Handler,
  verifier: ListVerifier<Rest<Parameters<Handler>>> 
}
```

src/extensions/lens-extension.ts
```diff
@@ -73,23 +81,20 @@ export class LensExtension {
...
-  get description() {
-    return this.manifest.description;
-  }
-
   @action
   async enable() {
     if (this.isEnabled) return;
     this.isEnabled = true;
-    this.onActivate();
+    this.onActivate?.();
```
nevalla, line 88: why `?.()` is needed?

nevalla (approved):
Added couple of minor questions. Good job with docs!

src/common/ipc/type-enforced-ipc.ts
Nokel81: Sure

src/extensions/lens-extension.ts
Nokel81: Haha oops, I had thought about making those methods be optional fields instead but then reverted it. I guess I forgot to revert this too

src/common/ipc/type-enforced-ipc.ts
stefcameron: Didn't we agree to remove the verifier concept??

stefcameron: I thought I saw a commit to this PR that removed it. 🤔

Nokel81: I did, this function is an internal one

stefcameron: Ah, OK. 👍

src/extensions/core-api/stores.ts
```diff
@@ -20,3 +20,5 @@
  */
 
 export { ExtensionStore } from "../extension-store";
+export { MainIpcStore } from "../main-ipc-store";
```
jakolehm, line 23: @Nokel81 somehow I missed this during reviews. Store is probably a wrong place for `Ipc` related things. Also class names indicate that these are stores (which they are not).

docs/extensions/guides/ipc.md
```diff
@@ -0,0 +1,131 @@
...
+
+It is also needed to create an instance to broadcast messages too.
+
+---
+
+`helpers/renderer.ts`:
+```typescript
+import { Store } from "@k8slens/extensions";
+
+export class IpcMain extends Store.RendererIpcStore {}
```
nevalla, line 104: Should be `IpcRenderer`

Nokel81: Yes, I'll make a new PR to fix these names.

src/extensions/core-api/stores.ts
Nokel81: Yes, I'll make a new PR to fix these names.

docs/extensions/guides/ipc.md
stefcameron: Can you link the new PR here please so we know what it is? Will these be more breaking changes in the next beta?

Nokel81: @stefcameron sure I'll ping you when I put it up. Yes it will be a breaking change, but it should be as simple as changing your side in this one place to be `extends Communication.IpcRenderer` or something similar.

stefcameron: Sounds good, thanks!

PR #2776, opened 2021-05-14
Add license header to more files

Nokel81:

I noticed that these files were missing the license header from #2763

### DIFF
--- a/src/jest.setup.ts
+++ b/src/jest.setup.ts
@@ -1,3 +1,24 @@
+/**
+ * Copyright (c) 2021 OpenLens Authors
+ *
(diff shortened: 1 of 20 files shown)


src/jest.setup.ts
```diff
@@ -1,3 +1,24 @@
+/**
+ * Copyright (c) 2021 Google LLC
```
jim-docker: Google?

Nokel81: Oh dear I botched that line, thanks

jim-docker (approved):

jakolehm (approved):

PR #2777, opened 2021-05-14
Turn on `noUnusedLocals`, `noImplicitReturns`, `importsNotUsedAsValues: error`, and `isolatedModules`

Nokel81:

These options should help keep our repo cleaner.


src/renderer/api/endpoints/metrics.api.ts
```diff
@@ -132,11 +132,11 @@ export function normalizeMetrics(metrics: IMetrics, frames = 60): IMetrics {
   return metrics;
 }
 
-export function isMetricsEmpty(metrics: { [key: string]: IMetrics }) {
+export function isMetricsEmpty(metrics: Record<string, IMetrics>) {
   return Object.values(metrics).every(metric => !metric?.data?.result?.length);
 }
 
-export function getItemMetrics(metrics: { [key: string]: IMetrics }, itemName: string): { [key: string]: IMetrics } {
+export function getItemMetrics(metrics: Record<string, IMetrics>, itemName: string): Record<string, IMetrics> | void {
```
jim-docker, line 139: why `void`? Why not `null`? (just trying to understand)

src/renderer/components/button/button.tsx
```diff
@@ -60,7 +57,7 @@ export class Button extends React.PureComponent<ButtonProps, {}> {
     // render as link
     if (this.props.href) {
       return (
-        <a {...btnProps} ref={e => this.link = e}>
+        <a {...btnProps}>
```
jim-docker, line 60: I guess this wasn't doing anything but should the intended behaviour still be added another way?

src/renderer/components/dock/dock-tabs.tsx
```diff
@@ -41,21 +36,22 @@ interface Props {
...
+      case TabKind.EDIT_RESOURCE:
+        return <DockTab value={tab} icon="edit" />;
+      case TabKind.INSTALL_CHART:
+      case TabKind.UPGRADE_CHART:
+        return <DockTab value={tab} icon={<Icon svg="install" />} />;
+      case TabKind.POD_LOGS:
+        return <DockTab value={tab} icon="subject" />;
+      case TabKind.TERMINAL:
+        return <TerminalTab value={tab} />;
     }
```
jim-docker, line 55: should there be a `default: return null;`?

src/renderer/components/dock/install-chart.store.ts
```diff
@@ -48,15 +48,15 @@ export class InstallChartStore extends DockTabStore<IChartInstallData> {
     autorun(() => {
       const { selectedTab, isOpen } = dockStore;
 
-      if (isInstallChartTab(selectedTab) && isOpen) {
-        this.loadData()
+      if (selectedTab.kind === TabKind.TERMINAL && isOpen) {
```
jim-docker:
```suggestion
      if (selectedTab.kind === TabKind.INSTALL_CHART && isOpen) {
```

src/renderer/components/dock/log.store.ts
```diff
@@ -45,10 +45,12 @@ export class LogStore {
     autorun(() => {
       const { selectedTab, isOpen } = dockStore;
 
-      if (isLogsTab(selectedTab) && isOpen) {
-        this.refresher.start();
-      } else {
-        this.refresher.stop();
+      if (selectedTab.kind === TabKind.TERMINAL) {
```
jim-docker:
```suggestion
      if (selectedTab.kind === TabKind.POD_LOGS) {
```

src/renderer/components/dock/log.store.ts
```diff
@@ -45,10 +45,12 @@ export class LogStore {
...
+        if (isOpen) {
+          this.refresher.start();
+        } else {
+          this.refresher.stop();
```
jim-docker: If you don't want to change behaviour this needs to be called for all other tab kinds too.

tsconfig.json
```diff
@@ -13,14 +13,15 @@
     "sourceMap": true,
     "strict": false,
```
jim-docker, line 14: one day?

src/renderer/api/endpoints/metrics.api.ts
Nokel81: That was mostly to have the smallest diff possible. This code is being changed in #2707 anyway.

src/renderer/components/button/button.tsx
Nokel81: I don't know what the intended behaviour way, and those fields were private anyway so it is not like they could have been accessed anyway else.

src/renderer/components/dock/dock-tabs.tsx
Nokel81: Because we are limiting it to an `enum` then typescript considers this fully covered. Either way it will fall through, return `undefined` and then go on to not render anything.

So I think it is fine.

tsconfig.json
Nokel81: Moving there, hopefully.

src/renderer/components/dock/log.store.ts
Nokel81: Good catch, thanks.

jim-docker (dismissed):

jakolehm (dismissed):
LGTM. I think we should also enable [isolatedModules](https://www.typescriptlang.org/tsconfig#isolatedModules) (in addition to `importsNotUsedAsValues`).

Nokel81:
@jakolehm and @jim-docker PTALA

jim-docker (dismissed):

jim-docker (dismissed):

jakolehm (dismissed):
LGTM but we probably should merge this after next beta is released.

jim-docker (approved):

nevalla (approved):

PR #2787, opened 2021-05-17
Add notification in renderer if lens protocol handler fails to find any routes

### LINKED ISSUE #2747: Exception Handling for Protocol Handler
When a Lens URL is not correct have a way to hand the exceptions

**Why is this needed**:
Currently Lens comes to the foreground but nothing happens. The user could be confused

Nokel81:

fixes #2747 

Some screenshots:
Screen Shot 2021-05-17 at 12 15 57 PM
Screen Shot 2021-05-17 at 12 16 14 PM
Screen Shot 2021-05-17 at 12 28 04 PM


src/renderer/protocol-handler/app-handlers.tsx
```diff
@@ -30,14 +31,25 @@ import { entitySettingsURL } from "../components/+entity-settings";
...
     .getInstance()
     .addInternalHandler("/preferences", ({ search: { highlight }}) => {
       navigate(preferencesURL({ fragment: highlight }));
     })
-    .addInternalHandler("/", () => {
+    .addInternalHandler("/", ({ tail }) => {
+      if (tail) {
+        Notifications.info((
+          <p>
+            Missing handler for <code>lens://app/{tail}</code>. Are you on the latest version?
```
jim-docker: Handlers might be removed in a "latest" version too, right? "correct" might be better than "latest"

jim-docker (commented):
There's no notification if I specify the wrong cluster id, like
```
open lens://app/cluster/minikube
```
but it is reported in the logs:
```
[0] info: [PROTOCOL ROUTER]: routing lens://app/cluster/minikube
[0] info: [PROTOCOL ROUTER]: No handler found {"url":"lens://app/cluster/minikube"}
```

Nokel81:
Those logs I assume are in `main`, so that is expected.

As for the renderer side, sure good idea.

jim-docker:
> As for the renderer side, sure good idea.

I guess it's not actually a failure to find a route, but from the user's perspective it's the same thing: no navigation, and no failure notification.

Nokel81:
I think you are confused by the log line. What it is reporting is that it cannot find a handler for the route `lens://app/cluster/minikube` on the `main` side.

On the `renderer` side it does find a handler and you should see a log line there saying: 

```
[APP-HANDLER]: cluster with given ID does not exist { clusterId: "minikube" }
```

jim-docker:
Yes, see above

src/renderer/protocol-handler/app-handlers.tsx
jim-docker: This looks like a log message. The user probably doesn't know/care about handlers, just that the action couldn't be completed.

src/renderer/protocol-handler/router.tsx
```diff
@@ -0,0 +1,121 @@
...
+      source: ipcRenderer,
+      verifier: verifyIpcArgs,
+      listener: async (event, rawUrl, mainAttemptResult) => {
+        const rendererAttempt = await this._routeToExtension(new Url(rawUrl, true));
+
+        switch (foldAttemptResults(mainAttemptResult, rendererAttempt)) {
+          case RouteAttempt.MISSING:
+            Notifications.info((
+              <p>
+                Missing handler for <code>{rawUrl}</code>. Are you on the latest version of the extensions?
```
jim-docker: Do you mean
```suggestion
                Missing handler for <code>{rawUrl}</code>. Are you on the latest version of the extension?
```

src/main/catalog-sources/kubeconfig-sync.ts
```diff
@@ -68,28 +67,26 @@ export class KubeconfigSyncManager extends Singleton {
...
+        for (const filePath of this.sources.keys()) {
+          this.stopOldSync(filePath);
+        }
 
-    catalogEntityRegistry.removeSource(KubeconfigSyncManager.syncName);
-    this.syncing = false;
+        catalogEntityRegistry.removeSource(KubeconfigSyncManager.syncName);
+        this.syncing = false;
+      }
+    );
```
jim-docker: Are these changes related to protocol handler error notification?

src/main/cluster-manager.ts
```diff
@@ -35,29 +35,36 @@ export class ClusterManager extends Singleton {
...
+      }),
+      () => {
+        for (const cluster of ClusterStore.getInstance().clusters.values()) {
+          cluster.disconnect();
+        }
       }
-    }, {
-      delay: 250
-    });
+    );
```
jim-docker: And this? Maybe you should've done an "Add `disposers()` to Singleton" PR first. Can you add this to the PR title/description?

src/main/exit-app.ts
```diff
@@ -26,19 +26,11 @@ import { ClusterManager } from "./cluster-manager";
...
-  clusterManager?.stop();
+
+  WindowManager.resetInstance();
+  ClusterManager.resetInstance();
   logger.info("SERVICE:QUIT");
-  setTimeout(() => {
-    app.exit();
-  }, 1000);
+
+  setTimeout(() => app.exit(), 1000);
```
jim-docker: And this?

src/renderer/protocol-handler/app-handlers.tsx
Nokel81: How about "Unknown action"?

src/main/cluster-manager.ts
Nokel81: Probably a good idea, will do that instead.

src/renderer/protocol-handler/app-handlers.tsx
Nokel81: Changed to "Unknown action"

jim-docker: Should you change it here too?

jim-docker (commented):
Were you going to add a notification when the cluster id can't be found?
i.e. for 
```
open lens://app/cluster/minkube
```

src/renderer/components/notifications/notifications.tsx
```diff
@@ -37,20 +37,27 @@ export class Notifications extends React.Component {
   static ok(message: NotificationMessage) {
     notificationsStore.add({
       message,
-      timeout: 2500,
+      timeout: 2_500,
```
Nachasic, line 40: There is an ESLint plugin that adds a `numeric-separators-style` rule that could be useful, I think
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/numeric-separators-style.md

Nachasic (dismissed):
LGMT, a couple of minor comments

src/renderer/components/notifications/notifications.tsx
Nokel81: I think that should be a separate PR.

jim-docker (approved):

src/main/index.ts
```diff
@@ -93,8 +93,11 @@ if (!app.requestSingleInstanceLock()) {
 
   for (const arg of process.argv) {
     if (arg.toLowerCase().startsWith("lens://")) {
-      lprm.route(arg)
-        .catch(error => logger.error(`${LensProtocolRouterMain.LoggingPrefix}: an error occured`, { error, rawUrl: arg }));
+      try {
```
jakolehm, line 96: Should we refactor this logic to a function? I'm seeing exactly the same code on lines 108-116.

jakolehm (approved):

src/main/index.ts
Nokel81: Sure, I'll do that in a follow up PR.

PR #2790, opened 2021-05-18
Show active item in hotbar & allow to pin/unpin

### LINKED ISSUE #2791: Hotbar icon active border & status led are not visible
Broken in master branch.

image

jakolehm:
Shows active catalog item in hotbar even if it's not pinned there. This mimics similar behaviour of MacOS dock / Windows taskbar. Current way of not-showing active item (unless pinned) seems to be very confusing for most users.

https://user-images.githubusercontent.com/1446224/118625431-af7aae00-b7d2-11eb-8fee-8e6135257825.mov

Also fixes #2791

### DIFF
--- a/src/common/hotbar-store.ts
+++ b/src/common/hotbar-store.ts
@@ -23,8 +23,8 @@ import { action, comparer, observable, toJS } from "mobx";
 import { BaseStore } from "./base-store";
 import migrations from "../migrations/hotbar-store";
 import * as uuid from "uuid";
-import { CatalogEntityItem } from "../renderer/components/+catalog/catalog-entity.store";
 import isNull from "lodash/isNull";
+import { CatalogEntity } from "./catalog";
 
 export interface HotbarItem {
   entity: {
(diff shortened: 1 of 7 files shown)


src/common/hotbar-store.ts
```diff
@@ -23,8 +23,8 @@ import { action, comparer, observable, toJS } from "mobx";
 import { BaseStore } from "./base-store";
 import migrations from "../migrations/hotbar-store";
 import * as uuid from "uuid";
-import { CatalogEntityItem } from "../renderer/components/+catalog/catalog-entity.store";
```
jakolehm, line 26: Why changed: it's not good practice to import from renderer to common. This change also simplifies things (see tests).

nevalla (approved):

src/renderer/components/hotbar/hotbar-menu.tsx
```diff
@@ -151,7 +181,7 @@ export class HotbarMenu extends React.Component<Props> {
     return (
       <div className={cssNames("HotbarMenu flex column", className)}>
         <div className="HotbarItems flex column gaps">
-          <DragDropContext onDragEnd={this.onDragEnd}>
+          <DragDropContext onDragEnd={this.onDragEnd.bind(this)}>
```
ixrock, line 184: nit: I'd better to use arrow-function for all event-handlers by default (or `@boundMethod`, but it would be available only after merging `mobx6` migration branch)

ixrock (approved):

PR #2796, opened 2021-05-18
Restyling extensions page with tailwindcss

aleksfront:
Refreshing extensions page look along with using `tailwind` css framework and `react-table`.

### Extensions page
Using consistent way of showing listed data.

extensions page
extensions light theme

https://user-images.githubusercontent.com/9607060/118618055-e7cabe00-b7cb-11eb-8ec1-0bee84afe53e.mp4

https://user-images.githubusercontent.com/9607060/118618063-e9948180-b7cb-11eb-8520-9c270f64998c.mp4

### Tailwind framework and CSS Modules
Implementing [tailwindcss](https://tailwindcss.com/) framework and [CSS Modules](https://github.com/css-modules/css-modules) for styling purposes. 

By using this libraries we solve most common problems we have in styling codebase such as **classname conflicts and isolating styles**

**Tailwindcss pros**
It allows us to dramatically decrease amount of css by using short classnames along with `@apply` keyword.
Inline example:
```
<h3 className="font-medium text-3xl mt-5 mb-2">
  There are no extensions installed.
</h3>
```
or css-file example
```
.notFound {
  @apply flex items-center justify-center h-24;
  color: var(--textColorDimmed);
}
```

Tailwindcss has good support inside VSCode and Webstorm.
tailwind intellisense

**CSS Modules pros**
No more classname conflicts - every classname became unique in the compile time.
```
import styles from "./react-table.module.css";
...
<div className={styles.thead}></div>
```

CSS Modules also have good support inside VSCode and Webstorm.
css modules intellisense

### react-table
Introducing first draft of `react-table` component we will use to replace our native one. No virtual support yet.


src/renderer/components/+extensions/install.tsx
```diff
@@ -0,0 +1,100 @@
...
+  return (
+    <section className="mt-2">
+      <SubTitle title={`Path or URL to an extension package (${supportedFormats.join(", ")})`}/>
+      <div className="flex">
+        <div className="flex-1">
+          <Input
+            className="box grow mr-6"
+            theme="round-black"
+            disabled={ExtensionInstallationStateStore.anyPreInstallingOrInstalling}
+            placeholder={"File path or URL"}
```
Nokel81: This is incorrect

src/renderer/components/list/list.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { Table } from "../table/react-table";
+export type SearchFilter = (item: object) => string | number;
+
+interface Props extends UseTableOptions<any> {
+  items: object[];
+  filters: SearchFilter[];
+  title?: React.ReactNode;
+}
+
+export function List({ columns, data, title, items, filters, }: Props) {
```
Nokel81, line 36: Is this the newer for of `ItemListLayout` that you have come up with?

types/react-table.config.d.ts
```diff
@@ -0,0 +1,141 @@
+/**
```
Nokel81, line 1: Are there no public types for this?

Nokel81 (commented):
Why do we want/need `src/renderer/fonts/MaterialIcons-Regular.ttf`?

The svgs should be good enough.

src/renderer/components/+extensions/extension-install.store.ts
```diff
@@ -209,13 +209,27 @@ export class ExtensionInstallationStateStore {
     return ExtensionInstallationStateStore.InstallingExtensions.size;
   }
 
+  /**
+   * The current number of extensions uninstalling
+   */
+   @computed static get uninstalling(): number {
```
ixrock: Please don't use any decorators with `static` (won't work in `mobx-6`)
See all migration pitfalls in https://github.com/lensapp/lens/pull/2718

src/extensions/extension-loader.ts
```diff
@@ -86,6 +86,10 @@ export class ExtensionLoader extends Singleton {
     return extensions;
   }
 
+  @computed get userExtensionList() {
```
ixrock: Why do we need `computed` for `computed` where actually nothing to compute? :)
Better to change to `@computed get userExtensions(): InstalledExtension[]` everywhere.

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -20,33 +20,47 @@
...
+  Disposer,
+  disposer,
+  downloadFile,
+  downloadJson,
+  ExtendableDisposer,
+  extractTar,
+  listTarEntries,
+  noop,
+  readFileFromTar,
+} from "../../../common/utils";
```
ixrock, line 45: Is it some auto-formating from vscode?

jim-docker (commented):
When uninstalling the last extension:

https://user-images.githubusercontent.com/40840436/118667125-b816b880-b7c1-11eb-8be8-b05e046b00b3.mov

Also, maybe related, there's a `0` displayed under the Pro tip when there are no extensions:

Screen Shot 2021-05-18 at 10 09 41 AM

src/renderer/components/+extensions/installed-extensions.tsx
```diff
@@ -0,0 +1,169 @@
...
+    (extension: InstalledExtension) => extension.manifest.version,
+  ];
+
+  const columns: Column<any>[] = useMemo(
+    () => [
+      {
+        Header: "Name",
+        accessor: "extension",
+        width: 200,
+        sortType: (rowA: Row, rowB: Row) => { // Custom sorting for extension name
```
ixrock, line 59: Maybe better to use `lodash/orderBy`?

src/renderer/components/app.tsx
```diff
@@ -18,7 +18,7 @@
  * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
-
+import "tailwindcss/tailwind.css";
```
ixrock: Why not imported into `app.scss`?

src/renderer/components/table/react-table.module.css
```diff
@@ -0,0 +1,43 @@
+.table {
```
ixrock, line 1: This file named with `.module.css` ending for some reason?

src/renderer/components/table/react-table.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { useCallback, useMemo } from "react";
+import { useFlexLayout, useSortBy, useTable, UseTableOptions } from "react-table";
+import { Icon } from "../icon";
+import { cssNames } from "../../utils";
+
+interface Props extends UseTableOptions<any> {
+  headless?: boolean;
+}
+
+export function Table({ columns, data, headless }: Props) {
```
ixrock: Please align exporting component name with filename (should be `ReactTable` in that case).

types/react-table.config.d.ts
ixrock: Also wondering why `@types/react-table` not enough.

src/renderer/components/table/react-table.module.css
ixrock: Ok, I've seen some notice in `mocks.d.ts` but.. why the file extension not `.scss`?

aleksfront:
> Why do we want/need `src/renderer/fonts/MaterialIcons-Regular.ttf`?
> 
> The svgs should be good enough.

We already had this file. In this PR I've updated it to new version. We use it for material icons.

src/renderer/components/+extensions/install.tsx
aleksfront: Why? You can paste local file path or URL to remote resource.

src/renderer/components/list/list.tsx
aleksfront: Yes, correct. Eventually it will replace `ItemListLayout`.

types/react-table.config.d.ts
aleksfront: The biggest pitfall of `react-table` is separated types. The author doesn't provided them on his own, so even latest `@types/react-table` not always correlate with available codebase. In my specific case some of the types related to `useSortBy` hook was missing so typescript won't allow to compile actually proper code.

Contents inside `react-table.config.d.ts` comes from community fixing the issue.

src/renderer/components/+extensions/extensions.tsx
aleksfront: In particular, `TS Hero` plugin 😃

aleksfront: Just checked, native "Organize imports" does about the same thing.

src/extensions/extension-loader.ts
aleksfront: We have usage of `userExtensions.get(id)`, `userExtensions.has(id)` in `extensions.tsx`. Not sure if we should replace this with array-like behavior.

src/renderer/components/+extensions/installed-extensions.tsx
aleksfront: No, `lodash/sortBy` will actually sort array, but we need here to provide sorting function instead. The reason for providing it is need to fetch extension names and compare them. Comparator function is using basic algorithm for string sorting https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#description

src/renderer/components/table/react-table.module.css
aleksfront: It is possible to use with `.scss` but this way not working in current PR. I was intentionally went thought css-only way to test styling without use of SCSS and check the possibility to remove it someday. I guess the main benefit of using scss nowadays is nested properties. But now we have [postcss nested plugin](https://github.com/postcss/postcss-nested) which can be used to replace this feature (not implemented also in this PR).

aleksfront:
@jim-docker Thanks for noticing bug with app crash. The reason was violating rules of hooks https://reactjs.org/docs/hooks-rules.html. Order of hooks was mixed caused by conditions. Probably need to set eslint rules for that.

removing successfull

src/extensions/extension-loader.ts
ixrock: How about extra method `public getById(id: string): LensExtension`?

src/renderer/components/+extensions/installed-extensions.tsx
ixrock: Second argument supports callback: https://lodash.com/docs/4.17.15#orderBy

src/renderer/components/+extensions/install.tsx
Nokel81: Or a name of an extension, did you ignore the work that I did in https://github.com/lensapp/lens/pull/2221 ? I can see the old text in the diff.

types/react-table.config.d.ts
Nokel81: Shouldn't the community then push a new version to `@types/react-table`? I have submitted several fixes to that repo and they tend to be merged very quickly.

src/renderer/components/app.tsx
aleksfront: Moved into `app.scss` but it requires 3 lines instead of 1:
```
@import "tailwindcss/base";
@import "tailwindcss/components";
@import "tailwindcss/utilities";
@import "~flex.box";
@import "fonts";
```

src/renderer/components/+extensions/installed-extensions.tsx
aleksfront: We are providing a callback here for `react-table` internal sorting. I still don't see how lodash can be used here. We're not sorting any values here, we just telling `react-table` how to do it.

src/extensions/extension-loader.ts
aleksfront: One thing touches another, there's mix of `extensions`, `userExtensions`, instances etc. There's need to refactor `extension-loader.ts` carefully. For this PR i've removed this `userExtensionList()` method for now.

src/renderer/components/+extensions/installed-extensions.tsx
```diff
@@ -0,0 +1,169 @@
...
+        Header: "",
+        accessor: "actions",
+        disableSortBy: true,
+        width: 20,
+        className: "actions"
+      }
+    ], []
+  );
+
+  const data = useMemo(
```
chenhunghan, line 87: [nit] Can be a separate React component with extensions as props?

src/renderer/components/table/react-table.tsx
```diff
@@ -0,0 +1,109 @@
...
+}
+
+export function ReactTable({ columns, data, headless }: Props) {
+  const defaultColumn = useMemo(
+    () => ({
+      minWidth: 20,
+      width: 100,
+    }),
+    []
+  );
```
chenhunghan, line 40: [nit] I think it's not needed to wrapped with `useMemo()` (as they are static value in every re-render) but no harm if wrapped.
```suggestion
  const defaultColumn = {
      minWidth: 20,
      width: 100,
  };
```

chenhunghan (dismissed):

src/renderer/components/app.tsx
ixrock: Why? You can import `.css` intro `.scss` afaik.. (just explicitly add file extension)

Nokel81 (dismissed):
This PR regresses some of the wording of the extension's settings page.

src/renderer/components/table/react-table.tsx
aleksfront: Yes, I guess it can be removed. I've used the "recommendations" from the `react-table` author. There are many things memorized in there https://react-table.tanstack.com/docs/examples/virtualized-rows

aleksfront:
> This PR regresses some of the wording of the extension's settings page.

Reverted contents of previous input placeholder and title above it.

extensions

chenhunghan (approved):

types/react-table.config.d.ts
ixrock: I would consider to use [material-table](https://github.com/mbrn/material-table) instead `react-table`. Wdyt?

src/renderer/components/table/react-table.module.css
ixrock: Could you remind me why we need `postcss` at all?

ixrock (approved):
If everyones in the team happy with `react-table` then this is probably fine.

src/renderer/components/table/react-table.module.css
aleksfront: It is used by `tailwind` for cases when styling goes into separate files `*.css`. We using `@apply` directive in there which is compiled by `postcss` into regular css... As you remember we don't have `@apply` keyword in raw css 😃 .

aleksfront:
> Reverted, still need to check if it works.

@Nokel81 I didn't change any extension install logic in this PR, but will double check this.

Nokel81:
> > Reverted, still need to check if it works.
> 
> @Nokel81 I didn't change any extension install logic in this PR, but will double check this.

I meant that I hadn't yet pulled this branch to check it out.

jim-docker:
File dialog does not come up when I click on the folder icon

Nokel81:
Except for the issue that @jim-docker found (did verify that it wasn't working too). This looks good.

Nokel81:
Thanks for starting the epic of using `react-table` too!

aleksfront:
@Nokel81 Install from registry works okay.

https://user-images.githubusercontent.com/9607060/119260044-b0984a80-bbd9-11eb-84ae-cb9e0f70c2a8.mov

types/react-table.config.d.ts
jakolehm: Local copy of types feels bad. And it also feels bad that `@types/react-table` is not up to date. Did we consider any alternatives?

PR #2806, opened 2021-05-19
Include all files as resources from extension dist folder

jakolehm:
An extension might have other build artifacts than just yml/js.

Also removes legacy "locales" configuration from `build.extraResources`.


nevalla (approved):

jakolehm:
Replaced by #2807

PR #2812, opened 2021-05-19
Typed catalog category events

jakolehm:
- adds typed event emitter to `CatalogCategory`
- adds new event, `onLoad`, which can be used to refresh/reload catalog sources


src/renderer/components/+catalog/catalog-add-button.tsx
```diff
@@ -50,7 +50,7 @@ export class CatalogAddButton extends React.Component<CatalogAddButtonProps> {
             menuItems: this.menuItems
           };
 
-          category.emit("onCatalogAddMenu", context);
+          category.onCatalogAddMenu(context);
```
chenhunghan: This seems to return a promise. Not sure it's a problem or not.

chenhunghan (dismissed):

src/common/catalog/catalog-entity.ts
```diff
@@ -59,6 +66,18 @@ export abstract class CatalogCategory extends EventEmitter {
   public getId(): string {
     return `${this.spec.group}/${this.spec.names.kind}`;
   }
+
+  public async onLoad(): Promise<void> {
+    for( const listener of this.listeners("onLoad")) {
```
chenhunghan: [nit] if linter is happy with this extra space, I am fine too.
```suggestion
    for(const listener of this.listeners("onLoad")) {
```

src/common/catalog/catalog-entity.ts
```diff
@@ -59,6 +66,18 @@ export abstract class CatalogCategory extends EventEmitter {
...
+      await listener();
+    }
+  }
+
+  public async onCatalogAddMenu(context: CatalogEntityAddMenuContext): Promise<void> {
+    for( const listener of this.listeners("onCatalogAddMenu")) {
```
chenhunghan: [nit] Extra space also
```suggestion
    for(const listener of this.listeners("onCatalogAddMenu")) {
```

src/common/catalog/catalog-entity.ts
```diff
@@ -47,7 +48,13 @@ export interface CatalogCategorySpec {
...
 }
 
-export abstract class CatalogCategory extends EventEmitter {
+export interface CatalogCategoryEvents {
+  onLoad: () => void;
+  onCatalogAddMenu: (context: CatalogEntityAddMenuContext) => void;
+  onContextMenuOpen: (entity: CatalogEntity, context: CatalogEntityContextMenuContext) => void;
+}
+
+export abstract class CatalogCategory extends (EventEmitter as new () => TypedEmitter<CatalogCategoryEvents>) {
```
Nokel81, line 57: I don't think that that an `EventEmitter` should "await" the events.

From that point of view, `onLoad` should just be emitted and should come with an `offLoad` as well. And both of those should just be "emitted".

panuhorsmalahti: I agree, Lens shouldn't wait for each extension onLoad event handler to be completed, as that could cause "slow" opening of the Catalog (if I understand this correctly). Instead we should display what we have, and then each extension can asynchronously add new/remove old catalog entities in the event handler.

jakolehm: I agree, it does not make sense to wait `onLoad` (I'll remove it). But does it make sense to wait `onContextMenuOpen`?

src/common/catalog/catalog-entity.ts
nevalla: should it be?

```suggestion
    for (const listener of this.listeners("onLoad")) {
```

src/common/catalog/catalog-entity.ts
nevalla:
```suggestion
    for (const listener of this.listeners("onCatalogAddMenu")) {
```

src/common/catalog/catalog-entity.ts
panuhorsmalahti: Good question.. I can see benefits/drawbacks for both. If we wait, then the UI can be sluggish if there is at least one extension doing some API calls to add something to the context menu. On the other hand, if we don't wait, then the context menu will open, and a moment later new stuff will be added.

Maybe ideal would be to add a "race", e.g. wait maximum of 50-100ms for the extensions to handle the context menu open event?

Nokel81: > I agree, it does not make sense to wait `onLoad` (I'll remove it). But does it make sense to wait `onContextMenuOpen`?

See https://github.com/lensapp/lens/pull/2838 for how I think this should be changed (with regards to menu handlers

package.json
```diff
@@ -354,6 +354,7 @@
     "ts-loader": "^7.0.5",
     "ts-node": "^8.10.2",
     "type-fest": "^1.0.2",
+    "typed-emitter": "^1.3.1",
```
ixrock, line 364: This package contains only `types` for standard `events`.
What if we add required types directly to our custom event-emitter in `common/utils/event-emitter.ts`?

src/common/catalog/catalog-entity.ts
jakolehm: These are now removed.

src/common/catalog/catalog-entity.ts
jakolehm: Refactored code to use just `emit`.

Nokel81 (commented):
Should we have the `componentWillUnmount` analogy too?

package.json
jakolehm: Not sure what it would look like ... could you provide an example?

stefcameron:
@jakolehm I happened to notice this PR. It seems intriguing to me, in that it sounds like something I need in order to handle a use case I'm not aware that I need to handle yet...

- How is this meant to be used in an extension?
- If it's meant to be used in an extension, it seems this PR is missing some documentation on how to use this new feature...

package.json
ixrock: Ok, I mean I don't understand purpose of extra types for `events` package when we have our custom event emitter written in Typescript (where we can add any types on purpose). 
Otherwise we could remove `common/utils/event-emitter` and use only standard with these extra-types if it works better for us (to keep usage of single emitter over whole project).

src/renderer/components/+catalog/catalog-entity.store.ts
```diff
@@ -106,6 +106,14 @@ export class CatalogEntityStore extends ItemStore<CatalogEntityItem> {
   }
 
   loadAll() {
+    if (this.activeCategory) {
+      this.activeCategory.emit("onLoad");
```
Nokel81: This reads weird when used, because the consumer would do `.on("onLoad", <listener>)`. I think it should be just `"load"`.

stefcameron: I agree

Nokel81 (approved):
This looks like a good improvement of the current design.

Nokel81:
Though the opposite, an `"unload"` event probably should also be added.

stefcameron:
I think this PR is missing documentation...

nevalla (approved):

PR #2816, opened 2021-05-19
Filter out unknown catalog entities

### LINKED ISSUE #2813: Pin Catatlog Entity to Hotbar crash the app on restart
**Describe the bug**
Pin a Catalog Entity to Hotbar causes crashes

**To Reproduce**
Pin a Catalog Entity to Hotbar

Screen Shot 2021-05-19 at 2 52 49 PM

Confirm it's in Hotbar
Screen Shot 2021-05-19 at 2 54 49 PM

Restart Lens.

See this
Screen Shot 2021-05-19 at 2 57 40 PM

In console
Screen Shot 2021-05-19 at 2 58 41 PM

Above related to 
https://github.com/lensapp/lens/blob/master/src/renderer/components/hotbar/hotbar-menu.tsx#L55

If I only fix `hotbar-menu.tsx`

I will see this error in console instead

Screen Shot 2021-05-19 at 3 11 26 PM

This need fix too
https://github.com/lensapp/lens/blob/master/src/renderer/api/catalog-entity-registry.ts#L68

**Expected behavior**

No crashes

Nokel81:
- Keep raw data
- But filter unknown types until a category is registered
- Fixes a race condition.

### DIFF
--- a/src/common/catalog/catalog-category-registry.ts
+++ b/src/common/catalog/catalog-category-registry.ts
@@ -20,7 +20,7 @@
  */
 
 import { action, computed, observable } from "mobx";
-import { Disposer, ExtendedMap, StrictMap } from "../utils";
+import { Disposer, ExtendedMap } from "../utils";
 import { CatalogCategory, CatalogEntityData, CatalogEntityKindData } from "./catalog-entity";
 
 export class CatalogCategoryRegistry {
@@ -33,10 +33,13 @@ export class CatalogCategoryRegistry {
   }
 
   @computed private get groupKindLookup(): Map<string, Map<string, CatalogCategory>> {
-    const res = new ExtendedMap<string, StrictMap<string, CatalogCategory>>(StrictMap.new);
+    // ExtendedMap has the convenience methods `getOrInsert` and `strictSet`
+    const res = new ExtendedMap<string, ExtendedMap<string, CatalogCategory>>();
 
     for (const category of this.categories) {
-      res.getOrDefault(category.spec.group).strictSet(category.spec.names.kind, category);
+      res
+        .getOrInsert(category.spec.group, ExtendedMap.new)
+        .strictSet(category.spec.names.kind, category);
     }
 
     return res;
--- a/src/common/catalog/catalog-category-registry.ts
+++ b/src/common/catalog/catalog-category-registry.ts
@@ -19,26 +19,35 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { action, computed, observable, toJS } from "mobx";
+import { action, computed, observable } from "mobx";
+import { Disposer, ExtendedMap, StrictMap } from "../utils";
 import { CatalogCategory, CatalogEntityData, CatalogEntityKindData } from "./catalog-entity";
 
 export class CatalogCategoryRegistry {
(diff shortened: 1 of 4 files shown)


chenhunghan (dismissed):
LGTM

chenhunghan (dismissed):

jim-docker (approved):

src/common/utils/extended-map.ts
```diff
@@ -22,32 +22,47 @@
...
     if (this.has(key)) {
-      return this.get(key);
+      throw new TypeError("Duplicate key in map");
     }
 
-    return this.set(key, getVal()).get(key);
+    return this.set(key, val);
   }
 
-  getOrDefault(key: K): V {
```
jakolehm, line 45: 👍

src/renderer/api/__tests__/catalog-entity-registry.test.ts
```diff
@@ -42,7 +42,7 @@ describe("CatalogEntityRegistry", () => {
         spec: {}
       }];
 
-      catalog.updateItems(items);
+      (catalog as any).rawItems.replace(items);
```
jakolehm, line 45: This feels bad, we could have testable subclass that allows replace.

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -49,23 +46,27 @@ export class CatalogEntityRegistry {
     return this._activeEntity;
   }
 
-  get items() {
-    return this._items;
+  @computed get items() {
```
jakolehm, line 49: I think we should have tests that confirm filtering (now that it starts to be even more complex).

jakolehm (commented):
This fixes problems on the receiving side (renderer) 👍 But why do we even sync "invalid" items from main to renderer?

jakolehm (approved):
I think we should merge this as is, I'll open follow-up PR for additional improvements.

src/common/catalog/catalog-category-registry.ts
```diff
@@ -19,26 +19,38 @@
...
 
-  @action add(category: CatalogCategory) {
-    this.categories.push(category);
+  @action add(category: CatalogCategory): Disposer {
+    this.categories.add(category);
+
+    return () => this.categories.delete(category);
   }
 
-  @action remove(category: CatalogCategory) {
```
jakolehm, line 32: @Nokel81  was this removed by accident?

jakolehm: Maybe not, would have been good to mention that we are breaking the api here ... 🙂

Nokel81:
> This fixes problems on the receiving side (renderer) 👍 But why do we even sync "invalid" items from main to renderer?

Because we sync everything from main to renderer and the category in renderer mighty not have been registered yet (race condition)

jakolehm:
> > This fixes problems on the receiving side (renderer) 👍 But why do we even sync "invalid" items from main to renderer?
> 
> Because we sync everything from main to renderer and the category in renderer mighty not have been registered yet (race condition)

I get that but we need also protection on `main` side. Created a follow-up pr (#2824) that should improve this.

src/common/catalog/catalog-category-registry.ts
Nokel81: Sorry, will start listing breaking API changes in my PR descriptions from now on.

PR #2820, opened 2021-05-19
Show event last-seen column in addition to age

### LINKED ISSUE #2819: Events: Last Seen column displays First Seen value
**Describe the bug**
The "Last Seen" column in the Events view on Lens currently displays the first seen value.

**To Reproduce**
1. Go to 'Events'
2. Show the "Last Seen" column
3. See incorrect value in Last Seen column

**Expected behavior**
The column should be the value of Last Seen.

**Screenshots**

- Lens Version: 4.2.4
- OS: OSX
- Installation method (e.g. snap or AppImage in Linux): Website

**Logs:**
```
N/A
```

**Kubeconfig:**
```
N/A
```

**Additional context**
N/A

Nokel81:

fixes #2819

### DIFF
--- a/src/renderer/api/endpoints/events.api.ts
+++ b/src/renderer/api/endpoints/events.api.ts
@@ -75,6 +75,10 @@ export class KubeEvent extends KubeObject {
 
     return formatDuration(diff, true);
   }
+
+  getLastSeenTimeDiffFromNow(): number {
+    return Date.now() - new Date(this.lastTimestamp).getTime();
+  }
 }
 
 export const eventApi = new KubeApi({
--- a/src/renderer/api/endpoints/events.api.ts
+++ b/src/renderer/api/endpoints/events.api.ts
@@ -75,10 +75,6 @@ export class KubeEvent extends KubeObject {
 
     return formatDuration(diff, true);
   }
-
-  getLastSeenTimeDiffFromNow(): number {
-    return Date.now() - new Date(this.lastTimestamp).getTime();
-  }
 }
 
 export const eventApi = new KubeApi({
--- a/src/renderer/components/+events/events.tsx
+++ b/src/renderer/components/+events/events.tsx
@@ -62,6 +62,7 @@ const defaultProps: Partial<Props> = {
 @observer
 export class Events extends React.Component<Props> {
   static defaultProps = defaultProps as object;
+  now = Date.now();
 
   @observable sorting: TableSortParams = {
     sortBy: columnId.age,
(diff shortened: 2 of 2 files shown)


aleksfront (dismissed):

src/renderer/components/+events/events.tsx
```diff
@@ -163,7 +163,7 @@ export class Events extends React.Component<Props> {
           { title: "Involved Object", className: "object", sortBy: columnId.object, id: columnId.object },
           { title: "Source", className: "source", id: columnId.source },
           { title: "Count", className: "count", sortBy: columnId.count, id: columnId.count },
-          { title: "Last Seen", className: "age", sortBy: columnId.age, id: columnId.age },
+          { title: "Age", className: "age", sortBy: columnId.age, id: columnId.age },
```
jakolehm, line 169: I think we should fix "Last Seen" instead of this.

jakolehm (changes requested):

src/renderer/components/+events/events.tsx
Nokel81: How about both?

jim-docker: what is meant by "Last Seen"? It can mean "when an event was last seen", which would be the age of the event. Or it can mean "show me the last event seen", so the default order would be from most recent to oldest(?) Or does it mean when a specific event was last seen, which would be an actual time value. I think "Age" makes it unambiguous.

Nokel81: @jim-docker See discussion on https://github.com/lensapp/lens/issues/2819. Some controllers increment the count and last seen time instead of creating a new one.

nevalla (dismissed):

src/renderer/api/endpoints/events.api.ts
```diff
@@ -75,6 +75,10 @@ export class KubeEvent extends KubeObject {
 
     return formatDuration(diff, true);
   }
+
+  getLastSeenTimeDiffFromNow(): number {
```
jakolehm: Just a note that these kind of helper methods should not be in the api itself.. but we already have similar method so maybe it's fine to add "one more".

jakolehm (dismissed):

src/renderer/api/endpoints/events.api.ts
Nokel81: Okay makes sense, we probably want to remove both in the future. Shall I not make it worse in this PR?

jakolehm (approved):

jim-docker (approved):

PR #2821, opened 2021-05-19
Switch to more specific IPC helpers

Nokel81:
Namely we should now have:

- `ipcMainOn` which is `ipcMain.on` + return disposer of `ipcMain.off`
- `ipcRendererOn` which is `ipcRenderer.on` + return disposer of `ipcRenderer.off`
- `ipcMain.handle` instead of `handleMain`

### DIFF
--- a/src/common/ipc/ipc.ts
+++ b/src/common/ipc/ipc.ts
@@ -26,20 +26,19 @@
 import { ipcMain, ipcRenderer, remote, webContents } from "electron";
 import { toJS } from "../utils/toJS";
 import logger from "../../main/logger";
-import { ClusterFrameInfo, clusterFrameMap } from "../cluster-frames";
+import { ClusterFrameInfo, clusterFrameMap } from "../cluster-frames";
+import type { Disposer } from "../utils";
 
 const subFramesChannel = "ipc:get-sub-frames";
 
-export function handleRequest(channel: string, listener: (event: Electron.IpcMainInvokeEvent, ...args: any[]) => any) {
-  ipcMain.handle(channel, async (event, ...args) => {
-    const payload = await listener(event, ...args);
-
-    return sanitizePayload(payload);
-  });
-}
-
 export async function requestMain(channel: string, ...args: any[]) {
   return ipcRenderer.invoke(channel, ...args.map(sanitizePayload));
+}
+
+export function ipcMainHandle(channel: string, listener: (event: Electron.IpcMainInvokeEvent, ...args: any[]) => any) {
+  ipcMain.handle(channel, async (event, ...args) => {
+    return sanitizePayload(await listener(event, ...args));
+  });
 }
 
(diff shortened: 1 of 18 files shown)


src/renderer/components/cluster-manager/cluster-status.tsx
```diff
@@ -54,7 +54,7 @@ export class ClusterStatus extends React.Component<Props> {
   }
 
   async componentDidMount() {
-    subscribeToBroadcast(`kube-auth:${this.cluster.id}`, (evt, res: KubeAuthProxyLog) => {
+    ipcRendererOn(`kube-auth:${this.cluster.id}`, (evt, res: KubeAuthProxyLog) => {
```
Nachasic, line 64: Shouldn't we clean up the subscription when the component unmounts?

src/common/ipc/ipc.ts
```diff
@@ -71,34 +68,18 @@ export function broadcastMessage(channel: string, ...args: any[]) {
...
 
-export function unsubscribeFromBroadcast(channel: string, listener: (...args: any[]) => any) {
-  if (ipcRenderer) {
-    ipcRenderer.off(channel, listener);
-  } else if (ipcMain) {
-    ipcMain.off(channel, listener);
-  }
-}
+export function ipcRendererOn(channel: string, listener: (event: Electron.IpcRendererEvent, ...args: any[]) => any): Disposer {
+  ipcRenderer.on(channel, listener);
```
jim-docker, line 85: We should still check if `ipcRenderer`, etc.  is valid in these functions?

jim-docker (dismissed):

src/common/ipc/ipc.ts
Nokel81: Oh do you mean for tests and all? Currently the checks are usually done in the preceding scopes.

jim-docker: If this function is exported to the extensions api it should be bulletproofed. Or regardless. Just because it's called `ipcRendererOn` doesn't mean it will be called from the right process.

Nokel81: This function is not directly exposed to the extensions API. See #2775

stefcameron:
Is this PR a continuation of https://github.com/lensapp/lens/pull/2775 and where names are changing again to move away from "Store"? Or is it all just internal stuff. Not clear to me from the code.

Nokel81:
@stefcameron This is just internal stuff. I haven't made the PR for the extensions code yet.

src/renderer/components/cluster-manager/cluster-status.tsx
Nachasic: Is leaving the disposer hanging here a non-issue?

Nachasic (dismissed):

Nachasic (approved):

PR #2822, opened 2021-05-19
Ban circular dependencies

Nokel81:
- ApiManager now creates the instances of all stores

- ReleaseStore is Singleton-like

- Move most types and helper functions into seperate files

Note: this currently uses webpack's `CircularDependencyPlugin` but once https://github.com/benmosher/eslint-plugin-import/pull/2083 has made its way into a release we should move to that as then we can get in-editor errors.


Nokel81:
Making this a draft since it requires some other PRs and I think I should focus on other things.

PR #2824, opened 2021-05-20
Refactor CatalogEntityRegistry from common to main

jakolehm:
Follow-up for #2816.

- moves `CatalogEntityRegistry` from `common` to `main` (it should have been there from the start).
- does `items`  filtering also on `main` (`renderer` already has filtering)
- refactors `groupKindLookup` to use cached `ExtendedMap` (because otherwise we are generating new map for each catalog item).
- more tests

### DIFF
--- a/src/common/catalog/catalog-category-registry.ts
+++ b/src/common/catalog/catalog-category-registry.ts
@@ -25,24 +25,22 @@ import { CatalogCategory, CatalogEntityData, CatalogEntityKindData } from "./cat
 
 export class CatalogCategoryRegistry {
   protected categories = observable.set<CatalogCategory>();
+  protected groupKinds = new ExtendedMap<string, ExtendedMap<string, CatalogCategory>>();
 
   @action add(category: CatalogCategory): Disposer {
     this.categories.add(category);
+    this.updateGroupKinds(category);
 
-    return () => this.categories.delete(category);
+    return () => {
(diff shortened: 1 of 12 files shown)


aleksfront (dismissed):

aleksfront (approved):

chenhunghan (approved):

src/common/catalog/catalog-category-registry.ts
```diff
@@ -25,32 +25,30 @@ import { CatalogCategory, CatalogEntityData, CatalogEntityKindData } from "./cat
...
+  protected groupKinds = new ExtendedMap<string, ExtendedMap<string, CatalogCategory>>();
 
   @action add(category: CatalogCategory): Disposer {
     this.categories.add(category);
+    this.updateGroupKinds(category);
 
-    return () => this.categories.delete(category);
+    return () => {
+      this.categories.delete(category);
+      this.groupKinds.clear();
```
Nokel81, line 36: This is wrong FYI @jakolehm  We don't want to clear the whole map.

Nokel81: The reason that I used `@computed` is because it is sort of like a computed index in a database table.

This should be:
```suggestion
      this.groupKinds.get(category.spec.group).delete(category.spec.names.kind);
```

PR #2828, opened 2021-05-20
Check if name is undefined

### LINKED ISSUE #2826: Adding Entities to Hotbar causing App Crash
**Describe the bug**
After adding / removing entities from Hotbar and moving them arround suddenly seen the following app crash.

Seem its related to the following piece of code:
https://github.com/lensapp/lens/blob/5ed4537979e05ee22727371393dcc7efda551ef4/src/renderer/components/hotbar/hotbar-icon.tsx#L70

**To Reproduce**
--> Still trying to reproduce it consistently. 

**Expected behavior**
no app crash

**Screenshots**
Bildschirmfoto 2021-05-20 um 09 25 36

- OpenLensDev: 5.0.0-beta.5 (latest master)
- Mac OS
- Installation method (e.g. snap or AppImage in Linux): self compiled from master

**Logs:**

```
Uncaught TypeError: Cannot read property 'split' of undefined
    at getNameParts (OpenLensDev.js:910824)
    at getIconString (OpenLensDev.js:910843)
    at OpenLensDev.js:910857
    at OpenLensDev.js:558701
    at OpenLensDev.js:559144
    at trackDerivedFunction (OpenLensDev.js:561002)
    at Reaction../node_modules/mobx/lib/mobx.module.js.Reaction.track (OpenLensDev.js:562036)
    at OpenLensDev.js:559142
    at useQueuedForceUpdateBlock (OpenLensDev.js:559191)
    at useObserver (OpenLensDev.js:559136)
    at OpenLensDev.js:558701
    at renderWithHooks (OpenLensDev.js:682823)
    at updateFunctionComponent (OpenLensDev.js:685194)
    at updateSimpleMemoComponent (OpenLensDev.js:685053)
    at updateMemoComponent (OpenLensDev.js:684930)
    at beginWork (OpenLensDev.js:686973)
```

chenhunghan:
Closes https://github.com/lensapp/lens/issues/2826

### DIFF
--- a/src/renderer/components/hotbar/hotbar-icon.tsx
+++ b/src/renderer/components/hotbar/hotbar-icon.tsx
@@ -67,7 +67,10 @@ function onMenuItemClick(menuItem: CatalogEntityContextMenu) {
   }
 }
 
-function getNameParts(name: string): string[] {
+function getNameParts(name?: string) {
+  if (!name) {
+    return [];
+  }
   const byWhitespace = name.split(/\s+/);
 
   if (byWhitespace.length > 1) {


src/renderer/components/hotbar/hotbar-icon.tsx
```diff
@@ -67,7 +67,10 @@ function onMenuItemClick(menuItem: CatalogEntityContextMenu) {
   }
 }
 
-function getNameParts(name: string): string[] {
+function getNameParts(name?: string) {
+  if (!name) {
+    return [];
```
chenhunghan, line 72: not sure if I should return `["", "", ""]` instead for https://github.com/lensapp/lens/blob/e97b30dae06758dc9381b551cea91d22bc7b2d08/src/renderer/components/hotbar/hotbar-icon.tsx#L99

msa0311 (approved):

aleksfront:
Eventually, we should replace it to https://www.npmjs.com/package/initialism for example.

src/renderer/components/hotbar/hotbar-icon.tsx
Nokel81: Why is name undefined? Should we be doing some better checks when items get added to the catalog?

chenhunghan: don't know, should we fix it at source like you say instead?

Nokel81:
@aleksfront I don't think so, that package isn't what we want (it only uses spaces for instance). Plus I don't think it handle's non-ascii characters at all.

PR #2832, opened 2021-05-20
Add (optional, enable by env) esbuild-loader

chenhunghan:
For faster build in dev mode.

- Add env `LENS_DEV_DISABLE_TS_CHECKER `, to disable type checking (as most editor already have it), default is on (as before)
- Add env `LENS_DEV_USE_ESBUILD_LOADER `, to enable `esbuild-loader`, default is off as before.

### DIFF
--- a/package.json
+++ b/package.json
@@ -313,6 +313,7 @@
     "electron": "^9.4.4",
     "electron-builder": "^22.10.5",
     "electron-notarize": "^0.3.0",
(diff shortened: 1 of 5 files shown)


chenhunghan:
No significant improvements after some basic benchmark (only reduce about ~1s, memory usage reduces ~200MB) , I will close this.

PR #2838, opened 2021-05-20
Split Catalog Categories and Registries

Nokel81:
- Main and Renderer now have different types
- No longer have unified class declarations
- Move towards a computed model on main, with the CatalogEntityRegistry
  folding over the declared handlers


Nokel81:
Closing as this is too far away from current.

PR #2841, opened 2021-05-21
Show all pod addresses

### LINKED ISSUE #2745: Lens should show ALL pod addresses
**Describe the bug**
Currently Lens shows only the first (?) IP for a pod. Pod can actually have many (at least two) addresses e.g. when running the cluster in dual-stack mode (IPv4&6)

**To Reproduce**
Connect Lens into cluster running with dual stack mode, e.g. like k0s supports it: https://docs.k0sproject.io/v1.21.0+k0s.0/dual-stack/

The pod json data has the addresses like so:
```
        "podIP": "10.244.115.129",
        "podIPs": [
            {
                "ip": "10.244.115.129"
            },
            {
                "ip": "fd00::3:e501"
            }
        ],
```

**Expected behavior**
Lens should show all Pod addresses

**Screenshots**
Screenshot 2021-05-11 at 11 41 00

- Lens Version: 4.2.4
- OS: OSX

vshakirova:
Screen Shot 2021-05-23 at 8 12 00 AM
Screen Shot 2021-05-23 at 8 12 16 AM

spec: 
https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/pkg/apis/core/types.go

fix: https://github.com/lensapp/lens/issues/2745

### DIFF
--- a/src/renderer/api/endpoints/pods.api.ts
+++ b/src/renderer/api/endpoints/pods.api.ts
@@ -275,6 +275,9 @@ export class Pod extends WorkloadKubeObject {
     }[];
     hostIP: string;
     podIP: string;
+    podIPs?: {
+      ip: string
+    }[];
     startTime: string;
     initContainerStatuses?: IPodContainerStatus[];
     containerStatuses?: IPodContainerStatus[];
@@ -484,6 +487,13 @@ export class Pod extends WorkloadKubeObject {
 
   getSelectedNodeOs(): string | undefined {
     return this.spec.nodeSelector?.["kubernetes.io/os"] || this.spec.nodeSelector?.["beta.kubernetes.io/os"];
+  }
+
+  getIPs(): string[] {
+    if(!this.status.podIPs) return [];
+    const podIPs = this.status.podIPs;
+
+    return podIPs.map(value => value.ip);
   }
 }
 
--- a/src/renderer/components/+network-services/service-details.tsx
+++ b/src/renderer/components/+network-services/service-details.tsx
@@ -77,6 +77,22 @@ export class ServiceDetails extends React.Component<Props> {
           {spec.clusterIP}
         </DrawerItem>
 
+        <DrawerItem name="Cluster IPs" hidden={!service.getClusterIps().length} labelsOnly>
+          {
+            service.getClusterIps().map(label => (
+              <Badge key={label} label={label}/>
+            ))
+          }
+        </DrawerItem>
+
+        <DrawerItem name="IP families" hidden={!service.getIpFamilies().length}>
+          {service.getIpFamilies().join(", ")}
+        </DrawerItem>
+
+        <DrawerItem name="IP family policy" hidden={!service.getIpFamilyPolicy()}>
+          {service.getIpFamilyPolicy()}
+        </DrawerItem>
+
         {service.getExternalIps().length > 0 && (
           <DrawerItem name="External IPs">
             {service.getExternalIps().map(ip => <div key={ip}>{ip}</div>)}
--- a/src/renderer/api/endpoints/service.api.ts
+++ b/src/renderer/api/endpoints/service.api.ts
@@ -59,12 +59,22 @@ export class Service extends KubeObject {
   spec: {
     type: string;
     clusterIP: string;
+    clusterIPs?: string[];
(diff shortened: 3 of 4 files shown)


src/renderer/components/+network-services/service-details.tsx
```diff
@@ -77,6 +77,24 @@ export class ServiceDetails extends React.Component<Props> {
           {spec.clusterIP}
         </DrawerItem>
 
+        {service.getClusterIps().length > 0 && (
+          <DrawerItem name="Cluster IPs">
+            {service.getClusterIps().join(", ")}
+          </DrawerItem>
+        )}
```
Nokel81: Nit, `DrawerItem` has a `hidden` prop, which I think should be used instead of doing the `&&`.

Furthermore, I think it would be better to encapsulate all the IPs in `<Badge>` instead of just joining with `", "`. This would also apply to the above item too.

Nokel81 (commented):
Thanks for doing this, just a visual recommendation.

src/renderer/components/+network-services/service-details.tsx
nevalla: How about just putting those in separate lines?

Nokel81: That could work too, I just felt that badge's would look nicer.

vshakirova: thank you, fixed it, but still have problems with test timeout.
```Timeout - Async callback was not invoked within the 60000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 60000 ms timeout specified by jest.setTimeout.Error:```

Comment (Nokel81) [reply src/renderer/components/+network-services/service-details.tsx]: @vshakirova If the tests pass locally on your machine, I would just rerun them. If you don't have permissions, just poke one of us.

File: src/renderer/api/endpoints/pods.api.ts [hunk 2]
```diff
@@ -485,6 +488,13 @@ export class Pod extends WorkloadKubeObject {
   getSelectedNodeOs(): string | undefined {
     return this.spec.nodeSelector?.["kubernetes.io/os"] || this.spec.nodeSelector?.["beta.kubernetes.io/os"];
   }
+
+  getIPs(): string[] {
+    if(!this.status.podIPs) return [];
+    const podIPs = this.status.podIPs;
+
+    return podIPs.map(value => value.ip);
+  }
```
Nokel81, line 497: nit:
```suggestion
  getIPs(): string[] {
    return this.status.podIPs?.map(({ ip }) => ip) ?? [];
  }
```

Nokel81 (approved):

nevalla (approved):

src/renderer/api/endpoints/pods.api.ts
```diff
@@ -275,6 +275,9 @@ export class Pod extends WorkloadKubeObject {
     }[];
     hostIP: string;
     podIP: string;
+    podIPs?: {
+      ip: string
+    }[];
```
BlackHole1, line 280: there must be podIPs, we can remove the podIP

If there is only one ip, then it will have the following format:

```
podIPs: [{
  ip: "ip"
}]
```

vshakirova: Hi @BlackHole1, yeah, we can. But the spec (https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L3638) still has this field. Maybe the spec still has this field for backward compatibility, PodIPs has been added in 1.16 kubernetes release (as I can see) https://github.com/kubernetes/kubernetes/blob/release-1.15/staging/src/k8s.io/api/core/v1/types.go

BlackHole1: If we want to follow the spec definition, then our `hostIP`, `podIP`, etc. also need to be changed to optional

BlackHole1: You're right.

Nokel81: I would prefer if we didn't touch the fields at all. We can modify the methods so that `getPodIps()` returns the unique ip strings as an array if we want.

PR #2844, opened 2021-05-21
Improve typings on `Table`, `ItemListLayout`, `KubeObjectListLayout`, `KubeApi`

Nokel81:

Breaking Changes:
- `KubeApi`'s constructor now requires `objectConstructor` (previously optional). This is done so that the resulting type of the `objectConstructor` (which returns `K extends KubeObject`) is "picked up" as the type parameter's value for `KubeApi`.
- Remove type parameter's default from `KubeApi`. This is required to help with the above.
- Remove type parameter's default from `IKubeObjectConstructor`. This is required to help with the first.

### DIFF
--- a/src/renderer/api/endpoints/resource-applier.api.ts
+++ b/src/renderer/api/endpoints/resource-applier.api.ts
@@ -30,7 +30,7 @@ export const resourceApplierApi = {
     "kubectl.kubernetes.io/last-applied-configuration"
   ],
 
-  async update<D extends KubeObject>(resource: object | string): Promise<D> {
+  async update<K extends KubeObject>(resource: object | string): Promise<K | null> {
     if (typeof resource === "string") {
       resource = jsYaml.safeLoad(resource);
     }
@@ -48,7 +48,7 @@ export const resourceApplierApi = {
           }
         });
 
-        return items.length === 1 ? items[0] : items;
+        return items[0] as K ?? null;
       });
   }
 };
--- a/src/renderer/api/kube-api.ts
+++ b/src/renderer/api/kube-api.ts
@@ -49,7 +49,7 @@ export interface IKubeApiOptions<T extends KubeObject> {
    */
   fallbackApiBases?: string[];
 
-  objectConstructor?: IKubeObjectConstructor<T>;
+  objectConstructor: IKubeObjectConstructor<T>;
   request?: KubeJsonApi;
   isNamespaced?: boolean;
   kind?: string;
@@ -115,7 +115,7 @@ export function forCluster<T extends KubeObject>(cluster: IKubeApiCluster, kubeC
   });
 }
 
-export function ensureObjectSelfLink(api: KubeApi, object: KubeJsonApiData) {
+export function ensureObjectSelfLink(api: KubeApi<KubeObject>, object: KubeJsonApiData) {
   if (!object.metadata.selfLink) {
     object.metadata.selfLink = createKubeApiURL({
       apiPrefix: api.apiPrefix,
@@ -127,7 +127,7 @@ export function ensureObjectSelfLink(api: KubeApi, object: KubeJsonApiData) {
   }
 }
 
-export type KubeApiWatchCallback = (data: IKubeWatchEvent, error: any) => void;
+export type KubeApiWatchCallback = (data: IKubeWatchEvent<KubeJsonApiData>, error: any) => void;
 
 export type KubeApiWatchOptions = {
   namespace: string;
@@ -135,7 +135,7 @@ export type KubeApiWatchOptions = {
   abortController?: AbortController
 };
 
-export class KubeApi<T extends KubeObject = any> {
+export class KubeApi<T extends KubeObject> {
   readonly kind: string;
   readonly apiBase: string;
   readonly apiPrefix: string;
@@ -152,7 +152,7 @@ export class KubeApi<T extends KubeObject = any> {
 
   constructor(protected options: IKubeApiOptions<T>) {
     const {
-      objectConstructor = KubeObject as IKubeObjectConstructor,
+      objectConstructor,
       request = apiKube,
       kind = options.objectConstructor?.kind,
       isNamespaced = options.objectConstructor?.namespaced
@@ -456,14 +456,14 @@ export class KubeApi<T extends KubeObject = any> {
 
             clearTimeout(timedRetry);
             timedRetry = setTimeout(() => { // we did not get any kubernetes errors so let's retry
-              this.watch({...opts, namespace, callback});
+              this.watch({ ...opts, namespace, callback });
             }, 1000);
           });
         });
 
         byline(nodeStream).on("data", (line) => {
           try {
-            const event: IKubeWatchEvent = JSON.parse(line);
+            const event: IKubeWatchEvent<KubeJsonApiData> = JSON.parse(line);
 
             if (event.type === "ERROR" && event.object.kind === "Status") {
               errorReceived = true;
@@ -474,7 +474,7 @@ export class KubeApi<T extends KubeObject = any> {
             this.modifyWatchEvent(event);
             callback(event, null);
           } catch (ignore) {
-          // ignore parse errors
+            // ignore parse errors
           }
         });
       })
@@ -487,7 +487,7 @@ export class KubeApi<T extends KubeObject = any> {
     return abort;
   }
 
-  protected modifyWatchEvent(event: IKubeWatchEvent) {
+  protected modifyWatchEvent(event: IKubeWatchEvent<KubeJsonApiData>) {
 
     switch (event.type) {
       case "ADDED":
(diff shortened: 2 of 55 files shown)


src/renderer/api/endpoints/resource-applier.api.ts
```diff
@@ -30,7 +30,7 @@ export const resourceApplierApi = {
     "kubectl.kubernetes.io/last-applied-configuration"
   ],
 
-  async update<D extends KubeObject>(resource: object | string): Promise<D> {
+  async update<KO extends KubeObject>(resource: object | string): Promise<KO | null> {
```
jakolehm: `KO` looks a bit odd, maybe just `K`.

src/renderer/api/endpoints/resource-applier.api.ts
```diff
@@ -48,7 +48,7 @@ export const resourceApplierApi = {
           }
         });
 
-        return items.length === 1 ? items[0] : items;
+        return items[0] as KO ?? null;
```
jakolehm: This changes behaviour but I have no idea why we have returned an item or an array based on length 🤔

src/renderer/api/kube-object.ts
```diff
@@ -265,8 +265,8 @@ export class KubeObject implements ItemObject {
   }
 
   // use unified resource-applier api for updating all k8s objects
-  async update<T extends KubeObject>(data: Partial<T>) {
-    return resourceApplierApi.update<T>({
+  async update<KO extends KubeObject>(data: Partial<KO>) {
```
jakolehm: `KO` looks odd also here ...

src/renderer/api/endpoints/resource-applier.api.ts
Nokel81: I agree that this changes behaviour, but if you look at the declared types. We never said we were returning an array. So I don't believe anyone was expecting an array.

jakolehm: Yeah, I agree.. old behaviour is a bit confusing.

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -186,16 +190,6 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
 
       return items;
     },
-
-    [FilterType.NAMESPACE]: items => {
```
nevalla: This is not needed or why is this removed?

nevalla (dismissed):

src/renderer/components/item-object-list/item-list-layout.tsx
Nokel81: Because `Items` don't necessarily have a `getNs()` command, only `KubeObjects` do. So this was moved to `KubeObjectListLayout` and passed down.

src/renderer/api/kube-api.ts
```diff
@@ -127,15 +127,15 @@ export function ensureObjectSelfLink(api: KubeApi, object: KubeJsonApiData) {
...
+export type KubeApiWatchCallback = (data: IKubeWatchEvent<KubeJsonApiData>, error: any) => void;
 
 export type KubeApiWatchOptions = {
   namespace: string;
   callback?: KubeApiWatchCallback;
   abortController?: AbortController
 };
 
-export class KubeApi<T extends KubeObject = any> {
+export class KubeApi<T extends KubeObject> {
```
jakolehm, line 138: This is a breaking change, right? `KubeApi` now requires `T` because it does not have default value `any`?

Would `export class KubeApi<T extends KubeObject = KubeObject> {` work?

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -46,8 +46,10 @@ import { namespaceStore } from "../+namespaces/namespace.store";
 
 // todo: refactor, split to small re-usable components
 
-export type SearchFilter<T extends ItemObject = any> = (item: T) => string | number | (string | number)[];
-export type ItemsFilter<T extends ItemObject = any> = (items: T[]) => T[];
+export type SearchFilter<Item extends ItemObject> = (item: Item) => string | number | (string | number)[];
```
jakolehm: In other places we are using single uppercase letter. Here we chose to use `Item`, any reason for that?

src/renderer/components/table/table.tsx
```diff
@@ -74,16 +74,16 @@ export const orderByUrlParam = createPageParam({
 });
 
 @observer
-export class Table extends React.Component<TableProps> {
-  static defaultProps: TableProps = {
+export class Table<Item> extends React.Component<TableProps<Item>> {
```
jakolehm, line 77: Table improvements are also breaking changes.

src/renderer/kube-object.store.ts
```diff
@@ -31,16 +31,16 @@ import { IKubeApiQueryParams, KubeApi, parseKubeApi } from "./api/kube-api";
...
+  api?: KubeApi<K>;
   reqInit?: RequestInit;
 }
 
-export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemStore<T> {
+export abstract class KubeObjectStore<K extends KubeObject> extends ItemStore<K> {
   static defaultContext = observable.box<ClusterContext>(); // TODO: support multiple cluster contexts
 
-  abstract api: KubeApi<T>;
+  abstract api: KubeApi<K>;
```
jakolehm: All changes in this file after this line are just because `T` is now `K`?

src/renderer/kube-object.store.ts
```diff
@@ -406,7 +407,7 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
       switch (type) {
         case "ADDED":
         case "MODIFIED":
-          const newItem = new api.objectConstructor(object);
+          const newItem = new api.objectConstructor(object) as K;
```
jakolehm: Do we need `as K` here?

src/renderer/api/kube-api.ts
Nokel81: Yes that would work. I can change it back.

src/renderer/components/item-object-list/item-list-layout.tsx
Nokel81: Not really, I guess I could change these to `I` instead.

src/renderer/api/kube-api.ts
Nokel81: It is a breaking change yes, though only if you take it as an argument, you don't need to specify it when doing `new KubeApi(...)`.

Furthermore, from a TS perspective it is better not to have this as a default parameter because then the passed in `objectConstructor` doesn't get to "set" this type parameter.

src/renderer/components/table/table.tsx
Nokel81: Are they? I don't think so, it isn't required to use the new `renderRow` function (just helps with typings) and it is not required to specify the type arguments on use. (if you don't specify any, currently it is all of them or none of them).

src/renderer/kube-object.store.ts
Nokel81: Mostly

src/renderer/kube-object.store.ts
Nokel81: Not if I removed the default type argument from `IKubeObjectConstructor`. Which imo is the better solution.

src/renderer/api/kube-api.ts
jakolehm: Please describe this in the PR description (maybe PR title also needs an update?). I also think we need to add `area/extension` label because this affects extension API.

src/renderer/components/table/table.tsx
jakolehm: Ok 👍  I somehow misinterpreted the diff here.

src/renderer/components/table/table.tsx
```diff
@@ -171,9 +171,20 @@ export class Table extends React.Component<TableProps> {
     });
   }
 
-  renderRows() {
-    const { sortable, noItems, children, virtual, customRowHeights, rowLineHeight, rowPadding, items, getTableRow, selectedItemId, className } = this.props;
+  getContent() {
```
jakolehm: Should this be private/protected if it's just used within `renderRows`?

src/renderer/kube-object.store.ts
jakolehm: These kind of changes makes reviewing hard. What's wrong with `T ` and why we wanted to change existing practice (single letter vs word) in this PR 🤔

src/renderer/components/table/table.tsx
Nokel81: Sure... for react components I haven't been able to figure out why `private` or `protected` are good ideas for the instance methods. Because it is not like we call the constructor ourselves.

src/renderer/kube-object.store.ts
Nokel81: I can revert them, I change it because `K` is closer to `KubeObject` which is what it represents.

PR #2845, opened 2021-05-21
Change where extension IPC is exported

Nokel81:
- Now under Ipc.IpcMain and Ipc.IpcRenderer

- Allow listener's to be cleaned up early by extensions

### DIFF
--- a/docs/extensions/guides/ipc.md
+++ b/docs/extensions/guides/ipc.md
@@ -1,10 +1,11 @@
 # Inter Process Communication
 
-A Lens Extension can utilize IPC to send information between its `LensRendererExtension` and its `LensMainExtension`.
+A Lens Extension can utilize IPC to send information between the `renderer` and `main` processes.
 This is useful when wanting to communicate directly within your extension.
+
 For example, if a user logs into a service that your extension is a facade for and `main` needs to know some information so that you can start syncing items to the `Catalog`, this would be a good way to send that information along.
 
-IPC channels are blocked off per extension.
+IPC channels are sectioned off per extension.
 Meaning that each extension can only communicate with itself.
 
 ## Types of Communication
@@ -14,22 +15,22 @@ There are two flavours of communication that are provided:
 - Event based (IPC)
 - Request based (RPC)
 
-### Event Based IPC
+### Event Based or IPC
 
 This is the same as an [Event Emitter](https://nodejs.org/api/events.html#events_class_eventemitter) but is not limited to just one Javascript process.
 This is a good option when you need to report that something has happened but you don't need a response.
 
 This is a fully two-way form of communication.
-Both `LensMainExtension` and `LensRendererExtension` can do this sort of IPC.
+Both `main` and `renderer` can do this sort of IPC.
 
-### Request Based IPC
+### Request Based or RPC
 
-This is more like a Remote Procedure Call (RPC).
-With this sort of IPC the caller waits for the result from the other side.
-This is accomplished by returning a `Promise<T>` which needs to be `await`-ed.
+This is more like a Remote Procedure Call (RPC) or Send-Receive-Reply (SRR).
+With this sort of communication the caller needs to wait for the result from the other side.
+This is accomplished by `await`-int the returned `Promise<any>`.
 
 This is a unidirectional form of communication.
-Only `LensRendererExtension` can initiate this kind of request, and only `LensMainExtension` can and respond this this kind of request.
+Only `renderer` can initiate this kind of request, and only `main` can and respond this this kind of request.
 
 ## Registering IPC Handlers and Listeners
 
@@ -75,7 +76,7 @@ function onInitialize(event: Types.IpcMainEvent, id: string) {
 ```
 
 In other files, it is not necessary to pass around any instances.
-It should be able to just call `getInstance()` everywhere in your extension as needed.
+It should be able to just call `IpcMain.getInstance()` anywhere it is needed in your extension.
 
 ---
 
@@ -109,13 +110,16 @@ It is necessary to create child classes of these `abstract class`'s in your exte
 ---
 
 As this example shows: the channel names *must* be the same.
-It should also be noted that "listeners" and "handlers" are specific to either `LensRendererExtension` and `LensMainExtension`.
+It should also be noted that "listeners" and "handlers" are specific to either `renderer` or `main`.
 There is no behind the scenes transfer of these functions.
 
-If you want to register a "handler" you would call `IpcMain.getInstance().handle(...)` instead.
+To register a "handler" call `IpcMain.getInstance().handle(...)`.
 The cleanup of these handlers is handled by Lens itself.
 
-Calling either `IpcRenderer.getInstance().broadcast(...)` or `IpcMain.getInstance().broadcast(...)` sends an event to all renderer frames and to main.
+The `listen()` methods on `Ipc.Main` and `Ipc.Renderer` return a `Disposer`, or more specifically, a `() => void`.
+This can be optionally called to remove the listener early.
+
+Calling either `IpcRenderer.getInstance().broadcast(...)` or `IpcMain.getInstance().broadcast(...)` sends an event to all `renderer` frames and to `main`.
 Because of this, no matter where you broadcast from, all listeners in `main` and `renderer` will be notified.
 
 ### Allowed Values
@@ -123,6 +127,6 @@ Because of this, no matter where you broadcast from, all listeners in `main` and
 This IPC mechanism utilizes the [Structured Clone Algorithm](developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) for serialization.
 This means that more types than what are JSON serializable can be used, but not all the information will be passed through.
 
-## Using RPC
+## Using Request Based Communication
 
 If you are meaning to do a request based call from `renderer`, you should do `const res = await IpcRenderer.getInstance().invoke(<channel>, ...<args>));` instead.
--- a/docs/extensions/guides/ipc.md
+++ b/docs/extensions/guides/ipc.md
@@ -27,10 +27,10 @@ Both `main` and `renderer` can do this sort of IPC.
 
 This is more like a Remote Procedure Call (RPC) or Send-Receive-Reply (SRR).
 With this sort of communication the caller needs to wait for the result from the other side.
-This is accomplished by `await`-int the returned `Promise<any>`.
+This is accomplished by `await`-ing the returned `Promise<any>`.
 
 This is a unidirectional form of communication.
-Only `renderer` can initiate this kind of request, and only `main` can and respond this this kind of request.
+Only `renderer` can initiate this kind of request, and only `main` can and respond to this kind of request.
 
 ## Registering IPC Handlers and Listeners
 
@@ -76,7 +76,7 @@ function onInitialize(event: Types.IpcMainEvent, id: string) {
 ```
 
 In other files, it is not necessary to pass around any instances.
-It should be able to just call `IpcMain.getInstance()` anywhere it is needed in your extension.
+You should be able to just call `IpcMain.getInstance()` anywhere it is needed in your extension.
 
 ---
 
--- a/docs/extensions/guides/ipc.md
+++ b/docs/extensions/guides/ipc.md
@@ -7,12 +7,12 @@ For example, if a user logs into a service that your extension is a facade for a
 IPC channels are blocked off per extension.
 Meaning that each extension can only communicate with itself.
 
-## Types of IPC
+## Types of Communication
 
-There are two flavours of IPC that are provided:
+There are two flavours of communication that are provided:
 
-- Event based
-- Request based
+- Event based (IPC)
+- Request based (RPC)
 
 ### Event Based IPC
 
@@ -42,8 +42,8 @@ To register either a handler or a listener, you should do something like the fol
 
 `main.ts`:
 ```typescript
-import { LensMainExtension, Interface, Types, Store } from "@k8slens/extensions";
-import { registerListeners, IpcMain } from "./helpers/main";
+import { LensMainExtension } from "@k8slens/extensions";
+import { IpcMain } from "./helpers/main";
 
 export class ExampleExtensionMain extends LensMainExtension {
   onActivate() {
@@ -59,9 +59,9 @@ Lens will automatically clean up that store and all the handlers on deactivation
 
 `helpers/main.ts`:
 ```typescript
-import { Store } from "@k8slens/extensions";
+import { Ipc, Types } from "@k8slens/extensions";
 
-export class IpcMain extends Store.MainIpcStore {
+export class IpcMain extends Ipc.IpcMain {
   constructor(extension: LensMainExtension) {
     super(extension);
 
@@ -81,7 +81,7 @@ It should be able to just call `getInstance()` everywhere in your extension as n
 
 `renderer.ts`:
 ```typescript
-import { LensRendererExtension, Interface, Types } from "@k8slens/extensions";
+import { LensRendererExtension } from "@k8slens/extensions";
 import { IpcRenderer } from "./helpers/renderer";
 
 export class ExampleExtensionRenderer extends LensRendererExtension {
@@ -99,9 +99,9 @@ It is also needed to create an instance to broadcast messages too.
 
 `helpers/renderer.ts`:
 ```typescript
-import { Store } from "@k8slens/extensions";
+import { Ipc } from "@k8slens/extensions";
 
-export class IpcMain extends Store.RendererIpcStore {}
+export class IpcRenderer extends Ipc.IpcRenderer {}
 ```
 
 It is necessary to create child classes of these `abstract class`'s in your extension before you can use them.
@@ -112,10 +112,10 @@ As this example shows: the channel names *must* be the same.
 It should also be noted that "listeners" and "handlers" are specific to either `LensRendererExtension` and `LensMainExtension`.
 There is no behind the scenes transfer of these functions.
 
-If you want to register a "handler" you would call `Store.MainIpcStore.handleIpc(...)` instead.
+If you want to register a "handler" you would call `Ipc.IpcMain.handleRpc(...)` instead.
 The cleanup of these handlers is handled by Lens itself.
 
-`Store.RendererIpcStore.broadcastIpc(...)` and `Store.MainIpcStore.broadcastIpc(...)` sends an event to all renderer frames and to main.
+`Ipc.IpcRenderer.broadcastIpc(...)` and `Ipc.IpcMain.broadcastIpc(...)` sends an event to all renderer frames and to main.
 Because of this, no matter where you broadcast from, all listeners in `main` and `renderer` will be notified.
 
 ### Allowed Values
@@ -128,4 +128,4 @@ This means that more types than what are JSON serializable can be used, but not 
 Calling IPC is very simple.
 If you are meaning to do an event based call, merely call `broadcastIpc(<channel>, ...<args>)` from within your extension.
 
-If you are meaning to do a request based call from `renderer`, you should do `const res = await Store.RendererIpcStore.invokeIpc(<channel>, ...<args>));` instead.
+If you are meaning to do a request based call from `renderer`, you should do `const res = await Ipc.IpcRenderer.invokeRpc(<channel>, ...<args>));` instead.
--- a/docs/extensions/guides/ipc.md
+++ b/docs/extensions/guides/ipc.md
@@ -61,11 +61,11 @@ Lens will automatically clean up that store and all the handlers on deactivation
 ```typescript
 import { Ipc, Types } from "@k8slens/extensions";
 
-export class IpcMain extends Ipc.IpcMain {
+export class IpcMain extends Ipc.Main {
   constructor(extension: LensMainExtension) {
     super(extension);
 
-    this.listenIpc("initialize", onInitialize);
+    this.listen("initialize", onInitialize);
   }
 }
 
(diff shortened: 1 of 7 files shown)


src/extensions/ipc/ipc-main.ts
```diff
@@ -19,27 +19,43 @@
...
+    return cleanup;
   }
 
-  listenIpc(channel: string, listener: (event: Electron.IpcMainEvent, ...args: any[]) => any): void {
+  /**
+   * Declare a RPC over `channel`. Lens will cleanup when the extension is disabled or uninstalled
+   * @param channel The name of the RPC
+   * @param handler The remote procedure that is called
+   */
+  handleIpc(channel: string, handler: (event: Electron.IpcMainInvokeEvent, ...args: any[]) => any): void {
```
stefcameron: Why aren't we returning a disposer in this case too?

src/extensions/ipc/ipc-renderer.ts
```diff
@@ -19,23 +19,40 @@
...
+    return cleanup;
   }
 
+  /**
+   * Request main to execute its function over the `channel` channel.
+   * @param channel The channel to invoke a RPC on
+   * @param args The arguments to pass to the RPC
+   * @returns A promise of the resulting value
+   */
   invokeIpc(channel: string, ...args: any[]): Promise<any> {
```
stefcameron: Do we call `invokeIpc()` regardless of how the channel was declared in `main` (i.e. same method whether it's a listener or a handler)? Probably not since there's still a `broadcastIpc()` in the `IpcRegistrar`, I suppose.

stefcameron (commented):
The listeners feel a lot like event listeners, so it feels like there should be a pair of methods to add and remove listeners, instead of the add returning a disposer that needs to be invoked (and therefore needs to be kept in scope if a component is adding a listener on mount and then removing it on unmount). Disposer is OK, but doesn't feel natural to me.

I think this PR should also be updating the docs here: https://github.com/lensapp/lens/pull/2775/files#diff-35cfa0896b5dd8347734ddd548f83c3869571d486e0651c7c0e7d016b5194bd2R34

Things have changed.

And we should take this opportunity to remove the unnecessary `registerListeners` import from this line too: https://github.com/lensapp/lens/pull/2775/files#diff-35cfa0896b5dd8347734ddd548f83c3869571d486e0651c7c0e7d016b5194bd2R46

src/extensions/ipc/ipc-main.ts
Nokel81: I can add some documentation here. But RPC is quite different than IPC. I think I will change the method names too.

Namely, because it is an error to call `invokeRpc` without an "alive" `handleRpc` it doesn't lend itself to being as transient as `listenIpc`. 

Furthermore, the use case surrounding early cleanup of `listenIpc` doesn't apply to `handleRpc` as `main` has no concept of components mounting and unmounting on the DOM.

src/extensions/ipc/ipc-renderer.ts
Nokel81: No, `invokeIpc` (I will rename this to `invokeRpc`) *only* interacts with `handleRpc` (formerly `handleIpc`).

Nokel81:
> I think this PR should also be updating the docs here:

Thanks, I forgot about doing that.

> The listeners feel a lot like event listeners, so it feels like there should be a pair of methods to add and remove listeners

They are a form of event listeners. I did it this was because the requested use case was the ability to tie the lifetime of a listener to the lifetime of a component being mounted.

This use case is easily handled by this change by something like the following:

```ts
componentDidMount() {
  disposeOnUnmount(this, [
    Ipc.IpcRenderer.listenIpc("channel", handler),
  ]);
}
```

src/extensions/ipc/ipc-main.ts
stefcameron: OK, makes sense. Some mention in the docs about this would good.

stefcameron:
> > I think this PR should also be updating the docs here:
> 
> Thanks, I forgot about doing that.

👍 

> > The listeners feel a lot like event listeners, so it feels like there should be a pair of methods to add and remove listeners
> 
> They are a form of event listeners. I did it this was because the requested use case was the ability to tie the lifetime of a listener to the lifetime of a component being mounted.
> 

Maybe that's too narrow of scope? What if there was a good use case for _not_ tie it to the lifetime of a component? What if it's in a Context Provider that survives component mounts and dismounts? I use those heavily in the CC extension.

But then even if we had a `removeListener()`, it would probably require a reference to the handler that was added in order to be removed, so that's not much different from remembering the disposer.

OK, seems fine then. 🙂

Nokel81:
@stefcameron yes you would require the exact reference to the listener, so as you said even from that point of view the disposer is just the same (from the POV of remembering it).

docs/extensions/guides/ipc.md
```diff
@@ -112,10 +112,10 @@ As this example shows: the channel names *must* be the same.
 It should also be noted that "listeners" and "handlers" are specific to either `LensRendererExtension` and `LensMainExtension`.
 There is no behind the scenes transfer of these functions.
 
-If you want to register a "handler" you would call `Store.MainIpcStore.handleIpc(...)` instead.
+If you want to register a "handler" you would call `Ipc.IpcMain.handleRpc(...)` instead.
 The cleanup of these handlers is handled by Lens itself.
 
-`Store.RendererIpcStore.broadcastIpc(...)` and `Store.MainIpcStore.broadcastIpc(...)` sends an event to all renderer frames and to main.
+`Ipc.IpcRenderer.broadcastIpc(...)` and `Ipc.IpcMain.broadcastIpc(...)` sends an event to all renderer frames and to main.
```
stefcameron: I would say, "sends an event" -> "send events", since we have 2 things we're talking about.

docs/extensions/guides/ipc.md
```diff
@@ -128,4 +128,4 @@ This means that more types than what are JSON serializable can be used, but not
 Calling IPC is very simple.
 If you are meaning to do an event based call, merely call `broadcastIpc(<channel>, ...<args>)` from within your extension.
```
stefcameron: "`broadcastIpc(<channel>, ...<args>)`" -> "`Ipc.IcpMain.broadcastIpc(<channel>, ...<args>)` (`IpcRenderer` could also be used here, as there would be no difference in the result)"

docs/extensions/guides/ipc.md
stefcameron: And then, couldn't these calls be `IcpRenderer.getInstance().broadcastIcp(...)` and `IcpMain.getInstance().broadcastIpc()`?

stefcameron (commented):
- [This title](https://github.com/lensapp/lens/pull/2845/files/ccac859b0ba585bde42f89735b77c01c37c70bd2..fa87d9edbdb109b194e2b0b926a6e9be48d2c04f#diff-35cfa0896b5dd8347734ddd548f83c3869571d486e0651c7c0e7d016b5194bd2R25) should be renamed "Request Based RPC" now.
- Should [this title](https://github.com/lensapp/lens/pull/2845/files/ccac859b0ba585bde42f89735b77c01c37c70bd2..fa87d9edbdb109b194e2b0b926a6e9be48d2c04f#diff-35cfa0896b5dd8347734ddd548f83c3869571d486e0651c7c0e7d016b5194bd2R126) now be "Using IPC and RPC"?

I'm not sure renaming `handleIpc` to `handleRpc` made things better. Technically, it's correct, I guess, since RPC is "Remote Procedure Call" and you would expect a response from calling that procedure.

But it has created this IPC/RPC jumble, like `Ipc.IpcRenderer.handleRpc()` -- wait, what?

Does the extension need to really understand the deep difference between the two? If I wanted to, I could use a handler to send a message, and just not `await` it, knowing that this particular channel handler just responds with `undefined` anyway (same as calling any function in JavaScript that doesn't explicitly return a value; all functions technically return a value even if they don't have a `return` statement in them). It would be no different from broadcasting to a single listener on a given channel.

What if we dropped the `Ipc` and `Rpc` postfixes (since we're renaming things anyway)? You either `listen()` or `handle()`, and then either `broadcast()` or `invoke()`. I don't think, as an extension developer, I really need to care about whether I'm doing IPC or RPC, do I? The gist of it is that I'm making some type of cross-process remote call. In either case, the same object serialization algorithm is at play for any parameters I pass. The only differences I see are:
- `broadcast()` sends to _all_ listeners of the same channel and never responds (so no point `await`-ing it)
- `invoke()` probably sends to the handler and responds (so I should probably `await` it)

`class IpcRenderer extends Ipc.IpcRenderer` -> could we have `Ipc.Renderer` here, dropping the redundant `Ipc`? Maybe it's the name of the class in the extension that's making this a bit confusing, because in the docs, we end-up with two distinct `IpcRenderer` classes to talk about. If not `Ipc.Renderer` (and `Ipc.Main`), then how about `class RemoteRenderer extends Ipc.Ipc.Renderer`?

And then it's still strange to me to do `RemoteRenderer.getInstance().invokeRpc()` (or even if it were `RemoteRenderer.getInstance().invoke()`). It feels like I'm invoking on the `renderer`, when I'm actually invoking on `main`. I want to do `RemoteMain.getInstance().invoke()` -- from my code running in `renderer` (because I'm invoking on the `main` thread, not on the `renderer` thread, because I know the handler is declared on `main`). Maybe that's already possible? Not sure.

Nokel81:
> What if we dropped the Ipc and Rpc postfixes?

I am warming up to this. There is still a difference but I think it should be fine.

> `class IpcRenderer extends Ipc.IpcRenderer` -> could we have `Ipc.Renderer` here, dropping the redundant `Ipc`?

Yes

> And then it's still strange to me to do `RemoteRenderer.getInstance().invokeRpc()`

Where is `RemoteRenderer` coming from? If you have a `class IpcRenderer extends Ipc.Renderer` then it would be `IpcRenderer.getInstance().invoke()`.

Would you be fine with dropping the `ipc` suffix, but keeping the `rpc` suffix? Technically it is similar to for [Send-Receive-Reply](https://en.wikipedia.org/wiki/SIMPL).

> because I'm invoking on the main thread, not on the renderer thread, because I know the handler is declared on main).

Well you are doing the invoke in the renderer process. Which is being serialized and then sent to main.

stefcameron:
> > And then it's still strange to me to do `RemoteRenderer.getInstance().invokeRpc()`
> 
> Where is `RemoteRenderer` coming from? If you have a `class IpcRenderer extends Ipc.Renderer` then it would be `IpcRenderer.getInstance().invoke()`.
> 

I was just using `RemoteRenderer` from my previous suggestion. If you're thinking of dropping the `Icp` prefix, then we could leave the example as `export class IpcRenderer extends Icp.Renderer`. Now there's a clearer distinction between `IpcRenderer` as the thing specific to my extension, and `Ipc.Renderer` as the thing that comes from Lens that I extended.

> Would you be fine with dropping the `ipc` suffix, but keeping the `rpc` suffix? Technically it is similar to for [Send-Receive-Reply](https://en.wikipedia.org/wiki/SIMPL).
> 

I think the name of the method `invoke()` is enough of an implication to what's going on. The JSDocs for `invoke()` in `Icp.Renderer` could certainly mention this is RPC instead of IPC to pull out the subtle distinction, but I don't think the method name needs it.

> > because I'm invoking on the main thread, not on the renderer thread, because I know the handler is declared on main).
> 
> Well you are doing the invoke in the renderer process. Which is being serialized and then sent to main.

I see it more as invoking the `main` thread to do something because that's where the handler is defined, but I guess it depends on how you think about it. You could say that when I call an API on a backend from a web app, I'm not getting the backend and then invoking that API. I'm getting a client-side HTTP facility (like `fetch` in the browser) and using that from the client (i.e. `renderer` in this case) to invoke that API on the backend (i.e. `main`). So `IpcRenderer` is akin to `fetch` in this scenario.

But in the Lens Extension context, unless I'm wrong, it seems I can get the backend, so to speak, by importing the `IpcMain` class and calling `IcpMain.getInstance().invoke()` from the `renderer` -- or will that not work because the singleton instance won't exist (crash, which would be good; a quick notice that I'm doing something wrong) or because a new instance will be created (bad, things will appear to work, but won't) because the original instance only exists on `main` and won't "cross over"?

Nokel81:
> I was just using RemoteRenderer from my previous suggestion. If you're thinking of dropping the Icp prefix, then we could leave the example as export class IpcRenderer extends Icp.Renderer. Now there's a clearer distinction between IpcRenderer as the thing specific to my extension, and Ipc.Renderer as the thing that comes from Lens that I extended.

I think we are on the same page here, good.

> I think the name of the method invoke() is enough of an implication to what's going on. The JSDocs for invoke() in Icp.Renderer could certainly mention this is RPC instead of IPC to pull out the subtle distinction, but I don't think the method name needs it.

👍 

> I see it more as invoking the main thread to do something because that's where the handler is defined, but I guess it depends on how you think about it. You could say that when I call an API on a backend from a web app, I'm not getting the backend and then invoking that API. I'm getting a client-side HTTP facility (like fetch in the browser) and using that from the client (i.e. renderer in this case) to invoke that API on the backend (i.e. main). So IpcRenderer is akin to fetch in this scenario.

Yes it is more akin to `fetch` but not entirely the same

> But in the Lens Extension context, unless I'm wrong, it seems I can get the backend, so to speak, by importing the IpcMain class and calling IcpMain.getInstance().invoke() from the renderer -- or will that not work because the singleton instance won't exist (crash, which would be good; a quick notice that I'm doing something wrong) or because a new instance will be created (bad, things will appear to work, but won't) because the original instance only exists on main and won't "cross over"?

Two things here, one `IpcMain.getInstance().invoke()` would throw a TypeError "invoke is not a function". `invoke` only exists on `Ipc.Renderer`.

Second, if you import `Ipc.Main` on the `renderer` then you would get crashes about `ipcMain` (the internal electron instance) being `undefined`

stefcameron:
> > But in the Lens Extension context, unless I'm wrong, it seems I can get the backend, so to speak, by importing the IpcMain class and calling IcpMain.getInstance().invoke() from the renderer -- or will that not work because the singleton instance won't exist (crash, which would be good; a quick notice that I'm doing something wrong) or because a new instance will be created (bad, things will appear to work, but won't) because the original instance only exists on main and won't "cross over"?
> 
> Two things here, one `IpcMain.getInstance().invoke()` would throw a TypeError "invoke is not a function". `invoke` only exists on `Ipc.Renderer`.
> 

Right. 🤦 

> Second, if you import `Ipc.Main` on the `renderer` then you would get crashes about `ipcMain` (the internal electron instance) being `undefined`

OK, good!

docs/extensions/guides/ipc.md
```diff
@@ -112,20 +112,17 @@ As this example shows: the channel names *must* be the same.
 It should also be noted that "listeners" and "handlers" are specific to either `LensRendererExtension` and `LensMainExtension`.
 There is no behind the scenes transfer of these functions.
 
-If you want to register a "handler" you would call `Ipc.IpcMain.handleRpc(...)` instead.
+If you want to register a "handler" you would call `IpcMain.getInstance().handle(...)` instead.
 The cleanup of these handlers is handled by Lens itself.
 
-`Ipc.IpcRenderer.broadcastIpc(...)` and `Ipc.IpcMain.broadcastIpc(...)` sends an event to all renderer frames and to main.
+Calling either `IpcRenderer.getInstance().broadcast(...)` or `IpcMain.getInstance().broadcast(...)` sends an event to all renderer frames and to main.
```
stefcameron: Since we used backticks on `main` everywhere else, the one reference here at the end of the sentence should have them too.

stefcameron (commented):
- Should [this title](https://github.com/lensapp/lens/pull/2845/files#diff-35cfa0896b5dd8347734ddd548f83c3869571d486e0651c7c0e7d016b5194bd2R25) be "Request Based RPC"?
- I think [this section](https://github.com/lensapp/lens/pull/2845/files#diff-35cfa0896b5dd8347734ddd548f83c3869571d486e0651c7c0e7d016b5194bd2R34) is still not updated for this IPC vs RPC distinction.
- We're missing mention of the disposer when declaring a listener, though I suppose that will come out in the docs based on the typings. Just mentioning in case you wanted to point this out now, since this PR also adds that feature.

Nokel81:
@stefcameron I think I managed to guess as to which sections you meant since your links don't reference the guide file at all.

I would recommend doing a line by line comment on github instead as it is easier for me to follow. Thanks.

docs/extensions/guides/ipc.md
```diff
@@ -14,22 +15,22 @@ There are two flavours of communication that are provided:
...
 
-### Request Based IPC
+### Request Based or RPC
 
-This is more like a Remote Procedure Call (RPC).
-With this sort of IPC the caller waits for the result from the other side.
-This is accomplished by returning a `Promise<T>` which needs to be `await`-ed.
+This is more like a Remote Procedure Call (RPC) or Send-Receive-Reply (SRR).
+With this sort of communication the caller needs to wait for the result from the other side.
+This is accomplished by `await`-int the returned `Promise<any>`.
```
stefcameron: "int" -> "ing"

docs/extensions/guides/ipc.md
```diff
@@ -14,22 +15,22 @@ There are two flavours of communication that are provided:
...
 
 This is a unidirectional form of communication.
-Only `LensRendererExtension` can initiate this kind of request, and only `LensMainExtension` can and respond this this kind of request.
+Only `renderer` can initiate this kind of request, and only `main` can and respond this this kind of request.
```
stefcameron: "respond this" -> "respond to this"

docs/extensions/guides/ipc.md
```diff
@@ -75,7 +76,7 @@ function onInitialize(event: Types.IpcMainEvent, id: string) {
 ```
 
 In other files, it is not necessary to pass around any instances.
-It should be able to just call `getInstance()` everywhere in your extension as needed.
+It should be able to just call `IpcMain.getInstance()` anywhere it is needed in your extension.
```
stefcameron: "It should be able" -> "You should be able"?

stefcameron (commented):
✅ I found a few more typos, but other than that, this is looking solid now! Thanks for working through all this with me. Hopefully the rest of the team likes these changes too.

Nokel81:
@stefcameron Thanks for all the comments on the docs. Writing is not one of my strong suits.

stefcameron:
> I would recommend doing a line by line comment on github instead as it is easier for me to follow. Thanks.

I would have, but GitHub didn't consider those lines as part of the review, so I couldn't add a comment on them. I don't know why GitHub doesn't let you comment on _any_ line in a file that's been modified. Instead, it prevents commenting on lines that aren't within, like +/- 5 lines of a line that's been touched. 🤷‍♂️

Nokel81:
Oh now I see, yes that is annoying.

jakolehm (approved):
LGTM

docs/extensions/guides/ipc.md
```diff
@@ -1,35 +1,36 @@
...
-This is more like a Remote Procedure Call (RPC).
-With this sort of IPC the caller waits for the result from the other side.
-This is accomplished by returning a `Promise<T>` which needs to be `await`-ed.
+This is more like a Remote Procedure Call (RPC) or Send-Receive-Reply (SRR).
+With this sort of communication the caller needs to wait for the result from the other side.
+This is accomplished by `await`-ing the returned `Promise<any>`.
 
 This is a unidirectional form of communication.
-Only `LensRendererExtension` can initiate this kind of request, and only `LensMainExtension` can and respond this this kind of request.
+Only `renderer` can initiate this kind of request, and only `main` can and respond to this kind of request.
```
jim-docker, line 33:
```suggestion
Only `renderer` can initiate this kind of request, and only `main` can receive and respond to this kind of request.
```

stefcameron:
Is it possible to either merge this, or rebase it to get the `mobx 6` updates that are on master? Also would be great to get this into a new Lens 5 beta... 😄

Nokel81:
@stefcameron I think merging it would be best and we want to do another beta release soon so this should hopefully be in that too.

jakolehm:
@nevalla @ixrock ping

docs/extensions/guides/ipc.md
jim-docker, line 33:
```suggestion
Only `renderer` can initiate this kind of request, and only `main` can receive and respond to this kind of request.
```

src/extensions/ipc/ipc-main.ts
```diff
@@ -19,27 +19,43 @@
...
-  handleIpc(channel: string, handler: (event: Electron.IpcMainInvokeEvent, ...args: any[]) => any): void {
+  /**
+   * Listen for broadcasts within your extension
+   * @param channel The channel to listen for broadcasts on
+   * @param listener The function that will be called with the arguments of the broadcast
+   * @returns An optional disopser, Lens will cleanup when the extension is disabled or uninstalled even if this is not called
+   */
+  listen(channel: string, listener: (event: Electron.IpcRendererEvent, ...args: any[]) => any): Disposer {
     const prefixedChannel = `extensions@${this[IpcPrefix]}:${channel}`;
+    const cleanup = once(() => ipcMain.removeListener(prefixedChannel, listener));
```
jim-docker, line 42: nice

nevalla (approved):

src/extensions/ipc/ipc-renderer.ts
```diff
@@ -0,0 +1,65 @@
...
+    extension[Disposers].push(() => IpcRenderer.resetInstance());
+  }
+
+  /**
+   * Listen for broadcasts within your extension.
+   * If the lifetime of the listener should be tied to the mounted lifetime of
+   * a component then putting the returned value in a `disposeOnUnmount` call will suffice.
+   * @param channel The channel to listen for broadcasts on
+   * @param listener The function that will be called with the arguments of the broadcast
+   * @returns An optional disopser, Lens will cleanup even if this is not called
```
jim-docker, line 40:
```suggestion
   * @returns An optional disposer, Lens will cleanup even if this is not called
```

jim-docker (approved):

Nokel81:
Will make a follow up PR with the docs changes.

PR #2846, opened 2021-05-21
Bump prometheus and kube-state-metrics for arm+other arch support

jonstelly:
- Support for running kube-state-metrics on all architectures
- Bump prometheus for security vuln. fix 

See [here](https://github.com/prometheus/prometheus/releases/tag/v2.27.1) for prometheus release note about security vulnerability

See [here](https://github.com/kubernetes/kube-state-metrics/releases/tag/v2.0.0-rc.1) for change to new authoritative source for kube-state-metrics image that supports arm, ppc64, etc...

Signed-off-by: Jon Stelly <967068+jonstelly@users.noreply.github.com>

### DIFF
--- a/extensions/metrics-cluster-feature/resources/03-statefulset.yml.hb
+++ b/extensions/metrics-cluster-feature/resources/03-statefulset.yml.hb
@@ -54,7 +54,7 @@ spec:
               mountPath: /var/lib/prometheus
       containers:
         - name: prometheus
-          image: quay.io/prometheus/prometheus:v2.26.0
+          image: quay.io/prometheus/prometheus:v2.27.1
           args:
             - --web.listen-address=0.0.0.0:9090
             - --config.file=/etc/prometheus/prometheus.yaml
--- a/extensions/metrics-cluster-feature/resources/14-kube-state-metrics-deployment.yml.hb
+++ b/extensions/metrics-cluster-feature/resources/14-kube-state-metrics-deployment.yml.hb
@@ -23,23 +23,15 @@ spec:
                   operator: In
                   values:
                   - linux
-                - key: kubernetes.io/arch
-                  operator: In
-                  values:
-                  - amd64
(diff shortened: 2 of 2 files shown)


Nokel81 (approved):

jonstelly:
I couldn't get lens building, I had WSL shell issues, some futex error, etc...

I did test this, but I had to do it by hand, I took all the lens-metrics yaml files converted the handlebars templating values and applied them all.  kube-state-metrics is now running on one of my arm64 nodes and everything came up cleanly.

NOTE: The kube-state-metrics image may be a release candidate, the release I linked to above is for RC1, but the image tag doesn't have RC in it.  But figured I should mention that for visibility.

Nokel81:
> NOTE: The kube-state-metrics image may be a release candidate, the release I linked to above is for RC1, but the image tag doesn't have RC in it. But figured I should mention that for visibility.

The github page seems to imply that we should be using https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-state-metrics instead.

jonstelly:
Do you mean lens should switch to using a helm chart to deploy kube-state-metrics or the whole lens-metrics extension?  The helm chart from the repository you link uses the same kube-state-metrics image as this PR.

FWIW, I like helm and would be fine with that.  Another thing that might warrant some discussion... the way that the lens-metrics extension is bundled up with lens, if there's a security issue found in one of the components (like mentioned above) then a new build/release of lens would need to be put together which doesn't seem ideal.  It also isn't something that's surfaced very well (is lens-metrics running the current/latest version of everything).  Instead of bundling the yaml files with a lens release, some way of showing and executing the helm command to deploy a lens-metrics chart that pulls the chart from a github repository would be ideal.  Security updates could be pushed to that chart repository without needing to rebuild lens just to pick up that update.  Just  a thought.

jakolehm (approved):
LGTM

PR #2853, opened 2021-05-24
Shorter kubeconfig-sync file label

jakolehm:
Takes less screen space and makes UI more calm.

image

### DIFF
--- a/src/main/catalog-sources/kubeconfig-sync.ts
+++ b/src/main/catalog-sources/kubeconfig-sync.ts
@@ -35,6 +35,7 @@ import { catalogEntityFromCluster } from "../cluster-manager";
 import { UserStore } from "../../common/user-store";
 import { ClusterStore, UpdateClusterModel } from "../../common/cluster-store";
 import { createHash } from "crypto";
+import { homedir } from "os";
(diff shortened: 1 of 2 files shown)


Nachasic (approved):

src/main/catalog-sources/kubeconfig-sync.ts
```diff
@@ -187,7 +188,9 @@ export function computeDiff(contents: string, source: RootSource, filePath: stri
 
           const entity = catalogEntityFromCluster(cluster);
 
-          entity.metadata.labels.file = filePath;
+          if (!filePath.startsWith(ClusterStore.storedKubeConfigFolder)) {
+            entity.metadata.labels.file = filePath.replace(homedir(), "~");
```
ixrock, line 192: Is it will be valid for Windows users too?

ixrock (approved):

src/main/catalog-sources/kubeconfig-sync.ts
jakolehm: At least it works on Powershell/Windows.

PR #2861, opened 2021-05-24
Show all node resources in node view

### LINKED ISSUE #2762: Show ALL node resources in node view
I'd like the node view to show ALL allocatable resource on the node, not just the basic CPU/mem/pod amounts.

**Why is this needed**:
In many cases kubelet reports devices, hugepages etc. as usable resources and having to dive into node yaml via edit button is, eh, bad experience. :)

Screenshot 2021-05-12 at 16 23 40

- Kubernetes distribution: k0s
- Desktop OS: OSX

vshakirova:
If I'm not mistaken, we don't always get the response in Ki

Screen Shot 2021-05-26 at 5 57 27 PM

Screen Shot 2021-05-26 at 5 59 04 PM

fixes: https://github.com/lensapp/lens/issues/2762

### DIFF
--- a/src/renderer/api/endpoints/nodes.api.ts
+++ b/src/renderer/api/endpoints/nodes.api.ts
@@ -56,39 +56,56 @@ export class Node extends KubeObject {
   static apiBase = "/api/v1/nodes";
 
   spec: {
-    podCIDR: string;
-    externalID: string;
+    podCIDR?: string;
+    podCIDRs?: string[];
+    providerID?: string;
+    /**
+     * @deprecated see https://issues.k8s.io/61966
+     */
+    externalID?: string;
     taints?: {
       key: string;
       value: string;
       effect: string;
+      timeAdded: string;
     }[];
     unschedulable?: boolean;
   };
   status: {
-    capacity: {
+    capacity?: {
       cpu: string;
+      ["ephemeral-storage"]: string;
+      ["hugepages-1Gi"]: string;
+      ["hugepages-2Mi"]: string;
       memory: string;
       pods: string;
     };
-    allocatable: {
+    allocatable?: {
       cpu: string;
+      ["ephemeral-storage"]: string;
+      ["hugepages-1Gi"]: string;
+      ["hugepages-2Mi"]: string;
       memory: string;
       pods: string;
     };
-    conditions: {
+    conditions?: {
       type: string;
-      status?: string;
+      status: string;
       lastHeartbeatTime?: string;
       lastTransitionTime?: string;
       reason?: string;
       message?: string;
     }[];
-    addresses: {
+    addresses?: {
       type: string;
       address: string;
     }[];
-    nodeInfo: {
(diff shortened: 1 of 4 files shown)


src/renderer/components/+nodes/node-details.tsx
```diff
@@ -90,13 +98,19 @@ export class NodeDetails extends React.Component<Props> {
         <KubeObjectMeta object={node} hideFields={["labels", "annotations", "uid", "resourceVersion", "selfLink"]}/>
         <DrawerItem name="Capacity">
           CPU: {capacity.cpu},{" "}
-          Memory: {Math.floor(parseInt(capacity.memory) / 1024)}Mi,{" "}
-          Pods: {capacity.pods}
+          Memory: {this.toMi(capacity.memory)},{" "}
+          Pods: {capacity.pods},{" "}
+          Ephemeral-storage: {this.toMi(capacity["ephemeral-storage"])},{" "}
+          Hugepages-1Gi: {capacity["hugepages-1Gi"]},{" "}
+          Hugepages-2Mi: {capacity["hugepages-2Mi"]}
```
Nokel81: Is it rather hard to read like this, could you make it into a table of sorts?

vshakirova: Wdyt about this?
Screen Shot 2021-05-26 at 11 37 05 AM

Nokel81: That looks amazing!

src/renderer/api/endpoints/nodes.api.ts
```diff
@@ -56,39 +56,53 @@ export class Node extends KubeObject {
   static apiBase = "/api/v1/nodes";
 
   spec: {
-    podCIDR: string;
-    externalID: string;
+    podCIDR?: string;
+    podCIDRs?: string[];
+    providerID?: string;
+    externalID: string; // Deprecated. see: https://issues.k8s.io/61966
```
Nokel81: If it deprecated then it should be marked as optional and maybe even add a comment.
```suggestion
    /**
     * @deprecated see https://issues.k8s.io/61966
     */
    externalID?: string;
```

src/renderer/components/+nodes/node-details-resources.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+interface Props {
+  node: Node;
+  type: "allocatable" | "capacity";
+}
+
+export class NodeDetailsResources extends React.Component<Props> {
+  toMi(resource: string) {
+    if (resource.endsWith("Ki")) {
+      return `${Math.floor(parseInt(resource) / 1024)}Mi`;
```
Nokel81: Would it be better to have some decimal points?
```suggestion
      return `${(parseInt(resource) / 1024).toFixed(1)}Mi`;
```

ixrock (approved):

Nokel81 (approved):

PR #2867, opened 2021-05-25
Fix makefile parallel jobs

jakolehm:
Otherwise `-jX` will error out when executed on a clean folder.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,12 +14,12 @@ else
     DETECTED_OS := $(shell uname)
 endif
 
-binaries/client: node_modules
-	yarn download-bins
-
 node_modules: yarn.lock
 	yarn install --frozen-lockfile --network-timeout=100000
 	yarn check --verify-tree --integrity
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -51,11 +51,6 @@ test: binaries/client
 
 .PHONY: integration-linux
 integration-linux: binaries/client build-extension-types build-extensions
-# ifdef XDF_CONFIG_HOME
```
jakolehm, line 54: @Nokel81 probably ok to remove these?

Nokel81: Probably, but if a developer has changed settings on a release of Lens then it might fail some integration tests.

jakolehm: I don't understand.. these lines were commented out 🤔

Nokel81: Oh oops, never mind then. It should be good to delete them.

ixrock (approved):

aleksfront (approved):

PR #2868, opened 2021-05-25
Remove unused Prettier package

Nokel81:


jim-docker (approved):

jakolehm:
I thought this was used by devs (via VSCode or similar)? Not sure if anyone is actually using it...

Nokel81:
@jakolehm Do you mean for formatting? We don't have a prettier config file. So I don't think it is being run ever. Though I might be wrong

jakolehm (approved):

PR #2879, opened 2021-05-26
Rework extension API into parts

Nokel81:
This is needed so that we can have separate versions of items, like the `CatalogEntityRegistry` and the `CatalogCategoryRegistry` for both `Renderer` and `Main` without having to come up with new and obtuse names.

- Main, Common, and Renderer

I have verified that this works on a fully clean build (including deleting all of the `OpenLensDev` Application Support folder.


extensions/kube-object-event-status/package/package.json
```diff
@@ -0,0 +1,25 @@
+{
```
jakolehm, line 1: Is this file added by accident?

extensions/kube-object-event-status/renderer.tsx
```diff
@@ -19,45 +19,45 @@
...
 import { resolveStatus, resolveStatusForCronJobs, resolveStatusForPods } from "./src/resolver";
 
-export default class EventResourceStatusRendererExtension extends LensRendererExtension {
+export default class EventResourceStatusRendererExtension extends Renderer.LensExtension {
   kubeObjectStatusTexts = [
     {
       kind: "Pod",
       apiVersions: ["v1"],
-      resolve: (pod: K8sApi.Pod) => resolveStatusForPods(pod)
+      resolve: (pod: Renderer.K8sApi.Pod) => resolveStatusForPods(pod)
```
jakolehm, line 30: We could make these less verbose with:

```ts
const K8sApi = Rendered.K8sApi;
```

src/extensions/common-api/index.ts
```diff
@@ -19,27 +19,21 @@
...
-export { LensMainExtension } from "../lens-main-extension";
-export { LensRendererExtension } from "../lens-renderer-extension";
-
 // APIs
 import * as App from "./app";
 import * as EventBus from "./event-bus";
 import * as Store from "./stores";
 import * as Util from "./utils";
-import * as Interface from "../interfaces";
+import * as Registrations from "./registrations";
```
jakolehm: This does not sounds good as top-level module. Would it make sense to have these as `Types`?

src/extensions/extension-discovery.ts
```diff
@@ -387,24 +387,19 @@ export class ExtensionDiscovery extends Singleton {
         await this.installPackage(extension.absolutePath);
       }
     }
-    const extensions = bundledExtensions.concat(userExtensions);
 
-    return this.extensions = new Map(extensions.map(extension => [extension.id, extension]));
+    return this.extensions = new Map(bundledExtensions.concat(userExtensions).map(extension => [extension.id, extension]));
```
jakolehm: These changes are out-of-scope for this PR. Could you revert unless they are really needed.

src/extensions/extension-loader.ts
```diff
@@ -337,7 +337,11 @@ export class ExtensionLoader extends Singleton {
     try {
       return __non_webpack_require__(extAbsolutePath).default;
     } catch (error) {
-      logger.error(`${logModule}: can't load extension main at ${extAbsolutePath}: ${error}`, { extension, error });
+      if (ipcRenderer) {
```
jakolehm, line 340: This also looks like unrelated change?

jakolehm (commented):
Great work! I think this restructure of extension api is needed. It clarifies api responsibilities and allows Lens to provide process specific modules (which are needed for Catalog improvements).

Are you going to provide docs update in a follow-up PR?

extensions/kube-object-event-status/renderer.tsx
ixrock: Or maybe better to split main/renderer via different module names to avoid verbosity? Wdyt?

```ts
import { K8sApi } from "@k8slens/extensions/main"
import { Components } from "@k8slens/extensions/renderer"
```

Nokel81: I tried to do what Roman was suggesting, but was not able to figure it out.

src/extensions/common-api/index.ts
Nokel81: Oh because they are not usually used by y themselves? Yeah in `Types` makes sense.

extensions/kube-object-event-status/package/package.json
Nokel81: Yes

src/extensions/extension-loader.ts
Nokel81: Sure I can revert these, though this error logging in particular I changed because I needed it while debugging why the extensions weren't being added.

Nokel81:
> Are you going to provide docs update in a follow-up PR?

Yes that is what I am thinking, since there was a possibility that they might be changes to the PR. So this way there is less work.

extensions/kube-object-event-status/renderer.tsx
Nokel81: @jakolehm unfortunately because `K8sApi` is a `namespace` there is no destructuring currently, or even aliasing. See https://github.com/microsoft/TypeScript/issues/13135

jakolehm (approved):

nevalla (approved):
LGTM

extensions/kube-object-event-status/package/package.json
jakolehm: @Nokel81 just noticed that this file was not removed ...

PR #2883, opened 2021-05-27
Change site url to api-docs.k8slens.dev

stevejr:
Signed-off-by: Steve Richards <srichards@mirantis.com>


Nokel81:
@stevejr Shouldn't most of this be generated and not stored in tree?

Nokel81:
Namely everything under `master/extensions/api`?

stevejr:
> Namely everything under `master/extensions/api`?

That is correct. The way GH pages works is that static HTML content is created by the mkdocs publish command. All of our docs live in master (and tagged releases) and then we also create the API Docs at build time (they never get checked back into the repo). 

However, as we have now split the docs and they are being served on 2 different URLs this PR is to make sure that any clickable items severed under https://api-docs.k8slens.dev resolve back to that url. For example, the lens icon at the top of every docs page has a href that comes from the `site_url` config value defined in mkdocs.yml. This value has been amended for the master branch and therefore the static html is correct but we are not cherry picking this update back into all released branches and re-deploying the docs. 

Does this make sense @Nokel81

Nokel81 (approved):

msa0311 (approved):

PR #2886, opened 2021-05-27
Fix: cluster commands disappeared from Command Palette

### LINKED ISSUE #2760: Cluster commands disappeared from Command Palette
There's no cluster commands available in list like
* `Cluster: View Secrets` 
* `Cluster: View Pods` 
* etc.

no cluster commands
no pods

ixrock:
**Global view:**
Screenshot 2021-05-27 at 14 55 47

**Cluster view:**
Screenshot 2021-05-27 at 14 56 04

fix #2760


src/renderer/components/cluster-manager/cluster-view.tsx
```diff
@@ -65,6 +67,7 @@ export class ClusterView extends React.Component {
         initView(clusterId); // init cluster-view (iframe), requires parent container #lens-views to be in DOM
         requestMain(clusterActivateHandler, clusterId, false); // activate and fetch cluster's state from main
         catalogEntityRegistry.activeEntity = catalogEntityRegistry.getById(clusterId);
+        this.store.setActive(clusterId);
```
ixrock: Setting active cluster was missing somehow at all (only in `ClusterStore.fromStore`)

src/extensions/registries/command-registry.ts
```diff
@@ -39,14 +38,6 @@ export interface CommandRegistration {
 }
 
 export class CommandRegistry extends BaseRegistry<CommandRegistration> {
-  @observable.ref activeEntity: CatalogEntity;
```
ixrock, line 42: Removed in favour `catalogEntityRegistry.activeEntity`

src/renderer/components/cluster-manager/cluster-view.tsx
ixrock: If we keep only this line and remove all other changes from the PR then cluster commands are shown, but navigation to corresponding pages is broken for some reasons..

Nokel81: Why do we need this, shouldn't it be possible to derive the active cluster from the active catalog entity?

jakolehm: I'm also wondering do we really need this?

ixrock: Question is why do we need `ClusterStore.setActive(id: string)`? 
It sets `ClusterStore.activeCluster` and used in:

```ts
  @computed get active(): Cluster | null {
    return this.getById(this.activeCluster);
  }
```

Current usages of `ClusterStore.active`:

image

ixrock: Should we use something like this instead?

```ts
  @computed get active(): Cluster | null {
    return this.getById(catalogEntityRegistry.activeEntity?.getId());
  }
```

Nokel81: I would get rid of `active()` from `ClusterStore` entirely. Since it has no use on `main` and from `renderer` it would be better to just use `ClusterStore.getInstance().getById(catalogEntityRegistry.activeEntity?.getId())` directly.

ixrock:
Refactored. @jakolehm @Nokel81 PTAL

src/common/cluster-store.ts
```diff
@@ -217,43 +214,14 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
...
-
   @computed get connectedClustersList(): Cluster[] {
     return this.clustersList.filter((c) => !c.disconnected);
   }
 
-  isActive(id: ClusterId) {
-    return this.activeCluster === id;
-  }
-
-  isMetricHidden(resource: ResourceType) {
```
ixrock, line 240: Moved to `cluster.ts`

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -39,14 +41,6 @@ export class CatalogEntityRegistry {
     });
   }
 
-  set activeEntity(entity: CatalogEntity) {
```
ixrock, line 42: This is removed cause doesn't add any extra logic on `set`. 
We can revert when we need more actions on assign.

jakolehm (approved):
LGTM

Nokel81 (approved):

PR #2887, opened 2021-05-27
Minor cleanup of ExtensionDiscovery

Nokel81:

### DIFF
--- a/src/extensions/extension-discovery.ts
+++ b/src/extensions/extension-discovery.ts
@@ -30,7 +30,7 @@ import { broadcastMessage, handleRequest, requestMain, subscribeToBroadcast } fr
 import { Singleton, toJS } from "../common/utils";
 import logger from "../main/logger";
 import { ExtensionInstallationStateStore } from "../renderer/components/+extensions/extension-install.store";
-import { extensionInstaller, PackageJson } from "./extension-installer";
+import { extensionInstaller } from "./extension-installer";
 import { ExtensionsStore } from "./extensions-store";
 import { ExtensionLoader } from "./extension-loader";
 import type { LensExtensionId, LensExtensionManifest } from "./lens-extension";
@@ -183,7 +183,7 @@ export class ExtensionDiscovery extends Singleton {
       .on("unlink", this.handleWatchUnlinkEvent);
   }
 
-  handleWatchFileAdd =  async (manifestPath: string) => {
+  handleWatchFileAdd = async (manifestPath: string) => {
     // e.g. "foo/package.json"
     const relativePath = path.relative(this.localFolderPath, manifestPath);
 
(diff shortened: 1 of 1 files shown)


src/extensions/extension-discovery.ts
```diff
@@ -387,24 +391,19 @@ export class ExtensionDiscovery extends Singleton {
...
    */
-  async installBundledPackages(packageJsonPath: string, extensions: InstalledExtension[]) {
-    const packagesJson: PackageJson = {
-      dependencies: {}
-    };
-
-    extensions.forEach((extension) => {
-      packagesJson.dependencies[extension.manifest.name] = extension.absolutePath;
-    });
+  installBundledPackages(packageJsonPath: string, extensions: InstalledExtension[]) {
```
nevalla: so this does not be to be async since `extensionInstaller.installPackages()` returns promise and await is not needed here?

nevalla (dismissed):

src/extensions/extension-discovery.ts
Nokel81: That is true. An `async function` wraps the body in a promise and treats thrown errors as `reject`'s. But also, in JS there are no "nested promises". If you write `Promise.resolve(Promise.resolve(4))` you have a `Promise<number>`.

jakolehm: I think we should add explicit return type here. It's not obvious now that `installBundledPackages` returns a promise.

src/extensions/extension-discovery.ts
```diff
@@ -234,24 +234,28 @@ export class ExtensionDiscovery extends Singleton {
       return;
     }
 
-    const extension = Array.from(this.extensions.values()).find((extension) => extension.absolutePath === filePath);
+    for (const extension of this.extensions.values()) {
```
jakolehm, line 237: This does not feel like an improvement.. why do we want `for` loop here?

src/extensions/extension-discovery.ts
```diff
@@ -387,31 +391,26 @@ export class ExtensionDiscovery extends Singleton {
         await this.installPackage(extension.absolutePath);
       }
     }
-    const extensions = bundledExtensions.concat(userExtensions);
 
-    return this.extensions = new Map(extensions.map(extension => [extension.id, extension]));
+    return this.extensions = new Map(bundledExtensions.concat(userExtensions).map(extension => [extension.id, extension]));
```
jakolehm: Does not really improve with readability.

src/extensions/extension-discovery.ts
```diff
@@ -125,15 +125,15 @@ export class ExtensionDiscovery extends Singleton {
   /**
    * Initializes the class and setups the file watcher for added/removed local extensions.
    */
-  async init() {
+  async init(): Promise<void> {
```
jakolehm, line 128: Explicit return types 👍

Nokel81: Would you be in favour of eventually requiring these using eslint?

src/extensions/extension-discovery.ts
Nokel81: Mainly because it is wasteful the create an array just to pick an item out of it. 

If you prefer I could have used a theoretical `iter.find()` function instead.

src/extensions/extension-discovery.ts
Nokel81: Okay will revert.

nevalla (commented):
LGTM

nevalla (approved):

PR #2888, opened 2021-05-27
Make a master tag of the extensions API

Nokel81:

### DIFF
--- a/.github/workflows/publish-master-npm.yml
+++ b/.github/workflows/publish-master-npm.yml
@@ -1,12 +1,16 @@
 name: Publish NPM Package `master`
 on:
-  push:
+  pull_request:
     branches:
       - master
+    types:
+      - closed
 jobs:
   publish:
     name: Publish NPM Package `master`
     runs-on: ubuntu-latest
+    if: github.event.pull_request.merged == true
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'area/extension') }}
     strategy:
       matrix:
         node-version: [12.x]
--- a/.github/workflows/publish-master-npm.yml
+++ b/.github/workflows/publish-master-npm.yml
@@ -1,0 +1,34 @@
+name: Publish NPM Package `master`
+on:
+  push:
+    branches:
+      - master
+jobs:
+  publish:
+    name: Publish NPM Package `master`
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [12.x]
+    steps:
+      - name: Checkout Release
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Using Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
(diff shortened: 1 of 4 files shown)


msa0311 (dismissed):

jakolehm:
Could someone explain how this actually works? If we publish a tag on each push to master what is the version where tag points to? We probably don't want to have a version for each push, right?

.github/workflows/publish-master-npm.yml
```diff
@@ -0,0 +1,34 @@
+name: Publish NPM Package `master`
+on:
+  push:
```
jakolehm: I think we don't want to publish a new npm package from every push. This should be "manually triggered" or something like that.

.github/workflows/publish-master-npm.yml
```diff
@@ -0,0 +1,34 @@
...
+          node-version: ${{ matrix.node-version }}
+
+      - name: Generate NPM package
+        run: |
+           make build-npm
+
+      - name: publish new release
+        if: contains(github.ref, 'refs/tags/v')
+        run: |
+           make publish-npm
```
jakolehm, line 35: `make publish-npm`  needs to generate different package.json version (for example `x.y.z-master.<git-sha>`) when `NPM_RELEASE_TAG=master`. This can be done in `build/set_npm_version.ts`.

jakolehm (changes requested):

.github/workflows/publish-master-npm.yml
Nokel81: Why not? Because master might be broken? If that is the case you can always choose a specific hash

jakolehm: It would flood npmjs with versions that have exactly the same contents (not a good practice). Not sure if we can detect here PR labels (on merge). If yes -> we could publish when PR has `area/extensions` label.

build/set_npm_version.ts
```diff
@@ -22,8 +22,20 @@ import * as fs from "fs";
...
+const version = new SemVer(appInfo.version);
 
-packageInfo.version = appInfo.version;
-fs.writeFileSync(packagePath, `${JSON.stringify(packageInfo, null, 2)}\n`);
+if (NPM_RELEASE_TAG !== "latest") {
+  const gitRef = execSync("git rev-parse --short HEAD", {
+    encoding: "utf-8",
+  });
+
+  version.inc("prerelease", gitRef.trim());
```
jakolehm:
```suggestion
  version.inc("prerelease", `git.${gitRef.trim()}`);
```

Nokel81:
@jakolehm From reading the github action docs I think I have come up with a way to do it on when `pull_request`'s to `master` are merged and have the correct label.

jakolehm (approved):
LGTM

jakolehm:
@msa0311 ptal

msa0311 (approved):

PR #2272, opened 2021-03-02
Add common app routes to the protocol renderer router

Nokel81:
- Add most of the links to the docs so that users can use them.

Signed-off-by: Sebastian Malton <sebastian@malton.name>

fixes #2236

PR #2890, opened 2021-05-27
Fix Navigation.navigate to work with fragments

nevalla:
Revert change made in https://github.com/lensapp/lens/pull/2272 so that now `Navigation.navigate` works also when navigating to another detailsUrl from detailsUrl.


src/renderer/navigation/helpers.ts
```diff
@@ -20,17 +20,20 @@
...
 import type { LocationDescriptor } from "history";
+import { createPath } from "history";
 import { matchPath, RouteProps } from "react-router";
 import { PageParam, PageParamInit } from "./page-param";
 import { clusterViewRoute, IClusterViewRouteParams } from "../components/cluster-manager/cluster-view.route";
 import { navigation } from "./history";
 
 export function navigate(location: LocationDescriptor) {
-  const currentLocation = navigation.location.pathname;
+  const currentLocation = createPath(navigation.location);
```
Nokel81, line 30: So this basically produces `/path?and=query#withHash`?

Nokel81 (approved):

src/renderer/navigation/helpers.ts
nevalla: I haven't tested `#withHash`, but at least query params should work now.

jim-docker (approved):

src/renderer/navigation/helpers.ts
Nokel81: I have tested and it does.

PR #2892, opened 2021-05-27
Move allowed-resource storage out of `Cluster`

Nokel81:
To facilitate point 4 on https://github.com/lensapp/lens/issues/2663 we need to remove the dependency of `renderer` on `Cluster` instances. This is part of that.

This works by caching the allowed resources per namespace on the `main` side and returning those values to `renderer` whenever it asks for it. 

Going forward we probably should get rid of `View all Namespces` for clusters that are not admins (since it is such a resource hog) so this still will facilitate that since `""` means `"all namespaces"` from the APIs POV.

Breaking Changes:
---
- The function `isAllowedResource` now takes the argument `type { apiBase: string }` instead of `string | string[]`. This is done so that we can easily do the following call: `isAllowedResource(Pod)` (the `Pod` type from `pods.api.ts`. This is done because we are now collecting a complete list of allowed resources, and `resourceName`'s are not unique globally. They are only unique for a given `groupVersion`.


src/common/cluster-ipc.ts
```diff
@@ -102,4 +102,30 @@ if (ipcMain) {
...
+  handleRequest(ClusterResourceIsAllowedChannel, async (event, clusterId: ClusterId, namespaces: string[]): Promise<[string, boolean][]> => {
+    const cluster = ClusterStore.getInstance().getById(clusterId);
+
+    if (!cluster) {
+      return [];
+    }
+
+    const isAllowed = new Map<string, boolean>();
+
+    await Promise.all(
```
jakolehm: Throttling happens somewhere in `cluster.getIsAllowedResources`, right?

src/common/utils/async-throttle.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+export function asyncThrottle<Fn extends (...args: any[]) => Promise<any>>(fn: Fn, cooldownPeriod: number): Fn {
```
jakolehm, line 22: I think we should always write test cases for util methods that are complex enough. This is complex enough.

src/common/utils/extended-map.ts
```diff
@@ -67,6 +67,10 @@ export class ExtendedMap<K, V> extends Map<K, V> {
 }
 
 export class ExtendedObservableMap<K, V> extends ObservableMap<K, V> {
+  static new<K, V>(initialData?: IObservableMapInitialValues<K, V>, enhancer?: IEnhancer<V>, name?: string): ExtendedObservableMap<K, V> {
```
jakolehm, line 73: Some explanation would be nice 🙂 (maybe in form of tests)

src/common/utils/async-throttle.ts
jakolehm: Also wondering do we really need this... `debounce` from `lodash` does the same thing, right?.

package.json
```diff
@@ -180,7 +180,7 @@
   "dependencies": {
     "@hapi/call": "^8.0.0",
     "@hapi/subtext": "^7.0.3",
-    "@kubernetes/client-node": "^0.12.0",
+    "@kubernetes/client-node": "^0.14.3",
```
jakolehm: Why this is needed?

src/main/utils/api-resources.ts
```diff
@@ -0,0 +1,73 @@
...
+  verbs: Set<string>,
+  shortNames: Set<string>,
+}
+
+/**
+ * Mapping between groupVersions and resource names and their information
+ */
+export type ApiResourceMap = Map<string, Map<string, ApiResource>>;
+
+export async function getClusterResources(kc: KubeConfig): Promise<ApiResourceMap> {
```
jakolehm: Needs a bit more explanation.

src/renderer/api/allowed-resources.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  constructor(protected clusterId: ClusterId, protected getNamespaces: () => NamespaceName[]) {
+    super();
+  }
+
+  async init() {
+    this.resources = await requestMain(ClusterGetResourcesChannel, this.clusterId);
+    this.allowedResourceMap.replace(await requestMain(ClusterResourceIsAllowedChannel, this.clusterId, this.getNamespaces()));
+
+    console.log(this.allowedResourceMap);
```
jakolehm: Leftover?

src/renderer/api/allowed-resources.ts
```diff
@@ -0,0 +1,93 @@
...
+import { ObservableTimer } from "../../common/utils/observable-timer";
+
+type NamespaceName = string;
+type ResourceName = string;
+
+export class AllowedResources extends Singleton {
+  protected allowedResourceMap = new ObservableMap<ResourceName, boolean>();
+  public resources: ApiResourceMap;
+  protected timer = new ObservableTimer(60 * 1000);
+  disopser: Disposer;
```
jakolehm:
```suggestion
  disposer: Disposer;
```

src/renderer/api/allowed-resources.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  async init() {
+    this.resources = await requestMain(ClusterGetResourcesChannel, this.clusterId);
+    this.allowedResourceMap.replace(await requestMain(ClusterResourceIsAllowedChannel, this.clusterId, this.getNamespaces()));
+
+    console.log(this.allowedResourceMap);
+
+    this.disopser = reaction(() => [this.timer.tickCount, this.getNamespaces()] as const, async ([, namespaces]) => {
+      this.allowedResourceMap.replace(await requestMain(ClusterResourceIsAllowedChannel, this.clusterId, namespaces));
```
jakolehm: Should we handle errors?

src/renderer/components/+workloads/workloads.tsx
```diff
@@ -115,7 +116,7 @@ export class Workloads extends React.Component {
 
   render() {
     return (
-      <TabLayout className="Workloads" tabs={Workloads.tabRoutes}/>
+      <TabLayout className="Workloads" tabs={Workloads.tabRoutes} />
```
jakolehm, line 121: Are you using some formatter for these whitespace fixes? It adds a bit noise to the PR.

types/@jcoreio/async-throttle.d.ts
```diff
@@ -0,0 +1,36 @@
...
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+declare module "@jcoreio/async-throttle";
+
+export default throttle;
+
+declare function throttle<F extends (...args: any[]) => Promise<any>>(
```
jakolehm: What is this and why it's included? 🤔

src/common/cluster-ipc.ts
Nokel81: Yes, and less throttling and more caching. But it should be the same effect.

src/common/utils/async-throttle.ts
Nokel81: Will most definitely write tests for this. The difference between this and `debounce` or `throttle` from `lodash` is that this function won't call `fn` again until the first promise has resolved. That way if the `cooldownPeriod` is less than the amount of time one of the invocations takes to complete we won't start having many inflight requests at a time.

src/common/utils/extended-map.ts
Nokel81: Do you mean for this function? Those are basically the types from the constructor of `observable.map`.

src/renderer/components/+workloads/workloads.tsx
Nokel81: I think I had vscode's default formatter turned on. I have turned it off. I'll revert these changes.

package.json
Nokel81: I upgraded because it was available and I was trying out to see if it naively supported the getting of all resource lists (including legacy APIs).

It didn't but since we now also support 1.21 for kube, it seems like we should keep the upgrade. But I can understand if it should be its own PR.

nevalla:
Are there anything that we can do to improve the flow? Currently I can see a loader, when auth proxy is starting, then empty screen for a moment until the dashboard starts loading (with loader). After the dashboard is visible, it will be refreshed again with all allowed resources. This is a bit clumsy now.

nevalla:
Tested this more. Flow is a bit better now. Noticed one bug. Now when cluster dashboard is opened, workloads overview is opened always even for cluster admin user.

Other thing that I'm worried about is the performance drop we get with this. With the latest beta, user will see cluster dashboard in ~ 3.5s. With these changes it will take ~8.5s. That's quite huge drop.

Edit: tested also `master` branch and result is ~5.5s.

Edit2: Played a bit also with built version and the difference is not that dramatic anymore: ~ 4.5s. 

Most of the times is used when waiting iframe onload event.

jakolehm:
> Tested this more. Flow is a bit better now. Noticed one bug. Now when cluster dashboard is opened, workloads overview is opened always even for cluster admin user.
> 
> Other thing that I'm worried about is the performance drop we get with this. With the latest beta, user will see cluster dashboard in ~ 3.5s. With these changes it will take ~8.5s. That's quite huge drop.
> 
> Edit: tested also `master` branch and result is ~5.5s.

IMHO sounds like too big performance regression.

src/renderer/components/app.tsx
```diff
@@ -112,9 +116,14 @@ export class App extends React.Component {
         preload: true,
       })
     ]);
+
+    setTimeout(() => {
```
nevalla: how about if `AllowedResources` had some `loading` observable and we would display loader while it's true?

src/renderer/components/app.tsx
```diff
@@ -112,9 +116,14 @@ export class App extends React.Component {
...
+      // This is here so that the rest of react can respond to AllowedResources loading
+      this.isLoading = false;
+    }, 2000);
   }
 
-  @observable startUrl = isAllowedResource(["events", "nodes", "pods"]) ? routes.clusterURL() : routes.workloadsURL();
+  @observable startUrl = isAllowedResources("events", "nodes", "pods") ? routes.clusterURL() : routes.workloadsURL();
```
nevalla: this is a bit problematic, since `<Redirect>` component rendered before allowed resources are resolved.

src/renderer/components/app.tsx
Nokel81: How about it is loading only during the first load?

src/renderer/components/app.tsx
Nokel81: Will fix this

Nokel81:
@jakolehm Lauri found that it wasn't as bad on a release build. Will see if I can make it less bad.

nevalla:
> @jakolehm Lauri found that it wasn't as bad on a release build. Will see if I can make it less bad.

Also that timeout might explain something

Nokel81:
> Also that timeout might explain something

Yes... I will see if tracking the initial load in `AllowedResources` is enough.

Nokel81:
@nevalla I have removed the timeout, and found that the issue was with needing to load the allowed resources after setting the contexts for the `namespaceStore`. 

On a dev build it is now `~6s` so I think that 2s delay was the major problem.

I also resolved the "workspaces" being open issue.

nevalla:
Please could you test also with limited access to the cluster, for example with `viewer` role. I have a feeling that something is off there.

Nokel81:
@nevalla sure I can test that

Nokel81:
@nevalla I found the problem, I should have been using the `resource` field instead of the `name` field in the `this.canI` call.

src/renderer/components/app.tsx
```diff
@@ -69,12 +68,14 @@ import { Nodes } from "./+nodes";
...
+import { AllowedResources, isAllowedResources } from "../api/allowed-resources";
+import { CubeSpinner } from "./spinner";
 
 @observer
 export class App extends React.Component {
-  constructor(props: {}) {
-    super(props);
-    makeObservable(this);
+  @computed
+  static get startUrl(): string {
```
ixrock, line 76: decorators doesn't work with `static` in `mobx-6`

src/renderer/components/+user-management/user-management.tsx
```diff
@@ -35,7 +36,7 @@ import { ClusterRoles } from "./+cluster-roles";
 
 @observer
 export class UserManagement extends React.Component {
-  static get tabRoutes() {
+  @computed static get tabRoutes() {
```
ixrock: decorators doesn't work with `static` in `mobx-6`

src/renderer/components/+workloads/workloads.tsx
```diff
@@ -31,21 +32,14 @@ import { DaemonSets } from "../+workloads-daemonsets";
...
-  static get tabRoutes(): TabLayoutRoute[] {
-    const tabs: TabLayoutRoute[] = [
-      {
-        title: "Overview",
-        component: WorkloadsOverview,
-        url: routes.overviewURL(),
-        routePath: routes.overviewRoute.path.toString()
-      }
-    ];
+  @computed static get tabRoutes(): TabLayoutRoute[] {
```
ixrock: here too

src/renderer/components/app.tsx
Nokel81: How so? No errors are thrown and it seems to work as intended.

The docs say "Decorating static class members" are not supported. A getter is not a class member.

ixrock: It works as intended just because it's a `getter` :) == just no caching/computations

ixrock: If i got correctly "as intended" == "still observable"?

Nokel81:
Closing due to the load time issues (a different solution is needed) and because of conflicts

PR #2893, opened 2021-05-27
Remove old extensions on build

Nokel81:
Why is this needed?:
-
This is needed to prevent build error when switching between branch the contain differing sets of in-tree extensions. This removes the folders that no longer include `package.json` files. This is to prevent this step from removing extensions that are currently being worked on.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,9 @@ $(extension_dists): src/extensions/npm/extensions/dist
 
 .PHONY: clean-old-extensions
 clean-old-extensions:
+ifndef CI
 	find extensions -d 1 -type d '!' -exec test -e '{}/package.json' \; -delete
+endif
 
 .PHONY: build-extensions
 build-extensions: node_modules clean-old-extensions $(extension_node_modules) $(extension_dists)
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ $(extension_dists): src/extensions/npm/extensions/dist
 
 .PHONY: clean-old-extensions
 clean-old-extensions:
-	find extensions -d 1 -type d '!' -exec test -e "{}/package.json" \; -delete
+	find extensions -d 1 -type d '!' -exec test -e '{}/package.json' \; -delete
 
 .PHONY: build-extensions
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -82,8 +82,12 @@ $(extension_node_modules): node_modules
 $(extension_dists): src/extensions/npm/extensions/dist
 	cd $(@:/dist=) && ../../node_modules/.bin/npm run build
 
+.PHONY: clean-old-extensions
+clean-old-extensions:
+	./scripts/clean-old-extensions.sh
```
jakolehm: Why separate file vs calling `find` directly here?

```suggestion
	find extensions -d 1 -type d '!' -exec test -e "{}/package.json" \; -exec rm -r {} \; || exit 0
```

Nokel81: I did it because I was initially using a different mechanism but then changed it to use `find`.

Had to do some more experimentation because `||` doesn't work in `Makefile` and I was getting errors without it.

jakolehm (dismissed):

jakolehm:
Tests seem to fail.

Nachasic (dismissed):

Makefile
```diff
@@ -82,8 +82,14 @@ $(extension_node_modules): node_modules
 $(extension_dists): src/extensions/npm/extensions/dist
 	cd $(@:/dist=) && ../../node_modules/.bin/npm run build
 
+.PHONY: clean-old-extensions
+clean-old-extensions:
+ifndef CI
+	find extensions -d 1 -type d '!' -exec test -e '{}/package.json' \; -delete
```
jakolehm: Why do we need to skip this on CI? Is this incompatible with GNU `find`? How about Windows?

Nokel81: Let me check this on my ubuntu VM.

In theory this doesn't need to run on CI ever, which is why I went with this solution, since this is really a developer helper when switching branches (which CI never does).

Nokel81: Okay so GNU find and BSD find are not compatible with these options. I have found a new version that should work.

Nachasic (approved):
Looks pretty great, doing this by hand was kind of a headache

PR #2894, opened 2021-05-27
Upgrade to typescript 4.3

Nokel81:

### DIFF
--- a/src/renderer/utils/readableStream.ts
+++ b/src/renderer/utils/readableStream.ts
@@ -21,6 +21,18 @@
 
 import { Readable } from "readable-stream";
 
+type TypeArray = Int8Array
+  | Uint8Array
+  | Uint8ClampedArray
+  | Int16Array
+  | Uint16Array
+  | Int32Array
+  | Uint32Array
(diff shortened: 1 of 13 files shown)


nevalla (approved):

src/renderer/utils/readableStream.ts
```diff
@@ -21,6 +21,18 @@
...
+  | Uint8ClampedArray
+  | Int16Array
+  | Uint16Array
+  | Int32Array
+  | Uint32Array
+  | Float32Array
+  | Float64Array
+  | BigInt64Array
+  | BigUint64Array;
+
```
ixrock, line 35: FYI: we have `type-fest` package in the project where this available as `TypedArray`.

ixrock (approved):

src/renderer/utils/readableStream.ts
Nokel81: Oh cool, I was slightly annoyed that it wasn't part of the default typings. I think we can probably merge this and then change it in a follow up PR.

PR #2902, opened 2021-05-28
Fix main->renderer catalog entity sync issue

jakolehm:
Previously renderer always created a new entity instance when `registry.items` was called. This had side-effects on observability.

## Changes

- update existing entity if found from registry
- debounce catalog broadcast on main, otherwise we trigger a broadcast on every entity update

### DIFF
--- a/src/main/catalog-pusher.ts
+++ b/src/main/catalog-pusher.ts
@@ -24,10 +24,17 @@ import { broadcastMessage } from "../common/ipc";
 import type { CatalogEntityRegistry } from "./catalog";
 import "../common/catalog-entities/kubernetes-cluster";
 import { toJS } from "../common/utils";
+import { debounce } from "lodash";
+import type { CatalogEntity } from "../common/catalog";
+
+
+const broadcaster = debounce((items: CatalogEntity[]) => {
+  broadcastMessage("catalog:items", items);
+}, 1_000, { trailing: true });
 
 export function pushCatalogToRenderer(catalog: CatalogEntityRegistry) {
   return reaction(() => toJS(catalog.items), (items) => {
-    broadcastMessage("catalog:items", items);
+    broadcaster(items);
   }, {
     fireImmediately: true,
   });
--- a/src/renderer/api/catalog-entity-registry.ts
+++ b/src/renderer/api/catalog-entity-registry.ts
@@ -27,8 +27,13 @@ import type { Cluster } from "../../main/cluster";
 import { ClusterStore } from "../../common/cluster-store";
 
 export class CatalogEntityRegistry {
-  protected _entities = observable.map<string, CatalogEntity>([], { deep: true });
   @observable.ref activeEntity: CatalogEntity;
+  protected _entities = observable.map<string, CatalogEntity>([], { deep: true });
+
+  /**
+   * Buffer for keeping entities that don't yet have CatalogCategory synced
+   */
+  protected rawEntities: (CatalogEntityData & CatalogEntityKindData)[] = [];
 
   constructor(private categoryRegistry: CatalogCategoryRegistry) {
     makeObservable(this);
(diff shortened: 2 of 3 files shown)


src/renderer/api/catalog-entity-registry.ts
```diff
@@ -19,38 +19,59 @@
...
 
   init() {
     subscribeToBroadcast("catalog:items", (ev, items: (CatalogEntityData & CatalogEntityKindData)[]) => {
-      this.rawItems.replace(items);
+      this.updateItems(items);
+    });
+  }
+
+  @action updateItems(items: (CatalogEntityData & CatalogEntityKindData)[]) {
+    const newIds = items.map((item) => item.metadata.uid);
```
Nokel81:
```suggestion
    const newIds = new Set(items.map((item) => item.metadata.uid));
```

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -19,38 +19,59 @@
...
+
+    Array.from(this._entities.keys()).forEach((uid) => {
+      if (!newIds.includes(uid)) this._entities.delete(uid);
+    });
```
Nokel81:
```suggestion
    for (const uid of this._entities.keys()) {
      if (!newIds.has(uid) {
        this._entities.delete(uid);
      }
    }
```

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -19,38 +19,59 @@
...
+
+    items.forEach((item) => {
+      const existing = this._entities.get(item.metadata.uid);
+
+      if (!existing) {
+        const entity = this.categoryRegistry.getEntityForData(item);
+
+        if (entity) this._entities.set(entity.metadata.uid, entity);
```
Nokel81: This breaks the race condition I was talking about earlier. We still want items that were previously added to appear.

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -19,38 +19,59 @@
...
+  }
+
+  @action updateItems(items: (CatalogEntityData & CatalogEntityKindData)[]) {
+    const newIds = items.map((item) => item.metadata.uid);
+
+    Array.from(this._entities.keys()).forEach((uid) => {
+      if (!newIds.includes(uid)) this._entities.delete(uid);
+    });
+
+    items.forEach((item) => {
```
Nokel81: I would use a `for..of` loop here as well.

src/renderer/api/catalog-entity-registry.ts
Nokel81: This would require keeping `rawItems` around

Nokel81 (changes requested):

src/renderer/api/catalog-entity-registry.ts
jakolehm: Good point, refactored.

Nokel81 (approved):

src/main/catalog-pusher.ts
```diff
@@ -24,10 +24,17 @@ import { broadcastMessage } from "../common/ipc";
...
+const broadcaster = debounce((items: CatalogEntity[]) => {
+  broadcastMessage("catalog:items", items);
+}, 1_000, { trailing: true });
 
 export function pushCatalogToRenderer(catalog: CatalogEntityRegistry) {
   return reaction(() => toJS(catalog.items), (items) => {
-    broadcastMessage("catalog:items", items);
+    broadcaster(items);
   }, {
     fireImmediately: true,
```
ixrock, line 39: Probably debounce can be reached with `delay: 1000`

```suggestion
    fireImmediately: true,
    delay: 1000,
```

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -19,38 +19,90 @@
...
 import { CatalogCategory, catalogCategoryRegistry, CatalogCategoryRegistry, CatalogEntity, CatalogEntityData, CatalogEntityKindData } from "../../common/catalog";
 import "../../common/catalog-entities";
-import { iter } from "../utils";
 import type { Cluster } from "../../main/cluster";
 import { ClusterStore } from "../../common/cluster-store";
 
 export class CatalogEntityRegistry {
-  protected rawItems = observable.array<CatalogEntityData & CatalogEntityKindData>();
-  @observable.ref activeEntity?: CatalogEntity;
+  @observable.ref activeEntity: CatalogEntity;
```
ixrock, line 30: Why removed `?` I guess `activeEntity` still could be `undefined`, right?

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -19,38 +19,90 @@
...
+  protected _entities = observable.map<string, CatalogEntity>([], { deep: true });
```
ixrock, line 31: nit:`{deep: true}` is default mobx' observability option :)

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -19,38 +19,90 @@
...
+
+  /**
+   * Buffer for keeping entities that don't yet have CatalogCategory synced
+   */
+  protected rawEntities: (CatalogEntityData & CatalogEntityKindData)[] = [];
```
ixrock, line 36: Maybe create a new `interface` / `type` and give a name to `CatalogEntityData & CatalogEntityKindData`?

ixrock (approved):
LGTM

src/renderer/api/catalog-entity-registry.ts
Nokel81: `RawCatalogEntityData`?

PR #2909, opened 2021-05-31
Improve how kubernetes cluster status is resolved

jakolehm:
- mark as active only when dashboard is visible
- mark as connected only if connection is in usable state

### DIFF
--- a/src/common/catalog-entities/kubernetes-cluster.ts
+++ b/src/common/catalog-entities/kubernetes-cluster.ts
@@ -53,8 +53,11 @@ export interface KubernetesClusterStatus extends CatalogEntityStatus {
 }
 
 export class KubernetesCluster extends CatalogEntity<CatalogEntityMetadata, KubernetesClusterStatus, KubernetesClusterSpec> {
-  public readonly apiVersion = "entity.k8slens.dev/v1alpha1";
-  public readonly kind = "KubernetesCluster";
+  static apiVersion = "entity.k8slens.dev/v1alpha1";
+  static kind = "KubernetesCluster";
+
+  public readonly apiVersion = KubernetesCluster.apiVersion;
+  public readonly kind = KubernetesCluster.kind;
 
   async connect(): Promise<void> {
     if (app) {
--- a/src/common/catalog-entities/kubernetes-cluster.ts
+++ b/src/common/catalog-entities/kubernetes-cluster.ts
@@ -53,11 +53,8 @@ export interface KubernetesClusterStatus extends CatalogEntityStatus {
 }
 
 export class KubernetesCluster extends CatalogEntity<CatalogEntityMetadata, KubernetesClusterStatus, KubernetesClusterSpec> {
-  static apiVersion = "entity.k8slens.dev/v1alpha1";
-  static kind = "KubernetesCluster";
-
-  public readonly apiVersion = KubernetesCluster.apiVersion;
-  public readonly kind = KubernetesCluster.kind;
+  public readonly apiVersion = "entity.k8slens.dev/v1alpha1";
+  public readonly kind = "KubernetesCluster";
 
   async connect(): Promise<void> {
     if (app) {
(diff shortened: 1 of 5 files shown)


src/main/cluster-manager.ts
```diff
@@ -101,6 +100,10 @@ export class ClusterManager extends Singleton {
     }
   }
 
+  protected updateEntityStatus(entity: KubernetesCluster, cluster: Cluster) {
+    entity.status.phase = (cluster.ready && cluster.available && cluster.allowedNamespaces.length > 0) ? "connected" : "disconnected";
```
nevalla: so we display cluster as disconnected when no allowed namespaces present?

jakolehm: Yeah, not bullet-proof but currently we mark cluster as connected even it's not connected at all (auth failures etc). If you have better idea to check "connection in usable state" then I'm happy to change this 🙂

Nokel81: What is wrong with using `Cluster.accessible`?

Nokel81: Because if that is wrong then we should change it instead. Though TBH we should have a lot more derived state (which should be possible once we remove the need to sync Clusters to renderer).

src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -53,8 +53,11 @@ export interface KubernetesClusterStatus extends CatalogEntityStatus {
 }
 
 export class KubernetesCluster extends CatalogEntity<CatalogEntityMetadata, KubernetesClusterStatus, KubernetesClusterSpec> {
-  public readonly apiVersion = "entity.k8slens.dev/v1alpha1";
-  public readonly kind = "KubernetesCluster";
+  static apiVersion = "entity.k8slens.dev/v1alpha1";
+  static kind = "KubernetesCluster";
```
Nokel81: Why does this class need them? To make sure that the catalogue registry is set up correctly?

src/main/cluster-manager.ts
jakolehm: > What is wrong with using `Cluster.accessible`?

Nothing, somehow missed that attribute. Now it's using just `cluster.accessible`.

nevalla (dismissed):

src/main/cluster-manager.ts
```diff
@@ -118,10 +121,7 @@ export class ClusterManager extends Singleton {
         cluster.kubeConfigPath = entity.spec.kubeconfigPath;
         cluster.contextName = entity.spec.kubeconfigContext;
 
-        entity.status = {
-          phase: cluster.disconnected ? "disconnected" : "connected",
-          active: !cluster.disconnected
-        };
+        this.updateEntityStatus(entity, cluster);
```
Nokel81, line 124: Wouldn't it be better to pass through `cluster.failureReason` too. That way `cluster-status` doesn't need to access `ClusterStore` at all.

src/common/catalog-entities/kubernetes-cluster.ts
jakolehm: This class does not need them, I added them so that other places can use these: https://github.com/lensapp/lens/pull/2909/files#diff-5f80bd77276864216b34ae50b0db19ba251466314ae4cc141f36b8169611c9beR58

src/main/cluster-manager.ts
jakolehm: Maybe, I'll leave that to a separate PR.

src/common/catalog-entities/kubernetes-cluster.ts
Nokel81: Couldn't that have been `entity.apiVersion` and `entity.kind`?

jakolehm: Oh, of course. Idk what I was thinking here. Changed.

src/main/cluster-manager.ts
Nokel81: Sounds good

Nokel81 (approved):

nevalla (approved):

PR #2919, opened 2021-05-31
Selecting one namespace at a time

### LINKED ISSUE #2773: Selecting one namespace at a time
This is about the namespace dropdown menu in this part of the UI:
image

I want to be able to select one namespace at a time, much like you would in a standard dropdown menu. This could be implemented as Alt+Left click for example.

**Why is this needed**:
I'm usually only interested in one namespace at the same time, so I always end up toggling one namespace on and the previous namespace off. This could be more efficient and it actually makes the toggle behavior of the dropdown menu less intuitive than `kubectl`'s behavior of showing one namespace unless you explicitly ask for more.

- Kubernetes distribution: Amazon EKS
- Desktop OS: Ubuntu

vshakirova:
select one namespace by default, using "shift clicking" as a way to select multiple namespaces.
select all namespaces in list when clicked `All namespaces`
duplicate Shift behavior to Cmd (on mac) and Ctrl on windows which is more intuitive on win.

react-select contains "onKeyDown" props, but doesn't contain "onKeyUp" and "onClick". Wdyt? 

fixes: https://github.com/lensapp/lens/issues/2773

### DIFF
--- a/src/renderer/components/+namespaces/namespace-select-filter.tsx
+++ b/src/renderer/components/+namespaces/namespace-select-filter.tsx
@@ -30,6 +30,8 @@ import { NamespaceSelect } from "./namespace-select";
 import { namespaceStore } from "./namespace.store";
 
 import type { SelectOption, SelectProps } from "../select";
+import { isLinux, isMac, isWindows } from "../../../common/vars";
+import { observable } from "mobx";
 
 const Placeholder = observer((props: PlaceholderProps<any, boolean>) => {
   const getPlaceholder = (): React.ReactNode => {
@@ -55,13 +57,16 @@ const Placeholder = observer((props: PlaceholderProps<any, boolean>) => {
 
 @observer
 export class NamespaceSelectFilter extends React.Component<SelectProps> {
+  static isMultiSelection = observable.box(false);
+  static isMenuOpen = observable.box(false);
+
   formatOptionLabel({ value: namespace, label }: SelectOption) {
     if (namespace) {
       const isSelected = namespaceStore.hasContext(namespace);
 
       return (
         <div className="flex gaps align-center">
-          <Icon small material="layers" />
+          <Icon small material="layers"/>
           <span>{namespace}</span>
           {isSelected && <Icon small material="check" className="box right"/>}
         </div>
@@ -72,26 +77,55 @@ export class NamespaceSelectFilter extends React.Component<SelectProps> {
   }
 
   onChange([{ value: namespace }]: SelectOption[]) {
-    if (namespace) {
+    if (NamespaceSelectFilter.isMultiSelection.get() && namespace) {
       namespaceStore.toggleContext(namespace);
+    } else if (!NamespaceSelectFilter.isMultiSelection.get() && namespace) {
+      namespaceStore.toggleSingle(namespace);
     } else {
-      namespaceStore.toggleAll(false); // "All namespaces" clicked
+      namespaceStore.toggleAll(true); // "All namespaces" clicked
     }
   }
 
+  onKeyDown = (e: React.KeyboardEvent<any>) => {
+    if (isMac && e.metaKey || (isWindows || isLinux) && e.ctrlKey) {
+      NamespaceSelectFilter.isMultiSelection.set(true);
+    }
+  };
+
+  onKeyUp = (e: React.KeyboardEvent<any>) => {
+    if (isMac && e.key === "Meta" || (isWindows || isLinux) && e.key === "Control") {
+      NamespaceSelectFilter.isMultiSelection.set(false);
+    }
+  };
+
+  onClick = () => {
+    if (!NamespaceSelectFilter.isMultiSelection.get()) {
+      NamespaceSelectFilter.isMenuOpen.set(!NamespaceSelectFilter.isMenuOpen.get());
+    }
+  };
+
+  reset = () => {
+    NamespaceSelectFilter.isMultiSelection.set(false);
+    NamespaceSelectFilter.isMenuOpen.set(false);
+  };
+
   render() {
     return (
-      <NamespaceSelect
-        isMulti={true}
-        components={{ Placeholder }}
-        showAllNamespacesOption={true}
-        closeMenuOnSelect={false}
-        controlShouldRenderValue={false}
-        placeholder={""}
-        onChange={this.onChange}
-        formatOptionLabel={this.formatOptionLabel}
-        className="NamespaceSelectFilter"
-      />
+      <div onKeyUp={this.onKeyUp} onKeyDown={this.onKeyDown} onClick={this.onClick}>
+        <NamespaceSelect
+          isMulti={true}
+          menuIsOpen={NamespaceSelectFilter.isMenuOpen.get()}
+          components={{ Placeholder }}
+          showAllNamespacesOption={true}
+          closeMenuOnSelect={false}
+          controlShouldRenderValue={false}
+          placeholder={""}
+          onChange={this.onChange}
+          onBlur={this.reset}
+          formatOptionLabel={this.formatOptionLabel}
+          className="NamespaceSelectFilter"
+        />
+      </div>
     );
   }
 }
--- a/src/renderer/components/+namespaces/namespace.store.ts
+++ b/src/renderer/components/+namespaces/namespace.store.ts
@@ -159,6 +159,12 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
   }
 
   @action
+  toggleSingle(namespace: string){
+    this.clearSelected();
+    this.selectNamespaces(namespace);
+  }
+
+  @action
   toggleAll(showAll?: boolean) {
     if (typeof showAll === "boolean") {
       if (showAll) {


src/renderer/components/+namespaces/namespace-select-filter.tsx
```diff
@@ -57,6 +57,8 @@ const Placeholder = observer((props: PlaceholderProps<any>) => {
 
 @observer
 export class NamespaceSelectFilter extends React.Component<SelectProps> {
+  static isMulti = false;
```
Nokel81: This should be observable
```suggestion
  static isMulti = observable.box(false);
```

src/renderer/components/+namespaces/namespace-select-filter.tsx
```diff
@@ -74,26 +76,42 @@ export class NamespaceSelectFilter extends React.Component<SelectProps> {
   }
 
   onChange([{ value: namespace }]: SelectOption[]) {
-    if (namespace) {
+    if (NamespaceSelectFilter.isMulti && namespace) {
       namespaceStore.toggleContext(namespace);
+    } else if (!NamespaceSelectFilter.isMulti && namespace) {
+      namespaceStore.toggleSingle(namespace);
     } else {
       namespaceStore.toggleAll(false); // "All namespaces" clicked
```
Nokel81: I don't think we should have "All Namespaces" anymore.

src/renderer/components/+namespaces/namespace-select-filter.tsx
```diff
@@ -74,26 +76,42 @@ export class NamespaceSelectFilter extends React.Component<SelectProps> {
   }
 
   onChange([{ value: namespace }]: SelectOption[]) {
-    if (namespace) {
+    if (NamespaceSelectFilter.isMulti && namespace) {
       namespaceStore.toggleContext(namespace);
+    } else if (!NamespaceSelectFilter.isMulti && namespace) {
+      namespaceStore.toggleSingle(namespace);
```
Nokel81: See below, but we won't need to check `&& namespace` anymore.

Nokel81 (commented):
This is good

src/renderer/components/+namespaces/namespace-select-filter.tsx
Nokel81: @jakolehm Is this fine?

jim-docker: I think we still need some sort of "Select All/Clear All" option. Equivalent to Ctrl-A. To select just my 200 test namespaces I can "Select All" then shift-click to deselect my 6 normal namespaces. 

Or maybe we need to be able to select ranges with modified clicks? Windows and Ubuntu commonly use shift-click to select the range between last clicked and current clicked. Ctrl-click adds individual items to the multi-selection. And Ctrl-Shift-click adds a range to the multi-selection.

I also think when all namespaces are selected they should all have check marks (instead of none having check marks, as it is now). Or a check mark beside "All namespaces"

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -158,6 +158,12 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
     }
   }
 
+  @action
+  toggleSingle(namespace: string){
+    this.clearSelected();
+    this.selectNamespaces(namespace);
```
jim-docker, line 164: This makes it so the user cannot deselect the last remaining selected namespace, which I guess is good? No point letting them pick no namespace?

Nokel81: Well currently "all namespaces" and "no namespaces" are equivalent. So shift-clicking the only selected namespace would act like selecting all.

src/renderer/components/+namespaces/namespace-select-filter.tsx
Nokel81: Well we could do that with shift-clicking the last remaining.

And then if you want to remove them you could shift-click the ones you want to remove. 

That way "shift-clicking" means toggle if this namespace is in the output set.

jakolehm: We definitely need a way to select/deselect "All Namespaces".

src/renderer/components/+namespaces/namespace-select-filter.tsx
```diff
@@ -74,26 +76,42 @@ export class NamespaceSelectFilter extends React.Component<SelectProps> {
   }
 
   onChange([{ value: namespace }]: SelectOption[]) {
-    if (namespace) {
+    if (NamespaceSelectFilter.isMulti && namespace) {
       namespaceStore.toggleContext(namespace);
+    } else if (!NamespaceSelectFilter.isMulti && namespace) {
```
jakolehm: I think we should automatically close the namespace select if user selects just one namespace.

src/renderer/components/+namespaces/namespace-select-filter.tsx
vshakirova: Thank you guys for your comments. I will  leave "All namespaces" as it is and will add closing the namespace select if user selects just one namespace, do I understand you correctly?

src/renderer/components/+namespaces/namespace-select-filter.tsx
Nokel81: Agreed.

src/renderer/components/+namespaces/namespace-select-filter.tsx
Nokel81: @jakolehm Should `"All Namespaces"` be available if `global watch` is `false`?

jakolehm: > @jakolehm Should `"All Namespaces"` be available if `global watch` is `false`?

Maybe not but hiding "All Namespaces" should be done in a separate PR.

aleksfront:
What I suggest to do in current PR:

1. Leave `All namespaces` as is.
2. Check (select/tick) all namespaces in list when clicked `All namespaces`. This will allows us to uncheck some of the namespaces from the full list.
all namespaces clicked
3. Duplicate `Shift` behavior to `Cmd` (on mac) and `Ctrl` on windows which is more intuitive on win.

@vshakirova @jim-docker wdyt?

jim-docker:
> What I suggest to do in current PR:
> 
> 1. Leave `All namespaces` as is.
> 2. Check (select/tick) all namespaces in list when clicked `All namespaces`. This will allows us to uncheck some of the namespaces from the full list.
>    all namespaces clicked
> 3. Duplicate `Shift` behavior to `Cmd` (on mac) and `Ctrl` on windows which is more intuitive on win.
> 
> @vshakirova @jim-docker wdyt?

Sounds good to me. If we feel there's a need for range selecting we could add it via shift-click (and Ctrl/Cmd-shift-click) in a follow-up PR

src/renderer/components/+namespaces/namespace-select-filter.tsx
```diff
@@ -72,26 +77,55 @@ export class NamespaceSelectFilter extends React.Component<SelectProps> {
...
+  };
+
+  onKeyUp = (e: React.KeyboardEvent<any>) => {
+    if (isMac && e.key === "Meta" || isWindows && e.key === "Ctrl" || isLinux && e.key === "Shift") {
+      NamespaceSelectFilter.isMultiSelection.set(false);
+    }
+  };
+
+  onClick = () => {
+    if (!NamespaceSelectFilter.isMultiSelection.get()) {
```
jim-docker, line 102: I guess there isn't a way to learn what the keyboard state was at the time of the click? (Instead of watching keyDown and keyUp)

src/renderer/components/+namespaces/namespace-select-filter.tsx
```diff
@@ -72,26 +77,55 @@ export class NamespaceSelectFilter extends React.Component<SelectProps> {
...
+    } else if (!NamespaceSelectFilter.isMultiSelection.get() && namespace) {
+      namespaceStore.toggleSingle(namespace);
     } else {
-      namespaceStore.toggleAll(false); // "All namespaces" clicked
+      namespaceStore.toggleAll(true); // "All namespaces" clicked
     }
   }
 
+  onKeyDown = (e: React.KeyboardEvent<any>) => {
+    if (isMac && e.metaKey || isWindows && e.ctrlKey || isLinux && e.shiftKey) {
```
jim-docker: I think Linux also uses ctrl (not shift) for multi-selection(?)

jim-docker (commented):
Works nicely, the UX is good.
I think this issue probably pre-existed, just easier to reproduce now:

https://user-images.githubusercontent.com/40840436/121565924-f717e280-c9ea-11eb-9506-70d4fc88fb4d.mov

src/renderer/components/+namespaces/namespace-select-filter.tsx
vshakirova: You are right, thank you! I'll fix it

src/renderer/components/+namespaces/namespace-select-filter.tsx
vshakirova: Not sure.. @aleksfront @Nokel81 Wdyt?

Also, unfortunately, I could not reproduce the issue. Maybe because I have only 20 namespaces.

aleksfront: Yes, we can fetch keyboard state from `onClick` event argument. There's `metaKey`, `ctrlKey` and `shiftKey` props as well.

onclick event

aleksfront: @vshakirova Proper type for event would be `React.MouseEvent`:
```
onClick = (event: React.MouseEvent) => {
```

aleksfront: After further investigating, it seems that watching `onKeyDown` & `onKeyUp` is the best option for now. `ReactSelect` doesn't provide `onClick` event to handle this so we can't actually fetch both meta keys and selected element at once.

aleksfront: @jim-docker Moved `Failed to fetch` error to separate issue https://github.com/lensapp/lens/issues/3350

aleksfront (approved):
LGTM

PR #2923, opened 2021-05-31
Add context menu entry for deleting local clusters

### LINKED ISSUE #3054: delete clusters from kubeconfig
Change all delete clusters to run kubectl config delete context --kubeconfig <path> (#2923)

Nokel81:
Part of #2692 


src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -101,38 +100,32 @@ export class KubernetesCluster extends CatalogEntity<CatalogEntityMetadata, Kube
...
+    context.menuItems.push(
       {
         title: "Settings",
         onlyVisibleForSource: "local",
-        onClick: async () => context.navigate(`/entity/${this.metadata.uid}/settings`)
+        onClick: () => context.navigate(`/entity/${this.metadata.uid}/settings`)
       },
-    ];
-
-    if (this.metadata.labels["file"]?.startsWith(ClusterStore.storedKubeConfigFolder)) {
```
jakolehm, line 113: I think we need different confirm message for internally stored kubeconfigs vs user managed.

src/common/cluster-ipc.ts
```diff
@@ -56,17 +59,42 @@ if (ipcMain) {
...
+    appEventBus.emit({ name: "cluster", action: "remove" });
+    const cluster = ClusterStore.getInstance().getById(clusterId);
+
+    if (!cluster) {
+      return;
+    }
+
+    cluster.disconnect();
+    clusterFrameMap.delete(cluster.id);
+    const kubectl = await cluster.ensureKubectl();
```
jakolehm: We should use bundled kubectl for these operations.

src/main/cluster.ts
```diff
@@ -88,12 +88,7 @@ export interface ClusterState {
 export class Cluster implements ClusterModel, ClusterState {
   /** Unique id for a cluster */
   public readonly id: ClusterId;
-  /**
-   * Kubectl
-   *
-   * @internal
-   */
-  public kubeCtl: Kubectl;
+  #kubeCtl: Kubectl;
```
jakolehm: We haven't really used `#`, maybe change to `private`?

src/main/cluster.ts
```diff
@@ -373,10 +368,12 @@ export class Cluster implements ClusterModel, ClusterState {
   /**
    * @internal
    */
-  protected async ensureKubectl() {
-    this.kubeCtl = new Kubectl(this.version);
+  async ensureKubectl() {
+    this.#kubeCtl ??= new Kubectl(this.version);
```
jakolehm: Was there a bug before? 🤔

src/main/kubectl.ts
```diff
@@ -89,16 +89,15 @@ export class Kubectl {
   public static readonly bundledKubectlVersion: string = bundledVersion;
   public static invalidBundle = false;
   private static bundledInstance: Kubectl;
+  private static readonly versionRegex = /^v?(\d+\.\d+)(.*)/;
```
jakolehm: Just noticed this regex.. just wondering why aren't we using `semver` package for parsing? 🤔 @nevalla do you remember?

src/common/cluster-ipc.ts
Nokel81: Always? okay.

src/main/cluster.ts
Nokel81: sure

src/main/cluster.ts
Nokel81: I think this is more of an optimization since `Kubectl` merely represents an on-disk binary. Though I did run into several crashes where `cluster.kubeCtl` was `undefined` while doing this.

src/main/kubectl.ts
Nokel81: I moved this from where we were doing it before since I wanted to reuse it (ended up not doing so). I would say that this probably predates `semver` being used. I can change it if you want.

src/common/catalog-entities/kubernetes-cluster.ts
Nokel81: How so? I don't delete the internally stored kubeconfigs. Just remove the `context`.

Nokel81:
Also @jakolehm FYI, this requires `simpler-add-clusters` too.

src/common/catalog-entities/kubernetes-cluster.ts
jakolehm: I was thinking pretty much the opposite. When we modify user managed configuration we need to be super clear what we are doing. When we are modifying internal kubeconfig we can even remove whole file when a kubeconfig has no more contexts.

src/main/kubectl.ts
jakolehm: Change should probably be a separate PR (I was just wondering why we had this regex).

jakolehm:
> Also @jakolehm FYI, this requires `simpler-add-clusters` too.

What stuff this requires from `simpler-add-clusters`?

src/common/catalog-entities/kubernetes-cluster.ts
Nokel81: I still don't understand why there needs to be a wording difference then. Unless you don't think that the current warning is good enough for the user-managed files.

I can add the logic for deleting the c&p files too.

jakolehm: Ok, let's keep it as is. We can iterate this in follow-up PR (if needed).

jakolehm (dismissed):

src/main/kubectl.ts
Nokel81: Okay I can factor that out.

stefcameron:
This is sorely needed! 😄

jakolehm (dismissed):

jim-docker:
Getting a JS error, after deleting a cluster from the cluster's details page (the details page should be closed once it is deleted. Instead I can pick other actions, like connect, which leads to a runtime error:

https://user-images.githubusercontent.com/40840436/121908699-3ef47d80-ccfb-11eb-9d04-e0561f6b7024.mov

Nokel81:
Thanks for the find.

Nokel81:
@jim-docker Fixed, this is what it looks like now:


https://user-images.githubusercontent.com/8225332/121912198-61d46100-ccfe-11eb-92e1-98009f87c2bb.mov

src/migrations/cluster-store/3.6.0-beta.1.ts
```diff
@@ -23,69 +23,68 @@
 // convert file path cluster icons to their base64 encoded versions
 
 import path from "path";
-import { app, remote } from "electron";
+import { app } from "electron";
```
jim-docker, line 26: are changes in this file related to deleting a cluster from the kubeconfig?

stefcameron:
> @jim-docker Fixed, this is what it looks like now:
> 
>  Screen.Recording.2021-06-14.at.10.49.19.AM.mov

Did you notice that rather long delay between deleting and when the item actually disappears from the list? I see that myself when my delete action is clicked to remove a CC cluster. I think that's related to the time required for mobx to react to the element being removed from the observed array. But it's __very__ long in terms of user perception.

Lens should show some type of "Deleting..." notification until the item actually goes away. I would in the CC extension, but I don't control that action from start to finish. All I can do is immediately remove the item from my Catalog source array. Everything else happens in Lens.

Nokel81:
> Did you notice that rather long delay between deleting and when the item actually disappears from the list?

I have, that is because everything is happening asynchronously. And there isn't ever an actual DELETE command issued to anything within Lens. This is what happens:

1. Read file from FS
2. Remove context from parsed YML of said file
3. Write file back to FS
4. FS watch detects the file was changed
5. Reads the file
6. computes diff from previous read version and removes the cluster from its internal source
7. Mobx detects that a computed value should be updated
8. We do an IPC send from `main` to `renderer`
9. That display updates.

I guess in theory I could have a notification, or change the entity's status to "Deleting".

src/migrations/cluster-store/3.6.0-beta.1.ts
Nokel81: Because I removed `kubeConfig: string;` from the `ClusterModel` I made `Pre360ClusterModel`.

The rest of the changes are releated to the fact that I removed `saveToAppFiles` as this is the only place that the `sync`-ness is needed and other places should use the promise based versions.

stefcameron:
> I guess in theory I could have a notification, or change the entity's status to "Deleting".

That sounds like a good idea to me given all the steps that need to take place, and since Lens itself doesn't know about the deletion per se. I'll implement this on my CC extension side as well.

Nokel81:
@jim-docker I have added the `deleting` phase. Though this video shows it before I fixed the css.

https://user-images.githubusercontent.com/8225332/121944195-627def00-cd20-11eb-9736-8e96c51bc958.mov

jim-docker:
> @jim-docker I have added the `deleting` phase. Though this video shows it before I fixed the css.
> 

I tried it out. I think you'll have to disable the corresponding item in the Catalog's ItemListLayout, I was able to click on it while it showed as "deleting" and bring up the details page, and get the same crash as before.

Nokel81:
Done, I have marked it as disabled. So it is even greyed out.

https://user-images.githubusercontent.com/8225332/121953908-218bd780-cd2c-11eb-93a7-c1e49eaa4181.mov

src/common/hotbar-store.ts
```diff
@@ -192,6 +192,39 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
...
+      if (index >= 0) {
+        undoItems.push([hotbar, index, hotbar.items[index]]);
+        hotbar.items[index] = null;
+      }
+    }
+
+    return () => runInAction(() => {
+      for (const [hotbar, index] of undoItems) {
+        if (hotbar.items[index] === null) {
+          console.warn("Stoped trying to undo a 'removeAllHotbarItems' call. One referent has been changed", { hotbarId: hotbar.id, index });
```
jakolehm:
```suggestion
          console.warn("Stopped trying to undo a 'removeAllHotbarItems' call. One referent has been changed", { hotbarId: hotbar.id, index });
```

src/renderer/components/+catalog/catalog-entity.store.ts
```diff
@@ -52,6 +52,10 @@ export class CatalogEntityItem implements ItemObject {
     return this.entity.status.phase;
   }
 
+  @computed get isDisabled() {
+    return this.entity.status.phase === "deleting";
```
jakolehm: Just wondering if this should be more generic (for example `this.entity.status.enabled`).

Nokel81: Probably a good idea. Do you want it to be a manditory field, or should it be `enabled?: boolean; // default true`?

src/common/hotbar-store.ts
```diff
@@ -192,6 +192,39 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
...
+
+      if (index >= 0) {
+        undoItems.push([hotbar, index, hotbar.items[index]]);
+        hotbar.items[index] = null;
+      }
+    }
+
+    return () => runInAction(() => {
+      for (const [hotbar, index] of undoItems) {
+        if (hotbar.items[index] === null) {
```
jim-docker: Did you mean `!==`?
Do we even need an undo? I think you're trying to do too much. If the delete fails for some reason it might be better to just leave the hotbar alone. It's ok that the cluster is no longer in the hotbar(s)

Nokel81: okay, I'll remove this. I just wanted the hotbar items to be still there if (for some reason) we failed due to missing write permissions

jim-docker: yeah, it's a commendable idea, but on the other hand the user asked us to remove the cluster, and taking it out of the hotbar(s) could be considered at least a partial removal...

src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -110,9 +111,17 @@ export class KubernetesCluster extends CatalogEntity<CatalogEntityMetadata, Kube
...
+          onClick: async () => {
+            const undo = HotbarStore.getInstance().removeAllHotbarItems(this.getId());
+            
+            try {
+              context.hideDetails();
+  
+              await requestMain(clusterDeleteHandler, this.metadata.uid);
+            } catch (err) {
+              undo();
+              throw err;
```
jim-docker: maybe just log instead of the undo and throw?

Nokel81: Sure

src/common/hotbar-store.ts
Nokel81: Fair, I was treating it more of a DB transaction where if something fails we try and back out of all the changes. This is probably too complex though.

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -227,7 +227,7 @@ export class Catalog extends React.Component<Props> {
           { title: "Status", className: css.statusCell, sortBy: sortBy.status },
         ]}
         customizeTableRowProps={(item: CatalogEntityItem) => ({
-          disabled: item.isDisabled,
+          disabled: !item.enabled,
```
jim-docker, line 205: Something is not working (the item is not grayed out/disabled while it is in the deleting phase any more). I don't see what's wrong though

Nokel81: Fixed, the problem was I wasn't doing the two writes in an action.

jim-docker (approved):
There's one odd behaviour still, probably minor. If you delete a cluster from the details page, it is deleted fine, details page is grayed out, then closed automatically. If you then externally add back the cluster to the kubeconfig, it gets picked up fine in the catalog, but then the details page for the re-added cluster is opened.

Since I started reviewing this there has been a fair bit of refactoring that could impact other parts of the code. We should keep an eye out for side effects...

Nokel81:
> it gets picked up fine in the catalog, but then the details page for the re-added cluster is opened. 

That is actually by design since we now keep track of the entity ID instead of the instance and grab the instance on demand.

src/common/cluster-store.ts
```diff
@@ -113,7 +109,7 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
   private static StateChannel = "cluster:state";
 
   static get storedKubeConfigFolder(): string {
-    return path.resolve((app || remote.app).getPath("userData"), "kubeconfigs");
+    return path.resolve(app.getPath("userData"), "kubeconfigs");
```
jakolehm, line 112: Why did you remove support for `renderer`? Now "add cluster" is broken...

PR #2928, opened 2021-06-01
Fix ClusterView bugs

### LINKED ISSUE #2911: Cluster view becomes empty
After switching between "active" and "disabled" clusters, status view for disabled one disappears.

https://user-images.githubusercontent.com/9607060/120154394-81e12c00-c1f8-11eb-8fbb-9953607121e8.mov

### LINKED ISSUE #2913: Fail to open disconnected cluster view
After some moment, it's unable to open disabled/disconnected cluster from hotbar or catalog items. Clicking it doesn't change view and basically does nothing.

https://user-images.githubusercontent.com/9607060/120164796-9bd43c00-c203-11eb-9ca0-6ba7d161d8c5.mov

Probably relates to #2911

jakolehm:
Fixes #2911
Fixes #2913

### DIFF
--- a/src/renderer/components/cluster-manager/cluster-view.tsx
+++ b/src/renderer/components/cluster-manager/cluster-view.tsx
@@ -30,20 +30,26 @@ import { ClusterStore } from "../../../common/cluster-store";
 import { requestMain } from "../../../common/ipc";
 import { clusterActivateHandler } from "../../../common/cluster-ipc";
 import { catalogEntityRegistry } from "../../api/catalog-entity-registry";
-import { getMatchedClusterId, navigate } from "../../navigation";
+import { navigate } from "../../navigation";
 import { catalogURL } from "../+catalog/catalog.route";
+import type { RouteComponentProps } from "react-router-dom";
+import type { IClusterViewRouteParams } from "./cluster-view.route";
+
+interface Props extends RouteComponentProps<IClusterViewRouteParams> {
+}
+
 
 @observer
-export class ClusterView extends React.Component {
+export class ClusterView extends React.Component<Props> {
   private store = ClusterStore.getInstance();
 
-  constructor(props: {}) {
+  constructor(props: Props) {
     super(props);
     makeObservable(this);
   }
 
-  get clusterId() {
-    return getMatchedClusterId();
+  @computed get clusterId() {
+    return this.props.match.params.clusterId;
   }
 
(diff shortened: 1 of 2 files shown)


nevalla (approved):
LGTM

src/renderer/components/cluster-manager/cluster-view.tsx
```diff
@@ -30,20 +30,26 @@ import { ClusterStore } from "../../../common/cluster-store";
 import { requestMain } from "../../../common/ipc";
 import { clusterActivateHandler } from "../../../common/cluster-ipc";
 import { catalogEntityRegistry } from "../../api/catalog-entity-registry";
-import { getMatchedClusterId, navigate } from "../../navigation";
+import { navigate } from "../../navigation";
 import { catalogURL } from "../+catalog/catalog.route";
+import type { RouteComponentProps } from "react-router-dom";
+import type { IClusterViewRouteParams } from "./cluster-view.route";
+
+interface Props extends RouteComponentProps<IClusterViewRouteParams> {
```
jakolehm, line 38: Without these props `navigate` did not work (it showed just blank page).

src/renderer/components/cluster-manager/cluster-view.tsx
```diff
@@ -87,7 +92,7 @@ export class ClusterView extends React.Component {
     const { clusterId, cluster, isReady } = this;
 
     if (cluster && !isReady) {
-      return <ClusterStatus clusterId={clusterId} className="box center"/>;
+      return <ClusterStatus key={clusterId} clusterId={clusterId} className="box center"/>;
```
jakolehm, line 95: Unique key so that `ClusterStatus` is not cached between clusters.

Nokel81: Are we only using a single iframe now?

jakolehm: This is not related to iframes at all. We are still having iframe per cluster. But we have just one status/error page that was shared between clusters -> in some cases it showed logs etc from previous cluster. `key` makes React to recreate the component instance on change.

Nokel81: Oh okay makes sense. I thought `ClusterStatus` was within the iframe.

Nokel81 (approved):

PR #2932, opened 2021-06-01
Show entity kind on catalog browse-all list

jakolehm:
image


nevalla (dismissed):

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -197,6 +198,85 @@ export class Catalog extends React.Component<Props> {
     );
   }
 
+  renderSingleCategoryList() {
+    return (
+      <ItemListLayout
+        renderHeaderTitle={this.catalogEntityStore.activeCategory?.metadata.name ?? "Browse All"}
+        isClusterScoped
```
Nokel81: Since this basically means "load namespace store" I don't think we need `isClusterScoped`.

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -197,6 +198,85 @@ export class Catalog extends React.Component<Props> {
...
+        renderItemMenu={this.renderItemMenu}
+      />
+    );
+  }
+
+  renderAllCategoriesList() {
+    return (
+      <ItemListLayout
+        renderHeaderTitle={this.catalogEntityStore.activeCategory?.metadata.name ?? "Browse All"}
+        isClusterScoped
```
Nokel81: Same here.

src/renderer/components/+catalog/catalog.tsx
jakolehm: Fixed.

Nokel81 (approved):

nevalla (approved):

Nokel81:
From my personal testing, I think CI is failing due to the shell sync issue.

Nokel81:
It has been happening to my PRs too.

src/renderer/components/+catalog/catalog.tsx
jakolehm: Except that we need this... otherwise "namespace selector" is visible 😞

Nokel81: That will be fixed by #2743 since it checks the stores (I think)

Nokel81: namely:

https://github.com/lensapp/lens/pull/2743/files#diff-e3cb3086ca8de521ccfb9f32c52693b212f25682682504df3d9fcd6515ec7cd1R194-R207

PR #2934, opened 2021-06-01
Expose NamespaceSelectFilter component to Extensions API

nevalla:
This PR exposes `NamespaceSelectFilter` component to extensions API and makes extension developers lives easier if they need to allow user to select namespaces.

**Example**
```html
        <header className="flex gaps align-center">
          <h2 className="flex gaps align-center">
            <span>Resource Map</span>
            <Renderer.Component.Icon material="info" tooltip={<KubeResourceChartLegend/>}/>
          </h2>
          <div className="box right">
            <Renderer.Component.NamespaceSelectFilter />
          </div>
        </header>
 ```

image


Nokel81:
Sure we can do this, is there a reason why this cannot just be dynamically added to the `<TopBar />` instead?

Nokel81 (approved):

jakolehm (approved):

PR #2936, opened 2021-06-01
Move requesting isGlobalWatchAllowed to renderer

Nokel81:

Continuing on from https://github.com/lensapp/lens/pull/2892

### DIFF
--- a/src/common/ipc/cluster.ipc.ts
+++ b/src/common/ipc/cluster.ipc.ts
@@ -26,6 +26,7 @@
 export const ClusterListNamespaceForbiddenChannel = "cluster:list-namespace-forbidden";
 export const ClusterGetResourcesChannel = "cluster:resources";
 export const ClusterResourceIsAllowedChannel = "cluster:resource:is-allowed";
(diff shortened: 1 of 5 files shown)


src/common/cluster-ipc.ts
```diff
@@ -128,4 +128,22 @@ if (ipcMain) {
 
     return Array.from(isAllowed);
   });
+
+  handleRequest(ClusterGlobalWatchChannel, async (event, clusterId: ClusterId): Promise<boolean> => {
```
jakolehm: Not related to this PR but I think we need to consider moving `handleRequest` handlers to `main`. We could leave the channel names to `common` because they are used from both sides.

jakolehm (approved):

src/common/cluster-ipc.ts
Nokel81: Already done in https://github.com/lensapp/lens/pull/2821

nevalla (approved):

Nokel81:
Closing due to the target PR being closed.

PR #2938, opened 2021-06-01
Fix shell-env timeout

### LINKED ISSUE #2764: Stuck in Syncing shell environment
**Describe the bug**
No window appears

**To Reproduce**
for view where the app stuck i have used the following command `gnome-terminal -- /bin/bash -c '/snap/bin/kontena-lens'`

**Screenshots**
```
info: 📟 Setting Lens as protocol client for lens://
info: 📟 failed ❗
info: 🚀 Starting Lens from "/home/lop/snap/kontena-lens/179/.config/Lens"
info: 🐚 Syncing shell environment
```

- Lens Version: 4.2.2
- OS: Ubuntu 20.04
- Installation method: snap

**Additional context**
if you run CTRL+c the following error appears:
```
^Cerror: shellEnv: TypeError: Cannot read property 'split' of undefined
```
the application proceed and the window appears

jim-docker:
fixes #2764

### DIFF
--- a/src/main/utils/shell-env.ts
+++ b/src/main/utils/shell-env.ts
@@ -19,42 +19,33 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { shellEnv } from "./utils/shell-env";
-import os from "os";
-import { app } from "electron";
+import shellEnvironment from "shell-env";
+import logger from "../logger";
 
-interface Env {
-  [key: string]: string;
+export interface EnvironmentVariables {
+  readonly [key: string]: string;
 }
 
-/**
- * shellSync loads what would have been the environment if this application was
- * run from the command line, into the process.env object. This is especially
- * useful on macos where this always needs to be done.
- */
-export async function shellSync() {
-  const { shell } = os.userInfo();
+let shellSyncFailed = false;
+
+export async function shellEnv(shell?: string, timeout?: number) : Promise<EnvironmentVariables> {
   let envVars = {};
 
-  envVars = await shellEnv(shell, 5_000);
-
-  const env: Env = JSON.parse(JSON.stringify(envVars));
-
-  if (!env.LANG) {
-    // the LANG env var expects an underscore instead of electron's dash
-    env.LANG = `${app.getLocale().replace("-", "_")}.UTF-8`;
-  } else if (!env.LANG.endsWith(".UTF-8")) {
-    env.LANG += ".UTF-8";
+  if (!shellSyncFailed) {
+    try {
+      envVars = await Promise.race([
+        shellEnvironment(shell),
+        new Promise((_resolve, reject) => setTimeout(() => {
+          reject(new Error("Resolving shell environment is taking very long. Please review your shell configuration."));
+        }, timeout))
+      ]);
+    } catch (error) {
+      logger.error(`shellEnv: ${error}`);
+      shellSyncFailed = true;
+    }
+  } else {
+    logger.error("shellSync(): Resolving shell environment took too long. Please review your shell configuration.");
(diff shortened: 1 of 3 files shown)


src/main/utils/shell-env.ts
```diff
@@ -0,0 +1,67 @@
+import execa from "execa";
```
jim-docker: This file is a copy of https://www.npmjs.com/package/shell-env, converted to typescript and added a timeout parameter

jakolehm:
@jim-docker what if we just mark "shell env" as "invalid" during boot and skip shell-env within terminal if we see an invalid shell?

src/main/utils/shell-env.ts
```diff
@@ -0,0 +1,51 @@
...
+          reject(new Error("Resolving shell environment is taking very long. Please review your shell configuration."));
+        }, timeout))
+      ]);
+    } catch (error) {
+      logger.error(`shellEnv: ${error}`);
+      shellSyncFailed = true;
+    }
+  } else {
+    logger.error("shellSync(): Resolving shell environment took too long. Please review your shell configuration.");
+  }
```
Nokel81, line 63: @jakolehm is this line skipping on subsequent calls?

src/main/utils/shell-env.ts
```diff
@@ -31,14 +31,20 @@ let shellSyncFailed = false;
...
-      ]);
+      if (timeout > 0 ) {
+        envVars = await Promise.race([
+          shellEnvironment(shell),
+          new Promise((_resolve, reject) => setTimeout(() => {
+            reject(new Error("Resolving shell environment is taking very long. Please review your shell configuration."));
+          }, timeout))
+        ]);
+      } else {
+        envVars = await shellEnvironment(shell);
```
Nokel81: Why would we want to not timeout?

jim-docker: To provide generic behaviour? (it is how we were using it before--the `shell-env` package does not support a timeout). Otherwise we need to force a minimum timeout. Or we could make `shellEnv()` less generic, just hard-code it with a 5s timeout.

Nokel81: Why do we need to force a minimum timeout? The above code handles the 0s timeout excellently (it will always fail).

jim-docker: How is that useful?

Nokel81: Safer I would argue. Since you would have to explicitly pass in `Number.PositiveInfinity`

src/main/utils/shell-env.ts
jakolehm: I don't understand this question 🤔

src/main/utils/shell-env.ts
jakolehm: I think we want always timeout... but if first call does timeout we should have a way to try again.

jim-docker: I thought the user would have to restart lens to try the shell sync again. This is what I do with vs code when the shell sync time out, though maybe I missed a way to do it without restarting.

jim-docker: > Safer I would argue. Since you would have to explicitly pass in `Number.PositiveInfinity`

But it is unnecessary to create a timer in this case. The timer would never get cleared, if the `shellEnv` function blocked

jim-docker: nvm, we always want to timeout

src/main/utils/shell-env.ts
Nokel81: It was in response to this comment made by you:

> @jim-docker what if we just mark "shell env" as "invalid" during boot and skip shell-env within terminal if we see an invalid shell?

Nokel81 (approved):

PR #2940, opened 2021-06-02
Upgrade `typedoc`

Nokel81:

closes https://github.com/lensapp/lens/pull/2518

~~blocked by https://github.com/TypeStrong/typedoc/issues/1589~~

This PR currently uses `typedoc@0.21.0-beta.2` which is a beta version, but the only item that is blocking the release is https://github.com/TypeStrong/typedoc/pull/1567. But that doesn't apply to us. It should be easy to upgrade to `0.21.0` itself once that is released.


typedoc.json
```diff
@@ -0,0 +1,11 @@
+{
```
jakolehm, line 1: Should this also have "entrypoints"? (like in #2518)

Nokel81: So I was basing this on the arguments directly passed to `yarn run typedoc` in `package.json`.

And that only is `--includes src/` which would refer to https://typedoc.org/guides/options/#includes in the documentation. Which we don't use.

`--entryPoints` seems (from my reading of the docs) to be the way to specify multiple files (or folders) as the starting points for the docs.

But we only want to document the extension API, iirc, and not the whole of `src`. So I don't think we do.

jakolehm: Did you test how the documentation looks like? In my PR I had to use entrypoints for proper generation (https://typedoc.org/guides/options/#entrypoints)

Nokel81: Yeah I did, it looked fine. I'll post some screenshots for you to look over.

Nokel81:
https://user-images.githubusercontent.com/8225332/120702364-fd094300-c481-11eb-9e5f-1d25e88ebe2c.mov

jakolehm (approved):

PR #2945, opened 2021-06-03
Implement extensions API for workloads overview details

nevalla:
This PR will add `workloadsOverviewDetailItems` property to `LensRendererExtension` and it allows extension developers to register items to Workload Overview page.

**Usage Example**

```typescript
export default class KubeResorceMapRenderer extends Renderer.LensExtension {
  kubeWorkloadsOverviewItems = [
    {
      priority: 25,
      components : {
        Details: () => { return (
          <div className="ResourceMapOverviewDetail">
            <div className="header flex gaps align-center">
              <h5 className="box grow">Resources</h5>
            </div>
            <div className="content">
              <KubeForceChart height={480} />
            </div>
          </div>
        )}
      }
    }
  ]
}
```

image

### DIFF
--- a/src/extensions/lens-renderer-extension.ts
+++ b/src/extensions/lens-renderer-extension.ts
@@ -40,9 +40,9 @@ export class LensRendererExtension extends LensExtension {
   statusBarItems: StatusBarRegistration[] = [];
   kubeObjectDetailItems: KubeObjectDetailRegistration[] = [];
   kubeObjectMenuItems: KubeObjectMenuRegistration[] = [];
+  kubeWorkloadsOverviewDetailItems: WorkloadsOverviewDetailRegistration[] = [];
   commands: CommandRegistration[] = [];
   welcomeMenus: WelcomeMenuRegistration[] = [];
-  workloadsOverviewDetailItems: WorkloadsOverviewDetailRegistration[] = [];
 
   async navigate<P extends object>(pageId?: string, params?: P) {
     const { navigate } = await import("../renderer/navigation");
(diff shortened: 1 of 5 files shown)


src/extensions/lens-renderer-extension.ts
```diff
@@ -42,6 +42,7 @@ export class LensRendererExtension extends LensExtension {
   kubeObjectMenuItems: KubeObjectMenuRegistration[] = [];
   commands: CommandRegistration[] = [];
   welcomeMenus: WelcomeMenuRegistration[] = [];
+  workloadsOverviewDetailItems: WorkloadsOverviewDetailRegistration[] = [];
```
jakolehm: Maybe we should prefix with kube.. for example `kubeWorkloadOverviewItems`?

nevalla: prefixed

src/renderer/components/+workloads-overview/overview.tsx
```diff
@@ -57,11 +58,34 @@ export class WorkloadsOverview extends React.Component<Props> {
...
 }
+
+workloadsOverviewDetailRegistry.add([
+  {
+    components: {
+      Details: (props: any) => <OverviewStatuses {...props} />,
+    }
+  },
+  {
+    priority: 5,
```
Nokel81, line 82: I know that the default priority is `50`, but since the range is technically `[Number.NEGATIVE_INFINITY, Number.POSITIVE_INFINITY]` (a closed but non-complete range). Would it make sense to have a PR which changes the default value to be `0`?

Nokel81 (approved):

nevalla:
@jakolehm PTAL again

jakolehm (approved):
LGTM (we need to create a doc PR also)

PR #2946, opened 2021-06-03
Set in-tree extensions package.json version on build

### LINKED ISSUE #2944: App update does not update in-tree extensions
**Describe the bug**

When built app is updated in-tree extensions won't get updated properly. This is because now we are bundling tgz files and package.json version never changes -> npm ignores the update because version did not change.

Note: this applies only to 5.0 beta series.

**Expected behavior**

In-tree extensions update properly.

jakolehm:
Fixes #2944

### DIFF
--- a/extensions/kube-object-event-status/package.json
+++ b/extensions/kube-object-event-status/package.json
@@ -1,6 +1,6 @@
 {
   "name": "kube-object-event-status",
-  "version": "0.1.0",
+  "version": "0.0.1",
   "description": "Adds kube object status from events",
   "renderer": "dist/renderer.js",
   "lens": {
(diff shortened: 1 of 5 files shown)


extensions/kube-object-event-status/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kube-object-event-status",
-  "version": "0.1.0",
+  "version": "0.0.1",
```
jakolehm, line 3: Marked these as `0.0.1` to indicate more clearly that this version does not matter.

nevalla (approved):
LGTM

extensions/kube-object-event-status/package.json
Nokel81: Wouldn't that be `0.0.0` then? Like the NPM package?

jim-docker (approved):

Issue #458, opened 2020-06-15
Switch to use Monaco editor


nevalla:
https://microsoft.github.io/monaco-editor/

Monaco editor provides better capabilities to enable k8s resource auto-completion and validation.

aleksfront:
We should use some plugin for supporting YAML format. https://www.npmjs.com/package/monaco-yaml can be a good candidate.

LucaDev:
Additionally I'd love to see the editor in more places, like for editing configmaps.

aleksfront:
Closed in #2949

PR #2949, opened 2021-06-03
Replace Ace editor with Monaco

pashevskii:
Peek 2021-06-22 17-47
Replaces Ace Editor with Monaco Editor

Implements https://github.com/lensapp/lens/issues/458


src/renderer/components/dock/create-resource.store.ts
```diff
@@ -29,6 +29,7 @@ import { autoBind } from "../../utils";
 import { DockTabStore } from "./dock-tab.store";
 import { dockStore, IDockTab, TabKind } from "./dock.store";
 
+
```
nevalla: redundant new line?

src/renderer/components/dock/create-resource.store.ts
```diff
@@ -82,9 +83,11 @@ export class CreateResourceStore extends DockTabStore<string> {
 export const createResourceStore = new CreateResourceStore();
 
 export function createResourceTab(tabParams: Partial<IDockTab> = {}) {
-  return dockStore.createTab({
+  const tab =  dockStore.createTab({
```
nevalla: unnecessary change?

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -33,6 +33,7 @@ export type DockTabStorageState<T> = Record<TabId, T>;
 export class DockTabStore<T> {
   protected storage?: StorageHelper<DockTabStorageState<T>>;
   protected data = observable.map<TabId, T>();
+
```
nevalla: redundant new line

src/renderer/components/dock/dock.store.ts
```diff
@@ -23,6 +23,8 @@ import MD5 from "crypto-js/md5";
 import { action, computed, IReactionOptions, makeObservable, observable, reaction } from "mobx";
 import { autoBind, createStorage } from "../../utils";
 import throttle from "lodash/throttle";
+import {monacoModelsStore} from "./monaco-model.store";
+
```
nevalla, line 27: also here extra new line

src/renderer/components/dock/edit-resource.store.ts
```diff
@@ -111,6 +112,11 @@ export class EditResourceStore extends DockTabStore<EditingResource> {
       dispose();
     });
   }
+
+  setData(tabId: TabId, data: EditingResource) {
+    super.setData(tabId, data);
+
```
nevalla: redundant new line

src/renderer/components/dock/upgrade-chart.store.ts
```diff
@@ -118,11 +119,13 @@ export class UpgradeChartStore extends DockTabStore<IChartUpgradeData> {
     const values = await getReleaseValues(releaseName, releaseNamespace, true);
 
     this.values.setData(tabId, values);
+    monacoModelsStore.getModel(tabId).setValue(values);
   }
 
   getTabByRelease(releaseName: string): IDockTab {
     return dockStore.getTabById(this.releaseNameReverseLookup.get(releaseName));
   }
+
```
nevalla: redundant new line

src/renderer/utils/createStorage.ts
```diff
@@ -53,7 +53,7 @@ export function createStorage<T>(key: string, defaultValue: T) {
 
   function init() {
     storage.initialized = true;
-
+    console.log("folder", folder);
```
nevalla: please remove `console.log`

nevalla (changes requested):
This looks awesome. Added this point some comments to remove empty lines to make diff smaller.

src/renderer/components/dock/edit-resource.store.ts
pashevskii: Hmm, setData should be deleted from there at all

src/renderer/components/dock/create-resource.store.ts
pashevskii: Yes, it is) It is the result of some experiments with monaco models. Fixed

package.json
```diff
@@ -225,6 +226,7 @@
     "proper-lockfile": "^4.1.2",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
+    "react-monaco-editor": "^0.43.0",
```
Nokel81: You should remove "ace-builds" from devDeps

src/renderer/components/+apps-releases/release-details.tsx
```diff
@@ -158,15 +158,15 @@ export class ReleaseDetails extends Component<Props> {
             onChange={value => this.showOnlyUserSuppliedValues = value}
             disabled={valuesLoading}
           />
-          <AceEditor
```
Nokel81, line 161: If we are replacing `<AceEditor>` then we should remove the files for that too.

src/renderer/components/dock/create-resource.tsx
```diff
@@ -160,8 +164,8 @@ export class CreateResource extends React.Component<Props> {
         />
         <EditorPanel
           tabId={tabId}
-          value={data}
           onChange={onChange}
+          value = {data}
         />
```
Nokel81: Please revert for smaller diff.

pashevskii:
Converted to draft for removing Ace dependency at all from the lens (not just for tabs)

pashevskii:
@Nokel81 @nevalla Ace Editor has been fully removed, please look at the changes

src/renderer/components/+add-cluster/add-cluster.tsx
```diff
@@ -115,10 +116,11 @@ export class AddCluster extends React.Component {
           Read more about adding clusters <a href={`${docsUrl}/clusters/adding-clusters/`} rel="noreferrer" target="_blank">here</a>.
         </p>
         <div className="flex column">
-          <AceEditor
-            autoFocus
-            showGutter={false}
-            mode="yaml"
+          <MonacoEditor
+            options = {{ automaticLayout: true, minimap: {enabled: false}, lineNumbers: "off"}}
```
Nokel81: Oooh, that would be some nice user preferences (for a future PR).

src/renderer/components/+custom-resources/crd-details.tsx
```diff
@@ -143,11 +144,12 @@ export class CRDDetails extends React.Component<Props> {
         {validation &&
         <>
           <DrawerTitle title="Validation"/>
-          <AceEditor
-            mode="yaml"
-            className="validation"
+          <MonacoEditor
+            options = {{ automaticLayout: true, minimap: {enabled: false}, readOnly: true}}
```
Nokel81: Would it be better to have a single place where the default options object is specified? Because this one doesn't turn off line numbers but the one in `add-cluster.tsx` does.

src/renderer/components/dock/create-resource.tsx
```diff
@@ -160,7 +164,7 @@ export class CreateResource extends React.Component<Props> {
...
         <EditorPanel
           tabId={tabId}
-          value={data}
+          value = {data}
```
Nokel81: nit: please revert.

src/renderer/components/dock/create-resource.tsx
```diff
@@ -88,7 +89,10 @@ export class CreateResource extends React.Component<Props> {
 
   onSelectTemplate = (item: SelectOption) => {
     this.currentTemplates.set(this.tabId, item);
-    fs.readFile(item.value,"utf8").then(v => createResourceStore.setData(this.tabId,v));
+    fs.readFile(item.value,"utf8").then(v => {
+      createResourceStore.setData(this.tabId,v);
+      monacoModelsStore.getModel(this.tabId).setValue(v ?? "");
```
Nokel81: Do we need both of these stores? Unless they do different tasks it seems like it is making more work for ourselves by having to keep them in sync.

src/renderer/components/dock/monaco-model.store.ts
```diff
@@ -0,0 +1,90 @@
...
+    get models(){
+      return this.storage.get().models;
+    }
+
+    set models(models: ModelEntry[]) {
+      this.storage.merge({ models });
+    }
+
+    @action
+    addModel(tabId: string, value = "", lang = "yaml"): monaco.editor.ITextModel {
```
Nokel81: For three parameters, especially since they have default values. I think that this should be the following. So that new arguments can be added in the future.

```ts
function addModel(tabId: string, { value = "", lang = "yaml" } = {}): monaco.editor.ITextModel;
```

src/renderer/theme.store.ts
```diff
@@ -26,6 +26,12 @@ import logger from "../main/logger";
 
 export type ThemeId = string;
 
+export enum MonacoTheme {
+  HC_BLACK = "hc-black",
```
Nokel81: If there is no way to specify `MonacoTheme.HC_BLACK` as of yet I don't think it should be an option here.

BTW is this a fixed list or a dynamic list?

src/renderer/utils/createStorage.ts
```diff
@@ -53,7 +53,7 @@ export function createStorage<T>(key: string, defaultValue: T) {
 
   function init() {
     storage.initialized = true;
-
+
```
Nokel81: please revert, this file doesn't need to be touched otherwise.

src/renderer/components/+add-cluster/add-cluster.tsx
pashevskii: Agree, I'm working on it right now

src/renderer/theme.store.ts
pashevskii: https://github.com/react-monaco-editor/react-monaco-editor/blob/bccb619c3ec5037bdda33afc268a616879334bfe/src/types.ts#L48

It is a fixed list of built-in themes

pashevskii: How about adding the checkbox in preferences like "Use Hight contrast editor theme (applicable for Dark theme)"

pashevskii: image

HC BLACK

src/renderer/components/dock/create-resource.tsx
pashevskii: Basically, monacoModelsStore is for managing monaco models, these entities are responsible for storing virtual in-memory "profiles" for the tabs. I tried to move all logic in one object, but, as I see, it requires more global refactoring. 

I can remove storing functions from monacoModelsStore (and rename this to something like monacoModelManager) and work with this within runtime, w/o (de)serialization of urls and languages.

src/renderer/components/dock/monaco-model.store.ts
pashevskii: It makes sense, thanks

src/renderer/components/dock/create-resource.tsx
Nokel81: If it is just for in memory remembering anyway. I think that just a model manager would be good.

src/renderer/theme.store.ts
Nokel81: Good to know. I think user preferences should be left to a separate PR

src/renderer/components/dock/create-resource.tsx
pashevskii: fixed

pashevskii:
@Nokel81 @nevalla  @aleksfront  I've added the ability to config editors by the second commit. Could you please take a look?

image

Nokel81:
@pashevskii Can you please sign your commits.

pashevskii:
> @pashevskii Can you please sign your commits.

Sorry, it is done

Nokel81:
> @Nokel81 @nevalla @aleksfront I've added the ability to config editors by the second commit. Could you please take a look?
> 
> image

Did you change styling? That doesn't look in line with how other settings look.

src/common/user-store.ts
```diff
@@ -97,6 +113,12 @@ export class UserStore extends BaseStore<UserStoreModel> {
    */
   hiddenTableColumns = observable.map<string, ObservableToggleSet<string>>();
 
+  editorConfiguration = observable.map<EditorType, EditorConfiguration>([
+    [EditorType.TABS, {miniMap: {enabled: true}, lineNumbers: "on", tabSize: 4}],
+    [EditorType.DETAILS, {miniMap: {enabled: false}, lineNumbers: "on", tabSize: 2}],
+    [EditorType.ADD_CLUSTER, {miniMap: {enabled: false}, lineNumbers: "on", tabSize: 2}],
+    [EditorType.KUBECONFIG, {miniMap: {enabled: false}, lineNumbers: "off", tabSize: 2}]]);
```
Nokel81: Please rebase/merge master as this sort of thing has now changed.

Furthermore I think it would be better if we just had settings for all kinds of editors to begin with, this fine detail seems a bit much.

pashevskii:
> > @Nokel81 @nevalla @aleksfront I've added the ability to config editors by the second commit. Could you please take a look?
> > image
> 
> Did you change styling? That doesn't look in line with how other settings look.

the styling was not changed, in the literal sense of the word, but:
1. I've added tabs for switching between editors (Config for tabs editor, releases/crd details, etc) because every tab contains similar inputs, IMHO listing all these configs makes the page huge
2.  also labels for selects was placed on the one line with the dropdown selects just to minimizing UI

I did some trying to bring it to "lens style" but IMHO it makes this page more ugly than now.

Nokel81:
@pashevskii How would a table look instead? Though to be honest do we really need 4 different settings?

src/common/user-store/preferences-helpers.ts
```diff
@@ -222,6 +243,17 @@ const syncKubeconfigEntries: PreferenceDescription<KubeconfigSyncEntry[], Map<st
...
 };
 
+const editorConfiguration: PreferenceDescription<[EditorType, EditorConfiguration][], Map<EditorType, EditorConfiguration>> = {
+  fromStore(val) {
+    return new Map(
+      (val??defaultEditorConfig).map(([type, config]) => [type, config])
+    );
+  },
+  toStore(val) {
+    return Array.from(val, ([type, config]) => ([ type, config ]));
```
Nokel81: toStore should remove items that are equal to the default config. So that we can change the defaults in the future.

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -111,6 +120,51 @@ export class Preferences extends React.Component {
     );
   }
 
+  renderEditorConfiguration(editorType: EditorType) {
+    if (this.activeEditorType == editorType)  
+    {return (<>
```
Nokel81: Please do it in the other order:

```ts
    if (this.activeEditorType !== editorType) {
      return null;
    }

    return (
      <>
        ...
      </>
    );
```

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -111,6 +120,51 @@ export class Preferences extends React.Component {
...
+          label="Line numbers"
+          control={ <Select
+            options={[{label: "On", value: "on"}, 
+              {label: "Off", value: "off"}, 
+              {label: "Relative", value: "relative"}, 
+              {label: "Interval", value: "interval"}]}
+            value={UserStore.getInstance().editorConfiguration.get(editorType).lineNumbers}
+            onChange={({ value }: SelectOption) => UserStore.getInstance().editorConfiguration.get(editorType).lineNumbers = value}
+            themeName="lens"
+          /> }
```
Nokel81: Please have a consistent indenting
```suggestion
          control={
            <Select
              options={[
                {label: "On", value: "on"}, 
                {label: "Off", value: "off"}, 
                {label: "Relative", value: "relative"}, 
                {label: "Interval", value: "interval"},
              ]}
              value={UserStore.getInstance().editorConfiguration.get(editorType).lineNumbers}
              onChange={({ value }: SelectOption) => UserStore.getInstance().editorConfiguration.get(editorType).lineNumbers = value}
              themeName="lens"
            /> 
          }
```

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -111,6 +120,51 @@ export class Preferences extends React.Component {
...
+        />
+      </section>
+      <section>
+        <FormSwitch
+          label="Line numbers"
+          control={ <Select
+            options={[{label: "On", value: "on"}, 
+              {label: "Off", value: "off"}, 
+              {label: "Relative", value: "relative"}, 
+              {label: "Interval", value: "interval"}]}
```
Nokel81: These options should probably be defined as a const object somewhere else and use an `enum`.

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -111,6 +120,51 @@ export class Preferences extends React.Component {
...
+            value={UserStore.getInstance().editorConfiguration.get(editorType).lineNumbers}
+            onChange={({ value }: SelectOption) => UserStore.getInstance().editorConfiguration.get(editorType).lineNumbers = value}
```
Nokel81: I think that these two should be helepr methods on `UserStore`:

```ts
    showLineNumbers(editorType: EditorType): LineNumbers;
```

```ts
   setLineNumbers(editorType: EditorType, setting: LineNumbers): void;
```

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -111,6 +120,51 @@ export class Preferences extends React.Component {
...
+      <section>
+        <FormSwitch
+          label="Tab size"
+          control={ <Select
+            options={[{label: "2 spaces", value: 2}, 
+              {label: "4 spaces", value: 4}]}
+            value={UserStore.getInstance().editorConfiguration.get(editorType).tabSize}
+            onChange={({ value }: SelectOption) => UserStore.getInstance().editorConfiguration.get(editorType).tabSize = value}
+            themeName="lens"
+          />}
```
Nokel81: This should be a direct input, users should be able to pick 3 if they want to.

src/renderer/components/+add-cluster/add-cluster.tsx
```diff
@@ -115,10 +118,11 @@ export class AddCluster extends React.Component {
...
         </p>
         <div className="flex column">
-          <AceEditor
-            autoFocus
-            showGutter={false}
-            mode="yaml"
+          <MonacoEditor
+            options = {{ ...UserStore.getInstance().getEditorOptions(EditorType.ADD_CLUSTER)}}
+            className={cssNames( "MonacoEditor")}
+            theme={ThemeStore.getInstance().activeTheme.monacoTheme}
```
Nokel81, line 122: This really is a kubeconfig, it should have the same config as that.

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -196,6 +250,21 @@ export class Preferences extends React.Component {
...
             </section>
+            <hr/>
+            <section>
+              <SubTitle title="Editor configuration" />
+              
+              <Tabs className="flex" scrollable={false} onChange={this.onEditorTypeChange} value={this.activeEditorType}>
+                <Tab value={EditorType.TABS} label="Tabs" data-testid="editor-tabs"/>
+                <Tab value={EditorType.DETAILS} label="Details" data-testid="editor-details"/>
+                <Tab value={EditorType.ADD_CLUSTER} label="Adding cluster" data-testid="editor-add-cluster"/>
+                <Tab value={EditorType.KUBECONFIG} label="Kubeconfig" data-testid="editor-kubeconfig"/>
```
Nokel81: Tabs is a little confusing since it is actual Dock entries and not tab components.

I still think that ADD_CLUSTER and KUBECONFIG should be one setting (since they are both kubeconfigs).

And in all actual fact I think we should have settings for "narrow views" and "wide views" instead. With "narrow views" being the details page and the wizard, and "wide views" being the dock and the "add-clusters". @jakolehm WDYT?

Nokel81 (dismissed):

src/common/user-store.ts
pashevskii: I'm not sure that is necessary. For instance, all tabs have the same position, size, and similar data, I believe that nobody will be in need to have different configs for creating and editing pods(deployments/etc).

pashevskii: Not sure, first - default settings (ace editor) are different, some places have no line-numbers
second - some places are too small to show minimap for navigation (imho)
therefore I would like to have the ability to configure each mount point of the editor. Probably we should leave only Tabs configs, wdyt?

Nokel81: How about a narrow/wide approach like https://github.com/lensapp/lens/pull/2949#discussion_r668750681?

src/renderer/components/+preferences/preferences.tsx
pashevskii: > I still think that ADD_CLUSTER and KUBECONFIG should be one setting (since they are both kubeconfigs).

Fixed

src/common/user-store/preferences-helpers.ts
Nokel81: This should still be changed.

src/common/user-store/user-store.ts
```diff
@@ -109,7 +116,41 @@ export class UserStore extends BaseStore<UserStoreModel> /* implements UserStore
...
+    UserStore.getInstance().editorConfiguration.get(type).tabSize = tabSize;
+  }
+
+  isMinimapEnabled(type: EditorType): boolean {
+    return UserStore.getInstance().editorConfiguration.get(type).miniMap.enabled;
+  }
+
+  enableMinimap(type: EditorType, miniMap: boolean) {
+    UserStore.getInstance().editorConfiguration.get(type).miniMap.enabled = miniMap;
   }
```
Nokel81: I think these can probably be reduced to a `getEditorConfiguration` function.

Furthermore, there is no need to call `UserStore.getInstance()` within `UserStore` methods. Just use `this`.

src/common/user-store/preferences-helpers.ts
pashevskii: Done

pashevskii:
image
The preferences page was refactored, now editor properties have default configs like before this PR but the user can override the default configs with global value. Also, Editor configuration was moved into a new tab. Thanks @aleksfront for the help with this!

src/common/user-store/user-store.ts
```diff
@@ -81,6 +83,11 @@ export class UserStore extends BaseStore<UserStoreModel> /* implements UserStore
    */
   hiddenTableColumns = observable.map<string, ObservableToggleSet<string>>();
 
+  /**
+   * Monaco editor configs
+   */
+   @observable editorConfiguration:EditorConfiguration = {tabSize: null, miniMap: null, lineNumbers: null};
```
Nokel81, line 88: If there are three different `EditorType`'s why so we only have one here?

src/common/user-store/preferences-helpers.ts
```diff
@@ -222,6 +241,16 @@ const syncKubeconfigEntries: PreferenceDescription<KubeconfigSyncEntry[], Map<st
   },
 };
 
+const editorConfiguration: PreferenceDescription<EditorConfiguration, EditorConfiguration> = {
+  fromStore(val) {
+    return val || {lineNumbers: null, miniMap: null, tabSize: 4};
```
Nokel81: I think we should store `Partial<EditorConfiguration>` but work with inside Lens always with `Required<EditorConfiguration>`. 

Having said that the `fromStore` function should populate the default values and `toStore` should remove the default values (even if some of the values have been set).

src/common/user-store/user-store.ts
```diff
@@ -109,7 +116,29 @@ export class UserStore extends BaseStore<UserStoreModel> /* implements UserStore
...
+
+  setLineNumbers(lineNumbers: monaco.editor.LineNumbersType) {
+    this.editorConfiguration.lineNumbers = lineNumbers;
+  }
+
+  setTabSize(tabSize: number) {
+    this.editorConfiguration.tabSize = tabSize;
+  }
+
+  enableMinimap(miniMap: Boolean) {
```
Nokel81: We should ban the class primative types, please use `boolean`.

src/common/user-store/user-store.ts
```diff
@@ -109,7 +116,29 @@ export class UserStore extends BaseStore<UserStoreModel> /* implements UserStore
...
+    this.editorConfiguration.miniMap = miniMap!= null ? {enabled: <boolean>miniMap} as monaco.editor.IEditorMinimapOptions : null;
```
Nokel81: Please don't use `<T>expr` casts. They are dangerous.

Furthermore, you shouldn't need to have a cast in a set config function.

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -42,21 +43,32 @@ import { KubeconfigSyncs } from "./kubeconfig-syncs";
...
   Other = "other"
 }
 
+enum EditorLineNumbersStyles {
+  default = "Default",
+  on = "On",
+  off = "Off",
+  relative = "Relative",
+  interval = "Interval"
+}
```
Nokel81: This enum does not correspond to the type `monaco.editor.LineNumbersType` (at least from the code in this PR). Namely, you have set the default value to be `"on"` but that value is not a valid variant of this enum.

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -70,6 +82,21 @@ export class Preferences extends React.Component {
     }));
   }
 
+  minimapOptions: SelectOption<Boolean>[] = [
+    {
+      label: "Default",
+      value: null
+    },
```
Nokel81: The default should be invisible to the user. Furthermore, if we remove this then we can use the toggle component instead of a dropdown.

src/renderer/components/dock/editor-panel.tsx
```diff
@@ -59,6 +61,14 @@ export class EditorPanel extends React.Component<Props> {
     ]);
   }
 
+  editorDidMount = (editor: monaco.editor.IStandaloneCodeEditor) =>{
+    this.editor = editor;
+    const model = monacoModelsManager.getModel(this.props.tabId);
+
+    monacoModelsManager.getModel(this.props.tabId).setValue(this.props.value ?? "");
```
Nokel81: why do you get it twice?

src/renderer/components/dock/monaco-model-manager.ts
```diff
@@ -0,0 +1,85 @@
...
+      let model;
+      const uri = this.getUri(tabId);
+
+      if(uri) {
+        model = monaco.editor.createModel(value, lang, uri); 
+      } else {
+        model = monaco.editor.createModel(value, lang); 
+      }
+        
+      !uri && (this.models = this.models.concat({ id: tabId, modelUri: model.uri, lang}));
```
Nokel81: Please do not use `&&` as an alternative to `if`.

src/renderer/components/dock/monaco-model-manager.ts
```diff
@@ -0,0 +1,85 @@
...
+    @action
+    addModel(tabId: string, { value = "", lang = "yaml" } = {}): monaco.editor.ITextModel {
+      let model;
+      const uri = this.getUri(tabId);
+
+      if(uri) {
+        model = monaco.editor.createModel(value, lang, uri); 
+      } else {
+        model = monaco.editor.createModel(value, lang); 
+      }
```
Nokel81: This could be a ternary.

src/renderer/components/dock/monaco-model-manager.ts
```diff
@@ -0,0 +1,85 @@
...
+      return this.models.find( model => model.id == tabId)?.modelUri;
+    }
+
+    @action
+    removeModel(tabId: string) {
+      const uri = this.getUri(tabId);
+
+      this.models = this.models.filter(v => v.id != tabId);
+      monaco.editor.getModel(uri).dispose();
+    }
```
Nokel81: If this gets called before the reactions and `componentDidMount` methods gets called the I would assume that `getModel` would return `undefined`.

src/common/user-store/user-store.ts
pashevskii: We decided do not to give control on each editor type to the user. Basically, this class stores global variables overriding default values.

src/common/user-store/user-store.ts
pashevskii: ok, how can I throw null within primitive type? This method contains 3 possible results, true/false and removing minimap model at all (default option)

Nokel81: `boolean | null`

src/renderer/components/+preferences/preferences.tsx
pashevskii: I set enum's value as a parameter in setLineNumbers(),  it works without issues:
image

Also, LineNumbersType allows to use "on":
  export type LineNumbersType = 'on' | 'off' | 'relative' | 'interval' | ((lineNumber: number) => string);

src/common/user-store/preferences-helpers.ts
pashevskii: Default values are pulled from export const defaultEditorConfig, store saves only changed values

src/renderer/components/+preferences/preferences.tsx
pashevskii: I'm not sure that it is really necessary. Default value hides minimap in the places, where we don't expect more data than one page (kube-config/ details).

src/renderer/components/+preferences/preferences.tsx
Nokel81: That is because select will just default to "string" in that case.

src/common/user-store/user-store.ts
Nokel81: That seems confusing, either we should have a single config for all (which we don't because some of the defaults are different by the "Type" of tab or we should allow configuring each type separately (as you did previously).

src/common/user-store/preferences-helpers.ts
Nokel81: That is not what those two lines do though... The goal should be to produce `undefined` if all the values are the "default".

src/common/user-store/preferences-helpers.ts
```diff
@@ -24,13 +24,32 @@ import path from "path";
...
 
+export interface EditorConfiguration {
+  miniMap?: monaco.editor.IEditorMinimapOptions;
+  lineNumbers?: monaco.editor.LineNumbersType;
+  tabSize?: number;
+}
+
+
+export enum EditorType {
+  TABS = "tabs",
```
aleksfront: Maybe `DOCK_TABS` is more preferable name which contains a bit more information about type.

src/common/user-store/preferences-helpers.ts
```diff
@@ -24,13 +24,32 @@ import path from "path";
...
+  DETAILS = "details",
+  KUBECONFIG = "kubeconfig",
+}
+
+export const defaultEditorConfig  = new Map<EditorType, EditorConfiguration>([
+  [EditorType.TABS, {miniMap: {enabled: true}, lineNumbers: "on"}],
```
aleksfront: `lineNumber: "on"` here. However, line number doesn't show up in Create Resource Tab with default mode on. A bug?

no line numbers
my editor config

src/common/user-store/user-store.ts
aleksfront: @Nokel81 This is trade-off before we get users's feedback about configuration and good UX to present configuration for each tab (if needed, right now not so many options to be configured)

aleksfront: @pashevskii Why we have `null` value for each option, but not the default configuration (`default`, `default`, `2`)?

src/common/user-store/user-store.ts
```diff
@@ -109,7 +116,29 @@ export class UserStore extends BaseStore<UserStoreModel> /* implements UserStore
...
+  setLineNumbers(lineNumbers: monaco.editor.LineNumbersType) {
+    this.editorConfiguration.lineNumbers = lineNumbers;
+  }
+
+  setTabSize(tabSize: number) {
+    this.editorConfiguration.tabSize = tabSize;
+  }
+
+  enableMinimap(miniMap: boolean | null) {
+    this.editorConfiguration.miniMap = miniMap!= null ? {enabled: miniMap} : null;
```
aleksfront: Code clean up thoughts: these methods shouldn't be in `user-store` directly. Or at least better to name them like `setEditorNumbers()`, `setEditorTabSize()` etc.

Maybe worth to move these to another store and create one more level of abstraction (in another refactoring PR).

src/renderer/components/+preferences/preferences.tsx
aleksfront: Why not use monaco type directly and remove this one?

```
export type LineNumbersType = 'on' | 'off' | 'relative' | 'interval' | ((lineNumber: number) => string);
```

And maybe this is personal preference, but I prefere literal types over enums where possible.

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -70,6 +82,21 @@ export class Preferences extends React.Component {
     }));
   }
 
+  minimapOptions: SelectOption<Boolean>[] = [
```
aleksfront: Please move this options to separate `<EditorSettings>` component to prevent Preferences to grow up.

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -284,6 +316,43 @@ export class Preferences extends React.Component {
             }
           </section>
         )}
+        {this.activeTab == Pages.Editor && (
+          <>
```
aleksfront: Please move this contents to separate `<EditorSettings>` component to prevent Preferences to grow up.

src/renderer/components/+workloads-pods/pod-details-affinities.tsx
```diff
@@ -42,11 +46,12 @@ export class PodDetailsAffinities extends React.Component<Props> {
       <DrawerItem name="Affinities" className="PodDetailsAffinities">
         <DrawerParamToggler label={affinitiesNum}>
           <div className="ace-container">
-            <AceEditor
-              mode="yaml"
+            <MonacoEditor
+              options = {{ readOnly: true, ...UserStore.getInstance().getEditorOptions(EditorType.DETAILS)}}
+              className={cssNames( "MonacoEditor")}
+              theme={ThemeStore.getInstance().activeTheme.monacoTheme}
+              language="yaml"
```
aleksfront: Please consider to create wrapper for `MonacoEditor` from `react-monaco-editor` in order to prevent passing duplicated props for every editor instance such as:
* className
* theme
* language

src/renderer/components/dock/monaco-model-manager.ts
```diff
@@ -0,0 +1,79 @@
...
+    id?: TabId;
+    modelUri?: monaco.Uri;
+    lang?: string;
+  }
+  
+export interface ModelsState {
+    models: ModelEntry[];
+  } 
+  
+
```
aleksfront, line 34: Redundant line.

src/renderer/components/dock/monaco-model-manager.ts
```diff
@@ -0,0 +1,79 @@
...
+export class MonacoModelsManager implements ModelsState {
```
aleksfront, line 35: `MonacoModelsManager` -> `MonacoModels`.

src/renderer/components/dock/monaco-model-manager.ts
```diff
@@ -0,0 +1,79 @@
...
+    private modelsArray = new Array<ModelEntry>();
```
aleksfront: `modelsArray` -> `models`

src/renderer/components/dock/monaco-model-manager.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @action
+    addModel(tabId: string, { value = "", lang = "yaml" } = {}): monaco.editor.ITextModel {
+      const uri = this.getUri(tabId);
+      const model = uri? monaco.editor.createModel(value, lang, uri) : monaco.editor.createModel(value, lang); 
+        
+      if(!uri) this.models = this.models.concat({ id: tabId, modelUri: model.uri, lang});
+
+      return model;
```
aleksfront: Why it returns something? By it's name it should add new model to `modelsArray` and shouldn't return anything. Also, the returned value doesn't get used anywhere.

src/renderer/components/dock/monaco-model-manager.ts
```diff
@@ -0,0 +1,79 @@
...
+    set models(models: ModelEntry[]) {
+      this.modelsArray = models;
+    }
+
+    @action
+    addModel(tabId: string, { value = "", lang = "yaml" } = {}): monaco.editor.ITextModel {
+      const uri = this.getUri(tabId);
+      const model = uri? monaco.editor.createModel(value, lang, uri) : monaco.editor.createModel(value, lang); 
+        
+      if(!uri) this.models = this.models.concat({ id: tabId, modelUri: model.uri, lang});
```
aleksfront: It changes `this.models` only if `uri` is undefined? Is it right behavior?

src/renderer/components/dock/monaco-model-manager.ts
```diff
@@ -0,0 +1,79 @@
...
+    
+    get models(){
+      return this.modelsArray;
+    }
+
+    set models(models: ModelEntry[]) {
+      this.modelsArray = models;
+    }
+
+    @action
```
aleksfront: We don't need `@action` here because we don't change any observables inside.

src/renderer/components/dock/monaco-model-manager.ts
```diff
@@ -0,0 +1,79 @@
...
+  } 
+  
+
+export class MonacoModelsManager implements ModelsState {
+    private modelsArray = new Array<ModelEntry>();
+
+    constructor() {
+      makeObservable(this);
+    }
+
```
aleksfront: We don't have any observables in class. This can be omitted.

aleksfront:
Found 2 issues while testing:
1. App crash on first Monaco editor start
2. Missing color scheme.

monaco app crash
missing monaco colors

src/common/user-store/preferences-helpers.ts
Nokel81: That is because it is supposed to be "On" I think. I mentioned that the default value is not a valid value in one of my reviews.

src/renderer/components/+preferences/preferences.tsx
Nokel81: You prefer literals? Also I don't think we should support the line number function to begin with.

src/common/user-store/user-store.ts
Nokel81: No I was meaning why have the different editor types at all? For a simpler initial we could just use all the same.

src/renderer/components/+workloads-pods/pod-details-affinities.tsx
Nokel81: I think that language should be always specified.

src/renderer/components/dock/monaco-model-manager.ts
Nokel81: Also we don't tend to use the array constructor directly.

src/renderer/components/+workloads-pods/pod-details-affinities.tsx
aleksfront: 100% of the time we use `yaml` so it can be just default value without passing it everywhere.

src/common/user-store/user-store.ts
Nokel81: Why would they be in a different store? The `UserStore` is really all about user preferences now. Though I agree the names could be better.

src/common/user-store/user-store.ts
aleksfront: Yes, it was decided because same preferences doesn't look good in every places. For example, minimap isnt look good on small editors... But thinking about it more maybe it's really better to stick with same values for every editor type? Let it just show minimap everywhere along with line numbers? @pashevskii @Nokel81

Nokel81: I think that would be best, at least for the first version.

src/common/user-store/preferences-helpers.ts
pashevskii: removed at all

src/common/user-store/user-store.ts
pashevskii: fixed

src/renderer/components/dock/monaco-model-manager.ts
pashevskii: Exactly, addModel can be called in two cases, adding a new tab or recovering an existing tab from the store The first case means that URI was not created earlier, therefore the model is not in models and we should add it into this.models object

src/renderer/components/dock/monaco-model-manager.ts
pashevskii: you right,  fixed

src/common/user-store/preferences-helpers.ts
```diff
@@ -222,6 +237,31 @@ const syncKubeconfigEntries: PreferenceDescription<KubeconfigSyncEntry[], Map<st
...
+const editorConfiguration: PreferenceDescription<EditorConfiguration, EditorConfiguration> = {
+  fromStore(val) {
+    const result: EditorConfiguration = {};
+
+    result.lineNumbers = val?.lineNumbers ?? defaultEditorConfig.lineNumbers;
+    result.tabSize = val?.tabSize ?? defaultEditorConfig.tabSize;
+    result.miniMap = val?.miniMap ?? defaultEditorConfig.miniMap;
+
+    return result;
+  },
```
aleksfront: We can use `merge` utility here to save all props at once https://lodash.com/docs/4.17.15#merge

src/common/user-store/preferences-helpers.ts
```diff
@@ -222,6 +237,31 @@ const syncKubeconfigEntries: PreferenceDescription<KubeconfigSyncEntry[], Map<st
...
+
+    if (val.lineNumbers != defaultEditorConfig.lineNumbers)
+      result.lineNumbers = val.lineNumbers;
+    if (val.tabSize != defaultEditorConfig.tabSize)
+      result.tabSize = val.tabSize;
+    if (val.miniMap?.enabled != defaultEditorConfig.miniMap?.enabled)
+      result.miniMap = val.miniMap;
+
+    return result;
+  },
```
aleksfront: Why we compare settings with default ones before saving them to store? I guess we need to skip this comparisons.

src/renderer/components/+apps-releases/release-details.tsx
```diff
@@ -158,15 +159,16 @@ export class ReleaseDetails extends Component<Props> {
...
-            mode="yaml"
+          <MonacoEditor
+            language="yaml"
             value={values}
             onChange={text => this.values = text}
-            className={cssNames({ loading: valuesLoading })}
-            readOnly={valuesLoading || this.showOnlyUserSuppliedValues}
+            theme={ThemeStore.getInstance().activeTheme.monacoTheme}
+            className={cssNames( "MonacoEditor", { loading: true })}
```
aleksfront: Is there a reason to always have `loading` class enabled?

src/renderer/components/+custom-resources/crd-details.tsx
```diff
@@ -143,11 +146,12 @@ export class CRDDetails extends React.Component<Props> {
         {validation &&
         <>
           <DrawerTitle title="Validation"/>
-          <AceEditor
-            mode="yaml"
-            className="validation"
+          <MonacoEditor
+            options = {{readOnly: true, ...UserStore.getInstance().getEditorOptions()}}
+            className={cssNames( "MonacoEditor", "validation")}
```
aleksfront: Please remove the space after `cssNames( ` here and in other MonacoEditor instances.

src/renderer/components/dock/editor-panel.tsx
```diff
@@ -59,6 +60,14 @@ export class EditorPanel extends React.Component<Props> {
     ]);
   }
 
+  editorDidMount = (editor: monaco.editor.IStandaloneCodeEditor) =>{
```
aleksfront: Add space between `=>{`

src/renderer/components/dock/editor-panel.tsx
```diff
@@ -70,17 +79,14 @@ export class EditorPanel extends React.Component<Props> {
 
   onTabChange = () => {
     this.editor.focus();
+    monacoModelsManager.getModel(this.props.tabId).setValue(this.props.value ?? "");
```
aleksfront: Nit: extract `monacoModelsManager.getModel(this.props.tabId)` to separate variable to avoid duplicating.

src/renderer/components/dock/monaco-model-manager.ts
```diff
@@ -18,75 +18,44 @@
...
-  }
+export class MonacoModelsManager implements ModelsState {
+    models: ModelEntry[] = [];
 
-  .ace_meta.ace_tag {
-    color: var(--dockEditorTag);
-  }
+    addModel(tabId: string, { value = "", lang = "yaml" } = {}) {
+      const uri = this.getUri(tabId);
+      const model = uri? monaco.editor.createModel(value, lang, uri) : monaco.editor.createModel(value, lang);
```
aleksfront: Wouldn't just `monaco.editor.createModel(value, lang, uri);` to be enough here?

src/common/user-store/preferences-helpers.ts
Nokel81: I requested this because the goal was to be able to change the defaults in the future.

nevalla:
Tests are failing:

```
SyntaxError: Cannot use import statement outside a module

      19 |  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
      20 |  */
    > 21 | import {monaco} from "react-monaco-editor";
         | ^
      22 |
      23 | export type TabId = string;
      24 |

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1350:14)
      at Object.<anonymous> (src/renderer/components/dock/monaco-model-manager.ts:21:1)
```

Otherwise looking good!

src/common/user-store/preferences-helpers.ts
aleksfront: Probably it's not a good idea to create this logic for some future possible scenarios. I suggest using migrations for such cases.

pashevskii:
> Tests are failing:
> 
> ```
> SyntaxError: Cannot use import statement outside a module
> 
>       19 |  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
>       20 |  */
>     > 21 | import {monaco} from "react-monaco-editor";
>          | ^
>       22 |
>       23 | export type TabId = string;
>       24 |
> 
>       at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1350:14)
>       at Object.<anonymous> (src/renderer/components/dock/monaco-model-manager.ts:21:1)
> ```
> 
> Otherwise looking good!

Thanks, 
I tried to fix it, but the cause of the errors is more complicated than was expected. I faced a number of similar threads ([un](https://github.com/react-monaco-editor/react-monaco-editor/issues/176), [dos](https://github.com/react-monaco-editor/react-monaco-editor/issues/133) ,[tres](https://github.com/Microsoft/monaco-editor/issues/996)) but it has not helped me. I tried to write mock, but it is rocket science for me. Could anybody please help me with it?

nevalla:
> I tried to fix it, but the cause of the errors is more complicated than was expected. I faced a number of similar threads ([un](https://github.com/react-monaco-editor/react-monaco-editor/issues/176), [dos](https://github.com/react-monaco-editor/react-monaco-editor/issues/133) ,[tres](https://github.com/Microsoft/monaco-editor/issues/996)) but it has not helped me. I tried to write mock, but it is rocket science for me. Could anybody please help me with it?

@aleksfront please could you help here

pashevskii:
image
thanks @aleksfront, unit tests are fixed

aleksfront (dismissed):
LGTM!

nevalla (approved):
LGTM

aleksfront (approved):

PR #2956, opened 2021-06-03
Make customizing ItemListLayout header explicitly reducable 

Nokel81:

This was inspired by wanting to move `<NamespaceSelectFilter>` into `KubeObjectListLayout` since it really didn't belong in `<ItemListLayout>`. However, to do that effectively and with the least amount of code in the consumers of either of those two components, `<ItemListLayout>` was changed to "reduce" or "fold" over all the customizers.

Video: (only minor visual changes)

https://user-images.githubusercontent.com/8225332/120715486-a3f5db00-c492-11eb-9dcb-a23a1866ce56.mov

### DIFF
--- a/src/renderer/components/+apps-helm-charts/helm-charts.tsx
+++ b/src/renderer/components/+apps-helm-charts/helm-charts.tsx
@@ -30,7 +30,6 @@ import type { HelmChart } from "../../api/endpoints/helm-charts.api";
 import { HelmChartDetails } from "./helm-chart-details";
 import { navigation } from "../../navigation";
 import { ItemListLayout } from "../item-object-list/item-list-layout";
-import { SearchInputUrl } from "../input";
 
 enum columnId {
   name = "name",
@@ -92,9 +91,12 @@ export class HelmCharts extends Component<Props> {
             (chart: HelmChart) => chart.getAppVersion(),
             (chart: HelmChart) => chart.getKeywords(),
           ]}
-          customizeHeader={() => (
-            <SearchInputUrl placeholder="Search Helm Charts" />
-          )}
+          customizeHeader={({ searchProps }) => ({
+            searchProps: {
+              ...searchProps,
+              placeholder: "Search Helm Charts...",
+            },
+          })}
           renderTableHeader={[
             { className: "icon", showWithColumn: columnId.name },
             { title: "Name", className: "name", sortBy: columnId.name, id: columnId.name },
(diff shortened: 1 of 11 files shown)


src/renderer/components/+custom-resources/crd-resources.tsx
```diff
@@ -101,6 +101,15 @@ export class CrdResources extends React.Component<Props> {
           (item: KubeObject) => item.getSearchFields(),
         ]}
         renderHeaderTitle={crd.getResourceTitle()}
+        customizeHeader={[
```
jim-docker: why is this one `HeaderCustomizer` in an array?

src/renderer/components/+apps-helm-charts/helm-charts.tsx
```diff
@@ -92,9 +91,12 @@ export class HelmCharts extends Component<Props> {
             (chart: HelmChart) => chart.getAppVersion(),
             (chart: HelmChart) => chart.getKeywords(),
           ]}
-          customizeHeader={() => (
-            <SearchInputUrl placeholder="Search Helm Charts" />
-          )}
+          customizeHeader={({ searchProps }) => ({
+            searchProps: {
+              ...searchProps,
+              placeholder: "Search Helm Charts",
```
jim-docker: Releases could use this too

jim-docker (dismissed):

src/renderer/components/+apps-helm-charts/helm-charts.tsx
Nokel81: Ah good point.

src/renderer/components/+custom-resources/crd-resources.tsx
Nokel81: forgot about this one.

Nachasic (dismissed):

extensions/kube-object-event-status/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kube-object-event-status",
-  "version": "0.1.0",
+  "version": "0.0.1",
```
jim-docker: Why do you need this? Something still fishy with the rebase?

Nokel81: yes fishy indeed.

Nokel81:
@jim-docker fixed.

jim-docker (approved):

PR #2963, opened 2021-06-04
Await `onActivate`/`onDeactive` of extensions

Nokel81:

We currently document these two handlers as being able to be `async`. However, in the code we do not treat them as such.

Furthermore there wasn't any defence against these functions throwing. Finally, since `LensExtension` has a list of disposers itself, we can just hook into that for the registering.

### DIFF
--- a/src/extensions/lens-extension.ts
+++ b/src/extensions/lens-extension.ts
@@ -20,11 +20,11 @@
  */
 
 import type { InstalledExtension } from "./extension-discovery";
-import { action, observable, reaction, makeObservable } from "mobx";
+import { action, observable, makeObservable } from "mobx";
 import { FilesystemProvisionerStore } from "../main/extension-filesystem";
 import logger from "../main/logger";
 import type { ProtocolHandlerRegistration } from "./registries";
-import { disposer } from "../common/utils";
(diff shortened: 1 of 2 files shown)


src/extensions/lens-extension.ts
```diff
@@ -83,59 +83,44 @@ export class LensExtension {
   }
 
   @action
-  async enable() {
-    if (this.isEnabled) return;
-    this.isEnabled = true;
-    this.onActivate?.();
-    logger.info(`[EXTENSION]: enabled ${this.name}@${this.version}`);
+  async enable(register: (ext: LensExtension) => Promise<Disposer[]>) {
```
jakolehm, line 86: Not directly related to this PR but could we move `enable/disable` methods to `ExtensionLoader`? It feels like they are in a wrong place atm.

jakolehm (approved):
LGTM

Nachasic (approved):

src/extensions/lens-extension.ts
Nokel81: Probably, see #2245 but I think we can and should do away with the registries as well (non-breaking change)

PR #2969, opened 2021-06-07
Always publish npm package on master branch updates

msa0311:


jakolehm (dismissed):

nevalla:
Does this work if package.version does not change?

jakolehm:
> Does this work if package.version does not change?

It should change here: https://github.com/lensapp/lens/blob/master/build/set_npm_version.ts#L31-L37

nevalla (dismissed):

PR #2976, opened 2021-06-07
Fix NPM_RELEASE_TAG in Makefile

jakolehm:


aleksfront (approved):

Nokel81 (changes requested):
Is this to fix the failing workflows currently? Because if so, this is not the fix.

Furthermore, aren't these two syntaxes equivalent? Or is there a difference to Make on linux?

jakolehm:
> Furthermore, aren't these two syntaxes equivalent? Or is there a difference to Make on linux?

Original errors out because it does not get any tag (even if I run it manually). It basically never defaults.

Nokel81:
Okay sounds good, though https://github.com/lensapp/lens/blob/master/.github/workflows/publish-master-npm.yml#L11 is still incorrect

Nokel81 (approved):

msa0311 (approved):

PR #2981, opened 2021-06-07
Improve Table typing and remove annotations on callbacks

Nokel81:

This is a new PR instead of https://github.com/lensapp/lens/pull/2844 (which claims that I merged, which I don't remember doing since it was failing).

I have since reverted those changes on master (https://github.com/lensapp/lens/pull/2980)

This PR also changes where the `CrdResourceDetails` component is added to `KubeObjectDetails`. It is now added to the `KubeObjectDetailsRegistry` just like everything else.

### DIFF
--- a/src/renderer/api/endpoints/resource-applier.api.ts
+++ b/src/renderer/api/endpoints/resource-applier.api.ts
@@ -30,7 +30,7 @@ export const resourceApplierApi = {
     "kubectl.kubernetes.io/last-applied-configuration"
   ],
 
-  async update<D extends KubeObject>(resource: object | string): Promise<D> {
+  async update<K extends KubeObject>(resource: object | string): Promise<K | null> {
     if (typeof resource === "string") {
       resource = jsYaml.safeLoad(resource);
     }
@@ -48,7 +48,7 @@ export const resourceApplierApi = {
           }
         });
 
-        return items.length === 1 ? items[0] : items;
+        return items[0] as K ?? null;
       });
   }
 };
--- a/src/renderer/api/kube-object.ts
+++ b/src/renderer/api/kube-object.ts
@@ -31,13 +31,13 @@ import { resourceApplierApi } from "./endpoints/resource-applier.api";
 import { hasOptionalProperty, hasTypedProperty, isObject, isString, bindPredicate, isTypedArray, isRecord } from "../../common/utils/type-narrowing";
 import _ from "lodash";
 
-export type IKubeObjectConstructor<T extends KubeObject = any> = (new (data: KubeJsonApiData | any) => T) & {
+export type KubeObjectConstructor<K extends KubeObject> = (new (data: KubeJsonApiData | any) => K) & {
   kind?: string;
   namespaced?: boolean;
   apiBase?: string;
 };
 
-export interface IKubeObjectMetadata {
+export interface KubeObjectMetadata {
   uid: string;
   name: string;
   namespace?: string;
@@ -63,7 +63,7 @@ export interface IKubeObjectMetadata {
   }[];
 }
 
-export interface IKubeStatus {
+export interface KubeStatusData {
   kind: string;
   apiVersion: string;
   code: number;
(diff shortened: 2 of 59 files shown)


src/renderer/api/endpoints/resource-applier.api.ts
```diff
@@ -48,7 +48,7 @@ export const resourceApplierApi = {
           }
         });
 
-        return items.length === 1 ? items[0] : items;
+        return items[0] as K ?? null;
```
jim-docker, line 51: this changes behaviour (no longer optionally returns `items`)

src/renderer/api/kube-object.ts
```diff
@@ -31,7 +31,7 @@ import { resourceApplierApi } from "./endpoints/resource-applier.api";
 import { hasOptionalProperty, hasTypedProperty, isObject, isString, bindPredicate, isTypedArray, isRecord } from "../../common/utils/type-narrowing";
 import _ from "lodash";
 
-export type IKubeObjectConstructor<T extends KubeObject = any> = (new (data: KubeJsonApiData | any) => T) & {
+export type IKubeObjectConstructor<K extends KubeObject> = (new (data: KubeJsonApiData | any) => K) & {
```
jim-docker: Doesn't the `I` imply interface?

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -252,10 +252,10 @@ export class Catalog extends React.Component<Props> {
           { title: "Labels", className: css.labelsCell, id: "labels" },
           { title: "Status", className: css.statusCell, sortBy: sortBy.status, id: "status" },
         ]}
-        customizeTableRowProps={(item: CatalogEntityItem<CatalogEntity>) => ({
+        customizeTableRowProps={item => ({
           disabled: !item.enabled,
         })}
-        renderTableContents={(item: CatalogEntityItem<CatalogEntity>) => [
+        renderTableContents={item => [
```
jim-docker, line 229: nit: change `item` to `entity`, to be consistent?

src/renderer/components/+workloads-overview/overview-statuses.tsx
```diff
@@ -47,7 +47,7 @@ const resources: KubeResource[] = [
 export class OverviewStatuses extends React.Component {
   @boundMethod
   renderWorkload(resource: KubeResource): React.ReactElement {
-    const store = workloadStores[resource];
+    const store = workloadStores.get(resource);
     const items = store.getAllByNs(namespaceStore.contextNamespaces);
```
jim-docker, line 56: store could be `undefined` so?
```suggestion
    const items = store?.getAllByNs(namespaceStore.contextNamespaces);
```

src/renderer/components/+workloads-pods/pod-tolerations.tsx
```diff
@@ -66,20 +59,24 @@ export function PodTolerations({ tolerations }: Props) {
     <Table
       tableId="workloads_pod_tolerations"
       selectable
+      items={tolerations}
```
jim-docker, line 64: does this change behaviour?

jim-docker (commented):
I'll try to do some testing tomorrow

src/renderer/api/endpoints/resource-applier.api.ts
Nokel81: I believe that was a bug given the explicit return type.

src/renderer/api/kube-object.ts
Nokel81: Yes, will change (since we don't export it to the extension API)

src/renderer/components/+catalog/catalog.tsx
Nokel81: Yes

src/renderer/components/+workloads-pods/pod-tolerations.tsx
Nokel81: No, it is specifically designed not to. This is just a better way (from a typescript perspective) for type inference.

jim-docker (dismissed):

jim-docker (approved):

PR #2986, opened 2021-06-07
RegEx and Wildcard Search in Item-List-Layout

### LINKED ISSUE #2479: Support wildcard in search-box
Use wildcard in the search-box

**Why is this needed**:
Enhance search capabilities on a large cluster.

Win 10

LucaDev:
This PR adds the ability to use * as wildcards and "regex:[expression]" to search through Item-Lists

I used indexOf instead of includes because testing showed it offers quite a bit better performance in chrome based browsers.

Fixes:  https://github.com/lensapp/lens/issues/2479

### DIFF
--- a/src/renderer/components/item-object-list/item-list-layout.tsx
+++ b/src/renderer/components/item-object-list/item-list-layout.tsx
@@ -171,7 +171,19 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
           return searchFilters.some(getTexts => {
             const sourceTexts: string[] = [getTexts(item)].flat().map(normalizeText);
 
-            return sourceTexts.some(source => searchTexts.some(search => source.includes(search)));
+            return searchTexts.some(search => {
+              if(search.startsWith("regex:")) {
+                const matcher = new RegExp(search.substring(6).trim());
+
+                return sourceTexts.some(source => matcher.test(source));
+              } else if(search.indexOf("*") !== -1) {
+                const matcher = new RegExp(search.split("*").map(string => string.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")).join(".*"));
+                
+                return sourceTexts.some(source => matcher.test(source));
+              } else {
+                return sourceTexts.some(source => source.indexOf(search) !== -1);
+              }
+            });
           });
         });
       }


src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -171,7 +171,19 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
           return searchFilters.some(getTexts => {
             const sourceTexts: string[] = [getTexts(item)].flat().map(normalizeText);
 
-            return sourceTexts.some(source => searchTexts.some(search => source.includes(search)));
+            return searchTexts.some(search => {
+              if(search.startsWith("regex:")) {
+                const matcher = new RegExp(search.substring(6).trim());
+
+                return sourceTexts.some(source => matcher.test(source));
+              } else if(search.indexOf("*") !== -1) {
```
BlackHole1, line 179: I personally feel that this should be removed.

Because this doesn't look obvious, it feels like a mix of "normal string" and "RegExp"

If it's for ease of input, I feel we should provide a button to selectively turn on whether to search in RegExp mode.

Just like idea:

图片

LucaDev: I was thinking about Googles advanced search queries like site:..., filletype:..., etc. You really don't often use them - that's why they are kept hidden. On the other hand.. Lens is for advanced users. There is not that big of a need to keep everything as simple as the google search.

Another interesting idea would be to autocomplete those advanced queries and allow the user to chain them(of course we would have to support more complex ones one the future). Like "regex:prometheus-[0-9] memory:>500M". A bit like Gitlab does it 
Bildschirmfoto von 2021-06-08 09-09-33

LucaDev: @Nokel81 @jakolehm @nevalla what do you think? Should I remove the "regex:" support or leave it in for now?

BlackHole1: I mean remove `*` , Instead of `regex:`

LucaDev: Oh really? I think a simple glob search is much more intuitive than regex in this case. We definitely need more opinions on this :D

BlackHole1: The main thing is that here, there will be a problem, as I said before: 

> Because this doesn't look obvious, it feels like a mix of "normal string" and "RegExp"

I think it might be better to use `glob:` instead of `*`

BlackHole1: Of course, I feel it would be better if a few buttons were provided

LucaDev: As a long-term Linux user, I'm kinda used to globs working everywhere. Nevertheless - I get your point. I'd love to hear a couple more opinions on this. In the end, we should find the best compromise between functionality, usability, and simplicity.

Nokel81: I think that just having glob search would be an improvement.

Furthermore, I would prefer having a UI and having structured search for more complex searching than trying to invent a new language.

So, if this PR could focus on adding glob support (which covers some of the regex case) I think it would be merged more quickly.

jakolehm: I also think that simple glob search is more intuitive for most users. As @Nokel81 said, we could have just glob search in this PR and then follow-up issue/PR about regex.

Nokel81:

PR #2995, opened 2021-06-08
Allow to close tabs using middle-clicks

LucaDev:
Most Browsers allow closing tabs by using the middle mouse button (usually pushing in on the scroll wheel) when hovering over a tab. I think the Labs in Lens should behave the same to make it even more intuitive.

Additionally, I explicitly added the Shift Key to the "Ctrl + W" label(to close a tab) to make it easier to understand

### DIFF
--- a/src/renderer/components/dock/dock-tab.tsx
+++ b/src/renderer/components/dock/dock-tab.tsx
@@ -23,7 +23,7 @@ import "./dock-tab.scss";
 
 import React from "react";
 import { observer } from "mobx-react";
-import { boundMethod, cssNames, prevDefault } from "../../utils";
+import { boundMethod, cssNames, prevDefault, isMiddleClick } from "../../utils";
 import { dockStore, IDockTab } from "./dock.store";
 import { Tab, TabProps } from "../tabs";
 import { Icon } from "../icon";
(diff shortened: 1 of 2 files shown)


src/renderer/components/dock/dock-tab.tsx
```diff
@@ -88,13 +88,13 @@ export class DockTab extends React.Component<DockTabProps> {
...
     const label = (
-      <div className="flex gaps align-center">
+      <div className="flex gaps align-center" onAuxClick={isMiddleClick(prevDefault(this.close))}>
         <span className="title" title={title}>{title}</span>
         {moreActions}
         {!pinned && (
           <Icon
             small material="close"
-            title="Close (Ctrl+W)"
+            title="Close (Ctrl+Shift+W)"
```
Nokel81, line 97: Is this what is currently used anyway?

LucaDev: Yeah. The Shift Key was indicated by the uppercase "W" instead of a lowercase one I suppose. On both Windows and Linux I wasn't able to close the tab without pressing the shift key

Nokel81: Oh I figure that any keys in such hints are case insensitive anyway. Thanks for this change.

Nokel81 (approved):

PR #2997, opened 2021-06-08
Adding Topbar extension registries

### LINKED ISSUE #2968: Add TopBar to extension full-page view
`TopBar` introduced in https://github.com/lensapp/lens/issues/2749 should be required for extensions that uses full-page views.

topbar

aleksfront:
Allowing extensions to add custom `TopBar` items.

Example of icon added by extension:
topbar items

Resolves #2968


src/renderer/components/+catalog/catalog.tsx
```diff
@@ -269,7 +264,7 @@ export class Catalog extends React.Component<Props> {
           item.kind,
           item.source,
           item.labels.map((label) => <Badge key={label} label={label} title={label} />),
-          { title: item.phase, className: kebabCase(item.phase) }
+          { title: item.phase, className: cssNames({ [styles.connected]: item.phase == "connected" }) }
```
jakolehm: Why did we change this? 🤔 I think the previous logic was more generic?

src/renderer/components/cluster-manager/topbar.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+import React from "react";
+import { render } from "@testing-library/react";
+import "@testing-library/jest-dom/extend-expect";
+
+jest.mock("../../../extensions/registries");
+
+import { topBarRegistry } from "../../../extensions/registries";
+import { TopBar } from "../layout/topbar";
+
+describe("<TopBar/>", () => {
```
jakolehm, line 28: Great to see component tests!

src/renderer/components/layout/topbar.module.css
```diff
@@ -42,4 +43,7 @@
   align-items: center;
   display: flex;
   height: 100%;
+}
+
+.extensionItem {
```
jakolehm: Is this needed?

src/renderer/components/cluster-manager/cluster-manager.tsx
```diff
@@ -35,12 +35,16 @@ import { Extensions, extensionsRoute } from "../+extensions";
...
 import { Welcome, welcomeRoute, welcomeURL } from "../+welcome";
+import { ClusterTopbar } from "./cluster-topbar";
+import { CatalogTopbar } from "./catalog-topbar";
 
 @observer
 export class ClusterManager extends React.Component {
   render() {
     return (
       <div className="ClusterManager">
+        <Route component={CatalogTopbar} {...catalogRoute} />
```
jakolehm: This looks a bit odd, why do we route to top bars?

src/renderer/components/+catalog/catalog.tsx
aleksfront: Such cases now needs a bit different approach, because of use CSS Modules. We can't just add `connected` classname in here because all classnames from `catalog.module.css` will be converted to unique ones, so literal `connected` will never meet `styles.connected` class name.

aleksfront: Basically previous logic can be untouched if using `*.scss` instead of `*.module.css` for styling.

src/renderer/components/cluster-manager/cluster-manager.tsx
aleksfront: If showing TopBar without route differentiation, we will just move it inside `<TopBar/>` component itself... It shows different content (label and close button) for Catalog as well as for Cluster views. So this way we would need parsing current route, match it to `catalogUrl()`, `/cluster/:clusterId` somewhere inside `componentDidMount()`. I found using `<Route/>` more clear and explicit.

src/renderer/components/+catalog/catalog.tsx
jakolehm: So for dynamic use cases we should use inline css?

src/renderer/components/layout/topbar.module.css
aleksfront: Was thinking about it as a placeholder for future styling. Removing for now.

src/renderer/components/cluster-manager/cluster-manager.tsx
jakolehm: Ok, for me it's a bit confusing but we can always refactor this later if needed.

src/renderer/components/+catalog/catalog.tsx
aleksfront: For dynamic use cases with different statuses we could specify which classnames could be used
```
{ title: item.phase, className: styles[kebabCase(item.phase) as "connected" | "disconnected"] }
```

jakolehm: IMHO that's not dynamic if we need to define all possible phases. This is something we need to figure out at some point.

aleksfront: I agree, should we do this in separate PR then?

Nokel81: Why are `*.module.css` better than `*.scss`?

aleksfront: `*.module.css` indicates this file is using CSS Modules and all it's classnames will be transformed to unique ones. If you're asking why using clear CSS instead of SCSS - currently it's the only way to work now (PostCSS isn't configured to use modules inside `scss` files). And it seems that we're okay using css for most of our tasks - we have native variables and nested css plugin.

Nokel81: okay sounds good, I really like the nested-ness of `scss`. But if we can continue to do that then 👍 from me.

aleksfront: @jakolehm Reverted to use dynamic classnames. It was made possible by some TS types fine-tuning. So, in files where you want to use such feature, you use `makeCss` utility:

```
const css = makeCss(styles);
```

src/extensions/lens-renderer-extension.ts
```diff
@@ -44,6 +45,7 @@ export class LensRendererExtension extends LensExtension {
   commands: CommandRegistration[] = [];
   welcomeMenus: WelcomeMenuRegistration[] = [];
   catalogEntityDetailItems: CatalogEntityDetailRegistration[] = [];
+  topBarItems: TopBarRegistration[] = [];
```
Nokel81, line 48: So you decided not to customize these over the active entity?

src/extensions/registries/topbar-registry.ts
```diff
@@ -0,0 +1,34 @@
...
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import type React from "react";
+import { BaseRegistry } from "./base-registry";
+
+interface TopBarComponents {
+  Item?: React.ComponentType;
```
Nokel81, line 26: Even if you don't support it above, we can easily support it here by passing in the activeEntity. Though TBH the extensions could do that too.

src/renderer/components/cluster-manager/catalog-topbar.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { Icon } from "../icon";
+import { TopBar } from "../layout/topbar";
+import { MaterialTooltip } from "../material-tooltip/material-tooltip";
+
+export function CatalogTopbar() {
+  return (
+    <TopBar label="Catalog">
+      <div>
+        <MaterialTooltip title="Close Catalog" placement="left">
+          <Icon style={{ cursor: "default" }} material="close" onClick={() => navigate(welcomeURL())}/>
```
Nokel81, line 34: Why are we overriding the cursor here?

src/renderer/components/cluster-manager/cluster-topbar.tsx
```diff
@@ -19,26 +19,33 @@
...
 
-interface Props {
-  cluster: Cluster
+interface Props extends RouteComponentProps<ClusterViewRouteParams> {
 }
 
-export function ClusterTopbar({ cluster }: Props) {
+export const ClusterTopbar = observer((props: Props) => {
+  const getCluster = (): Cluster | undefined => {
+    return ClusterStore.getInstance().getById(props.match.params.clusterId);
```
Nokel81, line 39: Shouldn't the props be `entityId` ?

src/renderer/components/layout/topbar.tsx
```diff
@@ -22,16 +22,44 @@
...
+          if (!registration?.components?.Item) {
+            return null;
+          }
+
+          return (
+            <div key={index}>
+              {registration.components.Item}
+            </div>
+          );
+        })}
```
Nokel81, line 51: nit: filter might be easier to read.

```suggestion
        {
          items
            .filter(registration => registration?.components?.Item)
            .map(({ components: { Item } }, index) => (
              <div key={index}>
                {Item}
              </div>
            ))
        }
```

Also, shouldn't the props be passed in somewhere?

src/renderer/components/cluster-manager/catalog-topbar.tsx
aleksfront: Because draggable area on OSX hovers part of the topbar, preventing to show `cursor: pointer` on any element. But, it hovers only part of it so if you hover quickly over the button, you'll see blinking cursor. This fix prevents any blinking and shows always a default cursor here.

src/extensions/lens-renderer-extension.ts
aleksfront: Active entity can be received inside extension in any place within https://api-docs.k8slens.dev/master/extensions/api/classes/renderer.catalog.catalogentityregistry/#activeentity property.

src/renderer/components/cluster-manager/cluster-topbar.tsx
aleksfront: Props are coming automatically from the `<Route/>` wrapper.

src/renderer/components/cluster-manager/catalog-topbar.tsx
Nokel81: Sounds good.

Nokel81: Maybe add a comment.

src/renderer/components/cluster-manager/cluster-topbar.tsx
Nokel81: True, but if other topbars are registered for non-clusters, it seems weird that it would pass in the `entityId` as `clusterId`.

jakolehm (approved):

PR #3016, opened 2021-06-09
Fix buildNumber to be consistent for all package.json's

Nokel81:

### DIFF
--- a/build/set_build_version.ts
+++ b/build/set_build_version.ts
@@ -18,7 +18,7 @@
  * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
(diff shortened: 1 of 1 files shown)


build/set_build_version.ts
```diff
@@ -18,38 +18,65 @@
...
-if (versionInfo.prerelease) {
-  if (versionInfo.prerelease.includes("alpha")) {
-    buildChannel = "alpha";
-  } else {
-    buildChannel = "beta";
+function getBuildChannel(): string {
+  switch (versionInfo.prerelease[0]) {
+    case "beta":
+      return "beta";
+    case "rc":
```
jakolehm: Rc was not included on purpose. We should have just `alpha`, `beta` and `latest` (stable).

jakolehm (changes requested):

build/set_build_version.ts
Nokel81: Okay will remove and add a comment about that.

jakolehm (approved):

PR #3035, opened 2021-06-11
Fix catalog entity context menu duplicate entries

### LINKED ISSUE #3005: Duplicate 'Unpin from Hotbar' menu items
unping from hotbar

jakolehm:
Fixes #3005

### DIFF
--- a/src/renderer/components/+catalog/catalog.tsx
+++ b/src/renderer/components/+catalog/catalog.tsx
@@ -79,7 +79,7 @@ export class Catalog extends React.Component<Props> {
 
   async componentDidMount() {
     this.contextMenu = {
-      menuItems: [],
+      menuItems: observable.array([]),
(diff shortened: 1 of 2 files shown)


src/renderer/components/+catalog/catalog.tsx
```diff
@@ -169,8 +169,14 @@ export class Catalog extends React.Component<Props> {
   }
 
   renderItemMenu = (item: CatalogEntityItem) => {
+    const onOpen = () => {
+      this.contextMenu.menuItems = [];
```
Nokel81, line 173: Shouldn't this be `this.contextMenu.clear()` and we should probably do a `runInAction` so that we don't get errors.

Nokel81 (approved):
Though having tested this. It works great.

PR #3038, opened 2021-06-11
Fix <Icon>'s not using tooltip and fix toolbar not being boolean

Nokel81:
For several months now, the toolbar prop that was being passed into `KubeObjectMenuRegistry`'s items was unconditionally truthy (from the `global.toolbar` variable).

This fixes this. In conjunction I have also made it so that `tooltip`'s are more nudged by making `title?: undefined`. Since we cannot style `title`'s. I believe that this is better.

Finally, I have made it so that in our extensions at least, the `tooltip` is not rendered when the menu item is not in `toolbar` form.

### DIFF
--- a/src/renderer/components/icon/icon.tsx
+++ b/src/renderer/components/icon/icon.tsx
@@ -43,6 +43,7 @@ export interface IconProps extends React.HTMLAttributes<any>, TooltipDecoratorPr
   focusable?: boolean;        // allow focus to the icon + show .active styles (default: "true", when icon is interactive)
   sticker?: boolean;
   disabled?: boolean;
+  title?: undefined;          // to suggest to TS users to use tooltip instead
 }
 
 @withTooltip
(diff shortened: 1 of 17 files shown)


Nokel81:
Partial fix of https://github.com/lensapp/lens/issues/3095

src/renderer/components/icon/icon.tsx
```diff
@@ -43,6 +43,7 @@ export interface IconProps extends React.HTMLAttributes<any>, TooltipDecoratorPr
   focusable?: boolean;        // allow focus to the icon + show .active styles (default: "true", when icon is interactive)
   sticker?: boolean;
   disabled?: boolean;
+  title?: undefined;          // to suggest to TS users to use tooltip instead
```
aleksfront: We shouldn't touch HTML native types. Replacing `title` to `tooltip` for icons should be enough.

aleksfront (dismissed):
Looks good. Thanks!

Issue #2994, opened 2021-06-08
LENS slows system down to a crawl when SSO login expires


karlschriek:
**Describe the bug**
We use SSO login with assume-role to establish a connection to our EKS clusters. 

Typically the kubeconfig user would look something like this:

```
- name: arn:aws:eks:eu-central-1:123456789012:cluster/my-cluster
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1alpha1
      args:
      - --region
      - eu-central-1
      - eks
      - get-token
      - --cluster-name
      - my-cluster
      - --role
      - arn:aws:iam::123456789012:role/kubernetes-admin-role
      command: aws
      env:
      - name: AWS_PROFILE
        value: my-profile
```
And the AWS profile would look like this:

```
[profile my-profile]
region = eu-central-1
sso_start_url = https://my-company.awsapps.com/start
sso_account_id = 123456789012
sso_role_name = PowerUserAccess
sso_region = eu-central-1
```

In order to establish a connection to the AWS cluster, the profile "my-profile" needs to first authenticate via SSO login. E.g.:

`aws sso login --profile my-profile`

At this point it is possible for LENS to connect to the cluster normally. This SSO login however expires once or twice per day. LENS does not appear to handle this expiry particularly well. Most mornings when I start up my workstation LENS is drawing almost all of my CPU resources (presumably in an attempt to connect to my cluster, but failing to do so since SSO login has expired). The issue is only resolved by renewing the SSO login.

**To Reproduce**

Login via SSO. Wait of SSO to expire.

**Expected behavior**
I would expect lens to either give up attempting to connect or in some other way limit the amount of CPU usage when unable to do so.

**Environment (please complete the following information):**
- Lens Version: 4.2.3
- OS: Windows

karlschriek:
It also appears to be spawning multiple threads:

![image](https://user-images.githubusercontent.com/25316920/121304949-4a136c00-c8fd-11eb-9145-0656ae2abeca.png)

Nokel81:
I believe that this is a duplicate of #2119

kdelisle-cs:
I'm wondering if I should make a separate issue for what I'm seeing, or if it's related since I'd imagine the fix is the same.
[...]

Issue #1952, opened 2021-01-13
MacOS Big Sur - Lens app causes extremely high host CPU usage when power mgmt kicks in and screen turns off.


mcooknu:
If Lens is open, minimised, or the "Lens Helper" is left running on the MacOS Big Sur laptop, once the screensaver kicks in and the screen turns off the laptop CPU starts to go up to a point that you hear all fans on the laptop at maximum forever until you return to the laptop and screen turns back on.  (laptop screen + monitor configuration).

**To Reproduce**
1. Install Lens
2. Open Lens 
3. Leave laptop alone and wait for screen to turn off as part of power management
4. Wait ~2mins and CPU will cause fans on laptop to go to maximum

**Expected behavior**
Lens does not generate any CPU load and certainly not max out the CPU to the point of overheating the laptop.

**Environment (please complete the following information):**
- Lens Version:
- OS: MacOS Big Sur
- Download from website

If I manually kill all "Lens" related processes, including the Lens Helper I see in the process list, then the laptop does not exhibit this maxed out CPU and fan issue.

cathy-kim:
It also happens on macOS Catalina 😢

jim-docker:
We're having trouble reproducing this, is there any more information you can provide. What cluster(s) are connected? Have you tried looking at the CPU usage with the macOS resource monitor? Is the behaviour consistent or intermittent? What are your power management settings?

mcooknu:
<img width="654" alt="Screen Shot 2021-01-25 at 8 50 35 AM" src="https://user-images.githubusercontent.com/61422545/105714478-654f8300-5eea-11eb-88f2-29617b9ef845.png">

1. Dual-montior (laptop + monitor) setup.
2. Noticed when this happens, that "com.docker.hyperkit" is at 50% CPU when I come back to my laptop and the screen turns on.  I dont see a Lens process with high CPU.
3. I usually always start Lens via command line, using "open -a Lens" after having issues AWS SSO commands to log into the account.
4. To switch between AWS Accounts, I kill Lens, re-issue new SSO for different AWS account then launch it again.
ps -ax  | grep -e "[L]ens" | awk '{print $1}' | xargs -t kill

Typically I only connect to one AWS EKS Cluster... although may move between 2-3 in the same AWS Account.

The behaviour is ~80-90% reproducable and I am now putting my laptop into Sleep mode to avoid the chance of returning to a Mac that sounds like a jet engine.  I really like Lens application so hopefully we can figure this out.

[...]

PR #3039, opened 2021-06-11
Don't try watching again if credentials have failed

### LINKED ISSUE #2119: high CPU usage in idle
**Describe the bug**
high CPU usage in idle (background)

**To Reproduce**
1. open lens
2. have 3-4 cluster config, some of them offline, all disconnected
2. go to another app like cmd

**Expected behavior**
CPU usage running low

**Screenshots**
image

- Lens Version: 4.0.8
- OS: win 10

Nokel81:

I believe that this fixes https://github.com/lensapp/lens/issues/2119 https://github.com/lensapp/lens/issues/2994 https://github.com/lensapp/lens/issues/1952

### DIFF
--- a/src/renderer/item.store.ts
+++ b/src/renderer/item.store.ts
@@ -20,7 +20,7 @@
  */
 
 import orderBy from "lodash/orderBy";
-import { autoBind, noop } from "./utils";
+import { autoBind } from "./utils";
 import { action, computed, observable, when, makeObservable } from "mobx";
 
 export interface ItemObject {
@@ -212,11 +212,6 @@ export abstract class ItemStore<T extends ItemObject = ItemObject> {
   }
 
   async removeSelectedItems?(): Promise<any>;
-
-  // eslint-disable-next-line unused-imports/no-unused-vars-ts
-  subscribe(...args: any[]) {
-    return noop;
-  }
 
   * [Symbol.iterator]() {
     yield* this.items;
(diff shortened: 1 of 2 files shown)


src/renderer/item.store.ts
```diff
@@ -213,11 +213,6 @@ export abstract class ItemStore<T extends ItemObject = ItemObject> {
 
   async removeSelectedItems?(): Promise<any>;
 
-  // eslint-disable-next-line unused-imports/no-unused-vars-ts
-  subscribe(...args: any[]) {
-    return noop;
-  }
-
```
Nokel81, line 220: We don't have a notion of subscribing these. So it is better to remove it.

src/renderer/kube-object.store.ts
```diff
@@ -345,8 +345,8 @@ export abstract class KubeObjectStore<T extends KubeObject = any> extends ItemSt
       if (!this.isLoaded || error instanceof DOMException) return;
 
       if (error instanceof Response) {
-        if (error.status === 404) {
-          // api has gone, let's not retry
+        if (error.status === 404 || error.status === 401) {
```
jakolehm, line 348: This worries me a bit because we don't recover from 401 (if credentials are refreshed after 401 happens).

jakolehm (approved):

src/renderer/kube-object.store.ts
Nokel81: This doesn't recover I agree. However, if the user moves to a different page then the credentials will be tried again. The thing is if the credentials are ENV bound then we would have to sync that  before trying again.

So we would have to sync from ENV every so often (I think) as well as retrying this. That is slightly more complicated though.

PR #3048, opened 2021-06-14
Fix manage cluster proxy settings

jakolehm:
- Click did go to "General" settings, not "Proxy" settings
- Changed button to link so that UI stays more calm (button was way too big compared to primary action, "Reconnect")

image


nevalla (approved):
LGTM

PR #3075, opened 2021-06-16
Make CatalogEntity labels clickable

### LINKED ISSUE #3061: Label Clusters from Workspaces during migration
When migrating the clusters from 4.x please add a label "workspace=xyz" to each of them.

- [x] make labels on overview and detail pane clickable
- [x] increase spacer between labels
- [x] highlight labels in white text color when hovering
- [x] make labels on detail pane clickable
- [x] on click add label text to search ("workspace=xyz")
- [x] clicking another label will result in clearing the search and adding the new label
- [ ] add label per workspace to migrated clusters

Nokel81:
- Transition workspace's to be labels

fixes #3061

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "productName": "OpenLens",
   "description": "OpenLens - Open Source IDE for Kubernetes",
   "homepage": "https://github.com/lensapp/lens",
-  "version": "5.0.0-beta.9",
+  "version": "5.0.0-beta.10",
   "main": "static/build/main.js",
   "copyright": "© 2021 OpenLens Authors",
   "license": "MIT",
--- a/src/common/cluster-store.ts
+++ b/src/common/cluster-store.ts
@@ -80,6 +80,11 @@ export interface ClusterModel {
   /** Metadata */
   metadata?: ClusterMetadata;
 
+  /**
+   * Labels for the catalog entity
+   */
+  labels?: Record<string, string>;
+
   /** List of accessible namespaces */
   accessibleNamespaces?: string[];
 }
--- a/src/main/cluster-manager.ts
+++ b/src/main/cluster-manager.ts
@@ -102,6 +102,8 @@ export class ClusterManager extends Singleton {
     const entity = catalogEntityRegistry.items[index] as KubernetesCluster;
 
     this.updateEntityStatus(entity, cluster);
+    
+    entity.metadata.labels = Object.assign({}, cluster.labels, entity.metadata.labels);
 
     if (cluster.preferences?.clusterName) {
       entity.metadata.name = cluster.preferences.clusterName;
--- a/src/migrations/helpers.ts
+++ b/src/migrations/helpers.ts
@@ -19,24 +19,37 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import type Config from "conf";
+import type Conf from "conf";
+import type { Migrations } from "conf/dist/source/types";
+import { ExtendedMap, iter } from "../common/utils";
 import { isTestEnv } from "../common/vars";
 
-export interface MigrationOpts {
-  version: string;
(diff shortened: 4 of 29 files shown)


src/migrations/helpers.ts
```diff
@@ -19,24 +19,37 @@
...
+  return Object.fromEntries(
+    iter.map(
+      migrations, 
+      ([v, fns]) => [v, (store: Conf<any>) => {
+        for (const fn of fns) {
+          fn(store);
+        }
+      }]
+    )
+  );
```
Nokel81, line 54: I needed to change this because the `snap` migration for `cluster-store` was overriding any new migration that I created. It was very annoying to debug.

package.json
```diff
@@ -3,7 +3,7 @@
   "productName": "OpenLens",
   "description": "OpenLens - Open Source IDE for Kubernetes",
   "homepage": "https://github.com/lensapp/lens",
-  "version": "5.0.0-beta.8",
+  "version": "5.0.0-beta.8.1",
```
jakolehm: Is this right?

src/common/cluster-store.ts
```diff
@@ -81,6 +81,11 @@ export interface ClusterModel {
   /** Metadata */
   metadata?: ClusterMetadata;
 
+  /**
+   * Labels for the catalog entity
+   */
+  labels?: Record<string, string>;
```
jakolehm, line 86: Why do we need to store labels?

src/main/cluster-manager.ts
```diff
@@ -95,6 +95,8 @@ export class ClusterManager extends Singleton {
     const entity = catalogEntityRegistry.items[index] as KubernetesCluster;
 
     this.updateEntityStatus(entity, cluster);
+    
+    entity.metadata.labels = Object.assign({}, cluster.labels, entity.metadata.labels);
```
jakolehm, line 106: Why not just `entity.metadata.labels.workspace = cluster.workspace`?

src/renderer/themes/lens-light.json
```diff
@@ -12,7 +12,7 @@
     "textColorPrimary": "#555555",
     "textColorSecondary": "#51575d",
     "textColorTertiary": "#555555",
-    "textColorAccent": "#333333",
+    "textColorAccent": "#222222",
```
jakolehm, line 15: Why change?

Nokel81: To match the delta on dark mode. Namely, `<Badge>`'s that are clickable now have their text change colour on hover. 

Without this change, on light mode, you can barely see the difference.

package.json
Nokel81: Yes, SemVer prerelease order means that `beta.9` > `beta.8.1` > `beta.8`.

src/common/cluster-store.ts
Nokel81: As answered elsewhere so that additional labels can be set in the future, and so that we don't need to constantly recheck the old `WorkspacesStore` file.

src/main/cluster-manager.ts
Nokel81: See above (as answered on slack), `cluster.workspace` should really have been `cluster.workspaceId`.

package.json
jakolehm: Aren't we already in `beta.9`, so this needs to be `rc.1`, right?

src/renderer/themes/lens-light.json
jakolehm: What kind of side-effects this causes (if any)?

package.json
Nokel81: We weren't it `beta.9` when I started this PR. The migration can stay where it is now though since it will still be run.

src/renderer/themes/lens-light.json
Nokel81: The text on material UI tooltips on light-mode will be darker and so will the text for `FormSwitcher`. But both of those backgrounds are quite light. So I don't think it will be noticeable.

msa0311 (approved):

PR #3079, opened 2021-06-16
Transition workspaces to hotbars

### LINKED ISSUE #3060: Migrate Workspaces to Hotbars
Migrate all existing workspaces to the hotbar and drop all clusters that do not fit to the hotbar. Users will be able to find those through the catalog.

Nokel81:

fixes #3060

### DIFF
--- a/src/migrations/hotbar-store/5.0.0-beta.11.ts
+++ b/src/migrations/hotbar-store/5.0.0-beta.11.ts
@@ -1,0 +1,85 @@
+/**
+ * Copyright (c) 2021 OpenLens Authors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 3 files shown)


src/migrations/hotbar-store/5.0.0-beta.11.ts
```diff
@@ -0,0 +1,85 @@
...
+          id: uuid.v4(), // don't use the old IDs as they aren't necessarily UUIDs
+          items: [],
+          name,
+        });
+      }
+
+      for (const cluster of clusters) {
+        const workspaceHotbar = workspaceHotbars.get(cluster.workspace);
+
+        if (workspaceHotbar && workspaceHotbars.get(cluster.workspace).items.length < defaultHotbarCells) {
```
jim-docker:
```suggestion
        if (workspaceHotbar?.items.length < defaultHotbarCells) {
```

jim-docker (commented):
How did you test this?

Nokel81:
> How did you test this?

I recreated a `lens-workspaces-store.json` file and added the IDs of the workspaces to some of my clusters within `lens-cluster-store.json`.

Making sure that the `lens-hotbar-store.json`'s migration version was such that the migration would happen.

And then I fired up a `make dev` instance, and the additional hotbars were present.

src/migrations/hotbar-store/5.0.0-beta.11.ts
```diff
@@ -0,0 +1,85 @@
...
+          workspaceHotbar.items.push({
+            entity: {
```
jim-docker, line 61: don't you need to set the kind, etc?

This is what I see:
Screen Shot 2021-06-18 at 11 48 02 AM

maybe I simulated the migration incorrectly...

Nokel81: That looks like a disabled hotbar entity because you no longer have the catalog entity associated with it.

jim-docker (approved):

PR #3082, opened 2021-06-16
Send LensMainExtension.nagivate calls to renderer before navigating

### LINKED ISSUE #3065: Can't navigate to global page from appMenu in Main.LensExtension
For an extension defined as:

```
import { Main } from "@k8slens/extensions";

export default class ExampleMainExtension extends Main.LensExtension {
  appMenus = [
    {
      parentId: "help",
      label: "Extension example",
      click: () => this.navigate("help")
    }
  ];

  async onActivate() {
    console.log('helloworld-sample activated');
  }

  onDeactivate() {
    console.log('helloworld-sample de-activated');
  }
}
```

when the menu item is clicked the following error appears in the main log:

```
[0] (node:51235) UnhandledPromiseRejectionWarning: TypeError: instance of GlobalPageRegistry is not created
[0]     at Function.getInstance (webpack-internal:///./src/common/utils/singleton.ts:61:19)
[0]     at getExtensionPageUrl (webpack-internal:///./src/extensions/registries/page-registry.ts:45:47)
[0]     at l.navigate (webpack-internal:///./src/extensions/lens-main-extension.ts:43:110)
[0]     at click (/Users/jimehrismann/git/lensapp/lens-extension-samples/helloworld-sample/dist/main.js:1:1223)
[0]     at MenuItem.click (electron/js2c/browser_init.js:65:1624)
[0]     at Function.executeCommand (electron/js2c/browser_init.js:73:1102)
[0] (node:51235) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
```

Nokel81:

fixes #3065

### DIFF
--- a/src/common/utils/iter.ts
+++ b/src/common/utils/iter.ts
@@ -140,3 +140,19 @@ export function* filterMapStrict<T, U>(src: Iterable<T>, fn: (from: T) => U | nu
     }
   }
 }
+
+/**
+ * Iterate through `src` until `match` returns a truthy value
+ * @param src A type that can be iterated over
+ * @param match A function that should return a truthy value for the item that you want to find
+ * @returns The first entry that `match` returns a truthy value for, or `undefined`
+ */
+export function find<T>(src: Iterable<T>, match: (i: T) => any): T | undefined {
+  for (const from of src) {
+    if (match(from)) {
+      return from;
+    }
+  }
+
+  return void 0;
+}
--- a/src/extensions/extension-loader.ts
+++ b/src/extensions/extension-loader.ts
@@ -355,6 +355,10 @@ export class ExtensionLoader extends Singleton {
     return this.extensions.get(extId);
   }
 
+  getInstanceById<E extends LensExtension>(extId: LensExtensionId): E {
+    return this.instances.get(extId) as E;
+  }
+
   toJSON(): Map<LensExtensionId, InstalledExtension> {
     return toJS(this.extensions);
   }
--- a/src/extensions/lens-main-extension.ts
+++ b/src/extensions/lens-main-extension.ts
@@ -21,7 +21,6 @@
 
 import { LensExtension } from "./lens-extension";
 import { WindowManager } from "../main/window-manager";
-import { getExtensionPageUrl } from "./registries/page-registry";
 import { catalogEntityRegistry } from "../main/catalog";
 import type { CatalogEntity } from "../common/catalog";
 import type { IObservableArray } from "mobx";
@@ -30,15 +29,8 @@ import type { MenuRegistration } from "./registries";
 export class LensMainExtension extends LensExtension {
   appMenus: MenuRegistration[] = [];
 
-  async navigate<P extends object>(pageId?: string, params?: P, frameId?: number) {
-    const windowManager = WindowManager.getInstance();
-    const pageUrl = getExtensionPageUrl({
-      extensionId: this.name,
-      pageId,
-      params: params ?? {}, // compile to url with params
-    });
-
-    await windowManager.navigate(pageUrl, frameId);
+  async navigate(pageId?: string, params?: Record<string, any>, frameId?: number) {
+    return WindowManager.getInstance().navigateExtension(this.id, pageId, params, frameId);
   }
 
   addCatalogSource(id: string, source: IObservableArray<CatalogEntity>) {
--- a/src/extensions/lens-renderer-extension.ts
+++ b/src/extensions/lens-renderer-extension.ts
@@ -19,33 +19,26 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import type {
-  AppPreferenceRegistration, CatalogEntityDetailRegistration, ClusterPageMenuRegistration, KubeObjectDetailRegistration, KubeObjectMenuRegistration,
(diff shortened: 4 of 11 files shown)


src/extensions/extension-loader.ts
```diff
@@ -355,6 +355,10 @@ export class ExtensionLoader extends Singleton {
     return this.extensions.get(extId);
   }
 
+  getInstanceById<E extends LensExtension>(extId: LensExtensionId): E {
+    return this.instances.get(extId) as E;
+  }
```
Nokel81, line 360: I don't know why this didn't exist before.

src/extensions/lens-main-extension.ts
```diff
@@ -30,15 +29,8 @@ import type { MenuRegistration } from "./registries";
...
-    const windowManager = WindowManager.getInstance();
-    const pageUrl = getExtensionPageUrl({
-      extensionId: this.name,
-      pageId,
-      params: params ?? {}, // compile to url with params
-    });
-
-    await windowManager.navigate(pageUrl, frameId);
+  async navigate(pageId?: string, params?: Record<string, any>, frameId?: number) {
+    return WindowManager.getInstance().navigateExtension(this.id, pageId, params, frameId);
```
Nokel81, line 33: `getExtensionPageUrl` really shouldn't have ever been accessible from `main` since neither `globalPageMenuRegistry` nor `pageMenuRegistry` would have been populated on `main` ever. 

So while this bug has only been recently noticed (because during #2779 I consciously didn't create instances on `main` for registries that aren't proper to `main`) this functionality has always been broken.

src/extensions/lens-renderer-extension.ts
```diff
@@ -19,33 +19,26 @@
...
 import { LensExtension } from "./lens-extension";
 import { getExtensionPageUrl } from "./registries/page-registry";
-import type { CommandRegistration } from "./registries/command-registry";
-import type { EntitySettingRegistration } from "./registries/entity-setting-registry";
-import type { TopBarRegistration } from "./registries/topbar-registry";
 
 export class LensRendererExtension extends LensExtension {
-  globalPages: PageRegistration[] = [];
-  clusterPages: PageRegistration[] = [];
-  globalPageMenus: PageMenuRegistration[] = [];
```
Nokel81, line 36: I removed this because they don't exist anymore

src/extensions/lens-renderer-extension.ts
```diff
@@ -19,33 +19,26 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import type {
-  AppPreferenceRegistration, CatalogEntityDetailRegistration, ClusterPageMenuRegistration, KubeObjectDetailRegistration, KubeObjectMenuRegistration,
-  KubeObjectStatusRegistration, PageMenuRegistration, PageRegistration, StatusBarRegistration, WelcomeMenuRegistration, WorkloadsOverviewDetailRegistration,
-} from "./registries";
+import type * as registries from "./registries";
```
Nokel81, line 22: I find this as easier to read, and makes adding future registries a smaller diff.

src/extensions/registries/page-menu-registry.ts
```diff
@@ -26,23 +26,20 @@ import type { PageTarget, RegisteredPage } from "./page-registry";
...
 }
 
-export interface PageMenuComponents {
+export interface ClusterPageMenuComponents {
   Icon: React.ComponentType<IconProps>;
 }
 
-export class PageMenuRegistry<T extends PageMenuRegistration> extends BaseRegistry<T> {
-  add(items: T[], ext: LensExtension) {
+export class ClusterPageMenuRegistry extends BaseRegistry<ClusterPageMenuRegistration> {
```
Nokel81, line 41: I consolidated this here because we no longer have `(Global)PageMenuRegistry` so we shouldn't leave this clutter around.

src/main/window-manager.ts
```diff
@@ -175,18 +181,30 @@ export class WindowManager extends Singleton {
     return this.mainWindow;
   }
 
-  sendToView({ channel, frameInfo, data = [] }: { channel: string, frameInfo?: ClusterFrameInfo, data?: any[] }) {
+  private sendToView({ channel, frameInfo, data = [] }: SendToViewArgs) {
```
Nokel81, line 184: This isn't used outside of this class so it really should be private, especially since the other helper functions cover all the cases.

jakolehm (approved):

src/common/utils/iter.ts
```diff
@@ -140,3 +140,19 @@ export function* filterMapStrict<T, U>(src: Iterable<T>, fn: (from: T) => U | nu
...
+ * @returns The first entry that `match` returns a truthy value for, or `undefined`
+ */
+export function find<T>(src: Iterable<T>, match: (i: T) => any): T | undefined {
+  for (const from of src) {
+    if (match(from)) {
+      return from;
+    }
+  }
+
+  return void 0;
```
nevalla, line 157: what is `void 0`? how is that different than `undefined`?

nevalla (approved):

src/common/utils/iter.ts
Nokel81: The `void` operator is a unary right-associative operator that "executes" the adjacent expression and returns the true undefined value.

Given that we are sensible, this shouldn't be different that `undefined`. But a JS querk is `undefined` is actually a global variable. Namely `var undefined = void 0;`, is it is possible for it to be reassigned.

PR #3083, opened 2021-06-17
Set correct namespace for the subject when creating new role binding

nevalla:
Currently when creating new role binding, selected namespace is set as service account subject's namespace. This PR will fix this and sets the namespace where service account is created.

### DIFF
--- a/src/renderer/components/+user-management/+role-bindings/dialog.tsx
+++ b/src/renderer/components/+user-management/+role-bindings/dialog.tsx
@@ -91,17 +91,15 @@ export class RoleBindingDialog extends React.Component<Props> {
     const serviceAccounts = Array.from(this.selectedAccounts, sa => ({
       name: sa.getName(),
       kind: "ServiceAccount" as const,
-      namespace: this.bindingNamespace,
+      namespace: sa.getNs(),
     }));
     const users = Array.from(this.selectedUsers, user => ({
       name: user,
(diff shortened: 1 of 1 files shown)


src/renderer/components/+user-management/+role-bindings/dialog.tsx
```diff
@@ -91,17 +91,15 @@ export class RoleBindingDialog extends React.Component<Props> {
     const serviceAccounts = Array.from(this.selectedAccounts, sa => ({
       name: sa.getName(),
       kind: "ServiceAccount" as const,
-      namespace: this.bindingNamespace,
+      namespace: sa.getNs(),
     }));
     const users = Array.from(this.selectedUsers, user => ({
       name: user,
-      kind: "User" as const,
-      namespace: this.bindingNamespace,
```
Nokel81, line 99: This is not what we want? Or am I missunderstanding what the namespace field in a subject binding is for?

nevalla: namespace field defines the namespace of the service account resource. There can be multiple service accounts with the same name in different namespaces.

Nokel81: Ah so it is only for service accounts. Thanks.

Nokel81 (approved):

PR #3086, opened 2021-06-17
Check are resources allowed before subscribing stores

nevalla:
Cluster dashboard start watching pods, events, nodes and namespace apis and that is causing issues if user does not have access to those resources. This PR will check are those resources allowed before subscribing related stores.

### DIFF
--- a/src/renderer/components/app.tsx
+++ b/src/renderer/components/app.tsx
@@ -107,8 +107,22 @@ export class App extends React.Component {
   }
 
   componentDidMount() {
+    const stores: Array<KubeObjectStore> = [namespaceStore];
+
+    if (isAllowedResource("nodes")) {
+      stores.push(nodesStore);
+    }
+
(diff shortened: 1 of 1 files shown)


src/renderer/components/app.tsx
```diff
@@ -107,8 +107,22 @@ export class App extends React.Component {
   }
 
   componentDidMount() {
+    const stores: Array<KubeObjectStore> = [namespaceStore];
```
nevalla, line 110: `namespaceStore` has logic not to start watching if there are no accessible namespaces available, but should we still have check here, if user has not configured any namespaces in cluster settings?

jakolehm: I think we need to fix this within `KubeObjectStore`, not in each react component.

nevalla:
Closed in favour of #3089

PR #3091, opened 2021-06-17
lens:// routing shouldn't ignore intree extensions

### LINKED ISSUE #3090: `lens://` routing ignores in-tree extensions
**Describe the bug**
If an intree extension registers a `lens://` protocol route. And a user tries to go to that route, a notification will be displayed that implies that no extension was found.

Nokel81:

fixes #3090

### DIFF
--- a/src/common/protocol-handler/router.ts
+++ b/src/common/protocol-handler/router.ts
@@ -179,7 +179,7 @@ export abstract class LensProtocolRouter extends Singleton {
     const { [EXTENSION_PUBLISHER_MATCH]: publisher, [EXTENSION_NAME_MATCH]: partialName } = match.params;
     const name = [publisher, partialName].filter(Boolean).join("/");
 
-    const extension = ExtensionLoader.getInstance().userExtensionsByName.get(name);
+    const extension = ExtensionLoader.getInstance().getInstanceByName(name);
 
     if (!extension) {
       logger.info(`${LensProtocolRouter.LoggingPrefix}: Extension ${name} matched, but not installed`);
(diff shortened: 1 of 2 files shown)


src/common/protocol-handler/router.ts
```diff
@@ -179,7 +179,7 @@ export abstract class LensProtocolRouter extends Singleton {
     const { [EXTENSION_PUBLISHER_MATCH]: publisher, [EXTENSION_NAME_MATCH]: partialName } = match.params;
     const name = [publisher, partialName].filter(Boolean).join("/");
 
-    const extension = ExtensionLoader.getInstance().userExtensionsByName.get(name);
+    const extension = ExtensionLoader.getInstance().getInstanceByName(name);
```
jim-docker, line 182: Couldn't you have just used the existing `getExtensionByName()`?

Nokel81: Do you mean instead of changing its name to `getInstanceByName`?

Nokel81: I changed the name because we have a difference between `InstalledExtension`'s (which use the name `extension` in the loader's method names and `LensExtension`'s (or instances) which tend to use the work `instance` in the loader's method names.

I updated this function name to be inline with that.

src/extensions/extension-loader.ts
```diff
@@ -81,28 +101,8 @@ export class ExtensionLoader extends Singleton {
...
-        continue;
-      }
-
-      extensions.set(val.manifest.name, val);
-    }
-
-    return extensions;
-  }
-
-  getExtensionByName(name: string): LensExtension | null {
```
jim-docker, line 98: This function as is doesn't do what you need? Did you have to add the `instancesByName` observable.map?

Nokel81: No, it was an optimization. Similar to those done in `CatalogEntityRegistry` in `renderer`. To avoid rebuilding the whole map every time a dependent is changed.

jim-docker (approved):

PR #3106, opened 2021-06-18
General catalog category

### LINKED ISSUE #3052: New Catalog Category "General"
Create a new catalog category called general with 2 entities:
- Catalog 
- Preferences

The first entity (catalog) should be by default on the index 0 of the hotbar.
Remove Catalog symbol from Status Bar.

**Notes from discussions:**
- add general section with catalog / preferences
- catalog should show the last thing you did on catalog preferrable (if not possible now, just show browse)
- catalog item is by default added to the first index
- catalog item is not on the other hotbars
- catalog item should be view-list material icon ui
- preferences should show system preferences on item click
- (category general list should only show description)

aleksfront:
* Adding `General` section into catalog categories
* Adding `Catalog` and `Preferences` general items which open associated pages
* Removing `Catalog` link from the bottom bar
* Setting `Catalog` item initially after creating first hotbar.

general category
general items in list

**Note**
To show `Catalog` entity for users migrated from 4.x, this item should be injected to migrations, probably after merge #3079.

Resolves #3052


src/common/catalog-entities/general.ts
```diff
@@ -0,0 +1,72 @@
...
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import { navigate } from "../../renderer/navigation";
+import { CatalogCategory, CatalogEntity, CatalogEntityMetadata, CatalogEntitySpec, CatalogEntityStatus } from "../catalog";
+import { catalogCategoryRegistry } from "../catalog/catalog-category-registry";
+
+type GeneralEntitySpec = {
+  path: string;
+} & CatalogEntitySpec;
```
Nokel81: I find this more readable, and is the style that we use elsewhere.

```suggestion
interface GeneralEntitySpec extends CatalogEntitySpec {
  path: string;
}
```

src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -42,12 +43,11 @@ export type KubernetesClusterPrometheusMetrics = {
 export type KubernetesClusterSpec = {
```
Nokel81: Same here

src/common/catalog/catalog-entity.ts
```diff
@@ -147,7 +147,13 @@ export interface CatalogEntityAddMenuContext {
...
 }
 
-export type CatalogEntitySpec = Record<string, any>;
+export type CatalogEntitySpec = {
+  icon?: {
+    src?: string;
+    material?: string;
+    background?: string;
+  }
+} & Record<string, any>;
```
Nokel81:
```suggestion
export interface CatalogEntitySpec {
  icon?: {
    src?: string;
    material?: string;
    background?: string;
  };
  [key: string]: any;
};
```

src/common/catalog/catalog-entity.ts
```diff
@@ -147,7 +147,14 @@ export interface CatalogEntityAddMenuContext {
   menuItems: CatalogEntityAddMenu[];
 }
 
-export type CatalogEntitySpec = Record<string, any>;
+export interface CatalogEntitySpec {
+  icon?: {
```
jakolehm: Why did we decide to add this to all entities? It also feels like we are leaking `HotbarIconProps` to entities...

src/main/index.ts
```diff
@@ -53,6 +53,7 @@ import { handleWsUpgrade } from "./proxy/ws-upgrade";
 import configurePackages from "../common/configure-packages";
 import { PrometheusProviderRegistry } from "./prometheus";
 import * as initializers from "./initializers";
+import { addGeneralEntities } from "./catalog-sources/general";
```
jakolehm: Should be imported via `initializers`.

src/common/catalog/catalog-entity.ts
aleksfront: Yes, it's definitely a case of [feature envy](https://refactoring.guru/smells/feature-envy). Removed icon reference from `CatalogEntitySpec`.

msa0311 (approved):

Nachasic (approved):

PR #3139, opened 2021-06-22
Use beta release channel for rc builds

jakolehm:


msa0311 (approved):

Nokel81:
This won't work because when we change to RC we reset the trailing numbers.

PR #3145, opened 2021-06-22
Transparent background for custom icons

### LINKED ISSUE #3135: Custom cluster icon has weird background color
image

aleksfront:
This PR fixes background color for `HotBarIcon` component. Icons with custom uploaded image will have transparent background from now on.

transparent background

Fixes #3135 

### DIFF
--- a/src/renderer/components/avatar/avatar.tsx
+++ b/src/renderer/components/avatar/avatar.tsx
@@ -72,9 +72,21 @@ function getIconString(title: string) {
 export function Avatar(props: Props) {
   const { title, width = 32, height = 32, colorHash, children, background, ...settings } = props;
 
+  const getBackgroundColor = () => {
+    if (background) {
+      return background;
+    }
+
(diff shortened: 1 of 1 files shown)


src/renderer/components/avatar/avatar.tsx
```diff
@@ -72,9 +72,21 @@ function getIconString(title: string) {
 export function Avatar(props: Props) {
   const { title, width = 32, height = 32, colorHash, children, background, ...settings } = props;
 
+  const getBackgroundColor = () => {
```
Nokel81, line 75: Why not just make this an immediately called function since the result never changes?

Nokel81 (approved):

src/renderer/components/avatar/avatar.tsx
aleksfront: Good point! Will change that in `HotbarIcon` -> `EntityIcon` refactoring.

PR #3147, opened 2021-06-22
Don't recreate the whole userExtensions Map on every change

Nokel81:

This is a similar solution to what was done with the instances by name mapping.

### DIFF
--- a/src/extensions/__tests__/extension-loader.test.ts
+++ b/src/extensions/__tests__/extension-loader.test.ts
@@ -132,18 +132,18 @@ describe("ExtensionLoader", () => {
     ExtensionLoader.resetInstance();
   });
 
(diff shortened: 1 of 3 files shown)


nevalla:
When trying to enable / disable an extension I get this error:
image

Nokel81:
@nevalla Thanks for that report, I forgot to test that flow.

nevalla:
There seems to be still some issues on disabling/enabling extensions:

https://user-images.githubusercontent.com/455844/129727467-b0cdb10f-48eb-4992-8890-6b095abd9543.mov

Nokel81:
After some more testing I don't think I can get this to work. Might come back to this later if need be.

PR #3066, opened 2021-06-15
Bring back support for custom cluster icons

jakolehm:
Fixes #3057

image
image

PR #3152, opened 2021-06-22
Use new EntityIcon for non-hotbar uses

### LINKED ISSUE #3126: Hot bar context menu not opening
How to reproduce:
1. Pin a new icon to the hot bar (can be e.g. web link or Space)
2. Right click on the icon in the hot bar to open the context menu

What should happen:
- Context menu opens

What actually happens:
- Context menu is not opened

Refreshing with CMD+R seems to fix the context menu. The menu works fine for older icons.

### LINKED ISSUE #3133: Catalog Icons have a `:hover` rule.
**Describe the bug**
Catalog icons currently have a hover rule, the same as in the hotbar. But they shouldn't.

https://user-images.githubusercontent.com/8225332/122774876-96ed3000-d277-11eb-8af1-7900676b5c72.mov

Nokel81:

fixes #3126 By not using the same component within the catalog. There are conflicts with multiple IDs that way. A better fix (for post 5.0) would be to have `<Menu>` be a child of its target always. That would mean that we don't need to use IDs at all.

fixes https://github.com/lensapp/lens/issues/3133 By making the icon hover boarder narrower (so it looks better) and by having the click action do the run command. I think it would be better if right clicking opened up the context menu. But I though that would be out of scope.

https://user-images.githubusercontent.com/8225332/122964470-4cda7c00-d355-11eb-97ad-811087f5c853.mov

### DIFF
--- a/src/renderer/components/+catalog/catalog.tsx
+++ b/src/renderer/components/+catalog/catalog.tsx
@@ -28,7 +28,7 @@ import { action, makeObservable, observable, reaction, runInAction, when } from 
 import { CatalogEntityItem, CatalogEntityStore } from "./catalog-entity.store";
 import { navigate } from "../../navigation";
 import { MenuItem, MenuActions } from "../menu";
-import type { CatalogEntityContextMenu, CatalogEntityContextMenuContext } from "../../api/catalog-entity";
+import { CatalogEntityContextMenu, CatalogEntityContextMenuContext, catalogEntityRunContext } from "../../api/catalog-entity";
 import { HotbarStore } from "../../../common/hotbar-store";
 import { ConfirmDialog } from "../confirm-dialog";
 import { catalogCategoryRegistry, CatalogEntity } from "../../../common/catalog";
@@ -41,7 +41,7 @@ import { makeCss } from "../../../common/utils/makeCss";
 import { CatalogEntityDetails } from "./catalog-entity-details";
 import { catalogURL, CatalogViewRouteParam } from "../../../common/routes";
 import { CatalogMenu } from "./catalog-menu";
-import { HotbarIcon } from "../hotbar/hotbar-icon";
+import { EntityIcon } from "../entity-icon";
 
 export const previousActiveTab = createAppStorage("catalog-previous-active-tab", "");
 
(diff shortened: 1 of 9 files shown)


src/renderer/components/+catalog/catalog.tsx
```diff
@@ -165,14 +165,14 @@ export class Catalog extends React.Component<Props> {
...
-        uid={item.getId()}
+      <EntityIcon
         title={item.getName()}
         source={item.source}
         src={item.entity.spec.icon?.src}
         material={item.entity.spec.icon?.material}
         background={item.entity.spec.icon?.background}
-        onClick={() => this.onDetails(item)}
+        onClick={() => item.onRun(catalogEntityRunContext)}
+        hoverWidth="1.5px"
```
jakolehm, line 182: This does not feel right.. why do we need this?

aleksfront:
> By making the icon hover boarder narrower (so it looks better) and by having the click action do the run command. I think it would be better if right clicking opened up the context menu. But I though that would be out of scope.

I disagree it looks better. I think we shouldn't change icon sizes. Smaller breaks whole "grid" and points too much eye attentions on it. Plus this differentiation in size might mean something, there must be a reason.

Nokel81:
> I disagree it looks better. I think we shouldn't change icon sizes. Smaller breaks whole "grid" and points too much eye attentions on it. Plus this differentiation in size might mean something, there must be a reason.

I didn't change the icon size at all.

src/renderer/components/+catalog/catalog.tsx
Nokel81: This fixes https://github.com/lensapp/lens/issues/3133 by the border width be appropriate to the size of the icon.

aleksfront:
Well, at least they seems to be smaller regarding to your recording.
Screenshot 2021-06-23 at 16 14 38

Icons with letter and icons with custom images should have same size and look same (by their proportions)

Nokel81:
> Well, at least they seems to be smaller regarding to your recording.

Smaller with regard to what? On master they are just as small. Ever since #3066

jakolehm:
> > Well, at least they seems to be smaller regarding to your recording.
> 
> Smaller with regard to what? On master they are just as small. Ever since #3066

I think that screenshot icon has padding.. at least for me custom icons fill up the space in master branch.

jakolehm:
> > > Well, at least they seems to be smaller regarding to your recording.
> > 
> > 
> > Smaller with regard to what? On master they are just as small. Ever since #3066
> 
> I think that screenshot icon has padding.. at least for me custom icons fill up the space in master branch.

I was wrong, they actually look smaller .. that's a bug.

Nachasic (dismissed):

aleksfront:
`EntityIcon` refactoring will be made in another PR taking part of this changes into account.

PR #3153, opened 2021-06-23
Show led only for clusters

jakolehm:
Led on hotbar icon does not really make sense to any other entities. We need to provide proper API for enabling this in future versions.


aleksfront (approved):
Make sense

msa0311 (approved):

PR #3178, opened 2021-06-24
Fix hotbar migration of workspaces

### LINKED ISSUE #3177: Migrating workspaces to hotbars does not work
**Describe the bug**
When upgrading from 4.x to 5, current workspaces should be migrated to hotbars. However, it's not working properly at this moment. It looks like migration will create hot bars two times. First hotbars do not contain any clusters.

Also cluster icons are not migrated at all.

**To Reproduce**
1. Install Lens 4.x
2. Add workspaces and clusters
3. Install Lens 5 beta
4. See that workspaces are not migrated to hot bars.

Nokel81:
- Don't use the cluster.id field as it was previously a uuidV4, and
  newer clusterIds are md5's of path:name


src/migrations/hotbar-store/5.0.0-alpha.0.ts
```diff
@@ -21,33 +21,18 @@
...
-        hotbars.push(hotbar);
-      }
-
-      hotbar.items.push({
-        entity: { uid: cluster.id },
-        params: {}
-      });
-    });
-
-    store.set("hotbars", hotbars);
```
Nokel81: This migration just doesn't make any sense... so I replaced it with one that makes sure that there is an initial hotbar.

src/renderer/components/+catalog/catalog-entity-details.tsx
```diff
@@ -91,6 +92,11 @@ export class CatalogEntityDetails<T extends CatalogEntity> extends Component<Pro
               <DrawerItem name="Labels">
                 {...item.getLabelBadges(this.props.hideDetails)}
               </DrawerItem>
+              {isDevelopment && (
+                <DrawerItem name="Id">
+                  {item.getId()}
+                </DrawerItem>
+              )}
```
Nokel81, line 99: This is useful for developers to know what the entity IDs are.

src/migrations/hotbar-store/5.0.0-beta.10.ts
```diff
@@ -49,7 +50,7 @@ export default {
         workspaceHotbars.set(id, {
           id: uuid.v4(), // don't use the old IDs as they aren't necessarily UUIDs
           items: [],
-          name,
+          name: `Workspace: ${name}`,
```
Nokel81, line 55: This prevents there being two hotbars with the name "default"

src/common/base-store.ts
```diff
@@ -48,16 +46,27 @@ export abstract class BaseStore<T = any> extends Singleton {
...
-  protected constructor(protected params: BaseStoreParams) {
+  protected constructor(protected params: BaseStoreParams<T>) {
     super();
     makeObservable(this);
+  }
+
+  /**
+   * This must be called after the last child's constructor is finished (or just before it finishes)
+   */
+  load() {
```
Nokel81, line 51: Given that this is now synchronous I have made it so that all the in-repo stores initialize in the constructor.

This several benefits:
1. It is inline with how the `conf` package works so we have less dissonance.
2. We can get rid of `BaseStore.isLoaded`
3. No need to `await` the loading so config loading is now always done in a static order.

It is fine to use synchronous code here because loading only happens once.

src/common/cluster-store.ts
```diff
@@ -137,31 +139,27 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
...
   }
 
-  async load() {
-    const initialStates = "cluster:states";
-
-    await super.load();
-
-    if (ipcRenderer) {
-      logger.info("[CLUSTER-STORE] requesting initial state sync");
+  async loadInitialOnRenderer() {
```
Nokel81, line 146: This had to be moved to its own function since `load()` is no longer `async`. I still believe that this is a better design.

src/common/hotbar-store.ts
```diff
@@ -92,30 +92,13 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
...
-    const initialItem = { entity: { uid, name, source }};
-
-    return [
-      initialItem,
-      ...Array.from(Array(defaultHotbarCells - 1).fill(null))
-    ];
-  }
-
-  get initialItems() {
+  static getInitialItems() {
```
Nokel81, line 91: This is static so that migrations can reuse this code.

src/common/hotbar-store.ts
```diff
@@ -92,30 +92,13 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
...
   }
 
-  @action protected async fromStore(data: Partial<HotbarStoreModel> = {}) {
-    if (data.hotbars?.length === 0) {
-      this.hotbars = [{
-        id: uuid.v4(),
-        name: "Default",
-        items: this.defaultHotbarInitialItems,
-      }];
-    } else {
```
Nokel81, line 116: This was moved to the `alpha.0` migration so that it is always added. This code doesn't work when moving from pre-alpha.0 because the migration of `beta.10` adds at least one hotbar.

src/extensions/extension-store.ts
```diff
@@ -26,13 +26,13 @@ import type { LensExtension } from "./lens-extension";
 export abstract class ExtensionStore<T> extends BaseStore<T> {
   protected extension: LensExtension;
 
-  async loadExtension(extension: LensExtension) {
+  loadExtension(extension: LensExtension) {
```
Nokel81, line 29: Though this is technically a breaking change (which I can easily revert) I think that it is better to leave it as synchronous so that the extension developers can gain the same benefits as ourselves.

src/main/catalog-sources/general.ts
```diff
@@ -67,6 +67,6 @@ const generalEntities = observable([
   preferencesEntity
 ]);
 
-export function initializeGeneralEntities() {
+export function syncGeneralEntities() {
```
Nokel81, line 70: We are not initializing anything here so I changed this function name (and the one for weblinks) and removed the exports from `./initializers/index.ts`

src/main/index.ts
```diff
@@ -122,9 +128,18 @@ app.on("ready", async () => {
...
-  await initializers.initializeStores();
-  initializers.initializeWeblinks();
+  logger.info("💾 Loading stores");
+
+  UserStore.createInstance();
+  ClusterStore.createInstance().provideInitialFromMain();
+  HotbarStore.createInstance();
+  ExtensionsStore.createInstance();
+  FilesystemProvisionerStore.createInstance();
+  WeblinkStore.createInstance();
```
Nokel81: Given that we are no longer awaiting anything the benefit of having a separate function is diminished (I think). But I can revert this to its own function if desirable.

src/migrations/hotbar-store/5.0.0-alpha.0.ts
```diff
@@ -20,34 +20,24 @@
...
 // Cleans up a store that had the state related data stored
-import type { Hotbar } from "../../common/hotbar-store";
-import { ClusterStore } from "../../common/cluster-store";
+import { Hotbar, HotbarStore } from "../../common/hotbar-store";
 import * as uuid from "uuid";
 import type { MigrationDeclaration } from "../helpers";
+import { catalogEntity } from "../../main/catalog-sources/general";
 
 export default {
   version: "5.0.0-alpha.0",
```
Nokel81, line 29: I believe that this is the best place to add this hotbar entity even though this is slightly anachronistic since the general category didn't exist in `alpha.0`.

That is because for all non-beta users this is guaranteed to run and simplifies the logic since before `alpha.0` we know that there was not a hotbar in existence.

src/migrations/hotbar-store/5.0.0-beta.10.ts
```diff
@@ -59,7 +60,7 @@ export default {
         if (workspaceHotbar?.items.length < defaultHotbarCells) {
           workspaceHotbar.items.push({
             entity: {
-              uid: cluster.id,
+              uid: createHash("md5").update(`${cluster.kubeConfigPath}:${cluster.contextName}`).digest("hex"),
```
Nokel81, line 77: This is needed because `cluster.id` was previously a black box value that could not be recreated (ie uuidV4 in most cases).

Nokel81: This is the line that fixes the hotbar items being disabled.

src/migrations/hotbar-store/5.0.0-beta.5.ts
```diff
@@ -20,7 +20,7 @@
  */
 
 import type { Hotbar } from "../../common/hotbar-store";
-import { catalogEntityRegistry } from "../../renderer/api/catalog-entity-registry";
+import { catalogEntityRegistry } from "../../main/catalog";
```
Nokel81, line 23: This was just wrong since the `renderer` migration would never fire because the main code would have already done the migration before hand. And thus this migration was removing all items from all hotbars.

src/migrations/hotbar-store/5.0.0-alpha.0.ts
```diff
@@ -20,34 +20,24 @@
...
-      const name = cluster.workspace;
+    const { metadata: { uid, name, source } } = catalogEntity;
 
-      if (!name) return;
+    hotbar.items[0] = { entity: { uid, name, source } };
 
-      let hotbar = hotbars.find((h) => h.name === name);
-
-      if (!hotbar) {
-        hotbar = { id: uuid.v4(), name, items: [] };
```
Nokel81, line 41: Furthermore, I don't know when this code was added but this is wrong since `name` is misleading: `cluster.workspace` is not the workspace name, but in fact its a workspace ID.

src/migrations/user-store/file-name-migration.ts
```diff
@@ -23,18 +23,18 @@ import fse from "fs-extra";
 import { app, remote } from "electron";
 import path from "path";
 
-export async function fileNameMigration() {
+export function fileNameMigration() {
```
Nokel81, line 26: Making this synchronous allows this to be done within the `UserStore`'s `constructor`.

src/renderer/theme.store.ts
```diff
@@ -20,9 +20,11 @@
  */
 
 import { computed, observable, reaction, makeObservable } from "mobx";
-import { autoBind, boundMethod, Singleton } from "./utils";
+import { autoBind, iter, Singleton } from "./utils";
 import { UserStore } from "../common/user-store";
 import logger from "../main/logger";
+import darkThemeBase from "./themes/lens-dark.json";
+import lightThemeBase from "./themes/lens-light.json";
```
Nokel81: Given that these are the only two themes we have, it seems better practice to just import them statically.

src/renderer/theme.store.ts
```diff
@@ -32,42 +34,46 @@ export enum ThemeType {
...
-  id: ThemeId; // filename without .json-extension
   type: ThemeType;
-  name?: string;
-  colors?: Record<string, string>;
-  description?: string;
-  author?: string;
+  name: string;
+  colors: Record<string, string>;
+  description: string;
+  author: string;
```
Nokel81, line 41: We can now make these non-optional because we are no longer post-loading the files.

src/renderer/theme.store.ts
```diff
@@ -32,42 +34,46 @@ export enum ThemeType {
...
+
+const darkTheme: Theme = {
+  ...darkThemeBase,
+  type: darkThemeBase.type as ThemeType,
+};
+
+const lightTheme: Theme = {
+  ...lightThemeBase,
+  type: lightThemeBase.type as ThemeType,
+};
```
Nokel81: This is more verbose than it might otherwise seem necessary because `type` is TS typed as `string` instead of `ThemeType`. 

I could have done `lightThemeBase as Theme` but that wasn't throwing errors if I removed the `author` field from the JSON file. So this mechanism is more resilient to changes to the `interface Theme` type.

src/renderer/theme.store.ts
```diff
@@ -32,42 +34,46 @@ export enum ThemeType {
...
   }
 
   @computed get activeTheme(): Theme {
-    const activeTheme = this.allThemes.get(this.activeThemeId) ?? this.allThemes.get("lens-dark");
-
-    return {
-      colors: {},
-      ...activeTheme,
-    };
+    return this.allThemes.get(this.activeThemeId) ?? this.allThemes.get("lens-dark");
```
Nokel81, line 66: No need to make sure that `colors` is defined since it is now a required field.

src/renderer/theme.store.ts
```diff
@@ -77,9 +83,9 @@ export class ThemeStore extends Singleton {
     autoBind(this);
 
     // auto-apply active theme
-    reaction(() => this.activeThemeId, async themeId => {
+    reaction(() => this.activeThemeId, themeId => {
       try {
-        this.applyTheme(await this.loadTheme(themeId));
+        this.applyTheme(this.getThemeById(themeId));
```
Nokel81, line 78: This no longer needs to be `async` because we have a static list. Furthermore, in the future when we support custom themes we can keep this synchronous because the loading of theme files will be done before the call the `applyTheme` is made.

src/renderer/theme.store.ts
```diff
@@ -89,38 +95,10 @@ export class ThemeStore extends Singleton {
...
-        existingTheme.colors = theme.colors;
-        existingTheme.description = theme.description;
-        existingTheme.name = theme.name;
-      }
-
-      return existingTheme;
-    } catch (err) {
-      throw new Error(`Can't load theme "${themeId}": ${err}`);
-    }
-  }
```
Nokel81, line 122: This function is made completely redundant by the code above.

src/common/__tests__/hotbar-store.test.ts
```diff
@@ -120,63 +127,61 @@ jest.mock("electron", () => {
 
 describe("HotbarStore", () => {
   beforeEach(() => {
-    ClusterStore.resetInstance();
+    mockFs({
+      "tmp": {
+        "lens-hotbar-store.json": JSON.stringify({})
+      }
+    });
```
Nokel81, line 134: This needs to be first so that the stores use the mocked file system.

src/common/user-store.ts
```diff
@@ -136,6 +112,24 @@ export class UserStore extends BaseStore<UserStoreModel> {
     return this.shell || process.env.SHELL || process.env.PTYSHELL;
   }
 
+  startMainReactions() {
```
Nokel81, line 115: Having this code not be part of either `load` or `constructor` means that we can ignore uncancellable reactions in unit tests.

src/migrations/hotbar-store/5.0.0-beta.10.ts
```diff
@@ -59,7 +62,7 @@ export default {
         if (workspaceHotbar?.items.length < defaultHotbarCells) {
```
aleksfront, line 74: Probably we should use 
```
if (workspaceHotbar?.items.length < defaultHotbarCells - 1) {
```
here since first item must be already occupied by Catalog entity.

src/renderer/theme.store.ts
```diff
@@ -32,42 +34,46 @@ export enum ThemeType {
 }
 
 export interface Theme {
-  id: ThemeId; // filename without .json-extension
   type: ThemeType;
```
aleksfront, line 37: We can use literal types here `type: "dark" | "light"` and set them explicitly below. To avoid verbose things like
```
const darkTheme: Theme = {
  ...darkThemeBase,
  type: darkThemeBase.type as ThemeType,
};
```

aleksfront (dismissed):

src/renderer/theme.store.ts
Nokel81: Okay sounds good

src/migrations/hotbar-store/5.0.0-beta.10.ts
Nokel81: These are all new hotbars so that fact doesn't apply.

aleksfront (approved):
Looking great!

PR #3179, opened 2021-06-24
Notify user of error if confirmation dialog rejects promise

Nokel81:

This will help with extensions and error reporting as a good default for when extension developers use the confirmation dialog. Of course they can catch the error themselves and report a customized notification.


nevalla:
Not sure should we display this to end users, or should we just output error to console?

Nokel81:
I was thinking that displaying and error n notification would be a better UX because then the user knows that the action failed. Of course an extension developer could catch the error himself and output a better error message. This is just supposed to be a decent default.

aleksfront (approved):
LGTM

nevalla (approved):

PR #3190, opened 2021-06-25
HotbarIcon's refering to unknown entities should be removalable

Nokel81:

fixes regression from #2923 

This PR creates a new type of hotbar icon prop called `inactive` (because the hotbar icon is used in several different places with different needs). 

When `inactive` is true the icon looks like it is disabled but still allows menu actions and run actions to be done.

### DIFF
--- a/src/renderer/components/hotbar/hotbar-icon.scss
+++ b/src/renderer/components/hotbar/hotbar-icon.scss
@@ -49,6 +49,10 @@
     cursor: default;
     filter: grayscale(0.7);
 
+    &.contextMenuAvailable {
+      cursor: context-menu;
+    }
+
     &:hover {
       &:not(.active) {
         box-shadow: none;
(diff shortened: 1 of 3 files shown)


src/renderer/components/hotbar/hotbar-icon.scss
```diff
@@ -56,6 +55,14 @@
     }
   }
 
+  &.disabled {
+    cursor: default;
+  }
+
+  &.inactive {
+    cursor: context-menu;
+  }
```
Nokel81: This is also new, I found that because it was disabled I didn't realize (initially) that there was supposed to be a context menu.

aleksfront:
Is there a reason why we hiding context menu for `disabled` hotbar icons? Since we're passing `disabled` along with context menu items inside `hotbar-menu.tsx`, I suggest just to remove this check 
https://github.com/lensapp/lens/blob/master/src/renderer/components/hotbar/hotbar-icon.tsx#L113.

Creating `inactive` along with `disabled` can lead to confusion. You can't say what's the difference just by looking at it.

msa0311 (approved):

aleksfront (approved):
LGTM

PR #3194, opened 2021-06-28
Fixing tooltips to use single component

### LINKED ISSUE #3095: Inconsistent Tooltips
In different places we're using "native" tooltip version along with new one which using `MaterialTooltip` under the hood. We should replace all tooltips with a consistent new look.

tooltips

### LINKED ISSUE #3133: Catalog Icons have a `:hover` rule.

aleksfront:
* Fine-tuning `<Tooltip/>` styles along with bugfixes (add delay period before showing tooltip)
* Removing reference to `<MaterialTooltip/>` and component itself.
* Removing hover effect in Catalog entity icons

https://user-images.githubusercontent.com/9607060/123600542-75ec8800-d7ff-11eb-92a3-d75b342dd7e3.mov

Fixes #3095 
Fixes #3133


src/renderer/components/+catalog/catalog.tsx
```diff
@@ -180,6 +180,7 @@ export class Catalog extends React.Component<Props> {
         material={item.entity.spec.icon?.material}
         background={item.entity.spec.icon?.background}
         onClick={() => this.onDetails(item)}
+        interactive={false}
```
Nokel81: This seems strange, given that we also have an `onClick` defined.

src/renderer/components/hotbar/hotbar-selector.tsx
```diff
@@ -40,14 +40,16 @@ export function HotbarSelector({ hotbar }: Props) {
...
       <div className="box grow flex align-center">
-        <MaterialTooltip arrow title={hotbar.name}>
-          <Badge
-            id="hotbarIndex"
-            small
-            label={hotbarDisplayIndex(store.activeHotbarId)}
-            onClick={() => CommandOverlay.open(<HotbarSwitchCommand />)}
-          />
-        </MaterialTooltip>
+        <Badge
```
Nokel81, line 43: Why are we going back to our in-tree version instead? Wouldn't it be better to get rid of that and move to `MaterialTooltip` instead?

aleksfront: Well, the issue was marked to `5.0.0` and revering changes with some restyling became fastest way to solve thing. Going to `MaterialTooltip` way need to refactor `withTooltip` HOC, which is slower and more complex. Probably we should make another attempt on `MaterialTooltip` later.

Nokel81: Yes it was marked as `5.0.0` initially.

I guess I thought that the issue was about moving directly to `MaterialTooltip` since we have had over a week to do so. WDYT @leenamba ?

src/renderer/components/+catalog/catalog.tsx
aleksfront: Removed this prop.

src/renderer/components/hotbar/hotbar-selector.tsx
aleksfront: This PR fixes the described issue, so I suggest to merge it for 5.0.1. We can outline to Material tooltip in the upcoming plan.

chenhunghan (approved):
LGTM

PR #3199, opened 2021-06-28
Fix extension enabled status after installation

### LINKED ISSUE #3198: Extension installed from UI is not always enabled in 5.x
When user installs an extension from the UI, the extension is sometimes not enabled, even though the UI says "Enabled".

Technical background:
This seems to happen sometimes, so could be caused by a race condition. The "enabled" property is "false" in `lens-extensions.json`. User can e.g. restart Lens enable the extension as a workaround.

The console has warning `Lens.js:3784 [mobx.array] Attempt to read an array index (0) that is out of bounds (0). Please check length first. Out of bound indices will not be tracked by MobX` which might be related.

Tested on `Lens: 5.0.0-beta.12.20210625.2`.

panuhorsmalahti:
* Fixes https://github.com/lensapp/lens/issues/3198
* The cause of the bug: the `reaction()` was not triggered after `isEnabled = true` because we edited the computed `userExtensions` instead of the original extension object. This made the UI and the store to be not in sync.
* Thanks for @nevalla for debugging help
* Not extensively tested yet..

### DIFF
--- a/src/extensions/extension-loader.ts
+++ b/src/extensions/extension-loader.ts
@@ -192,6 +192,10 @@ export class ExtensionLoader extends Singleton {
     }
   }
 
+  setIsEnabled(lensExtensionId: LensExtensionId, isEnabled: boolean) {
+    this.extensions.get(lensExtensionId).isEnabled = isEnabled;
+  }
+
   protected async initMain() {
     this.isLoaded = true;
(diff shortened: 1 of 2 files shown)


Nokel81 (approved):

nevalla (approved):
LGTM

PR #3200, opened 2021-06-28
Filtering out clusters by metadata.id in hotbar migrations

### LINKED ISSUE #3126: Hot bar context menu not opening

aleksfront:
`cluster.id` doesn't seems to be a good choice for filtering since it is always unique even for same clusters. However, `metadata.id` is same.

Fixes #3126 

clusters in 4

clusters in 5

### DIFF
--- a/src/migrations/hotbar-store/5.0.0-beta.10.ts
+++ b/src/migrations/hotbar-store/5.0.0-beta.10.ts
@@ -22,7 +22,7 @@
 import { createHash } from "crypto";
 import { app } from "electron";
 import fse from "fs-extra";
-import { isNull } from "lodash";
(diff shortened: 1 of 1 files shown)


Nokel81:
Nice fix. Though if you checked it in a migration for `beta.13` then we would also fix it for users of the beta.

But that might not be that big of a problem.

msa0311 (approved):

PR #3211, opened 2021-06-29
Correctly migrate duplicate clusters in workspaces

Nokel81:
- Keep track of the set of all workspaces that each duplicated entry was a part of

This correctly migrates users' workspaces to hotbars, for users that don't immediately upgrade to `5.0.0` from `4.2.5` (but do upgrade to a version that includes this). Previously duplicated clusters would "forget" all workspaces except for the first one.

Note: this won't fix users that already upgraded to `5.0.0` because the previous version of this migration was destructive of this data.

### DIFF
--- a/src/common/cluster-store.ts
+++ b/src/common/cluster-store.ts
@@ -71,6 +71,11 @@ export interface ClusterModel {
    */
   workspace?: string;
 
+  /**
+   * @deprecated this is used only for hotbar migrations from 4.2.X
+   */
+  workspaces?: string[];
+
   /** User context in kubeconfig  */
   contextName: string;
 
--- a/src/main/cluster.ts
+++ b/src/main/cluster.ts
@@ -121,6 +121,10 @@ export class Cluster implements ClusterModel, ClusterState {
    */
   @observable workspace: string;
   /**
+   * @deprecated
+   */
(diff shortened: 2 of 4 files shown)


aleksfront:
Migration goes well, duplicated clusters can now stay in different hotbars.

Found backward compatibility issue though: when opening back 4.2.5 after 5.0.0 duplicated clusters disappear from all workspaces except one. We should preserve same workspace/cluster order regardless of starting lens 5.0.0.

Example:
Before running Lens 5 same clusters was located in all 3 workspaces. After running Lens 5 clusters stays only in middle workspace.


https://user-images.githubusercontent.com/9607060/124083566-af6b0080-da56-11eb-8e9e-10c3fa5e4e67.mov

I can suggest that it's unrelated to current PR. If so, this could be approved.

nevalla (approved):

Nokel81:
> Found backward compatibility issue though: when opening back 4.2.5 after 5.0.0 duplicated clusters disappear from all workspaces except one. We should preserve same workspace/cluster order regardless of starting lens 5.0.0.

I think that would be a good goal but I don't think we can do so because the workspace field in the cluster model only holds one workspace ID.

msa0311 (approved):

PR #3214, opened 2021-06-30
Use env var for registry mirror

pitwegner:
In an offline environment, using docker.io as a hardcoded registry does not work. If there's a better alternative to environment variables, please let me know.

### DIFF
--- a/src/main/shell-session/node-shell-session.ts
+++ b/src/main/shell-session/node-shell-session.ts
@@ -77,7 +77,7 @@ export class NodeShellSession extends ShellSession {
           }],
           containers: [{
             name: "shell",
-            image: "docker.io/alpine:3.13",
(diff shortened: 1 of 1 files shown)


Nokel81:
The only problem I see is if kubernetes pulls the image itself from the node vs locally. If the node is "offline" how would you be able to access it.

Furthermore, this is probably better done as a cluster setting since it seems like it would be something on a per-cluster basis vs all clusters you have access to.

pitwegner:
The node itself is offline and all images are pulled through a registry proxy. I agree that this should probably be a cluster setting instead. I'll refactor.

jakolehm:
https://github.com/lensapp/lens/pull/3249 seems to be similar.

Nokel81:
I agree @jakolehm I think this sort of thing should be a cluster setting not an env var.

Nokel81:
Closing due to #3249

PR #3218, opened 2021-06-30
Temporary add manual update notification to Lens 5 users

nevalla:
Lens 5 uses different `appId` than Lens 4.x and currently Lens 4.x users cannot auto update to Lens 5.

This PR will add notification to Lens 5 users to manually update to the next version that will use the old appId.

When we have released this fix, we can revert Lens to use `electron.com.kontena-lens` as appId so 4.x users can install the  automatic update.

image

To test this on dev env, please modify `app-updater.ts`:

```
export const startUpdateChecking = once(function (interval = 1000 * 60 * 60 * 24): void {
  if (isDevelopment || isTestEnv) {
    broadcastMessage(ManualUpdateAvailableChannel, "", {version: "5.0.0"});

    return;
  }
```

### DIFF
--- a/src/common/ipc/update-available.ipc.ts
+++ b/src/common/ipc/update-available.ipc.ts
@@ -22,6 +22,7 @@
 import type { UpdateInfo } from "electron-updater";
 
 export const UpdateAvailableChannel = "update-available";
+export const ManualUpdateAvailableChannel = "manual-update-available";
 export const AutoUpdateLogPrefix = "[UPDATE-CHECKER]";
 
 export type UpdateAvailableFromMain = [backChannel: string, updateInfo: UpdateInfo];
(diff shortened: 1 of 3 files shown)


nevalla:
@msa0311 @leenamba please take a look on wording. Anything you would like to change there.

leenamba:
A new version of Lens 5.0.0 of Lens IDE is available and requires a one time manual download and update.

To update, please download the new version from our website k8slens.dev and install it manually. Do you want to download it now and quit Lens?

nevalla:
> A new version of Lens 5.0.0 of Lens IDE is available and requires a one time manual download and update.
> 
> To update, please download the new version from our website k8slens.dev and install it manually. Do you want to download it now and quit Lens?

This PR will download the binary directly, do we want to open k8slens.dev website instead?

leenamba:
OK, I didn't know clicking the button would download it directly. Then let's change to:

```
A new version of Lens 5.0.0 of Lens IDE is available and requires a one time manual download and update.

Do you want to download it now and quit Lens?
```

nevalla:
We figured out the better option to manage this.

PR #3220, opened 2021-06-30
Add esbuild/esbuild-loader to reduce memory usage when `make dev`

chenhunghan:
If using `ts-loader`, editing React components in renderer, I often see
```
[dev:renderer] FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
[dev:renderer]  1: 0x101203f65 node::Abort() (.cold.1) 
```

(this can be avoid by adding `--max-old-space-size` to node, but with `esbuild-loader`, seems it can reduce memory footprint overall (without `--max-old-space-size`, and keep usage `<2GB`).)

This PR adds `esbuild`/`esbuild-loader` to avoid `Allocation failed - JavaScript heap out of memory`, improve DX experience for Lens developer.

`esbuild-loader` is optional, and only enable via `LENS_DEV_USE_ESBUILD_LOADER ` env variable.

### DIFF
--- a/src/common/getTSLoader.ts
+++ b/src/common/getTSLoader.ts
@@ -1,0 +1,63 @@
+/**
+ * Copyright (c) 2021 OpenLens Authors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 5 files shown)


aleksfront (approved):
Working good, didn't see any memory errors so far.

src/common/getTSLoader.ts
```diff
@@ -0,0 +1,63 @@
...
+ * 
+ * depends on env LENS_DEV_USE_ESBUILD_LOADER to use esbuild-loader (faster) or good-old ts-loader
+ * 
+ * @param testRegExp - the regex for webpack to conditional find the files 
+ * @returns ts/tsx webpack loader configuration object
+ */
+const getTSLoader = (
+  testRegExp: RegExp, transpileOnly = true
+) => {
+  const useEsbuildLoader = process.env.LENS_DEV_USE_ESBUILD_LOADER === "true";
```
Nokel81, line 35: Given that we are all running into this issue. Why not just switch to `esbuild-loader` whole hog?

chenhunghan: Esbuild-loader is not production ready, I think it's better we use ts-loader (typescript) for production build, or anyone not concern about speed to avoid surprises

Nokel81: Ah good to know.

PR #3225, opened 2021-06-30
Unblock subsequent Singleton.createInstance()

Nokel81:
- Also fix clusters with unknown workspaces

### DIFF
--- a/src/common/utils/singleton.ts
+++ b/src/common/utils/singleton.ts
@@ -46,12 +46,15 @@ export class Singleton {
   static createInstance<T, R extends any[]>(this: StaticThis<T, R>, ...args: R): T {
     if (!Singleton.instances.has(this)) {
       if (Singleton.creating.length > 0) {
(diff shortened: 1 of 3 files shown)


aleksfront:
In the part of ignoring clusters without `workspace` attribute it works as expected 👍🏻 

However, can you explain a bit more why this singleton fixes is needed and what cases this PR fixes?

Nokel81:
The singleton changes are needed because if the constructor of a singleton fails, and even if something else catches that errors, subsequent attempts to make new singletons (even of different types) will all fail.

build/set_build_version.ts
```diff
@@ -37,6 +37,7 @@ function getBuildChannel(): string {
     case "beta":
       return "beta";
     case undefined:
+    case "latest":
```
jakolehm: Unrelated change?

jakolehm (approved):
LGTM

PR #3227, opened 2021-06-30
Automatically sync all kubeConfigPaths

### LINKED ISSUE #3221: Connection to clusters with external kubeconfig files lost after upgrade to 5.0.0
**Describe the bug**
Ability to connect to clusters through external kubeconfig files broken in 5.0.0

**To Reproduce**
1. Dialog displayed offering to upgrade to latest version 5.0.0.  No warning that accepting will lose all connectivity to clusters if using external kubeconfig files.
2. Click to upgrade.
3. After restart with 5.0.0 previous clusters are shown in "hotbar" but saying "unknown".  Clusters can be added but only with copy-paste content (and with warning that clusters will not be merged into .kubeconfig) and not by accessing existing files.
4. %AppData%/Roaming/Lens/lens-cluster-store.json still contains cluster setups with correct "kubeConfigPath", and the kubeconfig file still exists and with correct content.

**Expected behavior**
1. Support for multiple external kubeconfigs should be supported (I don't want or need cloud sync of single kubeconfig file).
2. Failing that at the very least a strong warning that connectivity will be lost should be displayed before upgrade when external kubeconfig files are configured.

- Lens Version: now 5.0.0
- OS: Windows
- Installation method (e.g. snap or AppImage in Linux): in-app upgrade from previous version.

**Logs:**
```
info: ≡ƒôƒ Setting Lens as protocol client for lens://
info: ≡ƒôƒ Protocol client register succeeded Γ£à
```

**Kubeconfig:**
```
n/a - previous configs worked fine in v4.x, current kubeconfigs are not even processed
```

**Additional context**
none, but need access to clusters via lens asap.

Nokel81:

fixes #3221


src/common/hotbar-store.ts
```diff
@@ -94,7 +94,7 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
   }
 
   @action
-  protected async fromStore(data: Partial<HotbarStoreModel> = {}) {
+  protected fromStore(data: Partial<HotbarStoreModel> = {}) {
```
Nokel81: These cannot be `async`. See https://github.com/lensapp/lens/pull/3178

src/migrations/user-store/5.0.1.ts
```diff
@@ -0,0 +1,66 @@
...
+
+      if (syncPaths.has(cluster.kubeConfigPath) || syncPaths.has(dirOfKubeconfig)) {
+        migrationLog(`Skipping ${cluster.id} because kubeConfigPath is already being synced`);
+        continue;
+      }
+
+      if (!existsSync(cluster.kubeConfigPath)) {
+        migrationLog(`Skipping ${cluster.id} because kubeConfigPath no longer exists`);
+        continue;
+      }
```
Nokel81: I know this is technically a race condition but I think it is better not to confuse users by adding files that don't exist at this point.

src/common/hotbar-store.ts
nevalla: Are we sure this change does not have any side effects?

src/common/user-store.ts
```diff
@@ -206,6 +204,7 @@ export class UserStore extends BaseStore<UserStoreModel> {
     }
 
     if (preferences.syncKubeconfigEntries) {
+      console.log(preferences.syncKubeconfigEntries);
```
nevalla: debug leftover here?

aleksfront, line 207: Forgotten `console.log`

aleksfront:
It works for different kubeconfig paths 🎉 , but migration is started only if app version number is moved to `5.0.1`. Probably tests are failing for the same reason.

msa0311:
@aleksfront mentioned that this branch is causing app crash on startup: https://github.com/lensapp/lens/issues/3242
Please lets make sure this is fixed before merging the PR. @Nokel81

src/common/hotbar-store.ts
Nokel81: This actually fixed the fromStore not working

src/migrations/user-store/5.0.1.ts
```diff
@@ -0,0 +1,66 @@
...
+
+      migrationLog(`Adding ${cluster.kubeConfigPath} from ${cluster.id} to sync paths`);
+      syncPaths.add(cluster.kubeConfigPath);
+    }
+
+    const updatedSyncEntries: KubeconfigSyncEntry[] = [...syncPaths].map(filePath => ({ filePath }));
+
+    migrationLog("Final list of syned paths", updatedSyncEntries);
```
jim-docker:
```suggestion
    migrationLog("Final list of synced paths", updatedSyncEntries);
```

src/migrations/user-store/5.0.1.ts
```diff
@@ -0,0 +1,66 @@
...
+import { MigrationDeclaration, migrationLog } from "../helpers";
+
+export default {
+  version: "5.0.1",
+  run(store) {
+    const { syncKubeconfigEntries = [], ...preferences }: UserPreferencesModel = store.get("preferences") ?? {};
+    const { clusters = [] }: ClusterStoreModel = readJsonSync(path.resolve(app.getPath("userData"), "lens-cluster-store.json"));
+    const syncPaths = new Set(syncKubeconfigEntries.map(s => s.filePath));
+
+    syncPaths.add(path.join(os.homedir(), ".kube"));
```
jim-docker: Does this get executed even for a new install of Lens?

Nokel81: Yes because the file doesn't exist yet

jim-docker: So migrations get run even for a new install? Adding ".kube" to the list of sync'ed paths is "setup" and it seems odd to have it done in a migration (and not easy to find)

jim-docker: In light of #3254 should we just add ".kube/config"? User can manually add ".kube" in that case. But as it is the user cannot remove ".kube" (because it will keep getting added back, right?) Using ".kube" folder to save backups is probably not a typical use case but certainly possible. Could be done in another PR.

Nokel81: That was the intentional design. I would actually choose to close that issue as "working as intended"

src/migrations/user-store/5.0.2-beta.1.ts
```diff
@@ -0,0 +1,66 @@
...
+import { app } from "electron";
+import { existsSync, readJsonSync } from "fs-extra";
+import path from "path";
+import os from "os";
+import { ClusterStore, ClusterStoreModel } from "../../common/cluster-store";
+import type { KubeconfigSyncEntry, UserPreferencesModel } from "../../common/user-store";
+import { MigrationDeclaration, migrationLog } from "../helpers";
+
+export default {
+  version: "5.0.2-beta.1",
```
jakolehm:
```suggestion
  version: "5.0.3-beta.1",
```

package.json
```diff
@@ -3,7 +3,7 @@
   "productName": "OpenLens",
   "description": "OpenLens - Open Source IDE for Kubernetes",
   "homepage": "https://github.com/lensapp/lens",
-  "version": "5.0.2",
+  "version": "5.0.2-beta.1",
```
jakolehm:
```suggestion
  "version": "5.0.3-beta.1",
```

src/migrations/user-store/5.0.3-beta.1.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    for (const cluster of clusters) {
+      const dirOfKubeconfig = path.dirname(cluster.kubeConfigPath);
+
+      if (dirOfKubeconfig === ClusterStore.storedKubeConfigFolder) {
+        migrationLog(`Skipping ${cluster.id} because kubeConfigPath is under ClusterStore.storedKubeConfigFolder`);
+        continue;
+      }
+
+      if (syncPaths.has(cluster.kubeConfigPath) || syncPaths.has(dirOfKubeconfig)) {
```
jakolehm: Should this check include if any parent dir is already in `syncPaths`?

Nokel81: No because we only sync shallowly. See https://github.com/lensapp/lens/blob/master/src/main/catalog-sources/kubeconfig-sync.ts#L265

src/migrations/user-store/5.0.1.ts
Nokel81: FYI this migration will only happen once. So once this migration happens a user will be able to remove `~/.kube` to their hearts desire.

src/migrations/hotbar-store/5.0.0-beta.10.ts
```diff
@@ -148,8 +148,8 @@ export default {
 
       store.set("hotbars", hotbars);
     } catch (error) {
-      if (!(error.code === "ENOENT" && error.path.endsWith("lens-workspace-store.json"))) {
-        // ignore lens-workspace-store.json being missing
+      if (error.code !== "ENOENT") {
+        // ignore files being missing
```
jim-docker, line 152: (old) nit: this comment is about not entering the if statement

src/migrations/user-store/5.0.3-beta.1.ts
```diff
@@ -0,0 +1,75 @@
...
+          continue;
+        }
+
+        if (!existsSync(cluster.kubeConfigPath)) {
+          migrationLog(`Skipping ${cluster.id} because kubeConfigPath no longer exists`);
+          continue;
+        }
+
+        migrationLog(`Adding ${cluster.kubeConfigPath} from ${cluster.id} to sync paths`);
+        syncPaths.add(cluster.kubeConfigPath);
```
jim-docker, line 59: can a file get added here and then subsequently the file's dir is added, and then the file should be removed (since it is covered by the dir)?

jim-docker (approved):

src/migrations/user-store/5.0.3-beta.1.ts
Nokel81: No because we never add `dirOfKubeconfig`

src/migrations/hotbar-store/5.0.0-beta.10.ts
Nokel81: I fix that in a follow up PR.

src/migrations/user-store/5.0.3-beta.1.ts
```diff
@@ -0,0 +1,75 @@
...
+  version: "5.0.3-beta.1",
+  run(store) {
+    try {
+      const { syncKubeconfigEntries = [], ...preferences }: UserPreferencesModel = store.get("preferences") ?? {};
+      const { clusters = [] }: ClusterStoreModel = readJsonSync(path.resolve(app.getPath("userData"), "lens-cluster-store.json")) ?? {};
+      const syncPaths = new Set(syncKubeconfigEntries.map(s => s.filePath));
+
+      syncPaths.add(path.join(os.homedir(), ".kube"));
+
+      for (const cluster of clusters) {
```
jakolehm, line 40: We should only loop through local clusters. Looping through all clusters will have nasty side-effect -> it adds extension generated kubeconfigs as kubeconfig watch paths.

PR #3231, opened 2021-06-30
Allow users to disconnect from clusters that fail to connect

Nokel81:

Without this if a cluster fails to connect there is no way to tell Lens to stop trying to connect to it, bar quitting lens itself.

### DIFF
--- a/src/common/catalog-entities/kubernetes-cluster.ts
+++ b/src/common/catalog-entities/kubernetes-cluster.ts
@@ -55,8 +55,10 @@ export interface KubernetesClusterSpec extends CatalogEntitySpec {
   };
 }
 
+export type KubernetesClusterStatusPhase = "connected" | "connecting" | "disconnected" | "deleting";
+
 export interface KubernetesClusterStatus extends CatalogEntityStatus {
-  phase: "connected" | "disconnected" | "deleting";
+  phase: KubernetesClusterStatusPhase;
 }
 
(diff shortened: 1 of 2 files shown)


src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -130,23 +116,27 @@ export class KubernetesCluster extends CatalogEntity<CatalogEntityMetadata, Kube
...
+          icon: "link",
+          onClick: () => context.navigate(`/cluster/${this.metadata.uid}`)
+        });
+        break;
     }
 
-    const category = catalogCategoryRegistry.getCategoryForEntity<KubernetesClusterCategory>(this);
-
-    if (category) category.emit("contextMenuOpen", this, context);
+    catalogCategoryRegistry
```
nevalla, line 137: I'm not really a fan of these changes. Both are totally valid code and the change is not necessary, they are just causing noise. Or did I miss something?

nevalla (approved):

src/common/catalog-entities/kubernetes-cluster.ts
Nokel81: Do you specifically mean the change to how the emit is handled or even the switch?

nevalla: I mean that isn't your change doing exactly the same as
```
const category = catalogCategoryRegistry.getCategoryForEntity<KubernetesClusterCategory>(this);

if (category) category.emit("contextMenuOpen", this, context);
```

PR #3234, opened 2021-07-01
Add kubesync shortcut to catalog

### LINKED ISSUE #3232: Category onAdd for kubernetes sync
An easier way for users to discover the ability to sync kubernetes files.

**Why is this needed**:
Users of Lens5 have complained about not being able to select contexts from cluster files since that was removed in Lens5. However, there is the better feature of syncing a file from disk which exists. This should be an accordance that such a feature exists.

### LINKED ISSUE #3499: How to add cluster
**Describe the bug**
Clicking on "Add cluster" opens a page. But it is not clear what/how to add a kubeconfig here

**To Reproduce**
1. Go to 'Add cluster'

**Expected behavior**
In version 4 there was the possibility to select a kubeconfig.

**Screenshots**

- Lens: 5.1.3-latest.20210722.1
- OS: [e.g. OSX] macOS 11.4
- Installation method (e.g. snap or AppImage in Linux): upgrade from version 4

Nokel81:
- Add ability to register new app preference groupings

- Switch all lens preferences to be retreived from the registry

- Dynamically render preference's navigation based on the existance of
  items in each grouping

fixes https://github.com/lensapp/lens/issues/3232

Note: there are only three sections here on the navigation because this screenshot is being taken from a `make dev` run without the telemetry extension installed.

Also, it might make sense for the telemetry extension to provide its own preferences kind instead of Lens doing so.

Screen Shot 2021-06-30 at 11 26 26 PM
Screen Shot 2021-06-30 at 11 26 51 PM

---

New video showing new version that @jakolehm suggested:

Note: the video shows two items. Those will only be visible on windows. Only all other OS's it will only show one. That is because on windows a file picker can only pick directories XOR files.

https://user-images.githubusercontent.com/8225332/125652584-efc68caa-593a-4862-836a-d26e0a3ed635.mov


src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -174,13 +174,18 @@ export class KubernetesClusterCategory extends CatalogCategory {
...
+      ctx.menuItems.push(
+        {
+          icon: "text_snippet",
+          title: "Add from kubeconfig",
+          onClick: () => ctx.navigate(addClusterURL()),
+        },
+        {
+          icon: "settings",
+          title: "Sync kubeconfig file(s)",
+          onClick: () => ctx.navigate(preferencesURL({ fragment: "kube-sync" })),
```
Nokel81: All of this was done so that this could be added.

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -22,249 +22,105 @@
...
-          document.getElementById(fragment)?.scrollIntoView();
+        if (fragment.length > 0) {
+          const settingsItem = AppPreferenceRegistry.getInstance().getItems().find(item => item.id === fragment);
+
+          if (settingsItem) {
+            this.activeTab = settingsItem.showInPreferencesTab;
+            setTimeout(() => {
+              document.getElementById(fragment)?.scrollIntoView();
+            }, 150);
+          }
```
Nokel81, line 57: And the moving out of the in-repo settings to the registry was done because this is the correct logic (the previous logic no longer worked because the id wasn't necessarily in the DOM.

Technically, it would be more correct to do this scrollIntoView after the next render but I was not able to find a succinct method of doing that.

yarn.lock
```diff
@@ -6361,20 +6361,6 @@ filelist@^1.0.1:
...
-
-filenamify@^4.1.0:
-  version "4.1.0"
-  resolved "https://registry.yarnpkg.com/filenamify/-/filenamify-4.1.0.tgz#54d110810ae74eebfe115c1b995bd07e03cf2184"
-  integrity sha512-KQV/uJDI9VQgN7sHH1Zbk6+42cD6mnQ2HONzkXUfPJ+K2FC8GZ1dpewbbHw0Sz8Tf5k3EVdHVayM4DoAwWlmtg==
-  dependencies:
-    filename-reserved-regex "^2.0.0"
-    strip-outer "^1.0.1"
-    trim-repeated "^1.0.0"
-
```
Nokel81: These changes seem to be required by yarn even for master.

Nokel81:
From discussion:

- Make syncing a file the default action
- Don't switch to preferences after adding a sync, but instead show a notification with a button to go to the settings.

aleksfront:
Part of this related to moving all Preference elements to registries are controversial and should be created within another PR opened for discussion. One of the reasons blocking this feature is much less control over styling Preference parts (here they tightly wrapper with same `<section className="small">`, `<hr className="small"/>` and so on. Not all elements will have similar grid and look same.

I will create new one with code extracted from here related to fixing the issue #3232.

Nokel81:
> Part of this related to moving all Preference elements to registries are controversial and should be created within another PR opened for discussion.

I had it as part of this PR because there needs to be a way to know which section and which page to jump to.

I didn't want to special case this.

aleksfront:
Closing in favor of #3573

PR #3249, opened 2021-07-02
Make node shell image configurable

### LINKED ISSUE #2007: node-shell : Failed to pull image "docker.io/alpine:3.12"
**Describe the bug**

node-shell-888ce968-f3b4-4b86-9bd6-ba4487a37604

Failed to pull image "docker.io/alpine:3.12": rpc error: code = Unknown desc = Error response from daemon: Get https://registry-1.docker.io/v2/: dial tcp: lookup registry-1.docker.io on [::1]:53: read udp [::1]:46565->[::1]:53: read: connection refused

**To Reproduce**
when k8s cluster network cann't connect to internat, k8s only can connect to local harbor registry，can not pull the internet image docker.io/alpine:3.12

**Code In**
in this code file:  lens/src/main/node-shell-session.ts  line 78, have this line  image: "docker.io/alpine:3.12", 

**Solve Suggestion**
Is this can make a config  to solve the problem;

or use K8S DaemonSet YAML, I can change the DaemonSet YAML in runtime to solve the problem;

vshakirova:
deploy node shell pod using an image from a private repo
Screenshot from 2021-07-08 17-01-43

settings
Screen Shot 2021-07-12 at 4 45 20 PM

fixes: https://github.com/lensapp/lens/issues/2007

### DIFF
--- a/src/main/cluster.ts
+++ b/src/main/cluster.ts
@@ -34,6 +34,7 @@ import { VersionDetector } from "./cluster-detectors/version-detector";
 import { detectorRegistry } from "./cluster-detectors/detector-registry";
 import plimit from "p-limit";
 import { toJS } from "../common/utils";
+import { initialNodeShellImage } from "../common/cluster-store";
 
 export enum ClusterStatus {
   AccessGranted = 2,
@@ -744,5 +745,13 @@ export class Cluster implements ClusterModel, ClusterState {
 
   isMetricHidden(resource: ClusterMetricsResourceType): boolean {
     return Boolean(this.preferences.hiddenMetrics?.includes(resource));
+  }
+
+  get nodeShellImage(): string {
+    return this.preferences.nodeShellImage || initialNodeShellImage;
+  }
+
+  get imagePullSecret(): string {
+    return this.preferences.imagePullSecret || "";
   }
 }
--- a/src/renderer/components/cluster-settings/components/cluster-node-shell-setting.tsx
+++ b/src/renderer/components/cluster-settings/components/cluster-node-shell-setting.tsx
@@ -1,0 +1,112 @@
+/**
+ * Copyright (c) 2021 OpenLens Authors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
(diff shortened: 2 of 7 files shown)


src/main/cluster.ts
```diff
@@ -736,4 +737,12 @@ export class Cluster implements ClusterModel, ClusterState {
...
     return Boolean(this.preferences.hiddenMetrics?.includes(resource));
   }
+
+  getNodeShellImage(): string {
+    return this.preferences.nodeShellImage || initialNodeShellImage;
+  }
+
+  getImagePullSecret(): string {
+    return this.preferences.imagePullSecret || "";
+  }
```
Nokel81: These two should be getters probably.

src/main/cluster-manager.ts
```diff
@@ -149,7 +149,8 @@ export class ClusterManager extends Singleton {
         this.store.addCluster({
           id: entity.metadata.uid,
           preferences: {
-            clusterName: entity.metadata.name
+            clusterName: entity.metadata.name,
+            nodeShellImage: initialNodeShellImage,
```
Nokel81: No need to set this here if the getter defaults to this.

src/renderer/components/cluster-settings/components/cluster-node-shell-setting.tsx
```diff
@@ -0,0 +1,140 @@
...
+          <FormSwitch
+            control={
+              <Switcher
+                checked={this.usePrivateRegistry}
+                onClick={this.onSwitchChange}
+                name="private-registry"
+              />
+            }
+            label="Use private registry"
+          />
```
Nokel81: I think this will be confusing for users because it is essentially a "clear imagePullSecret" button. I think it would be better to not have it and just add something like "Optional" to the hint for the `imagePullSecret`

src/renderer/components/cluster-settings/components/cluster-node-shell-setting.tsx
```diff
@@ -0,0 +1,140 @@
...
+          <SubTitle title="Image pull secret" id="image-pull-secret"/>
+          <Input
+            theme="round-black"
+            value={this.imagePullSecret}
+            onChange={this.onSecretChange}
+            onBlur={this.saveSecret}
+            iconRight={clearSecretIcon}
+          />
+          <small className="hint">
+            Image pull secret. Used for pulling image from a private registry.
```
Nokel81: This should specify that it is the name of a pre-existing secret object.

src/main/cluster.ts
vshakirova: fixed, thank you.

src/renderer/components/cluster-settings/components/cluster-node-shell-setting.tsx
vshakirova: Is it better? 
Screen Shot 2021-07-12 at 4 45 16 PM

Nokel81: Yes thanks.

src/renderer/components/cluster-settings/components/cluster-node-shell-setting.tsx
```diff
@@ -0,0 +1,114 @@
...
+    this.nodeShellImage = initialNodeShellImage; //revert to default
+  };
+
+  clearSecret = () => {
+    this.imagePullSecret = "";
+  };
+
+  render() {
+    const resetImageIcon = <Icon small material="close" onClick={this.resetImage} tooltip={"Reset"}/>;
+    const clearSecretIcon = <Icon small material="close" onClick={this.clearSecret} tooltip={"Clear"}/>;
```
Nokel81: Nit: the tooltip props can just be `tooltip="Reset"` (etc...)

Nit: these are only each used once, so I think it would be better to just declare them inline.

Nokel81 (commented):
One final thing and then LGTM!

Nokel81 (approved):

Nokel81:
closes https://github.com/lensapp/lens/pull/3214

meganeinu7:
@vshakirova I am wondering how I can access this feature. I downloaded the latest version - Lens: 5.4.5-latest.20220405.1
, but I can't find the UI that I see here in this PR.
I am not sure this is the right place to ask this question...

meganeinu7:
@vshakirova I could not find where to ask this question... 
I downloaded the latest version of Lens, but I was not able to access this feature.
Could you let me know where I can find any documentation on this feature?

PR #3274, opened 2021-07-05
Fix problem with in-tree extension load

### LINKED ISSUE #3266: Lens5 - no more logs and console buttons when a pod is selected?
**Describe the bug**
image

**To Reproduce**
Click on a POD ... I used to have a console and logs to inspect a pod, no more since Lens 5.

- Lens Version: 5
- OS: windows
- Installation method (e.g. snap or AppImage in Linux): Windows setup

jakolehm:
Fixes #3266

Reverts changes in #3261 and includes a fix from #3256

### DIFF
--- a/src/extensions/extension-discovery.ts
+++ b/src/extensions/extension-discovery.ts
@@ -365,7 +365,13 @@ export class ExtensionDiscovery extends Singleton {
       let isCompatible = isBundled;
 
       if (manifest.engines?.lens) {
-        isCompatible = semver.satisfies(appSemVer, manifest.engines.lens);
+        const appSemVerLatestImplied = appSemVer;
+
+        if (appSemVerLatestImplied.prerelease?.[0] === "latest") {
+          /* remove the "latest" prerelease tag so as not to require the extension to specify it */
+          appSemVerLatestImplied.prerelease = [];
+        }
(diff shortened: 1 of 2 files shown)


jakolehm:
Tested & verified by doing a local build and `.dmg` install:
image

msa0311 (approved):

jim-docker:
Did you test on Windows or Ubuntu? #3266 says mac was OK, problem happens on Windows and Ubuntu.

jakolehm:
> Did you test on Windows or Ubuntu? #3266 says mac was OK, problem happens on Windows and Ubuntu.

MacOS has the same problem. Did not test on Windows/Ubuntu because this shouldn't be OS specific issue (it's a version mismatch between Lens and extension).

src/extensions/extension-discovery.ts
```diff
@@ -365,7 +365,13 @@ export class ExtensionDiscovery extends Singleton {
       let isCompatible = isBundled;
 
       if (manifest.engines?.lens) {
```
jim-docker, line 367: Is this code actually relevant to the bundled extensions? If so should we skip this if the extension is bundled?
```
if (!isBundled && manifest.engines?.lens) {
```

jakolehm: You are right, we could skip the check for bundled extensions.

PR #3295, opened 2021-07-07
Added ability to add custom port on pod and service port forwarding

rdeepc:
Signed-off-by: rdeepc <12953177+rdeepc@users.noreply.github.com>


src/main/routes/port-forward-route.ts
```diff
@@ -44,10 +45,10 @@ class PortForward {
...
     return PortForward.portForwards.find((pf) => (
       pf.clusterId == forward.clusterId &&
-      pf.kind == forward.kind &&
-      pf.name == forward.name &&
-      pf.namespace == forward.namespace &&
-      pf.port == forward.port
+        pf.kind == forward.kind &&
+        pf.name == forward.name &&
+        pf.namespace == forward.namespace &&
+        pf.port == forward.port
```
Nokel81: Please revert formatting changes.

src/main/routes/port-forward-route.ts
```diff
@@ -74,7 +77,7 @@ class PortForward {
       "port-forward",
       "-n", this.namespace,
       `${this.kind}/${this.name}`,
-      `:${this.port}`
+      `${this.forwardPort??""}:${this.port}`
```
Nokel81, line 80: If `forwardPort` can be of type `null | undefined` then the fields should all be set to either `optional` or include those types. (I think optional would be better)

src/main/routes/port-forward-route.ts
```diff
@@ -118,33 +125,73 @@ class PortForward {
...
+
+    const portForward = PortForward.getPortforward({
+      clusterId: cluster.id, kind: resourceType, name: resourceName,
+      namespace, port, forwardPort,
+    });
 
-    respondJson(response, {});
+    portForward.stop().then( () => {
+      respondJson(response, {status: true});
+    }).catch(error => logger.error(error));
```
Nokel81: Please do not use promise functions in an async context.

src/main/routes/port-forward-route.ts
```diff
@@ -118,33 +125,73 @@ class PortForward {
...
+    let forwardedPort = -1;
 
-      if (!started) {
-        return respondJson(response, {
-          message: "Failed to open port-forward"
-        }, 400);
-      }
+    if (portForward) {
+      forwardedPort = Number(portForward.forwardPort);
     }
```
Nokel81: This can be simplified:

```suggestion
    const forwardedPort = Number(portForward?.forwardPort ?? -1);
```

src/renderer/components/+network-services/service-port-component.tsx
```diff
@@ -28,20 +28,44 @@ import { apiBase } from "../../api";
...
 }
 
+interface PortForwardResult {
+  port: number;
+}
+
 @observer
 export class ServicePortComponent extends React.Component<Props> {
   @observable waiting = false;
+  @observable forwardPort = -1;
```
Nokel81: I think it would be better to use `null` or `undefined` as the sentinel value instead of a special number.

src/renderer/components/+network-services/service-port-component.tsx
```diff
@@ -28,20 +28,44 @@ import { apiBase } from "../../api";
...
+  @observable isPortForwarded = false;
 
   constructor(props: Props) {
     super(props);
     makeObservable(this);
+    this.init();
+  }
+
+  init() {
+    this.checkExistingPortForwarding().then();
```
Nokel81: Please remove the `then`. If you are meaning to ignore the output, either do `.catch(noop)` or match `init()` an async function, await the result in a `try/catch` and then either log the error or add a comment indicating that you want to ignore it. Best case would be to print an error message.

src/renderer/components/+network-services/service-port-component.tsx
```diff
@@ -61,14 +104,32 @@ export class ServicePortComponent extends React.Component<Props> {
...
+        <text>to</text>
+        <Input className={"portInput"}
+          type="number"
+          min="0"
+          max="65535"
+          value= {String(this.forwardPort)}
+          disabled={this.isPortForwarded}
+          onChange={(value) => this.forwardPort = Number(value)}
+        />
+        <Button onClick={() => portForwardAction()}> {this.isPortForwarded ? "Stop":"Forward"} </Button>
```
Nokel81: Please add some screenshots of how this looks

rdeepc: Thanks for the feedback, I have made the changes as suggested

src/renderer/components/+network-services/service-port-component.tsx
```diff
@@ -61,14 +105,32 @@ export class ServicePortComponent extends React.Component<Props> {
   render() {
     const { port } = this.props;
 
+    if (this.forwardPort == null) {
+      this.forwardPort = port.port;
```
jim-docker: I think this is generally not a good idea. If the pod or service port is 80 then suggesting the forward be 80 is a problem. Also, this will cause problems for frequently used ports like 8080, when you want to forward more than one service that exposes 8080. 
If `this.forwardPort` is `null` I think the port input field should start as blank (or some placeholder text like "random"  or "arbitrary port"). If the user leaves it blank then Lens does what it did before (lets `kubectl` find a port)

src/renderer/components/+network-services/service-port-component.tsx
```diff
@@ -61,14 +105,32 @@ export class ServicePortComponent extends React.Component<Props> {
...
-          {this.waiting && (
-            <Spinner />
-          )}
-        </span>
+        {port.toString()}
+        {" "}
+        <text>to</text>
+        <Input className={"portInput"}
+          type="number"
+          min="0"
```
jim-docker, line 123:
```suggestion
          min="1"
```
I think for TCP, port number 0 is reserved and cannot be used. I guess it's debatable if any of the low numbers should be used

jim-docker (commented):
thanks for contributing this!

src/renderer/components/+network-services/service-port-component.tsx
rdeepc: Thanks for the suggestion I have added the random port option if a port is not given, with a Random placeholder

src/renderer/components/+network-services/service-port-component.tsx
rdeepc: I have used the min port as 0, as kubectl allocates random port for input 0, Thanks for your input

jim-docker (approved):
@rdeepc I've changed the target for this PR to `lensapp:feature/port-forward-dashboard` (which is currently even with `master`). In July I used your work as a starting point for a port-forward dashboard that I implemented in a hackweek project here at Mirantis (see https://github.com/jim-docker/lens/tree/port-forward-dashboard). I am now beginning a proper integration into Lens, starting with your work, which will eventually get merged with `master`.  Your PR will require some changes to work with the dashboard, and I assume you are OK with us continuing the work? Thanks again for this contribution

rdeepc:
@jim-docker Thanks for the update. I love using Lens and I am happy to contribute.

PR #3296, opened 2021-07-07
Lazy load comp on route to reduce Lens.js bundle size

chenhunghan:
Add default export, and lazy load React components on route. Reduce Lens.js from 60MB to about 37MB (with tailwind JIT mode on).

Development build
Screenshot 2021-07-07 at 08 47 26

Production build
Screenshot 2021-07-07 at 08 46 20

!! Further tests needed. This PR will stay at draft.


aleksfront:
Can't spot any visible difference while testing. Production build is quite fast with and without lazy loading. We still has some issues when opening a Catalog, but it's the rendering case.

chenhunghan:
I think we can close unless needed in the future.

PR #3308, opened 2021-07-07
Fix listing charts failing due to bad versions

### LINKED ISSUE #3250: Apps/Charts endless loading
**Describe the bug**
After few latest updates (including version 5) of Lens - Apps/Charts are in endless loading.
Where the Apps/Releases works fine.

**To Reproduce**
1. Go to 'Apps/Charts'

**Expected behavior**
Full list of Helm charts (based on installed repositories) in the UI

**Screenshots**
Screenshot 2021-07-02 at 11 52 20 AM

- Lens Version: 5.0.0-latest.20210629.3
- OS: macOS Big Sur 11.4

**Logs:**
Nothing interesting, except 
```
[CLUSTER]: refresh {"id":"*","name":"dev","ready":true,"online":true,"accessible":true,"disconnected":false}
```

**Kubeconfig:**
N/A

**Additional context**
Is there any option to enable debug mode, see more precise logs?

Nokel81:
- Add generic sorting function that uses the comparitors directory

- Because repos are 3rd party managed resources we cannot enforce strict
  semver compatibility. Therefore we should loosely coerce versions to
  semver versions.

- If a version is not even loosely coercable then we should just sort
  alphanumerically.

- Sort semver before non-semver

- Filter out a chart if any of the versions are marked as deprecated

fixes https://github.com/lensapp/lens/issues/3250

---
NOTE: please review each commit separately and do not squash this PR.

### DIFF
--- a/src/main/helm/helm-service.ts
+++ b/src/main/helm/helm-service.ts
@@ -19,13 +19,14 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import semver from "semver";
+import semver, { SemVer } from "semver";
 import type { Cluster } from "../cluster";
 import logger from "../logger";
 import { HelmRepoManager } from "./helm-repo-manager";
 import { HelmChartManager } from "./helm-chart-manager";
-import type { HelmChartList, RepoHelmChartList } from "../../renderer/api/endpoints/helm-charts.api";
+import type { HelmChart, HelmChartList, RepoHelmChartList } from "../../renderer/api/endpoints/helm-charts.api";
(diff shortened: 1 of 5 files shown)


src/main/helm/helm-service.ts
```diff
@@ -132,28 +133,55 @@ class HelmService {
...
+    return Object.fromEntries(
+      iter.filterMap(
+        Object.entries(chartGroups),
+        ([name, charts]) => {
+          for (const chart of charts) {
+            if (chart.deprecated) {
+              // ignore chart group if any chart is deprecated
+              return undefined;
+            }
+          }
```
jim-docker, line 145: Can you add this change of behaviour to the PR description?

jim-docker (approved):

PR #3312, opened 2021-07-07
Cleanup `ResourceApplier` class

Nokel81:

- Fix ResourceApplier to not use sync methods within an async context
  - This is important because calling synchronous IO functions within an async context still blocks the entire event loop.
- There should be no observable changes in behaviour because of this.
- Use the inspect method for the process paths so to correctly handle different types of quotes in the pathnames.


src/main/resource-applier.ts
```diff
@@ -21,56 +21,73 @@
...
 import { appEventBus } from "../common/event-bus";
 import { cloneJsonObject } from "../common/utils";
+import { inspect } from "util";
+import { promiseExec } from "./promise-exec";
+
+function sanitize(raw: string): string {
+  return inspect(raw, false, undefined, false);
+}
+
+function sanitizeObject(resource: KubernetesObject | any) {
```
jakolehm: How is this refactoring related to PR?

Nokel81: I considered this to be mostly a refactoring and correctness PR. But I realize that the PR title does not capture that. Will update the title.

jakolehm: Ok, so in the end we just moved a method from a class to a separate function. Is there a reason?

src/main/resource-applier.ts
```diff
@@ -21,56 +21,73 @@
...
 import path from "path";
 import * as tempy from "tempy";
 import logger from "./logger";
 import { appEventBus } from "../common/event-bus";
 import { cloneJsonObject } from "../common/utils";
+import { inspect } from "util";
+import { promiseExec } from "./promise-exec";
+
+function sanitize(raw: string): string {
+  return inspect(raw, false, undefined, false);
```
jakolehm: Why do we need to `sanitize` with `inspect`?

jakolehm: Oh, it was mentioned in description.

Nokel81: Yes, though I have now realized that this is incorrect. I will try and find a different way of doing this. 

`inspect` doesn't produce the result we would like if the raw string includes back ticks

src/main/resource-applier.ts
Nokel81: I moved it to a separate function because it doesn't need anything from the class

src/main/resource-applier.ts
```diff
@@ -75,25 +84,27 @@ export class ResourceApplier {
...
+    return this.kubectlApply(yaml.dump(sanitizeObject(resource)));
   }
 
   protected async kubectlApply(content: string): Promise<string> {
     const kubectl = await this.cluster.ensureKubectl();
     const kubectlPath = await kubectl.getPath();
     const proxyKubeconfigPath = await this.cluster.getProxyKubeconfigPath();
     const fileName = tempy.file({ name: "resource.yaml" });
+
+    await fse.writeFile(fileName, content);
```
jim-docker: You could put the `try` before this to maintain the catch/rethrow logic?

src/main/resource-applier.ts
```diff
@@ -126,47 +136,29 @@ export class ResourceApplier {
...
-
-          return;
-        }
-
-        resolve(stdout);
-      });
-    });
-  }
+    await Promise.all(
+      resources.map((resource, index) => fse.writeFile(path.join(tmpDir, `${index}.yaml`), resource))
```
jim-docker: what if there's an error writing the file?

src/main/resource-applier.ts
```diff
@@ -126,47 +136,29 @@ export class ResourceApplier {
...
+    const tmpDir = tempy.directory();
 
-    return new Promise((resolve, reject) => {
-      const tmpDir = tempy.directory();
-
-      // Dump each resource into tmpDir
-      resources.forEach((resource, index) => {
-        fs.writeFileSync(path.join(tmpDir, `${index}.yaml`), resource);
-      });
-      args.push("-f", `"${tmpDir}"`);
```
jim-docker, line 129: this is no longer necessary?

src/main/resource-applier.ts
Nokel81: Yes, I don't actually remember why I moved this.

src/main/resource-applier.ts
Nokel81: Do you mean to imply about cleanup?

src/main/resource-applier.ts
```diff
@@ -110,47 +105,50 @@ export class ResourceApplier {
...
+
+    const firstError = results.find(result => result.status === "rejected") as PromiseRejectedResult;
 
-  protected sanitizeObject(resource: KubernetesObject | any) {
-    resource = cloneJsonObject(resource);
-    delete resource.status;
-    delete resource.metadata?.resourceVersion;
-    const annotations = resource.metadata?.annotations;
+    if (firstError) {
+      await Promise.allSettled(
```
jim-docker: You could just delete `tmpDir`? Though I would've thought that it was cleaned up automatically via`tempy`. Except looking at the docs it's not clear. If you use `tempy.directory.task` for sure the temp dir is cleaned.

My initial concern about the `promise.all()` was that if fse.writeFile() threw an error you'd just have a rejected promise in the results, which weren't checked. Execution would've continued to do the kubectl apply, no?

Nokel81: `promise.all()` rejects as soon as any of the promises rejects and only resolves if all of them do

Nokel81: FYI I have now moved it so that we always clean up the folder, even on success.

src/main/resource-applier.ts
```diff
@@ -107,38 +107,32 @@ export class ResourceApplier {
...
+      } catch (error) {
+        logger.error(`[RESOURCE-APPLIER] cmd errored: ${error}`);
 
-      const splitError = String(error).split(`.yaml": `);
+        const splitError = String(error).split(`.yaml": `);
 
-      return splitError[1] ?? error;
+        return splitError[1] ?? error;
+      }
+    } finally {
```
jim-docker: shouldn't you catch here too? Assuming you don't want the error from the `Promise.all()` call to propagate up. Or maybe just do all this with one try/catch/finally?

Nokel81: Very good catch, I just realized that I did the transformation here wrong; the `splitError` line should be `throw` not return since in the original version is a `reject`.

I will write some tests for this.

Nokel81:
Closing due to no time to get this working.

PR #3313, opened 2021-07-08
Ignore extensions' clusters when adding to kubeConfigSync

Nokel81:
- To accomplish this we now ignore all configs that are logiaclly in a
  child directory of $userData/extension_data. This is done at the path
  level so as to minimize the number of FS interactions.

fixes bug in #3227

### DIFF
--- a/src/common/utils/paths.ts
+++ b/src/common/utils/paths.ts
@@ -33,3 +33,35 @@ function resolveTilde(filePath: string) {
 export function resolvePath(filePath: string): string {
   return path.resolve(resolveTilde(filePath));
 }
+
+/**
+ * Checks if `testPath` represents a potential filesystem entry that would be
+ * logically "within" the `parentPath` directory.
+ *
+ * This function will return `true` in the above case, and `false` otherwise.
+ * It will return `false` if the two paths are the same (after resolving them).
+ *
+ * The function makes no FS calls and is platform dependant. Meaning that the
+ * results are only guaranteed to be correct for the platform you are running
+ * on.
+ * @param parentPath The known path of a directory
+ * @param testPath The path that is to be tested
+ */
+export function isLogicalChildPath(parentPath: string, testPath: string): boolean {
+  parentPath = path.resolve(parentPath);
+  testPath = path.resolve(testPath);
+
+  if (parentPath === testPath) {
+    return false;
+  }
+
+  while (testPath.length >= parentPath.length) {
+    if (testPath === parentPath) {
+      return true;
+    }
+
+    testPath = path.dirname(testPath);
+  }
+
+  return false;
+}
--- a/src/migrations/user-store/5.0.3-beta.1.ts
+++ b/src/migrations/user-store/5.0.3-beta.1.ts
@@ -20,7 +20,7 @@
  */
 
 import { app } from "electron";
-import { existsSync, readJsonSync } from "fs-extra";
+import { existsSync, readFileSync } from "fs";
 import path from "path";
 import os from "os";
 import { ClusterStore, ClusterStoreModel } from "../../common/cluster-store";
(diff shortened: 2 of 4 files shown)


src/migrations/user-store/5.0.3-beta.1.ts
```diff
@@ -50,6 +52,13 @@ export default {
           continue;
         }
 
+        const relativeToExtensionData = path.relative(cluster.kubeConfigPath, extensionDataDir);
+
+        if (relativeToExtensionData === "" || relativeToExtensionData.match(/^(\.\.)([\\\/]\.\.)*$/)) {
+          migrationLog(`Skipping ${cluster.id} because kubeConfigPath is placed under an extension_data folder`);
+          continue;
+        }
```
jim-docker: without regex: https://stackoverflow.com/a/45242825

jim-docker (commented):
Does this cover extensions in `~/.k8slens/extensions/` too?

Nokel81:
`~/.k8slens/extensions/` is where the code of the extensions are, not the `extensions_data` folder (which we provide to extensions via the `LensExtension.getExtensionFileFolder()` method.

src/migrations/user-store/5.0.3-beta.1.ts
Nokel81: I agree that solution is nice but it fails if the folder or file name starts with `..`.

jim-docker:
> `~/.k8slens/extensions/` is where the code of the extensions are, not the `extensions_data` folder (which we provide to extensions via the `LensExtension.getExtensionFileFolder()` method.

That's good. So if an extension dev writes outside of the extension file folder, it's at their own risk.

Nokel81:
Yes there is only so much we can do at in terms of checking in this migration. If I had done this both correctly and initially then we wouldn't have needed these checks.

src/migrations/user-store/5.0.3-beta.1.ts
Nokel81: I have done some more testing and have figured out a more robust way of checking this. Will update PR with tests.

src/migrations/user-store/5.0.3-beta.1.ts
```diff
@@ -32,9 +32,11 @@ export default {
   run(store) {
     try {
       const { syncKubeconfigEntries = [], ...preferences }: UserPreferencesModel = store.get("preferences") ?? {};
-      const { clusters = [] }: ClusterStoreModel = readJsonSync(path.resolve(app.getPath("userData"), "lens-cluster-store.json")) ?? {};
+      const { clusters = [] }: ClusterStoreModel = JSON.parse(readFileSync(path.resolve(app.getPath("userData"), "lens-cluster-store.json"), "utf-8")) ?? {};
+      const extensionDataDir = path.resolve(app.getPath("userData"), "extension_data");
       const syncPaths = new Set(syncKubeconfigEntries.map(s => s.filePath));
 
+      console.log(extensionDataDir);
```
jim-docker: debug?

src/migrations/user-store/5.0.3-beta.1.ts
jim-docker: What path starts with `..`? `extensionDataDir` is an absolute path, right?

Nokel81: A folder or file.

src/common/utils/paths.ts
```diff
@@ -33,3 +33,35 @@ function resolveTilde(filePath: string) {
...
+ */
+export function isLogicalChildPath(parentPath: string, testPath: string): boolean {
+  parentPath = path.resolve(parentPath);
+  testPath = path.resolve(testPath);
+
+  if (parentPath === testPath) {
+    return false;
+  }
+
+  while (testPath.length >= parentPath.length) {
```
jakolehm, line 58: Instead of while loop why not just `testPath.startsWith(parentPath)`?

Nokel81: Because of the following case:

```ts
const testPath = "/foo/bar";
const parentPath = "/fo";
```

src/migrations/user-store/5.0.3-beta.1.ts
jim-docker: when will you encounter a file or folder that starts with `..`? `path.resolve()` always returns an absolute path, and that's what you're starting with. `extensionDataDir` and `cluster.kubeConfigPath` are absolute. I don't see what's wrong with https://stackoverflow.com/a/45242825

Nokel81: I mean the name of the folder is `..blah` so the path could be `/foo/bar/..blah`

jim-docker: oh, ugh.

src/common/utils/paths.ts
jim-docker: `testPath.startsWith(parentPath + path.sep)`?

Nokel81: Nope because on windows both \ and / are valid separators. I could check if I am on windows first though

jakolehm (approved):
LGTM

PR #3314, opened 2021-07-08
Sort pod names by convering sub-parts

### LINKED ISSUE #134: sort statefulset pods from 0 to .....n
1. scale statefulset to 10
2. see pod-0, pod-1, ..., pod-10
3. scale statefulset to 1000
4. see pod-0, pod-1, pod-10, pod-100, pod-101.....

imo would make sense to "keep" the sort and not do it lexiographically

screenshot related:
Lens

Nokel81:
- Switches to a hand-rolled sorting function for `<Table>` so that the
  behaviour when a sortingCallback returns an array of elements can be
  defined to be sorting each item.

- Require `<Table>` to have tableId if sortable is provided

fixes #134 

---

[Notes]:
- When using `<Table>`; if the props field `tableId` is not provided and the `sortable` field is. An error will now be thrown. Previously only an error would be logged to the console. However, this would lead to a bug where the sorting state was not persisted correctly.

### DIFF
--- a/src/common/utils/array.ts
+++ b/src/common/utils/array.ts
@@ -19,21 +19,26 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-// Common usage utils & helpers
+export type Tuple<T, N extends number> = N extends N ? number extends N ? T[] : _TupleOf<T, N, []> : never;
+type _TupleOf<T, N extends number, R extends unknown[]> = R["length"] extends N ? R : _TupleOf<T, N, [T, ...R]>;
 
-export * from "../../common/utils";
-export * from "../../common/event-emitter";
+/**
+ *
+ * @param sources The source arrays
+ * @yields A tuple of the next element from each of the sources
+ * @returns The tuple of all the sources as soon as at least one of the sources is exausted
+ */
+export function* zipStrict<T, N extends number>(...sources: Tuple<T[], N>): Iterator<Tuple<T, N>, Tuple<T[], N>> {
+  const maxSafeLength = sources.reduce((prev, cur) => Math.min(prev, cur.length), Number.POSITIVE_INFINITY);
 
-export * from "./copyToClipboard";
-export * from "./createStorage";
-export * from "./cssNames";
-export * from "./cssVar";
-export * from "./display-booleans";
-export * from "./interval";
-export * from "./isMiddleClick";
-export * from "./isReactNode";
-export * from "./metricUnitsToNumber";
-export * from "./name-parts";
-export * from "./prevDefault";
-export * from "./saveFile";
-export * from "./storageHelper";
+  if (!isFinite(maxSafeLength)) {
+    // There are no sources, thus just return
+    return [] as Tuple<T[], N>;
+  }
+
+  for (let i = 0; i < maxSafeLength; i += 1) {
+    yield sources.map(source => source[i]) as Tuple<T, N>;
+  }
+
+  return sources.map(source => source.slice(maxSafeLength)) as Tuple<T[], N>;
+}
--- a/src/renderer/components/table/__tests__/getSorted.test.ts
+++ b/src/renderer/components/table/__tests__/getSorted.test.ts
@@ -1,0 +1,155 @@
+/**
+ * Copyright (c) 2021 OpenLens Authors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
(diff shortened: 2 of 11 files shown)


src/renderer/components/table/table.tsx
```diff
@@ -92,7 +91,7 @@ export class Table extends React.Component<TableProps> {
     const { sortable, tableId } = this.props;
 
     if (sortable && !tableId) {
-      console.error("[Table]: sorted table requires props.tableId to be specified");
+      throw new Error("Table must have props.tableId if props.sortable is specified");
```
jakolehm: Isn't this a breaking change?

src/renderer/components/table/table.tsx
```diff
@@ -136,14 +135,50 @@ export class Table extends React.Component<TableProps> {
     return headElem;
   }
 
-  getSorted(items: any[]) {
-    const { sortBy, orderBy: order } = this.sortParams;
-    const sortingCallback = this.props.sortable[sortBy] || noop;
+  getSorted(rawItems: ItemObject[]) {
```
jakolehm: This method is now complex enough that we really need tests to confirm it works.

jakolehm (dismissed):

src/renderer/components/table/table.tsx
Nokel81: Yes I guess so, I did this so that it is in line with the other checks (like in `ItemListLayout`.

Nokel81: I don't think that this will break any extension developers though...

nevalla: so previously if extension has used sortable table and not given any table id, sorting has not worked and end user has seen the error in console, now instead of displaying error in console, end user will see error notification?

nevalla (dismissed):
Tested and seemed to work as expected. Left comment about breaking change raised by @jakolehm though

src/renderer/components/table/table.tsx
jakolehm: > now instead of displaying error in console, end user will see error notification?

I thought they will see error/crash page... that's why I was worried about breaking change.

nevalla (dismissed):

src/renderer/components/table/__tests__/getSorted.test.ts
```diff
@@ -0,0 +1,155 @@
...
+
+    it("should sort numerically asc (by defaul) and not touch the original list", () => {
+      const i = [1, 2, 4, 3];
+
+      expect(getSorted(i, v => v, "foobar")).toStrictEqual([1, 2, 3, 4]);
+      expect(i).toStrictEqual([1, 2, 4, 3]);
+    });
+
+    describe("multi-part", () => {
+      it("should treat fewer parts as LESS", () => {
```
jakolehm: I don't completely get why fewer parts should be considered as less. Also kubernetes object names don't have dots, does this also work with `-` as part separator?

src/common/utils/array.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export type Tuple<T, N extends number> = N extends N ? number extends N ? T[] : _TupleOf<T, N, []> : never;
+type _TupleOf<T, N extends number, R extends unknown[]> = R["length"] extends N ? R : _TupleOf<T, N, [T, ...R]>;
+
+/**
+ *
+ * @param collection
+ * @param chunkSize
+ */
+export function* chunkView<T, N extends number>(collection: T[], chunkSize: N): Iterable<Tuple<T, N>> {
```
jakolehm: Was this file added by accident? (it's not used anywhere)

src/renderer/components/table/__tests__/getSorted.test.ts
Nokel81: So the `getSorted` function is agnostic to anything to do with kubernetes. It also does not do any splitting itself. That is the the `v => v.split(".")` in this test.

As for this test..., looking at the actual test case, the declared expected sorted version doesn't even satisfy the test case name.

I think I should fix these.

src/common/utils/array.ts
Nokel81: I think I used this during development but then removed all uses of it. Will remove this file.

src/renderer/utils/index.ts
```diff
@@ -22,19 +22,19 @@
...
 export * from "../../common/utils";
-
-export * from "./cssVar";
-export * from "./cssNames";
 export * from "../../common/event-emitter";
-export * from "./saveFile";
-export * from "./prevDefault";
-export * from "./storageHelper";
+
+
```
nevalla: NIT: redundant empty line here

nevalla (dismissed):
Looks good to me. @jakolehm you left comments previously PTAL also?

nevalla (approved):

PR #3252, opened 2021-07-02
Add Sentry to capture exceptions and console.error messages, add a checkbox in preference to enable/disable error reporting

chenhunghan:
- Add Sentry to capture exceptions and console.error messages in `main`/`renderer` process
- Add `config.sentryDsn` to `package.json`, default to an empty string.
- Initialise Sentry (`Sentry.init()`) if `dsn` is a valid url.
- Add a new `on/off` checkbox to preference and `allowErrorReporting` to `UserStore`, check `allowErrorReporting` in `Sentry.init({ beforeSend })` hook.
Screenshot 2021-07-06 at 14 57 11

- `allowErrorReporting` defaults to `true` if `dsn` is a valid url.

Part of https://github.com/lensapp/lens/issues/3215 (CI part is not in this PR)

PR #3315, opened 2021-07-08
Cleanup sentry integration

Nokel81:
- Don't setup sentry until after UserStore has been initialized. It
  won't send anything and furthermore UserStore.getInstance() will throw

- Remove isValidURL as we already have input_validators.isUrl

- Remove isValidSentryDsn as Sentry will report that for us.
  Furthermore, the user setting can just assume that it being set in the
  package.json is good enough.

- Move to using import() for the sentry parts as it types correctly in
  TS.


src/renderer/components/+preferences/preferences.tsx
```diff
@@ -259,7 +259,7 @@ export class Preferences extends React.Component {
           <section id="telemetry">
             <h2 data-testid="telemetry-header">Telemetry</h2>
             {telemetryExtensions.map(this.renderExtension)}
-            {sentryDsnIsValid ? (
+            {sentryDsn ? (
```
chenhunghan: This is not the same as original intend, `sentryDsn` can be "ftp://whatever` and the preference section still renders.

src/common/sentry.ts
```diff
@@ -21,95 +21,78 @@
...
+  }
 };
 
-export const environment = isProduction ? "production" : "development";
+const environment = isProduction ? "production" : "development";
 
-const logInitError = (reason: string) => {
+function logInitError(reason: string) {
   logger.warn(`⚠️  [SENTRY-INIT]: ${reason}, Sentry is not initialized.`);
-};
```
chenhunghan: I wouldn't change these unless we have linter rule to force using function over const.

src/renderer/components/+preferences/preferences.tsx
Nokel81: Yes, but we set `sentryDsn`. For all intents and purposes it is static. I figured that the purpose of hiding the setting was so that on OpenLens the setting isn't there but on Lens it is there.

src/common/sentry.ts
Nokel81: Okay I can revert. We should have a discussion about some more lint rules

src/common/sentry.ts
```diff
@@ -21,95 +21,78 @@
...
-import { sentryDsn, sentryDsnIsValid, isProduction } from "./vars";
+import { sentryDsn, isProduction } from "./vars";
 import { UserStore } from "./user-store";
 import logger from "../main/logger";
 
 // https://www.electronjs.org/docs/api/process#processtype-readonly
-const electronProcessType = ["browser", "renderer", "worker"] as const;
+type ElectronProcessType = "browser" | "renderer" | "worker";
+
+const processType = process.type as ElectronProcessType;
```
chenhunghan: I would avoid type casting at best

Nokel81: Okay, even you were too? I guess we don't really care about this type cast.

chenhunghan: I am not fully sure, but constant assertion seems to be different than normal type assertion (type casting)?
https://dev.to/mremanuel/short-and-sweet-typescript-const-assertions-1d84

src/common/sentry.ts
```diff
@@ -21,95 +21,78 @@
...
   return null;
-};
-
-export const SentryInit = () => {
-  if (!sentryDsnIsValid) {
-    logInitError("Sentry DSN is not valid");
+}
 
-    return;
+async  function requireSentry() {
```
chenhunghan: Seems to be an extra space?
```suggestion
async function requireSentry() {
```

src/common/sentry.ts
Nokel81: They produce different code. Namely the constant assertion produces an array in the JS which can then be used. Similar to how `enum`'s produce an object in the JS.

However, the previous code was doing a type cast anyway. All type casts in TS are purely a compile time construct even if the type came from a constant assertion.

src/renderer/components/+preferences/preferences.tsx
chenhunghan: There are some discussion in original PR related to this, might provide context why isValidDsn is needed
https://github.com/lensapp/lens/pull/3252#issuecomment-875583696

A good example is our CI in lens-ide is broken, and inject some invalid value in package.json

Nokel81: But if you catch the error, like what it done in this PR then you don't have that `uncaught exception` you have the following instead:
Screen Shot 2021-07-08 at 4 08 44 PM

src/common/sentry.ts
```diff
@@ -21,95 +21,78 @@
...
+      });
+      logger.info(`✔️  [SENTRY-INIT]: Sentry for ${processType} is initialized.`);
+    } catch (error) {
+      return logInitError(`Sentry.init() error ${error?.message}`);
+    }
   } catch (error) {
-    logInitError(`Sentry.init() error ${error?.message}`);
-
-    return;
+    return logInitError(`Error loading Sentry module ${error?.message ?? error}`);
```
chenhunghan: I wouldn't return `logInitError()` here, might not matter here because `logInitError` also doesn't return, but I think it's better to return just what intended to return in the case of early return. (For example, if `logInitError()` returns `string`, `SentryInit` will be void or string, most linter doesn't like that..

Nokel81: I reverted this.

src/common/sentry.ts
```diff
@@ -20,96 +20,68 @@
...
-    }
-
-    // if in renderer process
-    if (processType === "renderer") {
-      Sentry = require("@sentry/electron/dist/renderer");
+    const Sentry = await requireSentry();
+
+    try {
+      Sentry.init({
+        beforeSend: event => {
```
Nokel81, line 54: Note I moved this here because the typings now work and we don't need to import the `SentryEvent` type anymore.

src/common/sentry.ts
```diff
@@ -20,96 +20,68 @@
...
-    if (processType === "browser") {
-      Sentry = require("@sentry/electron/dist/main");
-    }
-
-    // if in renderer process
-    if (processType === "renderer") {
-      Sentry = require("@sentry/electron/dist/renderer");
+    const Sentry = await requireSentry();
+
+    try {
```
Nokel81, line 52: This is a nested try because I prefer `const` over `let`. I really wish that JS had switch as expressions but they are only in the stage 2 for ecmascript proposals.

src/common/sentry.ts
chenhunghan: Nice :)

src/renderer/components/+preferences/preferences.tsx
chenhunghan: Yes, Sentry won't be init, but the UI in preference is still there, and there is no point to show the UI if dsn is not valid.

Please try to change `sentryDsn` in package.json to `http://whatever` and `make dev`
You will still see this in preference
Screenshot 2021-07-08 at 23 20 30

Nokel81: Ah makes sense, I think it would be better to wait until `Sentry.init()` succeeds.

chenhunghan (approved):
LGTM, thanks.

PR #3317, opened 2021-07-08
ignore Lens's prerelease tag when checking extension compatibility

jim-docker:
- ignore Lens's prerelease tag when checking extension compatibility
- assume bundled extensions are compatible

### DIFF
--- a/src/extensions/extension-discovery.ts
+++ b/src/extensions/extension-discovery.ts
@@ -34,9 +34,8 @@ import { extensionInstaller } from "./extension-installer";
 import { ExtensionsStore } from "./extensions-store";
 import { ExtensionLoader } from "./extension-loader";
 import type { LensExtensionId, LensExtensionManifest } from "./lens-extension";
-import semver from "semver";
-import { appSemVer } from "../common/vars";
 import { isProduction } from "../common/vars";
+import { isCompatibleExtension } from "./extension-compatibility";
 
 export interface InstalledExtension {
   id: LensExtensionId;
@@ -362,13 +361,7 @@ export class ExtensionDiscovery extends Singleton {
       const extensionDir = path.dirname(manifestPath);
       const npmPackage = path.join(extensionDir, `${manifest.name}-${manifest.version}.tgz`);
       const absolutePath = (isProduction && await fse.pathExists(npmPackage)) ? npmPackage : extensionDir;
-      let isCompatible = isBundled;
-
-      /* assume bundled extensions are compatibile */
-      if (!isBundled && manifest.engines?.lens) {
-        /* include Lens's prerelease tag in the matching so the extension's compatibility is not limited by it */
-        isCompatible = semver.satisfies(appSemVer, manifest.engines.lens, { includePrerelease: true });
-      }
+      const isCompatible = isBundled || isCompatibleExtension(manifest);
 
       return {
         id,
--- a/src/extensions/extension-discovery.ts
+++ b/src/extensions/extension-discovery.ts
@@ -364,13 +364,12 @@ export class ExtensionDiscovery extends Singleton {
       const absolutePath = (isProduction && await fse.pathExists(npmPackage)) ? npmPackage : extensionDir;
       let isCompatible = isBundled;
 
-      if (manifest.engines?.lens) {
+      /* assume bundled extensions are compatibile */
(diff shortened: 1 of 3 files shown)


src/extensions/extension-discovery.ts
```diff
@@ -364,13 +364,12 @@ export class ExtensionDiscovery extends Singleton {
       const absolutePath = (isProduction && await fse.pathExists(npmPackage)) ? npmPackage : extensionDir;
       let isCompatible = isBundled;
 
-      if (manifest.engines?.lens) {
+      /* assume bundled extensions are compatibile */
+      if (!isBundled && manifest.engines?.lens) {
         const appSemVerLatestImplied = appSemVer;
```
Nokel81: Does this actually copy it, if not why not just clear it directly?

src/extensions/extension-discovery.ts
```diff
@@ -364,13 +364,12 @@ export class ExtensionDiscovery extends Singleton {
...
 
-        if (appSemVerLatestImplied.prerelease?.[0] === "latest") {
-          /* remove the "latest" prerelease tag so as not to require the extension to specify it */
-          appSemVerLatestImplied.prerelease = [];
-        }
+        /* remove any prerelease tag so the extension's compatibility is not limited by it */
+        appSemVerLatestImplied.prerelease = [];
         isCompatible = semver.satisfies(appSemVerLatestImplied, manifest.engines.lens);
```
Nokel81: Therefor this can be simplified to 
```suggestion
        const isCompatible = isBundled || semver.satisfies(appSemVer, manifest.engines.lens);
```

src/extensions/extension-discovery.ts
jakolehm: We probably don't want to mutate `appSemVer` and that's why it's copied.

jakolehm (dismissed):

src/extensions/extension-discovery.ts
Nokel81: But this is JS all we are doing here is copying the object pointer.

example:

```js
const x = {
	foo: [1, 2, 3]
};

console.log(x);

const y = x;

y.foo = [];

console.log(y);
console.log(x);
console.log(y === x);
```

jim-docker: pointer? :)

Yes, javascript trips me up again. I will rework...

Nokel81: Yup pointer, all object references are pointers which are always passed by value

jim-docker: If it's a pointer I should be able to do this:
```
y->foo = [];
```

I'm just kidding, you're talking about pointers in javascript. (I think I know what a reference is :) )

src/extensions/extension-discovery.ts
```diff
@@ -366,11 +366,8 @@ export class ExtensionDiscovery extends Singleton {
 
       /* assume bundled extensions are compatibile */
       if (!isBundled && manifest.engines?.lens) {
-        const appSemVerLatestImplied = appSemVer;
-
-        /* remove any prerelease tag so the extension's compatibility is not limited by it */
-        appSemVerLatestImplied.prerelease = [];
-        isCompatible = semver.satisfies(appSemVerLatestImplied, manifest.engines.lens);
+        /* include Lens's prerelease tag in the matching so the extension's compatibility is not limited by it */
+        isCompatible = semver.satisfies(appSemVer, manifest.engines.lens, { includePrerelease: true });
```
Nokel81: Are you sure that that we want to set `includePrerelease: true` here?

jim-docker: I think so, without it an extension will be deemed incompatible if the Lens version has a tag (and the extension's range does not have a tag, which would normally be the case). With it, Lens versions having a tag will be included in the matching criteria to the extensions range.

Do you see a problem?

Nokel81: nope, maybe this should be factored out into a separate function that we have unit tests for. Maybe something with the signature:

```js
function checkCompatability(manifest: LensExtensionManifest, isBundled: boolean): boolean;
```

I guess it does make sense (a bit) that `includePrerelease: true` for the range `>= 5.0.1` would include `5.0.2-beta.1` but I think tests would still help.

Nokel81 (approved):

PR #3323, opened 2021-07-09
Sync activeEntity to cluster renderer

### LINKED ISSUE #3322: ActiveEntity not available within cluster renderer
`activeEntity` is not properly synced to cluster renderer (frame) which makes it really hard to use for extensions that are only focused on a single cluster.

jakolehm:
Fixes #3322

### DIFF
--- a/src/renderer/components/app.tsx
+++ b/src/renderer/components/app.tsx
@@ -19,7 +19,7 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 import React from "react";
-import { observable, makeObservable } from "mobx";
+import { observable, makeObservable, reaction } from "mobx";
 import { disposeOnUnmount, observer } from "mobx-react";
 import { Redirect, Route, Router, Switch } from "react-router";
 import { history } from "../navigation";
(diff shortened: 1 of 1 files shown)


src/renderer/components/app.tsx
```diff
@@ -86,6 +88,15 @@ export class App extends React.Component {
...
     await requestMain(clusterSetFrameIdHandler, clusterId);
     await getHostedCluster().whenReady; // cluster.activate() is done at this point
+
+    const activeEntityDisposer = reaction(() => catalogEntityRegistry.getById(clusterId), (entity) => {
+      if (!entity) {
+        return;
+      }
+      catalogEntityRegistry.activeEntity = entity;
+      activeEntityDisposer();
+    }, {fireImmediately: true});
```
Nokel81, line 98: Why not just override `activeEntity` with a getter?

```suggestion
    Object.defineProperty(catalogEntityRegistry, "activeEntity", {
      get() {
        return catalogEntityRegistry.getById(clusterId);
      },
    });
```

Nokel81: Actually for that matter, why is it not a getter for that class to begin with?

src/renderer/components/app.tsx
```diff
@@ -86,6 +88,15 @@ export class App extends React.Component {
 
     await requestMain(clusterSetFrameIdHandler, clusterId);
     await getHostedCluster().whenReady; // cluster.activate() is done at this point
+
+    const activeEntityDisposer = reaction(() => catalogEntityRegistry.getById(clusterId), (entity) => {
```
aleksfront, line 92: Maybe more reasonable to move this disposer code into `catalogEntityRegistry.init()` as it relates to registry itself.

aleksfront (approved):

PR #3324, opened 2021-07-09
New spinner for cluster connect status page

jakolehm:
https://user-images.githubusercontent.com/1446224/125062591-3c493600-e0b7-11eb-989a-3f5d28297d88.mov

After this change `CubeSpinner` is not used anywhere and can be removed.


Nokel81 (approved):
Looks good

PR #3325, opened 2021-07-09
Fix Sentry integration bugs

chenhunghan:
- Remove `sentryIsIntitilzied`, only hide UI when dsn is undefined/null/empty string.
- Avoid this uncaught exception by moving `SentryInit()` before `logger.*` in main process.
Screenshot 2021-07-09 at 08 47 28 

- Remove `requireSentry` hack, as we had `nodeIntegration: true` for renderer process, this https://docs.sentry.io/platforms/javascript/guides/electron/#webpack-configuration is not needed.
- After removing `requireSentry`, `Sentry` is not an `async` function. And call be called at the begging of main process to avoid the uncaught exception above.
- Add `try...catch` for `UserStore.getInstance()?.allowErrorReporting`, and set `allowErrorReporting` default to `false`, so `SentryInit()` can be called anywhere, even before `UserStore` having instances.

Recommended tests:
1. Remove `package.config.sentryDsn` https://github.com/lensapp/lens/blob/master/package.json#L53, the preference for auto error reporting shouldn't appear
2. Set `package.config.sentryDsn` to an empty string `""`, the preference for auto error reporting shouldn't appear
3. Set `package.config.sentryDsn` to a valid dsn `"https://673ac8dc2d4649b5bc498639765f995b@o625296.ingest.sentry.io/5753768"`, the preference for auto error reporting should appearer (Preference > Telemetry)
124985454-17b37680-e043-11eb-9347-c7bc2f46579d

4. With valid dsn set, you should be able to toggle on/off in preference, try to `console.error` in developer console, and see if the error message shows in https://sentry.io/organizations/k8slens/issues/?project=5753768 if on.
5. Test switching off auto error reporting, error shouldn't be transport to https://sentry.io/organizations/k8slens/issues/?project=5753768
6. Check if issue in Sentry has correct context, proces = 'main' || 'renderer'.
Screenshot 2021-07-09 at 15 00 58
7. Close Lens, and restart, see if settings are persistent.


src/common/sentry.ts
```diff
@@ -20,73 +20,75 @@
...
 
-          logger.info(`🔒  [SENTRY-BEFORE-SEND-HOOK]: allowErrorReporting: false. Sentry event is caught but not sent to server.`);
-          logger.info("🔒  [SENTRY-BEFORE-SEND-HOOK]: === START OF SENTRY EVENT ===");
-          logger.info(event);
-          logger.info("🔒  [SENTRY-BEFORE-SEND-HOOK]: ===  END OF SENTRY EVENT  ===");
+        try {
+          allowErrorReporting = UserStore.getInstance()?.allowErrorReporting;
+        } catch {
+          // ignore the TypeError throw by UserStore.getInstance()
+        }
```
Nokel81: You can just do `UserStore.getInstance(false)?.allowErrorReporting ?? false;`

src/common/sentry.ts
```diff
@@ -20,73 +20,75 @@
...
+      initialScope: {
+        tags: {
+
+          "process": processName
+        }
+      },
+      environment: isProduction ? "production" : "development",
+    });
+
+    if (sentryDsnIsValid) {
```
Nokel81: If this is only needed for the `""` case can't we just check for that?

src/main/index.ts
```diff
@@ -61,6 +61,10 @@ import { ExtensionsStore } from "../extensions/extensions-store";
 import { FilesystemProvisionerStore } from "./extension-filesystem";
 import { SentryInit } from "../common/sentry";
 
+// This has to be called before start using winton-based logger
+// For example, before any logger.log 
+SentryInit();
```
Nokel81, line 66: Would you be able to expound a bit more as to why this has to be called before using winston?

chenhunghan: There will be an uncaught exception as in the PR description... I don't know why.

Nokel81: Okay sounds good

src/common/sentry.ts
chenhunghan: Yes.. but I guess regex covers more cases and more future proof?

Nokel81: But from looking at the sentry code they already run this regex so how is it more future proof?

chenhunghan: For example, Sentry changes their source code in 7.0 and don't throw error from Sentry.init() when dsn is not valid, we will have side effects for showing the UI even if dsn is not valid after upgrading to 7.0.

Nokel81: What if Sentry changes its format of the dsn in v7? I don't find the argument in a future version they may no longer throw an error persuasive. I actually find the fact that they don't throw an error on `""` strange.

chenhunghan: Ideally they export the regex check for dsn somewhere, but at least I can't find it for now. I don't know why they didn't throw when "". This happened to be our case.

Nokel81: They do report an error to console, I am seeing the following on master (with `sentryDsn: ""` in package.json):

```
SentryError: Attempted to enable Electron native crash reporter but no DSN was supplied
[0]     at new SentryError (/repos/lens/node_modules/@sentry/utils/dist/error.js:9:28)
[0]     at MainBackend.<anonymous> (/repos/lens/node_modules/@sentry/electron/dist/main/backend.js:152:27)
[0]     at step (/repos/lens/node_modules/@sentry/electron/node_modules/tslib/tslib.js:143:27)
[0]     at Object.next (/repos/lens/node_modules/@sentry/electron/node_modules/tslib/tslib.js:124:57)
[0]     at /repos/lens/node_modules/@sentry/electron/node_modules/tslib/tslib.js:117:75
[0]     at new Promise (<anonymous>)
[0]     at Object.__awaiter (/repos/lens/node_modules/@sentry/electron/node_modules/tslib/tslib.js:113:16)
[0]     at MainBackend._installNativeHandler (/repos/lens/node_modules/@sentry/electron/dist/main/backend.js:146:24)
[0]     at new MainBackend (/repos/lens/node_modules/@sentry/electron/dist/main/backend.js:66:34)
[0]     at MainClient.BaseClient (/repos/lens/node_modules/@sentry/core/dist/baseclient.js:52:25) {
[0]   name: 'SentryError'
[0] }
```

chenhunghan: Yes, but it didn't throw error from Sentry.init(), at least the try...catch can't catch it

Nokel81: It is this line https://github.com/getsentry/sentry-electron/blob/edd2775048d4658f88fd23a989574da913e6762a/src/main/backend.ts#L101

Nokel81: I think this is a bug, it should reject the init

chenhunghan: Forget = Consumes the promise and logs the error when it rejects.

https://github.com/getsentry/sentry-javascript/blob/5ed31354/packages/utils/src/async.ts#L6

Nice name...

chenhunghan: I am not sure if it's a bug, if DSN can be set dynamically at runtime, there is no need to reject promise if the initial dsn is not valid.

Nokel81: Yeah... a really bad coding practice.

Nokel81: Submitted https://github.com/getsentry/sentry-electron/issues/348

Nokel81: How would that work, calling `init` again? I cannot find anything in the docs to state that you can pass in a dsn without calling `init`

chenhunghan: Yeah, then it's probably a bug, or bad-by-design

Nokel81: should be easy to fix too.

Nokel81: Submitted fix https://github.com/getsentry/sentry-electron/pull/349

chenhunghan: So what's the plan, we should wait for Sentry to merge your PR and update Sentry libs and remove the "" check in this PR?

Nokel81: I think that would take to long. How about keeping just the `""` check (since we do get errors thrown on invalid URLs of length > 0) and reverting to assigning to a `sentryIsInitialized` variable. 

All the while adding a `// TODO:` that references that PR.

Nokel81: WDYT?

chenhunghan: I am thinking just remove the Sentry init check, and always shows the UI (checkbox) in preference page, for users, it doesn't matter Sentry has initiated or not.

I imagine users only care if they can disable exceptions transporting.

Nokel81: Yes probably, and if it isn't initialized then even if it is true it won't send anything.

jakolehm: > I am thinking just remove the Sentry init check, and always shows the UI (checkbox) in preference page, for users, it doesn't matter Sentry has initiated or not.
> 
> I imagine users only care if they can disable exceptions transporting.

I think we should keep UI hidden if `sentryDsn` is not set (or is empty).

chenhunghan: > > I am thinking just remove the Sentry init check, and always shows the UI (checkbox) in preference page, for users, it doesn't matter Sentry has initiated or not.
> > I imagine users only care if they can disable exceptions transporting.
> 
> I think we should keep UI hidden if `sentryDsn` is not set (or is empty).

Ok, let me summerise a bit...

if we want:

> Keep UI hidden if `sentryDsn` is not set (or is empty).

We have two ways

1. Since `Sentry.init` won't throw exception if DSN is not set (or is empty).
    - Keep `sentryDsnIsValid` and use that to hide UI if false.
    - Remove `sentryIsInitialized` as it will always be true because `Sentry.init()` never throws.

2. Wait for Sentry to merge their PR https://github.com/getsentry/sentry-electron/issues/348, and update Sentry to the version with "throw exception" behaviour
    - Remove `sentryDsnIsValid` as not needed, we can use `Sentry.init`.
    - Keep `sentryIsInitialized` and use `sentryIsInitialized` to hide UI if false.

aleksfront: https://github.com/getsentry/sentry-electron/pulls has only 2 PRs opened by this time, so big chance that the issue will be solved quickly. I'd prefer second variant if it will be resolved within 1-2 d. Make option 1 otherwise (and keep and eye on PR)

src/common/sentry.ts
chenhunghan: You mean removing the try...catch and just use `UserStore.getInstance(false)?.allowErrorReporting ?? false;`?

Just wonder is it possible that `getInstance` is `undefined` or not a function?

src/common/sentry.ts
chenhunghan: a third option: only hide preference UI if sentryDsn is an empty string or undefined/null, remove try...catch for `Sentry.init()` (should be caught by tests instead).  https://github.com/lensapp/lens/pull/3325/commits/ec8c3bc26d2a5ebe0622122bbbec2a3a215bf302

src/common/sentry.ts
Nokel81: It is a static function. Is don't think we need to be that paranoid.

chenhunghan: Ok, edited, thanks for confirmation https://github.com/lensapp/lens/pull/3325/commits/a0296d606deaf297bd0042170f9034009170309a

src/common/sentry.ts
aleksfront: Might be the best one. Is think its rare case to have invalid dsn value whatsoever.

Nokel81 (approved):

PR #3333, opened 2021-07-09
Add ability to trim the value from <Input>

### LINKED ISSUE #3306: Add helm repo should trim input before reporting errors
**Describe the bug**
A custom helm repo name is a DNS name so tailing or leading whitespace is always wrong. It should be trimmed before errors are reported.

**To Reproduce**
1. Go to preferences
2. Go to kubernetes
3. Go to helm repos
4. click add custom repo
5. add name with trailing whitespace
6. see error

**Expected behavior**
Lens should ignore trailing or leading whitespace

**Screenshots**
Screen Shot 2021-07-07 at 9 33 24 AM

Nokel81:
- If set (default = false) then the input will be trimmed of whitespace
  before validation and before submit

- Use trim on all input's that use the systemName validator

fixes #3306

### DIFF
--- a/src/renderer/components/+config-resource-quotas/add-quota-dialog.tsx
+++ b/src/renderer/components/+config-resource-quotas/add-quota-dialog.tsx
@@ -175,6 +175,7 @@ export class AddQuotaDialog extends React.Component<Props> {
               <Input
                 required autoFocus
                 placeholder="ResourceQuota name"
+                trim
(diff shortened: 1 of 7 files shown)


Nokel81:
Another option could be that the default for `trim` is `!multiline`

aleksfront (approved):
LGTM

PR #3334, opened 2021-07-09
Remove conflicting classNames under quota-entries

Nokel81:
- Both flex and inline modify the display rule and inline was winning


jakolehm:
Does this fix a UI bug?

Nokel81:
No, which is why I marked it chore. It seems to be an artifact of the before time.

jakolehm (approved):

PR #3337, opened 2021-07-09
Wait for shell to bring up prompt before sending commands

### LINKED ISSUE #3279: Draining a node opens a terminal but no command is sent
**Describe the bug**
In the Nodes tab of a cluster, when I click the breadcrumbs/3dots next to a node and select "Drain" a new terminal opens, but no command gets sent to that terminal from lens, which means that effectively nothing happens.

**To Reproduce**
1. Go to Nodes tab in a cluster
2. (Optional, works as expected) On a node select the breadcrumbs/3 dots to the far right, and click cordon
3. On a node select the breadcrumbs/3 dots to the far right, and click drain
4. New but empty terminal opens in lens, without a command being sent to that terminal by lens

**Expected behavior**
New terminal opens and a command gets sent, eg `kubectl drain <node>` similar to how Cordon sends `kubectl cordon <node>`.

- Lens Version: doesnt work in 5.0.0-latest.20210629.3, 5.0.2-latest.20210705.2
- OS: OSX
- Installation method (e.g. snap or AppImage in Linux): downloaded .dmg from website

Nokel81:

fixes https://github.com/lensapp/lens/issues/3279

### DIFF
--- a/src/renderer/api/terminal-api.ts
+++ b/src/renderer/api/terminal-api.ts
@@ -19,11 +19,13 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { stringify } from "querystring";
 import { boundMethod, base64, EventEmitter } from "../utils";
 import { WebSocketApi } from "./websocket-api";
 import isEqual from "lodash/isEqual";
 import { isDevelopment } from "../../common/vars";
+import url from "url";
+import { makeObservable, observable } from "mobx";
+import type { ParsedUrlQueryInput } from "querystring";
 
 export enum TerminalChannels {
   STDIN = 0,
@@ -55,7 +57,9 @@ export class TerminalApi extends WebSocketApi {
   protected size: { Width: number; Height: number };
 
   public onReady = new EventEmitter<[]>();
-  public isReady = false;
+  @observable public isReady = false;
+  @observable public rcFinished = false;
+  public readonly url: string;
 
   constructor(protected options: TerminalApiQuery) {
     super({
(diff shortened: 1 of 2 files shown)


src/renderer/api/terminal-api.ts
```diff
@@ -55,42 +57,43 @@ export class TerminalApi extends WebSocketApi {
...
 
-    return `${wss}${hostname}${port}/api?${stringify(query)}`;
+    this.url = url.format({
+      protocol: protocol.includes("https") ? "wss" : "ws",
+      hostname,
+      port,
+      pathname: "/api",
+      query,
+      slashes: true,
+    });
```
Nokel81, line 89: This was done because `getUrl()` both doesn't need to be `async` and because it should never change, so we might as well only do the work once.

src/renderer/api/terminal-api.ts
```diff
@@ -106,6 +109,17 @@ export class TerminalApi extends WebSocketApi {
     this.onReady.removeAllListeners();
   }
 
+  _onRcReady = (data: string) => {
+    if (!data) {
+      return;
+    }
+
+    if (data.match(/\r?\n/) === null && data.endsWith(" ")) {
```
Nokel81: This is the most finicky part, basically I assume that all prompts will have a space and be only a single line.

src/renderer/api/terminal-api.ts
```diff
@@ -55,42 +57,43 @@ export class TerminalApi extends WebSocketApi {
...
 
-  async connect() {
-    const apiUrl = await this.getUrl();
-
+  connect() {
     this.emitStatus("Connecting ...");
     this.onData.addListener(this._onReady, { prepend: true });
-
-    return super.connect(apiUrl);
+    this.onData.addListener(this._onRcReady);
```
aleksfront: Can you make names like `_onRcReady`, `rcIsFinished` more self-explanatory? What is RC here and why we're firing it after on `onData`?

src/renderer/api/terminal-api.ts
aleksfront: Need an explanation what are we doing here. Probably by moving to separate function with explanatory name or just by writing a comment.

src/renderer/components/dock/terminal.store.ts
```diff
@@ -104,18 +105,42 @@ export class TerminalStore extends Singleton {
...
+        Notifications.info("Terminal shell is taking a long time to complete startup. Please check your .rc file.", {
+          timeout: 4_000,
+        });
+      }, 2_500);
+      const notifyVeryLong = setTimeout(() => {
+        rcIsFinished.cancel();
+        Notifications.info("Bypassing shell completion check.", {
+          timeout: 4_000,
+        });
+      }, 7_500);
```
aleksfront: Seems like between 2 timeouts just 5 seconds of difference. Maybe worth combine those 2 notifications into 1?

src/renderer/api/terminal-api.ts
Nokel81: Sounds good, will do both.

src/renderer/components/dock/terminal.store.ts
Nokel81: Sounds good.

src/renderer/api/terminal-api.ts
Nokel81: How about with the name `_onShellRunCommandsFinished` and `shellRunCommandsFinished`?

aleksfront (approved):

PR #3338, opened 2021-07-09
Cleanup <EventEmitter>

Nokel81:

IMO this is a misuse of `.every()` because we don't care about its return value.

### DIFF
--- a/src/common/event-emitter.ts
+++ b/src/common/event-emitter.ts
@@ -33,12 +33,8 @@ export class EventEmitter<D extends [...any[]]> {
 
   addListener(callback: Callback<D>, options: Options = {}) {
     if (options.prepend) {
-      const listeners = [...this.listeners];
-
-      listeners.unshift([callback, options]);
-      this.listeners = new Map(listeners);
-    }
-    else {
+      this.listeners = new Map([[callback, options], ...this.listeners]);
+    } else {
       this.listeners.set(callback, options);
     }
   }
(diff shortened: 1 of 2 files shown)


src/common/event-emitter.ts
```diff
@@ -52,12 +48,14 @@ export class EventEmitter<D extends [...any[]]> {
...
-    [...this.listeners].every(([callback, options]) => {
-      if (options.once) {
+    for (const [callback, { once }] of this.listeners) {
+      if (once) {
         this.removeListener(callback);
       }
 
-      return callback(...data) !== false;
-    });
+      if (callback(...data) === false) {
```
chenhunghan, line 54: This reads like stop calling rest of the callback if one of the callback returns false, doesn't seems be the same before refactoring?

src/common/event-emitter.ts
```diff
@@ -52,12 +48,14 @@ export class EventEmitter<D extends [...any[]]> {
   }
 
   emit(...data: D) {
-    [...this.listeners].every(([callback, options]) => {
```
chenhunghan, line 55: I agree that .every is a misuse but will be great if adding tests to ensure .emit do the same before and after refactoring.

src/common/event-emitter.ts
Nokel81: It is what the previous had, the docs of every say so.

src/common/event-emitter.ts
Nokel81: I will add tests

aleksfront (approved):
LGTM

PR #3339, opened 2021-07-09
Prefer SHELL over PTYSHELL

Nokel81:

At least on macOS the SHELL env var is set instead of the PTYSHELL env var if you change your default shell.

### DIFF
--- a/src/main/shell-session/local-shell-session.ts
+++ b/src/main/shell-session/local-shell-session.ts
@@ -32,12 +32,11 @@ export class LocalShellSession extends ShellSession {
   }
 
   public async open() {
-
     const env = await this.getCachedShellEnv();
-    const shell = env.PTYSHELL;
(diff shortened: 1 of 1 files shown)


src/main/shell-session/local-shell-session.ts
```diff
@@ -32,12 +32,11 @@ export class LocalShellSession extends ShellSession {
   }
 
   public async open() {
-
     const env = await this.getCachedShellEnv();
-    const shell = env.PTYSHELL;
+    const shell = env.SHELL || env.PTYSHELL;
```
jakolehm, line 36: `this.getCachedShellEnv()` is setting `env.PTYSHELL` based on user preferences (with fallback to `env.SHELL`). I don't see any reason to modify it again here. Could you describe what bug this actually fixes?

Nokel81: You are right. I had it set...

PR #3340, opened 2021-07-09
Provide better typing for dock tabs

Nokel81:
- Change name from `IDockTab` to `DockTabDescriptor`

- Make all fields required

- Add new types for the `createTab()` and the `create(Type)Tab()` functions

### DIFF
--- a/src/renderer/components/dock/dock.store.ts
+++ b/src/renderer/components/dock/dock.store.ts
@@ -19,7 +19,7 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import MD5 from "crypto-js/md5";
+import * as uuid from "uuid";
 import { action, computed, IReactionOptions, makeObservable, observable, reaction } from "mobx";
 import { autoBind, createStorage } from "../../utils";
 import throttle from "lodash/throttle";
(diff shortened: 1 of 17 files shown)


src/renderer/components/dock/dock.store.ts
```diff
@@ -35,16 +35,24 @@ export enum TabKind {
   POD_LOGS = "pod-logs",
 }
 
-export interface IDockTab {
-  id?: TabId;
+export interface DockTabDescriptor {
```
aleksfront: I found new name a bit verbose then it should be. Why not use `DockTab`? `Descriptor` part here is like `Items`, `Object` and similar names that can be omitted.

src/renderer/components/dock/dock.store.ts
```diff
@@ -35,16 +35,24 @@ export enum TabKind {
...
+  id: TabId;
   kind: TabKind;
-  title?: string;
-  pinned?: boolean; // not closable
+  title: string;
+  pinned: boolean; // not closable
```
aleksfront: I guess it's good that we'll make all props required.

src/renderer/components/dock/dock.store.ts
```diff
@@ -35,16 +35,24 @@ export enum TabKind {
...
+
+  /**
+   * DockTabData may contain extra data
+   */
+  [key: string]: any;
 }
 
+export type CreateDockTabDescriptor = Partial<Omit<DockTabDescriptor, "kind">> & Pick<DockTabDescriptor, "kind">;
```
aleksfront: Do we really need this new type?

From the quick view the typings a bit confusing too. We omit `kind` field and then picking same field again?

src/renderer/components/dock/dock.store.ts
```diff
@@ -35,16 +35,24 @@ export enum TabKind {
...
+export type SpecializedCreateDockTabDescriptor = Omit<CreateDockTabDescriptor, "kind">;
```
aleksfront: Maybe specified type is better than `Partial<IDockTab>` but it's super verbose here. You can't say what it does and how it differs from `CreateDockTabDescriptor` when founding it somewhere in the code.

aleksfront (changes requested):
If we're improving typings here, maybe we can eliminate new types and use `DockTab` (`DockTabDescriptor`) everywhere. By looking at the code we don't have much places where those `tabParams: Partial<IDockTab> = {}` are used. Wdyt?

src/renderer/components/dock/dock.store.ts
Nokel81: It should throw an error if the `kind` field is specified. That is the desire because the functions that use this type specify the `kind` field.

src/renderer/components/dock/dock.store.ts
Nokel81: Sure I can change the name to `DockTab`

Nokel81: I chose this name because `DockTab` is already a component, I agree `DockTabDescriptor` is a tad long.

aleksfront (approved):

PR #3349, opened 2021-07-11
Fix broken shell when workdir not exists

### LINKED ISSUE #125: Shell not working if working directory not exists.
**Describe the bug**
"Pod shell" and "Node shell" console hangs if terminal working directory not exists.

**To Reproduce**
1. Create "xxx" local directory
2. Go to cluster settings
3. Set "Terminal Working Directory" as "xxx"
4. Delete directory "xxx"
5. Go to nodes or pods. Select any and press "Exec" button
6. Console just hangs on "Connecting..."

**Expected behavior**
Probably it should use some default working directory.
Or show user friendly error to set proper working directory.

- Lens Version: 3.01
- OS: Windows
- Installation method (e.g. snap or AppImage in Linux): Setup.exe

devodev:
- validate shell session cwd exists, otherwise use default: `env.HOME`.

Resolves: https://github.com/lensapp/lens/issues/125

### DIFF
--- a/src/main/shell-session/shell-session.ts
+++ b/src/main/shell-session/shell-session.ts
@@ -19,6 +19,7 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
+import fse from "fs-extra";
 import type { Cluster } from "../cluster";
 import { Kubectl } from "../kubectl";
 import type * as WebSocket from "ws";
@@ -50,9 +51,7 @@ export abstract class ShellSession {
   protected kubectlBinDirP: Promise<string>;
   protected kubeconfigPathP: Promise<string>;
 
-  protected get cwd(): string | undefined {
-    return this.cluster.preferences?.terminalCWD;
-  }
+  protected abstract get cwd(): string | undefined;
 
   constructor(protected websocket: WebSocket, protected cluster: Cluster) {
     this.kubectl = new Kubectl(cluster.version);
@@ -60,10 +59,14 @@ export abstract class ShellSession {
     this.kubectlBinDirP = this.kubectl.binDir();
   }
 
-  open(shell: string, args: string[], env: Record<string, any>): void {
+  protected async open(shell: string, args: string[], env: Record<string, any>) {
+    const cwd = (this.cwd && await fse.pathExists(this.cwd))
+    	? this.cwd
+    	: env.HOME;
+
     this.shellProcess = pty.spawn(shell, args, {
       cols: 80,
-      cwd: this.cwd || env.HOME,
+      cwd,
       env,
       name: "xterm-256color",
       rows: 30,
--- a/src/main/shell-session/local-shell-session.ts
+++ b/src/main/shell-session/local-shell-session.ts
@@ -31,8 +31,11 @@ export class LocalShellSession extends ShellSession {
     return [helmCli.getBinaryDir()];
   }
 
-  public async open() {
+  protected get cwd(): string | undefined {
+    return this.cluster.preferences?.terminalCWD;
+  }
 
+  public async open() {
     const env = await this.getCachedShellEnv();
     const shell = env.PTYSHELL;
     const args = await this.getShellArgs(shell);
--- a/src/main/shell-session/node-shell-session.ts
+++ b/src/main/shell-session/node-shell-session.ts
@@ -32,6 +32,10 @@ export class NodeShellSession extends ShellSession {
   protected podId = `node-shell-${uuid()}`;
   protected kc: KubeConfig;
 
+  protected get cwd(): string | undefined {
+    return undefined;
+  }
+
   constructor(socket: WebSocket, cluster: Cluster, protected nodeName: string) {
     super(socket, cluster);
   }


src/main/utils/shell-env.ts
```diff
@@ -64,3 +64,13 @@ export async function shellEnv(shell?: string, forceRetry = false) : Promise<Env
 
   return envVars;
 }
+
+export function resolveEnv(value: string, env: Record<string, any>): string {
+  // resolves both $VAR and ${VAR} syntax, and makes sure to
+  // ignore malformed syntax, such as begins with digit or lower case characters.
+  return value.replace(/\$\{?([A-Z_][A-Z0-9_]*)(?:(?<=\$\{\1)\})?/g, function (match) {
```
Nokel81: It is incorrect that lowercase values are invalid in the environment.

Furthermore shouldn't this regex has a `^` at the end? Lastly this regex is incorrect as it matches `${A`. Furthermore there is no need for it to be a global regex.

To continue bracket matching is not a regular language and requires at least a context free grammer to properly match.

src/main/shell-session/shell-session.ts
```diff
@@ -50,23 +52,33 @@ export abstract class ShellSession {
...
+    if (value) {
+      const resolvedEnvValue = resolveEnv(value, env);
+
+      if (fs.existsSync(resolvedEnvValue)) {
+        cwd = resolvedEnvValue;
+      }
+    }
+
+    return cwd;
+  }
```
Nokel81: This could be a free function instead.

src/main/utils/shell-env.ts
Nokel81: I think that we should just support `$...` and not `${..}` and this information should be added to the hint of the "Working Directory" setting.

devodev: Agreed, will update, thanks.

src/main/shell-session/shell-session.ts
devodev: Will update.

src/main/utils/shell-env.ts
devodev: I pushed some changes to now only match on simple syntax (`$HOME`, `$var`), and I added some tests to validate the logic.

Also the reason for a global regex and no line start anchor (`^`) is I thought it would be a good idea while at it to support more than one var, although I have no problem restricting it.

Finally, I updated the hint of "Working Directory" setting to mention environment variable expansion.

Nokel81: Will rereview. If we wanted to do some sort of hierarchy then I think we would switch to an array. No need to do weird string parsing.

devodev: Yeah the more I think of it, the less I think its useful. The idea was just to allow arbitrary vars expansion, like: `$HOME/myfolder/$ENV` to resolve to `/home/alice/myfolder/development`.

Let me know if you just want me to update this to only validate that cwd exists instead. Thanks.

Nokel81: After discussing it with some other team members we think that the expansion of vars is a very good idea but probably should be its own PR because it is is a feature enhancement whereas the rest of this PR is a bug fix.

src/main/shell-session/shell-session.ts
```diff
@@ -50,23 +51,19 @@ export abstract class ShellSession {
   protected kubectlBinDirP: Promise<string>;
   protected kubeconfigPathP: Promise<string>;
 
-  protected get cwd(): string | undefined {
```
Nokel81, line 53: Probably making this an `abstract get cwd()` would be a good, then we don't need to change the arguments of `open()`

src/main/utils/shell-env.ts
devodev: Awesome I will do the split and keep only the bug fix here.

I'll open another PR to add var expansion, and we can discuss there the best way to implement, thanks :)

src/main/shell-session/shell-session.ts
devodev: Makes much more sense, will update.

devodev:
I applied the changes requested, and opened a new PR for the env var expansion feature here: https://github.com/lensapp/lens/pull/3386

src/main/shell-session/shell-session.ts
```diff
@@ -61,9 +60,15 @@ export abstract class ShellSession {
   }
 
   open(shell: string, args: string[], env: Record<string, any>): void {
+    let cwd = env.HOME;
+
+    if (this.cwd && fs.existsSync(this.cwd)) {
+      cwd = this.cwd;
+    }
```
Nokel81: Nit, this could be 
```suggestion
    const cwd = (this.cwd && await fse.pathExists(this.cwd)) 
    	? this.cwd 
    	: env.HOME;
```

Also, given that both child classes have an `async open` function. I think it would be fine to change this function to async too. Don't want to use blocking FS calls within an async context.

devodev: Agreed, this is an oversight on my part about the async logic.

And I also prefer the ternary. Will update.

Nokel81 (approved):

PR #3359, opened 2021-07-12
Only create TerminalStore after ThemeStore

Nokel81:
- Use Proxy for backwards compatability

Without this an uncaught exception from mobx happens.

### DIFF
--- a/src/renderer/bootstrap.tsx
+++ b/src/renderer/bootstrap.tsx
@@ -48,6 +48,7 @@ import { ExtensionsStore } from "../extensions/extensions-store";
 import { FilesystemProvisionerStore } from "../main/extension-filesystem";
 import { ThemeStore } from "./theme.store";
 import { SentryInit } from "../common/sentry";
+import { TerminalStore } from "./components/dock/terminal.store";
 
 configurePackages();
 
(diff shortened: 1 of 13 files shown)


src/renderer/bootstrap.tsx
```diff
@@ -89,18 +90,22 @@ export async function bootstrap(App: AppComponent) {
...
-  await ClusterStore.createInstance().loadInitialOnRenderer();
+  const cs = ClusterStore.createInstance();
+
+  await cs.loadInitialOnRenderer();
+
   HotbarStore.createInstance();
   ExtensionsStore.createInstance();
   FilesystemProvisionerStore.createInstance();
   ThemeStore.createInstance();
+  TerminalStore.createInstance();
```
jim-docker, line 107: Maybe a comment indicating that the order is important here?

Nokel81: Good idea.

Nokel81:
@jim-docker I have added comments and fixed all unit tests. Testing a `make dev` build it seems to work!

jim-docker:
> * Use Proxy for backwards compatability
> 
> Signed-off-by: Sebastian Malton [sebastian@malton.name](mailto:sebastian@malton.name)
> 
> Without this an uncaught exception from mobx happens.

Is there an issue? Does it always happen?

Nokel81:
The mobx error is a race condition. To reproduce have an open local terminal session with the dock open and then restart lens.

Nokel81:
@jim-docker This is the error I see 
Screen Shot 2021-07-13 at 12 32 56 PM

jim-docker (approved):
I managed to reproduce and then confirm that this fixes it

PR #3362, opened 2021-07-12
Remove win-ca/pem before publish & catch inject error

### LINKED ISSUE #3357: Installing for all users causes error: JavaScript error occurred in the main process: Uncaught exception: Error: EPERM: operation not permitted
**Describe the bug**

When I open Lens, I get this error: 

JavaScript error occurred in the main process

Uncaught exception: 
Error: EPERM: operation not permitted, open 'C:\Program Files\Lens\resources\app.asar.unpacked\node_modules\win-ca\pem\root.pem'

Error Lens

**To Reproduce**
1. Update to Lens 5.0.2, install for all users
2. Open Lens 
3. See error

**Expected behavior**
1. Update to Lens 5.0.2, install for all users
2. Open Lens
3. Do not see error

**Screenshots**
See attached 

- Lens Version: 5.0.2-latest
- OS: Windows 10 Version 21H1, OS build 19043.1083
- Installation method: Windows install

Nokel81:

fixes https://github.com/lensapp/lens/issues/3357

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -49,7 +49,7 @@ jobs:
         displayName: Customize config
 
       - script: make node_modules
-        displayName: Install dependencies=
+        displayName: Install dependencies
 
       - script: make build-npm
         displayName: Generate npm package
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -49,12 +49,7 @@ jobs:
         displayName: Customize config
 
       - script: make node_modules
-        displayName: Install dependencies
-
-      - base: |
-          rm -rf node_modules/win-ca/pem
-        displayName: Remove win-ca/pem
-        # https://github.com/ukoloff/win-ca#clear-pem-folder-on-publish
+        displayName: Install dependencies=
 
       - script: make build-npm
         displayName: Generate npm package
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -24,12 +24,12 @@ jobs:
           Write-Output ("##vso[task.setvariable variable=CI_BUILD_TAG;]$CI_BUILD_TAG")
         condition: "and(succeeded(), startsWith(variables['Build.SourceBranch'], 'refs/tags/'))"
         displayName: Set the tag name as an environment variable
-      
+
       - task: NodeTool@0
         inputs:
           versionSpec: $(node_version)
         displayName: Install Node.js
-      
+
       - task: Cache@2
         inputs:
           key: 'yarn | "$(Agent.OS)"" | yarn.lock'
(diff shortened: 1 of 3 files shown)


jim-docker:
I still see this:

Screen Shot 2021-07-12 at 7 18 46 PM

I ran:
```
rm -rf node_modules/win-ca/pem
make build
'./dist/OpenLens Setup 5.0.3-beta.1.1626124453802.exe'
```

Oh, maybe `make build reinstalls the pem`

.azure-pipelines.yml
```diff
@@ -47,13 +47,18 @@ jobs:
         env:
           GH_TOKEN: $(LENS_IDE_GH_TOKEN)
         displayName: Customize config
-      
+
       - script: make node_modules
         displayName: Install dependencies
-      
+
+      - base: |
```
jakolehm: Probably would make sense to do this within `make build`:

```Makefile
build: node_modules binaries/client
	yarn run npm:fix-build-version
	$(MAKE) build-extensions -B
	yarn run compile
ifeq "$(DETECTED_OS)" "Windows"
	rm -rf node_modules/win-ca/pem
	yarn run electron-builder --publish onTag --x64 --ia32
else
	yarn run electron-builder --publish onTag
endif
```

Nokel81: Sounds good will change

Nokel81:
@jim-docker PTALA?

src/common/system-ca.ts
```diff
@@ -33,5 +33,9 @@ if (isMac) {
 }
 
 if (isWindows) {
-  winca.inject("+"); // see: https://github.com/ukoloff/win-ca#caveats
+  try {
+    winca.inject("+"); // see: https://github.com/ukoloff/win-ca#caveats
```
jim-docker, line 37: this is documented as experimental

.azure-pipelines.yml
```diff
@@ -47,13 +47,13 @@ jobs:
         env:
           GH_TOKEN: $(LENS_IDE_GH_TOKEN)
         displayName: Customize config
-      
+
       - script: make node_modules
-        displayName: Install dependencies
-      
+        displayName: Install dependencies=
```
jim-docker: what does the `=` do?

Nokel81: oops that was accidental when reverting the previous change.

src/common/system-ca.ts
Nokel81: Do you think we should just not do it instead?

jim-docker: If this fix works without using `"+"` then yes. Otherwise it's risky depending on an experimental feature

Nokel81: I believe that it is a separate concern to this EPERM issue. I added this `try/catch` as just some extra safety.

jim-docker (approved):
No more error notification! Still took a few tries to connect to the cluster (oops warning) but I assume that's another issue.

Nokel81:
yes that is another issue. We need to solve it but in a different PR.

PR #3364, opened 2021-07-13
Make Sentry captures Winston based logger message

chenhunghan:
In addition to `console.error` and `exception`, this PR allows Sentry to capture `logger.error` error messages.

To verify it works as expected, please try `logger.error(any string or object)` in both renderer process and main process.

### DIFF
--- a/src/main/logger.ts
+++ b/src/main/logger.ts
@@ -21,6 +21,7 @@
 
 import { app, remote } from "electron";
 import winston from "winston";
+import * as Sentry from "@sentry/electron";
 import { isDebugging, isTestEnv } from "../common/vars";
 
 const logLevel = process.env.LOG_LEVEL ? process.env.LOG_LEVEL : isDebugging ? "debug" : "info";
@@ -48,4 +49,55 @@ const logger = winston.createLogger({
   ],
 });
 
-export default logger;
+type Logger = ReturnType<typeof winston.createLogger>;
+type LoggerKeys = keyof typeof logger;
+type LoggerError = typeof logger.error;
+
+/**
+ * Type guard to ensure unknown is logger.error function
+ */
+const isLoggerError = (unknown: unknown, key: LoggerKeys): unknown is LoggerError =>
+  typeof unknown === "function" && key === "error";
+
+/**
+ * Proxied version of logger
+ * 
+ * Captures error message using Sentry.captureException(...params) when logger.error(...params)
+ */
+const proxiedLogger: Logger = new Proxy(logger, {
+  get(target: Logger, key: LoggerKeys) {
+    const property = target[key];
+
+    if (isLoggerError(property, key)) {
+      return (...params: Parameters<LoggerError>) => {
+        // do logger.error(...params)
+        property(...params);
+
+        const tags = {
+          process: process.type,
+          logger: "winston"
+        };
+
+        try {
+          const [message, ...extra] = params;
+
+          // toString() is added because Parameters<LoggerError> doesn't seems to work as expected
+          // (infer "message" as an object but should be a string)
+          Sentry.captureMessage(message.toString(), {
+            // need to explicitly assign level because default is Sentry.Severity.Info
+            level: Sentry.Severity.Error,
+            tags,
+            extra: extra ? { ...extra } : null
+          });
+        } catch {
+          // fallback to just captureException(params) (issue will have 'unknown' title in Sentry dashbaird)
+          Sentry.captureException(params, { tags });
+        }
+      };
+    }
+
+    return property;
+  }
+});
+
+export default proxiedLogger;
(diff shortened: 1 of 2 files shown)


chenhunghan:
Not sure why tests failed..

aleksfront (dismissed):

src/main/logger.ts
```diff
@@ -48,4 +49,55 @@ const logger = winston.createLogger({
...
 
-export default logger;
+type Logger = ReturnType<typeof winston.createLogger>;
+type LoggerKeys = keyof Logger;
+type LoggerError = Logger["error"];
+
+/**
+ * Type guard to ensure unknown is logger.error function
+ */
+const isLoggerError = (unknown: unknown, key: LoggerKeys): unknown is LoggerError =>
```
Nokel81: nit using `unknown` as an arg name is confusing

src/main/logger.ts
```diff
@@ -48,4 +49,55 @@ const logger = winston.createLogger({
...
+        } catch {
+          // fallback to just captureException(params) (issue will have 'unknown' title in Sentry dashboard)
+          Sentry.captureException(params, { tags });
+        }
+      };
+    }
+
+    return property;
+  }
+});
```
Nokel81: Is it not possible to just have another transport? This also doesn't work if someone uses `logger.log("error", <message>)`.

Furthermore there already is a Sentry transport: https://github.com/winstonjs/winston/blob/master/docs/transports.md#sentry-transport

Nokel81 (changes requested):

src/main/logger.ts
chenhunghan: I looked `winston-transport-sentry-node` before, but it says its for `@sentry/node` (not support Electron? not sure) , and `winston-transport-sentry-node` doesn't seems to be well-maintained... that was my concerns.

If we want to included `logger.log`, the typeguard can be changed from
```ts
const isLoggerError = (unknown: unknown, key: LoggerKeys): unknown is LoggerError =>
  typeof unknown === "function" && key === "error";
```
to
```ts
const isLoggerErrorAndLog = (unknown: unknown, key: LoggerKeys): unknown is LoggerError =>
  typeof unknown === "function" && (key === "error" || key === "log" );
```
but sending logger.log to Sentry is a bit too verbose for us IMHO.

Nokel81: I agree, I mostly was wondering why we couldn't just make a new transport instead of this proxy business.

chenhunghan: Mostly because I don't have knowledge of how to build a Winston transport... or not in 2 days.

Nokel81: https://github.com/winstonjs/winston-transport#readme seems to make it pretty easy.

chenhunghan: Ok, if transport is better, I can take a look. But why's Proxy not recommended? We are using mobx, which is basically a huge Proxy so I thought Proxy is acceptable.

Nokel81: I view proxy as a last resort, especially in this case where we are already given a way to inject a handler. 

Yes, mobx is proxy but it needs to be to do all that it does.

chenhunghan: Well, Proxy is just another JavaScript feature, we don't need Babel or Typescript to make it works, and we are using Proxy as it's designed to be used.

Transport api might change, Winston might dead one day (in fact, looks like Winston is in a stale state for about 1 year already...) but Proxy will continue working as expected even if we change the underlying logger to... maybe pino.

Nokel81: How you do know that if/when we change loggers the new API won't be different? I am not really convinced we should use a "mixin" feature when we are given an API that works.

src/main/logger.ts
chenhunghan: Fixed in https://github.com/lensapp/lens/pull/3364/commits/146891e4315bf5315745753f37101a769009bbbd

src/main/logger.ts
chenhunghan: Not sure I understand correctly, most of the logger I have used has similar APIs like `logger.error`, and we can keep the same "Proxy as transport" code when changing logger, for example, if we want to change logger to `pino`

We only need to change
```ts
const logger = winston.createLogger();
type Logger = ReturnType<typeof winston.createLogger>;
```

to
```ts
const logger = pino();
type Logger = ReturnType<typeof pino>;
```

If we want to use NestJS logger

```ts
import { Logger as NestLogger } from "@nestjs/common";
const logger = new NestLogger();
type Logger = InstanceType<typeof NestLogger>;
```

If we want to use good old console

```ts
const logger = console;
type Logger = typeof console;
```

We can use the same Proxy as the "transport" for each logger. IMHO Proxy is more generic than building transports for each logger.

PR #3369, opened 2021-07-13
Removing limits and requests from Pod metrics

### LINKED ISSUE #2951: Pod Summary Metrics vs Limit are confusing with 1 limited container, 1 unlimited container
**Describe the bug**
When a pod has multiple containers and one container has a limit or request while another does not, the summary graph on top of the pod flyout shows the limit/request values where usage can be substantially above limit.

NOTE: This also happens on the StatefulSet graph (and I'm assuming Deployments, DaemonSets, ReplicaSets, Jobs but I haven't verified those)

**To Reproduce**
1. Deploy a pod with multiple containers, 1 with a limit and 1 without
2. Enable lens-metrics
3. View pod's details

The [prometheus kube stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack) is a good helm chart as an example of this, you can install with default options to see the behavior mentioned here.  The prometheus stateful set has a prometheus container with no limits/requests and a config-reloader with limits/requests of 0.1 CPU and 50Mi memory.  So you see usages like the screenshot below of 800Mi memory usage with a 50Mi limit.

**Expected behavior**
I'd think that for any graph that shows limits/requests as a sum of several children, if any child has a null value for limit or request then the result would be a null limit/request so usage would be displayed but the null limit/request would not.

I could also see showing the limit/request values on the summary graph with an asterisk in the popup and a callout under the graph like `* - This pod contains a mix of containers with and without limits/requests`.  Personally I think that gets a bit noisy but it would at least make the situation more obvious to the user and wouldn't lose the limit/request information.

This also seems like one of those things where other people might think of this differently so I'm fine if this gets closed as rejected, but it felt a bit off to me so I just wanted to get some discussion on it.

**Screenshots**
image

- Lens Version: 4.2.4
- OS: Windows

aleksfront:
Remove limits/requests from pod overview (those are really confusing and basically misleading information)

Extracting code from mixed PR #1944 

Fixes #2951 

### DIFF
--- a/src/renderer/components/+workloads-pods/container-charts.tsx
+++ b/src/renderer/components/+workloads-pods/container-charts.tsx
@@ -27,6 +27,7 @@ import { isMetricsEmpty, normalizeMetrics } from "../../api/endpoints/metrics.ap
 import { NoMetrics } from "../resource-metrics/no-metrics";
 import { IResourceMetricsValue, ResourceMetricsContext } from "../resource-metrics";
 import { ThemeStore } from "../../theme.store";
+import { mapValues } from "lodash";
 
 type IContext = IResourceMetricsValue<any, { metrics: IPodMetrics }>;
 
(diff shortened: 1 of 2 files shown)


Nokel81 (approved):

Nokel81:
@aleksfront does this fix #2951 ?

aleksfront:
> @aleksfront does this fix #2951 ?

Yes, updated the description.

PR #3371, opened 2021-07-13
remove unused filenamify package

jim-docker:
instead of #3343 


aleksfront (approved):
Good catch!

PR #3377, opened 2021-07-14
Catalog Helm repositories

jakolehm:
image


Nokel81:
I know this is still a draft, but do we actually want all possible helm repos (even those which have not been locally added) to be part of the catalog? Wouldn't it make more sense to have only those repos that have been locally added be part of the catalog?

Nokel81:
Closing this as it is quite out of date.

PR #3379, opened 2021-07-14
Make sure that there are only ever 12 slots in a hotbar

### LINKED ISSUE #3255: on 12 clusters listed in hotbar
**Describe the bug**
adding a 13 cluster to the hotbar will do nothing except show a greyed out box as tho it was waiting for the user to add a cluster

**To Reproduce**

**Expected behavior**
clusters keep getting added. maybe the pagination works even

**Screenshots**
Screen Shot 2021-07-02 at 12 45 03 PM

- Lens Version: 5.0
- OS: [e.g. OSX] osx 10.15.7

**Logs:**
```
Your logs go here...
```

**Kubeconfig:**
```
your kubeconfig here
```

**Additional context**

Nokel81:
- Send notification if hotbar is full

- Filter out on every fromStore call

fixes #3255 

Screen Shot 2021-07-14 at 12 15 21 PM

### DIFF
--- a/src/common/hotbar-store.ts
+++ b/src/common/hotbar-store.ts
@@ -69,6 +69,42 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
     return this.hotbarIndex(this.activeHotbarId);
   }
 
+  @action
+  protected fromStore(data: Partial<HotbarStoreModel> = {}) {
+    if (!data.hotbars || !data.hotbars.length) {
+      const hotbar = getEmptyHotbar("Default");
+      const { metadata: { uid, name, source } } = catalogEntity;
+      const initialItem = { entity: { uid, name, source } };
+
+      hotbar.items[0] = initialItem;
+
+      this.hotbars = [hotbar];
+    } else {
+      this.hotbars = data.hotbars;
+    }
+
+    this.hotbars.forEach(ensureExactHotbarItemLength);
+
+    if (data.activeHotbarId) {
+      if (this.getById(data.activeHotbarId)) {
+        this.activeHotbarId = data.activeHotbarId;
+      }
+    }
+
+    if (!this.activeHotbarId) {
+      this.activeHotbarId = this.hotbars[0].id;
+    }
+  }
+
+  toJSON(): HotbarStoreModel {
+    const model: HotbarStoreModel = {
+      hotbars: this.hotbars,
+      activeHotbarId: this.activeHotbarId
+    };
+
+    return toJS(model);
+  }
+
   getActive() {
     return this.getById(this.activeHotbarId);
   }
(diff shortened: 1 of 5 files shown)


src/common/hotbar-store.ts
```diff
@@ -93,29 +94,6 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
     return [...Array.from(Array(defaultHotbarCells).fill(null))];
   }
 
-  @action
```
Nokel81: I moved this down so that it would be closer to `toJSON`

src/common/hotbar-store.ts
```diff
@@ -175,18 +153,24 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
...
-      const emptyCellIndex = hotbar.items.findIndex(isNull);
+      const emptyCellIndex = hotbar.items.indexOf(null);
 
       if (emptyCellIndex != -1) {
         hotbar.items[emptyCellIndex] = newItem;
       } else {
-        // Add new item to the end of list
-        hotbar.items.push(newItem);
+        if (ipcRenderer) {
+          import("../renderer/components/notifications")
```
Nokel81: This is needed because it doesn't exist on `main`. I really wish we would split these into main and renderer sides.

jakolehm: I think we should just broadcast a message and act on that within renderer. Would keep store cleaner.

Nokel81: Okay, it just seems like hotbar-store is really a renderer-only thing but I can change it.

src/common/hotbar-store.ts
```diff
@@ -175,18 +153,19 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
...
 
       if (emptyCellIndex != -1) {
         hotbar.items[emptyCellIndex] = newItem;
       } else {
-        // Add new item to the end of list
-        hotbar.items.push(newItem);
+        broadcastMessage(HotbarTooManyItems);
       }
-    } else {
+    } else if (0 <= cellIndex && cellIndex <= hotbar.items.length) {
```
jim-docker: Shouldn't it be
```suggestion
    } else if (0 <= cellIndex && cellIndex < hotbar.items.length) {
```

Nokel81: yes it should

src/common/hotbar-store.ts
```diff
@@ -175,18 +153,19 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
...
 
       if (emptyCellIndex != -1) {
         hotbar.items[emptyCellIndex] = newItem;
       } else {
-        // Add new item to the end of list
-        hotbar.items.push(newItem);
+        broadcastMessage(HotbarTooManyItems);
       }
-    } else {
+    } else if (0 <= cellIndex && cellIndex < hotbar.items.length) {
```
jakolehm, line 187: Would be nice to see tests for this.

chenhunghan:
On this branch, I got this when trying to pin `WebLink: Kubernetes Documentation` to hotbar
Screenshot 2021-08-02 at 15 16 02

src/common/hotbar-store.ts
```diff
@@ -163,39 +141,52 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
...
+    if (typeof uid !== "string") {
+      throw new TypeError("CatalogEntity.getId() must return a string");
+    }
+
+    if (typeof name !== "string") {
+      throw new TypeError("CatalogEntity.getName() must return a string");
+    }
+
     const newItem = { entity: {
-      uid: item.metadata.uid,
```
aleksfront, line 145: Not all entities currently have `getItem()` method, so `safeCall` won't help here. And that's why @chenhunghan has an error when trying to pin some of the Links to hotbar.

aleksfront: Actually, **none** of the entities can be pinned to Hotbar. 🐞

chenhunghan (approved):
Tested locally and confirm the TypeError when adding items to catalog is gone.

PR #3385, opened 2021-07-14
Treat default weblinks same as any other weblinks

### LINKED ISSUE #3354: Delete default weblinks in catalog
Be able to delete default weblinks in catalog in Lens 5

Nokel81:

fixes #3354

### DIFF
--- a/src/main/catalog-sources/weblinks.ts
+++ b/src/main/catalog-sources/weblinks.ts
@@ -19,43 +19,19 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { observable, reaction } from "mobx";
+import { computed, observable, reaction } from "mobx";
 import { WeblinkStore } from "../../common/weblink-store";
 import { WebLink } from "../../common/catalog-entities";
 import { catalogEntityRegistry } from "../catalog";
 import got from "got";
-import logger from "../logger";
-import { docsUrl, slackUrl } from "../../common/vars";
-
-const defaultLinks = [
-  { title: "Lens Website", url: "https://k8slens.dev" },
-  { title: "Lens Documentation", url: docsUrl },
-  { title: "Lens Community Slack", url: slackUrl },
-  { title: "Kubernetes Documentation", url: "https://kubernetes.io/docs/home/" },
-  { title: "Lens on Twitter", url: "https://twitter.com/k8slens" },
-  { title: "Lens Official Blog", url: "https://medium.com/k8slens" }
-].map((link) => (
-  new WebLink({
-    metadata: {
-      uid: link.url,
-      name: link.title,
-      source: "app",
-      labels: {}
-    },
-    spec: {
-      url: link.url
-    },
-    status: {
-      phase: "available",
-      active: true
-    }
-  })
-));
+import type { Disposer } from "../../common/utils";
+import { random } from "lodash";
 
 async function validateLink(link: WebLink) {
   try {
     const response = await got.get(link.spec.url, {
-      throwHttpErrors: false
+      throwHttpErrors: false,
+      timeout: 20_000,
     });
 
     if (response.statusCode >= 200 && response.statusCode < 500) {
@@ -63,7 +39,7 @@ async function validateLink(link: WebLink) {
     } else {
       link.status.phase = "unavailable";
     }
-  } catch(error) {
+  } catch {
     link.status.phase = "unavailable";
   }
 }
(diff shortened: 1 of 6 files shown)


src/main/catalog-sources/weblinks.ts
```diff
@@ -19,84 +19,75 @@
...
-      status: {
-        phase: "available",
-        active: true
-      }
-    })));
-    weblinks.push(...defaultLinks);
+    const seenWeblinks = new Set<string>();
+
+    for (const weblinkData of links) {
+      seenWeblinks.add(weblinkData.id);
```
jim-docker, line 70: looks like you're not actually doing anything with this

src/migrations/weblinks-store/5.1.1-beta.1.ts
```diff
@@ -0,0 +1,42 @@
...
+import { docsUrl, slackUrl } from "../../common/vars";
+import type { WeblinkData } from "../../common/weblink-store";
+import type { MigrationDeclaration } from "../helpers";
+
+export default {
+  version: "5.1.1-beta.1",
+  run(store) {
+    const weblinks: WeblinkData[] = store.get("weblinks");
+
+    weblinks.push(
```
jim-docker: weblinks can be undefined

jim-docker (commented):
I get this for `make dev`:
```
[0] Running 5.1.1-beta.1 migration for /Users/jimehrismann/Library/Application Support/OpenLensDev/lens-cluster-store.json
[0] info: [STORE]: LOADED from /Users/jimehrismann/Library/Application Support/OpenLensDev/lens-cluster-store.json
[0] info: [STORE]: LOADED from /Users/jimehrismann/Library/Application Support/OpenLensDev/lens-hotbar-store.json
[0] info: [STORE]: LOADED from /Users/jimehrismann/Library/Application Support/OpenLensDev/lens-extensions.json
[0] info: [STORE]: LOADED from /Users/jimehrismann/Library/Application Support/OpenLensDev/lens-filesystem-provisioner-store.json
[0] Running 5.1.1-beta.1 migration for /Users/jimehrismann/Library/Application Support/OpenLensDev/lens-weblink-store.json
[0] (node:9054) UnhandledPromiseRejectionWarning: Error: Something went wrong during the migration! Changes applied to the store until this failed migration will be restored. TypeError: Cannot read property 'push' of undefined
[0]     at Conf._migrate (/Users/jimehrismann/git/lensapp/lens/node_modules/conf/dist/source/index.js:407:23)
[0]     at new Conf (/Users/jimehrismann/git/lensapp/lens/node_modules/conf/dist/source/index.js:144:18)
[0]     at WeblinkStore.load (webpack-internal:///./src/common/base-store.ts:74:28)
[0]     at new WeblinkStore (webpack-internal:///./src/common/weblink-store.ts:62:14)
[0]     at Function.createInstance (webpack-internal:///./src/common/utils/singleton.ts:48:47)
[0]     at App.eval (webpack-internal:///./src/main/index.ts:175:73)
[0]     at processTicksAndRejections (internal/process/task_queues.js:94:5)
[0] (node:9054) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 8)
[0] (node:9054) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
It was "migrating" from master (I pulled master and ran `make dev` before pulling this branch)

Nokel81:
That is probably the "weblinks" was undefined.

jim-docker:
> That is probably the "weblinks" was undefined.

yes, but shouldn't there have been the default weblinks in the store?

Nokel81:
@jim-docker That is what this PR is trying to implement. There weren't any default weblinks in the store.

jim-docker:
> @jim-docker That is what this PR is trying to implement. There weren't any default weblinks in the store.

So you didn't test this?

Nokel81:
I did, but I guess I didn't test it enough.

Nokel81:
Namely, if I already had ever added a weblink that issue doesn't appear.

src/main/catalog-sources/weblinks.ts
```diff
@@ -19,84 +19,83 @@
...
   }
 }
 
 
 export function syncWeblinks() {
   const weblinkStore = WeblinkStore.getInstance();
-  const weblinks = observable.array(defaultLinks);
+  const webLinkEntities = observable.map<string, [WebLink, Disposer]>();
+
+  function periodicallyCheckLink(link: WebLink): Disposer {
```
jim-docker, line 52: This looks scary to me. A timer for each weblink? An extension could programmatically generate a lot of links? I think this is outside the scope of the issue anyway

src/migrations/weblinks-store/5.1.1-beta.1.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import { docsUrl, slackUrl } from "../../common/vars";
+import type { WeblinkData } from "../../common/weblink-store";
+import type { MigrationDeclaration } from "../helpers";
+
+export default {
+  version: "5.1.1-beta.1",
+  run(store) {
+    const weblinksRaw: any = store.get("weblinks");
+    const weblinks = (Array.isArray(weblinksRaw) ? weblinksRaw : []) as WeblinkData[];
```
jim-docker: seems like it could be more concise?

Nokel81: It could be. All depends on how data paranoid we want to be. In theory I could even do some validation of the contents of that array. But I felt that throwing away items that were clearly wrong (like strings) was good enough.

src/main/catalog-sources/weblinks.ts
Nokel81: By "outside the scope" do you mean you think I should revert this and only check each link once?

jim-docker: I guess you want `setInterval`?
I think I would revert it, seems more of a feature. They are running all the time? Or just when the catalog is displayed?

Nokel81: I did mean `setInterval` yes. They would be running all the time, I guess I could make it so that there is an initial spread so that there is some smoothness to the checking (and not spikes every 10 minutes).

I would still say it is a bug because the catalog displays the phase and it should show as "unavailable" if it is actually unavailable.

jim-docker: How often do weblinks become unavailable? Maybe I'm missing something but I just don't see the value in using cycles to track this.

Nokel81: Anytime the website goes down. it would become unavailable.

Nokel81: @jakolehm Should we check them periodically or only once per set change (which is what it was before).

Nokel81: Jari says that once initially and then every hour or so.

jim-docker (approved):
The extraneous hotbar field in lens-weblink-store.json should probably be cleaned up in another PR?

Nokel81:
@jim-docker Yes I think it should, probably by fixing `BaseStore` itself.

PR #3386, opened 2021-07-15
Expand env vars in cwd passed to shell sessions

devodev:
- A proposal to add environment variable expansion capability for cwd passed to shell sessions.

Signed-off-by: devodev <abalexandrebarone@gmail.com>

### DIFF
--- a/src/common/utils/paths.ts
+++ b/src/common/utils/paths.ts
@@ -22,9 +22,9 @@
 import path from "path";
 import os from "os";
 
-function resolveTilde(filePath: string) {
+export function resolveTilde(filePath: string, home?: string) {
   if (filePath[0] === "~" && (filePath[1] === "/" || filePath.length === 1)) {
-    return filePath.replace("~", os.homedir());
+    return filePath.replace("~", home || os.homedir());
   }
 
   return filePath;
--- a/src/main/__test__/shell-session.test.ts
+++ b/src/main/__test__/shell-session.test.ts
@@ -24,6 +24,7 @@
  */
 
 import { clearKubeconfigEnvVars } from "../utils/clear-kube-env-vars";
+import { resolveEnv } from "../utils/shell-env";
 
 describe("clearKubeconfigEnvVars tests", () => {
   it("should not touch non kubeconfig keys", () => {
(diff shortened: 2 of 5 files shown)


Nokel81:
Will review this once https://github.com/lensapp/lens/pull/3349 is merged so that the diff is correct.

src/main/shell-session/shell-session.ts
```diff
@@ -60,16 +60,22 @@ export abstract class ShellSession {
   }
 
   protected async open(shell: string, args: string[], env: Record<string, any>) {
-    const cwd = (this.cwd && await fse.pathExists(this.cwd))
-    	? this.cwd
-    	: env.HOME;
```
Nokel81, line 65: Did you mean to revert this?

devodev: Yes as I was not sure where to insert the resolveEnv function. Would you prefer running it using the cwd computed var directly when passing it to the shell func?

Like that ?

```
    const cwd = (this.cwd && await fse.pathExists(this.cwd))
    	? this.cwd
    	: env.HOME;

    this.shellProcess = pty.spawn(shell, args, {
      cols: 80,
      cwd: resolveEnv(cwd, env),
      env,
      name: "xterm-256color",
      rows: 30,
    });
    this.running = true;
```

Its mostly to prevent having to run it when we dont have to (on env.HOME) but your call :)

src/main/__test__/shell-session.test.ts
```diff
@@ -38,3 +39,24 @@ describe("clearKubeconfigEnvVars tests", () => {
     expect(clearKubeconfigEnvVars({ a: 1, kubeconfig: "1", kUbeconfig: "1" })).toStrictEqual({ a: 1 });
   });
 });
+
+describe("resolveEnv tests", () => {
```
Nokel81, line 43: Some extra test cases that I think we should cover:

1. `~/foo` should resolve to be the same as `$HOME/foo`. I would use our `resolvePath` function in `utils/paths`


I think for a first pass this should be good enough. Maybe in the future we will support `${foo:-default}` but for the time being this is good.

devodev: I like that, will update.

src/main/shell-session/shell-session.ts
Nokel81: Why not have it `resolveEnd(this.cwd)`?

devodev: Because it would need to be run both with pathExists and on assignment, which I tried to avoid work duplication.

Nokel81: okay cool

src/common/utils/paths.ts
```diff
@@ -22,9 +22,9 @@
 import path from "path";
 import os from "os";
 
-function resolveTilde(filePath: string) {
+export function resolveTilde(filePath: string, home?: string) {
```
devodev, line 25: I take home as optional param because on windows, `os.homedir()` returns `$USERPROFILE` instead of `$HOME`.

I wonder if we should instead update `env.HOME` to `os.homedir()` in `shell.open` and update the hint to print `$USERPROFILE` when on windows? I don't know if these two could differ?

Please advise 😃

src/main/__test__/shell-session.test.ts
devodev: I tried using `resolvePath`, but it seems to prepend the result with the CWD whenever the path is not absolute, which can occur when var expansion fails.

```
resolveEnv tests › should not resolve invalid var (starts with digit)
    Expected: "$0HOME"
    Received: "C:\\dev\\github.com\\devodev\\lens\\$0HOME"
```

I assume this will still fail the `pathExists()` test though, so might be enough?

src/common/utils/paths.ts
Nokel81: So I would just use `%USERPROFILE%` (for `~`) as that is guaranteed to exist and seems to be what powershell does. If it doesn't work for a user then they can just use `%HOME%` themselves.

Nokel81:
Closing this as it has conflicts and hasn't been touched in over a year. Please feel free to open this again if you so desire.

PR #3391, opened 2021-07-15
Use declaritive system for creating singletons

Nokel81:
- Let the code say the requirements of each singleton is instead of
  comments

### DIFF
--- a/src/common/utils/singleton.ts
+++ b/src/common/utils/singleton.ts
@@ -19,6 +19,7 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
+
 type StaticThis<T, R extends any[]> = { new(...args: R): T };
 
 export class Singleton {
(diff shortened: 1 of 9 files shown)


jakolehm:
If this does not fix a bug I would move it to next minor version.

src/common/utils/singleton.ts
```diff
@@ -88,3 +88,94 @@ export class Singleton {
...
+
+interface CreateRequirements<T extends typeof Singleton> {
+  builds: T,
+  args?: ConstructorParameters<T> | (() => ConstructorParameters<T>),
+  requires?: Iterable<typeof Singleton>,
+}
+
+interface DeclareOpts {
+  requires?: Iterable<typeof Singleton>,
+
```
panuhorsmalahti, line 98: [nit] extra empty line

src/common/utils/singleton.ts
```diff
@@ -88,3 +88,94 @@ export class Singleton {
...
+}
+
+/**
+ * A builder class for declaring the dependency graph of a set of singletons
+ *
+ * This is useful because it allows the code to declare the requirements
+ */
+export class CreateSingletons {
+  rawRequirements: CreateRequirements<typeof Singleton>[] = [];
```
panuhorsmalahti: This should probably be marked as private?

src/renderer/bootstrap.tsx
```diff
@@ -84,40 +84,42 @@ export async function bootstrap(App: AppComponent) {
...
 
   // Register additional store listeners
-  cs.registerIpcListener();
+  ClusterStore.getInstance().registerIpcListener();
 
   // init app's dependencies if any
-  if (App.init) {
-    await App.init();
-  }
+  await App.init?.();
```
panuhorsmalahti: I don't think change improves readability

panuhorsmalahti (dismissed):

src/common/utils/singleton.ts
Nokel81: Yes

src/renderer/bootstrap.tsx
Nokel81: Okay I'll revert.

Nokel81:
Closing because we will probably be moving to DI

PR #3397, opened 2021-07-15
Only warn if an extension fails to extend an undefined value

### LINKED ISSUE #3396: [EXTENSIONS-LOADER]: can't load renderer for "example-extension": TypeError: Class extends value ...
Sentry Issue: [K8SLENS-BT](https://sentry.io/organizations/k8slens/issues/2514446191/?referrer=github_integration)

```
[EXTENSIONS-LOADER]: can't load renderer for "example-extension": TypeError: Class extends value undefined is not a constructor or null
    at Object.<anonymous> (C:\Users\ntshoe_i\AppData\Local\Programs\kontena-lens\resources\extensions\example-exte...
```

Nokel81:

fixes #3396

### DIFF
--- a/src/extensions/extension-compatibility.ts
+++ b/src/extensions/extension-compatibility.ts
@@ -23,8 +23,12 @@ import semver from "semver";
 import { appSemVer } from "../common/vars";
 import type { LensExtensionManifest } from "./lens-extension";
 
-export function isCompatibleExtension(manifest: LensExtensionManifest): boolean {
+export function isCompatibleExtension(manifest: LensExtensionManifest, isBundled: boolean): boolean {
   if (manifest.engines?.lens) {
+    if (isBundled) {
+      return true;
+    }
+
     /* include Lens's prerelease tag in the matching so the extension's compatibility is not limited by it */
     return semver.satisfies(appSemVer, manifest.engines.lens, { includePrerelease: true });
   }
--- a/src/extensions/extension-compatibility.ts
+++ b/src/extensions/extension-compatibility.ts
@@ -25,6 +25,7 @@ import type { LensExtensionManifest } from "./lens-extension";
 
 export function isCompatibleExtension(manifest: LensExtensionManifest, isBundled: boolean): boolean {
   if (manifest.engines?.lens) {
+    // We only started having bundled extensions specify the engine in v5
     if (isBundled) {
       return true;
     }
--- a/src/extensions/extension-loader.ts
+++ b/src/extensions/extension-loader.ts
@@ -372,8 +372,14 @@ export class ExtensionLoader extends Singleton {
     } catch (error) {
       const safeName = inspect(extension.manifest.name, false, null, false);
 
-      if (error instanceof TypeError && error.message.match(/^Class extends value .* is not a constructor or null$/) !== null) {
-        console.warn(`${logModule}: ${entryPointName} for ${safeName} does not support ${packageJson.version}`);
+      if (isExtensionIncompatibleError(error)) {
+        const message = `${logModule}: ${entryPointName} for ${safeName} does not support ${packageJson.version}`;
+
+        if (ipcRenderer) {
+          console.warn(message, extension);
+        } else {
+          logger.warn(message, { extension });
+        }
       } else {
         const message = `${logModule}: can't load ${entryPointName} for ${safeName}`;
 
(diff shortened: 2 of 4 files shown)


src/extensions/extension-loader.ts
```diff
@@ -368,10 +370,18 @@ export class ExtensionLoader extends Singleton {
     try {
       return __non_webpack_require__(extAbsolutePath).default;
     } catch (error) {
-      if (ipcRenderer) {
-        console.error(`${logModule}: can't load ${entryPointName} for "${extension.manifest.name}": ${error.stack || error}`, extension);
+      const safeName = inspect(extension.manifest.name, false, null, false);
+
+      if (error instanceof TypeError && error.message.match(/^Class extends value .* is not a constructor or null$/) !== null) {
+        console.warn(`${logModule}: ${entryPointName} for ${safeName} does not support ${packageJson.version}`);
```
jim-docker: no need to test for `ipcRenderer`?

src/extensions/extension-loader.ts
```diff
@@ -368,10 +370,18 @@ export class ExtensionLoader extends Singleton {
     try {
       return __non_webpack_require__(extAbsolutePath).default;
     } catch (error) {
-      if (ipcRenderer) {
-        console.error(`${logModule}: can't load ${entryPointName} for "${extension.manifest.name}": ${error.stack || error}`, extension);
+      const safeName = inspect(extension.manifest.name, false, null, false);
+
+      if (error instanceof TypeError && error.message.match(/^Class extends value .* is not a constructor or null$/) !== null) {
```
jim-docker: not pretty :(

jim-docker (commented):
Does it fix this kind too: https://sentry.io/organizations/k8slens/issues/2514638720/?project=5753768&query=is%3Aunresolved

src/extensions/extension-loader.ts
Nokel81: Good point.

src/extensions/extension-loader.ts
Nokel81: I can try and making it better.

Nokel81:
> Does it fix this kind too: [sentry.io/organizations/k8slens/issues/2514638720/?project=5753768&query=is%3Aunresolved](https://sentry.io/organizations/k8slens/issues/2514638720/?project=5753768&query=is%3Aunresolved)

No, and I think I might try

src/extensions/extension-compatibility.ts
```diff
@@ -23,8 +23,12 @@ import semver from "semver";
...
 
-export function isCompatibleExtension(manifest: LensExtensionManifest): boolean {
+export function isCompatibleExtension(manifest: LensExtensionManifest, isBundled: boolean): boolean {
   if (manifest.engines?.lens) {
+    if (isBundled) {
+      return true;
+    }
+
     /* include Lens's prerelease tag in the matching so the extension's compatibility is not limited by it */
     return semver.satisfies(appSemVer, manifest.engines.lens, { includePrerelease: true });
```
Nokel81, line 34: @jim-docker Is this still needed after your fix of the semver thing? can we just treat bundled extensions as normal for this purpose?

jim-docker: This is the fix. `appSemVer` almost always has a prelease tag, and to avoid requiring extension developers from having to specify one in their range we need to set `{ includePrerelease: true }`

src/extensions/extension-compatibility.ts
```diff
@@ -23,8 +23,12 @@ import semver from "semver";
 import { appSemVer } from "../common/vars";
 import type { LensExtensionManifest } from "./lens-extension";
 
-export function isCompatibleExtension(manifest: LensExtensionManifest): boolean {
+export function isCompatibleExtension(manifest: LensExtensionManifest, isBundled: boolean): boolean {
   if (manifest.engines?.lens) {
+    if (isBundled) {
+      return true;
+    }
+
```
jim-docker: Comment that, since Lens v5, bundled extensions must specify the lens engine?
This assumes that we aren't going to generate any v4 versions from this code :)

jim-docker (dismissed):

src/extensions/extension-compatibility.ts
Nokel81: That is correct, we only ever generate a previous patch version from the tag (plus additonal commits) of the most recent patch version.

PR #3417, opened 2021-07-17
Allow clicking on selector badge to navigate to filtered pods/nodes page

### LINKED ISSUE #41: Labels annotations and selectors should be clickable
I'm looking at this service 
image

And I see the label `app: rxindicator`

Down below I even see the selector `app=rxindicator` but no way to get to the pods it represents.

It seems to me that clicking a label, annotation, or selector should take you to a universal resource list page with that selector applied and show all matching resources.

-----

Barring that, paying special attention to Services with a selector, those ones should definitely be linked to the Pods list with that selector applied.

devodev:
### Description
- Add `BadgeNavigatable` component that navigates to a resource page on click
- Update the `.Badge.clickable:hover` style to now have a primary-colored outline
- Use `stringifyLabels` in `getNodeSelectors` of `WorkloadKubeObject` to align with other `getNodeSelectors` (use `=` separator)

Replaces `Selector` and `Node Selector` badges of: `Services`, `Pods`, `Daemonsets`, `Deployments`, `Replicasets` and `Statefulsets`.

### Screenshots

Service selector on hover

image

Pod node selector on hover

image

Filtered pods (Pod node selector)

image

Service annotations on hover (update to `.Badge.clickable:hover`)

image

---

Resolves https://github.com/lensapp/lens/issues/41

### DIFF
--- a/src/renderer/api/endpoints/pods.api.ts
+++ b/src/renderer/api/endpoints/pods.api.ts
@@ -19,6 +19,7 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
+import { KubeObject } from "../kube-object";
 import { IAffinity, WorkloadKubeObject } from "../workload-kube-object";
 import { autoBind } from "../../utils";
 import { IMetrics, metricsApi } from "./metrics.api";
@@ -409,7 +410,7 @@ export class Pod extends WorkloadKubeObject {
   getNodeSelectors(): string[] {
     const { nodeSelector = {} } = this.spec;
 
-    return Object.entries(nodeSelector).map(values => values.join(": "));
+    return KubeObject.stringifyLabels(nodeSelector);
   }
 
   getTolerations() {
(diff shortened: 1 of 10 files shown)


Nokel81:
Since we already have a style for clickable badges I don't think we should change that in a PR like this.

I don't think that we need the full annotation in a tooltip since users can already click the label to expand it.

I don't think we should use the primary mouse button for this sort of navigation since it might conflict with the expansion. How about a second mouse button context menu (WDYT @aleksfront ?)

devodev:
> Since we already have a style for clickable badges I don't think we should change that in a PR like this.
> 
> I don't think that we need the full annotation in a tooltip since users can already click the label to expand it.
> 
> I don't think we should use the primary mouse button for this sort of navigation since it might conflict with the expansion. How about a second mouse button context menu (WDYT @aleksfront ?)

Just to clarify: The annotations tooltip is already there on master. I included the screenshot only to show the new primary-colored outline on clickable items

I agree that we might want to make it more discernable between clickable/expandable and navigatable badge. Please advise if you have any ideas, I did this on the spot :)

Nokel81:
Ah I see, how about a new `<Menu>` (context menu)?

devodev:
> Ah I see, how about a new `<Menu>` (context menu)?

I believe it would be a great enhancement. Would you also use it for expanding the badge? I saw there is some logic on click to check if the text is selected before shrinking it back. I wonder how we would handle that if we use a context menu instead.

Nokel81:
I don't think we should use the menu for expanding the badge.

devodev:
Ok I will move this to a draft and hack on it a bit to see what I can come up with. Will keep you in the loop :)

Nokel81:

PR #3418, opened 2021-07-18
Add Pod IP column in pods table

devodev:
- Add `GetIP()` method in `PodsApi` class
- Add `Pod IP` column in pods list
- Re-order `columnId` enum and `KubeObjectListLayout.sortingCallbacks` to match default column ordering

Signed-off-by: devodev <abalexandrebarone@gmail.com>

Resolves https://github.com/lensapp/lens/issues/709

PR #3419, opened 2021-07-18
Allow having table columns hidden by default

devodev:
Based on https://github.com/lensapp/lens/pull/3418

- Introduces `hiddenColumns` attribute in `ItemListLayoutProps`
- Make `QoS` column in pods table hidden by default (effectively replaces `QoS` with `Pod IP`)

Breaking change: Format of `hiddenTableColumns` in user-store updated from `Map<string, Set<string>>` to `Map<string, Map<string, boolean>>`

table-columns-hidden-by-default

Signed-off-by: devodev abalexandrebarone@gmail.com

### DIFF
--- a/src/common/user-store/user-store.ts
+++ b/src/common/user-store/user-store.ts
@@ -21,7 +21,7 @@
 
 import { app, remote } from "electron";
 import semver from "semver";
-import { action, computed, observable, reaction, makeObservable } from "mobx";
+import { action, computed, observable, reaction, makeObservable, ObservableMap } from "mobx";
 import { BaseStore } from "../base-store";
 import migrations from "../../migrations/user-store";
 import { getAppVersion } from "../utils/app-version";
@@ -29,7 +29,7 @@ import { kubeConfigDefaultPath } from "../kube-helpers";
 import { appEventBus } from "../event-bus";
 import path from "path";
 import { fileNameMigration } from "../../migrations/user-store";
-import { ObservableToggleSet, toJS } from "../../renderer/utils";
+import { toJS } from "../../renderer/utils";
 import { DESCRIPTORS, KubeconfigSyncValue, UserPreferencesModel } from "./preferences-helpers";
 import logger from "../../main/logger";
 
@@ -79,7 +79,7 @@ export class UserStore extends BaseStore<UserStoreModel> /* implements UserStore
    * The column IDs under each configurable table ID that have been configured
    * to not be shown
    */
-  hiddenTableColumns = observable.map<string, ObservableToggleSet<string>>();
+  hiddenTableColumns = observable.map<string, ObservableMap<string, boolean>>();
 
   /**
    * The set of file/folder paths to be synced
@@ -136,12 +136,12 @@ export class UserStore extends BaseStore<UserStoreModel> /* implements UserStore
   /**
    * Toggles the hidden configuration of a table's column
    */
-  toggleTableColumnVisibility(tableId: string, columnId: string) {
-    if (!this.hiddenTableColumns.get(tableId)) {
-      this.hiddenTableColumns.set(tableId, new ObservableToggleSet());
+  toggleTableColumnVisibility(tableId: string, columnId: string, defaultHidden: boolean) {
+    if (this.hiddenTableColumns.get(tableId)?.get(columnId) === undefined){
+      this.hiddenTableColumns.get(tableId)?.set(columnId, !defaultHidden);
+    } else if (!!this.hiddenTableColumns.get(tableId)?.get(columnId) !== defaultHidden) {
+      this.hiddenTableColumns.get(tableId)?.delete(columnId);
     }
-
-    this.hiddenTableColumns.get(tableId).toggle(columnId);
   }
 
   @action
--- a/src/renderer/components/item-object-list/item-list-layout.tsx
+++ b/src/renderer/components/item-object-list/item-list-layout.tsx
@@ -23,11 +23,11 @@ import "./item-list-layout.scss";
 import groupBy from "lodash/groupBy";
 
 import React, { ReactNode } from "react";
-import { computed, makeObservable } from "mobx";
+import { computed, makeObservable, observable } from "mobx";
 import { observer } from "mobx-react";
 import { ConfirmDialog, ConfirmDialogParams } from "../confirm-dialog";
 import { Table, TableCell, TableCellProps, TableHead, TableProps, TableRow, TableRowProps, TableSortCallbacks } from "../table";
-import { boundMethod, createStorage, cssNames, IClassName, isReactNode, noop, ObservableToggleSet, prevDefault, stopPropagation } from "../../utils";
+import { boundMethod, createStorage, cssNames, IClassName, isReactNode, noop, prevDefault, stopPropagation } from "../../utils";
 import { AddRemoveButtons, AddRemoveButtonsProps } from "../add-remove-buttons";
 import { NoItems } from "../no-items";
 import { Spinner } from "../spinner";
@@ -68,6 +68,7 @@ export interface ItemListLayoutProps<I extends ItemObject> {
   searchFilters?: SearchFilter<I>[];
   /** @deprecated */
   filterItems?: ItemsFilter<I>[];
+  hiddenColumns?: string[];
 
   // header (title, filtering, searching, etc.)
   showHeader?: boolean;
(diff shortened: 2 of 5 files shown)


src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -136,14 +138,17 @@ export class ItemListLayout extends React.Component<ItemListLayoutProps> {
...
 
     if (isConfigurable && !tableId) {
       throw new Error("[ItemListLayout]: configurable list require props.tableId to be specified");
     }
 
     if (isConfigurable && !UserStore.getInstance().hiddenTableColumns.has(tableId)) {
       UserStore.getInstance().hiddenTableColumns.set(tableId, new ObservableToggleSet());
+      hiddenColumns.forEach(columnId => (
+        UserStore.getInstance().toggleTableColumnVisibility(tableId, columnId))
+      );
```
Nokel81: This is incorrect, it will toggle the columns visibility on ever mount

devodev: Ohhhh I realize my mistake. Could you advise where I should put this logic so that it executes only once?

devodev: I just re-checked, and it should only execute once when there is no `hiddenTableColumns` instance associated to the tableId.

Nokel81: But then there will be on the next time it is rendered.

devodev: Let me re-test that, but I assumed that the observable was linked to the tableId once on the first mount, then it would remain for the lifetime of the app since UserStore is a singleton.

Nokel81: Okay so it is slightly more complicated. If you reset your hidden tables to "default" (which is currently `[]`) then the next time it is round-tripped through the persistence layer that `tableId` will be removed.

Nokel81: If we do this correctly, we are able to change the defaults

devodev: Ok I see! Mmmm I'll need to think about it a bit then If you have any insight, Im all ears.

I'm moving it to a draft to work on it a bit more.

Nokel81: The proper place for this is in the `src/common/user-store/preference-helpers.ts` file. I guess.

devodev: Thanks, I'll look into it, appreciated!

devodev:
@Nokel81 I updated `hiddenTableColumns` from `ObservableToggleSet` to `ObservableMap` to be able to set a boolean for whether the column should be shown or hidden. Not sure about the code or if we want to introduce breaking changes for such small PRs, but Im happy I was able to make it work :)

Could you please let me know what you think?

Thank you!

jakolehm (approved):
LGTM

src/common/user-store/user-store.ts
```diff
@@ -136,12 +136,12 @@ export class UserStore extends BaseStore<UserStoreModel> /* implements UserStore
   /**
    * Toggles the hidden configuration of a table's column
    */
-  toggleTableColumnVisibility(tableId: string, columnId: string) {
-    if (!this.hiddenTableColumns.get(tableId)) {
-      this.hiddenTableColumns.set(tableId, new ObservableToggleSet());
+  toggleTableColumnVisibility(tableId: string, columnId: string, defaultHidden: boolean) {
+    if (this.hiddenTableColumns.get(tableId)?.get(columnId) === undefined){
```
Nokel81, line 140: nit: this should use `has`

Nokel81: further nit, should we either warn or initialize `this.hiddenTableColumns.get(tableId)` if it is not present?

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -68,6 +68,7 @@ export interface ItemListLayoutProps<I extends ItemObject> {
   searchFilters?: SearchFilter<I>[];
   /** @deprecated */
   filterItems?: ItemsFilter<I>[];
+  hiddenColumns?: string[];
```
Nokel81, line 71: I think it would be better for this to be called `columnsHiddenByDefault`

src/renderer/components/item-object-list/item-list-layout.tsx
```diff
@@ -476,13 +478,13 @@ export class ItemListLayout<I extends ItemObject> extends React.Component<ItemLi
   }
 
   showColumn({ id: columnId, showWithColumn }: TableCellProps): boolean {
-    const { tableId, isConfigurable } = this.props;
+    const { tableId, hiddenColumns, isConfigurable } = this.props;
 
-    return !isConfigurable || !UserStore.getInstance().isTableColumnHidden(tableId, columnId, showWithColumn);
+    return !isConfigurable || !(UserStore.getInstance().isTableColumnHidden(tableId, columnId, showWithColumn) !== hiddenColumns.includes(columnId));
```
Nokel81, line 483: This seems to imply that a stored value of `true` means something different for a column that has been marked as `hidden`.

That will certainly trip someone up. I think I better check would be to pass `hiddenColumns` to `isTableColumnHidden`. And if it isn't set then return the computed value from the hidden columns list.

That way, it truely acts as a default. The other benefit of this is that we don't have to add a migration or change the user-store.

Nokel81 (changes requested):
I like this idea but it is currently broken as we change a store type without a new migration.

src/common/user-store/user-store.ts
devodev: Yes, `has` is cleaner I agree, dont know why I didnt think about it. Will update.

Second point, in theory it should be present from being set in the onMount hook, but to be safe we could log a warning, yeah.

Nokel81: I think if we move to initialize, we no longer need the onmount hook at all.

src/renderer/components/item-object-list/item-list-layout.tsx
devodev: I agree that having to add a migration is not the ideal solution, but then the name `hiddenTableColumns` is also confusing if there is an entry for `qos` lets say, you would think its hidden just by looking at the saved state, when in fact, its now shown because its default is hidden.

I also agree my implementation with the booleans is a bit confusing and requires a migration, so I'll try what you proposed.

Nokel81: Good point... maybe a combination? Having a migration is not the end of the world. We just need one with this current change set. And I admit that it would be good for a user to be able to override the default.

src/common/user-store/user-store.ts
devodev: Sure, let me update that.

src/renderer/components/item-object-list/item-list-layout.tsx
devodev: Ok I'll update to pass hiddenColumns to `isTableColumnHidden`. How would you update the func signature? It currently is `isTableColumnHidden(tableId: string, ...columnIds: string[])` and it feels weird to pass hiddenColumns on either side of tableId.

What about a method in item-list-layout that does the conditional instead?

Nokel81: A method in `ItemListLayout` probably would be better. But I would change the return type of `isTableColumnHidden` to `boolean | undefined` and return `undefined` in the cases where the fields don't exist.

devodev: gotcha, will update :)

devodev: Also Im not sure how I can create a migration? Do you have docs or maybe just an example I can get inspired from? Or you'd prefer to deal with this yourself?

Thanks

Nokel81: There already exist migrations in the `src/migrations/user-store` folder. I would follow their lead. You need to specify a version (I would pick `5.1.4-beta.1` even though we don't plan on releasing a version with that number).

vfjerlkj (approved):
It is great

devodev:
Sorry I was on vacation and thought I had put that back as a draft.

maxpain:
Any updates?

Nokel81:

PR #3420, opened 2021-07-19
Skip previously bundled extensions

### LINKED ISSUE #3396: [EXTENSIONS-LOADER]: can't load renderer for "example-extension": TypeError: Class extends value ...

jakolehm:
Fixes #3396

### DIFF
--- a/src/extensions/extension-compatibility.ts
+++ b/src/extensions/extension-compatibility.ts
@@ -31,3 +31,7 @@ export function isCompatibleExtension(manifest: LensExtensionManifest): boolean 
 
   return false;
 }
+
+export function isCompatibleBundledExtension(manifest: LensExtensionManifest): boolean {
+  return manifest.version === appSemVer.raw;
+}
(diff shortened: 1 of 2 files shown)


aleksfront (approved):

Nokel81:
Alternative is https://github.com/lensapp/lens/pull/3397 but I think this PR probably will work better.

src/extensions/extension-compatibility.ts
```diff
@@ -31,3 +31,7 @@ export function isCompatibleExtension(manifest: LensExtensionManifest): boolean
 
   return false;
 }
+
+export function isCompatibleBundledExtension(manifest: LensExtensionManifest): boolean {
+  return manifest.version === appSemVer.raw;
```
jim-docker, line 36: Does this work when `appSemVer` has a prerelease tag? (Does `manifest.version` have the matching tag, at least for bundled v5 extensions?)

Nokel81: Yes do (I believe). We set them to the exact version that the `package.json` was changed to

jakolehm: During build the app version is synced to the bundled extensions: https://github.com/lensapp/lens/blob/master/build/set_build_version.ts#L45

Nokel81:
closes #3397

PR #3422, opened 2021-07-19
Catalog render optimizations

### LINKED ISSUE #3389: Catalog list poor performance
Switching between catalog categories goes with visible delay, causing poor UX. Rendering process here should be optimised.

https://user-images.githubusercontent.com/9607060/125783224-755c2fdb-8ed4-47bf-9611-ab9134f58ebe.mov

aleksfront:
Optimising Catalog list rendering by:
1. Using same `ItemListLayout` for each section
2. Moving slow components (components which user doesn't need to see instantly, e. g. submenus, add clusters button etc) to newly introduced `<RenderDelay/>`. It renders its own children only when browser goes to idle state (not busy rendering something).

Dev mode:

https://user-images.githubusercontent.com/9607060/126120105-64139e3a-a7c2-4c0a-bdca-2930dfedbd97.mov

Production mode:

https://user-images.githubusercontent.com/9607060/126120154-f1c07862-639b-4dbc-8382-43f22bfba4f8.mov

Fixes #3389

### DIFF
--- a/src/renderer/components/+catalog/catalog.tsx
+++ b/src/renderer/components/+catalog/catalog.tsx
@@ -44,6 +44,7 @@ import { CatalogMenu } from "./catalog-menu";
 import { HotbarIcon } from "../hotbar/hotbar-icon";
 import { RenderDelay } from "../render-delay/render-delay";
 import { CatalogTopbar } from "../cluster-manager/catalog-topbar";
+import type { TableSortCallback } from "../table";
 
 export const previousActiveTab = createAppStorage("catalog-previous-active-tab", "");
 
@@ -194,8 +195,17 @@ export class Catalog extends React.Component<Props> {
 
   renderList() {
     const { activeCategory } = this.catalogEntityStore;
+    const tableId = activeCategory ? `catalog-items-${activeCategory.metadata.name.replace(" ", "")}` : "catalog-items";
+    let sortingCallbacks: { [sortBy: string]: TableSortCallback } = {
+      [sortBy.name]: (item: CatalogEntityItem<CatalogEntity>) => item.name,
+      [sortBy.source]: (item: CatalogEntityItem<CatalogEntity>) => item.source,
+      [sortBy.status]: (item: CatalogEntityItem<CatalogEntity>) => item.phase,
+    };
 
-    const tableId = activeCategory ? `catalog-items-${activeCategory?.metadata.name.replace(" ", "")}` : "catalog-items";
+    sortingCallbacks = activeCategory ? sortingCallbacks : {
+      ...sortingCallbacks,
+      [sortBy.kind]: (item: CatalogEntityItem<CatalogEntity>) => item.kind,
+    };
 
     if (this.activeTab === undefined) {
       return null;
@@ -209,12 +219,7 @@ export class Catalog extends React.Component<Props> {
         isConfigurable={true}
         className="CatalogItemList"
         store={this.catalogEntityStore}
-        sortingCallbacks={{
-          [sortBy.name]: (item: CatalogEntityItem<CatalogEntity>) => item.name,
-          [sortBy.kind]: (item: CatalogEntityItem<CatalogEntity>) => item.kind,
-          [sortBy.source]: (item: CatalogEntityItem<CatalogEntity>) => item.source,
-          [sortBy.status]: (item: CatalogEntityItem<CatalogEntity>) => item.phase,
-        }}
+        sortingCallbacks={sortingCallbacks}
         searchFilters={[
           (entity: CatalogEntityItem<CatalogEntity>) => entity.searchFields,
         ]}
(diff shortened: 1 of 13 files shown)


src/renderer/components/menu/menu-actions.tsx
```diff
@@ -124,30 +125,32 @@ export class MenuActions extends React.Component<MenuActionsProps> {
...
-          )}
-          {removeAction && (
-            <MenuItem onClick={this.remove}>
-              <Icon material="delete" interactive={toolbar} tooltip="Delete"/>
-              <span className="title">Remove</span>
-            </MenuItem>
-          )}
-        </Menu>
+        <RenderDelay>
+          <Menu
```
aleksfront, line 129: While menu trigger icon is rendered and visible instantly, menu itself which appears only on user click may be postponed to render.

aleksfront: Decided to use `Menu` with `RenderDelay` because it triggers a lot of actions such as bunch of `addEventListener` for each menu instance.

src/renderer/components/menu/menu.tsx
```diff
@@ -260,6 +260,8 @@ export class Menu extends React.Component<MenuProps, State> {
   }
 
   onBlur() {
+    if (!this.isOpen) return;
```
aleksfront: Performance analysis shows that menu instances triggers a lot of `onBlur()` events when unmounting.

src/renderer/components/render-delay/render-delay.tsx
```diff
@@ -0,0 +1,61 @@
...
+export class RenderDelay extends React.Component<Props> {
+  @observable isVisible = false;
+
+  constructor(props: Props) {
+    super(props);
+    makeObservable(this);
+  }
+
+  componentDidMount() {
+    window.requestIdleCallback(this.showContents);
```
aleksfront: https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -144,7 +146,11 @@ export class Catalog extends React.Component<Props> {
   };
 
   renderNavigation() {
-    return <CatalogMenu activeItem={this.activeTab} onItemClick={this.onTabChange}/>;
+    return (
+      <RenderDelay>
+        <CatalogMenu activeItem={this.activeTab} onItemClick={this.onTabChange}/>
```
Nokel81: I noticed this flashes in the videos you show, maybe this shouldn't be delayed

src/renderer/components/menu/menu.tsx
Nokel81: Maybe add a comment to that effect then.

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -172,69 +178,37 @@ export class Catalog extends React.Component<Props> {
...
-        ]}
-        onDetails={this.onDetails}
-        renderItemMenu={this.renderItemMenu}
-      />
-    );
-  }
+  renderList() {
+    const { activeCategory } = this.catalogEntityStore;
+
+    const tableId = activeCategory ? `catalog-items-${activeCategory?.metadata.name.replace(" ", "")}` : "catalog-items";
```
jim-docker: you don't need the second `?` ?

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -172,69 +178,37 @@ export class Catalog extends React.Component<Props> {
...
+        tableId={tableId}
+        renderHeaderTitle={activeCategory?.metadata.name || "Browse All"}
         isSelectable={false}
         isConfigurable={true}
         className="CatalogItemList"
         store={this.catalogEntityStore}
-        tableId="catalog-items"
         sortingCallbacks={{
           [sortBy.name]: (item: CatalogEntityItem<CatalogEntity>) => item.name,
           [sortBy.kind]: (item: CatalogEntityItem<CatalogEntity>) => item.kind,
```
jim-docker: Should you use `!activeCategory && ` ... here too?

src/renderer/components/render-delay/render-delay.tsx
jim-docker: should you set the timeout option here? (maybe you tried it...)

jim-docker (commented):
I didn't notice any issues while testing

src/renderer/components/+catalog/catalog.tsx
aleksfront: Yes, but this component affect performance. Probably we should go away from Material `<TreeView/>` for sidebar. Some css-in-js operations affect rendering, as it turned out.

src/renderer/components/render-delay/render-delay.tsx
aleksfront: Set the `timeout: 1000` option.

Nokel81 (approved):

PR #3424, opened 2021-07-19
Improving metric queries

aleksfront:
* Metric query improvements extracted from #1944
* Metrics added to namespaces and jobs
* Metric loaders are moved from stores to React components. The reason here because metrics doesn't really relate to stores. See this comment https://github.com/lensapp/lens/pull/1944#discussion_r556622245

node metrics

### DIFF
--- a/src/renderer/api/endpoints/metrics.api.ts
+++ b/src/renderer/api/endpoints/metrics.api.ts
@@ -60,7 +60,7 @@ export interface IMetricsReqParams {
   namespace?: string;             // rbac-proxy validation param
 }
 
-export interface IResourceMetrics<T = IMetrics> {
+export interface IResourceMetrics<T extends IMetrics> {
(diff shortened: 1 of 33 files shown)


src/renderer/api/endpoints/metrics.api.ts
```diff
@@ -60,6 +60,15 @@ export interface IMetricsReqParams {
   namespace?: string;             // rbac-proxy validation param
 }
 
+export interface IResourceMetrics<T = IMetrics> {
```
Nokel81: Shouldn't this really be `T extends IMetrics`?

jakolehm:
Moving to 5.2.0 because it's an enhancement.

Nokel81 (approved):
Please do not merge, I'll merge this after the release of 5.1.3

PR #3395, opened 2021-07-15
getPortFrom should log the lines it retrived if it fails to get a port

Nokel81:
Signed-off-by: Sebastian Malton <sebastian@malton.name>

need more information to fix https://sentry.io/organizations/k8slens/issues/2514284178/?project=5753768

PR #3425, opened 2021-07-19
Log as errors the places where getPortFrom can fail

Nokel81:

follow up to #3395


jim-docker (approved):
LGTM

PR #3428, opened 2021-07-19
Create c&p folder before starting kube sync

### LINKED ISSUE #3407: Issues trying to add clusters to Lens
**Describe the bug**
Cannot add  clusters that are not in ~/.kube/config file. Regardless of what I paste in the input filed, the only cluster that appears in the cluster list is the one in ~/.kube/config file. If the ~/.kube/config file does not exist, no cluster is added.

**To Reproduce**
1. Go to 'clusters'
2. Click on '+'
3. Paste Kubeconfig contents into input field

**Expected behavior**
A new cluster should appear on the cluster list

**Screenshots**
The only cluster that appears is the one on ~/.kbue/config

image

**Environment**
OS: OSX
Lens: 5.0.2-latest.20210705.2
Electron: 9.4.4
Chrome: 83.0.4103.122
Node: 12.14.1

Nokel81:

fixes #3407

### DIFF
--- a/src/main/catalog-sources/kubeconfig-sync.ts
+++ b/src/main/catalog-sources/kubeconfig-sync.ts
@@ -22,10 +22,9 @@
 import { action, observable, IComputedValue, computed, ObservableMap, runInAction, makeObservable, observe } from "mobx";
 import type { CatalogEntity } from "../../common/catalog";
 import { catalogEntityRegistry } from "../../main/catalog";
-import { watch } from "chokidar";
+import { FSWatcher, watch } from "chokidar";
 import fs from "fs";
 import path from "path";
-import fse from "fs-extra";
 import type stream from "stream";
 import { Disposer, ExtendedObservableMap, iter, Singleton } from "../../common/utils";
 import logger from "../logger";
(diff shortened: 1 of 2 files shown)


src/main/catalog-sources/kubeconfig-sync.ts
```diff
@@ -258,37 +253,59 @@ function diffChangedConfig(filePath: string, source: RootSource): Disposer {
...
-    })
-    .on("unlink", (childFilePath) => {
-      stoppers.get(childFilePath)();
-      stoppers.delete(childFilePath);
-      rootSource.delete(childFilePath);
-    })
-    .on("error", error => logger.error(`${logPrefix} watching file/folder failed: ${error}`, { filePath }));
+  let watcher: FSWatcher;
+
+  (async () => {
```
nevalla, line 276: How about implementing this as `watch` method?

Nokel81: That doesn't work, none of the OS's or the libraries that I tried reported any events when watching a path that doesn't exist at the time of call.

nevalla (approved):

PR #3433, opened 2021-07-19
Fix permissions issue on ServiceDetails under RBAC

### LINKED ISSUE #3402: Service details pane tries loading endpoints from default namespace
**Describe the bug**
The service details pane tries to load endpoints and fails to display any endpoint when the user does not have access to the default namespace (even when default namespace is not selected).

Navigating to `Network/endpoints` does display the endpoints correctly, and I get no errors trying to load from the default namespace.

**To Reproduce**
1. Switch to a user that does not have access to default namespace
2. Go to `Network/services`
3. Select 1 namespace you are allowed to access
4. Click on any service to show the service details pane
5. See error

**Expected behavior**
Service details pane does not load endpoints for the default namespace if not selected.

**Screenshots**
image

- Lens Version: 5.1.0-latest.20210715.1
- OS: OSX

Nokel81:
- Only watch the namespace that the service is in instead of all of them

fixes #3402

### DIFF
--- a/src/renderer/components/+network-services/service-details.tsx
+++ b/src/renderer/components/+network-services/service-details.tsx
@@ -39,9 +39,12 @@ interface Props extends KubeObjectDetailsProps<Service> {
 @observer
 export class ServiceDetails extends React.Component<Props> {
   componentDidMount() {
+    const { object: service } = this.props;
+
     disposeOnUnmount(this, [
       kubeWatchApi.subscribeStores([endpointStore], {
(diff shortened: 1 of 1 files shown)


Nokel81:
@devodev does this fix your issue? I think it should.

devodev:
> @devodev does this fix your issue? I think it should.

At first glance, it should yes.

jakolehm (approved):

PR #3452, opened 2021-07-20
Ignore certain file globs when under dir kube sync

Nokel81:
- .DS_Store
- *.lock
- *.swp


src/main/catalog-sources/kubeconfig-sync.ts
```diff
@@ -36,9 +37,23 @@ import { UserStore } from "../../common/user-store";
...
 import { homedir } from "os";
+import globToRegExp from "glob-to-regexp";
+import { inspect } from "util";
 
 const logPrefix = "[KUBECONFIG-SYNC]:";
 
+/**
+ * This is the list of globs of which files are ignored when under a folder sync
+ */
+const ignoreGlobs = [
```
nevalla, line 48: Should we allow users to add more globs in future PR?

nevalla (approved):
LGTM

jakolehm (approved):

src/main/catalog-sources/kubeconfig-sync.ts
Nokel81: That sounds like a good idea in the future. Though we might wait until someone asks for it.

Issue #3226, opened 2021-06-30
Windows Lens v5.0.0 first connection attempt fails


Kolossi:
**Describe the bug**
After adding external "sync" configs to lens for windows v5.0.0, first attempt to connect to the cluster always fails with "Invalid Credentials error"

**To Reproduce**
1. Add a cluster external config file as a "Kubeconfig Sync" in preferences "Kubernetes" tab
2. restart Lens
3. Attempt to connect to cluster 
4. Error "invalid credentials" is shown
5. Attempt to connect again and connection can suceed

**Expected behavior**
Where config is valid, first attempt should suceed without error.

**Screenshots**

Screenshot, including failed attempt to get debug logs for win version using WSL terminal:

![image](https://user-images.githubusercontent.com/5367728/123981046-eccd8080-d9b9-11eb-9f76-554a98634e2b.png)

**Environment (please complete the following information):**
- Lens Version: v5.0.0
- OS: Windows
- Installation method : auto upgraded from v5.0.0

**Logs:**
When you run the application executable from command line you will see some logging output. Please paste them here:

Unable to get debug logs from windows lens - see later part of discussion on https://k8slens.slack.com/archives/C1U90NQN8/p1625045050141900 and screenshot in this post.

**Kubeconfig:**
Quite often the problems are caused by malformed kubeconfig which the application tries to load. Please share your kubeconfig, remember to remove any secret and sensitive information.
```
N/A: connection works on second attempt
first time fail occurs with multiple clusters and with valid files using both ServiceAccount and Dex/LDAP login.
```
** Additional **

Would like to grab and post debug logging, will do so when advised how to do it in windows version of lens.

jeffjones-kbx:
I am also getting the failed connect at initial connection to a cluster. Hit Reconnect a few times and got further but received the output in the screenshot below.  I am able to view the cluster resources now though.

![image](https://user-images.githubusercontent.com/77127429/123982229-76645a80-d988-11eb-9a30-c5fe580820c6.png)

Kolossi:
Additional observation:

When performing the initial failing connect attempt there is only the message "Connecting..." and no message "Authentication proxy started" at any stage.

When clicking to try again there is the message "Reconnecting..." and then after a while "Authentication proxy started" after which the connection then succeeds.

Nokel81:
That is a good hint thanks

andyliddle:
Can take 2 or 3 times to connect and it's not quick compared to previous versions.

[...]

PR #3458, opened 2021-07-20
Fix kube auth proxy race condition

jim-docker:
fix race condition where the first call to `ensureServer()` initiates the kube auth proxy  but a subsequent call to `ensureServer()` does not wait for the initiation to complete. Relying on mobx `when()` to indicate when the kube auth proxy is really ready.

fixes some cluster connection issues, hopefully some reported in #3226

### DIFF
--- a/src/main/kube-auth-proxy.ts
+++ b/src/main/kube-auth-proxy.ts
@@ -67,53 +67,53 @@ export class KubeAuthProxy {
   }
 
   public async run(): Promise<void> {
-    if (!this.proxyProcess) {
-      const proxyBin = await this.kubectl.getPath();
-      const args = [
-        "proxy",
-        "-p", "0",
-        "--kubeconfig", `${this.cluster.kubeConfigPath}`,
-        "--context", `${this.cluster.contextName}`,
-        "--accept-hosts", this.acceptHosts,
-        "--reject-paths", "^[^/]"
-      ];
-
-      if (process.env.DEBUG_PROXY === "true") {
-        args.push("-v", "9");
-      }
-      logger.debug(`spawning kubectl proxy with args: ${args}`);
-
-      this.proxyProcess = spawn(proxyBin, args, { env: this.env, });
-      this.proxyProcess.on("error", (error) => {
-        this.sendIpcLogMessage({ data: error.message, error: true });
-        this.exit();
-      });
-
-      this.proxyProcess.on("exit", (code) => {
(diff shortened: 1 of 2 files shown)


src/main/kube-auth-proxy.ts
```diff
@@ -58,52 +62,58 @@ export class KubeAuthProxy {
...
-      onFind: () => this.sendIpcLogMessage({ data: "Authentication proxy started\n" }),
-    });
-
-    this.proxyProcess.stdout.on("data", (data: any) => {
-      this.sendIpcLogMessage({ data: data.toString() });
-    });
-
-    return waitUntilUsed(this.port, 500, 10000);
+    
+    return this.whenReady;
```
Nokel81: Since you removed the `await run()` shouldn't this not return this promise?

src/main/kube-auth-proxy.ts
```diff
@@ -58,52 +62,58 @@ export class KubeAuthProxy {
...
-      "-p", "0",
-      "--kubeconfig", `${this.cluster.kubeConfigPath}`,
-      "--context", `${this.cluster.contextName}`,
-      "--accept-hosts", this.acceptHosts,
-      "--reject-paths", "^[^/]"
-    ];
-
-    if (process.env.DEBUG_PROXY === "true") {
-      args.push("-v", "9");
+  public async run(): Promise<void> {
```
Nokel81, line 69: I think a better name would probably be `start` now.

src/main/kube-auth-proxy.ts
jim-docker: Not sure I understand. I removed the `await` from the `run()` call in `ensureServer()` just because `ensureServer()` has to finally await the `whenReady` anyway. I could leave it as `await run()`. Does it matter?

`run()` has to return this promise, for when it gets called when `this.proxyProcess` is defined already, right?

Nokel81: Yes I see why you removed the `await`. I think that is correct. However, because you did that this getter is creating a new promise which is then wrapped into the upper promise and then thrown away because no one await it.

I think it would be better to just not return anything. Even better it would be nice if the promise inside `run()` didn't leak at all out of it. So that the return type was `void`. This would make it inline with a `start()` function.

The corollary of that could be an `ensureStart()` which calls start and then awaits the `whenReady` getter promise. But that might be a bit too much.

Nokel81 (approved):

PR #3469, opened 2021-07-21
Add ability to rename hotbars

### LINKED ISSUE #2731: support renaming of hotbars
allow hotbars to be renamed within lens

vshakirova:
Suggestion: Maybe we could use `value.trim()` in `uniqueHotbarName()` func, line:
`  validate: value => !HotbarStore.getInstance().getByName(value))`? To avoid "name" and "name         " case. Wdyt? 

https://user-images.githubusercontent.com/38247153/126641485-3380b847-d849-4bc5-9d81-4a9ec708a507.mp4

fixes: https://github.com/lensapp/lens/issues/2731

### DIFF
--- a/src/common/hotbar-store.ts
+++ b/src/common/hotbar-store.ts
@@ -154,6 +154,19 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
   }
 
   @action
+  setHotbarName(id: string, name: string) {
+    const index = this.hotbars.findIndex((hotbar) => hotbar.id === id);
+
+    if(index < 0) {
+      console.warn(`[HOTBAR-STORE]: cannot setHotbarName: unknown id`, { id });
+
+      return;
+    }
+
+    this.hotbars[index].name = name;
+  }
+
+  @action
   remove(hotbar: Hotbar) {
     this.hotbars = this.hotbars.filter((h) => h !== hotbar);
 
--- a/src/renderer/components/hotbar/hotbar-rename-command.tsx
+++ b/src/renderer/components/hotbar/hotbar-rename-command.tsx
@@ -21,43 +21,93 @@
 
 import React from "react";
 import { observer } from "mobx-react";
+import { Select } from "../select";
+import { action, computed, makeObservable, observable } from "mobx";
 import { HotbarStore } from "../../../common/hotbar-store";
+import { hotbarDisplayLabel } from "./hotbar-display-label";
+import { Input } from "../input";
+import { uniqueHotbarName } from "./hotbar-add-command";
 import { CommandOverlay } from "../command-palette";
-import { Input, InputValidator } from "../input";
-
-export const uniqueHotbarName: InputValidator = {
-  condition: ({ required }) => required,
-  message: () => "Hotbar with this name already exists",
-  validate: value => !HotbarStore.getInstance().getByName(value),
-};
 
 @observer
-export class HotbarAddCommand extends React.Component {
+export class HotbarRenameCommand extends React.Component {
+  @observable hotbarId = "";
+  @observable hotbarName = "";
+
+  constructor(props: {}) {
+    super(props);
+    makeObservable(this);
+  }
+
+  @computed get options() {
+    return HotbarStore.getInstance().hotbars.map((hotbar) => {
+      return { value: hotbar.id, label: hotbarDisplayLabel(hotbar.id) };
+    });
+  }
+
+  @action onSelect = (id: string) => {
+    this.hotbarId = id;
+    this.hotbarName = HotbarStore.getInstance().getById(this.hotbarId).name;
+  };
+
   onSubmit = (name: string) => {
     if (!name.trim()) {
       return;
     }
 
-    HotbarStore.getInstance().add({ name }, { setActive: true });
+    const hotbarStore = HotbarStore.getInstance();
+    const hotbar = HotbarStore.getInstance().getById(this.hotbarId);
+
+    if (!hotbar) {
+      return;
+    }
+
+    hotbarStore.setHotbarName(this.hotbarId, name);
     CommandOverlay.close();
   };
 
-  render() {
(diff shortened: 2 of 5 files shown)


src/common/hotbar-store.ts
```diff
@@ -153,6 +153,13 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
     }
   }
 
+  @action
+  setHotbarName(id: string, name: string) {
+    const idx = this.hotbars.findIndex((hotbar) => hotbar.id === id);
```
Nokel81: This should check for unknown id

src/renderer/components/hotbar/hotbar-add-command.tsx
```diff
@@ -25,7 +25,7 @@ import { HotbarStore } from "../../../common/hotbar-store";
 import { CommandOverlay } from "../command-palette";
 import { Input, InputValidator } from "../input";
 
-const uniqueHotbarName: InputValidator = {
+export const uniqueHotbarName: InputValidator = {
   condition: ({ required }) => required,
   message: () => "Hotbar with this name already exists",
   validate: value => !HotbarStore.getInstance().getByName(value),
```
Nokel81, line 31: I agree that this should trim the input.

src/renderer/components/hotbar/hotbar-rename-command.tsx
```diff
@@ -0,0 +1,113 @@
...
+          escapeClearsValue={false}
+          placeholder="Rename hotbar"/>
+      </>
+    );
+  }
+
+  renderNameInput() {
+    return (
+      <>
+        <Input
```
Nokel81, line 88: This input probably should have `trim` on it.

src/common/hotbar-store.ts
vshakirova: Oki, should I throw an error or just `return` undefined?

Nokel81: I would log a warning and then return undefined

src/renderer/components/hotbar/hotbar-rename-command.tsx
vshakirova: What if I do it inside `setHotbarName()`?    
e.g.  `this.hotbars[idx].name = name.trim();`

Nokel81: if you use `trim` then the duplicate check is done after. So having both `trim` and `required` means that the input of `"   "` won't even call your functions.

src/common/hotbar-store.ts
vshakirova: Okay, not sure we have to check for unknown id, because before calling `setTorbarName` we check hotbar existence  here.. 
https://github.com/lensapp/lens/pull/3469/files/62e42c9774d8e2c16449e172119c98fb914c1ac3#diff-9c1d174748a991f0ca88f6194e58e846475d55bf4228295f3a741482582e870fR59

src/renderer/components/hotbar/hotbar-rename-command.tsx
aleksfront: @vshakirova Trim can be made right on `<Input/>`. When we solve it, this PR is good to go 🚀

vshakirova: oh I see. Thank you!

src/common/hotbar-store.ts
Nokel81: It is just better practice. We may add calls in the future that don't do that check, or they might be some other weirdness.

src/renderer/components/hotbar/hotbar-add-command.tsx
```diff
@@ -25,10 +25,10 @@ import { HotbarStore } from "../../../common/hotbar-store";
 import { CommandOverlay } from "../command-palette";
 import { Input, InputValidator } from "../input";
 
-const uniqueHotbarName: InputValidator = {
+export const uniqueHotbarName: InputValidator = {
   condition: ({ required }) => required,
   message: () => "Hotbar with this name already exists",
-  validate: value => !HotbarStore.getInstance().getByName(value),
+  validate: value => !HotbarStore.getInstance().getByName(value.trim()),
```
Nokel81: This doesn't need trim anymore since the value passed already has been trimmed (if trim is true on input).

vshakirova: my bad, thank you

src/common/hotbar-store.ts
```diff
@@ -153,6 +153,18 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
     }
   }
 
+  @action
+  setHotbarName(id: string, name: string) {
+    const index = this.hotbars.findIndex((hotbar) => hotbar.id === id);
+
+    if(index < 0) {
+      console.warn("Hotbar with id=" + id + " was not found");
```
Nokel81: nit, please follow the standard store logging format:

```
console.warn(`[HOTBAR-STORE]: cannot setHotbarName: unknown id`, { id });
```

extensions/pod-menu/package-lock.json
```diff
@@ -1248,6 +1249,12 @@
           "integrity": "sha512-lBN9zLN/oAf68o3zNXYrdCt1kP8WsiGW8Oo2ka41b2IM5JL/S1CTyX1rW0mb/zSuJun0ZUrDxx4sqvYS2FWzPA==",
           "dev": true
         },
+        "typed-emitter": {
```
Nokel81: Is it possible to revert these lock changes? if not don't worry.

vshakirova: yep, sorry for that. sometimes it happens

Nokel81: That it does.

Nokel81 (approved):

PR #3473, opened 2021-07-21
Fix local cluster delete flow

### LINKED ISSUE #3500: bug on deleting a cluster
**Describe the bug**
Can not remove cluster from lens on macos.

**To Reproduce**
1. Right click in cluster in the left panel.
2. Click on 'Delete'
3. It says `Deleting` forever. Cluster never gone. If I add the cluster again, it will keep disconnected whenever I switch to another cluster.

**Expected behavior**
Cluster is deleted.

**Screenshots**
Screen Shot 2021-07-23 at 19 26 18

- Lens Version:  5.1.1-latest.20210716.1
- OS: OSX

Nokel81:
- Now checks for the cluster being the current cluster and confirms the
  action

- `<ConfirmDialog>` is fixed to work with several confirmations happening
  at once

https://user-images.githubusercontent.com/8225332/126526469-6395fc86-e39a-4f08-a068-15fd355cee25.mov


aleksfront:
A few comments related to changes:
1. Need some more description about problem context. Do we have an issue what this PR solves?
2. Why we need to replace deleted context with new one? As a Lens user I find current way of deleting seems more preferable, because you "just need to remove the cluster" (maybe wrong). @jakolehm Ptal ^
3. We shouldn't use 2 separate confirmations for removing with repeatable `ConfirmDialog`. If we need to specify something on deletion its better to create a new dialog component.
4. Second confirmation layout needs to be fixed: better not to stack description and selector in one line. Let the flow goes from top to bottom with appropriate paddings. (I can make changes later if needed, but this is important).

Nokel81:
> A few comments related to changes:
> 
>     1. Need some more description about problem context. Do we have an issue what this PR solves?

Sure, we were seeing errors in Sentry (and then Jari noticed it himself) that if the cluster being removed was the `current-cluster` then the delete logic would fail. This PR is a fix for that. 


>     2. Why we need to replace deleted context with new one? As a Lens user I find current way of deleting seems more preferable, because you "just need to remove the cluster" (maybe wrong). @jakolehm Ptal ^

So the flow is the same if the cluster being deleted is not the `current-cluster`. But now that you say this. I remember that I forgot to check if the file being deleted from is one of the c&p files. In that case we might be able to just remove the file.

Furthermore, I decided that it was better to replace the `current-cluster` then just unset it because if it is unset then `kubectl` will start to error out when used by the user.

>     3. We shouldn't use 2 separate confirmations for removing with repeatable `ConfirmDialog`. If we need to specify something on deletion its better to create a new dialog component.

I figured that having the first confirmation up as soon as possible was preferable to waiting for lens to read the file. However, I do see your point about a new dialog component. If we do move in that direction I think I will spin off the fixes to `ConfirmDialog` into a seperate PR.

>     4. Second confirmation layout needs to be fixed: better not to stack description and selector in one line. Let the flow goes from top to bottom with appropriate paddings. (I can make changes later if needed, but this is important).

I agree that the layout needs to be fixed. And I am not completely convinced with how I am handing the picking of "clear current-cluster".

src/main/initializers/ipc.ts
```diff
@@ -77,27 +74,24 @@ export function initIpcMainHandlers() {
...
-    appEventBus.emit({ name: "cluster", action: "remove" });
+  ipcMainHandle(clusterDeleteHandler, (event, clusterId: ClusterId) => {
     const cluster = ClusterStore.getInstance().getById(clusterId);
 
     if (!cluster) {
       return;
     }
 
-    ClusterManager.getInstance().deleting.add(clusterId);
+    appEventBus.emit({ name: "cluster", action: "delete_frame" });
```
nevalla: do we really need this?

src/renderer/initializers/catalog.tsx
```diff
@@ -20,14 +20,160 @@
...
+      });
+    });
+}
+
 export function initCatalog() {
   initWebLinks();
+  initKubernetesClusters();
+}
+
+export async function deleteLocalCluster(clusterId: ClusterId): Promise<void> {
```
nevalla: This is quite a huge method, maybe refactor into smaller pieces?

nevalla:
Added couple of comments. Also, please coordinate with @aleksfront the confirmation dialog changes. What comes the confirmation itself, I've a mixed feelings. Current context is a bit out of Lens scope, but I do understand that kubectl won't work after this operation out-of-the-box.

src/main/initializers/ipc.ts
Nokel81: Nope, I can remove this.

src/renderer/initializers/catalog.tsx
Nokel81: Fair enough, will make into smaller pieces.

Nokel81:
I have split the delete function, but I am still not happy with how the second confirm dialog looks.

aleksfront:
Made refactoring as well as custom cluster delete dialog.


https://user-images.githubusercontent.com/9607060/130416966-3a72442e-5655-494f-ba3a-45d5ebb1bb4e.mov

@Nokel81 ptal for review ^

package.json
```diff
@@ -346,6 +346,7 @@
     "jest-canvas-mock": "^2.3.0",
     "jest-fetch-mock": "^3.0.3",
     "jest-mock-extended": "^1.0.16",
+    "json-to-pretty-yaml": "^1.2.2",
```
aleksfront, line 350: Used to keep spaces in tabs for kubeconfig file after rewriting its contents.

src/renderer/components/confirm-dialog/confirm-dialog.scss
```diff
@@ -33,7 +33,8 @@
     max-width: 50vw;
     min-width: 45 * $unit;
     background-color: white;
-    outline: $unit solid rgba(255, 255, 255, .15);
+    border-radius: $radius;
```
aleksfront, line 36: Some fine-tunings in dialog-related styles to keep all dialogs views consistent (rounded a bit).

src/renderer/components/delete-cluster-dialog/save-config.ts
```diff
@@ -0,0 +1,39 @@
...
+export async function saveKubeconfig(config: KubeConfig, path: string) {
+  lockFile.lock(path).then(async (release) => {
+    const tmpFilePath = tempy.file();
+    const contents = YAML.stringify(JSON.parse(config.exportConfig()));
+
+    await fs.promises.writeFile(tmpFilePath, contents);
+    await fs.promises.rename(tmpFilePath, path);
+    release();
+  }).catch((e) => {
+    throw new Error(`Failed to aquire lock file.\n${e}`);
```
jim-docker:
```suggestion
    throw new Error(`Failed to acquire lock file.\n${e}`);
```

jim-docker: temp file might still exist, should we delete it here?

src/renderer/components/delete-cluster-dialog/delete-cluster-dialog.tsx
```diff
@@ -0,0 +1,195 @@
...
+    const { cluster, config, isOpen } = dialogState;
+
+    if (!cluster || !config) return null;
+
+    const isCurrentContext = config.currentContext == cluster.contextName;
+    const contexts = config.contexts.filter(context => context.name !== cluster.contextName);
+
+    const warning = isCurrentContext ? (
+      <>
+        <p data-testid="context-warning">The <b>current-context</b> field from the kubeconfig file indicates a minikube context
```
jim-docker: When you remove the context that is the current-context using `kubectl` you get this message:
```
warning: this removed your active context, use "kubectl config use-context" to select a different one
```
We could phrase this similarly, like
This will remove your active context in your kubeconfig which will affect the operation of kubectl

Also, why is "minikube" specified here?

src/renderer/components/delete-cluster-dialog/delete-cluster-dialog.tsx
```diff
@@ -0,0 +1,195 @@
...
+        that will cease to exist after the change. This will affect the operation of kubectl.{" "}
+        {contexts.length > 0 && (
+          <b
+            className="cursor-pointer underline"
+            onClick={() => this.showContextSwitch = !this.showContextSwitch}
+          >Replace current context</b>
```
jim-docker: and this could be 
Select a different one

jim-docker (commented):
As @Nokel81 mentioned above do we need to show the delete dialog for contexts added by copy/paste?

Screen Shot 2021-08-23 at 7 25 39 PM

Seems odd to show the path to the Lens-internal kubeconfig, and to call it the user's ("your kubeconfig"). 

We should be able to reset the current-context to whatever we like?

Maybe "Are you sure you want to delete it? It can be re-added through the copy/paste mechanism" or something better worded. Though I guess if there are a number of contexts it gets messy re-adding one or more...

src/renderer/components/delete-cluster-dialog/delete-cluster-dialog.tsx
```diff
@@ -0,0 +1,195 @@
...
+        )}
+        </p>
+      </>
+    ) : (
+      <p>The contents of your kubeconfig file will be changed!</p>
```
jim-docker: This may be nit-picky but when there are no other contexts we could give the user the option to "Delete current context" (similar to "Replace current context") . This assumes that the user knows this refers to `current-context` in the kubeconfig. Or "Delete current-context" 

Though either way this leaves the kubeconfig with an empty contexts field, and cripples `kubectl`

aleksfront: Decided to skip this since both options will corrupt `kubectl`. Also, `kubectl` itself won't allow you to set empty `current-context`

Nokel81: Fair but you can do `kubectl config unset current-context`

aleksfront:
@jim-docker Made some refactoring and spell changes addressed in review.

1. Removing context which is `current-context` (checkbox is checked by default here):
removing current-context

2. Removing "regular" context:
removing regular context

3. Removing the latest context:
no contexts left

src/main/cluster.ts
```diff
@@ -712,4 +712,8 @@ export class Cluster implements ClusterModel, ClusterState {
   get imagePullSecret(): string {
     return this.preferences.imagePullSecret || "";
   }
+
+  isInLocalKubeconfig() {
+    return this.kubeConfigPath.includes(storedKubeConfigFolder());
```
Nokel81: This is wrong, I would say a more correct version would be to use `startsWith`

src/main/initializers/ipc.ts
```diff
@@ -79,27 +76,23 @@ export function initIpcMainHandlers() {
     }
   });
 
-  ipcMainHandle(clusterDeleteHandler, async (event, clusterId: ClusterId) => {
-    appEventBus.emit({ name: "cluster", action: "remove" });
```
Nokel81, line 83: Do we no longer want the event?

src/renderer/components/delete-cluster-dialog/delete-cluster-dialog.tsx
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  static open({ config, cluster }: Partial<DialogState>) {
+    dialogState.isOpen = true;
+    dialogState.config = config;
+    dialogState.cluster = cluster;
+  }
+
+  static close() {
+    dialogState.isOpen = false;
```
Nokel81, line 73: Why not make `isOpen` a `@computed`?

src/renderer/components/delete-cluster-dialog/delete-cluster-dialog.tsx
```diff
@@ -0,0 +1,227 @@
...
+    if (!this.showContextSwitch) return null;
+    const { cluster, config } = dialogState;
+    const contexts = config.contexts.filter(context => context.name !== cluster.contextName);
+
+    const options = [
+      ...contexts.map(context => ({
+        label: context.name,
+        value: context.name,
+      })),
+    ];
```
Nokel81, line 151: This is extraneous and wasteful, there is no need to spread a map as `.map` always creates a new array.

src/renderer/components/delete-cluster-dialog/save-config.ts
```diff
@@ -0,0 +1,42 @@
...
+    const contents = YAML.stringify(JSON.parse(config.exportConfig()));
+
+    await fs.promises.writeFile(tmpFilePath, contents);
+    await fs.promises.rename(tmpFilePath, path);
+    release();
+  }).catch(async (e) => {
+    await fs.unlink(tmpFilePath, noop);
+    throw new Error(`Failed to acquire lock file.\n${e}`);
+  });
+}
```
Nokel81: please do not mix promises and async functions

src/main/initializers/ipc.ts
aleksfront: Probably some merging conflict, don't remember removing those.

src/renderer/components/delete-cluster-dialog/save-config.ts
aleksfront: Can you be more specific here? What part of code wrong?

Nokel81: I think it is just ban form.

src/renderer/components/delete-cluster-dialog/delete-cluster-dialog.tsx
aleksfront: I don't see why `isOpen` should be computed. Afaik computed property is derived from other observable value. But here `isOpen` is observable by itself.

Nokel81: yes but there is no benefit of separating them. Namely, I suggest that we instead have something along the lines of:

```ts
private static readonly _isOpen = computed(() => Boolean(dialogState.cluster && dialogState.config));
static get isOpen() {
  return DeleteClusterDialog._isOpen.get();
}
```

jim-docker:
> @jim-docker Made some refactoring and spell changes addressed in review.
> 
> 1. Removing context which is `current-context` (checkbox is checked by default here):
>    removing current-context
> 2. Removing "regular" context:
>    removing regular context
> 3. Removing the latest context:
>    no contexts left

I like the first two but the third one (removing last context) still seems clunky. The warning could be "This will remove the last context in kubeconfig. There will be no active context." And just remove the "Select current-context" option and dropdown. ?

jim-docker:
I'm seeing this:
Screen Shot 2021-08-31 at 3 39 14 PM

jim-docker:
> As @Nokel81 mentioned above do we need to show the delete dialog for contexts added by copy/paste?
> 
> Screen Shot 2021-08-23 at 7 25 39 PM
> 
> Seems odd to show the path to the Lens-internal kubeconfig, and to call it the user's ("your kubeconfig").
> 
> We should be able to reset the current-context to whatever we like?
> 
> Maybe "Are you sure you want to delete it? It can be re-added through the copy/paste mechanism" or something better worded. Though I guess if there are a number of contexts it gets messy re-adding one or more...

I think this is still an issue too.

src/renderer/components/delete-cluster-dialog/delete-cluster-dialog.tsx
aleksfront: I guess we can leave it separated for now. If we want to change it, it's worth refactor few dialogs at once.

aleksfront:
> I'm seeing this:
> Screen Shot 2021-08-31 at 3 39 14 PM

Fixed.

aleksfront:
@jim-docker Latest changes regarding to warning messages:

1. Deleting internal cluster
internal cluster

2. No more contexts left
no more contexts

3. Removing "regular" cluster
regular context deleting

nevalla:
Not sure do we need to display `[ ] Select current context` if the context of the cluster is not the current context in Kubeconfig? But we can improve this in follow-up PR if needed.

nevalla (dismissed):

src/renderer/components/delete-cluster-dialog/delete-cluster-dialog.tsx
```diff
@@ -0,0 +1,249 @@
...
+    return (
+      <Dialog
+        className={styles.dialog}
+        isOpen={isOpen}
+        close={this.onClose}
+        onOpen={this.onOpen}
+      >
+        <div className={styles.dialogContent}>
+          <div>
+            Delete the <b>{cluster.getMeta().name}</b> context from <b>{cluster.kubeConfigPath}</b>?
```
jim-docker: I still think that if the cluster is in a Lens local config we shouldn't display the kubeconfig path. If `cluster.isInLocalKubeconfig()` then:
```
            Delete the <b>{cluster.getMeta().name}</b> context from the Lens local config file?
```

Not a blocker though

jim-docker:
When I'm about to delete the context which is also `current-context` I shouldn't be able to click the Delete Context button until I actually select a different one from the drop-down, right? Since I can click the Delete Context button this leaves my config in a broken state (stale `current-context`)

Screen Shot 2021-09-07 at 10 33 16 AM

jim-docker (approved):

PR #3475, opened 2021-07-21
Display hint to user on how to add items in an EditableList

### LINKED ISSUE #3451: Cannot add ClusterRoleBinding with user/group target
Create button is only enabled if I select ServiceAccount.

image

Nokel81:

fixes #3451 

https://user-images.githubusercontent.com/8225332/126553289-3de489ae-2844-4c89-9af6-daa4f4f31406.mov

Note: I think that the other display issues related to the items should be part of a different PR.

### DIFF
--- a/src/renderer/components/input/input.tsx
+++ b/src/renderer/components/input/input.tsx
@@ -27,7 +27,6 @@ import { Icon } from "../icon";
 import { Tooltip, TooltipProps } from "../tooltip";
 import * as Validators from "./input_validators";
 import type { InputValidator } from "./input_validators";
-import isString from "lodash/isString";
 import isFunction from "lodash/isFunction";
 import isBoolean from "lodash/isBoolean";
 import uniqueId from "lodash/uniqueId";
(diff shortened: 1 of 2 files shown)


src/renderer/components/input/input.tsx
```diff
@@ -262,6 +272,10 @@ export class Input extends React.Component<InputProps, State> {
         this.props.onSubmit?.(this.getValue(), evt);
         this.setDirtyOnChange.cancel();
         this.setState({ submitted: true });
+
+        if (this.input && typeof this.props.value !== "string") {
+          this.input.value = "";
+        }
```
aleksfront, line 278: Do we have use cases the `value` became non-string? It's clearly marked as `string` in props.

aleksfront (approved):
LGTM, but one question about input value check needs to be resolved (answered)

src/renderer/components/input/input.tsx
Nokel81: With the second check I try to check if this is a controlled vs uncontrolled instance. Since I can only reset the value in the uncontrolled state.

aleksfront: Thinking more about this maybe the method `getValue()` became misleading for now. As we see from it's name it should return a value, but shouldn't do any side effects like `this.input.value = "";`.

aleksfront: And is this check relates to showing and icon for user? (I'm just wanted to reconsider do we really need this dropping value operation)?

Nokel81: This isn't part of `getValue()` (though I admit that the github diff implies that it is). It is part of `onKeyDown`

aleksfront: Oh yes, my fault. Good to go!

Issue #3481, opened 2021-07-22
Remove legacy plus sign (+) from component directories


aleksfront:
For legacy purposes some of the directories inside `/components` are named with `+`. The meaning was originally to indicate components which get opened by `Routes`.

However, this decision creates problems to testing. Jest matcher doesn't understand plus sing in folder names and unable to find any tests there.

```
In /Users/alexfront/Projects/lens
  900 files checked.
  testMatch: **/__tests__/**/*.[jt]s?(x), **/?(*.)+(spec|test).[tj]s?(x) - 60 matches
  testPathIgnorePatterns: /node_modules/ - 900 matches
  testRegex:  - 0 matches
Pattern: src/renderer/components/+workloads-deployments/deployment-scale-dialog.test.tsx - 0 matches
```

![legacy folders](https://user-images.githubusercontent.com/9607060/126615099-980f0ea0-8c91-47d5-8eb8-baeef2931165.png)

We need to rename all folders to be without `+`.

jakolehm:
@aleksfront does it work if you escape `+` in a pattern match (`\+`).

aleksfront:
Thanks for the solution!

It works if 
1. Test address gets wrapped with quotes.
2. Plus sign gets escaped.

So this command will work:
```
yarn jest "src/renderer/components/\+welcome/__test__/welcome.test.tsx"
```

PR #3486, opened 2021-07-22
Fix workload overview status sorting

### LINKED ISSUE #3423: Alignment of counters for Workloads/Overview
**Describe the bug**
Is the alignment of Running/Pending counters should be shifted?
Previously it wasn't. If yes, please describe the order/behavior of it.

**To Reproduce**
1. Go to 'Workloads/Overview'

**Expected behavior**
All counters are aligned (and possibly sorted) into one equal column.

**Screenshots**
Screenshot 2021-07-19 at 11 34 02 AM

- Lens Version: `5.1.1-latest.20210716.1`
- OS: MacOS Big Sur 11.4
- Installation method (e.g. snap or AppImage in Linux): `brew`

**Logs:**
Nothing was found in debug mode.

**Kubeconfig:**
N/A

**Additional context**
N/A

aleksfront:
1. Sorting workload statuses in same manner (`Running`, `Succeeded`, `Scheduled` goes on top).
2. Fixed badge alignment styles.

Before:
misaligned badges

After:
aligned badges

Fixes #3423

### DIFF
--- a/src/renderer/components/+workloads-cronjobs/cronjob.store.ts
+++ b/src/renderer/components/+workloads-cronjobs/cronjob.store.ts
@@ -34,7 +34,7 @@ export class CronJobStore extends KubeObjectStore<CronJob> {
   }
 
   getStatuses(cronJobs?: CronJob[]) {
-    const status = { suspended: 0, scheduled: 0 };
+    const status = { scheduled: 0, suspended: 0 };
 
     cronJobs.forEach(cronJob => {
       if (cronJob.spec.suspend) {
--- a/src/renderer/components/+workloads-pods/pods.store.ts
+++ b/src/renderer/components/+workloads-pods/pods.store.ts
@@ -80,7 +80,7 @@ export class PodsStore extends KubeObjectStore<Pod> {
   }
 
   getStatuses(pods: Pod[]) {
-    return countBy(pods.map(pod => pod.getStatus()));
+    return countBy(pods.map(pod => pod.getStatus()).sort().reverse());
   }
 
   getPodKubeMetrics(pod: Pod) {
--- a/src/renderer/components/__tests__/cronjob.store.test.ts
+++ b/src/renderer/components/__tests__/cronjob.store.test.ts
@@ -1,0 +1,115 @@
+/**
+ * Copyright (c) 2021 OpenLens Authors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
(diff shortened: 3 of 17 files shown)


src/renderer/components/+workloads-pods/pods.store.ts
```diff
@@ -80,7 +80,7 @@ export class PodsStore extends KubeObjectStore<Pod> {
   }
 
   getStatuses(pods: Pod[]) {
-    return countBy(pods.map(pod => pod.getStatus()));
+    return countBy(pods.map(pod => pod.getStatus()).sort().reverse());
```
aleksfront, line 83: This will automatically sort statuses in a needed order:
```
Succeeded -> Running -> Pending -> Failed -> Evicted
```

src/renderer/components/__tests__/cronjob.store.test.ts
```diff
@@ -0,0 +1,115 @@
+/**
```
aleksfront, line 1: All tests sit in parent `/components/` folder until this #3481  won't be resolved. Otherwise it's hard run (read impossible) tests locally.

src/renderer/components/+workloads-cronjobs/cronjob.store.ts
```diff
@@ -34,7 +34,7 @@ export class CronJobStore extends KubeObjectStore<CronJob> {
   }
 
   getStatuses(cronJobs?: CronJob[]) {
-    const status = { suspended: 0, scheduled: 0 };
```
Nokel81, line 37: Why does the order matter?

Nokel81: This is an object, so I think we should specify the order by extracting to an array. Or at least add a comment above all of these saying that the order is important.

aleksfront: Since we're decided to put all "green" statuses on top (show them first), I've moved `scheduled` to be first one since it's "green".

Nokel81: I still think that should be something that is commented or specified somewhere else. Since objects are key/value structures which don't really have a conceptual ordering.

aleksfront: Yes, objects didn't have conceptual ordering, but they're actually have an order for now. The results are always the same + we have tests now checking ordering.

Here's mentions about object ordering https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#objects_vs._maps

As an alternative, I've converted statuses to use Maps, but this looks more messy and verbose without visible benefits.

Nokel81: Okay, I was thinking more of having a defined ordering in `overview-workload-status.tsx` but that might be a bit too much. Maybe something for the future.

I don't particularly like the fact that this ordering is required to be static but I concur that should be for the future.

jakolehm (approved):
LGTM

PR #3487, opened 2021-07-22
Add accessibleNamespaces to KubernetesClusterModel

### LINKED ISSUE #3485: Include accessible namespaces in KubernetesCluster Catalog model
We like to manage clusters with a extensions. We can create a cluster, but it is not possible to configure automatically the accessible namespaces.

**Why is this needed**:
We can get these namespaces from our backend. We don't want to allow users to list namespaces.

- Kubernetes distribution: Kubeadm
- Desktop OS: Linux

farodin91:
Fixes #3485

### DIFF
--- a/src/common/catalog-entities/kubernetes-cluster.ts
+++ b/src/common/catalog-entities/kubernetes-cluster.ts
@@ -43,6 +43,7 @@ export interface KubernetesClusterPrometheusMetrics {
 export interface KubernetesClusterSpec extends CatalogEntitySpec {
   kubeconfigPath: string;
   kubeconfigContext: string;
+  accessibleNamespaces?: string[];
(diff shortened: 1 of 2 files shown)


Nokel81:
I don't think we should display the accessible cluster settings if the entities' model field is set.

farodin91:
Cluster Settings are already hidden, if you set source to non local. Isn't this enough?

Nokel81:
We have discussed this internally and have decided that all `KubernetesCluster` will have a settings menu item. And extension developers can opt into hiding the "all source" settings on a per settings basis. 

However, I do see that this work would probably fit better into the PR that implements that so I think that this PR is good to go.

Nokel81 (approved):

PR #3492, opened 2021-07-22
Validate and set defaults for HelmChart creation

### LINKED ISSUE #3491: TypeError: Cannot read property 'toLowerCase' of undefined
Sentry Issue: [K8SLENS-53V](https://sentry.io/organizations/k8slens/issues/2527718540/?referrer=github_integration)

```
React ErrorBoundary Error: Cannot read property 'toLowerCase' of undefined
  at None (/build/Lens.js:475926:16)
  at List (/build/Lens.js:506006:30)
  at AutoSizer (/build/Lens.js:507234:5)
  at virtual_list_VirtualList (/build/Lens.js:507408:9)
  at Table (/build/Lens.js:507602:9)
...
(18 additional frame(s) were not displayed)

TypeError: Cannot read property 'toLowerCase' of undefined
  at renderTableContents (/build/Lens.js:535551:85)
  at ItemListLayout.getRow (/build/Lens.js:519254:13)
  at None (/build/Lens.js:507488:17)
  at None (/build/Lens.js:475926:49)
  at None (/build/Lens.js:475890:25)
...
(27 additional frame(s) were not displayed)
```

Nokel81:

fixes #3491


src/renderer/api/endpoints/helm-charts.api.ts
```diff
@@ -83,81 +85,303 @@ export async function getChartValues(repo: string, name: string, version: string
...
+    this.urls = value.urls;
+    this.dependencies = value.dependencies;
+    this.type = value.type;
+
     autoBind(this);
   }
 
-  static create(data: any) {
-    return new HelmChart(data);
+  static create(data: RawHelmChart, { onError = "throw" }: HelmChartCreateOpts = {}): HelmChart | undefined {
```
nevalla: since we are not handling onError == "log", should we use `{ throw: true }` instead?

nevalla (commented):
Changes looks good to me. By just looking the code, I'm not sure how is this fixing the issue. What is happening if chart is not valid? It's not added to charts list?

src/renderer/api/endpoints/helm-charts.api.ts
Nokel81: Sure, I meant the `console.warn` on line 339 to be the handling of `onError = "log"`

Nokel81:
> By just looking the code, I'm not sure how is this fixing the issue. What is happening if chart is not valid? It's not added to charts list?

So https://github.com/lensapp/lens/pull/3492/files#diff-57f9f6ee587d04a73c86ae4a9d6c6d2ac70848d5747ad0f07089a9f71e410ad9R50-R51 catches and logs any errors about creating the HelmCharts. So the invalid chart (that doesn't have a repo set) would just not be shown in the list.

src/renderer/api/endpoints/helm-charts.api.ts
nevalla: yes it does, and it handles also `onError="foo"`. So, since only `onError="throw"` has effect, it sounds more like boolean value, so passing `{ throw:true|false }` would be more logical. But this is only semantics

nevalla (approved):

PR #3505, opened 2021-07-23
Fix auto cleanup of Extension's IPC classes

Nokel81:
- Add logging for registering/removing IPC handlers & listeners

- Fix navigation logs

fixes a bug where uninstalling and then reinstalling the same extension removes but then doesn't add IPC handlers.

### DIFF
--- a/src/extensions/ipc/ipc-main.ts
+++ b/src/extensions/ipc/ipc-main.ts
@@ -30,6 +30,8 @@ import logger from "../../main/logger";
 export abstract class IpcMain extends IpcRegistrar {
   constructor(extension: LensMainExtension) {
     super(extension);
+
+    // Call the static method on the bottom child class.
     extension[Disposers].push(() => (this.constructor as typeof IpcMain).resetInstance());
   }
 
--- a/src/extensions/ipc/ipc-main.ts
+++ b/src/extensions/ipc/ipc-main.ts
@@ -25,11 +25,12 @@ import type { LensMainExtension } from "../lens-main-extension";
 import type { Disposer } from "../../common/utils";
 import { once } from "lodash";
 import { ipcMainHandle } from "../../common/ipc";
+import logger from "../../main/logger";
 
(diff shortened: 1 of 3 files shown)


src/extensions/ipc/ipc-main.ts
```diff
@@ -25,11 +25,12 @@ import type { LensMainExtension } from "../lens-main-extension";
 import type { Disposer } from "../../common/utils";
 import { once } from "lodash";
 import { ipcMainHandle } from "../../common/ipc";
+import logger from "../../main/logger";
 
 export abstract class IpcMain extends IpcRegistrar {
   constructor(extension: LensMainExtension) {
     super(extension);
-    extension[Disposers].push(() => IpcMain.resetInstance());
+    extension[Disposers].push(() => (this.constructor as typeof IpcMain).resetInstance());
```
jim-docker, line 35: that's not obvious :)

src/extensions/ipc/ipc-renderer.ts
```diff
@@ -28,7 +28,7 @@ import { once } from "lodash";
 export abstract class IpcRenderer extends IpcRegistrar {
   constructor(extension: LensRendererExtension) {
     super(extension);
-    extension[Disposers].push(() => IpcRenderer.resetInstance());
+    extension[Disposers].push(() => (this.constructor as typeof IpcRenderer).resetInstance());
```
jakolehm, line 33: Is this the fix?

jim-docker (dismissed):

src/extensions/ipc/ipc-renderer.ts
Nokel81: Yes

src/extensions/ipc/ipc-main.ts
Nokel81: Not it isn't which is why I did the other thing before and no one caught it 😆

jim-docker (approved):

PR #3506, opened 2021-07-23
Use user configured kubectl path in node and pod menu items

### LINKED ISSUE #2948: "Shell" command doesn't respect kubectl location
**Describe the bug**
My kubectl binary is set in preferences with the "Path to kubectl binary" setting.  However, in the Pods screen, when I click "Shell", it opens a terminal window using just `kubectl` (instead of the full path to my kubectl binary).

**To Reproduce**
1. Go to settings, set a value for "Path to kubectl binary"
2. Go to Workloads > Pods. Click the menu next to a pod and click "Shell"
3. The terminal opens with the command `exec kubectl exec ...`. (This fails because `kubectl` cannot be found)

**Expected behavior**
The terminal opens with the command `exec /my/kubectl/path exec ...`

- Lens Version:
- OS: OSX
- Installation method (e.g. snap or AppImage in Linux): not sure

**Additional context**
Don't fully understand the codebase, but it seems like the menu bar does not attempt to retrieve the kubectl path from the settings: https://github.com/lensapp/lens/blob/master/extensions/pod-menu/src/shell-menu.tsx#L52

Nokel81:

fixes #2948 

Note:
- This PR adds to the extension API so that extensions can have a read-only view of user settings.

### DIFF
--- a/extensions/node-menu/src/node-menu.tsx
+++ b/extensions/node-menu/src/node-menu.tsx
@@ -21,7 +21,6 @@
 
 import React from "react";
 import { Common, Renderer } from "@k8slens/extensions";
-import { inspect } from "util";
 
 type Node = Renderer.K8sApi.Node;
 
@@ -48,7 +47,7 @@ export function NodeMenu(props: NodeMenuProps) {
 
   if (!node) return null;
   const nodeName = node.getName();
-  const kubectlPath = inspect(UserPreferences.getKubectlPath(), false, null, false);
+  const kubectlPath = JSON.stringify(UserPreferences.getKubectlPath()); // this call escapes spaces and quotes
 
   const sendToTerminal = (command: string) => {
     terminalStore.sendCommand(command, {
--- a/extensions/node-menu/src/node-menu.tsx
+++ b/extensions/node-menu/src/node-menu.tsx
@@ -45,9 +45,12 @@ export interface NodeMenuProps extends Renderer.Component.KubeObjectMenuProps<No
 export function NodeMenu(props: NodeMenuProps) {
   const { object: node, toolbar } = props;
 
-  if (!node) return null;
+  if (!node) {
+    return null;
+  }
+
   const nodeName = node.getName();
-  const kubectlPath = JSON.stringify(App.Preferences.getKubectlPath()); // this call escapes spaces and quotes
+  const kubectlPath = App.Preferences.getKubectlPath() || "kubectl";
 
   const sendToTerminal = (command: string) => {
     terminalStore.sendCommand(command, {
--- a/extensions/node-menu/src/node-menu.tsx
+++ b/extensions/node-menu/src/node-menu.tsx
@@ -20,7 +20,8 @@
  */
 
 import React from "react";
-import { Renderer } from "@k8slens/extensions";
+import { Common, Renderer } from "@k8slens/extensions";
+import { inspect } from "util";
 
 type Node = Renderer.K8sApi.Node;
 
@@ -34,6 +35,9 @@ const {
   },
   Navigation
 } = Renderer;
+const {
+  UserPreferences,
+} = Common;
 
 
 export interface NodeMenuProps extends Renderer.Component.KubeObjectMenuProps<Node> {
@@ -44,6 +48,7 @@ export function NodeMenu(props: NodeMenuProps) {
 
   if (!node) return null;
   const nodeName = node.getName();
+  const kubectlPath = inspect(UserPreferences.getKubectlPath(), false, null, false);
 
   const sendToTerminal = (command: string) => {
     terminalStore.sendCommand(command, {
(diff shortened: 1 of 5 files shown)


extensions/node-menu/src/node-menu.tsx
```diff
@@ -62,15 +67,15 @@ export function NodeMenu(props: NodeMenuProps) {
   };
 
   const cordon = () => {
-    sendToTerminal(`kubectl cordon ${nodeName}`);
+    sendToTerminal(`${kubectlPath} cordon ${nodeName}`);
```
jim-docker, line 72: Is this going to work if `kubectlPath` has a space in it?

Nokel81: The call to `inspect` above was supposed to catch any possible cases like that. And it does, but how it handles strings with both `'` and `"` in them (by using `) is wrong and I should fix it.

nevalla:
Previously when `kubectl` command is sent to the terminal, it will use versions for kubectl that matches with kubernetes server version. Now this will change it so that it will use bundled kubectl instead. I think this is unwanted change?

chenhunghan:
> Previously when `kubectl` command is sent to the terminal, it will use versions for kubectl that matches with kubernetes server version. Now this will change it so that it will use bundled kubectl instead. I think this is unwanted change?

Looks like it will fallback to `bundledKubectlPath` which should match the server version, see `getKubectlPath`

src/extensions/common-api/index.ts
```diff
@@ -34,4 +35,5 @@ export {
   Store,
   Types,
   Util,
+  UserPreferences,
```
jakolehm: I think `App.Preferences` would make more sense (and we wouldn't need to create another top-level namespace). Wdyt?

nevalla:
> > Previously when `kubectl` command is sent to the terminal, it will use versions for kubectl that matches with kubernetes server version. Now this will change it so that it will use bundled kubectl instead. I think this is unwanted change?
> 
> Looks like it will fallback to `bundledKubectlPath` which should match the server version, see `getKubectlPath`

Lens sets kubectl binary folder into `PATH` for terminal and when running `kubectl` command it will use correct version. Now we are switching to use bundled kubectl binary for these commands.

With Lens 5.1.3
```
➜  ~ which kubectl                                                                                             
/Users/lauri/Library/Application Support/Lens/binaries/kubectl/1.19.7/kubectl
➜  ~ kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.7", GitCommit:"1dd5338295409edcfff11505e7bb246f0d325d15", GitTreeState:"clean", BuildDate:"2021-01-13T13:23:52Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.10+k3s1", GitCommit:"72e8196c9b9eb277f56951db6d2d9e5f9f70feac", GitTreeState:"clean", BuildDate:"2021-04-16T18:36:38Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
➜  ~ exec kubectl exec -i -t -n lens-metrics prometheus-0 -c prometheus "--" sh -c "clear; (bash || ash || sh)"
```

With this PR
```
➜  ~ kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.7", GitCommit:"1dd5338295409edcfff11505e7bb246f0d325d15", GitTreeState:"clean", BuildDate:"2021-01-13T13:23:52Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.10+k3s1", GitCommit:"72e8196c9b9eb277f56951db6d2d9e5f9f70feac", GitTreeState:"clean", BuildDate:"2021-04-16T18:36:38Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
➜  ~ exec "/Users/lauri/Projects/lens/binaries/client/darwin/x64/kubectl" exec -i -t -n lens-metrics prometheus-0 -c prometheus -- sh -c "clear; (bash || ash || sh)"
➜  ~ /Users/lauri/Projects/lens/binaries/client/darwin/x64/kubectl version                                                                                           
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.15", GitCommit:"73dd5c840662bb066a146d0871216333181f4b64", GitTreeState:"clean", BuildDate:"2021-01-13T13:22:41Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.10+k3s1", GitCommit:"72e8196c9b9eb277f56951db6d2d9e5f9f70feac", GitTreeState:"clean", BuildDate:"2021-04-16T18:36:38Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
```

src/extensions/common-api/index.ts
Nokel81: sure

Nokel81:
@nevalla Then there seems to be a different issue at play here too because if you look at the original issue, the user reported that their configured path was not being used.

Nokel81:
> Previously when `kubectl` command is sent to the terminal, it will use versions for kubectl that matches with kubernetes server version. Now this will change it so that it will use bundled kubectl instead. I think this is unwanted change?

Ah I think I understand your concern now. I want to use the setting if specified but not use the fallback if it isn't.

Nokel81:
> Lens sets kubectl binary folder into PATH for terminal and when running kubectl command it will use correct version. Now we are switching to use bundled kubectl binary for these commands.

I am not switching to the bundled version though. I am switching to the specified version in settings that the user said to use instead.

nevalla (approved):
LGTM

PR #3511, opened 2021-07-23
increase timeout for getPortFrom()

jim-docker:
5s may not be enough time to get a port through `kubectl proxy`, particularly on Windows, where a number of "failed to retrieve port from stream" Sentry events are occurring. Since it has been hard to reproduce the issue this is a first attempt at a solution.


Nokel81 (approved):

Nokel81:
If you run `kubectl proxy` manually. How long does it take to start?

jim-docker:
> If you run `kubectl proxy` manually. How long does it take to start?

About a second or so, but that was when I couldn't reproduce the issue. The one time I saw it was right after a cold start of my windows machine, but haven't timed kubectl proxy then. Will try. Also, looking at the recent Sentry events, the breadcrumbs often (most times?) show `[LENS-VIEW]: remove dashboard` log message just before the port failure exception, which is odd...

PR #3526, opened 2021-07-29
[Proposal]: Change topbar X to back button

jakolehm:
image
image


Nokel81:
I like the concept but the current implementation is a bit confusing to me since it seems to imply that I click it to go back to the text listed. So I find it confusing that on the Catalog it has a "< Catalog" button.

aleksfront:
Closing this in favor of #3696

PR #3529, opened 2021-07-29
Simplify most of the KubeObjectDetailRegistry initial values 

Nokel81:

blocked by https://github.com/lensapp/lens/pull/3528

### DIFF
--- a/src/renderer/initializers/kube-object-detail-registry.tsx
+++ b/src/renderer/initializers/kube-object-detail-registry.tsx
@@ -20,10 +20,8 @@
  */
 
 import React from "react";
-import type * as endpoints from "../api/endpoints";
-import type { KubeObject } from "../api/kube-object";
 import { KubeObjectDetailRegistry } from "../api/kube-object-detail-registry";
-import { HpaDetails } from "../components/+config-autoscalers";
+import { HpaDetails, HpaDetailsProps } from "../components/+config-autoscalers";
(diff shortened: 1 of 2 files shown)


src/renderer/initializers/kube-object-detail-registry.tsx
```diff
@@ -64,393 +61,395 @@ export function intiKubeObjectDetailRegistry() {
...
       },
       {
         kind: "HorizontalPodAutoscaler",
         apiVersions: ["autoscaling/v2beta1"],
         priority: 5,
         components: {
-          Details: (props: KubeObjectDetailsProps<KubeObject>) => <KubeEventDetails {...props}/>,
-        }
+          // Note: this line is left in the long form as a validation that this usecase is valid
+          Details: (props: KubeObjectDetailsProps) => <KubeEventDetails {...props}/>,
```
nevalla, line 74: 🤔 do we need props here? All other cases does not have any props for `KubeEventDetails`

nevalla (approved):
LGTM

src/renderer/initializers/kube-object-detail-registry.tsx
Nokel81: As I tried to explain with the comment on line 73 I left these two lines with props as a way to make sure that we don't break extension developers that use either of these styles. We have in the past so I think it would be good to have all three versions in our code.

PR #3530, opened 2021-07-29
Move cluster related types and function into seperate files

Nokel81:
- This is with a mind to ban circular dependencies

- With this change all the types and helper functions can be imported
  from anywhere without circular imports

Signed-off-by: Sebastian Malton <sebastian@malton.name>

Part of splitting apart #3410

PR #3532, opened 2021-07-29
Move Hotbar types to seperate file

Nokel81:

Blocked by #3530

### DIFF
--- a/src/common/hotbar-store.ts
+++ b/src/common/hotbar-store.ts
@@ -22,37 +22,16 @@
 import { action, comparer, observable, makeObservable } from "mobx";
 import { BaseStore } from "./base-store";
 import migrations from "../migrations/hotbar-store";
-import * as uuid from "uuid";
 import isNull from "lodash/isNull";
 import { toJS } from "./utils";
 import { CatalogEntity } from "./catalog";
 import { catalogEntity } from "../main/catalog-sources/general";
-
-export interface HotbarItem {
-  entity: {
-    uid: string;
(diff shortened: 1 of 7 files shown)


src/common/hotbar-store.ts
```diff
@@ -89,18 +68,16 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
...
-
   @action
   protected fromStore(data: Partial<HotbarStoreModel> = {}) {
     if (!data.hotbars || !data.hotbars.length) {
-      this.hotbars = [{
-        id: uuid.v4(),
-        name: "Default",
-        items: this.defaultHotbarInitialItems,
-      }];
+      const hotbar = getEmptyHotbar("default");
```
jim-docker: Was "Default", now is "default" ? I guess we've always set or required the name so "Default" never actually appears anywhere?

jim-docker (dismissed):

src/common/hotbar-store.ts
Nokel81: Good catch, I think I should leave it as `Default` so as to not change.

jim-docker (approved):
When would I see the "Default" hotbar in Lens?

Nokel81:
> When would I see the "Default" hotbar in Lens?

On a fresh install of lens, though the `"default"` hotbar from `src/migrations/hotbar-store/5.0.0-alpha.0.ts` might be shown instead. Perhaps I should change that name too?

PR #3536, opened 2021-07-30
Rollout support

### LINKED ISSUE #498: Rollout Support

Support for rollouts for deployments, statefulsets, and daemonsets.

A minimal implementation could be additional context/kebab menu items for deployments/etc with restart; pause; resume; undo rollout entries.

**Why is this needed**:

* Pause/resume/undo rollouts (including making changes without applying them immediately)
* Do a rolling restart of a deployment/statefulset/daemonset.
* See status for in-progress rollouts — this is mostly in place already.

- Kubernetes distribution: Kops
- Desktop OS: macOS

vshakirova:
The attempt to implement rollout to Deployments/Daemonsets/Replicasets. 
If I'm not mistaken API for restart/pause/rollback/resume e.t.c doesn't exist for now and in this case, verbose output helped me. 
The main problem is rollback. https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/staging/src/k8s.io/kubectl/pkg/polymorphichelpers/rollback.go#L99
pause/resume deployment 
https://github.com/kubernetes/kubernetes/blob/5b671d41c2fa89314eb9710dddd4f6a22c530952/test/integration/deployment/util.go#L51

Implementation for deployments has complex logic and I thought that using the terminal, in this case, might be better. 
Need your advice. Thanks!

fixes: https://github.com/lensapp/lens/issues/498


aleksfront:
@vshakirova I don't see anything bad with your design. Just thinking if we can move duplicated parts of rollout methods into separate `abstract` class, say `WorkloadKubeApi`. To avoid copypasting `restart`, `undo`, `history` and other commands.

So `DeploymentApi` can be declared like this
```
export class DeploymentApi extends WorkloadKubeApi<Deployment> {}
```

src/renderer/components/+workloads/rollout-menu.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { Icon } from "../icon";
+import { ScaleDialog } from "./scale-dialog";
+import { apiManager } from "../../api/api-manager";
+
+interface Props<T extends WorkloadKubeObject> {
+  workloadKubeObject: T;
+  interactive?: boolean;
+}
+
+export function renderRestart<T extends WorkloadKubeObject>(workloadKubeObject: T, api: any, interactive?: boolean) {
```
aleksfront: I guess we need to provide some information to `api` argument type, not `any`

src/renderer/components/+workloads/rollout-menu.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  if (!workloadKubeObject) {
+    return null;
+  }
```
aleksfront: This can be removed. Check is already done in parent `RolloutMenu` component.

src/renderer/components/+workloads/rollout-menu.tsx
```diff
@@ -0,0 +1,100 @@
...
+        <p>
+          Are you sure you want to restart {workloadKubeObject.kind} <b>{workloadKubeObject.getName()}</b>?
+        </p>
+      ),
+    })}>
+      <Icon material="autorenew" tooltip="Restart" interactive={interactive}/>
+      <span className="title">Restart</span>
+    </MenuItem>
+  );
+}
```
aleksfront: This should be moved to separate component as all of the rollout commands will share exact same code.

src/renderer/components/+workloads/rollout-menu.tsx
vshakirova: Yeah, everything did not go according to plan) 
Different resources have similar logic, but not the same.

src/common/k8s-api/endpoints/controller-revision.api.ts
```diff
@@ -0,0 +1,53 @@
...
+import type { KubeJsonApiData } from "../kube-json-api";
+
+export class ControllerRevisionApi extends KubeApi<ControllerRevision> {
+
+  async getRevisions(params: { namespace: string; name: string }) {
+    return await this.list({ namespace: params.namespace }, { labelSelector: `name=${params.name}` });
+  }
+}
+
+export class ControllerRevision extends WorkloadKubeObject {
```
vshakirova, line 33: I will change it to KubeObject

src/common/k8s-api/endpoints/controller-revision.api.ts
```diff
@@ -0,0 +1,53 @@
...
+  declare data: string;
+
+  getRevisionNumber() {
+    return this.revision;
+  }
+}
+
+export const controllerRevisionApi = new ControllerRevisionApi({
+  objectConstructor: ControllerRevision,
+});
```
Nokel81, line 53: This will need to be updated to be like the other ones. Such as in `pods.api.ts`

src/common/k8s-api/endpoints/daemon-set.api.ts
```diff
@@ -22,11 +22,14 @@
...
 import { autoBind } from "../../utils";
-import { KubeApi } from "../kube-api";
 import { metricsApi } from "./metrics.api";
 import type { KubeJsonApiData } from "../kube-json-api";
 import type { IPodContainer, IPodMetrics } from "./pods.api";
 import { isClusterPageContext } from "../../utils/cluster-id-url-parsing";
+import { WorkloadKubeApi } from "../../../renderer/components/+workloads/api/workload-kube-api";
+
+export class DaemonSetApi extends WorkloadKubeApi<DaemonSet> {
+}
```
Nokel81, line 32: Why was this changed?

src/common/k8s-api/endpoints/deployment.api.ts
```diff
@@ -51,21 +49,32 @@ export class DeploymentApi extends KubeApi<Deployment> {
...
-  restart(params: { namespace: string; name: string }) {
+  pause(params: { namespace: string; name: string }) {
     return this.request.patch(this.getUrl(params), {
       data: {
         spec: {
-          template: {
-            metadata: {
-              annotations: {"kubectl.kubernetes.io/restartedAt" : moment.utc().format()}
-            }
-          }
```
Nokel81, line 62: Please revert, this is a breaking change.

src/common/k8s-api/endpoints/replica-set.api.ts
```diff
@@ -117,6 +120,14 @@ export class ReplicaSet extends WorkloadKubeObject {
     return this.status.readyReplicas || 0;
   }
 
+  getRevisionNumber(): number {
+    return parseInt(this.metadata.annotations[revisionAnnotation]);
```
Nokel81, line 124: Do we really want to return `NaN` if that annotation is missing?

src/common/k8s-api/endpoints/stateful-set.api.ts
```diff
@@ -21,13 +21,13 @@
...
 import { autoBind } from "../../utils";
-import { KubeApi } from "../kube-api";
 import { metricsApi } from "./metrics.api";
 import type { IPodMetrics } from "./pods.api";
 import type { KubeJsonApiData } from "../kube-json-api";
 import { isClusterPageContext } from "../../utils/cluster-id-url-parsing";
+import { WorkloadKubeApi } from "../../../renderer/components/+workloads/api/workload-kube-api";
 
-export class StatefulSetApi extends KubeApi<StatefulSet> {
+export class StatefulSetApi extends WorkloadKubeApi<StatefulSet> {
```
Nokel81, line 30: Why are you changing these?

src/renderer/components/+workloads-daemonsets/daemonset-details.tsx
```diff
@@ -58,6 +62,17 @@ export class DaemonSetDetails extends React.Component<Props> {
     this.metrics = null;
   });
 
+  @disposeOnUnmount
+  revisionLoader = autorun(async () => {
+    const { object: daemonSet } = this.props;
+    const revisionHistory = await controllerRevisionApi.getRevisions({
+      name: daemonSet.getName(),
+      namespace: daemonSet.getNs()
+    }) as ControllerRevision[];
```
Nokel81, line 71: This cast should be unnecessary `getRevisions`'s return type should be `ControllerRevision[] | null`

src/renderer/components/+workloads-daemonsets/daemonset-details.tsx
```diff
@@ -58,6 +62,17 @@ export class DaemonSetDetails extends React.Component<Props> {
...
+
+    this.revisionHistory = observable.array(revisionHistory);
```
Nokel81, line 73: This is redundent, please change to `this.revisionHistory.replace(revisionHistory)`

src/renderer/components/+workloads-daemonsets/daemonsets.tsx
```diff
@@ -91,7 +93,18 @@ export class DaemonSets extends React.Component<Props> {
...
 }
+
+export function DaemonSetMenu(props: KubeObjectMenuProps<DaemonSet>) {
+  const { object } = props;
+
+  return (
+    <>
+      <RolloutMenu workloadKubeObject={object}/>
+    </>
+  );
```
Nokel81, line 109: nit

```suggestion
  return <RolloutMenu workloadKubeObject={object}/>;
```

src/renderer/components/+workloads-deployments/deployments.store.ts
```diff
@@ -67,6 +68,10 @@ export class DeploymentStore extends KubeObjectStore<Deployment> {
       .getByLabel(deployment.getTemplateLabels())
       .filter(pod => pod.getNs() === deployment.getNs());
   }
+
+  getRelatedReplicas(deployment: Deployment) {
+    return replicaSetStore.getReplicaSetsByOwner(deployment);
+  }
```
Nokel81, line 74: Would you mind doing this in a different PR? I don't see why it is necessary for rollout support.

src/renderer/components/+workloads-deployments/deployment-details.tsx
```diff
@@ -143,7 +141,7 @@ export class DeploymentDetails extends React.Component<Props> {
         <PodDetailsTolerations workload={deployment}/>
         <PodDetailsAffinities workload={deployment}/>
         <ResourceMetricsText metrics={this.metrics}/>
-        <DeploymentReplicaSets replicaSets={replicaSets}/>
```
Nokel81, line 146: Do we no longer have replica sets?

Nokel81 (changes requested):
Start of a review

src/common/k8s-api/endpoints/daemon-set.api.ts
vshakirova: Scale and restart work similar for daemontSet/DeploymentSet/Statetfulset. It is just an attempt to get rid of duplicate code. I've moved `Restart` functionality to `WorkloadKubeApi` abstract class.

src/common/k8s-api/endpoints/deployment.api.ts
vshakirova: we still have restart functionality, because we extend `WorkloadKubeApi`. These are just additional functions.

src/common/k8s-api/endpoints/stateful-set.api.ts
vshakirova: My intention was to get rid of duplicate code

src/renderer/components/+workloads-daemonsets/daemonsets.tsx
vshakirova: much better, thank you!

src/renderer/components/+workloads-deployments/deployment-details.tsx
vshakirova: We have, I just rename it because it seems like here (https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/staging/src/k8s.io/kubectl/pkg/polymorphichelpers/rollback.go#L212) they get a list of replicaSets and chose appropriate for rollback. I guess there is no need to duplicate the same information.

src/renderer/components/+workloads-deployments/deployments.store.ts
vshakirova: yes of course, I'll revert it

src/common/k8s-api/endpoints/replica-set.api.ts
vshakirova: Agree. Maybe `parseInt(this.metadata.annotations[revisionAnnotation]) || ""`. Wdyt?

Nokel81: Given that this is only really used for display, I don't see the benefit of parsing at all. That way we could display `"-"` when missing.

src/common/k8s-api/endpoints/stateful-set.api.ts
Nokel81: Okay that seems fair, I would have made a free function that such methods could just call instead.

src/renderer/components/+workloads-deployments/deployment-revision-history.tsx
```diff
@@ -19,89 +19,90 @@
...
   name = "name",
   namespace = "namespace",
   pods = "pods",
   age = "age",
+  revision = "revision",
 }
 
 interface Props {
-  replicaSets: ReplicaSet[];
+  object: ReplicaSet[];
```
Nokel81, line 43: why the rename? Also if it is a list then the parameter should be plural

src/renderer/components/+workloads-statefulsets/statefulset-details.tsx
```diff
@@ -58,6 +62,17 @@ export class StatefulSetDetails extends React.Component<Props> {
...
 
+  @disposeOnUnmount
+  revisionLoader = autorun(async () => {
+    const { object: statefulSet } = this.props;
+    const revisionHistory = await controllerRevisionApi.getRevisions({
+      name: statefulSet.getName(),
+      namespace: statefulSet.getNs()
+    }) as ControllerRevision[];
+
+    this.revisionHistory = observable.array(revisionHistory);
```
Nokel81, line 73: nit: could be
```suggestion
    this.revisionHistory.replace(await controllerRevisionApi.getRevisions({
      name: statefulSet.getName(),
      namespace: statefulSet.getNs()
    }));
```

src/renderer/components/+workloads/workload-rollback-dialog.tsx
```diff
@@ -0,0 +1,174 @@
...
+    this.revision = this.revisions[0];
+    this.isLoading = false;
+  };
+
+  rollback = async () => {
+    const shell = createTerminalTab({
+      title: `Rollback: ${this.workloadKubeObject.getName()} (namespace: ${this.workloadKubeObject.getNs()})`
+    });
+
+    TerminalStore.getInstance().sendCommand(`kubectl rollout undo ${this.workloadKubeObject.kind.toLocaleLowerCase()}` +
```
Nokel81, line 114: why `toLocaleLowerCase()`? even more `kubectl` accepts kind directly it seems.

src/common/k8s-api/endpoints/stateful-set.api.ts
vshakirova: I can change my approach, but what do you mean by saying 'free function'?

Nokel81:
Plan to try and finish a better and more complete review today, would be great if you could rebase.

Nokel81:
Marking as draft since it has been few months since this was last worked on.

Nokel81:

PR #3538, opened 2021-07-30
Make DetectoryRegistry a Singleton and initialize in a function

Nokel81:

part of splitting out ban-cir-deps

### DIFF
--- a/src/main/initializers/cluster-metadata-detectors.ts
+++ b/src/main/initializers/cluster-metadata-detectors.ts
@@ -27,7 +27,7 @@ import { NodesCountDetector } from "../cluster-detectors/nodes-count-detector";
 import { VersionDetector } from "../cluster-detectors/version-detector";
 
 export function initClusterMetadataDetectors() {
-  DetectorRegistry.getInstance()
+  DetectorRegistry.createInstance()
     .add(ClusterIdDetector)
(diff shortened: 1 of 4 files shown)


src/main/initializers/cluster-metadata-detectors.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import { ClusterIdDetector } from "../cluster-detectors/cluster-id-detector";
+import { DetectorRegistry } from "../cluster-detectors/detector-registry";
+import { DistributionDetector } from "../cluster-detectors/distribution-detector";
+import { LastSeenDetector } from "../cluster-detectors/last-seen-detector";
+import { NodesCountDetector } from "../cluster-detectors/nodes-count-detector";
+import { VersionDetector } from "../cluster-detectors/version-detector";
+
+export function initClusterMetadataDetectors() {
+  DetectorRegistry.getInstance()
```
jim-docker: You could create the instance here too? It could be considered part of the initialization and eliminates the chance that this gets called before the singleton is created (someone could move code around in the future)

Nokel81: Fair, enough will do.

jim-docker (approved):

PR #3540, opened 2021-07-30
Move lookupApi into apiManager

Nokel81:
- Don't export kube-api-parse from kube-api (cir-deps)

part of splitting apart ban-cir-deps

### DIFF
--- a/src/renderer/components/dock/create-resource.tsx
+++ b/src/renderer/components/dock/create-resource.tsx
@@ -105,7 +105,7 @@ export class CreateResource extends React.Component<Props> {
     await Promise.all(
       resources.map(data => {
         return resourceApplierApi.update(data)
-          .then(item => createdResources.push(item.getName()))
+          .then(item => createdResources.push(item.metadata.name))
           .catch((err: JsonApiErrorParsed) => errors.push(err.toString()));
       })
     );
--- a/src/renderer/kube-object.store.ts
+++ b/src/renderer/kube-object.store.ts
@@ -27,7 +27,8 @@ import { KubeObject, KubeStatus } from "./api/kube-object";
 import type { IKubeWatchEvent } from "./api/kube-watch-api";
 import { ItemStore } from "./item.store";
 import { apiManager } from "./api/api-manager";
-import { ensureObjectSelfLink, IKubeApiQueryParams, KubeApi, parseKubeApi } from "./api/kube-api";
+import { ensureObjectSelfLink, IKubeApiQueryParams, KubeApi } from "./api/kube-api";
+import { parseKubeApi } from "./api/kube-api-parse";
 import type { KubeJsonApiData } from "./api/kube-json-api";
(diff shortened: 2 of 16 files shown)


src/renderer/components/dock/create-resource.tsx
```diff
@@ -105,7 +105,7 @@ export class CreateResource extends React.Component<Props> {
     await Promise.all(
       resources.map(data => {
         return resourceApplierApi.update(data)
-          .then(item => createdResources.push(item.getName()))
+          .then(item => createdResources.push(item.metadata.name))
```
jim-docker, line 108: why?

src/renderer/kube-object.store.ts
```diff
@@ -279,7 +280,8 @@ export abstract class KubeObjectStore<T extends KubeObject> extends ItemStore<T>
   }
 
   async update(item: T, data: Partial<T>): Promise<T> {
-    const newItem = await item.update(data);
+    const rawItem = await item.update(data);
+    const newItem = new this.api.objectConstructor(rawItem);
```
jim-docker, line 284: this behaviour seems a little different than it was in resource-applier.api, is this ok?

src/renderer/components/dock/create-resource.tsx
Nokel81: Because `item` is of type `KubeJsonApiData` and doesn't have any methods

src/renderer/kube-object.store.ts
Nokel81: So I changed resource applier because it was doing needless extra work (and wasn't type correct).

In this case, we actually end up with the same result. One of the reasons that I wanted to get away from having multiple APIs for a single store.

In the other case that you commented on, we don't actually care about the `KubeObject` instance, so I don't bother creating one (which in the previous version of the code was the `new KubeObject` branch.

jim-docker:
Not sure if this is related to something in this chain of PRs:

https://user-images.githubusercontent.com/40840436/128207314-3987e00f-aa33-43c7-958e-1c280bc35d13.mov

Clicking on a definition sometimes brings up the details page, when it should bring up the list of CRD instances?

Nokel81:
@jim-docker Do you mean that it is strange that the drawer opens when you click the row near the name of the CRD but opens the list of resources if you click the name?

I think that has always been the situation.

jim-docker:
> @jim-docker Do you mean that it is strange that the drawer opens when you click the row near the name of the CRD but opens the list of resources if you click the name?
> 
> I think that has always been the situation.

Ah, I see, I was clicking just below the word "Prometheus", such that it wasn't highlighted (wasn't white instead of blue), hence the details came up

Nokel81:
Yes I agree, that it is a bit of a bad design and should probably open the list of resources when clicking that entire cell (and not just the text), but that is for a different PR.

jim-docker (approved):

PR #3542, opened 2021-07-30
Fix custom resources not having a fallback details component

### LINKED ISSUE #3541: Generic CRD details are empty
Regression from #2981

Nokel81:

fixes #3541

### DIFF
--- a/src/renderer/api/kube-object.ts
+++ b/src/renderer/api/kube-object.ts
@@ -87,7 +87,7 @@ export class KubeStatus {
 }
 
 export interface KubeObjectStatus {
-  conditions: {
+  conditions?: {
     lastTransitionTime: string;
     message: string;
     reason: string;
--- a/src/renderer/api/kube-object.ts
+++ b/src/renderer/api/kube-object.ts
@@ -86,6 +86,16 @@ export class KubeStatus {
   }
 }
 
+export interface KubeObjectStatus {
+  conditions: {
+    lastTransitionTime: string;
+    message: string;
+    reason: string;
+    status: string;
+    type?: string;
+  }[];
+}
+
 export type KubeMetaField = keyof KubeObjectMetadata;
 
 export class KubeObject<Metadata extends KubeObjectMetadata = KubeObjectMetadata, Status = any, Spec = any> implements ItemObject {
--- a/src/renderer/components/+custom-resources/crd-resource-details.tsx
+++ b/src/renderer/components/+custom-resources/crd-resource-details.tsx
@@ -68,14 +68,14 @@ export class CrdResourceDetails extends React.Component<Props> {
     ));
   }
 
-  renderStatus(crd: KubeObject<KubeObjectMetadata, KubeObjectStatus, any>, columns: AdditionalPrinterColumnsV1[]) {
-    const showStatus = !columns.find(column => column.name == "Status") && crd.status?.conditions;
+  renderStatus(customResource: KubeObject<KubeObjectMetadata, KubeObjectStatus, any>, columns: AdditionalPrinterColumnsV1[]) {
+    const showStatus = !columns.find(column => column.name == "Status") && Array.isArray(customResource.status?.conditions);
 
     if (!showStatus) {
       return null;
     }
 
-    const conditions = crd.status.conditions
+    const conditions = customResource.status.conditions
       .filter(({ type, reason }) => type || reason)
       .map(({ type, reason, message, status }) => ({ kind: type || reason, message, status }))
       .map(({ kind, message, status }, index) => (
--- a/src/renderer/components/+custom-resources/crd-resource-details.tsx
+++ b/src/renderer/components/+custom-resources/crd-resource-details.tsx
@@ -24,18 +24,18 @@ import "./crd-resource-details.scss";
 import React from "react";
 import jsonPath from "jsonpath";
 import { observer } from "mobx-react";
-import { computed, makeObservable } from "mobx";
 import { cssNames } from "../../utils";
 import { Badge } from "../badge";
 import { DrawerItem } from "../drawer";
 import type { KubeObjectDetailsProps } from "../kube-object";
-import { crdStore } from "./crd.store";
 import { KubeObjectMeta } from "../kube-object/kube-object-meta";
 import { Input } from "../input";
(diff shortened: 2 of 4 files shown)


src/renderer/api/kube-object.ts
```diff
@@ -86,6 +86,16 @@ export class KubeStatus {
...
 
+export interface KubeObjectStatus {
+  conditions: {
+    lastTransitionTime: string;
+    message: string;
+    reason: string;
+    status: string;
+    type?: string;
+  }[];
+}
```
Nokel81: This is the starting of a similar type to `KubeObjectMetadata`. In the future we should be able to use it everywhere because kube uses this type for object (even custom objects) it seems.

src/renderer/components/+custom-resources/crd-resource-details.tsx
```diff
@@ -24,18 +24,18 @@ import "./crd-resource-details.scss";
...
-import { crdStore } from "./crd.store";
 import { KubeObjectMeta } from "../kube-object/kube-object-meta";
 import { Input } from "../input";
 import type { AdditionalPrinterColumnsV1, CustomResourceDefinition } from "../../api/endpoints/crd.api";
 import { parseJsonPath } from "../../utils/jsonPath";
+import type { KubeObject, KubeObjectMetadata, KubeObjectStatus } from "../../api/kube-object";
 
-interface Props extends KubeObjectDetailsProps<CustomResourceDefinition> {
+interface Props extends KubeObjectDetailsProps<KubeObject> {
+  crd: CustomResourceDefinition;
```
Nokel81, line 38: I changed to passing in the crd (which should be the parent crd of `object`) because the single use of this component already needs to have it.

src/renderer/components/+custom-resources/crd-resource-details.tsx
```diff
@@ -60,24 +60,15 @@ function convertSpecValue(value: any): any {
...
-    super(props);
-    makeObservable(this);
-  }
-
-  @computed get crd() {
-    return crdStore.getByObject(this.props.object);
-  }
-
-  renderAdditionalColumns(crd: CustomResourceDefinition, columns: AdditionalPrinterColumnsV1[]) {
+  renderAdditionalColumns(resource: KubeObject, columns: AdditionalPrinterColumnsV1[]) {
```
Nokel81, line 63: This component was incorrect because it said it was dealing with `CustomResourceDefenition`'s as its main type. It is not, it is dealing with custom resources (which are from Lens' perspective, just `KubeObject`'s).

src/renderer/components/+custom-resources/crd-resource-details.tsx
```diff
@@ -60,24 +60,15 @@ function convertSpecValue(value: any): any {
...
     return columns.map(({ name, jsonPath: jp }) => (
       <DrawerItem key={name} name={name} renderBoolean>
-        {convertSpecValue(jsonPath.value(crd, parseJsonPath(jp.slice(1))))}
+        {convertSpecValue(jsonPath.value(resource, parseJsonPath(jp.slice(1))))}
       </DrawerItem>
     ));
   }
 
-  renderStatus(crd: CustomResourceDefinition, columns: AdditionalPrinterColumnsV1[]) {
+  renderStatus(crd: KubeObject<KubeObjectMetadata, KubeObjectStatus, any>, columns: AdditionalPrinterColumnsV1[]) {
```
Nokel81: Hopefully in the future, we won't need to specify `KubeObjectStatus` and we can just specify `KubeObject` but we are not there yet.

src/renderer/components/kube-object/kube-object-details.tsx
```diff
@@ -169,6 +166,18 @@ export class KubeObjectDetails extends React.Component {
...
+    if (details.length === 0) {
+      const crd = crdStore.getByObject(object);
+
+      /**
+       * This is a fallback so that if a custom resource object doesn't have
+       * any defined details we should try and display at least some details
+       */
+      if (crd) {
+        details.push(<CrdResourceDetails key={object.getId()} object={object} crd={crd} />);
+      }
```
Nokel81, line 179: By fixing the typing on `CrdResourceDetails` (to not saying we want a `CustomResourceDefenition` as `object`, because we don't) it is no longer necessary to use a type cast.

src/renderer/components/kube-object/kube-object-details.tsx
```diff
@@ -169,6 +166,18 @@ export class KubeObjectDetails extends React.Component {
...
       ));
 
+    if (details.length === 0) {
+      const crd = crdStore.getByObject(object);
+
+      /**
+       * This is a fallback so that if a custom resource object doesn't have
+       * any defined details we should try and display at least some details
+       */
+      if (crd) {
```
jakolehm, line 177: Why do we need to get & check `crd`? `object` already is the crd, right?

Nokel81: No object is a resource that is described by a CRD (I would call it a CR).

I guess we should fall back to just `KubeObjectDetails` if `crd` is false.

src/renderer/api/kube-object.ts
jakolehm: Are we sure all kubernetes objects have these?

jakolehm: ~~Also I think this interface does not match actual implementation, conditions seems to be an array of objects.~~

src/renderer/api/kube-object.ts
```diff
@@ -86,6 +86,16 @@ export class KubeStatus {
   }
 }
 
+export interface KubeObjectStatus {
+  conditions: {
```
jakolehm:
```suggestion
  conditions?: {
```

src/renderer/api/kube-object.ts
Nokel81: So I am not sure if all kube controllers do use this (or something like this...)

After reading:
 - https://github.com/kubernetes/kubernetes/issues/50798
 - https://github.com/kubernetes/kubernetes/issues/7856
 - https://github.com/kubernetes/community/pull/4521

and https://maelvls.dev/kubernetes-conditions/

It seems clear that not everything does use this type (or at least MUST use something like this). But that it is still the recommended way of displaying the history of data.

jakolehm (approved):
LGTM

PR #3545, opened 2021-07-30
Remove circular dependency with isAllowedResource function

Nokel81:

### DIFF
--- a/src/common/utils/allowed-resource.ts
+++ b/src/common/utils/allowed-resource.ts
@@ -19,27 +19,23 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import React from "react";
-import { isAllowedResource } from "../../common/utils/allowed-resource";
-import { WorkloadsOverviewDetailRegistry } from "../../extensions/registries";
-import { Events } from "../components/+events";
-import { OverviewStatuses } from "../components/+workloads-overview/overview-statuses";
+import { ClusterStore } from "../cluster-store";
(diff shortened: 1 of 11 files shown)


src/common/utils/allowed-resource.ts
```diff
@@ -0,0 +1,41 @@
...
+  const resources = [resource].flat();
+  const cluster = ClusterStore.getInstance().getById(getHostedClusterId());
+
+  if (!cluster?.allowedResources) {
+    return false;
+  }
+
+  if (resources.length === 0) {
+    return true;
+  }
```
jim-docker, line 36: guess you could do this before getting the cluster

src/common/utils/allowed-resource.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  const allowedResources = new Set(cluster.allowedResources);
+
+  return resources.every(resource => allowedResources.has(resource));
```
jim-docker, line 40: Why is this better than:
```suggestion
  return resources.every(resource => cluster.allowedResources.includes(resource));
```

jim-docker (approved):

src/common/utils/allowed-resource.ts
Nokel81: Fair enough

src/common/utils/allowed-resource.ts
Nokel81: Generally because I would figure it is less work (though I don't have the data to back it up).

Nokel81:
Will make a follow up PR once everything is merged.

PR #3546, opened 2021-07-30
Inject handlers into LensProxy to remove circular-deps

Nokel81:

### DIFF
--- a/src/main/proxy-functions/kube-api-request.ts
+++ b/src/main/proxy-functions/kube-api-request.ts
@@ -1,0 +1,84 @@
+/**
+ * Copyright (c) 2021 OpenLens Authors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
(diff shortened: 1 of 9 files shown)


src/main/proxy-functions/kube-api-request.ts
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+import { chunk } from "lodash";
+import net from "net";
+import url from "url";
+import { apiKubePrefix } from "../../common/vars";
+import { ClusterManager } from "../cluster-manager";
+import type { ProxyApiRequestArgs } from "./types";
+
+const skipRawHeaders = new Set(["Host", "Authorization"]);
```
jim-docker, line 29: could just be an array and use `Array.includes()` below instead of `Set.has()`? Maybe there's an advantage to using `Set` here?

jim-docker (approved):

src/main/proxy-functions/kube-api-request.ts
Nokel81: I generally figure that a hashset has complexity of a hash and then a comparison.

Whereas the array would have a complexity of the length of the string being compared by the length of the array.

PR #3547, opened 2021-07-30
Ban circular dependencies

Nokel81:

This is the last in a long chain of PRs which all need to be merged before this one will pass. This was done to make reviewing each part easier.

The PRs are as follows and must be merged in order

1. [x] https://github.com/lensapp/lens/pull/3530
2. [x] https://github.com/lensapp/lens/pull/3532
3. [x] https://github.com/lensapp/lens/pull/3533
4. [x] https://github.com/lensapp/lens/pull/3538
5. [x] https://github.com/lensapp/lens/pull/3540
6. [x] https://github.com/lensapp/lens/pull/3544
7. [x] https://github.com/lensapp/lens/pull/3545
8. [x] https://github.com/lensapp/lens/pull/3546

The following PR: must also be merged before this one, but can be done out of order of the above:
1. [x] https://github.com/lensapp/lens/pull/3531

blocked on above


src/common/user-store/preferences-helpers.ts
```diff
@@ -22,9 +22,11 @@
 import moment from "moment-timezone";
 import path from "path";
 import os from "os";
-import { ThemeStore } from "../../renderer/theme.store";
+import type { ThemeId } from "../../renderer/theme.store";
 import { ObservableToggleSet } from "../utils";
 
+export const defaultThemeName: ThemeId = "lens-dark";
```
jim-docker: is this to avoid a circular dep? Seems unfortunate not be able to use the already defined default value

src/renderer/components/dock/dock.store.ts
```diff
@@ -282,6 +282,7 @@ export class DockStore implements DockStorageState {
 
         if (newTab?.kind === TabKind.TERMINAL) {
           // close the dock when selected sibling inactive terminal tab
+          // eslint-disable-next-line import/no-cycle
```
jim-docker: ok because it's dynamically imported? No other workaround?

src/renderer/utils/createStorage.ts
```diff
@@ -73,11 +72,6 @@ export function createAppStorage<T>(key: string, defaultValue: T, clusterId?: st
       equals: comparer.structural, // save only when something really changed
     });
 
-    // remove json-file when cluster deleted
```
jim-docker: Does it get removed some other way now?

src/renderer/components/dock/dock.store.ts
Nokel81: No, it is fine because it is a dynamic import. 

In a previous iteration I had tried to refactor the docks, dock tabs, and terminals to not need this. But it was way too large of a change. I think I will come back at a later point to remove this.

src/renderer/utils/createStorage.ts
Nokel81: So I removed this for two reasons, first of all because the current logic is broken. No cluster will ever be part of `removedClusters` because of how sync works.

Second because it was a circular import. I think a better form would be a "force delete" option when removing a cluster.

src/common/user-store/preferences-helpers.ts
```diff
@@ -58,10 +56,10 @@ const shell: PreferenceDescription<string | undefined> = {
 
 const colorTheme: PreferenceDescription<string> = {
   fromStore(val) {
-    return val || defaultThemeName;
+    return val || ThemeStore.defaultTheme;
```
jim-docker: Can the `defaultTheme` be moved to vars (or somewhere appropriate) and imported from there?

src/common/cluster-store.ts
```diff
@@ -142,35 +141,19 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
...
-    // update removed clusters
-    currentClusters.forEach(cluster => {
-      if (!newClusters.has(cluster.id)) {
-        removedClusters.set(cluster.id, cluster);
-      }
-    });
-
-    this.clusters.replace(newClusters);
-    this.removedClusters.replace(removedClusters);
   }
```
jim-docker: Are you sure you want to do this in this PR? (I mean all the changes in this file)

Nokel81: Fair enough, probably not

src/common/user-store/preferences-helpers.ts
Nokel81: Yup

.eslintrc.js
```diff
@@ -21,6 +21,17 @@
...
 const isInVscode = Boolean(process.env.VSCODE_PID);
+const isInGitHubActions = Boolean(process.env.GITHUB_ACTIONS);
+
+const noCycleMaxDepth = (
+  isInVscode
+    ? 10 // very fast for editors
+    : (
+      isInGitHubActions
+        ? undefined // on CI check everything
+        : 30 // somewhat fast for local lint running
```
jim-docker: nice, I was going to ask if this could be done

src/renderer/theme.store.ts
```diff
@@ -46,7 +46,6 @@ export interface ThemeItems extends Theme {
 }
 
 export class ThemeStore extends Singleton {
-  static readonly defaultTheme = "lens-dark";
```
jim-docker, line 43: Is this a breaking change? It could be kept, assigned to `defaultTheme` from vars?

.eslintrc.js
Nokel81: I think I should reduce it a bit further

src/renderer/theme.store.ts
Nokel81: It is only a breaking change if we export `ThemeStore`, which do don't we only export `src/extensions/renderer-api/theming.ts`

Nokel81:
While the `import/no-cycle` rule produced nice errors within vscode. It takes significantly longer than the webpack plugin to run (~10 minutes instead of ~20s) so I think we shouldn't use it until it gets faster.

src/common/ipc/ipc.ts
```diff
@@ -30,7 +30,15 @@ import { ClusterFrameInfo, clusterFrameMap } from "../cluster-frames";
 import type { Disposer } from "../utils";
 import type remote from "@electron/remote";
 
-const electronRemote = ipcMain ? null : require("@electron/remote");
+const electronRemote = (() => {
+  if (ipcRenderer) {
```
Iku-turso, line 34: This is non-env-agnostic code in common, which is bad. However, the tooling to solve something like this is not here yet -> ok for now.

src/common/ipc/ipc.ts
```diff
@@ -30,7 +30,15 @@ import { ClusterFrameInfo, clusterFrameMap } from "../cluster-frames";
...
+    try {
```
Iku-turso, line 35: Hmm. Add comment to explain the underlying very good reason for this wild-card try/catch.

src/common/k8s-api/kube-api.ts
```diff
@@ -34,6 +34,9 @@ import type { IKubeWatchEvent } from "./kube-watch-api";
 import { KubeJsonApi, KubeJsonApiData } from "./kube-json-api";
 import { noop } from "../utils";
 import type { RequestInit } from "node-fetch";
+
+// BUG: https://github.com/mysticatea/abort-controller/pull/22
```
Iku-turso, line 38: Could explain more about the relevancy of this bug?

src/common/k8s-api/kube-object.store.ts
```diff
@@ -30,6 +30,9 @@ import { ensureObjectSelfLink, IKubeApiQueryParams, KubeApi } from "./kube-api";
 import { parseKubeApi } from "./kube-api-parse";
 import type { KubeJsonApiData } from "./kube-json-api";
 import type { RequestInit } from "node-fetch";
+
+// BUG: https://github.com/mysticatea/abort-controller/pull/22
```
Iku-turso, line 34: Same as before.

src/common/ipc/ipc.ts
Nokel81: This code will be going away as soon as #4625 is merged so I don't think it is really necessary.

src/renderer/components/kube-object-details/kube-object-details.tsx
```diff
@@ -62,7 +61,7 @@ export class KubeObjectDetails extends React.Component {
         .getStore(this.path)
         ?.getByPath(this.path);
     } catch (error) {
-      logger.error(`[KUBE-OBJECT-DETAILS]: failed to get store or object: ${error}`, { path: this.path });
+      console.error(`[KUBE-OBJECT-DETAILS]: failed to get store or object: ${error}`, { path: this.path });
```
Iku-turso, line 64: Hmm. Should not use explicit console.error instead of abstracted logger.

src/common/k8s-api/kube-api.ts
Nokel81: So there is the rule `import/no-named-as-default` from `eslint-import` which errors if (according to the typings) an default import uses the same name as a named non-default import because it might mean that you meant to import that instead.

However, in this case there is a bug in the "abort-controller" package where (even though according to the typings there is an export called "AbortController") in fact it is only default exported in the JS.

src/renderer/components/kube-object-details/kube-object-details.tsx
Nokel81: Fair enough, I don't exactly remember why I did this since in #4317 I revert this back.... will revert.

Iku-turso (approved):
Added mostly superficial comments. Mergeable without action.

jim-docker (approved):
for me, `make lint` now takes 67s (was 17s) on mac,  54s (was 12s) on linux. Much better than 10m

PR #3553, opened 2021-08-02
Refactor k8s api to common

jakolehm:
Allows to use kubernetes api library from `main`. We probably should do the same for KubeObject stores but I'll leave that to follow-up PR.

Also improves winston log output because we cannot see api requests in devtools network tab (because everything is executed within nodejs runtime).

image
image


src/common/k8s-api/json-api.ts
```diff
@@ -194,11 +205,12 @@ export class JsonApi<D = JsonApiData, P extends JsonApiParams = JsonApiParams> {
   protected writeLog(log: JsonApiLog) {
     if (!this.config.debug) return;
     const { method, reqUrl, ...params } = log;
-    let textStyle = "font-weight: bold;";
+    // let textStyle = "font-weight: bold;";
```
nevalla: Do we want to leave these commented lines?

src/common/k8s-api/json-api.ts
```diff
@@ -194,11 +205,12 @@ export class JsonApi<D = JsonApiData, P extends JsonApiParams = JsonApiParams> {
...
 
-    if (params.data) textStyle += "background: green; color: white;";
-    if (params.error) textStyle += "background: red; color: white;";
-    console.log(`%c${method} ${reqUrl}`, textStyle, params);
+    // if (params.data) textStyle += "background: green; color: white;";
+    // if (params.error) textStyle += "background: red; color: white;";
+    //console.log(`%c${method} ${reqUrl}`, textStyle, params);
+    logger.info(`[JSON-API] request ${method} ${reqUrl}`, params);
```
jakolehm, line 197: Will cleanup this after we have decided right logging level. Would `debug` be better?

src/common/k8s-api/kube-api.ts
```diff
@@ -173,7 +173,6 @@ export class KubeApi<T extends KubeObject> {
     this.request = request;
     this.objectConstructor = objectConstructor;
 
-    this.checkPreferredVersion();
```
nevalla, line 171: we don't need this anymore?

src/common/k8s-api/kube-api.ts
```diff
@@ -479,7 +478,7 @@ export class KubeApi<T extends KubeObject> {
         });
       })
       .catch(error => {
-        if (error instanceof DOMException) return; // AbortController rejects, we can ignore it
+        if (typeof DOMException === "function" && error instanceof DOMException) return; // AbortController rejects, we can ignore it
```
jakolehm: @Nokel81 do you remember why we need to handle `DOMException`?

src/common/k8s-api/kube-object.store.ts
```diff
@@ -192,9 +191,9 @@ export abstract class KubeObjectStore<T extends KubeObject> extends ItemStore<T>
 
       return items;
     } catch (error) {
-      if (error.message) {
-        Notifications.error(error.message);
-      }
+      // if (error.message) {
```
jakolehm: @Nokel81 is it ok to remove this Notification? Mixing UI with lower level classes is not a good idea (and it does not work in `main`).

nevalla: some lines commented out here too

nevalla (dismissed):
added couple of comments about some lines commented out, otherwise LGTM and by quick testing everything seems to work on renderer side.

src/common/k8s-api/json-api.ts
jakolehm: No, I left these on purpose to get feedback... see https://github.com/lensapp/lens/pull/3553#discussion_r681474267

src/common/k8s-api/kube-api.ts
jakolehm: I asked the same question on Slack few days ago. Everything seems to work without -> calling this in a constructor has serious side-effects because it will trigger api requests when api is instantiated ... at that point it might not even have a context where it should make the api call. Feels like a wrong place for this call (maybe it's some legacy?).

src/common/k8s-api/kube-object.store.ts
jakolehm: See https://github.com/lensapp/lens/pull/3553#discussion_r681475098

src/common/k8s-api/kube-api.ts
nevalla: yes, agree

src/common/k8s-api/json-api.ts
nevalla: probably, log lines we are getting now are a bit too verbose for `info` level.

jakolehm:
There seems to be a problem with "watch" requests, will investigate.

src/common/k8s-api/kube-api.ts
Nokel81: DOMException is the type of exception that the `abort-controller` package throws when the `cancel` function is called.

src/common/k8s-api/kube-object.store.ts
Nokel81: I think it should be fine. My only concern is that because no downstream consumer of this function call will be able to observe this error so a consumer on the renderer wouldn't be able to show the notification anyway.

jakolehm: I think we could emit an error or something like that. It's hard to return an error with current interface.

src/common/k8s-api/kube-api.ts
jakolehm: Does it throw DOMException also with nodejs environment?

src/common/k8s-api/kube-object.store.ts
Nokel81: Fair enough.

src/common/k8s-api/json-api.ts
```diff
@@ -49,9 +51,9 @@ export interface JsonApiLog {
 
 export interface JsonApiConfig {
   apiBase: string;
+  serverAddress: string;
```
jakolehm, line 54: `node-fetch` requires full address for requests.

src/renderer/components/+workloads-replicasets/replicaset-scale-dialog.tsx
```diff
@@ -30,9 +30,10 @@ import { Icon } from "../icon";
 import { Slider } from "../slider";
 import { Notifications } from "../notifications";
 import { cssNames } from "../../utils";
-import { ReplicaSet, replicaSetApi } from "../../api/endpoints/replica-set.api";
+import { ReplicaSet, ReplicaSetApi, replicaSetApi } from "../../../common/k8s-api/endpoints/replica-set.api";
 
 interface Props extends Partial<DialogProps> {
+  replicaSetApi: ReplicaSetApi
```
jakolehm, line 36: Easier to test if we don't use global stores. And with this PR those global stores are not automatically initialized so had to make these changes.

msa0311 (dismissed):

nevalla (approved):
LGTM

PR #3572, opened 2021-08-05
Upgrade to electron 12 and Node 14

### LINKED ISSUE #3665: Tray icon disappears on window close
**Describe the bug**
Tray icon disappears when Lens main window is closed. 

**Expected behavior**
Tray icon is kept until Lens app is quit.

**Additional context**
Most likely a regression from https://github.com/lensapp/lens/pull/3546

Nokel81:

This PR upgrades our electron version to the current latest 12.X and the current latest node 14.X.

I tried upgrading to electron 13.X but ran into trouble with the `@electron/remote` library. I think we should probably try and remove that before upgrading as it is considered harmful to security.

Notes:

- I have removed `spectron` as our integration testing library as it doesn't work with subframes in electron@12.X. 
- I have rewrote our integration tests using `playwright`. This did require changing some of the tests because playwright doesn't have some of the indepth electron APIs. I believe that this leads to better tests because the interactions that the tests can use are closer to what the users can do.


package.json
```diff
@@ -365,7 +367,7 @@
     "react-window": "^1.8.5",
     "sass-loader": "^8.0.2",
     "sharp": "^0.26.1",
-    "spectron": "11.0.0",
+    "spectron": "14",
```
jakolehm: Seems that we don't need spectron anymore, probably should be removed from here?

extensions/pod-menu/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "lens-pod-menu",
-  "version": "0.0.1",
+  "version": "5.1.4-alpha.1628102440259.1628102553246",
```
jakolehm: These are probably added by accident.

Nokel81:
Blocked on #3602 as some of the conflicts are because of that regression.

jakolehm:
@Nokel81 tests are still failing.

aleksfront:
Seems some of them keeps executing forever (that's why here Tests are marked as cancelled)
integration tests

src/renderer/components/+apps-helm-charts/helm-charts.tsx
```diff
@@ -86,17 +86,19 @@ export class HelmCharts extends Component<Props> {
...
             chart => chart.getKeywords(),
           ]}
-          customizeHeader={({ searchProps }) => ({
+          customizeHeader={({ searchProps, ...headerPlaceholders }) => ({
             searchProps: {
               ...searchProps,
               placeholder: "Search Helm Charts...",
             },
+            ...headerPlaceholders
           })}
```
aleksfront: Helm Charts page was designed without header title and expanded search box by purpose. This changes breaks initial layout.

Before the changes (looks like expected):
helm charts before

Current view (broken):
helm charts after

We can just discard changes in this file.

Nokel81: Okay, I will have to revisit the tests then because I needed that `Helm Charts` title.

Nokel81:
@aleksfront @jakolehm I have fixed the tests, PTAL!

integration/helpers/utils.ts
```diff
@@ -69,71 +34,45 @@ export function describeIf(condition: boolean) {
...
+  await window.waitForSelector("div.TableCell");
+  await window.click("div.TableCell >> text='minikube'");
+  await window.waitForSelector("div.drawer-title-text >> text='KubernetesCluster: minikube'");
+  await window.click("div.EntityIcon div.HotbarIcon div div.MuiAvatar-root");
 
-type HelmRepository = {
-  name: string;
-  url: string;
-};
+  const minikubeFrame = await window.waitForSelector("#cluster-frame-484e864bad9b84ce5d6b4fff704cc0e4");
```
nevalla: where does this id come from?

src/renderer/components/+apps-releases/releases.tsx
```diff
@@ -138,7 +138,7 @@ export class HelmReleases extends Component<Props> {
             },
             ...headerPlaceholders,
           })}
-          renderHeaderTitle="Releases"
+          renderHeaderTitle="Helm Releases"
```
nevalla: This change is not related to Electron version bump(?)

src/renderer/utils/createStorage.ts
```diff
@@ -47,15 +46,11 @@ export function createStorage<T>(key: string, defaultValue: T) {
...
+  const folder = path.resolve(getPath("userData"), "lens-local-storage");
   const fileName = `${clusterId ?? "app"}.json`;
   const filePath = path.resolve(folder, fileName);
 
-  if (!storage.initialized) {
-    init(); // called once per cluster-view
-  }
-
-  function init() {
+  if (!storage.initialized && !process.argv.includes("--integration-testing")) {
```
nevalla: ideally this `!process.argv.includes("--integration-testing")` should be in helper method that is called here

src/renderer/ipc/index.tsx
```diff
@@ -117,7 +117,7 @@ function HotbarTooManyItemsHandler(): void {
   Notifications.error(`Cannot have more than ${defaultHotbarCells} items pinned to a hotbar`);
 }
 
-export function registerIpcHandlers() {
+export function registerIpcListeners() {
```
nevalla, line 120: Not sure if this change is needed either, but probably makes sense better now.

nevalla (dismissed):
Generally looking good and could not see any errors while testing this. For some reason integration tests are failing. I triggered re-run, but if those will fail again, there must be some real issue. 

We should probably have helper method to resolve is app running on integration tests to reduce code repetition.

src/renderer/utils/createStorage.ts
Nokel81: Sounds good

src/renderer/components/+apps-releases/releases.tsx
Nokel81: No and I will revert

integration/helpers/utils.ts
Nokel81: From line 205 in `src/main/catalog-sources/kubeconfig-sync.ts`:

```ts
const clusterId = createHash("md5").update(`${filePath}:${contextName}`).digest("hex");
```

I guess I should probably make it so that the the ID is computed on demand for each OS

src/main/window-manager.ts
```diff
@@ -75,6 +75,12 @@ export class WindowManager extends Singleton {
       app.dock?.show();
 
       const { width, height, x, y } = this.windowState;
+      const additionalArguments = [];
+
+      if (isIntegrationTesting) {
```
jakolehm: Why do we need this?

src/renderer/components/+apps-helm-charts/helm-charts.tsx
```diff
@@ -86,6 +86,7 @@ export class HelmCharts extends Component<Props> {
             [columnId.name]: chart => chart.getName(),
             [columnId.repo]: chart => chart.getRepository(),
           }}
+          renderHeaderTitle="Helm Charts"
```
jakolehm: Unrelated change?

src/renderer/utils/createStorage.ts
```diff
@@ -22,16 +22,16 @@
...
 import { StorageHelper } from "./storageHelper";
 import { ClusterStore } from "../../common/cluster-store";
 import logger from "../../main/logger";
-import { getHostedClusterId } from "../../common/utils";
+import { getHostedClusterId, getPath, noop } from "../../common/utils";
+import { isIntegrationTesting } from "../../common/vars";
 
 const storage = observable({
-  initialized: false,
+  initialized: isIntegrationTesting,
```
jakolehm: This feels odd, we should not change logic while running integration tests.

integration/helpers/utils.ts
```diff
@@ -69,71 +37,69 @@ export function describeIf(condition: boolean) {
...
-  const app =  new Application({
-    path: AppPaths[process.platform], // path to electron app
-    args: [],
-    startTimeout: 60000,
-    waitTimeout: 10000,
-    env: {
-      CICD: "true"
+  for (let i = 0; i < 10; i += 1) {
+    try {
+      return await electron.launch({
```
jakolehm: This is missing temporary path for user data. Now it's pretty dangerous (and probably flaky) to run integration tests because it will read/modify users real environment. Most likely `--integration-testing` flag can be removed after we fix this.

jakolehm (dismissed):
We need to use temporary path for user data during tests.

integration/helpers/utils.ts
Nokel81: Not quite, but we don't need to pass it to the renderer if we change the user data path. I initially introduced this flag so as to not block the app from quitting (which we do under normal circumstances).

I guess a proper solution would be to send a SIGTERM to the process, I might look into that.

src/renderer/utils/createStorage.ts
Nokel81: Fair enough, the changing of userData path is probably better.

build/download_kubectl.ts
```diff
@@ -117,7 +117,7 @@ class KubectlDownloader {
 }
 
 const downloadVersion = packageInfo.config.bundledKubectlVersion;
-const baseDir = path.join(process.env.INIT_CWD, "binaries", "client");
+const baseDir = path.join(__dirname, "..", "binaries", "client");
```
jakolehm, line 120: What's the reason behind this change?

integration/__tests__/app.tests.ts
```diff
@@ -25,63 +25,55 @@
...
 
-jest.setTimeout(2 * 60 * 1000); // 2 minutes so that we can get better errors from spectron
+describe("preferences page tests", () => {
+  it('shows "preferences" and can navigate through the tabs', async () => {
+    const { window, cleanup } = await utils.start();
 
-// FIXME (!): improve / simplify all css-selectors + use [data-test-id="some-id"] (already used in some tests below)
-describe("Lens integration tests", () => {
-  let app: Application;
+    try {
```
jakolehm: Try/finally should be done using before/afterEach to avoid repetition in tests.

src/common/ipc/ipc.ts
```diff
@@ -46,6 +48,7 @@ function getSubFrames(): ClusterFrameInfo[] {
 }
 
 export function broadcastMessage(channel: string, ...args: any[]) {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
```
jakolehm: Why this is needed? 🤔

src/common/utils/getPath.ts
```diff
@@ -0,0 +1,38 @@
...
+ * calls getPath either on app or on the remote's app
+ *
+ * @deprecated Use a different method for accessing the getPath function
+ */
+export function getPath(name: Parameters<typeof app["getPath"]>[0]): string {
+  if (app) {
+    return app.getPath(name);
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
```
jakolehm: Eslint-disable seems odd also here, is it required?

src/renderer/utils/createStorage.ts
```diff
@@ -68,32 +63,35 @@ export function createAppStorage<T>(key: string, defaultValue: T, clusterId?: st
       });
 
     // bind auto-saving data changes to %storage-file.json
-    reaction(() => toJS(storage.data), saveFile, {
-      delay: 250, // lazy, avoid excessive writes to fs
-      equals: comparer.structural, // save only when something really changed
-    });
+    reaction(
+      () => toJS(storage.data),
+      async (state: Record<string, any> = {}) => {
```
jakolehm: Was this change necessary? Usually code blocks are refactored to smaller logical chunks, this goes opposite way.

src/common/ipc/ipc.ts
Nokel81: I don't know, I think a relic of a revert from master.

build/download_kubectl.ts
Nokel81: Because on windows `INIT_CWD` is not defined.

src/main/menu.ts
```diff
@@ -80,10 +81,11 @@ export function buildMenu(windowManager: WindowManager) {
       { type: "separator" },
       {
         label: "Preferences",
+        id: "Preferences",
```
jim-docker: Why do some get an `id` and not others (like Extensions)? For testing purposes?

extensions/kube-object-event-status/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kube-object-event-status",
-  "version": "0.0.1",
+  "version": "5.2.0-beta.2.1629813675276",
```
jim-docker: still need to revert this?

integration/__tests__/app-preferences.tests.ts
```diff
@@ -0,0 +1,74 @@
...
+import * as utils from "../helpers/utils";
+
+describe("preferences page tests", () => {
+  let window: Page, cleanup: () => Promise<void>;
+
+  beforeEach(async () => {
+    ({ window, cleanup } = await utils.start());
+    await utils.clickWelcomeButton(window);
+    await window.keyboard.press("Meta+,");
+  }, 10*60*1000);
```
jim-docker, line 38: 10 minutes seems rather long?

src/common/utils/getPath.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+import { app, ipcMain } from "electron";
+
+const remote = ipcMain ? null : require("@electron/remote");
+
+/**
+ * calls getPath either on app or on the remote's app
+ *
+ * @deprecated Use a different method for accessing the getPath function
```
jim-docker, line 29: Adding a new function that is immediately deprecated? What different method should be used?

integration/__tests__/app-preferences.tests.ts
Nokel81: Fair however each statement has a 30s timeout by default. And I would much prefer it if `playwright` times out then jest because the error messages are much better.

src/main/menu.ts
Nokel81: I will be removing these once I get CI to pass as I no longer need them. Hopefully in the future we will be able to interact with the menu but it doesn't seem to work yet.

extensions/kube-object-event-status/package.json
Nokel81: Yup

src/common/utils/getPath.ts
Nokel81: Given that these shouldn't be changed once they are set. The better solution would be to initialize them for each frame once. But I wanted to leave that sort of thing to follow up PRs

integration/helpers/utils.ts
```diff
@@ -69,71 +40,77 @@ export function describeIf(condition: boolean) {
...
+  // Make sure that the directory is clear
+  await remove(CICD).catch(noop);
+  await mkdirp(CICD);
 
-  await app.stop();
+  const app = await electron.launch({
+    args: ["--integration-testing"], // this argument turns off the blocking of quit
+    executablePath: AppPaths[process.platform],
+    bypassCSP: true,
+    env: { CICD },
```
jim-docker: what does this do, just sets one env var, an empty path?

Nokel81: It shouldn't be empty. It should be a part to a new temporary directory. This allows each integration test to start with a clean slate in terms of local storage (so dock tabs and the sidebar open status for instance).

build/download_kubectl.ts
jakolehm: For future readers: `INIT_CWD` is not defined anymore on node14.

package.json
```diff
@@ -28,7 +28,7 @@
     "build:linux": "yarn run compile && electron-builder --linux --dir",
     "build:mac": "yarn run compile && electron-builder --mac --dir",
     "build:win": "yarn run compile && electron-builder --win --dir",
-    "integration": "jest --runInBand integration",
+    "integration": "jest --runInBand --detectOpenHandles --forceExit integration",
```
jakolehm, line 31: I hope we can remove `--forceExit` in follow-up PRs.

src/common/utils/getPath.ts
jakolehm: Deprecated new function sounds strange but let's fix this in follow-up PR.

src/main/index.ts
```diff
@@ -209,7 +211,7 @@ app.on("ready", async () => {
   logger.info("🖥️  Starting WindowManager");
   const windowManager = WindowManager.createInstance();
 
-  cleanup.push(
+  onQuitCleanup.push(
```
jakolehm, line 214: Note: these fixes are also in a separate PR.

src/renderer/utils/createStorage.ts
```diff
@@ -68,32 +63,35 @@ export function createAppStorage<T>(key: string, defaultValue: T, clusterId?: st
       });
 
     // bind auto-saving data changes to %storage-file.json
-    reaction(() => toJS(storage.data), saveFile, {
-      delay: 250, // lazy, avoid excessive writes to fs
-      equals: comparer.structural, // save only when something really changed
-    });
+    reaction(
+      () => toJS(storage.data),
+      async state => {
```
jakolehm: My previous comment has somehow got lost. This change does not really make sense to me. It does not alter the behaviour but just does the opposite of usual refactoring.. it bundles more logic into one (usually we refactor code blocks to smaller chunks).

package.json
Nokel81: Yes I completely agree, I am going to try and upgrade to the newer version of Jest in a follow-up PR.

src/main/index.ts
Nokel81: Shall we merge that other one first and then rebase this one?

src/renderer/utils/createStorage.ts
Nokel81: I will revert this and put it into a separate PR. I disagree that it is bundling more logic because I am removing function declarations which are only ever used once.

jakolehm: Just noticed that those functions were inline functions. Now the refactoring makes more sense.. anyway better to do outside of this PR because it is an unrelated change.

jakolehm (approved):

PR #3573, opened 2021-08-05
Sync kubeconfigs from catalog

### LINKED ISSUE #3232: Category onAdd for kubernetes sync

aleksfront:
1. Adding `Sync kubeconfig(s)` item in catalog blue add button using the code from https://github.com/lensapp/lens/pull/3234
2. Added success notification with a link to `Kubernetes` section inside `Preferences`.
3. Refactored Preferences page and child components to use `<Route/>` for navigation. This is needed for option 2 to work.

https://user-images.githubusercontent.com/9607060/128371453-c613699c-d18c-4838-86cf-357f6d0d9e30.mov

Should be merged replacing #3234 

Fixes #3232

### DIFF
--- a/src/renderer/components/+preferences/preferences.tsx
+++ b/src/renderer/components/+preferences/preferences.tsx
@@ -47,6 +47,8 @@ import { Application } from "./application";
 import { Kubernetes } from "./kubernetes";
 import { LensProxy } from "./proxy";
 import { Telemetry } from "./telemetry";
+import { Extensions } from "./extensions";
+import { sentryDsn } from "../../../common/vars";
 
 @observer
 export class Preferences extends React.Component {
@@ -58,7 +60,8 @@ export class Preferences extends React.Component {
   }
 
   renderNavigation() {
-    const extensions = AppPreferenceRegistry.getInstance().getItems().filter(e => !e.showInPreferencesTab);
+    const extensions = AppPreferenceRegistry.getInstance().getItems();
+    const telemetryExtensions = extensions.filter(e => e.showInPreferencesTab == "telemetry");
     const currentLocation = navigation.location.pathname;
     const isActive = (route: RouteProps) => !!matchPath(currentLocation, { path: route.path, exact: route.exact });
 
@@ -68,8 +71,10 @@ export class Preferences extends React.Component {
         <Tab value={appURL()} label="Application" data-testid="application-tab" active={isActive(appRoute)}/>
         <Tab value={proxyURL()} label="Proxy" data-testid="proxy-tab" active={isActive(proxyRoute)}/>
         <Tab value={kubernetesURL()} label="Kubernetes" data-testid="kube-tab" active={isActive(kubernetesRoute)}/>
-        <Tab value={telemetryURL()} label="Telemetry" data-testid="telemetry-tab" active={isActive(telemetryRoute)}/>
-        {extensions.length > 0 &&
+        {telemetryExtensions.length > 0 || !!sentryDsn &&
+          <Tab value={telemetryURL()} label="Telemetry" data-testid="telemetry-tab" active={isActive(telemetryRoute)}/>
+        }
+        {extensions.filter(e => !e.showInPreferencesTab).length > 0 &&
           <Tab value={extensionURL()} label="Extensions" data-testid="extensions-tab" active={isActive(extensionRoute)}/>
         }
       </Tabs>
@@ -88,7 +93,7 @@ export class Preferences extends React.Component {
           <Route path={proxyURL()} component={LensProxy}/>
           <Route path={kubernetesURL()} component={Kubernetes}/>
           <Route path={telemetryURL()} component={Telemetry}/>
-          <Route path={extensionURL()} component={Telemetry}/>
+          <Route path={extensionURL()} component={Extensions}/>
           <Redirect exact from={`${preferencesURL()}/`} to={appURL()}/>
         </Switch>
       </SettingLayout>
--- a/src/renderer/components/+preferences/preferences.tsx
+++ b/src/renderer/components/+preferences/preferences.tsx
@@ -23,7 +23,7 @@ import "./preferences.scss";
 import { makeObservable, observable } from "mobx";
 import { observer } from "mobx-react";
 import React from "react";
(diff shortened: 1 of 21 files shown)


src/renderer/components/+preferences/preferences.tsx
```diff
@@ -18,279 +18,95 @@
...
-            <h2>Extensions</h2>
-            {extensions.filter(e => !e.showInPreferencesTab).map(this.renderExtension)}
-          </section>
-        )}
+        <Switch>
+          <Route path={appURL()} component={Application}/>
+          <Route path={proxyURL()} component={LensProxy}/>
+          <Route path={kubernetesURL()} component={Kubernetes}/>
+          <Route path={telemetryURL()} component={Telemetry}/>
+          <Route path={extensionURL()} component={Telemetry}/>
```
Nokel81: Shouldn't this be `Extensions`?

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -18,279 +18,95 @@
...
-          <section id="extensions">
-            <h2>Extensions</h2>
-            {extensions.filter(e => !e.showInPreferencesTab).map(this.renderExtension)}
-          </section>
-        )}
+        <Switch>
+          <Route path={appURL()} component={Application}/>
+          <Route path={proxyURL()} component={LensProxy}/>
+          <Route path={kubernetesURL()} component={Kubernetes}/>
+          <Route path={telemetryURL()} component={Telemetry}/>
```
Nokel81, line 95: I don't think we should show Telemetry if there isn't any preferences for Telemetry

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -18,279 +18,95 @@
...
-        <Tab value={Pages.Kubernetes} label="Kubernetes" data-testid="kube-tab"/>
-        <Tab value={Pages.Telemetry} label="Telemetry" data-testid="telemetry-tab"/>
+        <Tab value={appURL()} label="Application" data-testid="application-tab" active={isActive(appRoute)}/>
+        <Tab value={proxyURL()} label="Proxy" data-testid="proxy-tab" active={isActive(proxyRoute)}/>
+        <Tab value={kubernetesURL()} label="Kubernetes" data-testid="kube-tab" active={isActive(kubernetesRoute)}/>
+        <Tab value={telemetryURL()} label="Telemetry" data-testid="telemetry-tab" active={isActive(telemetryRoute)}/>
         {extensions.length > 0 &&
-          <Tab value={Pages.Extensions} label="Extensions" data-testid="extensions-tab"/>
+          <Tab value={extensionURL()} label="Extensions" data-testid="extensions-tab" active={isActive(extensionRoute)}/>
         }
```
Nokel81: I know you brought this up https://github.com/lensapp/lens/pull/3234#issuecomment-892552066 but I am still wondering where you see the controversy, or do you mean that you would prefer more control over how the settings pages are rendered?

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -18,279 +18,95 @@
...
+          <Route path={telemetryURL()} component={Telemetry}/>
+          <Route path={extensionURL()} component={Telemetry}/>
+          <Redirect exact from={`${preferencesURL()}/`} to={appURL()}/>
+        </Switch>
       </SettingLayout>
     );
   }
 }
+
+export function ExtensionSettings({ title, id, components: { Hint, Input } }: RegisteredAppPreference) {
```
Nokel81, line 104: Why is this exported here?

src/renderer/components/+preferences/preferences.tsx
aleksfront: Made navigation link hidden if no telemetry extensions provided.

src/renderer/components/+preferences/preferences.tsx
aleksfront: This is simple. Problems as I see in #3234
1. We have reduced the control over the styling of elements. In here https://github.com/lensapp/lens/pull/3234/files#diff-8111c23d78b73e3e7dac1c6d8a6060060af33487ccaa99d78326dd9865602d08R108 we iterate every "preference" items with predefined html elements: a title, a hint and a `<hr/>` elements. This looks quite nice, but in reality some sections may look different, some of them may not contain `<hr/>`, some may have more paddings etc. #3234 makes impossible to fine-tune such a things.
2. By moving all preference sections into "registry" we have violated the natural order of the display according to the React. Instead of placing an element inside JSX like so:
```
<section id='sentry' className="small">
  <SubTitle title='Automatic Error Reporting' />
  <Checkbox
    label="Allow automatic error reporting"
    value={UserStore.getInstance().allowErrorReporting}
    onChange={value => {
      UserStore.getInstance().allowErrorReporting = value;
    }}
 />
</section>
```
#3234 produces one level of abstraction to point React where to place elements:
```
{
  id: "sentry",
  title: "Automatic Error Reporting",
  components: {
          Input: () => (
            <Checkbox
              label="Allow automatic error reporting"
              value={UserStore.getInstance().allowErrorReporting}
              onChange={value => {
                UserStore.getInstance().allowErrorReporting = value;
              }}
            />
          ),
          Hint: () => (
            <>
              Automatic error reports provide vital information about issues and application crashes.{" "}
              It is highly recommended to keep this feature enabled to ensure fast turnaround for issues you might encounter.
            </>
          )
        },
        showInPreferencesTab: AppPreferenceKind.Telemetry,
        hide: Boolean(sentryDsn),
}
```

src/renderer/components/+preferences/preferences.tsx
aleksfront: This is used inside 2 separate components. But thanks for noticing, found some other bugs.

aleksfront:
Fixed `Telemetry` and `Extensions` links rendering regarding of available extensions.
extension tabs

nevalla (approved):
LGTM and seems to work as expected. I was thinking what labels we should put to + icons, so that it would be obvious to user what to do. But this is something we can improve later.

PR #3575, opened 2021-08-06
Add the ability to copy k8s object name

mostafahussein:
This PR makes it easier to copy any of the pod names , configmaps and so on by adding a button to the k8s object menu

### DIFF
--- a/src/renderer/components/drawer/drawer.scss
+++ b/src/renderer/components/drawer/drawer.scss
@@ -80,6 +80,17 @@
       padding-right: $padding;
     }
 
+    .drawer-title-text .Icon {
+      opacity: 0;
+      font-weight: normal;
+      margin-left: 8px;
+    }
+     
+    .drawer-title-text:hover .Icon {
+      opacity: 1;
+      transition: opacity 250ms;
+    }
+
     .MenuActions.toolbar .Icon {
       color: $drawerTitleText;
     }
--- a/src/renderer/components/drawer/drawer.tsx
+++ b/src/renderer/components/drawer/drawer.tsx
@@ -22,6 +22,7 @@
 import "./drawer.scss";
 
 import React from "react";
+import { clipboard } from "electron";
 import { createPortal } from "react-dom";
 import { cssNames, noop } from "../../utils";
 import { Icon } from "../icon";
@@ -48,7 +49,12 @@ const defaultProps: Partial<DrawerProps> = {
   onClose: noop,
 };
 
+interface State {
+  isCopied: boolean;
+}
+
 export class Drawer extends React.Component<DrawerProps> {
+
   static defaultProps = defaultProps as object;
 
   private mouseDownTarget: HTMLElement;
@@ -59,6 +65,10 @@ export class Drawer extends React.Component<DrawerProps> {
   private stopListenLocation = history.listen(() => {
     this.restoreScrollPos();
   });
+
+  public state: State = {
+    isCopied: false
+  };
 
   componentDidMount() {
     // Using window target for events to make sure they will be catched after other places (e.g. Dialog)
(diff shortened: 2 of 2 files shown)


aleksfront (changes requested):
@mostafahussein Thanks for contribution! Your addition can be really useful, but it needs to have some of the UI/UX changes.

Separate menu `Copy` menu item may be omitted because it doesn't clearly show what exactly getting copied here. By clicking `Copy`, the pod itself can be copied in some way, it's contents, the name etc.

Instead, it's good to add copy button next to resource (Pod, Deployment, Secret, etc) which will appear on mouse hover.

place for copy button

Is it possible to implement this change?

mostafahussein:
Hello @aleksfront, thank you for your feedback, I have moved the copy functionality as suggested, can you review it again please?

src/renderer/components/drawer/drawer.tsx
```diff
@@ -137,7 +144,10 @@ export class Drawer extends React.Component<DrawerProps> {
         <div className={className} style={style} ref={e => this.contentElem = e}>
           <div className="drawer-wrapper flex column">
             <div className="drawer-title flex align-center">
-              <div className="drawer-title-text">{title}</div>
+              <div className="drawer-title-text">
```
aleksfront: Add some extra class names to make sure copy icon will be placed correctly with some super-long object name


```
<div className="drawer-title-text flex gaps align-center">
```

src/renderer/components/drawer/drawer.scss
```diff
@@ -80,6 +80,17 @@
...
+    .drawer-title-text .Icon {
+      visibility: hidden;
+    }
+
+    .drawer-title-text:hover .Icon {
+      font-weight: normal;
+      flex-grow: 2;
+      visibility: visible;
+      box-shadow: none;
+    }
```
aleksfront, line 92: Some fine-tuning needed to make sure our copy icon looks good. The main thing here is to replace `visibility` prop with `opacity` to make it possible to animate transition on hover.


```
.drawer-title-text .Icon {
  opacity: 0;
  font-weight: normal;
  margin-left: 8px;
}
 
.drawer-title-text:hover .Icon {
  opacity: 1;
  transition: opacity 250ms;
}
```

aleksfront (commented):
Added few style-related comments which will allow icon to look like this
copy button

Also, would be nice if click event will trigger tooltip to be `Copied` instead of `Copy` to say user that feature worked as expected. This can be done by replacing `Copy` -> `Copied` in Icon `onClick` event, and then replacing back `Copied` -> `Copy` in `<div className="drawer-title-text">` `onMouseOut` event.

mostafahussein:
@aleksfront , thanks, that was really helpful. The requested changes are done.

Regarding changing the tooltip suggestion, I decided to implement it similar to what Github is doing when you copy the branch name.

- The icon will change to "done"
- The tooltip will change to "Copied!"
- Everything will be reverted to its original state based on "setTimeOut" instead of using "onMouseOut" event.
- 
What do you think?

mostafahussein:
@aleksfront , thanks, that was really helpful. The requested changes are done.

Regarding what should happen when you click the copy button, I decided to implement it similar to what Github is doing when you copy the branch name or file path.

- The icon will be changed to "done"
- The tooltip will be changed to "Copied!"
- Everything will be reverted to its original state based on setTimeOut instead of using onMouseOut event.

What do you think?

src/renderer/components/drawer/drawer.tsx
```diff
@@ -125,9 +137,21 @@ export class Drawer extends React.Component<DrawerProps> {
...
   };
 
+  copyK8sObjName = () => {
+    const { title } = this.props;
+    const k8sObjName = title.toString().split(': ')[1];
+    clipboard.writeText(k8sObjName);
+    this.setState({copyIcon: "done", isCopied: true});
+    setTimeout(() => {
+      this.setState({copyIcon: "content_copy", isCopied: false});
+    }, 400);
```
aleksfront: Please increase timeout to 3 or 4 seconds. Now it changes too fast:
fast copy button

src/renderer/components/drawer/drawer.tsx
```diff
@@ -125,9 +137,21 @@ export class Drawer extends React.Component<DrawerProps> {
...
+  };
+
   render() {
     const { open, position, title, animation, children, toolbar, size, usePortal } = this.props;
     let { className, contentClass } = this.props;
+    const { isCopied, copyIcon } = this.state;
+    const copyTooltip = isCopied? "Copied!" : "Copy";
```
aleksfront, line 153: We can create also `copyIcon` constant here to keep component state clean (only with 1 value `isCopied`).

aleksfront:
@mostafahussein Added a few tiny comments, please take a look. Otherwise, looks good!

mostafahussein:
@aleksfront  all done 🎉 ,  you can check it now

aleksfront:
There's 2 linter errors over there. Please run `yarn lint:fix`.

aleksfront (approved):
Looking good!

PR #3595, opened 2021-08-12
Fix apiBase initialization on main

jakolehm:


src/common/k8s-api/index.ts
```diff
@@ -23,19 +23,28 @@ import { JsonApi } from "./json-api";
...
-    apiBase: apiPrefix,
-    debug: isDevelopment || isDebugging,
-  }, {
-    headers: {
-      "Host": `localhost:${process.env.LENS_PROXY_PORT}`
-    }
+  appEventBus.addListener((event) => {
+    if (event.name !== "lens-proxy" && event.action !== "listen") return;
+
+    const params = event.params as { port?: number };
```
nevalla, line 35: could be also `const { port } = event.params as { port?: number };`

nevalla (approved):

PR #3598, opened 2021-08-12
Replace no-unused-imports rule to "warning" in dev environment

aleksfront:
Switching from `error` to `warning` for `no-unused-imports` rules to make development process faster. Before this PR, the app can't be recompiled if some imports are left unused, so it's needed to spend some time cleaning imports just to test one thing or another.

Added separation for environments by adding `NODE_ENV=production` rule into `yarn lint` command.

So after the changes this one:
error imports

Becomes this in editor:
warning imports

`yarn lint:fix` still cleaning imports even if `warning` status provided.
linter is fixing

However, PR request lint process will fail as it will use `NODE_ENV=production`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "download:kubectl": "yarn run ts-node build/download_kubectl.ts",
     "download:helm": "yarn run ts-node build/download_helm.ts",
     "build:tray-icons": "yarn run ts-node build/build_tray_icon.ts",
-    "lint": "NODE_ENV=production yarn run eslint --ext js,ts,tsx --max-warnings=0 .",
+    "lint": "ENV=production yarn run eslint --ext js,ts,tsx --max-warnings=0 .",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -37,7 +37,7 @@
     "download:kubectl": "yarn run ts-node build/download_kubectl.ts",
     "download:helm": "yarn run ts-node build/download_helm.ts",
     "build:tray-icons": "yarn run ts-node build/build_tray_icon.ts",
-    "lint": "yarn run eslint --ext js,ts,tsx --max-warnings=0 .",
+    "lint": "NODE_ENV=production yarn run eslint --ext js,ts,tsx --max-warnings=0 .",
```
Nokel81: This seems like an abuse of a `NODE_` prefixed environment variable

aleksfront: Replaced to just `ENV`

Nokel81 (approved):

PR #3602, opened 2021-08-12
Add back sentry winston logger transport

Nokel81:

reverts regression from #3553

### DIFF
--- a/src/common/logger.ts
+++ b/src/common/logger.ts
@@ -20,62 +20,64 @@
  */
 
 import { app, ipcMain, remote } from "electron";
-import winston from "winston";
+import winston, { format } from "winston";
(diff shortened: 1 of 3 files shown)


src/common/logger.ts
```diff
@@ -20,62 +20,64 @@
...
+    })
+  );
 } else {
-  consoleOptions = {
-    handleExceptions: false,
-    level: logLevel,
-    format: winston.format.combine(
-      winston.format.colorize({ level: true, message: false}),
-    )
-  };
```
Nokel81, line 59: This is never used as the options are only used if `ipcMain` is truthy.

jakolehm (approved):

PR #3603, opened 2021-08-12
Polishing Kubernetes section view in Preferences

aleksfront:
* Using same components for iterable elements inside Kubesync and Helm Charts sections
* Moving some of the styles to CSS Modules
* Adding empty states for Kubesync and Helm Charts sections

kubernetes page

syncs empty state
helm charts empty state


Nokel81 (approved):

src/renderer/components/layout/setting-layout.scss
```diff
@@ -228,7 +228,7 @@
       margin-top: 0;
       margin-bottom: 8px;
       padding-bottom: 0;
-      font-size: 12px;
```
Nokel81, line 231: Why this change?

aleksfront: For better readability. I found this font too small to read comfortably.

Nokel81: Sounds good

PR #3623, opened 2021-08-17
Upgrade js-yaml to 4.X

Nokel81:

closed #3479


src/common/k8s-api/endpoints/helm-releases.api.ts
```diff
@@ -116,7 +116,7 @@ export async function createRelease(payload: IReleaseCreatePayload): Promise<IRe
   const { repo, ...data } = payload;
 
   data.chart = `${repo}/${data.chart}`;
-  data.values = jsYaml.safeLoad(data.values);
+  data.values = yaml.load(data.values);
```
Nokel81: @jakolehm The typings for js-yaml@4 now have `yaml.load` return `string | number | object`. However, `data.values` is supposed to be a string, do you remember why the previous call was being made? 

Note: the migration guide basically says that the default should now be the safe version https://github.com/nodeca/js-yaml/blob/49baadd52af887d2991e2c39a6639baa56d6c71b/migrate_v3_to_v4.md

jakolehm: @nevalla ping

nevalla: I think it took values as string and converted it as object

nevalla (dismissed):
Code it self looks good

nevalla (approved):

PR #1125, opened 2020-10-23
Allow extensions to define cluster features

jakolehm:
Fixes #837

PR #3624, opened 2021-08-17
Support window nodes for node-shells

### LINKED ISSUE #2823: windows containers exec error (sh not found)
when I try exec into windows container it tries next and fails(there is no sh on this):
```sh
kubectl exec -i -t -n default gs-zlxrz -c gs "--" sh -c "clear; (bash || ash || sh)"
````

but next works well:
```powershell
kubectl exec -i -t -n default gs-zlxrz -c gs "--" cmd
```

not sure how to autoswitch. i see nodes has osType windows, but what if windows node can run Linux containers?

Nokel81:

fixes #2823

### DIFF
--- a/src/common/k8s-api/endpoints/nodes.api.ts
+++ b/src/common/k8s-api/endpoints/nodes.api.ts
@@ -213,6 +213,7 @@ export class Node extends KubeObject {
 
   getOperatingSystem(): string {
     return this.metadata?.labels?.["kubernetes.io/os"]
+      || this.metadata?.labels?.["beta.kubernetes.io/os"]
       || this.status?.nodeInfo?.operatingSystem
       || "linux";
   }
--- a/src/common/k8s-api/endpoints/nodes.api.ts
+++ b/src/common/k8s-api/endpoints/nodes.api.ts
@@ -212,13 +212,9 @@ export class Node extends KubeObject {
   }
 
   getOperatingSystem(): string {
-    const label = this.getLabels().find(label => label.startsWith("kubernetes.io/os="));
-
-    if (label) {
-      return label.split("=", 2)[1];
-    }
-
-    return "linux";
+    return this.metadata?.labels?.["kubernetes.io/os"]
+      || this.status?.nodeInfo?.operatingSystem
+      || "linux";
   }
 
   isUnschedulable() {
--- a/src/common/k8s-api/kube-api.ts
+++ b/src/common/k8s-api/kube-api.ts
@@ -221,10 +221,7 @@ export class KubeApi<T extends KubeObject> {
       isNamespaced = options.objectConstructor?.namespaced
     } = options || {};
 
-    if (!options.apiBase) {
-      options.apiBase = objectConstructor.apiBase;
-    }
-    const { apiBase, apiPrefix, apiGroup, apiVersion, resource } = parseKubeApi(options.apiBase);
+    const { apiBase, apiPrefix, apiGroup, apiVersion, resource } = parseKubeApi(options.apiBase || objectConstructor.apiBase);
 
     this.kind = kind;
     this.isNamespaced = isNamespaced;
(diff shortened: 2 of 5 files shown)


src/common/k8s-api/endpoints/nodes.api.ts
```diff
@@ -212,13 +212,9 @@ export class Node extends KubeObject {
...
 
   getOperatingSystem(): string {
-    const label = this.getLabels().find(label => label.startsWith("kubernetes.io/os="));
-
-    if (label) {
-      return label.split("=", 2)[1];
-    }
-
-    return "linux";
+    return this.metadata?.labels?.["kubernetes.io/os"]
```
Nokel81, line 215: There is no need to serialize all the labels and then split them again, I also added the status fallback since that is documented to be present as well.

src/common/k8s-api/kube-json-api.ts
```diff
@@ -70,6 +73,20 @@ export interface KubeJsonApiError extends JsonApiError {
...
+  static forCluster(cluster: Cluster): KubeJsonApi {
+    const port = LensProxy.getInstance().port;
+
+    return new this({
+      serverAddress: `http://127.0.0.1:${port}`,
+      apiBase: apiKubePrefix,
+      debug: isDebugging,
+    }, {
+      headers: {
+        "Host": `${cluster.id}.localhost:${port}`,
```
Nokel81, line 85: Would have used `x-cluster-id` instead but it didn't work, it was bypassing the `getProxyTarget` check in `LensProxy` 😕 that is why I removed the `x-cluster-id` handling from below as it seems to be legacy anyway.

src/common/k8s-api/endpoints/nodes.api.ts
nevalla: previously also `beta.kubernetes.io/os` label has been taken into an account.

src/common/k8s-api/kube-api.ts
```diff
@@ -156,10 +156,7 @@ export class KubeApi<T extends KubeObject> {
       isNamespaced = options.objectConstructor?.namespaced
     } = options || {};
 
-    if (!options.apiBase) {
-      options.apiBase = objectConstructor.apiBase;
-    }
-    const { apiBase, apiPrefix, apiGroup, apiVersion, resource } = parseKubeApi(options.apiBase);
+    const { apiBase, apiPrefix, apiGroup, apiVersion, resource } = parseKubeApi(options.apiBase || objectConstructor.apiBase);
```
jakolehm, line 224: We don't anymore mutate `options`.. was that intentional / does it cause any side-effects?

Nokel81: You added it in https://github.com/lensapp/lens/pull/1125 so I don't know if it is intentional. 

Technically this change could be outside observable because we pass in an object reference as options however I don't think we should consider this a breaking change because reusing an options object after using it seems like a definitely unsupported use case. Other than that this shouldn't have any side effects (unless someone passes in a proxy object and want to react to this specific assignment for some reason).

src/main/shell-session/node-shell-session.ts
```diff
@@ -55,10 +57,27 @@ export class NodeShellSession extends ShellSession {
...
+
+    switch (nodeOs) {
+      case "linux":
+        args.push("sh", "-c", "((clear && bash) || (clear && ash) || (clear && sh))");
+        break;
+      case "windows":
+        args.push("powershell");
+        break;
+      default:
+        throw new Error(`Cannot open node shell for ${this.nodeName}: unknown operating sytem ${nodeOs}`);
```
jim-docker: nit: previous behaviour just tried linux style regardless, you could just log a message here and proceed as linux

jim-docker (dismissed):
As suggested by @Nokel81, see if the reporter of #2823 is willing to test on a 5.3.0-alpha

jim-docker (approved):

PR #3635, opened 2021-08-18
Fix filtering deprecated charts

### LINKED ISSUE #3515: Lens doesn't find/list all expected charts
**Describe the bug**
Hi. I have added some repositories like grafana

**To Reproduce**
1. Go to 'Preferences >  Kubernetes > Helm Charts > Select 'grafana' from the list'
2. Click on cluster on the sidebar
3. Scroll down to Apps > Charts
4. Search for Grafana
5. some charts are shown from the grafana repo but not all. The chart for grafana itself doesn't show up. I have tried `helm repo update` and closing/reopening the app but no difference.

**Expected behavior**
I would expect Lens to list all the charts in the enabled repos.

- Lens Version: 5.1.3-latest.20210722.1
- OS: [e.g. OSX] macOS Big Sur 11.5
- Installation method (e.g. snap or AppImage in Linux): Just downloaded the app from the website

**Logs:**
````
https://pastebin.ubuntu.com/p/zwFQKy3hTk/
```

**Kubeconfig:**
```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: ...
    server: https://x.x.x.x:6443
  name: alpha
contexts:
- context:
    cluster: alpha
    user: alpha
  name: alpha
current-context: alpha
kind: Config
preferences: {}
users:
- name: alpha
  user:
    client-certificate-data: ...
    client-key-data: ...e
```

**Additional context**
Not sure of what other info to provide. Thanks in advance!

Nokel81:
- Always show charts, just disable the option to select them in the
  details view

- Optimize the loading of repo charts to sort and normalize once

- Using the structured clone algorithm instead of JSON and fix the cache
  placement

fixes #3515

### DIFF
--- a/src/common/utils/sort-compare.ts
+++ b/src/common/utils/sort-compare.ts
@@ -19,7 +19,9 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import semver, { SemVer } from "semver";
+import semver, { coerce, SemVer } from "semver";
+import * as iter from "./iter";
+import type { RawHelmChart } from "../k8s-api/endpoints/helm-charts.api";
 
 export function sortCompare<T>(left: T, right: T): -1 | 0 | 1 {
   if (left < right) {
@@ -52,4 +54,33 @@ export function sortCompareChartVersions(left: ChartVersion, right: ChartVersion
   }
 
   return sortCompare(left.version, right.version);
+}
+
+
+
+export function sortCharts(charts: RawHelmChart[], log?: (...args: any[]) => void) {
+  interface ExtendedHelmChart extends RawHelmChart {
+    __version: SemVer;
+  }
+
+  const chartsWithVersion = Array.from(
+    iter.map(
+      charts,
+      (chart => {
+        const __version = coerce(chart.version, { includePrerelease: true, loose: true });
+
+        if (!__version) {
+          log?.(`[HELM-SERVICE]: Version from helm chart is not loosely coercable to semver.`, { name: chart.name, version: chart.version, repo: chart.repo });
+        }
+
+        (chart as ExtendedHelmChart).__version = __version;
+
+        return chart as ExtendedHelmChart;
+      })
+    ),
+  );
+
+  return chartsWithVersion
+    .sort(sortCompareChartVersions)
+    .map(chart => (delete chart.__version, chart));
 }
--- a/src/common/utils/sort-compare.ts
+++ b/src/common/utils/sort-compare.ts
@@ -22,6 +22,7 @@
 import semver, { coerce, SemVer } from "semver";
 import * as iter from "./iter";
 import type { RawHelmChart } from "../k8s-api/endpoints/helm-charts.api";
+import logger from "../logger";
 
 export function sortCompare<T>(left: T, right: T): -1 | 0 | 1 {
   if (left < right) {
@@ -58,7 +59,7 @@ export function sortCompareChartVersions(left: ChartVersion, right: ChartVersion
 
 
 
-export function sortCharts(charts: RawHelmChart[], log?: (...args: any[]) => void) {
+export function sortCharts(charts: RawHelmChart[]) {
   interface ExtendedHelmChart extends RawHelmChart {
     __version: SemVer;
   }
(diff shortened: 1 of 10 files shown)


src/main/helm/helm-chart-manager.ts
```diff
@@ -20,86 +20,100 @@
...
+              })),
+              logger.warn,
+            ),
+          ] as const)
+          .filter(([, charts]) => !charts.every(chart => chart.deprecated))
+      );
+
+      if (this.repo.name === "grafana") {
+        console.log(JSON.stringify(normalized));
       }
```
aleksfront: Probably left from debugging?

src/main/helm/helm-chart-manager.ts
```diff
@@ -20,86 +20,100 @@
...
-
-type CachedYaml = {
-  entries: RepoHelmChartList
-};
+import { sortCharts } from "../../common/utils";
 
 export class HelmChartManager {
-  protected cache: any = {};
-  protected repo: HelmRepo;
+  static #cache = new Map<string, Buffer>();
```
aleksfront, line 33: We're not using such style anywhere else. Probably better to omit the hash sign?

src/main/helm/helm-chart-manager.ts
```diff
@@ -20,86 +20,100 @@
...
+        Object.entries(entries)
+          .map(([name, charts]) => [
+            name,
+            sortCharts(
+              charts.map(chart => ({
+                ...chart,
+                created: Date.parse(chart.created).toString(),
+                repo: this.repo.name,
+              })),
+              logger.warn,
```
aleksfront: Just to make sure. Will `logger.warm` be okay for both prod and dev environment?

aleksfront (commented):
Left few minor comments. While testing works okay, fixes the issues listed in #3515. However, unable to find any deprecated chart versions to test this one (versions with `line-through` visibility) https://github.com/lensapp/lens/pull/3635/files#diff-4ada9d82cde72ab9fa11079e8dd1c571c9d3de88a5a5ff0494ac8c04eea92a25

src/main/helm/helm-chart-manager.ts
Nokel81: This is a better version of `private` as it is actually enforced at runtime

src/main/helm/helm-chart-manager.ts
Nokel81: Yes of course, though this was mostly don't so that we don't have warnings in the tests, I think I will change this.

Nokel81:
> However, unable to find any deprecated chart versions to test this one (versions with line-through visibility)

Sure, you can add the `grafana             	https://grafana.github.io/helm-charts` repo, then the `grafana` chart has versions 5.5.7 and 5.5.6 which are deprecated.

src/main/helm/helm-chart-manager.ts
aleksfront: Why is this better? And why mark `static` property as private? Asking because static property can not be called on instance of a class.

src/common/utils/sort-compare.ts
```diff
@@ -53,3 +56,32 @@ export function sortCompareChartVersions(left: ChartVersion, right: ChartVersion
...
+export function sortCharts(charts: RawHelmChart[]) {
+  interface ExtendedHelmChart extends RawHelmChart {
+    __version: SemVer;
+  }
+
+  const chartsWithVersion = Array.from(
+    iter.map(
+      charts,
+      (chart => {
+        const __version = coerce(chart.version, { includePrerelease: true, loose: true });
```
aleksfront, line 71: What is the meaning to use underscores before variable name here? `__version`?

src/main/helm/helm-chart-manager.ts
```diff
@@ -20,86 +20,95 @@
...
+      return await this.cachedYaml();
     } catch(error) {
       logger.error("HELM-CHART-MANAGER]: failed to list charts", { error });
 
       return {};
     }
   }
 
-  public async getReadme(name: string, version = "") {
+  private async command(action: string, name: string, version?: string) {
```
aleksfront:
```suggestion
  private async executeCommand(action: string, name: string, version?: string) {
```

src/main/helm/helm-chart-manager.ts
Nokel81: Same reason to mark an instance property as `private`, so to better model the encapsulation.

A `static` property that is `#` (JS private) cannot be accessed by code that is not a static member or an instance member.

src/common/utils/sort-compare.ts
Nokel81: I meant it as both "temporary" and less likely to conflict with other fields. I chose `__version` because it is related to the `version` field but I don't want to override that field.

src/main/helm/helm-chart-manager.ts
Nokel81: sure

aleksfront (approved):
LGTM

PR #3647, opened 2021-08-19
Support filtering catalog entities

### LINKED ISSUE #3679: Catalog Filter Extension API
Create Extension API to filter Catalog entities.

The API could look something like this:
```
LensMainExtension.addCatalogFilter = (entity: Common.Catalog.CatalogEntity): boolean;
```

If the method returns false, the catalog entity is filtered out and therefore not displayed in the catalog.

All catalog entities are filtered through every extension's `addCatalogFilter`, and if one returns false, the entity is filtered.

Nokel81:
- This allows extensions and OpenLens to restrict which of the entities
  the sources have provied are "visible" to the rest of Lens
- Fixes https://github.com/lensapp/lens/issues/3679

### DIFF
--- a/src/common/utils/iter.ts
+++ b/src/common/utils/iter.ts
@@ -156,3 +156,19 @@ export function find<T>(src: Iterable<T>, match: (i: T) => any): T | undefined {
 
   return void 0;
 }
+
+/**
+ * Iterate over `src` calling `reducer` with the previous produced value and the current
+ * yielded value until `src` is exausted. Then return the final value.
+ * @param reducer A function for producing the next item from an accumilation and the current item
+ * @param initial The initial value for the iteration
+ */
+export function reduce<T, R = T>(src: Iterable<T>, reducer: (acc: R, cur: T) => R, initial: R): R {
+  let acc = initial;
+
+  for (const item of src) {
+    acc = reducer(acc, item);
+  }
+
+  return acc;
+}
--- a/src/common/utils/iter.ts
+++ b/src/common/utils/iter.ts
@@ -160,6 +160,7 @@ export function find<T>(src: Iterable<T>, match: (i: T) => any): T | undefined {
 /**
  * Iterate over `src` calling `reducer` with the previous produced value and the current
  * yielded value until `src` is exausted. Then return the final value.
+ * @param src The value to iterate over
  * @param reducer A function for producing the next item from an accumilation and the current item
  * @param initial The initial value for the iteration
  */
--- a/src/common/utils/iter.ts
+++ b/src/common/utils/iter.ts
@@ -163,6 +163,8 @@ export function find<T>(src: Iterable<T>, match: (i: T) => any): T | undefined {
  * @param reducer A function for producing the next item from an accumilation and the current item
  * @param initial The initial value for the iteration
  */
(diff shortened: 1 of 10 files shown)


src/extensions/lens-main-extension.ts
```diff
@@ -33,6 +34,19 @@ export class LensMainExtension extends LensExtension {
     return WindowManager.getInstance().navigateExtension(this.id, pageId, params, frameId);
   }
 
+  /**
+   * Add a filtering function for the catalog. This will be removed if the extension is disabled.
+   * @param fn The function which should return a truthy value for those entities which should be kepted
+   * @returns A function to clean up the filter
+   */
+  addCatalogFilter(fn: EntityFilter): Disposer {
```
jakolehm: I'm not sure if this is the correct place for filters. I would thought that filter only applies to visible items, not to whole catalog data. Doing filtering at this level might cause unintended side-effects to other extensions.

jakolehm: My suggestion: let's do filtering on renderer side, maybe only for Catalog UI.

Nokel81: Sure I guess, I would have figured that this sort of filtering was supposed to be pervasive. That way it actually feels like switching.

jakolehm: It cannot be pervasive because it would mean that filtered items would show as disabled on a hotbar.

Nokel81: Ah ok, I'll update the PR to only be on the render side and only be for the catalog views

src/common/utils/iter.ts
```diff
@@ -156,3 +156,21 @@ export function find<T>(src: Iterable<T>, match: (i: T) => any): T | undefined {
 
   return void 0;
 }
+
+/**
+ * Iterate over `src` calling `reducer` with the previous produced value and the current
+ * yielded value until `src` is exausted. Then return the final value.
+ * @param reducer A function for producing the next item from an accumilation and the current item
```
nevalla, line 164: doc block missing `src` param

src/common/utils/iter.ts
```diff
@@ -156,3 +156,21 @@ export function find<T>(src: Iterable<T>, match: (i: T) => any): T | undefined {
...
+ * @param initial The initial value for the iteration
+ */
+export function reduce<T, R>(src: Iterable<T>, reducer: (acc: Iterable<R>, cur: T) => Iterable<R>, initial: Iterable<R>): Iterable<R>;
+
+export function reduce<T, R = T>(src: Iterable<T>, reducer: (acc: R, cur: T) => R, initial: R): R {
```
nevalla, line 169: Ideally would be nice to have unit tests for this method. Probably `CatalogEntityRegistryTest` is covering this also, but in general.

nevalla (dismissed):
Added couple of minor comments, but in general LGTM

src/common/utils/iter.ts
panuhorsmalahti: Fixed

src/common/utils/iter.ts
panuhorsmalahti: Added few basic tests

nevalla (approved):

aleksfront (approved):

jakolehm (approved):

PR #3658, opened 2021-08-23
Fix @material-ui/core console errors

aleksfront:
@material-ui/lab dependency should be bumped after https://github.com/lensapp/lens/pull/3554 to avoid some warnings.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -256,7 +256,7 @@
     "@emeraldpay/hashicon-react": "^0.4.0",
     "@material-ui/core": "^4.12.3",
     "@material-ui/icons": "^4.11.2",
-    "@material-ui/lab": "^4.0.0-alpha.57",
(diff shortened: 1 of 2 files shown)


chenhunghan (approved):

aleksfront:
Can't make `Require Release Category Labels / label` have although I have all labels set.

aleksfront:
Can't make `Require Release Category Labels / label` have although I have all labels set.

aleksfront:
@Nokel81 I found that setting both `chore` and `bug` breaks the `Require Release Category Labels` check. How do you think is this an issue or not?

Nokel81:
@aleksfront Those labels are used for the release drafter which is why the need to be XOR. I would label this a bug

PR #3663, opened 2021-08-23
Add WelcomeBanner.width extension API

chenhunghan:
Previously, the width of welcome banner is hard-coded at `320px`, this PR remove the restriction and allow extension to register the welcome banner with width. Also remove some unused CSS.

Example:
```tsx
import { Renderer } from "@k8slens/extensions";
import { css } from "@emotion/css";

const width = 840;

export default class LensCloudExtension extends Renderer.LensExtension {
  welcomeBanners = [{
      width,
      Banner: () => (
        <webview
          src="https://k8slens.dev/spaces"
          className={css`
            width: ${840}px;
          `}
        >
        </webview>
      )
    }]
}
```

Screenshot 2021-08-23 at 17 33 51

### DIFF
--- a/src/renderer/components/+welcome/welcome.tsx
+++ b/src/renderer/components/+welcome/welcome.tsx
@@ -29,7 +29,7 @@ import { WelcomeMenuRegistry } from "../../../extensions/registries";
 import { WelcomeTopbar } from "../cluster-manager/welcome-topbar";
 import { WelcomeBannerRegistry } from "../../../extensions/registries";
 
-const defaultWidth = 320;
+export const defaultWidth = 320;
 
 @observer
 export class Welcome extends React.Component {
@@ -90,5 +90,3 @@ export class Welcome extends React.Component {
     );
   }
 }
-
-export { defaultWidth };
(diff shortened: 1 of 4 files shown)


src/renderer/components/+welcome/welcome.tsx
```diff
@@ -29,46 +29,66 @@ import { WelcomeMenuRegistry } from "../../../extensions/registries";
...
+    // if there is banner with specified width, use it to calculate the width of the container
+    const maxWidth = welcomeBanner.reduce((acc, curr) => {
+      const currWidth = curr.width ?? 0;
+
+      if (acc > currWidth) {
+        return acc;
+      }
+
+      return currWidth;
+    }, defaultWidth);
```
Nokel81, line 48: nit:

```ts
const maxWidth = Math.max(defaultWidth, ...welcomeBanner.map(b => b.width ?? 0));
```

src/renderer/components/+welcome/welcome.tsx
```diff
@@ -29,46 +29,66 @@ import { WelcomeMenuRegistry } from "../../../extensions/registries";
...
+              </div>
+            </div>
           </div>
         </div>
       </>
     );
   }
 }
+
+export { defaultWidth };
```
Nokel81: why not just use `export const` above?

chenhunghan: Old habit to export at the bottom...  I will change.

chenhunghan: Done https://github.com/lensapp/lens/pull/3663/commits/ae30eb3aeb74b8e9146296358859e8704df8883e

jakolehm (approved):

PR #3668, opened 2021-08-24
Fix node shell secret pulling

Nokel81:
- On some KE errors are thrown if trying to get a secret with an empty
  name, instead of just ignoring the entry

- Clean up the node shell settings

Once this is merged, I will start a quick 5.1.4 release. And merge it there too.

### DIFF
--- a/src/renderer/components/cluster-settings/components/cluster-node-shell-setting.tsx
+++ b/src/renderer/components/cluster-settings/components/cluster-node-shell-setting.tsx
@@ -20,11 +20,11 @@
  */
 
 import type { Cluster } from "../../../../main/cluster";
-import { autorun, makeObservable, observable } from "mobx";
+import { makeObservable, observable } from "mobx";
 import { SubTitle } from "../../layout/sub-title";
 import React from "react";
 import { Input } from "../../input/input";
-import { disposeOnUnmount, observer } from "mobx-react";
+import { observer } from "mobx-react";
 import { Icon } from "../../icon/icon";
 import { initialNodeShellImage } from "../../../../common/cluster-types";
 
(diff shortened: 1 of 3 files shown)


src/renderer/components/cluster-settings/components/cluster-node-shell-setting.tsx
```diff
@@ -34,59 +34,42 @@ interface Props {
 
 @observer
 export class ClusterNodeShellSetting extends React.Component<Props> {
-  @observable nodeShellImage = "";
-  @observable imagePullSecret = "";
+  @observable nodeShellImage = this.props.cluster.preferences?.nodeShellImage || "";
+  @observable imagePullSecret = this.props.cluster.preferences?.imagePullSecret || "";
```
nevalla, line 38: probably more logical place for these assignments would be in constructor

nevalla (approved):

jakolehm (approved):

src/renderer/components/cluster-settings/components/cluster-node-shell-setting.tsx
```diff
@@ -95,15 +78,25 @@ export class ClusterNodeShellSetting extends React.Component<Props> {
...
+                    onClick={() => this.imagePullSecret = ""}
+                    tooltip="Clear"
+                  />
+                )
+                : undefined
+            }
           />
           <small className="hint">
-            Name of a pre-existing secret (optional). Used for pulling image from a private registry.
+            Name of a pre-existing secret. An optional setting. Used for pulling image from a private registry.
```
kcak11, line 99: Will it be a good idea that this message also specifies the "Namespace" for the pre-existing secret i.e. kube-system ??

See: https://github.com/lensapp/lens/pull/3732

PR #3681, opened 2021-08-26
Remove leading + in component folder names

### LINKED ISSUE #3481: Remove legacy plus sign (+) from component directories

Nokel81:

fixes #3481


jakolehm:
I think targeting 5.2 is a bit too optimistic given how many files this moves?

Nokel81:
@jakolehm I just targeted the same version as the issue. But I agree I don't think it will happen.

PR #3682, opened 2021-08-26
Support MacOS arm64 build

jakolehm:


build/download_kubectl.ts
```diff
@@ -115,15 +116,21 @@ class KubectlDownloader {
...
+
+const downloads = [];
+
+if (isMac) {
+  downloads.push({ platform: "darwin", arch: binaryArch, target: path.join(baseDir, "darwin", binaryTargetArch, "kubectl") });
+} else if (isLinux) {
+  downloads.push({ platform: "linux", arch: binaryArch, target: path.join(baseDir, "linux", binaryTargetArch, "kubectl") });
+} else {
+  downloads.push({ platform: "windows", arch: "amd64", target: path.join(baseDir, "windows", binaryTargetArch, "kubectl.exe") });
+  downloads.push({ platform: "windows", arch: "386", target: path.join(baseDir, "windows", "ia32", "kubectl.exe") });
```
Nokel81, line 132: Why do we download both here but not for either of the above? is this also so that we don't download the versions for the other OS's?

package.json
```diff
@@ -29,8 +29,7 @@
     "build:mac": "yarn run compile && electron-builder --mac --dir",
     "build:win": "yarn run compile && electron-builder --win --dir",
     "integration": "jest --runInBand --detectOpenHandles --forceExit integration",
-    "dist": "yarn run compile && electron-builder --publish onTag",
-    "dist:win": "yarn run compile && electron-builder --publish onTag --x64 --ia32",
+    "dist": "yarn run compile && electron-builder --publish onTag --x64",
```
Nokel81: Why do we pass in any arch arg here?

jakolehm: We shouldn't, I'll fix it.

build/download_kubectl.ts
jakolehm: We have downloaded all binaries for all OSes in the past for no reason. Now that I had to improve this anyway I did this small optimization (it also makes `make dev` etc a bit faster).

Note that Windows is still a special case because we support old 32bit systems there.

Nokel81: Ah but we don't build on a 32-bit VM so we have to special case it, makes sense.

Nokel81 (dismissed):

Nokel81 (approved):

PR #3692, opened 2021-08-26
Add ability to search and activate entities via the command dialog

### LINKED ISSUE #1011: Selecting a cluster by search

A text box to select clusters. As you type to narrow your search a dropdown shows the clusters that match your current input and you can select the cluster you want. The matching can be wild curd, regex or fuzzy search.

**Why is this needed**:

I work regularly with tens of clusters that I need to switch between (multiple types of clusters, each type in multiple geographical regions and staging and production clusters in each region).

The icons on the sidebar are super cool, but I have to scroll down the list and hover over each icon to discover the cluster I want. With 20+ clusters this is super tedious.

- Kubernetes distribution: GKE
- Desktop OS: macOS Catalina

Nokel81:

fixes https://github.com/lensapp/lens/issues/1011

### DIFF
--- a/src/extensions/registries/base-registry.ts
+++ b/src/extensions/registries/base-registry.ts
@@ -25,7 +25,7 @@ import { Singleton } from "../../common/utils";
 import { LensExtension } from "../lens-extension";
 
 export class BaseRegistry<T, I = T> extends Singleton {
-  private items = observable.map<T, I>([], {deep: false});
+  private items = observable.map<T, I>([], { deep: false });
 
   constructor() {
     super();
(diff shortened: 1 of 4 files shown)


ixrock (approved):
LGTM.
One side note, currently we can go-to entity via 2 steps from command-dialog:

Screenshot 2021-09-23 at 14 33 33
Screenshot 2021-09-23 at 14 33 40

Maybe better to make a flat list of all commands (with some prefix), like in `ClusterView: ...`? Otherwise I cannot find my cluster right away by name and should remember to type "active entity..". Wdyt?

Screenshot 2021-09-23 at 14 34 00

Nokel81:
I think for this PR we should leave it as it is because I don't think we should pollute the top level of the command dialog too much.

If that becomes a problem then I think we could have a different shortcut to directly open the second dialog.

How does that sound @ixrock ?

ixrock:
> I think for this PR we should leave it as it is because I don't think we should pollute the top level of the command dialog too much.
> If that becomes a problem then I think we could have a different shortcut to directly open the second dialog.> 
> How does that sound @ixrock ?

Sounds good.

PR #3696, opened 2021-08-30
Topbar history buttons

### LINKED ISSUE #3617: Improve Top Bar
- remove close button for clusters
- add recently used drop down to show last used catalog entities
- back / forward arrows
- adding cluster icon + name to cluster view (iframe)

image (5)
image (4)

Screens:
https://docs.google.com/presentation/d/1ER32kCkGaXE80cEoC-Tj5YWVGHHxLoaW8ykd7qiX38Q/edit#slide=id.gc955d27163_0_38

aleksfront:
* Adding back and next buttons on topbar as part of #3617 
* Removing close button from the top-right corner
* Moving `Catalog` or cluster name to Sidebar
* Using in-browser `goBack` and `goForward` methods (so triggering them equal to hitting `cmd+[` and `cmd+]`)

Updating buttons enabled/disabled state was made by throwing ipc events. The reason is to update buttons from both main and iframe views.

https://user-images.githubusercontent.com/9607060/131298703-7aab5865-124c-4287-887b-fcd4b78b21bf.mov


src/renderer/components/layout/sidebar.tsx
```diff
@@ -181,6 +183,9 @@ export class Sidebar extends React.Component<Props> {
 
     return (
       <div className={cssNames(Sidebar.displayName, "flex column", className)}>
+        <div className="cluster-name">
+          {ClusterStore.getInstance().getById(App.clusterId).preferences.clusterName}
```
jim-docker: this should be
```suggestion
          {ClusterStore.getInstance().getById(App.clusterId)?.name}
```

jakolehm: I agree with @jim-docker

src/renderer/components/icon/flat_arrow.svg
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
```
jakolehm: Just wondering why we didn't use material icon arrows here.

src/renderer/components/layout/sidebar.tsx
aleksfront: Fixed.

src/renderer/components/icon/flat_arrow.svg
aleksfront: To fully match the wireframes we have.

aleksfront:
Fixed to use material icons instead of custom ones.

material history icons

jakolehm (approved):
LGTM

PR #3706, opened 2021-09-01
Allow to configure default namespace used in terminal

nevalla:
This PR allows to configure default namespace in cluster settings. This namespace is set as default namespace for kubectl running in terminal.

image

image


src/main/cluster.ts
```diff
@@ -216,6 +216,18 @@ export class Cluster implements ClusterModel, ClusterState {
     return toJS({ prometheus, prometheusProvider });
   }
 
+  /**
+   * defaultNamespace preference
+   *
+   * @computed
+   * @internal
```
jakolehm, line 223: internal not needed because `Cluster` is not anymore directly exposed to the extension api.

jakolehm (approved):

PR #3717, opened 2021-09-02
Make ipc.boradcast not stopping at uncaught exceptions, improve typing, improve error logs

chenhunghan:
- Make "broadcast" not stopping at an uncaught exception. Change/fix the `try...catch` block logic, make it continue to `view.sendToFrame` even if `view.send` throw, and continue if one of `view.sendToFrame` throws.
- Improve type of `views`.
- Better error message with `channel`, `view.type`, `view.id` and `subframe.id` for easier debugging.


jakolehm (approved):
LGTM

PR #3718, opened 2021-09-02
Add CatalogCategory filter to the extension API

panuhorsmalahti:
* API:
```
  /**
   * Add a new filter to the set of category filters
   * @param fn The function that should return a truthy value if that category should be displayed
   * @returns A function to remove that filter
   */
  addCatalogCategoryFilter(fn: CategoryFilter): Disposer ;
```
* Example usage of the API:
```
this.addCatalogCategoryFilter(category => category.metadata.name === "General");
```
* This API will hide Catalog categories from the Catalog menu according to the filter.
* `addCatalogCategoryFilter` returns a dispose function, which if called, will remove the filter.
* Filters are automatically removed on extension deactivation.
* If an added filter removes the currently active Catalog category, the active category is automatically switched to the first category in the menu, if one exists

### DIFF
--- a/src/common/__tests__/catalog-category-registry.test.ts
+++ b/src/common/__tests__/catalog-category-registry.test.ts
@@ -72,4 +72,34 @@ describe("CatalogCategoryRegistry", () => {
     d2();
     expect(registry.items.length).toBe(0);
   });
+
+  it("doesn't return items that are filtered out", () => {
+    const registry = new TestCatalogCategoryRegistry();
+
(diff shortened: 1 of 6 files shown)


jakolehm:
@panuhorsmalahti tests are still failing.

jakolehm:
Moved to 5.3.0.

chenhunghan (dismissed):

Nokel81 (approved):

PR #3722, opened 2021-09-02
Add catalogAddMenu filter to CatalogCategory

panuhorsmalahti:
* Add filter for `catalogAddMenu` context menu items. Extension can e.g. filter certain menu items from the context menu.
* Example use:
```
const disposer = kubernetesClusterCategory.addMenuFilter(item => item.title === "Add from kubeconfig");
```
The filter can be removed with the returned disposer:
```
disposer();
```
* This is supported for all `CatalogCategory` classes.


chenhunghan (dismissed):
LGTM

src/common/catalog-entities/__tests__/kubernetes-cluster.test.ts
```diff
@@ -0,0 +1,62 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+import { kubernetesClusterCategory }  from "../kubernetes-cluster";
+
+describe("kubernetesClusterCategory", () => {
+  describe("filteredItems", () => {
```
chenhunghan, line 24: maybe it's more useful to add test for `<CatalogAddButton  />`, but I guess this is fine too..

panuhorsmalahti: True, tests added

chenhunghan (dismissed):
Looks great

src/common/catalog/catalog-entity.ts
```diff
@@ -72,6 +83,32 @@ export abstract class CatalogCategory extends (EventEmitter as new () => TypedEm
...
+
+  /**
+   * Add a filter for menu items of catalogAddMenu
+   * @param fn The function that should return a truthy value if that menu item should be displayed
+   * @returns A function to remove that filter
+   */
+  public addMenuFilter(fn: AddMenuFilter): Disposer {
+    this.filters.add(fn);
+
+    return once(() => void this.filters.delete(fn));
```
Nokel81, line 95: Only nit would be that if the category is removed, these should be removed too.

Nokel81 (approved):

src/common/catalog/catalog-entity.ts
panuhorsmalahti: Why? Do you mean if I call the `Disposer` returned by `CatalogCategoryRegistry::add`, I should remove the filters from the `CatalogCategory` instance? Doesn't sound right to me

Nokel81: Yeah probably not.

PR #3740, opened 2021-09-06
TopBar home icon

### LINKED ISSUE #3703: Home Icon on Top Bar to open Catalog
- [x] Add grey home icon on top bar, next to back / forward buttons (left hand side)
- [x] For now it will open up Catalog
- [x] Catalog should get highlighted in case it is an entity on the hot bar
- [x] Add new Welcome Page Entry on General Catalog Category (Icon: ?)

aleksfront:
* Adding `home` icon in addition to history buttons which leads to `Catalog` page
* Highlight general entities in hotbar if their route match to current `location`
* Adding `Welcome` page entity which can be pinned to hotbar

Resolves #3703 

https://user-images.githubusercontent.com/9607060/132209006-906f390c-ff1d-4990-bee0-9761a30e5094.mov

### DIFF
--- a/src/common/catalog-entities/general.ts
+++ b/src/common/catalog-entities/general.ts
@@ -19,7 +19,9 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { navigate } from "../../renderer/navigation";
+import { reaction, when } from "mobx";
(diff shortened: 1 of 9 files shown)


msa0311 (approved):

nevalla:
If we have now Home icon that opens catalog on top bar, do we still need catalog pinned on hotbar? Just a question, not related to this PR.

nevalla (approved):

PR #3749, opened 2021-09-07
don't allow delete of the active cluster before setting to new active cluster

jim-docker:
- "Delete Context" button is disabled if the "select current-context" checkbox is checked and no context has been selected from the dropdown.
- The "select current-context" checkbox cannot be unchecked if the cluster to delete is the active cluster
- The file path of a Lens private config kept in Lens's app data is not shown in the delete dialog message


aleksfront:
Added a quick fix to enable/disable delete button logic to avoid this state when "Delete context" is disabled:

delete last context

aleksfront (approved):
LGTM

jim-docker:
> Added a quick fix to enable/disable delete button logic to avoid this state when "Delete context" is disabled:


good catch!

PR #3751, opened 2021-09-08
Use entity name as cluster name in sidebar

### LINKED ISSUE #3741: Cluster name may differs in sidebar and hotbar
image (1)

Cluster sidebar takes `cluster.preferences.clusterName` to show the name, but hotbar uses `entity.metadata.name`.

aleksfront:
Using entity name in cluster sidebar and forcing catalog items broadcast on iframe load (to catch entites).

cluster name

changing cluster name

Fixes #3741

### DIFF
--- a/src/renderer/components/app.tsx
+++ b/src/renderer/components/app.tsx
@@ -19,7 +19,7 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 import React from "react";
-import { observable, makeObservable } from "mobx";
+import { observable, makeObservable, reaction } from "mobx";
 import { disposeOnUnmount, observer } from "mobx-react";
 import { Redirect, Route, Router, Switch } from "react-router";
 import { history } from "../navigation";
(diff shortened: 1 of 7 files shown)


src/renderer/components/app.tsx
```diff
@@ -92,19 +92,12 @@ export class App extends React.Component {
     logger.info(`[APP]: Init dashboard, clusterId=${App.clusterId}, frameId=${frameId}`);
     await Terminal.preloadFonts();
     await requestMain(clusterSetFrameIdHandler, App.clusterId);
+    await requestMain(catalogBroadcastHandler);
 
     const cluster = ClusterStore.getInstance().getById(App.clusterId);
 
     await cluster.whenReady; // cluster.activate() is done at this point
 
-    const activeEntityDisposer = reaction(() => catalogEntityRegistry.getById(App.clusterId), (entity) => {
```
jakolehm: I think we need to keep this... or how does `activeEntity` get set without?

src/renderer/components/cluster-settings/components/cluster-name-setting.tsx
```diff
@@ -43,7 +45,7 @@ export class ClusterNameSetting extends React.Component<Props> {
   componentDidMount() {
     disposeOnUnmount(this,
       autorun(() => {
-        this.name = this.props.cluster.preferences.clusterName;
+        this.name = this.props.cluster.preferences.clusterName || this.props.entity.metadata.name;
```
jakolehm, line 48: Not sure if this is needed, probably ok to have it.

jakolehm (approved):

src/renderer/components/cluster-settings/components/cluster-name-setting.tsx
aleksfront: I found locally that for some reason `clusterName` prop was `undefined` for my minikube cluster. That was caused empty input field and errors in console.

PR #3795, opened 2021-09-14
Fix shell connection readiness heuristic

### LINKED ISSUE #3757: Terminal shell is taking a long time to complete startup
**Describe the bug**
I regularly see the message `Terminal shell is taking a long time to complete startup` followed by a suggestion to check my .rc. file.

How do I diagnose what is causing the slow startup?

Thanks.

**To Reproduce**
1. Go to 'nodes' view
2. Click on a node and select an action (drain or cordon).
3. Dismiss 'are you sure' prompt
4. Terminal appears sitting on the 'PS C:\Users\[username]' prompt.
5. Wait 5 seconds
6. Blue popup containing the above message appears
7. Requested command executes in the terminal window.

**Expected behavior**

Skip step 5. 

Note that on other screens (like connecting to a POD shell) there is no wait.

**Screenshots**

- Lens Version: 5.2.0
- OS: Windows 10 x64
- Installation method: Lens Setup 5.2.0-latest.20210908.1.exe

Note, 5.1.5 has the same issue.  4.x never had this issue.  Unfortunately I had no reason to try any 5.x betas

Nokel81:

fixes #3757

### DIFF
--- a/src/renderer/api/terminal-api.ts
+++ b/src/renderer/api/terminal-api.ts
@@ -119,9 +119,9 @@ export class TerminalApi extends WebSocketApi {
      * its own rc file (or RunCommands file) such as `.bashrc` or `.zshrc`.
      *
      * This heuistic assumes that the prompt line of a terminal is a single line
-     * and ends with a whitespace character.
+     * and ends with a whitespace character or the > character (windows's CMD).
      */
-    if (data.match(/\r?\n/) === null && data.match(/\s$/)) {
+    if (data.match(/\r?\n/) === null && data.match(/(\s|>)$/)) {
       this.shellRunCommandsFinished = true;
       this.onData.removeListener(this._onShellRunCommandsFinished);
     }
--- a/src/renderer/api/terminal-api.ts
+++ b/src/renderer/api/terminal-api.ts
@@ -58,7 +58,6 @@ export class TerminalApi extends WebSocketApi {
 
   public onReady = new EventEmitter<[]>();
   @observable public isReady = false;
-  @observable public shellRunCommandsFinished = false;
   public readonly url: string;
 
   constructor(protected options: TerminalApiQuery) {
@@ -92,7 +91,6 @@ export class TerminalApi extends WebSocketApi {
   connect() {
     this.emitStatus("Connecting ...");
     this.onData.addListener(this._onReady, { prepend: true });
-    this.onData.addListener(this._onShellRunCommandsFinished);
     super.connect(this.url);
   }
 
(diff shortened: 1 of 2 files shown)


src/renderer/api/terminal-api.ts
```diff
@@ -119,9 +119,9 @@ export class TerminalApi extends WebSocketApi {
      * its own rc file (or RunCommands file) such as `.bashrc` or `.zshrc`.
      *
      * This heuistic assumes that the prompt line of a terminal is a single line
-     * and ends with a whitespace character.
+     * and ends with a whitespace character or the > character (windows's CMD).
```
jim-docker: The prompt is customizable so this may not be true anyway...

src/renderer/components/dock/terminal.store.ts
```diff
@@ -120,7 +120,7 @@ export class TerminalStore extends Singleton {
       const rcIsFinished = when(() => this.connections.get(tab.id).shellRunCommandsFinished);
       const notifyVeryLong = setTimeout(() => {
         rcIsFinished.cancel();
-        Notifications.info("Terminal shell is taking a long time to complete startup. Please check your .rc file. Bypassing shell completion check.", {
+        Notifications.info("Terminal shell is taking a long time to complete startup. Bypassing shell completion check.", {
```
jim-docker: Or:
```suggestion
        Notifications.info("Terminal shell is taking a long time to complete startup. Please check your .rc file, if applicable. Bypassing shell completion check.", {
```

Nokel81: Fair enough

src/renderer/api/terminal-api.ts
Nokel81: That is why I consider it a heuristic. It is very real race condition when sending a command to a newly created terminal tab. 

I don't know any other way of doing it. This is an attempt to try and make sure that the connection is both alive and that the backing process (wherever it might be) is ready to receive input.

jim-docker: I see. Then I think the notification should be different

src/renderer/components/dock/terminal.store.ts
jim-docker: Maybe something like:
```suggestion
        Notifications.info("If terminal shell is not ready please check your shell init files, if applicable. You may see this message if you have set a custom shell prompt", {
```

Nokel81: Sounds good, nice and system agnostic.

jim-docker (dismissed):

src/renderer/components/dock/terminal.store.ts
```diff
@@ -117,15 +117,18 @@ export class TerminalStore extends Singleton {
...
-      const rcIsFinished = when(() => this.connections.get(tab.id).shellRunCommandsFinished);
+      const shellIsReady = when(() => this.connections.get(tab.id).isReady);
       const notifyVeryLong = setTimeout(() => {
-        rcIsFinished.cancel();
-        Notifications.info("Terminal shell is taking a long time to complete startup. Please check your .rc file. Bypassing shell completion check.", {
-          timeout: 4_000,
-        });
+        shellIsReady.cancel();
+        Notifications.info(
+          "If terminal shell is not ready please check your shell init files, if applicable. You may see this message if you have set a custom shell prompt.",
```
jim-docker: custom shell prompt is no longer relevant?

Nokel81: yeah fair enough, it might still be relevant but probably not.

jim-docker (approved):

PR #3799, opened 2021-09-14
Upgrade all testing linux CI machines to Ubuntu-20.04

### LINKED ISSUE #3025: Upgrade to a newer version of Ubuntu
As per https://github.com/actions/virtual-environments/issues/3287

We should upgrade to a newer version of ubuntu.

Nokel81:

### DIFF
--- a/.azure-pipelines-k8s-matrix.yml
+++ b/.azure-pipelines-k8s-matrix.yml
@@ -15,7 +15,7 @@ trigger: none
 jobs:
   - job: Linux
     pool:
-      vmImage: ubuntu-16.04
+      vmImage: ubuntu-20.04
     strategy:
       matrix:
         kube_1.16:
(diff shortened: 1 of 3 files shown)


.azure-pipelines-k8s-matrix.yml
```diff
@@ -46,7 +46,7 @@ jobs:
           curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64
           sudo install minikube-linux-amd64 /usr/local/bin/minikube
           sudo minikube start --driver=none --kubernetes-version $(kubernetes_version)
-          # Although the kube and minikube config files are in placed $HOME they are owned by root
+          sudo mv /root/.kube /root/.minikube $HOME
```
Nokel81, line 49: My only concern is that this won't work as expected. I search around but could not find the equivalent to the github-action for azure.

jim-docker: The comment said they are already in `$HOME`. Are they no longer put there, after this minikube install/run is executed on ubuntu 20.04?

Nokel81: According to this run I don't think so...

https://github.com/lensapp/lens/runs/3602487897?check_suite_focus=true

jim-docker: OK, if it works...

jim-docker (approved):

Nokel81:
@jakolehm Is there a way to run that azure pipeline?

jakolehm (approved):

PR #3800, opened 2021-09-14
Draggable region on macOS should be the same as the main header

Nokel81:

This way there is a visual queue to where the user can drag the window around from.


jweak (approved):
Looks good!

PR #3807, opened 2021-09-15
Redirect from /catalog to last catalog page

### LINKED ISSUE #3642: Catalog should always remember its last state
As a User returning to the catalog I would like to always see the Catalog in the state that I last time left it after navigating away  (e.g opening a cluster, add cluster, ...?)

**Details:**
- this already works pretty fine when closing a cluster using X button on top right corner
- It does not yet properly remember the catalog state when opening a cluster and going back to catalog using the catalog hotbar item
- It does not remember the state when adding a new cluster (+) and closing / submitting the form

aleksfront:
Setting redirect from `/catalog` url to `/catalog` + last page from storage (or `browse`).

https://user-images.githubusercontent.com/9607060/133383419-73a7f472-781e-4f2a-8840-68dd63817031.mov

Fixes #3642

### DIFF
--- a/src/common/routes/catalog.ts
+++ b/src/common/routes/catalog.ts
@@ -31,3 +31,5 @@ export const catalogRoute: RouteProps = {
 };
 
 export const catalogURL = buildURL<CatalogViewRouteParam>(catalogRoute.path);
+
+export const browseCatalogTab = "browse";
--- a/src/common/routes/catalog.ts
+++ b/src/common/routes/catalog.ts
@@ -31,7 +31,14 @@ export const catalogRoute: RouteProps = {
 };
 
 export const getPreviousTabUrl = (path: string) => {
-  return `${catalogURL()}/${path || browseCatalogTab}`;
+  const [group, kind] = path.split("/");
+
+  return catalogURL({
+    params: {
+      group: group || browseCatalogTab,
+      kind
+    }
+  });
 };
 
 export const catalogURL = buildURL<CatalogViewRouteParam>(catalogRoute.path);
--- a/src/common/routes/catalog.ts
+++ b/src/common/routes/catalog.ts
@@ -30,6 +30,10 @@ export const catalogRoute: RouteProps = {
   path: "/catalog/:group?/:kind?"
 };
 
+export const getPreviousTabUrl = (path: string) => {
+  return `${catalogURL()}/${path || browseCatalogTab}`;
+};
+
 export const catalogURL = buildURL<CatalogViewRouteParam>(catalogRoute.path);
 
 export const browseCatalogTab = "browse";
--- a/src/renderer/components/+catalog/catalog.tsx
+++ b/src/renderer/components/+catalog/catalog.tsx
@@ -44,7 +44,8 @@ import { CatalogMenu } from "./catalog-menu";
 import { HotbarIcon } from "../hotbar/hotbar-icon";
 import { RenderDelay } from "../render-delay/render-delay";
 
-export const previousActiveTab = createAppStorage("catalog-previous-active-tab", "");
+export const browseCatalogPage = "browse";
+export const previousActiveTab = createAppStorage("catalog-previous-active-tab", browseCatalogPage);
 
 enum sortBy {
   name = "name",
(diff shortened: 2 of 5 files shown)


src/renderer/components/+catalog/catalog.tsx
```diff
@@ -44,7 +44,8 @@ import { CatalogMenu } from "./catalog-menu";
 import { HotbarIcon } from "../hotbar/hotbar-icon";
 import { RenderDelay } from "../render-delay/render-delay";
 
-export const previousActiveTab = createAppStorage("catalog-previous-active-tab", "");
+export const browseCatalogPage = "browse";
```
ixrock: Why this constant is here? Should we keep all known groups for catalog in `routes/catalog.ts`?

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -153,7 +154,7 @@ export class Catalog extends React.Component<Props> {
     if (activeCategory) {
       navigate(catalogURL({ params: {group: activeCategory.spec.group, kind: activeCategory.spec.names.kind }}));
     } else {
-      navigate(catalogURL());
+      navigate(`${catalogURL()}/browse`);
```
ixrock: This url-builder supports parameters: `catalogURL({group: browseCatalogPage})`

src/renderer/components/cluster-manager/cluster-manager.tsx
```diff
@@ -56,6 +57,7 @@ export class ClusterManager extends React.Component {
         <main>
           <div id="lens-views"/>
           <Switch>
+            <Redirect exact from={catalogURL()} to={`${catalogURL()}/${previousActiveTab.get() || browseCatalogPage}`}/>
```
ixrock: `catalogURL({group: previousActiveTab.get()})` + maybe add `browseCatalogPage` as default value for `previousActiveTab`?

ixrock (changes requested):

src/renderer/components/+catalog/catalog.tsx
aleksfront: Moving this to `/common/routes/catalog.ts`. All of the other groups are dynamic and generated on demand from `catalogCategoryRegistry`.

src/renderer/components/+catalog/catalog.tsx
aleksfront: Yes, it supports parameters but the form is more verbose, I'm not sure if it's fit better:
```
catalogURL({ params: { group: browseCatalogPage }})
```

src/renderer/components/cluster-manager/cluster-manager.tsx
aleksfront: It is a default value, but adding extra check here for users already having empty string ("") for `previousActiveTab` in their local storage. This check overrides such case.

src/renderer/components/+catalog/catalog.tsx
aleksfront: Yes, using your suggested form for sake of consistency.

src/common/routes/catalog.ts
```diff
@@ -30,4 +30,10 @@ export const catalogRoute: RouteProps = {
   path: "/catalog/:group?/:kind?"
 };
 
+export const getPreviousTabUrl = (path: string) => {
+  return `${catalogURL()}/${path || browseCatalogTab}`;
+};
```
aleksfront: @ixrock Left url constructor untouched for the sake of backward compatibility. Users already has this form saved in their local storage. And `to` prop inside `<Redirect/>` can take `string` safely.

ixrock: Let's say we will change this route path in future to just:

```ts
export const catalogRoute: RouteProps = {
  path: "/catalog"
};
```
And will serve params like `/catalog?group=someGroup&kind=kind`, wouldn't it break backward compatibility with that users?

In other words I mean it's better and safer to parse already that params and serve separately in used URLs.
But `Redirect` can and should accept strings of course :)

ixrock (approved):

PR #3810, opened 2021-09-15
Pin icon for Catalog list items

### LINKED ISSUE #3754: Hot Bar Pin Item next to Cluster Name in catalog view
Show the Pin icon next to the cluster name (and catalog) to allow users to pin it to the hotbar easily!

aleksfront:
To make adding to the Hotbar an intuitive action, we add a pin icon next to the name of each object in the catalog table.

https://user-images.githubusercontent.com/9607060/133427001-73f09c9e-12b8-420f-a147-9544653faaff.mov

Resolves #3754


src/common/hotbar-store.ts
```diff
@@ -274,6 +274,14 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
 
     hotbarStore.activeHotbarId = hotbarStore.hotbars[index].id;
   }
+
+  /**
+   * Checks if entity already pinned to hotbar
+   * @returns boolean
+   */
+  entityPinnedToHotbar(entity: CatalogEntity) {
```
ixrock: active == pinned, right? Probably better to use single terminology.

src/renderer/components/+catalog/catalog.module.css
```diff
@@ -19,10 +19,40 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
+.list :global(.TableRow) {
```
ixrock, line 22: I see lots of errors due nested selectors and `&` in `.css` file in my editor (WebStorm):

Screenshot 2021-09-22 at 22 54 42

Nesting is supported by css-modules? This disappears if I rename extension to `.scss` (I would prefer actually to use everywhere `*.scss`)

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -186,6 +187,26 @@ export class Catalog extends React.Component<Props> {
...
+        <Icon
+          small
+          disabled={disabledIcon}
+          className={styles.pinIcon}
+          material="push_pin"
+          tooltip={disabledIcon ? "" : "Pin to Hotbar"}
+          onClick={(event) => {
+            event.stopPropagation();
+            this.addToHotbar(item);
+          }}/>
```
ixrock: _nit_: we could use also `src/renderer/utils/prevDefault.ts` to prevent event propagation:

```suggestion
          onClick={prevDefault(() => this.addToHotbar(item))}
```

ixrock: Also I would better to allow these icons to **unpin** instead of no-action when pinned already:

Screenshot 2021-09-22 at 23 09 14

src/common/hotbar-store.ts
Nokel81: No active (in the sense of `this.getActive()`) means the currently visible hotbar. 

Though I would probably change this to be `isPinnedToActive(entity: CatalogEntity)`

aleksfront:
* Removed any mention about "Pin/Unpin". Replaced with "Add/Remove" words since we're not adding anything to hotbar automatically and `pin` phrases doesn't make sense anymore.
* Pin icon now toggles item to hotbar, not just adding it.

https://user-images.githubusercontent.com/9607060/136749380-a651da8f-c962-4e98-a273-f48dea75f1ab.mov

src/renderer/components/+catalog/catalog.module.css
aleksfront: I can't compile an app if just change `*.css` to `*.scss`. These tweaks pointed to work CSS Modules with SCSS needs additional investigation. Will take a look on this.

src/renderer/components/+catalog/catalog-entity-drawer-menu.tsx
```diff
@@ -125,3 +117,19 @@ export class CatalogEntityDrawerMenu<T extends CatalogEntity> extends React.Comp
     );
   }
 }
+
+function HotbarToggleItem(props: { entity: CatalogEntity }) {
+  const store = HotbarStore.getInstance(false);
+  const add = () => store?.addToHotbar(props.entity);
```
ixrock: Why `store` might not exist?

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -194,13 +201,36 @@ export class Catalog extends React.Component<Props> {
...
-        </MenuItem>
+        {!isItemInHotbar ? (
+          <MenuItem key="add-to-hotbar" onClick={() => this.addToHotbar(item)}>
+            Add to Hotbar
+          </MenuItem>
+        ) : (
+          <MenuItem key="remove-from-hotbar" onClick={() => this.removeFromHotbar(item)}>
+            Remove from Hotbar
+          </MenuItem>
+        )}
```
ixrock: Can we reuse newly created `<HotbarToggleItem/>` to render proper menu item?

ixrock (dismissed):

src/renderer/components/+catalog/catalog-entity-drawer-menu.tsx
```diff
@@ -125,3 +117,19 @@ export class CatalogEntityDrawerMenu<T extends CatalogEntity> extends React.Comp
...
+  const remove = () => store?.removeFromHotbar(props.entity.getId());
+  const [itemInHotbar, setItemInHotbar] = useState(store?.isAddedToActive(props.entity));
+
+  return (
+    <MenuItem onClick={() => {
+      itemInHotbar ? remove() : add();
+      setItemInHotbar(!itemInHotbar);
+    }}>
+      <Icon material="push_pin" small tooltip={itemInHotbar ? "Remove from Hotbar" : "Add to Hotbar"}/>
```
Nokel81: Shouldn't the icon change if the action is `remove`?

src/renderer/components/+catalog/catalog-entity-drawer-menu.tsx
aleksfront: It is made for tests to pass (in specific `catalog.test.tsx`). For some unknown reason instances won't get created by the time test code execution. So this is a workaround.

aleksfront:
@Nokel81 Resolved requested comments and added custom "unpin" icon
pin unpin

src/renderer/components/+catalog/catalog-entity-drawer-menu.tsx
```diff
@@ -99,17 +95,20 @@ export class CatalogEntityDrawerMenu<T extends CatalogEntity> extends React.Comp
     }
 
     items.push(
-      <MenuItem key="add-to-hotbar" onClick={() => this.addToHotbar(entity) }>
-        <Icon material="playlist_add" small tooltip="Add to Hotbar" />
-      </MenuItem>
+      <HotbarToggleMenuItem
+        key="hotbar-toggle"
+        entity={entity}
+        addContent={<Icon material="push_pin" small tooltip={"Add to Hotbar"}/>}
```
Nokel81, line 101: nit you don't need the `{...}`, just quotes will do.

src/renderer/components/+catalog/hotbar-toggle-menu-item.tsx
```diff
@@ -0,0 +1,42 @@
...
+ */
+import React, { ReactNode, useState } from "react";
+
+import { HotbarStore } from "../../../common/hotbar-store";
+import { MenuItem } from "../menu";
+
+import type { CatalogEntity } from "../../api/catalog-entity";
+
+export function HotbarToggleMenuItem(props: { entity: CatalogEntity, addContent: ReactNode, removeContent: ReactNode }) {
+  const store = HotbarStore.getInstance(false);
```
Nokel81, line 29: I agree with Roman, why might this be not created here?

Nokel81 (dismissed):

src/renderer/components/+catalog/hotbar-toggle-menu-item.tsx
aleksfront: https://github.com/lensapp/lens/pull/3810#discussion_r729513504

Btw do you have an idea why instances can be empty in tests?

Nokel81: Yes I do actually, you need to call `HotbarStore.createInstance()` in the `beforeEach` and then `HotbarStore.resetInstance()` in the `afterEach()

src/renderer/components/+catalog/catalog-entity-drawer-menu.tsx
ixrock: So, you should probably figure out the reason, instead of workaround it just to pass the test?

src/renderer/components/+catalog/hotbar-toggle-menu-item.tsx
Nokel81: no need for the false above as well

Nokel81 (approved):
Last two minor nits will be fixed in a follow up PR, LGTM

PR #3820, opened 2021-09-16
Electron 13.4.0

jakolehm:


aleksfront (dismissed):
LGTM, works without issues.

jweak (commented):
Checked with breaking changes log from Electron. Everything else checks out but I found that we use a deprecated event on window manager.

https://www.electronjs.org/docs/breaking-changes#deprecated-webcontents-new-window-event

https://github.com/lensapp/lens/blob/d448017f334a39b0ee9d974d867e236b4e0104e2/src/main/window-manager.ts#L110-L112

jakolehm:
> Checked with breaking changes log from Electron. Everything else checks out but I found that we use a deprecated event on window manager.
> 
> https://www.electronjs.org/docs/breaking-changes#deprecated-webcontents-new-window-event
> 
> https://github.com/lensapp/lens/blob/d448017f334a39b0ee9d974d867e236b4e0104e2/src/main/window-manager.ts#L110-L112

Good catch!

jakolehm:
@aleksfront @jweak ptal again.

jweak (approved):
LGTM

PR #3822, opened 2021-09-16
Fix test GH action to run on all branches

Nokel81:

reference: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet


jim-docker (approved):
worked on release/v5.2 branch

PR #3842, opened 2021-09-20
Emit contextMenuOpen for WebLink CatalogEntity

panuhorsmalahti:
This makes it possible for e.g. extension to add items to the Weblink CatalogEntity context menu. Currently this is possible for clusters but not weblinks.


Nokel81 (approved):
I think that this should be done automatically but this is good for the time being.

PR #3849, opened 2021-09-21
Closing dock tabs with cmd+w on mac

### LINKED ISSUE #3673: Option to unbind/rebind the cmd+w 'close window' keyboard shortcut
An Option to reconfigure the cmd+w keybinding that closes the Lens window. Ideally I could configure this key combination to close open tabs containing logs or terminals.

**Why is this needed**:
I regularly use the cmd+w keybinding to close tabs in other applications, e.g. Browser, Iterm2, VSCode, IntelliJ. Due to muscle memory, I usually try to use the same shortcut to close open logs or terminals in Lens, but it closes the whole window instead. 
This behaviour is annoying, since it takes a couple of seconds to restart and reconnect to the Kubernetes clusters. Additionally it also does not close the open log or terminal tabs.
Since I have not found a way to open multiple lens windows yet, closing the window with cmd+w or the whole app with cmd+q leads to basically the same state, especially since Lens5 removed the possibility to access a cluster directly through the menu bar icon.

- Kubernetes distribution:  GKE
- Desktop OS: OSX 11.3.1

Slightly related Issue on vscode:
https://github.com/microsoft/vscode/issues/54492

aleksfront:
* Using `cmd+w` shortcut key to close dock tabs
* Changing `Close Window` menu command accelerator from `cmd+w` to `shift+cmd+w` (usual shortcut among apps)

https://user-images.githubusercontent.com/9607060/134120014-52c5bc73-6c6b-43b8-be97-370839dd7d07.mov

Fixes #3673


src/main/menu.ts
```diff
@@ -135,13 +135,18 @@ export function buildMenu(windowManager: WindowManager) {
...
         }
       ]),
+
       { type: "separator" },
-      {
-        role: "close",
-        label: "Close Window"
-      },
+
+      ...(isMac ? [
```
aleksfront, line 153: We don't need "Close window" command on win. Otherwise, it will just duplicate the `Exit` one.

src/renderer/components/dock/dock.tsx
```diff
@@ -46,19 +46,31 @@ interface Props {
...
 
-    if (!selectedTab) return;
-    const { code, ctrlKey, shiftKey } = evt.nativeEvent;
+  componentDidMount() {
+    document.addEventListener("keydown", this.onKeyDown);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener("keydown", this.onKeyDown);
+  }
```
aleksfront, line 57: Listening `keydown` globally allowing to close dock tabs in quick sequence, by repeating pressing `cmd+w`. Before the change it was impossible since the focus was jumped after closing each tab, so the event was no longer listened to.

src/renderer/components/dock/terminal.ts
```diff
@@ -225,11 +225,6 @@ export class Terminal {
       }
     }
 
-    // Pass the event above in DOM for <Dock/> to handle common actions
-    if (!evt.defaultPrevented) {
-      this.elem.dispatchEvent(new KeyboardEvent(type, evt));
-    }
-
```
aleksfront, line 255: Removing this because we moved `keydown` listener to the global scope so `<Dock/>` won't listen any events anymore.

src/renderer/components/dock/dock.tsx
```diff
@@ -46,19 +46,31 @@ interface Props {
...
+
+  onKeyDown = (evt: KeyboardEvent) => {
+    const { close, selectedTab, closeTab } = dockStore;
+    const { code, ctrlKey, metaKey, shiftKey } = evt;
+    // Determine if user working inside <Dock/> or using any other areas in app
+    const dockIsFocused = this.element?.current.contains(document.activeElement);
+
+    if (!selectedTab || !dockIsFocused) return;
```
Nokel81, line 65: I think that if a dock is not focused then we should close the window so as to preserve the functionality.

aleksfront: I found that `shift+cmd+w` is also very common keyboard shortcut among the apps. My guess it is used in the applications where some kind of tabs are presented. I would keep `shift+cmd+w` for closing window since it's not required to write extra code preventing triggering global shortcuts. And there might be situation when document focused element is unclear and attempt to close tab will lead to closing whole window as described in #3673 .

jweak: Should we somehow indicate where the focus is? Maybe the color under the tab could be gray when not in focus and blue when it is? Or maybe the whole dock could have an outline when there's focus?

src/renderer/components/dock/dock.tsx
```diff
@@ -46,19 +46,31 @@ interface Props {
...
 
     if (shiftKey && code === "Escape") {
       close();
     }
 
-    if (ctrlKey && code === "KeyW") {
-      if (selectedTab.pinned) close();
```
ixrock, line 60: Why this logic removed? If tab is pinned we're still closing it?

src/renderer/components/dock/dock.tsx
ixrock: Maybe this for consistency?

_Windows_: `Shift+Ctrl+W` => exit app (suggested), `Ctrl+W` closing active tab (and currently whole window if tab is not focused, right?)
_Mac_: `Shift+Cmd+W` => exit app (like now), `Cmd+W` closing active tab only

ixrock (commented):
LGTM

src/renderer/components/dock/dock.tsx
Nokel81: The problem is that we aren't exiting the app, just closing the window.

src/renderer/components/dock/dock.tsx
aleksfront: We're not using `pinned` feature anywhere in the code, so I suggest to clean it up. Also, `closeTab()` method from dock store has this check
```
if (!tab || tab.pinned) {
  return;
}
```

aleksfront:
Made some fixes and added focused state as @jweak suggested. Active tab line now goes gray if dock isn't focused.

dock focus state

@ixrock @jweak Ptal.

jweak (approved):
LGTM

PR #3858, opened 2021-09-21
Fix kube-api watch leak on window unload

jakolehm:
`node-fetch` connection is not always automatically cleaned when window reloads. This PR aims to fix that by unmounting app when window unloads.


ixrock (approved):
Not sure if this fix is really works and how `node-fetch` related to `ReactDOM.unmountComponentAtNode` 😂 but let's try it out 🤞
Code LGTM.

PR #3866, opened 2021-09-22
Fix menu rendering on when slightly off screen

### LINKED ISSUE #3383: Dancing context menu
It can't decide if it should go up or go down:

https://user-images.githubusercontent.com/40840436/125684709-aecfebb4-fe39-49e7-9d2d-2ed3c7bb04c3.mov

happens in 5.1.0-beta.1 on mac

jweak:
Fixes: #3383 

Menu would bounce around in some cases because the
refreshPosition() would calculate different results
based on different state of the component and previous
render cycle CSS classes.

Also I found that the render function calls causes
in an infinite render loop while open because of refreshPosition
will cause a state change which will rerender the function.
Fixed that with the shouldComponentUpdate which checks
if state has actually changed in between renders.


src/renderer/components/menu/menu.tsx
```diff
@@ -85,6 +86,10 @@ export class Menu extends React.Component<MenuProps, State> {
     return !!this.props.isOpen;
   }
 
+  shouldComponentUpdate(nextProps: MenuProps, nextState: State) {
```
jim-docker: Do we really need this? The real problem is invoking `refreshPosition()` from `render()`, right? Why does that cause an infinite render loop? if it doesn't change state or props then `shouldComponentUpdate()` shouldn't be needed(?)

jim-docker (commented):
I tested it, couldn't find any issues

src/renderer/components/menu/menu.tsx
jweak: ```setState``` will always cause a rerender even if the state is the same (without this method), but the virtual DOM won't do any operations if the result of the render does not change. 

The underlying reason is calling ```refreshPosition``` from the ```render```, and that was added apparently because sometimes the content will change asynchronously. The whole problem stems from manually moving the DOM element in the ```refreshPosition``` so the actual position state is in the DOM but not in the component state. So in order to sync the state in the DOM we are calling the ```refreshPosition``` way too much.

Actually now I realised the better solution is to calculate the position on the render instead of calling setState and remove the ```refreshPosition``` method call. I'll try this approach. Thanks!

ixrock: I'd rather to use `mobx.observable` to handle state changes.
Let try to avoid old-funky `setState` as possible? This way we don't need `shouldComponentUpdate` at all.

jim-docker: @ixrock, glad you're back!

jweak: Removed `shouldComponentUpdate` and put the style in the state so they'll be in sync.

jim-docker (approved):
LGTM, fixes the problem. Did you consider the mobx suggestion from @ixrock? We are using mobx instead of React state for most of our React components now. Could be a good followup PR...

jweak:
> LGTM, fixes the problem. Did you consider the mobx suggestion from @ixrock? We are using mobx instead of React state for most of our React components now. Could be a good followup PR...

Thanks. Yeah I considered, but I think the whole component could use a bit of refactoring and maybe then it'd be good to reconsider observables to manage the position state 👍

PR #3869, opened 2021-09-22
Use browser fetch on renderer

jakolehm:
For performance reasons.

### DIFF
--- a/src/common/k8s-api/json-api.ts
+++ b/src/common/k8s-api/json-api.ts
@@ -21,7 +21,6 @@
 
 // Base http-service / json-api class
 
-import { randomBytes } from "crypto";
 import { merge } from "lodash";
 import nodeFetch, { Response as NodeResponse, RequestInit as NodeRequestInit } from "node-fetch";
 import { stringify } from "querystring";
@@ -82,8 +81,17 @@ export class JsonApi<D = JsonApiData, P extends JsonApiParams = JsonApiParams> {
     return this.request<T>(path, params, { ...reqInit, method: "get" });
   }
 
+  protected getRequestUrl(path: string, useSubdomain = false) {
+    if (window) {
+
+      return `${this.config.apiBase}${path}`;
+    } else {
+      return `${this.config.serverAddress}${this.config.apiBase}${path}`;
+    }
+  }
+
   getResponse(path: string, params?: P, init: JsonRequestInit = {}): Promise<JsonResponse> {
-    let reqUrl = this.buildRequestUrl(path, true);
+    let reqUrl = this.getRequestUrl(path, true);
     const reqInit = merge({}, this.reqInit, init);
     const { query } = params || {} as P;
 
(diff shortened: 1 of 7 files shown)


src/common/k8s-api/json-api.ts
```diff
@@ -65,22 +69,33 @@ export class JsonApi<D = JsonApiData, P extends JsonApiParams = JsonApiParams> {
...
-  get<T = D>(path: string, params?: P, reqInit: RequestInit = {}) {
+  get<T = D>(path: string, params?: P, reqInit: RequestInit | NodeRequestInit = {}) {
     return this.request<T>(path, params, { ...reqInit, method: "get" });
   }
 
-  getResponse(path: string, params?: P, init: RequestInit = {}): Promise<Response> {
-    let reqUrl = `${this.config.serverAddress}${this.config.apiBase}${path}`;
-    const reqInit: RequestInit = merge({}, this.reqInit, init);
+  protected getRequestUrl(path: string, useSubdomain = false) {
+    if (window) {
```
ixrock, line 86: I guess it's not gonna work in node env. 
You cannot check global variable existence this way, use `typeof` at least:

```
% node -i
Welcome to Node.js v14.17.6.
Type ".help" for more information.
> window
Uncaught ReferenceError: window is not defined
> typeof window
'undefined'
```

src/common/k8s-api/kube-object.store.ts
```diff
@@ -348,7 +348,7 @@ export abstract class KubeObjectStore<T extends KubeObject> extends ItemStore<T>
     const { signal } = abortController;
 
     const callback = (data: IKubeWatchEvent<T>, error: any) => {
-      if (!this.isLoaded || error?.type === "aborted") return;
+      if (!this.isLoaded || error instanceof DOMException || error?.type === "aborted") return;
```
ixrock, line 351: Probably just having `error instanceof DOMException` should be enough for aborted requests.
Or `error?.type === "aborted"` could come not only from DOM's `AbortController`?

src/main/lens-proxy.ts
```diff
@@ -185,15 +185,25 @@ export class LensProxy extends Singleton {
...
   protected async handleRequest(req: http.IncomingMessage, res: http.ServerResponse) {
+    // allow to fetch apis in "clusterId.localhost:port" from "localhost:port"
+    // this should be safe because we have already validated cluster uuid
+        
+    res.setHeader("Access-Control-Allow-Origin", "*");
+    res.setHeader("Access-Control-Allow-Methods", "GET, POST, PUT, DELETE, PATCH, OPTIONS");
+    res.setHeader("Access-Control-Allow-Headers", "Content-Type, X-Cluster-Id");
+
+    if (req.method === "OPTIONS") {
+      res.end();
```
ixrock, line 196: What the point to allow `OPTIONS` from `Access-Control-Allow-Methods` to end it right after?

jakolehm: Renderer (browser) asks CORS via `OPTIONS`, we just need to return headers and nothing else.

jakolehm:
Closing this because it seems that node-fetch seems to be ok (it was logger causing performance issues).

PR #3882, opened 2021-09-23
Update immer to version 9.x

ixrock:
_Prerequisites:_

image

https://github.com/lensapp/lens/security/dependabot/package.json/immer/open


src/renderer/components/dock/__test__/log-tab.store.test.ts
```diff
@@ -41,7 +41,8 @@ podsStore.items.push(new Pod(dockerPod));
 podsStore.items.push(new Pod(deploymentPod1));
 podsStore.items.push(new Pod(deploymentPod2));
 
-describe("log tab store", () => {
+// FIXME: broken since mobx@6 migration
+describe.skip("log tab store", () => {
```
Nokel81: I don't see how this is related to updating immer.

ixrock: Tests are failing

ixrock: Otherwise I can just remove this file if you want (useless atm)

Nokel81: tests failed because of the immer upgrade?

src/renderer/utils/storageHelper.ts
```diff
@@ -151,7 +150,7 @@ export class StorageHelper<T> {
   @action
   merge(value: Partial<T> | ((draft: Draft<T>) => Partial<T> | void)) {
     const nextValue = produce(this.toJSON(), (state: Draft<T>) => {
-      const newValue = isFunction(value) ? value(state) : value;
+      const newValue = (isFunction(value) ? value(state) : value) as unknown;
```
Nokel81: why do we need to cast it to `unknown`?

ixrock: Because it could be `void` or `Draft<T>` or even `Partial<T>` _(all these values should be acceptable from a callback)._  So, don't see any reason to strict type about that..

Nokel81: why is the merge function typed `| void`? That also seems weird.

If it returns `void` then we shouldn't use it at all.

ixrock: Refactored. PTAL.

src/renderer/utils/storageHelper.ts
```diff
@@ -149,16 +148,23 @@ export class StorageHelper<T> {
...
 
   @action
-  merge(value: Partial<T> | ((draft: Draft<T>) => Partial<T> | void)) {
-    const nextValue = produce(this.toJSON(), (state: Draft<T>) => {
-      const newValue = isFunction(value) ? value(state) : value;
+  merge(value: Partial<T> | ((draft: Draft<T>) => Draft<T> | void)) {
+    const nextValue = produce<T>(this.toJSON(), (draft: Draft<T>) => {
+
+      if (typeof value == "function") {
+        const newValue = value(draft) as Draft<T>;
```
Nokel81: Still don't need this cast.

ixrock: Removed.

src/renderer/utils/storageHelper.ts
```diff
@@ -149,16 +148,23 @@ export class StorageHelper<T> {
...
+
+      if (typeof value == "function") {
+        const newValue = value(draft);
 
-      return isPlainObject(newValue)
-        ? Object.assign(state, newValue) // partial updates for returned plain objects
-        : newValue;
+        // merge returned plain objects from `value-as-callback` usage
+        // otherwise `draft` can be just modified inside a callback without returning any value (void)
+        if (newValue && !isDraft(newValue)) {
```
Nokel81, line 159: Just so we are clear, if they return a `Draft<T>` we assume that they modified the input value, right?

ixrock: Callback usually _should not_ return a draft. Usages:

```ts
helper.merge(draft => {
 draft.update = 123;
})
```

This should produce same result as `helper.merge(() => ({update: 123}))`

Nokel81: Then why is the type `(d: Draft<T>) => Draft<T> | void`? Shouldn't it be `(() => Draft<T>) | ((d: Draft<T>) => void)`?

ixrock: Nope. Draft is still provided in arguments for convenience to update-on-fly in form `(draft: Draft<T>) => Partial<T>)`.
See examples from `storageHelper.test.ts`:

```ts
    it("merge() does partial data tree updates", () => {
      storageHelper.init();
      storageHelper.merge({ message: "updated" });

      expect(storageHelper.get()).toEqual({ ...storageHelper.defaultValue, message: "updated" });
      expect(remoteStorageMock.get(storageKey)).toEqual({ ...storageHelper.defaultValue, message: "updated" });

      // `draft` modified inside, returning `void` is expected
      storageHelper.merge(draft => {
        draft.message = "updated2";
      });
      expect(storageHelper.get()).toEqual({ ...storageHelper.defaultValue, message: "updated2" });

      // returning object modifies `draft` as well
      storageHelper.merge(draft => ({
        message: draft.message.replace("2", "3")
      }));
      expect(storageHelper.get()).toEqual({ ...storageHelper.defaultValue, message: "updated3" });
    });
```

ixrock: So, the correct definition for the `callback` should be: `(draft: Draft<T>) => Partial<T> | void)`.

Nokel81: So why do we need to check if it is a `Draft<T>`?

ixrock: To avoid accidentally assigning draft to itself (probably `immer` even might throw error on that). 
So technically you can return `Draft<T>` as `Partial<T>` from the callback.

Nokel81: This seems to imply that you can return a draft https://immerjs.github.io/immer/return

Nokel81 (approved):
While I think that checking if the return value of the merger function is a bit much, we can relax it in the future

Nokel81:
@ixrock FYI, I encountered some errors just after the PR was merged which I didn't encounter before hand. I think that the merge logic is broken when `createStorage` deals with non-object types (such as `string`). For instance the `"catalog-previous-active-tab"` storage never seems to be saved to disk after this update.

jim-docker:
I'm also seeing lots of errors. the saved port-forwards are no longer loading.

```
Uncaught (in promise) TypeError: Cannot assign to read only property '0' of object '[object String]'
    at Function.assign (<anonymous>)
    at storageHelper.ts:163
    at produce (immer.esm.js:1)
    at StorageHelper.merge (storageHelper.ts:152)
    at executeAction (mobx.esm.js:1128)
    at StorageHelper.merge (mobx.esm.js:1112)
    at value (storageHelper.ts:83)
```

immer is in the call stack

src/renderer/utils/storageHelper.ts
```diff
@@ -150,15 +149,22 @@ export class StorageHelper<T> {
...
+          Object.assign(draft, newValue);
+        }
+      } else {
+        Object.assign(draft, value);
```
jim-docker, line 163: Shouldn't there be some "plain Object" logic here too?

ixrock: Yes, should be `else if(isPlainObject(value)`

ixrock:
> @ixrock FYI, I encountered some errors just after the PR was merged which I didn't encounter before hand. I think that the merge logic is broken when `createStorage` deals with non-object types (such as `string`). For instance the `"catalog-previous-active-tab"` storage never seems to be saved to disk after this update.

What is broken?  If you're talking about`merge()` it's not intended to work with plain data types..

Nokel81:
@ixrock Yes I realize that, but somehow it is being called... I don't know how, but I wasn't seeing the errors before this PR

PR #3890, opened 2021-09-24
Detect log follow requests

### LINKED ISSUE #3878: Terminal: kubectl logs --follow times out
**Describe the bug**

`kubectl logs <pod> --follow`  times out with error:

```
error: unexpected EOF
```

jakolehm:
Fixes #3878


src/main/lens-proxy.ts
```diff
@@ -40,6 +40,24 @@ export interface LensProxyFunctions {
...
 }
 
+const truthyParams = ["1", "true", ""];
+const watchParam = "watch";
+const followParam = "follow";
+
+export function isLongRunningRequest(reqUrl: string) {
+  const url = new URL(reqUrl, "http://localhost");
+
+  if (url.searchParams.has(watchParam) && truthyParams.includes(url.searchParams.get(watchParam)) ) {
```
Nokel81: I would use the `getBoolean` method from `src/main/routes/utils/` (and maybe move it) unless you feel that this shouldn't prioritize false.

jakolehm: I didn't know that function.... but that function returns false if value is empty. Is that expected behaviour also here 🤔

Nokel81: No it doesn't. The function `get` return `null` if the key is not set

jakolehm: Right, refactored to use `getBoolean`.

Nokel81 (approved):

Issue #3909, opened 2021-09-29
Change Catalog default action of Entity to "open"


msa0311:
In order to make catalog more intuitive, lets change the catalog so that the default action will immediately open / connect to the Cluster / WebLink. 

Add the details panel as a link to the burger menu (three dots) as "Details".

PR #3911, opened 2021-09-29
Before hook for onRun of Catalog entities

### LINKED ISSUE #3916: Hook for onRun of Catalog Entities
In order to allow Extensions to hook into the "Run"-Event (e.g. Connect to Cluster) we need a hook in the extension API that is triggered if someone "connects" / "runs" a catalog api entity.

chenhunghan:
- Add extension api `Renderer.Catalog.catalogEntities.addOnRunHook`

Can be used like
```tsx
Renderer.Catalog.catalogEntities.addOnRunHook(catalogEntityUid, (entity) => {
   // return true => invoke onRun, false => don't invoke onRun;
   return true;
});
```

- Add `prettier` to be able to use `.toMatchInlineSnapshot()`


package.json
```diff
@@ -357,6 +357,7 @@
     "postcss": "^8.3.6",
     "postcss-loader": "4.3.0",
     "postinstall-postinstall": "^2.1.0",
+    "prettier": "^2.4.1",
```
Nokel81, line 360: Why is this added?

src/common/catalog/catalog-entity.ts
```diff
@@ -230,7 +231,21 @@ export abstract class CatalogEntity<
     return this.metadata.name;
   }
 
+  /**
+	 * Trigger when user click on the icon in details panel
```
chenhunghan: not sure why github renders like this, this looks normal (without extra indents) in my VSCode.

src/renderer/components/+catalog/catalog-entity.store.tsx
```diff
@@ -19,103 +19,13 @@
...
 import { autoBind } from "../../../common/utils";
-import { Badge } from "../badge";
-import { navigation } from "../../navigation";
-import { searchUrlParam } from "../input";
-import { makeCss } from "../../../common/utils/makeCss";
-import { KubeObject } from "../../../common/k8s-api/kube-object";
-
-const css = makeCss(styles);
-
-export class CatalogEntityItem<T extends CatalogEntity> implements ItemObject {
```
chenhunghan, line 39: Move `CatalogEntityItem` to a separate file (easier to mock in tests)

src/common/catalog/catalog-entity.ts
```diff
@@ -230,7 +231,21 @@ export abstract class CatalogEntity<
...
   public abstract onRun?(context: CatalogEntityActionContext): void | Promise<void>;
   public abstract onContextMenuOpen(context: CatalogEntityContextMenuContext): void | Promise<void>;
+
+	/**
+	 * Trigger when user click on the icon in details panel
+	 * 
+	 * @remarks
+	 * Priority > this.onRun, if presents, onRun won't be called.
+	 */
+	public abstract onClickDetailIcon?(context: CatalogEntityActionContext): void;
```
Nokel81: I disagree that this is the best API for this, especially since https://github.com/lensapp/lens/issues/3909 has now been asked and because I assume you want to do the work before any `onRun` call.

Furthermore, "detail icon" is too much of an implementation detail and locks us in one design.

src/main/catalog-sources/weblinks.ts
```diff
@@ -75,7 +75,9 @@ export function syncWeblinks() {
             uid: weblinkData.id,
             name: weblinkData.name,
             source: "local",
-            labels: {}
+            labels: {
+              triggerOnRunAfterOnClickDetailIcon: "true",
+            }
```
Nokel81: I highly disagree that we should use labels as a means for doing this sort of logic.

src/common/catalog-entities/web-link.ts
```diff
@@ -46,6 +46,12 @@ export class WebLink extends CatalogEntity<CatalogEntityMetadata, WebLinkStatus,
     return;
   }
 
+  onClickDetailIcon(context: CatalogEntityActionContext) {
+    catalogCategoryRegistry
+      .getCategoryForEntity<WebLinkCategory>(this)
+      ?.emit("onClickDetailIcon", this, context);
+  }
```
Nokel81: From talking to @msa0311 I was under the impression that we were wanting to block the "onRun" if any of these pre-hooks calls the equivalent of `preventDefault` which this form cannot support because the handlers are run concurrently.

Nokel81 (changes requested):

package.json
chenhunghan: To be able to use `matchInlineSnapshot()`

src/main/catalog-sources/weblinks.ts
chenhunghan: Yeah, just noticed labels will show up in the UI, so I guess we need to use other

src/common/catalog/catalog-entity.ts
Nokel81: Are you using tab characters?

src/main/catalog-sources/weblinks.ts
chenhunghan: Move under `.spec`

src/renderer/components/+catalog/catalog-entity-details.tsx
```diff
@@ -75,7 +75,7 @@ export class CatalogEntityDetails<T extends CatalogEntity> extends Component<Pro
                     item.onRun(catalogEntityRunContext);
                   }
 
-                  if (typeof item.entity.onClickDetailIcon === "function" && Boolean(item.entity.metadata?.labels?.triggerOnRunAfterOnClickDetailIcon)) {
+                  if (typeof item.entity.onClickDetailIcon === "function" && item.entity.spec?.triggerOnRunAfterOnClickDetailIcon) {
                     item.onRun(catalogEntityRunContext);
```
Nokel81: 1. This should probably be in the above `if` so as to not check twice.
2. I don't think this does what you want it to do because `onClickDetailIcon` is very much an asyncronous function which will return immediately. I think it would be better to have a "beforeOnRun" hook that gets passed a type that can be `preventDefault`-ed.

chenhunghan: 1. It's a bit hard to explain why order like this, maybe it's easier to look at the test case "prioritize onClickDetailIcon over onRun, but always trigger onRun if triggerOnRunAfterOnClickDetailIcon='true'" and "prioritize onClickDetailIcon over onRun"

"prioritize onClickDetailIcon over onRun" is a test case for
```tsx
if (typeof item.entity.onClickDetailIcon === "function") {
    item.onClickDetailIcon(catalogEntityRunContext);
  }	else {
    item.onRun(catalogEntityRunContext);
  }
```
and
"prioritize onClickDetailIcon over onRun, but always trigger onRun if triggerOnRunAfterOnClickDetailIcon='true'" is for this part
```tsx
if (typeof item.entity.onClickDetailIcon === "function" && item.entity.spec?.triggerOnRunAfterOnClickDetailIcon) {
    item.onRun(catalogEntityRunContext);
  }
```

2. `onClickDetailIcon` is a sync function (see typings) that doesn't return. What you describe: "beforeOnRun" + preventDefaulted equals to having `onClickDetailIcon` and set `triggerOnRunAfterOnClickDetailIcon=false`

Nokel81: It is a sync function yes, but you are doing an "emit" which, while sync. Returns immediately and not after all the listeners return.

chenhunghan: Not sure I understand, can you elaborate, like which functions return?

jim-docker:
I don't understand the need for this but instead of adding `onClickDetailIcon` can't you add something to the run context that lets `onRun` determine if the icon was clicked? Breaking change?

package.json
Nokel81: It is a peer dependency then?

src/renderer/components/+catalog/catalog-entity-details.tsx
Nokel81: Take https://github.com/lensapp/lens/pull/3911/files/aa70ca4154f945c0ee369167e4826248108856b7#diff-5099912721d6845c3d29e2acaa3fd4a50285eec049fd7aa76e8f025e363b4b2bR49-R53 for instance, any method that has been registered with the Weblinks category will be added to the JS event loop and then the `emit` call will return immediately but because of how the event loop works the return will happen and then continue executing before any of those listeners (or handlers) are executed.

Nokel81:
> I don't understand the need for this but instead of adding `onClickDetailIcon` can't you add something to the run context that lets `onRun` determine if the icon was clicked? Breaking change?

That would also be a possible solution. However, according to @msa0311 this is desirable to happen on attempts for `onRun`. So from that POV the name of even this proposed function is wrong.

package.json
chenhunghan: We need prettier in dependencies because we are using jest v26. v27 doesn't need that. See https://github.com/facebook/jest/pull/11566
And https://github.com/facebook/jest/issues/8467

src/renderer/components/+catalog/catalog-entity-details.tsx
panuhorsmalahti: That's not true, at least for Node.js events and the "events" package for npm. (Not sure which eventemitter we're using here).

```
var EventEmitter = require('events')
 
var ee = new EventEmitter()
ee.on('message', function () {
  console.log("2");
})
console.log("1");
ee.emit('message', 'hello world')
console.log("3");
```

This prints 1, 2 and then 3.

Node.js docs: "When the EventEmitter object emits an event, all of the functions attached to that specific event are called synchronously. Any values returned by the called listeners are ignored and discarded."

src/common/catalog/catalog-entity.ts
chenhunghan: Thanks, fixed.

chenhunghan:
> I don't understand the need for this but instead of adding `onClickDetailIcon` can't you add something to the run context that lets `onRun` determine if the icon was clicked? Breaking change?

Because we need to add the hooks to onRun via extension, from extension there is no way of changing `onRun`...

src/renderer/components/+catalog/catalog-entity-details.tsx
Nokel81: okay fair enough. However async function return immediately anyway so if any they aren't awaited.

src/common/catalog-entities/general.ts
```diff
@@ -39,6 +39,10 @@ export class GeneralEntity extends CatalogEntity<CatalogEntityMetadata, CatalogE
     navigate(this.spec.path);
   }
 
+  onClickDetailIcon() {
```
Nokel81: Still don't like this name.

src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -52,6 +52,7 @@ export interface KubernetesClusterSpec extends CatalogEntitySpec {
     background?: string;
   };
   accessibleNamespaces?: string[];
+	triggerOnRunAfterOnClickDetailIcon?: boolean
```
Nokel81: I think this should just happen

src/common/catalog-entities/general.ts
chenhunghan: Just rebased to master, no new commits yet :)

src/renderer/components/+catalog/catalog-entity-details.tsx
chenhunghan: Not sure about that too, but I guess it's out of scope of this PR..

src/renderer/components/+catalog/catalog-entity.store.tsx
chenhunghan: Do it in https://github.com/lensapp/lens/pull/3927

src/common/catalog-entities/kubernetes-cluster.ts
chenhunghan: Removed

src/extensions/renderer-api/catalog.ts
```diff
@@ -48,6 +48,33 @@ export class CatalogEntityRegistry {
...
+
+  /**
+   * Returns all catalog entities' onRun hooks
+   */
+  getAllOnRunHooks(): Array<{
+    catalogEntityUid: CatalogEntity["metadata"]["uid"];
+    onRunHook: CatelogEntityOnRunHook;
+  }> {
+    return registry.getAllOnRunHooks();
+  }
```
Nokel81: I don't think we need this API.

src/extensions/renderer-api/catalog.ts
```diff
@@ -48,6 +48,33 @@ export class CatalogEntityRegistry {
...
+    return registry.addOnRunHook(catalogEntityUid, onRunHook);
+  }
+
+  /**
+   * Returns one catalog entity onRun hook by catalog entity uid
+   */
+  getOnRunHook(catalogEntityUid: CatalogEntity["metadata"]["uid"]): CatelogEntityOnRunHook | undefined {
+    return registry.getOnRunHook(catalogEntityUid);
+  }
+
```
Nokel81: We should support several onRunHooks. You sort of do already but not very well since you only ever return the first hook.

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -27,15 +27,25 @@ import type { Cluster } from "../../main/cluster";
 import { ClusterStore } from "../../common/cluster-store";
 import { Disposer, iter } from "../utils";
 import { once } from "lodash";
+import type { CatalogEntityItem } from "../../renderer/components/+catalog/catalog-entity-item";
 
 export type EntityFilter = (entity: CatalogEntity) => any;
+export type CatelogEntityOnRunHook = (entity: CatalogEntity | CatalogEntityItem<CatalogEntity>) => boolean | Promise<boolean>;
```
Nokel81: spelling mistake: should be catalog...

Nokel81: Also I think that leaking `CatalogEnityItem` to the extension API is a very bad idea.

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -27,15 +27,25 @@ import type { Cluster } from "../../main/cluster";
...
+
+type CatalogEntityUid = CatalogEntity["metadata"]["uid"];
```
Nokel81, line 36: I would just use `string` directly.

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -27,15 +27,25 @@ import type { Cluster } from "../../main/cluster";
...
   protected _entities = observable.map<string, CatalogEntity>([], { deep: true });
   protected filters = observable.set<EntityFilter>([], {
     deep: false,
   });
+  protected entityOnRunHooks = observable.set<{
+    catalogEntityUid: CatalogEntityUid;
+    onRunHook: CatelogEntityOnRunHook
+  }>([], {
+    deep: false,
+  });
```
Nokel81: This is a very strange type, why is it not `observable.map<string, CatalogEnitytOnRunHook[]>([], { deep: false });`?

src/renderer/components/+catalog/catalog-entity-details.tsx
```diff
@@ -68,8 +68,12 @@ export class CatalogEntityDetails<T extends CatalogEntity> extends Component<Pro
                 material={item.entity.spec.icon?.material}
                 background={item.entity.spec.icon?.background}
                 disabled={!item?.enabled}
-                onClick={() => item.onRun(catalogEntityRunContext)}
-                size={128} />
+                onClick={() => {
+                  this.props.onClickDetailPanelIcon(item);
```
Nokel81: I don't think we should pass this in as props, we should have an unexported function that takes an `CatalogEntity` and does the work once (DRY as it were).

src/renderer/components/+catalog/catalog-entity-item.tsx
```diff
@@ -0,0 +1,127 @@
...
+    if (!onRunHook) {
+      this.entity.onRun(ctx);
+
+      return;
+    }
+
+    if (typeof onRunHook === "function") {
+      // if onRunHook() returns a Promise, we wait for it to resolve
+      // if not, just take whatever it returns
+      Promise.resolve(onRunHook(toJS(this.entity))).then((shouldRun) => {
```
Nokel81: This can lead to `UnhandledPromiseRejectionError` which is bad design.

src/renderer/components/+catalog/catalog-entity-item.tsx
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  onRun(onRunHook: CatelogEntityOnRunHook | undefined, ctx: CatalogEntityActionContext) {
+    if (!onRunHook) {
+      this.entity.onRun(ctx);
+
+      return;
+    }
+
+    if (typeof onRunHook === "function") {
```
Nokel81: These two checks can be combined:

```js
if (typeof onRunHook !== "function") {
  ...
  return;
}

...
```

src/extensions/renderer-api/catalog.ts
```diff
@@ -48,6 +48,33 @@ export class CatalogEntityRegistry {
...
     return registry.getItemsForCategory(category);
   }
+
+  /**
+   * Add a onRun hook to a catalog entity.
+   * @param catalogEntityUid The uid of the catalog entity
+   * @param onRunHook The function that should return a boolean if the onRun of catalog entity should be triggered.
+   * @returns A function to remove that hook
+   */
+  addOnRunHook(catalogEntityUid: CatalogEntity["metadata"]["uid"], onRunHook: CatelogEntityOnRunHook) {
```
Nokel81: I think this should be instead called `onBeforeRun` so as to be more descriptive of when it will be run.

src/extensions/renderer-api/catalog.ts
chenhunghan: Can you explain why? Intention: allow extension developer to check if hooks are already added.

src/renderer/components/+catalog/catalog-entity-item.tsx
chenhunghan: Maybe, but with less readability IMHO.

src/extensions/renderer-api/catalog.ts
Nokel81: Why would it be the extension developer care? We should just use a `Set` then.

chenhunghan: At least in space extension we need this to check when the clusters are synced from the cloud.

Nokel81: They should just add the handler again (using the built-in set type logic) or you should add a method to do that check. I don't see the point of this method in particular.

chenhunghan: realised we had `getOnRunHook` already, we can remove this.

src/extensions/renderer-api/catalog.ts
chenhunghan: I think this should be another PR,

src/renderer/api/catalog-entity-registry.ts
chenhunghan: Thanks, fix the typo. About `CatalogEnityItem`, it more about reflecting the true types... but I guess we can remove it.

src/renderer/api/catalog-entity-registry.ts
chenhunghan: IMHO this is more descriptive we are using uid from `catalog.metadata.uid`

src/extensions/renderer-api/catalog.ts
chenhunghan: If we want to support multiple hooks in the future, `onBeforeRun` is less descriptive IMO.

src/renderer/api/catalog-entity-registry.ts
chenhunghan: We need `catalogEntityUid` because there are `getOnRunHook `

src/renderer/components/+catalog/catalog-entity-item.tsx
chenhunghan: `UnhandledPromiseRejectionError` is only in node.js not in React, right? Good point about the error handling though, added some more test cases https://github.com/lensapp/lens/pull/3911/commits/24f5f9950afb432c98c815387c9f41786f1ea667

src/renderer/components/+catalog/catalog-entity-details.tsx
chenhunghan: Can the refactoring be another PR?

src/renderer/components/icon/icon.tsx
```diff
@@ -114,14 +114,14 @@ export class Icon extends React.PureComponent<IconProps> {
     };
 
     // render as inline svg-icon
-    if (svg) {
+    if (typeof svg === "string") {
```
chenhunghan, line 117: This is here just for easier mocking in tests (less crashes in components)

src/extensions/lens-renderer-extension.ts
```diff
@@ -77,7 +77,7 @@ export class LensRendererExtension extends LensExtension {
   }
 
   /**
-   * Add a filtering function for the catalog catogries. This will be removed if the extension is disabled.
+   * Add a filtering function for the catalog categories. This will be removed if the extension is disabled.
```
chenhunghan, line 80: Was a typo.

src/extensions/renderer-api/catalog.ts
Nokel81: No because this will have to be supported until the next major

src/renderer/api/catalog-entity-registry.ts
Nokel81: Ok

src/renderer/api/catalog-entity-registry.ts
Nokel81: The top level map's key would be `CatalogEntityUid`

src/renderer/components/+catalog/catalog-entity-item.tsx
Nokel81: no that is part of the JS spec. If you open up devtools and type `Promise.reject(4)` into the console you should see something like this. Sure it isn't the same name but it is still bad design
Screen Shot 2021-10-05 at 8 42 11 AM

src/extensions/renderer-api/catalog.ts
Nokel81: How is it less descriptive? It better shows when we plan on running them

chenhunghan: Well, it's just my opinion, if you think "onBeforeRun" is better, I would just change it

src/renderer/components/+catalog/catalog-entity-item.tsx
chenhunghan: It has been fixed.

src/extensions/renderer-api/catalog.ts
Nokel81: I meant the argument would be `onBeforeRun` and this method would be `addOnBeforeRun`

src/renderer/api/catalog-entity-registry.ts
chenhunghan: `string` is not as obvious as `catalogEntityUid`, but again, just my opinion, maybe you know more about the convention in Lens.

src/extensions/renderer-api/catalog.ts
chenhunghan: OK, `addOnBeforeRun` makes sense.

src/renderer/api/catalog-entity-registry.ts
Nokel81: We don't really have a convention... (namely we use both styles). What I meant was I should have suggested the type `observable.map<CatalogEntityUid, IObservableSet<CatelogEntityOnRunHook>>`

src/extensions/renderer-api/catalog.ts
chenhunghan: Edited https://github.com/lensapp/lens/pull/3911/commits/74eb89c361ceff7f396f405ddbf7eb08e4da3b99

src/extensions/renderer-api/catalog.ts
chenhunghan: At least in current timeframe I don't think I can do this before 5.3. @Nokel81 would you like to to take over on the multiple hooks part?

src/renderer/api/catalog-entity-registry.ts
chenhunghan: OK, thanks for clarify.

chenhunghan: Changed https://github.com/lensapp/lens/pull/3911/commits/c8631346ca0d734c233a97d3134f1a9362b6900f

jim-docker (commented):
reviewed @Nokel81 's commits, LGTM but my head is spinning. @Nokel81 should give the final approval...

src/extensions/renderer-api/catalog.ts
```diff
@@ -50,21 +51,14 @@ export class CatalogEntityRegistry {
   }
 
   /**
-   * Add a onBeforeRun hook to a catalog entity.
+   * Add a onRun hook to a catalog entity. If `onBeforeRun` was previously added then it will not be added again
```
jim-docker: Why change to `onRun`? ("Add a onRun hook to...")

Nokel81: Good catch, I actually copied this over from the non-exposed registry. So I have fixed it there too.

Nokel81 (approved):

PR #3913, opened 2021-09-29
Add better error message if the proxy test fails

### LINKED ISSUE #3897: Initial load after installation show error dialog
**Describe the bug**
When opening lens for the first time after upgrade from 5.1.x the following error dialog shows up:
`static://splash.html` seems to be missing according to the error.
image

**To Reproduce**
1. Open lens after installation

**Expected behavior**
No error dialog

**Screenshots**
See above

Lens: 5.2.4-latest.20210923.1
Electron: 12.0.17
Chrome: 89.0.4389.128
Node: 14.16.0
- OS: [e.g. OSX] : Windows 10
- Installation method (e.g. snap or AppImage in Linux): exe file

**Additional context**
This doesn't happen on second launch, it happened to me only the first time I opened the app.
I haven't tried to reproduce it by uninstalling and installing again though, might be a single time glitch

Nokel81:

Several users have noted that after upgrading they are no longer able to use Lens. After debugging with them their problems were resolved but it didn't seem to be related entirely to the upgrade of lens. Maybe some internal APIs of electron have changed.

I have condensed these debugging tactics into a better error dialog. Furthermore, I have noticed that `app.exit()` returns which leads to more things breaking. So I have added `return` to with the calls of `app.exit()`.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -207,7 +207,7 @@ app.on("ready", async () => {
       `Failed connection test: ${error}`,
       "Check to make sure that no other versions of Lens are running",
       `Check ${hostsPath} to make sure that it is clean and that the localhost loopback is at the top and set to 127.0.0.1`,
-      "If you have HTTP_PROXY or http_proxy set in your environment, make sure that that localhost and the ipv4 loopback addr 127.0.0.1 are added to the NO_PROXY environment variable.",
+      "If you have HTTP_PROXY or http_proxy set in your environment, make sure that the localhost and the ipv4 loopback address 127.0.0.1 are added to the NO_PROXY environment variable.",
     ];
 
     dialog.showErrorBox("Lens Proxy Error", message.join("\n\n"));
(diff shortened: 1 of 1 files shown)


src/main/index.ts
```diff
@@ -193,13 +192,27 @@ app.on("ready", async () => {
...
-    app.exit();
+
+    const hostsPath = isWindows
+      ? "C:\\windows\\system32\\drivers\\etc\\hosts"
+      : "/etc/hosts";
+    const message = [
+      `Failed connection test: ${error}`,
+      "Check to make sure that no other versions of Lens are running",
+      `Check ${hostsPath} to make sure that it is clean and that the localhost loopback is at the top and set to 127.0.0.1`,
+      "If you have HTTP_PROXY or http_proxy set in your environment, make sure that that localhost and the ipv4 loopback addr 127.0.0.1 are added to the NO_PROXY environment variable.",
```
jim-docker:
```suggestion
      "If you have HTTP_PROXY or http_proxy set in your environment, make sure that the localhost and the ipv4 loopback address 127.0.0.1 are added to the NO_PROXY environment variable.",
```

jim-docker (approved):

PR #3922, opened 2021-10-01
Feature/port forward dashboard

jim-docker:
This PR implements a basic port-forward "dashboard". See video to see how to access it.

https://user-images.githubusercontent.com/40840436/135726766-a2c01c5b-ae6d-41d1-bca4-8129c77efed2.mov

- user can start port forwards from pods/services details page (as before)
- user can specify/change local port
-  port forwards are saved (across app restarts), restarted when cluster is connected
- dashboard lists saved port forwards, user can open in browser, edit, and remove

- user can't stop a port forward without removing it from saved list (no concept of active/inactive in dashboard)


src/common/routes/port-forwards.ts
```diff
@@ -0,0 +1,33 @@
...
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import type { RouteProps } from "react-router";
+import { buildURL } from "../utils/buildUrl";
+
+export const portForwardsRoute: RouteProps = {
+  path: "/port-forwards/:forwardport?"
```
Nokel81: I don't think this should end with a `?`

src/main/routes/port-forward-route.ts
```diff
@@ -104,58 +108,108 @@ class PortForward {
...
+        clusterId: f.clusterId,
+        kind: f.kind,
+        namespace: f.namespace,
+        name: f.name,
+        port: f.port,
+        forwardPort: f.forwardPort,
+      };
+
+      portForwards.push(pf);
+    });
```
Nokel81: nit: this could be a `PortForward.portForwards.map(f => ...)`

src/main/routes/port-forward-route.ts
```diff
@@ -104,58 +108,108 @@ class PortForward {
...
+    const portForward = PortForward.getPortforward({
+      clusterId: cluster.id, kind: resourceType, name: resourceName,
+      namespace, port, forwardPort,
+    });
+
+    try {
+      await portForward.stop();
+      respondJson(response, {status: true});
+    } catch (error) {
+      logger.error(error);
```
Nokel81: we should respond here so that the request doesn't hang.

src/renderer/components/+network-services/service-port-component.tsx
```diff
@@ -38,19 +41,100 @@ interface Props {
...
+      this.watch(),
+    ]);
+  }
+
+  watch() {
+    const disposers: IReactionDisposer[] = [
+      reaction(() => portForwardStore.portForwards, () => this.init()),
+    ];
+
+    return () => disposers.forEach((dispose) => dispose());
```
Nokel81: nit: this could be changed to the following:

```js
return disposer(
  reaction(() => portForwardStore.portForwards, () => this.init()),
);
```

src/renderer/components/+workloads-pods/pod-container-port.tsx
```diff
@@ -75,6 +175,8 @@ export class PodContainerPort extends React.Component<Props> {
             <Spinner />
           )}
         </span>
+        {" "}
```
Nokel81: I think this should be done via CSS instead.

src/renderer/components/+workloads-pods/pod-container-port.tsx
```diff
@@ -63,10 +147,26 @@ export class PodContainerPort extends React.Component<Props> {
...
+      } else {
+        const portForward: ForwardedPort = {
+          kind: "pod",
+          name: pod.getName(),
+          namespace: pod.getNs(),
+          port: port.containerPort.toString(),
+          forwardPort: this.forwardPort.toString()
+        };
+        
+        PortForwardDialog.open(portForward, true);
```
Nokel81: nit: you wouldn't need to import the type if you just directly created the object in the function call. Also, what does the `true` mean? Could we make it an options object instead?

src/renderer/components/+workloads-pods/pod-container-port.tsx
```diff
@@ -42,19 +45,100 @@ interface Props {
...
+      name: pod.getName(),
+      namespace: pod.getNs(),
+      port: port.containerPort.toString(),
+      forwardPort: this.forwardPort.toString()
+    };
+    let activePort = await getPortForward(portForward);
+
+    if (!activePort) {
+      activePort = 0;
+    }
```
Nokel81: nit: 
```suggestion
    const activePort = await getPortForward(portForward) ?? 0;
```

src/renderer/components/+workloads-pods/pod-container-port.tsx
```diff
@@ -42,19 +45,100 @@ interface Props {
...
+      namespace: pod.getNs(),
+      port: port.containerPort.toString(),
+      forwardPort: this.forwardPort.toString()
+    };
 
     this.waiting = true;
 
+    if (!this.forwardPort) {
+      this.forwardPort = 0;
+    }
```
Nokel81: Why is this done here? Can `forwardPort` be `undefined` ever?

src/renderer/components/+workloads-pods/pod-container-port.tsx
```diff
@@ -42,19 +45,100 @@ interface Props {
...
+      this.watch(),
+    ]);
+  }
+
+  watch() {
+    const disposers: IReactionDisposer[] = [
+      reaction(() => portForwardStore.portForwards, () => this.init()),
+    ];
+
+    return () => disposers.forEach((dispose) => dispose());
```
Nokel81: nit: this can use `disposer` too.

Nokel81: or, actually. I don't think we need this method at all, I think just calling the `reaction` above is good enough. That is pattern we use a lot already.

src/renderer/port-forward/port-forward.store.ts
```diff
@@ -0,0 +1,218 @@
...
+import { Notifications } from "../components/notifications";
+import logger from "../../common/logger";
+
+export interface ForwardedPort {
+  clusterId?: string;
+  kind: string;
+  namespace: string;
+  name: string;
+  port: string;
+  forwardPort: string;
```
Nokel81: Why are these two strings?

src/renderer/port-forward/port-forward-item.ts
```diff
@@ -0,0 +1,75 @@
...
+import type { ItemObject } from "../../common/item.store";
+import { autoBind } from "../../common/utils";
+
+export class PortForwardItem implements ItemObject {
+  clusterId: string;
+  kind: string;
+  namespace: string;
+  name: string;
+  port: string;
+  forwardPort: string;
```
Nokel81: Same here, why are these strings?

src/renderer/port-forward/port-forward.store.ts
```diff
@@ -0,0 +1,218 @@
...
+
+        port.clusterId = pf.clusterId;
+        port.kind = pf.kind;
+        port.namespace = pf.namespace;
+        port.name = pf.name;
+        port.port = pf.port;
+        port.forwardPort = pf.forwardPort;
+
+        return addPortForward(port);
+      }));
```
Nokel81: 1. Shouldn't `PortForwardItem` have a constructor to make itself from a `ForwardedPort`?
2. Why are you creating `PortForwardItem`'s at all if `addPortForard` takes only a `ForwardedPort`?
3. This could be simiplified to `await Promise.all(savedPortForwards.map(addPortForward))`

src/renderer/port-forward/port-forward.store.ts
```diff
@@ -0,0 +1,218 @@
...
+
+      this.portForwards = [];
```
Nokel81: Also, why are we throwing away the saved port forwards?

src/renderer/port-forward/port-forward.store.ts
```diff
@@ -0,0 +1,218 @@
...
+  @computed get selectedItem() {
+    return this.portForwards.find((e: ItemObject) => e.getId() === this.selectedItemId);
+  }
+
+  watch() {
+    const disposers: IReactionDisposer[] = [
+      reaction(() => this.portForwards, () => this.loadAll()),
+    ];
+
+    return () => disposers.forEach((dispose) => dispose());
```
Nokel81: nit: this could use `disposer`

src/renderer/port-forward/port-forward.store.ts
```diff
@@ -0,0 +1,218 @@
...
+        port.clusterId = pf.clusterId;
+        port.kind = pf.kind;
+        port.namespace = pf.namespace;
+        port.name = pf.name;
+        port.port = pf.port;
+        port.forwardPort = pf.forwardPort;
+        this.portForwards.push(port);
+      });
+
+      return this.portForwards;
```
Nokel81, line 73: should we really return `undefined` if `getPortForwards` returns `undefined`?

src/renderer/port-forward/port-forward.store.ts
```diff
@@ -0,0 +1,218 @@
...
+        const port = new PortForwardItem;
+
+        port.clusterId = pf.clusterId;
+        port.kind = pf.kind;
+        port.namespace = pf.namespace;
+        port.name = pf.name;
+        port.port = pf.port;
+        port.forwardPort = pf.forwardPort;
+        this.portForwards.push(port);
+      });
```
Nokel81: This should be a `for..of` or maybe a `this.portForwards.push(...portForwards.map(pf => ...))`

src/renderer/port-forward/port-forward.store.ts
```diff
@@ -0,0 +1,218 @@
...
+
+    return () => disposers.forEach((dispose) => dispose());
+  }
+
+  loadAll() {
+    return this.loadItems(async () => {
+      let portForwards = await getPortForwards();
+
+      // filter out any not for this cluster
+      portForwards = portForwards?.filter(pf => pf.clusterId == getHostedClusterId());
```
Nokel81: Shouldn't this be part of the `get6PortForwards` call itself?

src/renderer/port-forward/port-forward.store.ts
```diff
@@ -0,0 +1,218 @@
...
+
+interface PortForwardsResult {
+  portForwards: ForwardedPort[];
+}
+
+export async function addPortForward(portForward: ForwardedPort): Promise<number> {
+  let response: PortForwardResult;
+
+  try {
+    response = await apiBase.post<PortForwardResult>(`/pods/${portForward.namespace}/${portForward.kind}/${portForward.name}/port-forward/${portForward.port}/${portForward.forwardPort}`, {});
```
Nokel81: This seems error prone, wouldn't it be better to use the `buildURL` methods directly?

src/renderer/port-forward/port-forward.store.ts
```diff
@@ -0,0 +1,218 @@
...
+    
+    if (response?.port && response.port != +portForward.forwardPort) {
+      logger.warn(`specified ${portForward.forwardPort} got ${response.port}`);
+    }
+  } catch (error) {
+    logger.warn(error); // don't care, caller must check 
+  }
+  portForwardStore.reset();
```
Nokel81, line 106: Why are we resetting the store after a successful (or for that matter unsuccessful) add?

src/renderer/port-forward/port-forward.store.ts
```diff
@@ -0,0 +1,218 @@
...
+
+  return response?.port;
+}
+
+export async function getPortForward(portForward: ForwardedPort): Promise<number> {
+  let response: PortForwardResult;
+
+  try {
+    response = await apiBase.get<PortForwardResult>(`/pods/${portForward.namespace}/${portForward.kind}/${portForward.name}/port-forward/${portForward.port}/${portForward.forwardPort}`, {});
```
Nokel81: nit: no need to pass an `{}` here.

src/renderer/port-forward/port-forward.store.ts
```diff
@@ -0,0 +1,218 @@
...
+        namespace: portForward.namespace,
+        name: portForward.name,
+      });
+      Notifications.error(`Failed to open ${browseTo} in browser`);
+    }
+    );
+
+}
+
+export const portForwardStore = new PortForwardStore();
```
Nokel81, line 179: This should be a singleton and not be exported here.

src/renderer/port-forward/port-forward-dialog.tsx
```diff
@@ -0,0 +1,178 @@
...
+            />
+          </div>
+        </div>
+      </>
+    );
+  }
+
+  render() {
+    const { className, ...dialogProps } = this.props;
+    const resourceName = this.portForward ? this.portForward.name : "";
```
Nokel81: nit: `this.portForward?.name ?? ""`

src/renderer/port-forward/port-forward-dialog.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+  renderContents() {
+    return (
+      <>
+        <div className="flex gaps align-center">
+          <div className="input-container flex align-center">
+            <div className="current-port" data-testid="current-port">
+              Local port to forward from:
+            </div>
+            <Input className={"portInput"}
```
Nokel81: nit: no need for `{}`

src/renderer/port-forward/port-forward-dialog.tsx
```diff
@@ -0,0 +1,178 @@
...
+              type="number"
+              min="0"
+              max="65535"
+              value={this.desiredPort === 0 ? "" : String(this.desiredPort)}
```
Nokel81, line 132: Why does this need to be converted to a string?

src/renderer/port-forward/port-forward-dialog.tsx
```diff
@@ -0,0 +1,178 @@
...
+  };
+
+  @computed get scaleMax() {
+    const { currentPort } = this;
+    const defaultMax = 50;
+
+    return currentPort <= defaultMax
+      ? defaultMax * 2
+      : currentPort * 2;
+  }
```
Nokel81: This is unused

src/renderer/components/+network-services/service-port-component.tsx
```diff
@@ -69,6 +169,8 @@ export class ServicePortComponent extends React.Component<Props> {
             <Spinner />
           )}
         </span>
+        {" "}
```
Nokel81: same here, CSS would be better

src/renderer/port-forward/port-forward-dialog.tsx
```diff
@@ -0,0 +1,178 @@
...
+  @observable ready = false;
+  @observable currentPort = 0;
+  @observable desiredPort = 0;
+
+  constructor(props: Props) {
+    super(props);
+    makeObservable(this);
+  }
+
+  static open(portForward: ForwardedPort, openInBrowser = false) {
```
Nokel81: nit: I think it would be better to require this `openInBrowser` to be part of an object.

src/renderer/components/+network-services/service-port-component.tsx
```diff
@@ -38,19 +41,100 @@ interface Props {
...
+      namespace: service.getNs(),
+      port: port.port.toString(),
+      forwardPort: this.forwardPort.toString()
+    };
 
     this.waiting = true;
 
+    if (!this.forwardPort) {
+      this.forwardPort = 0;
+    }
```
Nokel81: can `forwardPort` be falsy here? (though tbh we set it to a falsy value anyway)

src/main/router.ts
```diff
@@ -180,7 +180,10 @@ export class Router {
     this.router.add({ method: "get", path: `${apiPrefix}/metrics/providers` }, MetricsRoute.routeMetricsProviders);
 
     // Port-forward API
-    this.router.add({ method: "post", path: `${apiPrefix}/pods/{namespace}/{resourceType}/{resourceName}/port-forward/{port}` }, PortForwardRoute.routePortForward);
+    this.router.add({ method: "post", path: `${apiPrefix}/pods/{namespace}/{resourceType}/{resourceName}/port-forward/{port}/{forwardPort}` }, PortForwardRoute.routePortForward);
+    this.router.add({ method: "get", path: `${apiPrefix}/pods/{namespace}/{resourceType}/{resourceName}/port-forward/{port}/{forwardPort}` }, PortForwardRoute.routeCurrentPortForward);
+    this.router.add({ method: "get", path: `${apiPrefix}/port-forwards` }, PortForwardRoute.routeAllPortForwards);
+    this.router.add({ method: "delete", path: `${apiPrefix}/pods/{namespace}/{resourceType}/{resourceName}/port-forward/{port}/{forwardPort}` }, PortForwardRoute.routeCurrentPortForwardStop);
```
Nokel81: Would it be better to use query params for some of these instead of path sections?

src/main/routes/port-forward-route.ts
jim-docker: good catch

src/renderer/components/+workloads-pods/pod-container-port.tsx
jim-docker: I guess that's the more modern way but I find this more readable. Also, I think the `portForward()` method is cleaner using the local instance, so the import would be needed still.

src/renderer/components/+workloads-pods/pod-container-port.tsx
jim-docker: At one time in the evolution of this code it was possible, maybe not anymore.

src/renderer/components/+workloads-pods/pod-container-port.tsx
jim-docker: not sure I understand, how does the reaction() get stopped when the object is destroyed?

Nokel81: I meant the reaction in the above call to disposeOnUnmount

src/renderer/port-forward/port-forward.store.ts
jim-docker: missed this after refactoring the hacktoring

src/renderer/port-forward/port-forward.store.ts
jim-docker: this is a reset(), to force the list to be refreshed

src/renderer/port-forward/port-forward.store.ts
jim-docker: could be, but I envision a global page showing all port-forwards for all clusters

src/renderer/port-forward/port-forward.store.ts
jim-docker: `getPortForwards()` no longer returns undefined

src/renderer/port-forward/port-forward.store.ts
jim-docker: this is how the list/details pages, etc. know to refresh. Maybe there is a more efficient way. But the port-forwards are managed in main, and more than one cluster could be trying to add/remove/etc them at the same time. Thought it is safer to just "loadAll"

src/renderer/port-forward/port-forward.store.ts
jim-docker: How so? All the other ItemStores and KubeObjectStores are done like this.

src/renderer/port-forward/port-forward-dialog.tsx
jim-docker: Get this error otherwise:
```
Type 'string | number' is not assignable to type '(string | number | readonly string[]) & string'.
```

src/main/router.ts
jim-docker: yes. With https option coming...

src/renderer/components/+network-services/service-port-component.tsx
jim-docker: left over from the hackweek project, thought it wouldn't hurt to do this, probably could come out

src/renderer/port-forward/port-forward.store.ts
jim-docker: legacy I guess. Much of the time they are just used as strings. I'll try changing them to numbers...

jim-docker:
@Nokel81 thanks for the great comments

src/renderer/port-forward/port-forward.store.ts
```diff
@@ -0,0 +1,218 @@
...
+import { waitUntilFree } from "tcp-port-used";
+import { Notifications } from "../components/notifications";
+import logger from "../../common/logger";
+
+export interface ForwardedPort {
+  clusterId?: string;
+  kind: string;
+  namespace: string;
+  name: string;
+  port: string;
```
Nokel81: I still think these should be numbers

src/renderer/port-forward/port-forward.store.ts
Nokel81: Why are we resetting within an init function?

src/renderer/port-forward/port-forward.store.ts
Nokel81: I just think it is a bit strange for this component (which is built for in-cluster use) to get all the port forwards and then throw some away.

src/renderer/port-forward/port-forward.store.ts
Nokel81: Do we have that view in this PR? I did not see it. Nevertheless I think that think should really be an actual watch with either polling or server push but that is too big for this PR. As long as we don't expose any of this to the extension API I am fine with leaving this as it is for the time being.

src/renderer/port-forward/port-forward.store.ts
Nokel81: Fair enough, something for the future.

Nokel81 (commented):
Last few things, I think this is good to merge for the time being.

src/renderer/port-forward/port-forward.store.ts
jim-docker: Did I miss some?

Nokel81: I think so...

jim-docker: > I just think it is a bit strange for this component (which is built for in-cluster use) to get all the port forwards and then throw some away.

In the future we should probably check which local ports are already being used, which requires all the port-forwards or an additional function to return all reserved ports.

jim-docker: > Do we have that view in this PR? I did not see it. Nevertheless I think that think should really be an actual watch with either polling or server push but that is too big for this PR. As long as we don't expose any of this to the extension API I am fine with leaving this as it is for the time being.

By view do you mean "don't care, caller must check"? That's part of the reason for reseting and gathering all the port-forwards again. I wasn't sure I'd have time to do more specific error handling (what happens when port is already used? pod dies? etc)

jim-docker: > Why are we resetting within an init function?

I've removed the call, it was needed once, no more.

Nokel81: I meant if you were planning on adding that global port forward view (one that isn't cluster specific) in this PR.

jim-docker: No time for this PR. Still have some bugs

Nokel81: Fair enough

src/main/router.ts
```diff
@@ -179,8 +179,11 @@ export class Router {
     this.router.add({ method: "post", path: `${apiPrefix}/metrics` }, MetricsRoute.routeMetrics);
     this.router.add({ method: "get", path: `${apiPrefix}/metrics/providers` }, MetricsRoute.routeMetricsProviders);
 
-    // Port-forward API
-    this.router.add({ method: "post", path: `${apiPrefix}/pods/{namespace}/{resourceType}/{resourceName}/port-forward/{port}` }, PortForwardRoute.routePortForward);
+    // Port-forward API (the container port and local forwarding port are obtained from the query parameters)
+    this.router.add({ method: "post", path: `${apiPrefix}/pods/port-forward/{namespace}/{resourceType}/{resourceName}` }, PortForwardRoute.routePortForward);
+    this.router.add({ method: "get", path: `${apiPrefix}/pods/port-forward/{namespace}/{resourceType}/{resourceName}` }, PortForwardRoute.routeCurrentPortForward);
+    this.router.add({ method: "get", path: `${apiPrefix}/pods/port-forwards` }, PortForwardRoute.routeAllPortForwards);
```
Nokel81, line 185: Since this method gets all port-forwards (for all clusters) should it really be under `/pods`?

src/main/routes/port-forward-route.ts
```diff
@@ -96,6 +98,8 @@ class PortForward {
     try {
       await tcpPortUsed.waitUntilUsed(this.internalPort, 500, 15000);
 
+      this.forwardPort = this.internalPort;
```
Nokel81: What is the difference between `forwardPort` and `internalPort`. It seemed initially that `forwardPort` is the number requested and `internalPort` is the number received. So I am a bit confused as to why you would do this assignment.

src/renderer/components/+network-services/service-port-component.tsx
Nokel81: I think it should be removed and then in the `catch` below we could set `isPortForwarded` to `false.

src/renderer/components/+workloads-pods/pod-container-port.tsx
```diff
@@ -42,39 +45,125 @@ interface Props {
 @observer
 export class PodContainerPort extends React.Component<Props> {
```
Nokel81, line 46: Is it possible to join all these `Port` components? They seem to only differ in the `kind` passed to the `*PortForward` functions.

src/renderer/port-forward/port-forward-dialog.tsx
Nokel81: Ah I see ok

src/main/router.ts
jim-docker: I think so. Looks more consistent to me. Port-forwards are associated with pods (containers in the pods). Or we could lose `/pods` altogether...?

Nokel81: No I don't think we should loose the `/pods`... It just feels odd to me that this one route is about all clusters and the others are about single clusters. But I don't know how to remedy that. So lets leave it.

jim-docker: Hmm, they're not explicitly about single clusters but maybe they should be (we could add cluster-id as a path segment or query param?). Anyway, I think this can be resolved when we add a global page (port-forwards in the catalog?). This api is strictly for internal ipc, right? We can change it anytime.

Nokel81: Yes we can change it at any time, technically any extension could just use it but we don't really expose it.

I meant cluster scoped because you use the attached cluster object in the other route handlers.

src/main/routes/port-forward-route.ts
jim-docker: if `forwardPort` is 0 then `kubectl port-forward` will find an available port.

Nokel81: Ah I see, maybe a comment here might be a good idea then

jim-docker: I don't see `internalPort` needing to be a class field, removed it, testing, seems ok

src/renderer/components/+workloads-pods/pod-container-port.tsx
jim-docker: There are only two. They differ in the props too (one has a pod the other has a service and servicePort). Could do some more code sharing but not sure it's worth it, particularly in this PR?

src/renderer/port-forward/port-forward.store.ts
```diff
@@ -0,0 +1,179 @@
...
+
+export async function modifyPortForward(portForward: ForwardedPort, desiredPort: number): Promise<number> {
+  let port = 0;
+  
+  try {
+    await removePortForward(portForward);
+    portForward.forwardPort = desiredPort;
+    port = await addPortForward(portForward);
+  } catch (error) {
+    logger.warn(error); // don't care, caller must check
```
jweak, line 131: How would the caller check if you don't rethrow the exception? Same applies to other functions.

jweak:
Looks good! Random port works everytime but I think a nicer error on here would be good to have.

https://user-images.githubusercontent.com/774344/136353305-4636d1da-89ca-4063-adfb-167656ff1e74.mov

src/renderer/components/+workloads-pods/pod-container-port.tsx
Nokel81: Fair enough that can be a future PR

src/renderer/port-forward/port-forward.store.ts
jim-docker: Caller would query to see if the port-forward exists (is running). 

I wasn't sure I could do a better UX for errors in time for 5.3.0. I was hoping this PR could soak in master for a bit and do refinements in follow up PRs

jweak: Alright makes sense!

Nokel81 (approved):

PR #3927, opened 2021-10-01
Add tests for <Catalog /> and <CatalogEntityDetails />

chenhunghan:
- Set `safeDescriptors:false` in jest setup to be able to use `jest.spyOn`
- Add tests to `<Catalog />` and `<CatalogEntityDetails />`
- Bug fixes for `<Icon />` and `<CatalogMenu />` (found during adding tests)


src/renderer/components/icon/icon.tsx
```diff
@@ -121,7 +121,7 @@ export class Icon extends React.PureComponent<IconProps> {
     }
 
     // render as material-icon
-    if (material) {
+    if (material && typeof material === "string") {
```
jweak: Why is this check needed? Isn't the type of `material` string or undefined so it should always be a string if it's defined?

chenhunghan: yeah, if users (extension developer) are using typescript (some just use javascript). In my test, I have tried to mock a category which has `category.metadata.icon = {}` and managed to pass `{}` to material props here (and crash the component) 

https://github.com/lensapp/lens/blob/da3c8d00c7ebe96c67b8c1be09878b4234ba6bf6/src/renderer/components/%2Bcatalog/catalog-menu.tsx#L48

src/renderer/components/+catalog/catalog-entity-details.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { mockWindow } from "../../../../__mocks__/windowMock";
+ 
+// avoid TypeError: window.requestIdleCallback is not a function
+mockWindow();
+ 
+describe("<CatalogEntityDetails />", () => {
+ 
+  beforeEach(() => {
+    CatalogEntityDetailRegistry.createInstance();
+  });
```
Nokel81, line 36: You should reset the instance after each test as well. Just in case we want to add more tests in the future.

src/renderer/components/+catalog/catalog-menu.tsx
```diff
@@ -43,7 +43,7 @@ function getCategories() {
 function getCategoryIcon(category: CatalogCategory) {
   if (!category.metadata?.icon) return null;
 
-  return category.metadata.icon.includes("<svg")
+  return category.metadata.icon.includes?.("<svg")
```
Nokel81, line 46: why is this here? would it be better to just check if the `typeof` above is correct?

src/renderer/components/icon/icon.tsx
Nokel81: shouldn't you also do the check for `svg`? And wouldn't it be good enough to just do a `typeof` check?

chenhunghan: That's true, thanks.

src/renderer/components/+catalog/catalog-entity-details.test.tsx
chenhunghan: Added https://github.com/lensapp/lens/pull/3927/commits/a89ef63477b2fabfdda0dd12689e6e2fbfcf7f65

src/renderer/components/icon/icon.tsx
chenhunghan: Added https://github.com/lensapp/lens/pull/3927/commits/c55b2fa1ae0cec757985c30b2bf4f8058b1ed9f3

src/renderer/components/+catalog/catalog-menu.tsx
chenhunghan: Not sure that (adding `typeof` checks to `category.metadata?.icon`), will be the original intent of the code, there was no spec. But I guess both should be fine...

Nokel81:
Why did you close this?

PR #3935, opened 2021-10-01
Remove directly depending on openid-client

Nokel81:

closes #3932 

We don't use this directly in our code so we shouldn't pull it in as a dep


package.json
```diff
@@ -202,6 +202,7 @@
     "filehound": "^1.17.4",
     "fs-extra": "^9.0.1",
     "glob-to-regexp": "^0.4.1",
+    "got": "^11.8.2",
```
jakolehm, line 205: Please remove from this PR.

jakolehm (dismissed):

package.json
jakolehm: Was not obvious that we already had this as a dependency.

jakolehm (approved):

PR #3292, opened 2021-07-06
Refine UserStore from/to usage

Nokel81:
- This change allows new fields to be added while being more DRY about
  it

Signed-off-by: Sebastian Malton <sebastian@malton.name>

follow up to #2708

PR #3956, opened 2021-10-05
Disable kubectl download mirror options when not available

### LINKED ISSUE #3814: Azure mirror does not have kubectl binaries for MacOS
Since Azure mirror does not have kubectl binaries for MacOS, we probably should disable that option on MacOS

Nokel81:

fixes #3814

### DIFF
--- a/src/common/user-store/preferences-helpers.ts
+++ b/src/common/user-store/preferences-helpers.ts
@@ -135,12 +135,32 @@ const allowErrorReporting: PreferenceDescription<boolean> = {
   },
 };
 
+export interface DownloadMirror {
+  url: string;
+  label: string;
+  platforms: Set<NodeJS.Platform>;
+}
+
+const defaultPackageMirror = "default";
+const packageMirrors = new Map<string, DownloadMirror>([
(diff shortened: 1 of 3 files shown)


src/common/user-store/preferences-helpers.ts
```diff
@@ -289,3 +309,8 @@ export const DESCRIPTORS = {
   editorConfiguration,
   terminalCopyOnSelect,
 };
+
+export const CONSTANTS = {
```
aleksfront: Maybe it's just, but it feels like an old-schoolish way.

Why not export each of the constants separately and import them like this:

```
import { defaultPackageMirror, packageMirrors } from "../common/user-store/preferences-helpers";
```

Nokel81: okay can do

Nokel81: I want the descriptors to be separate because of https://github.com/lensapp/lens/pull/3292

aleksfront (dismissed):
LGTM

PR #3957, opened 2021-10-05
Fix taints on nodes view

### LINKED ISSUE #3847: Make taint hover text consistent with overlay

#3646 changed the node overlay to show each taint's value. However, the hover text was not changed, so if you hover over the count in the Taints column on the Nodes screen, it still only shows "\<key>: \<effect>".

**Why is this needed**:

For consistency. Also whatever motivations were behind #3646 would apply.

- Desktop OS: macos

Nokel81:

fixes #3847


jweak (approved):
LGTM

PR #3974, opened 2021-10-06
Remove request and request-promise-native in favour of got

### LINKED ISSUE #459: Get rid of "request" and "request-promise-native" (deprecation notice)
Prerequisites:
https://github.com/request/request/issues/3142

Suggested replacement:
https://github.com/mikeal/bent

Nokel81:

fixes #459


jakolehm (commented):
Why not `node-fetch`? (we are already using that)

Nokel81:
Because I think that got's API is better.

src/main/cluster-detectors/detector-registry.ts
```diff
@@ -21,43 +21,49 @@
...
 
   async detectForCluster(cluster: Cluster): Promise<ClusterMetadata> {
-    const results: {[key: string]: ClusterDetectionResult } = {};
+    const results = new Map<string, ClusterDetectionResult>();
+    const detections = this.registry.map(async DetectorClass => {
+      const detector = new DetectorClass(cluster);
 
-    for (const detectorClass of this.registry) {
-      const detector = new detectorClass(cluster);
+      return [detector.key, await detector.detect()] as const;
```
ixrock, line 28: What means `as const` and why do we need it here?

src/main/cluster-detectors/detector-registry.ts
```diff
@@ -21,43 +21,49 @@
...
     }
 
-    return metadata;
+    return Object.fromEntries(
+      iter.map(
+        Object.entries(results),
+        ([key, { value }]) => [key, value]
+      )
+    );
   }
```
ixrock: Why do we need `iter.map` here? Is it different somehow from:

```ts
return Object.fromEntries(
  Object.entries(results).map(([key, { value }]) => [key, value])
);
```

src/renderer/components/+preferences/helm-charts.tsx
```diff
@@ -34,6 +34,16 @@ import { observer } from "mobx-react";
 import { RemovableItem } from "./removable-item";
 import { Notice } from "../+extensions/notice";
 import { Spinner } from "../spinner";
+import got from "got";
+import { orderBy } from "lodash";
+
+async function loadAvailableHelmRepos(): Promise<HelmRepo[]> {
+  const { body } = await got.get("https://github.com/lensapp/artifact-hub-repositories/releases/download/latest/repositories.json", {
```
ixrock: Body is always a `string`? Can `got` do something like `got.json()` with already parsed body? (noticed usage of `JSON.parse` in many places where `got.get|*` is used)

ixrock (dismissed):
LGTM

ixrock:
> Why not `node-fetch`? (we are already using that)

Good point btw. Or just use `window.fetch` in renderer + just `fetch` is available in service-workers. 
 As far as I remember main limitation for [browser's Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) that it's limited to 4 requests per host, right? But in most cases we could use `fetch` or `node-fetch` everywhere. Wdyt?

jakolehm:
> Because I think that got's API is better.

@Nokel81 I meant that we are using `node-fetch` extensively in our JsonApi/KubeApi. If that's not good enough then we probably should consider replacing that also (it would cause breaking change).

https://github.com/lensapp/lens/blob/master/src/common/k8s-api/json-api.ts#L25

jakolehm:
> Good point btw. Or just use `window.fetch` in renderer + just `fetch` is available in service-workers.
> As far as I remember main limitation for [browser's Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) that it's limited to 4 requests per host, right? But in most cases we could use `fetch` or `node-fetch` everywhere. Wdyt?

We are already using `node-fetch` on both main & renderer because of browser fetch limitations 🙂

src/main/cluster-detectors/detector-registry.ts
Nokel81: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

It is the way to say that you want a tuple of distinct types instead of an array of the union of all possible types.

Nokel81:
@jakolehm Here is a comparison between the two (yes I admit it is from the GOT side) https://github.com/sindresorhus/got#comparison

src/renderer/components/+preferences/helm-charts.tsx
Nokel81: I think there is, let me try something.

src/main/cluster-detectors/detector-registry.ts
Nokel81: Doesn't allocate a new array. That is all (since `fromEntries` wants an iterable). I can change it if you would like.

src/renderer/components/+preferences/helm-charts.tsx
Nokel81: @ixrock Take another look?

jakolehm:
> @jakolehm Here is a comparison between the two (yes I admit it is from the GOT side) https://github.com/sindresorhus/got#comparison

I have been using got in other projects and it's fine choice but does not really warrant yet-another library included if we already have node-fetch. I'm fine with this PR because it removes deprecated `request` library (although just partially, kubernetes javascript client still uses it) -> we can change `got` -> `node-fetch` in follow-up if that feels better. Bottom line: less deps is better.

Nokel81:
I am going to close this PR, instead move to `node-fetch`. Since according to https://github.com/kubernetes-client/javascript/issues/754 that is what `kubernetes/client-javascript` will be doing as well.

PR #3983, opened 2021-10-07
Fix NamespaceSelectFilter all namespaces button

Nokel81:

Fixes some weirdness when lots of namespaces are being created in rapid succession.

### DIFF
--- a/src/renderer/components/+namespaces/namespace-select-filter.tsx
+++ b/src/renderer/components/+namespaces/namespace-select-filter.tsx
@@ -121,7 +121,7 @@ export class NamespaceSelectFilter extends React.Component<SelectProps> {
         namespaceStore.toggleSingle(namespace);
       }
     } else {
-      namespaceStore.toggleAll(true); // "All namespaces" clicked
+      namespaceStore.selectAll();
     }
   };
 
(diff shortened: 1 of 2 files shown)


src/renderer/components/+namespaces/namespace-select-filter.tsx
```diff
@@ -121,7 +121,7 @@ export class NamespaceSelectFilter extends React.Component<SelectProps> {
         namespaceStore.toggleSingle(namespace);
       }
     } else {
-      namespaceStore.toggleAll(true); // "All namespaces" clicked
+      namespaceStore.selectAll();
```
jakolehm: couldn't we just call `namespaceStore.toggleAll(false);` here?

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -167,8 +167,18 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
     this.selectNamespaces(namespace);
   }
 
+  @action
+  selectAll() {
+    this.selectNamespaces([]); // Empty array means "all"
+  }
+
+  /**
+   * @deprecated This function is broken, `showAll` is opposite of what
```
jakolehm: Not sure why we added another function if using `toggleAll(false)` does the same thing?

src/renderer/components/+namespaces/namespace-select-filter.tsx
Nokel81: Yes we could have, however the `showAll` parameter is broken. It doesn't actually do what it claims to and it doesn't really make sense to call it ever with `true`. I could fix it, but I figured that is technically a breaking change. 

Furthermore, we don't really support "toggling all" anyway since we don't remember what the previous selection was.

jakolehm (approved):

PR #3985, opened 2021-10-07
Don't attempt to sync files larger than 2MiB or non UTF-8 files

### LINKED ISSUE #3959: Scanning massive files on startup leads to 100% memory consumption and crash
**Describe the bug**
Lens on Linux will consume all memory after loading and then crash if the user home directory has massive files in it.  

For whatever reason Lens tries to read ALL files in the users home directory indiscriminately.  When it reaches a very large file it tries to read it into memory so it can parse it.  I had a 200g disk image file that was in my home directory.  After trying several Lens versions I was able to solve my issue by moving the file out of the home directory and into a subfolder.

**To Reproduce**
Place a very large (50g+) file in the home directory.  Name it "diskimage" to completely reproduce my setup.
Launch Lens and watch all of your system memory quickly be consumed.

**Expected behavior**
Lens in general probably should not be reading files in the home directory that it has no business accessing.  It should discriminate by file type and/or at least enforce a size limit before attempting to do so.

**Screenshots**
image

- Lens Version: 5.2.5-latest.20211001.2
- OS: Linux Mint  20
- Installation method (e.g. snap or AppImage in Linux):  Tested with both the deb package and AppImage installations.

**Logs:**
```
The "log" contains sensitive information such as the contents of .bash_history because it was in fact trying to parse those types of files and was generating format errors for them.  So I won't include it here.
```

**Kubeconfig:**
```
Never had a chance to create a kubeconfig file because the crash prevented it.
```

**Additional context**

Nokel81:

fixes #3959

### DIFF
--- a/src/main/catalog-sources/kubeconfig-sync.ts
+++ b/src/main/catalog-sources/kubeconfig-sync.ts
@@ -60,7 +60,8 @@ const ignoreGlobs = [
  * Even if you have a cert-file, key-file, and client-cert files that is only
  * 12kb of extra data (at 4096 bytes each) which allows for around 150 entries.
  */
-const maxAllowedFileReadSize = 2 * 1024 * 1024; // 2 MiB
+const folderSyncMaxAllowedFileReadSize = 2 * 1024 * 1024; // 2 MiB
+const fileSyncMaxAllowedFileReadSize = 16 * folderSyncMaxAllowedFileReadSize; // 32 MiB
 
 export class KubeconfigSyncManager extends Singleton {
   protected sources = observable.map<string, [IComputedValue<CatalogEntity[]>, Disposer]>();
@@ -236,7 +237,14 @@ export function computeDiff(contents: string, source: RootSource, filePath: stri
   });
 }
 
-function diffChangedConfig(filePath: string, source: RootSource, stats: fs.Stats): Disposer {
+interface DiffChangedConfigArgs {
+  filePath: string;
+  source: RootSource;
+  stats: fs.Stats;
+  maxAllowedFileReadSize: number;
+}
+
+function diffChangedConfig({ filePath, source, stats, maxAllowedFileReadSize }: DiffChangedConfigArgs): Disposer {
   logger.debug(`${logPrefix} file changed`, { filePath });
 
   if (stats.size >= maxAllowedFileReadSize) {
@@ -303,6 +311,9 @@ function watchFileChanges(filePath: string): [IComputedValue<CatalogEntity[]>, D
       const stat = await fs.promises.stat(filePath);
       const isFolderSync = stat.isDirectory();
       const cleanupFns = new Map<string, Disposer>();
+      const maxAllowedFileReadSize = isFolderSync
+        ? folderSyncMaxAllowedFileReadSize
+        : fileSyncMaxAllowedFileReadSize;
 
       watcher = watch(filePath, {
         followSymlinks: true,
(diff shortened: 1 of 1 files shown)


src/main/catalog-sources/kubeconfig-sync.ts
```diff
@@ -228,9 +228,18 @@ export function computeDiff(contents: string, source: RootSource, filePath: stri
   });
 }
 
-function diffChangedConfig(filePath: string, source: RootSource): Disposer {
+const maxAllowedFileReadSize = 16 * 1024 * 1024; // 16 MiB
```
jakolehm: Is there any realistic possibility to kubeconfig be over 1MiB? 16MiB sounds quite big text file to me 🤔

Nokel81: It is, I can shorten it. Maybe I should also try and detect if the input is going to be valid utf-8. WDYT?

Nokel81: @jakolehm I have also added the utf-8 checking. PTAL

Nokel81:
Also @jim-docker I looked into what happens now if I try and sync a non UTF-8 file (such as a PNG). The conversion to UTF-8 to "never throw" because of this line in the node docs

> If encoding is 'utf8' and a byte sequence in the input is not valid UTF-8, then each invalid byte is replaced with the replacement character U+FFFD.

so it fails at the yaml parsing step instead, which is not as good of an error message.

src/main/catalog-sources/kubeconfig-sync.ts
```diff
@@ -228,15 +236,23 @@ export function computeDiff(contents: string, source: RootSource, filePath: stri
   });
 }
 
-function diffChangedConfig(filePath: string, source: RootSource): Disposer {
+function diffChangedConfig(filePath: string, source: RootSource, stats: fs.Stats): Disposer {
   logger.debug(`${logPrefix} file changed`, { filePath });
 
+  if (stats.size >= maxAllowedFileReadSize) {
```
jim-docker, line 250: If the file is part of a folder sync this is ok to do. But if the file was explicitly sync'd I don't think it should be rejected because of size.

Nokel81: How about a larger size limit then? Just worried about accidental syncs of very large items.

jim-docker: Doesn't the utf-8 decoding short-circuit the reading if it's not utf-8? Should happen quickly on a binary file? You're concerned there might be a 200M utf-8 file? My feeling is if the user specifies the file explicitly it's "buyer beware". 

Anyway, it's fine, I'll approve shortly, just testing on windows...

src/main/catalog-sources/kubeconfig-sync.ts
```diff
@@ -236,7 +237,14 @@ export function computeDiff(contents: string, source: RootSource, filePath: stri
   });
 }
 
-function diffChangedConfig(filePath: string, source: RootSource, stats: fs.Stats): Disposer {
+interface DiffChangedConfigArgs {
+  filePath: string;
+  source: RootSource;
+  stats: fs.Stats;
```
jim-docker, line 243: this could just be `size`...

Nokel81: It could yes, shall I change it? I think, iirc, that I passed in stats because I thought I might check other things. But I cannot now think of what else I would check.

jim-docker (approved):
nice improvement!

PR #3990, opened 2021-10-07
Refactor Nodes view metrics renderer

Nokel81:
- Use a common base function to simplify logic

### DIFF
--- a/src/renderer/components/+nodes/nodes.tsx
+++ b/src/renderer/components/+nodes/nodes.tsx
@@ -56,11 +56,13 @@ enum columnId {
 interface Props extends RouteComponentProps<NodesRouteParams> {
 }
 
+type MetricsTooltipFormatter = (metrics: [number, number]) => string;
+
 interface UsageArgs {
   node: Node;
   title: string;
   metricNames: [string, string];
-  secondFormater: (metrics: [number, number]) => string;
+  formatters: MetricsTooltipFormatter[];
 }
 
 @observer
@@ -85,18 +87,17 @@ export class Nodes extends React.Component<Props> {
     if (isEmpty(this.metrics)) {
       return [];
     }
+
     const nodeName = node.getName();
 
     return metricNames.map(metricName => {
       try {
         const metric = this.metrics[metricName];
-        const result = metric.data.result.find(result => {
-          return [
-            result.metric.node,
-            result.metric.instance,
-            result.metric.kubernetes_node,
-          ].includes(nodeName);
-        });
+        const result = metric.data.result.find(({ metric: { node, instance, kubernetes_node }}) => (
+          nodeName === node
+          || nodeName === instance
+          || nodeName === kubernetes_node
+        ));
 
         return result ? parseFloat(result.values.slice(-1)[0][1]) : 0;
       } catch (e) {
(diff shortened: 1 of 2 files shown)


src/renderer/components/+nodes/nodes.tsx
```diff
@@ -98,74 +98,45 @@ export class Nodes extends React.Component<Props> {
...
-    const metrics = this.getLastMetricValues(node, ["cpuUsage", "cpuCapacity"]);
-
-    if (!metrics || !metrics[1]) return <LineProgress value={0}/>;
-    const usage = metrics[0];
-    const cores = metrics[1];
-    const cpuUsagePercent = Math.ceil(usage * 100) / cores;
-    const cpuUsagePercentLabel: String = cpuUsagePercent % 1 === 0
-      ? cpuUsagePercent.toString()
-      : cpuUsagePercent.toFixed(2);
+  renderUsage(node: Node, title: string, metricNames: [string, string], secondFormater: (metrics: [number, number]) => string) {
```
aleksfront: If this number of arguments is necessary, we should convert them into an object.

From Robert C. Martin:
> Too Many Arguments

> Functions should have a small number of arguments. No argument is best, followed by one, two, and three. More than three is very questionable and should be avoided with prejudice.

Nokel81: Why is no arguments the best? I would say that is categorically worse because it means that you have to go searching other places for a functions input.

nevalla: IMHO when having lots of arguments, it's start to get painful to call those methods. If you don't believe, please try to use kubernetes client library 😄

Nokel81: I switched it to a destructed object. I was mostly asking about the zero arguments. Because I think that functions should be as pure as possible to help reduce the mental load when reading the code and if a function has no arguments then it is almost certainly not pure.

src/renderer/components/+nodes/nodes.tsx
```diff
@@ -98,74 +105,60 @@ export class Nodes extends React.Component<Props> {
...
-      <LineProgress
-        max={capacity}
-        value={usage}
-        tooltip={{
-          preferredPositions: TooltipPosition.BOTTOM,
-          children: `Disk: ${Math.ceil(usage * 100 / capacity)}%, ${bytesToUnits(usage, 3)}`,
-        }}
-      />
-    );
+  renderDiskUsage(node: Node): any {
```
jweak: `: any` 

is unnecessary?

src/renderer/components/+nodes/nodes.tsx
```diff
@@ -56,6 +56,13 @@ enum columnId {
 interface Props extends RouteComponentProps<NodesRouteParams> {
 }
 
+interface UsageArgs {
+  node: Node;
+  title: string;
+  metricNames: [string, string];
+  secondFormater: (metrics: [number, number]) => string;
```
jweak: Should this be spelled `secondFormatter`? 

Maybe the argument name would be a bit clearer if both formatter would be passed, even though it's same for all of them for now.

src/renderer/components/+nodes/nodes.tsx
Nokel81: Yes it is unnecessary, will remove.

src/renderer/components/+nodes/nodes.tsx
Nokel81: > Should this be spelled `secondFormatter`?

yes

> Maybe the argument name would be a bit clearer if both formatter would be passed, even though it's same for all of them for now.

sure.

jweak (approved):

PR #3999, opened 2021-10-08
Use the served and storage version of a CRD instead of the first

### LINKED ISSUE #3988: CRDs fail to load
please check lens when it working with certmanager.
when i install cert-manager. lens can't load custom resource that certmanager create

image

image

_Originally posted by @nvthuong1996 in https://github.com/lensapp/lens/issues/3903#issuecomment-935823553_

Nokel81:

fixes #3988

### DIFF
--- a/src/common/k8s-api/endpoints/crd.api.ts
+++ b/src/common/k8s-api/endpoints/crd.api.ts
@@ -40,11 +40,19 @@ type AdditionalPrinterColumnsV1Beta = AdditionalPrinterColumnsCommon & {
   JSONPath: string;
 };
 
+export interface CRDVersion {
+  name: string;
+  served: boolean;
+  storage: boolean;
+  schema?: object; // required in v1 but not present in v1beta
+  additionalPrinterColumns?: AdditionalPrinterColumnsV1[];
+}
+
 export interface CustomResourceDefinition {
   spec: {
     group: string;
     /**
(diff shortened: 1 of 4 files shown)


src/common/k8s-api/endpoints/crd.api.ts
```diff
@@ -125,9 +149,36 @@ export class CustomResourceDefinition extends KubeObject {
...
+      return {
+        name: this.spec.version,
+        served: true,
+        storage: true,
+        schema: this.spec.validation,
+        additionalPrinterColumns,
+      };
+    }
+
+    throw new Error(`Failed to find a version for CustomResourceDefinition ${this.metadata.name}`);
```
jim-docker, line 177: theres something to catch this?

jim-docker (approved):

src/common/k8s-api/endpoints/crd.api.ts
Nokel81: Only the `<ErrorBoundary>` in the general case. I added this because it is more descriptive then `cannot read property "name" of undefined`. 

I was following the documentation from the kube website so I don't know what else I should do in this case.

jim-docker: I was just wondering how graceful the UI handles it

Nokel81: Probably not very well if it ever happens

Issue #2340, opened 2021-03-15
Port-Forward Enhancements


mrseanr:
**What would you like to be added**: 
Better port-forwarding tools.  It's common to open multiple port-forwards to multiple pods. In the current implementation, there is no way to know if a pod has an active forward. Also, if a forward needs to be stopped, the whole Lens App needs restarting. Lastly, not all forwards are HTTP related, opening a browser is convenient for most, however a customize button to allow mapping from host to local and suppress the browser opening would be productive.

**Why is this needed**:
The Lens App has become a daily driver for K8s management, however, regular usage of port forwards to perform tasks is cumbersome. Often reverting to command line to perform multiple port forwards.  Adding a bit of instrumentation to the UI, showing active forwards, enable stopping a forward, and customization of a forward, would be hugely beneficial.

**Environment you are Lens application on:**

- Kubernetes distribution: Common to all 
- Desktop OS: Common to all

jim-docker:
related: #2282, #2230, #542, #541

sherifabdlnaby:
Bump 👍🏻 Our Development teams are using Lens + K9S, where k9S use case is mostly for port-forwarding. Having everything in one tool will be very helpful.

PR #4000, opened 2021-10-09
poll for pods and services disappearing to remove dead port-forwards

jim-docker:

related to #2340

### DIFF
--- a/src/renderer/port-forward/port-forward-dialog.tsx
+++ b/src/renderer/port-forward/port-forward-dialog.tsx
@@ -29,9 +29,8 @@ import { Wizard, WizardStep } from "../components/wizard";
 import { Input } from "../components/input";
 import { Notifications } from "../components/notifications";
 import { cssNames } from "../utils";
-import { addPortForward, modifyPortForward } from "./port-forward.store";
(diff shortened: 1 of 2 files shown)


jim-docker:
Is there a better way than polling?

Nokel81:
> Is there a better way than polling?

You probably could start up a `watch` request. But I wonder if it is really necessary, does the port forwarding not fail if the pod is removed?

jim-docker:
> > Is there a better way than polling?
> 
> You probably could start up a `watch` request. But I wonder if it is really necessary, does the port forwarding not fail if the pod is removed?

When a pod is terminated the `kubectl port-forward` process does not stop, and the port is not released. This happens when run from the command line too. The process does report errors however, like:
```
E1013 11:48:34.876047   37778 portforward.go:400] an error occurred forwarding 40400 -> 3000: error forwarding port 3000 to pod ...
```

Nokel81:
So you could just hook into the pod store watch, right?

jim-docker:
right, thanks

jim-docker:
needs redoing due to port-forward refactoring, plus will try using pod/service watch instead

Issue #4018, opened 2021-10-12
Blank Screen when accessing one (out of three) clusters


darkpixel:
I have 3 clusters on Digital Ocean under various accounts.
I have always been able to connect to them with Lens.
About a week ago, I suddenly couldn't connect to one of the clusters.
I've tried deleting and re-adding the config in kubectl, I've tried deleting and re-adding it to Lens.
I just get a blank screen.  kubectl works fine, it's just Lens that isn't showing anything.
Lens has no trouble with the other clusters.  They are all running the same version.

Here's a screenshot of Lens and the terminal I launched it from when I try to connect to the cluster that Lens is having issues with:

![Screenshot from 2021-10-12 09-42-00](https://user-images.githubusercontent.com/158380/136996871-d8276707-7b6f-487b-9f18-234ec49a1431.png)

- Lens Version: 5.2.5
- OS: Linux (Mint)
- Installation method (e.g. snap or AppImage in Linux): AppImage

```
apiVersion: v1                                                                                                          
clusters:                                                                                                               
- cluster:                                                                                                              
    certificate-authority-data: --redacted--
    server: https://--redacted--.k8s.ondigitalocean.com                                         
  name: do-sfo2-aaronkube                                                                                               
- cluster:                                                                                                              
    certificate-authority-data: --redacted--
    server: https://--redacted--.k8s.ondigitalocean.com                                         
  name: do-sfo2-caitkube                                                                                                
- cluster:                                                                                                              
    certificate-authority-data: --redacted--
    server: https://--redacted--.k8s.ondigitalocean.com                                         
  name: do-sfo3-personalkube                                                                                            
contexts:                                                                                                               
- context:                                                                                                              
    cluster: do-sfo2-aaronkube                                                                                          
    user: do-sfo2-aaronkube-admin                                                                                       
  name: do-sfo2-aaronkube                                                                                               
- context:                                                                                                              
    cluster: do-sfo2-caitkube                                                                                           
    user: do-sfo2-caitkube-admin                                                                                        
  name: do-sfo2-caitkube                                                                                                
- context:                                                                                                              
    cluster: do-sfo3-personalkube                                                                                       
    user: do-sfo3-personalkube-admin                                                                                    
  name: do-sfo3-personalkube                                                                                            
current-context: do-sfo2-aaronkube                                                                                      
kind: Config                                                                                                            
preferences: {}                                                                                                         
users:                                                                                                                  
- name: do-sfo2-aaronkube-admin                                                                                         
[...]

PR #4020, opened 2021-10-12
Improve bootstrap logging for help with iframe debugging

Nokel81:

This should help with issues like https://github.com/lensapp/lens/issues/4018

### DIFF
--- a/src/common/base-store.ts
+++ b/src/common/base-store.ts
@@ -43,6 +43,8 @@ export abstract class BaseStore<T> extends Singleton {
   protected storeConfig?: Config<T>;
   protected syncDisposers: Disposer[] = [];
 
+  readonly displayName: string = this.constructor.name;
+
   protected constructor(protected params: BaseStoreParams<T>) {
     super();
     makeObservable(this);
@@ -56,6 +58,10 @@ export abstract class BaseStore<T> extends Singleton {
    * This must be called after the last child's constructor is finished (or just before it finishes)
    */
   load() {
+    if (!isTestEnv) {
+      logger.info(`[${kebabCase(this.displayName).toUpperCase()}]: LOADING from ${this.path} ...`);
+    }
+
     this.storeConfig = new Config({
       ...this.params,
       projectName: "lens",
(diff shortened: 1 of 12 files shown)


src/common/base-store.ts
```diff
@@ -43,6 +43,8 @@ export abstract class BaseStore<T> extends Singleton {
   protected storeConfig?: Config<T>;
   protected syncDisposers: Disposer[] = [];
 
+  readonly displayName: string = this.constructor.name;
```
ixrock, line 46: Remember info for class name would be useless/muggled when app is built with webpack's `--production=true` && `optimization.minimized = true`

Nokel81: Yes I know, which is why for all of our built-in `T extends BaseStore` I override this. I would have wanted to make it `abstract` but that would have been a breaking change.

src/common/base-store.ts
```diff
@@ -66,13 +72,13 @@ export abstract class BaseStore<T> extends Singleton {
     const res: any = this.fromStore(this.storeConfig.store);
 
     if (res instanceof Promise || (typeof res === "object" && res && typeof res.then === "function")) {
-      console.error(`${this.constructor.name} extends BaseStore<T>'s fromStore method returns a Promise or promise-like object. This is an error and must be fixed.`);
+      console.error(`${this.displayName} extends BaseStore<T>'s fromStore method returns a Promise or promise-like object. This is an error and must be fixed.`);
```
jim-docker, line 75: errors must go through console?

src/renderer/root-frame.tsx
```diff
@@ -55,11 +58,10 @@ export class LensApp extends React.Component {
 
     registerIpcListeners();
 
-    window.onbeforeunload = () => {
-      logger.info("[App]: Unload app");
-
+    window.addEventListener("beforeunload", () => {
+      logger.info(`${RootFrame.logPrefix} Unload app`);
```
jim-docker, line 62:
```suggestion
      logger.info(`${RootFrame.logPrefix} Unload root frame`);
```
?

jim-docker (approved):

src/common/base-store.ts
Nokel81: Do you mean instead of `logger`? That will be resolved by #4317

src/renderer/root-frame.tsx
Nokel81: I don't think that is necessary given that `logPrefix` is already `"[ROOT-FRAME]:"`

PR #4030, opened 2021-10-13
Fix auto-update channel checking and pinning

Nokel81:

Fixes error from auto-update `nextUpdateChannel` and from the auto-updater not using the current channel by default.

### DIFF
--- a/src/common/user-store/preferences-helpers.ts
+++ b/src/common/user-store/preferences-helpers.ts
@@ -23,9 +23,10 @@ import moment from "moment-timezone";
 import path from "path";
 import os from "os";
 import { ThemeStore } from "../../renderer/theme.store";
-import { ObservableToggleSet } from "../utils";
+import { getAppVersion, ObservableToggleSet } from "../utils";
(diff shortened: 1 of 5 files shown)


src/common/user-store/preferences-helpers.ts
```diff
@@ -272,6 +272,7 @@ const updateChannels = new Map([
     label: "Alpha"
   }],
 ]);
+const defaultUpdateChannel = new SemVer(getAppVersion()).prerelease[0].toString();
```
jim-docker: For dev builds `.prerelease` might not be defined?

Nokel81: Ah good catch, should default to `"latest"` then

jim-docker (approved):
LGTM, but didn't test, see if it works for next prereleases

Nokel81:
@jim-docker sounds good, it isn't really possible to test any other way as there doesn't seem to be a way to change where the updater looks for updates from.

PR #4036, opened 2021-10-13
Only allow extensions to use non-lens specific cluster phases

### LINKED ISSUE #4035: Entity status phase does not get reset after disconnecting
**Describe the bug**
A connected cluster, when disconnected, does not have its status updated. This is a regression from https://github.com/lensapp/lens/pull/4009

Nokel81:

fixes #4035

### DIFF
--- a/src/common/catalog-entities/kubernetes-cluster.ts
+++ b/src/common/catalog-entities/kubernetes-cluster.ts
@@ -54,6 +54,13 @@ export interface KubernetesClusterSpec extends CatalogEntitySpec {
   accessibleNamespaces?: string[];
 }
 
+export enum LensKubernetesClusterStatus {
+  DELETING = "deleting",
+  CONNECTING = "connecting",
+  CONNECTED = "connected",
+  DISCONNECTED = "disconnected"
+}
+
 export interface KubernetesClusterMetadata extends CatalogEntityMetadata {
   distro?: string;
   kubeVersion?: string;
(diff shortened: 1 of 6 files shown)


src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -59,13 +59,7 @@ export interface KubernetesClusterMetadata extends CatalogEntityMetadata {
   kubeVersion?: string;
 }
 
-export type KubernetesClusterStatusPhase = "connected" | "connecting" | "disconnected" | "deleting";
-
-export interface KubernetesClusterStatus extends CatalogEntityStatus {
-  phase: KubernetesClusterStatusPhase;
-}
-
-export class KubernetesCluster extends CatalogEntity<KubernetesClusterMetadata, KubernetesClusterStatus, KubernetesClusterSpec> {
```
Nokel81: This was done because it is just not true anymore

src/main/cluster-manager.ts
```diff
@@ -27,14 +27,23 @@ import logger from "./logger";
...
 import type { ClusterId } from "../common/cluster-types";
 
 const logPrefix = "[CLUSTER-MANAGER]:";
 
+enum LensClusterStatus {
+  DELETING = "deleting",
+  CONNECTING = "connecting",
+  CONNECTED = "connected",
+  DISCONNECTED = "disconnected"
+}
```
Nokel81: This should remove at least some instances of special string values, which should be avoided

src/main/cluster.ts
```diff
@@ -405,6 +404,7 @@ export class Cluster implements ClusterModel, ClusterState {
     this.allowedNamespaces = [];
     this.resourceAccessStatuses.clear();
     this.pushState();
+    logger.info(`[CLUSTER]: disconnect`, this.getMeta());
```
Nokel81, line 407: This was changed because otherwise the meta values were wrong slightly, they were too old to be useful.

src/main/cluster-manager.ts
ixrock: Why moving status types definition here from `kubernetes-cluster.ts`? 
Isn't it comes purely from k8s/cluster apis? What if we need to re-use this in another module?

ixrock (dismissed):

src/main/cluster-manager.ts
Nokel81: You are right, I should move them back. Not use them as the type but use them all over as a means to reduce spelling errors.

chenhunghan (dismissed):
This looks better, tested in space extension.

jim-docker (approved):

Issue #3982, opened 2021-10-07
Display current hotbar name


LFrobeen:
**What would you like to be added**:
Display the name of the currently active hotbar somewhere.

**Why is this needed**:
I think a common use-case for grouping clusters in a hotbar is to separate staging from prod clusters or to create a separation between different environments. Since it might not always be immediately clear from the clusters in a hotbar which environment one is in, it would be really nice if Lens displayed the hotbar name somewhere (e.g. the window title bar or lower status bar).

**Environment you are Lens application on:**

- Kubernetes distribution: EKS, GKE, Rancher / k3s / k3d
- Desktop OS: MacOS Big Sur 11.2.3

sujith0408:
What was the solution for this problem. Can someone guide me on this? Not able to identify the solution from the issue that has been closed.

Nokel81:
@sujith0408 It was decided after we initially implemented it that having it in the status bar wasn't a good fit. We moved it to a tooltip in https://github.com/lensapp/lens/pull/4679 but with trying it I see that it doesn't work on 5.5.2 either. So I will open an issue for that.

Nokel81:
I have opened https://github.com/lensapp/lens/issues/5517

PR #4044, opened 2021-10-14
Add context menu into sidebar

### LINKED ISSUE #3802: Lack of cluster interaction ability without hotbar pin
**Bugs relating the cluster interation:**

1. Cannot open settings panel (or for that matter anything to do with the context menu
2. ~~Cannot pin a cluster to the hotbar (now that it doesn't temporarily appear)~~ (https://github.com/lensapp/lens/issues/3754)
3. The cluster icon appears to be interactable, but isn't

### LINKED ISSUE #3811: Pin Icon on left hand side menu (next to name / icon)
Let's add a hotbar pin icon next to the cluster (catalog) name / icon on left hand side menu.

e.g. next to minikube:
image

Nokel81:

Fixes #3802

### DIFF
--- a/src/common/routes/apps.ts
+++ b/src/common/routes/apps.ts
@@ -23,7 +23,7 @@ import type { RouteProps } from "react-router";
 import { buildURL } from "../utils/buildUrl";
 
 export const appsRoute: RouteProps = {
-  path: "/apps",
+  path: "/cluster/apps",
 };
 
 export const appsURL = buildURL(appsRoute.path);
--- a/src/common/routes/apps.ts
+++ b/src/common/routes/apps.ts
@@ -23,7 +23,7 @@ import type { RouteProps } from "react-router";
 import { buildURL } from "../utils/buildUrl";
 
 export const appsRoute: RouteProps = {
-  path: "/cluster/apps",
+  path: "/apps",
 };
 
 export const appsURL = buildURL(appsRoute.path);
--- a/src/renderer/components/layout/sidebar.tsx
+++ b/src/renderer/components/layout/sidebar.tsx
@@ -211,8 +211,8 @@ export class Sidebar extends React.Component<Props> {
             const hotbarStore = HotbarStore.getInstance();
             const isAddedToActive = HotbarStore.getInstance().isAddedToActive(this.clusterEntity);
             const title = isAddedToActive
-              ? `Remove from hotbar ${hotbarStore.getActive().name}`
-              : `Add to hotbar ${hotbarStore.getActive().name}`;
+              ? "Remove from hotbar"
+              : "Add to hotbar";
             const onClick = isAddedToActive
               ? () => hotbarStore.removeFromHotbar(metadata.uid)
               : () => hotbarStore.addToHotbar(this.clusterEntity);
--- a/src/renderer/components/layout/sidebar.tsx
+++ b/src/renderer/components/layout/sidebar.tsx
@@ -42,9 +42,11 @@ import * as routes from "../../../common/routes";
 import { Config } from "../+config";
 import { catalogEntityRegistry } from "../../api/catalog-entity-registry";
 import { HotbarIcon } from "../hotbar/hotbar-icon";
-import { observable } from "mobx";
+import { makeObservable, observable } from "mobx";
 import type { CatalogEntityContextMenuContext } from "../../../common/catalog";
 import { HotbarStore } from "../../../common/hotbar-store";
+import { broadcastMessage } from "../../../common/ipc";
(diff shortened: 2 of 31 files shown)


src/renderer/components/layout/sidebar.tsx
```diff
@@ -194,6 +200,8 @@ export class Sidebar extends React.Component<Props> {
           src={spec.icon?.src}
           className="mr-5"
           onClick={() => navigate(routes.clusterURL())}
+          menuItems={this.contextMenu.menuItems}
+          onMenuOpen={() => this.clusterEntity.onContextMenuOpen(this.contextMenu)}
```
Nokel81: I don't think we need the `add to hotbar` items here because of https://github.com/lensapp/lens/issues/3754

aleksfront:
When I imagine how this context menu should look like, I outline view with dropdown arrow next to the cluster name where all badge will have hover effect.
dropdown arrow
This will be more obvious for user to see if we have some context menu.

aleksfront:
Related to what we have right now, the menu is somewhat buggy:
1. It doesn't update Add/Remove menu item properly as well as hotbar name. Tip: we can remove hotbar name here for sake of simplicity.

https://user-images.githubusercontent.com/9607060/137486374-ecafd427-454c-4cea-8d0a-cd721e2bfcd3.mov


2. It doesn't show menu when opening cluster for the first time.

https://user-images.githubusercontent.com/9607060/137486568-249cc95c-0b98-4016-984d-9a5436e1a0d2.mov


3. `Settings` and `Delete` options doesn't work.
broken settings

Nokel81:
Thanks for the review, investigating.

Nokel81:
@aleksfront Everything has now been fixed that you have found, PTALA!

https://user-images.githubusercontent.com/8225332/137499658-cfef0f30-47e5-4ead-9be6-4163bcae96e0.mov

Nokel81:
@aleksfront How important is it for the UX to be a chevron? I think it is still consistent if it is only a context menu because the icon looks like the hotbar icons.

aleksfront:
> @aleksfront How important is it for the UX to be a chevron? I think it is still consistent if it is only a context menu because the icon looks like the hotbar icons.

I guess not important for this specific PR, but for the future one.

aleksfront (commented):
It works now, but I'm not sure about changing all of the cluster routes along with adding redirects. If the one problem left and it's navigating to `Settings` page, maybe we will navigate it through your new `IpcRendererNavigationEvents.NAVIGATE_IN_APP` method right inside the `kubernetes-cluster.ts`?

```
async onContextMenuOpen(context: CatalogEntityContextMenuContext) {
    if (!this.metadata.source || this.metadata.source === "local") {
      context.menuItems.push({
        title: "Settings",
        icon: "edit",
        onClick: () => broadcastMessage(IpcRendererNavigationEvents.NAVIGATE_IN_APP, `/entity/${this.metadata.uid}/settings`)
      });
    }
...
}
```

src/common/routes/apps.ts
```diff
@@ -23,7 +23,7 @@ import type { RouteProps } from "react-router";
 import { buildURL } from "../utils/buildUrl";
 
 export const appsRoute: RouteProps = {
-  path: "/apps",
+  path: "/cluster/apps",
```
jakolehm: Why do we change routes in this PR?

aleksfront:
I'm also voting for removing hotbar name from menu. Let's just keep `Add to Hotbar` option without any specific name.

src/common/routes/apps.ts
Nokel81: As a means for determining if we should use navigate in cluster or navigate in app. But Alex showed a much better solution

Nokel81:
> I'm also voting for removing hotbar name from menu. Let's just keep `Add to Hotbar` option without any specific name.

Why? IMO we don't use the names enough (and should use them more). Even the users want a more visible way of seeing the name of the hotbar they are currently viewing, see #3982

aleksfront:
> > I'm also voting for removing hotbar name from menu. Let's just keep `Add to Hotbar` option without any specific name.
> 
> Why? IMO we don't use the names enough (and should use them more). Even the users want a more visible way of seeing the name of the hotbar they are currently viewing, see #3982

Yes, but seeing currently active hotbar name is a bit separate issue. `Add to Hotbar default` is adding more cognitive load in my perspective. This should lead to some purpose then. For example, in if we want to add option adding to some else hotbar.

aleksfront:
Code works and looks good. Suggesting to merge after we decide with hotbar name in menu.

aleksfront (approved):
LGTM

PR #4046, opened 2021-10-14
Add URL display to weblink URLs

### LINKED ISSUE #3840: User-created WebLink details should display url in properties
We should display the url of the weblink when the catalog entity properties are opened in the catalog. At the moment user can't see the url at all.

Nokel81:

fixes #3840

Screen Shot 2021-10-14 at 11 57 29 AM

### DIFF
--- a/src/renderer/initializers/catalog-entity-detail-registry.tsx
+++ b/src/renderer/initializers/catalog-entity-detail-registry.tsx
@@ -20,7 +20,7 @@
  */
 
 import React from "react";
-import { KubernetesCluster } from "../../common/catalog-entities";
+import { KubernetesCluster, WebLink } from "../../common/catalog-entities";
 import { CatalogEntityDetailRegistry, CatalogEntityDetailsProps } from "../../extensions/registries";
 import { DrawerItem, DrawerTitle } from "../components/drawer";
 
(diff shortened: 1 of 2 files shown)


src/renderer/initializers/catalog-entity-detail-registry.tsx
```diff
@@ -45,6 +45,20 @@ export function initCatalogEntityDetailRegistry() {
...
+          Details: ({ entity }: CatalogEntityDetailsProps<WebLink>) => (
+            <>
+              <DrawerTitle title="More Information" />
+              <DrawerItem name="URL">
+                {entity.spec.url}
+              </DrawerItem>
+            </>
+          ),
+        },
+      },
```
aleksfront, line 62: Ideally, this file shouldn't contain additional information for different kind of entities. Such details must be added through the specific entity files.

src/renderer/initializers/catalog-entity-detail-registry.tsx
```diff
@@ -45,6 +45,20 @@ export function initCatalogEntityDetailRegistry() {
...
         },
-      }
+      },
+      {
+        apiVersions: [WebLink.apiVersion],
+        kind: WebLink.kind,
+        components: {
+          Details: ({ entity }: CatalogEntityDetailsProps<WebLink>) => (
+            <>
+              <DrawerTitle title="More Information" />
```
aleksfront, line 55: Maybe just `More`?

aleksfront (approved):

src/renderer/initializers/catalog-entity-detail-registry.tsx
Nokel81: I don't understand, how is this different than, say, the file `src/renderer/initializers/kube-object-details-registry.tsx`?

src/renderer/initializers/catalog-entity-detail-registry.tsx
Nokel81: I don't like one word titles like that, though I also don't like the title I have now...

I would have wanted it to be in the same format as the lines above but couldn't get it to be. Probably a place for improvement in the future.

PR #4048, opened 2021-10-14
Fix accessible namespaces notification not navigating to correct settings

### LINKED ISSUE #4034: Clicking on the accessible namespaces button in notification does not go to the correct setting
**Describe the bug**
If a cluster needs accessible namespaces specified then a notification is displayed with a button to go to the settings to add them. However, clicking on that button goes only to settings and not to the namespaces part of the settings.

This seems to be a regression from #3156

Nokel81:
- Cleaned up the broadcasting of failure to that the debounding is only
  on renderer. This leads to faster initial notification.

- Improved logging to not log entire request.Response object

fixes #4034

### DIFF
--- a/src/jest.setup.ts
+++ b/src/jest.setup.ts
@@ -38,6 +38,6 @@ fetchMock.enableMocks();
 // Mock __non_webpack_require__ for tests
 globalThis.__non_webpack_require__ = jest.fn();
 
-process.on("unhandledRejection", (err) => {
+process.on("unhandledRejection", (err: any) => {
   fail(err);
 });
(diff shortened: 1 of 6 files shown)


Nokel81:
FYI This doesn't actually fix all the bugs in this because we can no longer scroll into view a sub-setting. I think we should switch this to something similar to what https://github.com/lensapp/lens/pull/3573 did from user preferences.

aleksfront (changes requested):
All good instead need to fix `Proxy` to `proxy` in `cluster-status.ts`.

```
manageProxySettings = () => {
    navigate(entitySettingsURL({
      params: {
        entityId: this.props.clusterId,
      },
      fragment: "proxy",
    }));
  };
```

Minor but cluster status link doesn't work properly.

Nokel81:
@aleksfront Fixed, WDYT about a future where the top level menus are routes?

aleksfront:
> @aleksfront Fixed, WDYT about a future where the top level menus are routes?

Sure it should be done as in Preferences page.

aleksfront (approved):

PR #4066, opened 2021-10-18
Further tweaks to release guide

jim-docker:

### DIFF
--- a/RELEASE_GUIDE.md
+++ b/RELEASE_GUIDE.md
@@ -2,15 +2,17 @@
 
 Lens releases are built by CICD automatically on git tags. The typical release process flow is the following:
 
+1. It is recommended to perform the release process from a folder used solely meant for creating releases (i.e. not your dev folder), with the Lens repo initialized (origin set to https://github.com/lensapp/lens).
+
 1. If doing a patch release checkout the `release/vMAJOR.MINOR` branch for the appropriate `MAJOR`/`MINOR` version and manually `cherry-pick` the PRs required for the patch that were commited to master. If there are any conflicts they must be resolved manually. If necessary, get assistance from the PR authors.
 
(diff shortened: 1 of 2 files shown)


RELEASE_GUIDE.md
```diff
@@ -2,23 +2,25 @@
 
 Lens releases are built by CICD automatically on git tags. The typical release process flow is the following:
 
+1. It is recommended to perform the release process from a folder used solely meant for creating releases (i.e. not your dev folder), with the Lens repo initialized (origin set to https://github.com/lensapp/lens).
```
Nokel81, line 5: What do you mean by this? I am slightly confused by this line. Also does the `REMOTE` env var not work?

RELEASE_GUIDE.md
```diff
@@ -2,23 +2,25 @@
...
+
 1. If doing a patch release checkout the `release/vMAJOR.MINOR` branch for the appropriate `MAJOR`/`MINOR` version and manually `cherry-pick` the PRs required for the patch that were commited to master. If there are any conflicts they must be resolved manually. If necessary, get assistance from the PR authors.
 
-1. From a clean and up to date `master` (or `release/vMAJOR.MINOR` if doing a patch release) run `make release-version <version-type>` where `<version-type>` is one of the following:
```
Nokel81, line 7: I would also just remove that make target completely

RELEASE_GUIDE.md
jim-docker: I mean it's best to do the release process from a clean (git) folder, and not your usual dev folder. Keep the interests separate. The release can sometimes take a while, you might be tempted to work on something else in the meantime, you might accidentally change something that then gets pushed to the release in a build fix...

GIT_REMOTE works, it's just more prone to user error (might forget to specify it)

Nokel81: Ah I see. I have never really had a problem with it taking a long time as I just will switch between branches.

Nokel81 (approved):

PR #4078, opened 2021-10-19
Stop using @electron/remote to obtain app.getPath()

Nokel81:

Fixes a bug where, for some reason, `remote.app` throws an error (during early lifetime of a frame). This fixes the issue by instead initializing the whole set in `bootstrap()` which is called far enough along that it shouldn't observe that behaviour.

The changes to `createStorage` are required because they were the initial error case, they now wait for bootstrap to call `AppPaths.init()` and then they will initialize themselves.


src/common/app-paths.ts
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  /**
+   * An alternative to `app.getPath()` for use in renderer and common.
+   * This function throws if called before initialization.
+   * @param name The name of the path field
+   */
+  static get(name: PathName): string {
+    if (!AppPaths.paths.get()) {
+      throw new Error("AppPaths.init() has not been called");
```
jweak, line 102: it could also call init itself and not throw?

Nokel81: `init()` is an async function... If the caller can, it should call `getAsync()` instead.

jweak: oh it's not marked async?

src/common/app-paths.ts
```diff
@@ -0,0 +1,114 @@
...
+if (isWindows) {
+  pathNames.push("recent");
+}
+
+export class AppPaths {
+  private static paths = observable.box<Record<PathName, string> | undefined>();
+  private static readonly ipcChannel = "get-app-paths";
+
+  /**
+   * Inititalizes the local copy of the paths from electron
```
jweak: [nit] spelling, should be Initializes

src/common/app-paths.ts
Nokel81: Sorry I misspoke. We don't want to call `ipcRenderer.sendSync` too early in the lifespan of the iframe. That is the main bug that this PR tries to fix. I guess I should make that clearer.

src/common/app-paths.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    if (ipcMain) {
+      AppPaths.initMain();
+    } else {
+      await AppPaths.initRenderer();
+    }
+  }
+
+  private static initMain(): void {
```
jim-docker, line 74: This gets called for sure before `initRenderer()`?

Nokel81: I hope so, it gets called in the main process whereas `initRenderer` gets called in each frame.

src/main/kubectl.ts
```diff
@@ -81,7 +81,7 @@ export class Kubectl {
   protected dirname: string;
 
   static get kubectlDir() {
-    return path.join(getPath("userData"), "binaries", "kubectl");
+    return path.join(app.getPath("userData"), "binaries", "kubectl");
```
jim-docker: why not `AppPaths.get()` here?

src/renderer/utils/createStorage.ts
```diff
@@ -38,65 +36,48 @@ const storage = observable({
...
-}
-
-export function createAppStorage<T>(key: string, defaultValue: T, clusterId?: string | undefined) {
   const { logPrefix } = StorageHelper;
-  const folder = path.resolve(getPath("userData"), "lens-local-storage");
-  const fileName = `${clusterId ?? "app"}.json`;
-  const filePath = path.resolve(folder, fileName);
 
   if (!storage.initialized) {
+    storage.initialized = true;
```
jim-docker, line 46: `init()` has no guard  from being called more than once now, if it is called from elsewhere

src/renderer/utils/createStorage.ts
```diff
@@ -38,65 +36,48 @@ const storage = observable({
...
     init(); // called once per cluster-view
   }
 
-  function init() {
-    storage.initialized = true;
+  async function init() {
+    const filePath = await StorageHelper.getLocalStoragePath();
```
jim-docker: should this be in the `try` block?

src/main/kubectl.ts
Nokel81: no need because `kubectl` is only ever used in main

src/renderer/utils/createStorage.ts
Nokel81: yes that is true, but since it is a private function defined within this function it cannot be called outside of this function anyway.

src/renderer/utils/createStorage.ts
Nokel81: I can add a log line to the `init()` call instead

src/main/kubectl.ts
jim-docker: seems weird to me that you wouldn't always use AppPaths. Abstract away `app.getPath()`

Nokel81: Ah fair enough. I'll change all of them then

Nokel81: Won't change all of them because that would introduce some circular dependencies on initialization but I will change most of them.

Nokel81:
@jim-docker PTALA

jim-docker (approved):

PR #4080, opened 2021-10-19
added port-forward details

### LINKED ISSUE #4061: Port-forward table rows are not clickable
**Describe the bug**
Nothing happens if user clicks "Network -> Port Forwarding" rows.

**Expected behavior**
Details panel should open (like with any other kube list view).

jim-docker:

fixes #4061

### DIFF
--- a/src/common/routes/port-forwards.ts
+++ b/src/common/routes/port-forwards.ts
@@ -27,7 +27,7 @@ export const portForwardsRoute: RouteProps = {
 };
 
 export interface PortForwardsRouteParams {
-  forwardport: string;
+  forwardport?: string;
 }
 
 export const portForwardsURL = buildURL<PortForwardsRouteParams>(portForwardsRoute.path);
--- a/src/renderer/components/+network-port-forwards/port-forward-details.tsx
+++ b/src/renderer/components/+network-port-forwards/port-forward-details.tsx
@@ -28,6 +28,7 @@ import { Drawer, DrawerItem } from "../drawer";
 import { cssNames } from "../../utils";
 import { podsApi, serviceApi } from "../../../common/k8s-api/endpoints";
 import { getDetailsUrl } from "../kube-detail-params";
+import { PortForwardMenu } from "./port-forward-menu";
 
 interface Props {
   portForward: PortForwardItem;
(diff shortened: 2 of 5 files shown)


src/common/routes/port-forwards.ts
```diff
@@ -23,10 +23,11 @@ import type { RouteProps } from "react-router";
 import { buildURL } from "../utils/buildUrl";
 
 export const portForwardsRoute: RouteProps = {
-  path: "/port-forwards"
+  path: "/port-forwards/:forwardport?"
 };
 
 export interface PortForwardsRouteParams {
+  forwardport: string;
```
Nokel81: Shouldn't this technically be optional?

src/renderer/components/+network-port-forwards/port-forward-details.tsx
```diff
@@ -0,0 +1,88 @@
...
+        <DrawerItem name="Status">
+          <span className={cssNames("status", portForward.getStatus().toLowerCase())}>{portForward.getStatus()}</span>
+        </DrawerItem>
+      </div>
+    );
+  }
+
+  render() {
+    const { hideDetails, portForward } = this.props;
+    const title = "Port Forward";
```
Nokel81: nit: why have this as a variable?

src/renderer/components/+network-port-forwards/port-forward-details.tsx
```diff
@@ -0,0 +1,88 @@
...
+}
+
+export class PortForwardDetails extends React.Component<Props> {
+
+  renderContent() {
+    const { portForward } = this.props;
+    const api = {
+      "service": serviceApi,
+      "pod": podsApi
+    }[portForward.kind];
```
Nokel81, line 46: I think this can throw, because we use `!!portForward` as the drawer open

src/renderer/components/+network-port-forwards/port-forward-details.tsx
jim-docker: right, not needed, copied from a more complex details page and missed this after simplification

src/renderer/components/+network-port-forwards/port-forward-details.tsx
jim-docker: I don't think the render code gets called if the drawer is not open, but I've added more guards to be sure

src/renderer/components/+network-port-forwards/port-forwards.tsx
```diff
@@ -36,15 +40,46 @@ enum columnId {
...
       portForwardStore.watch(),
     ]);
   }
 
+  get selectedPortForward() {
+    const { match: { params: { forwardport } } } = this.props;
+
+    return portForwardStore.items.find(pf => {
+      return pf.getForwardPort() == Number(forwardport);
+    });
```
Nokel81: This seems like it could easily be a method on `PortForwardStore`

Nokel81 (dismissed):
Looks good, one last minor point

src/renderer/components/+network-port-forwards/port-forwards.tsx
jim-docker: yes...

Nokel81 (approved):

PR #4087, opened 2021-10-20
Moving Catalog Details panel under TopBar

aleksfront:
Now it won't hide top bar components and will look like details panel in cluster view (which is also located bit under the TopBar).

Before:
details over topbar

After:
details under topbar


jweak (approved):
LGTM

PR #4089, opened 2021-10-20
Cleanup dependentStores

nevalla:
This PR removes unnecessary dependent stores from workload list components.

### DIFF
--- a/src/renderer/components/+workloads-deployments/deployments.tsx
+++ b/src/renderer/components/+workloads-deployments/deployments.tsx
@@ -31,9 +31,6 @@ import { Icon } from "../icon";
 import { DeploymentScaleDialog } from "./deployment-scale-dialog";
 import { ConfirmDialog } from "../confirm-dialog";
 import { deploymentStore } from "./deployments.store";
-import { replicaSetStore } from "../+workloads-replicasets/replicasets.store";
-import { podsStore } from "../+workloads-pods/pods.store";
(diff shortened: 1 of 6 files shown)


aleksfront (approved):
LGTM

src/renderer/components/+workloads-deployments/deployments.tsx
```diff
@@ -79,7 +76,7 @@ export class Deployments extends React.Component<Props> {
         isConfigurable
         tableId="workload_deployments"
         className="Deployments" store={deploymentStore}
-        dependentStores={[replicaSetStore, podsStore, nodesStore, eventStore]}
```
Nokel81, line 82: So looking at it, we actually do use `podsStore` and `replicaSetStore` in `DeploymentDetails` (though that component calls `reloadAll()` for some reason. I think it would be better to remove that reload calls instead.

Nokel81 (changes requested):
I am requesting changes so we can discuss this a bit more

src/renderer/components/+workloads-deployments/deployments.tsx
nevalla: User can navigate to Deployment details elsewhere too.

Nokel81: okay fair enough

Nokel81 (approved):

PR #4090, opened 2021-10-20
Add option to disable automatic apiManager registration for KubeApi

panuhorsmalahti:
If KubeApi is used by e.g. an extension, it usually shouldn't be registered to apiManager. This PR adds an option for disabling the automatic registration.


Nokel81:
If it is a bug, wouldn't it be better to just remove that line and add the calls to all the places we call the constructor?

panuhorsmalahti:
Then the risk is that a developer forgets to call the apiManager registration.. but I'm fine with it if we want to do that.

I wouldn't call this a bug necessarily, but it's a real problem for extensions.

Nokel81:
Fair enough, I don't think that we should register during a call of `forCluster` anyway...

PR #4093, opened 2021-10-20
Fix captitalization in CrdResources component

### LINKED ISSUE #4081: Inconsistent casing of CRDs
In #3257 the casing of CRDs in the sidebar was fixed to match kubectl. However, the title bar was missed. The search bar text should also be fixed.

image

Nokel81:
- Title should just be .spec.names.kind like the sidebar

- Searching should just be .spec.names.plural

fixes #4081

### DIFF
--- a/src/renderer/components/+custom-resources/crd-resources.tsx
+++ b/src/renderer/components/+custom-resources/crd-resources.tsx
@@ -93,7 +93,7 @@ export class CrdResources extends React.Component<Props> {
         customizeHeader={({ searchProps, ...headerPlaceholders }) => ({
           searchProps: {
             ...searchProps,
-            placeholder: `Search ${crd.getNames().plural}...`,
+            placeholder: `${crd.getResourceKind()} search ...`,
           },
           ...headerPlaceholders
         })}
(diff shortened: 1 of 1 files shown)


jim-docker:
Not quite?

Screen Shot 2021-10-20 at 12 03 12 PM

The title is not plural, and the search placeholder is the same?

Nokel81:
The title is the same as sidebar (which is not plural either). I think that it would be better to use the plural form of the name for search.

jim-docker:
> Not quite?
> 
 
> The title is not plural, and the search placeholder is the same?

Or that is what you intended?

jim-docker:
> The title is the same as sidebar (which is not plural either). I think that it would be better to use the plural form of the name for search.

So title is now singular, and search bar is still plural but all lowercase?

Nokel81:
Yes that is the change, I can make the search singlular too if you think that would be better. It would be very cool (IMO) if we did the following transform `"ClusterIssuer" -> "Cluster Issuers"` but that is probably a bit too tricky

jim-docker:
Can you make the placeholder "ClusterIssuer search..." ?

Nokel81:
Sure

jim-docker (approved):

PR #4094, opened 2021-10-20
Fix race condition in refreshViews

Nokel81:

I believe that this is a fix for the lens views not showing up some of the time. I noticed that if `refreshViews` was called before the cluster is ready and available then none of the views would be shown.

### DIFF
--- a/src/main/cluster-manager.ts
+++ b/src/main/cluster-manager.ts
@@ -41,6 +41,8 @@ export class ClusterManager extends Singleton {
   private store = ClusterStore.getInstance();
   deleting = observable.set<ClusterId>();
 
+  @observable visibleCluster: ClusterId | undefined = undefined;
+
   constructor() {
     super();
     makeObservable(this);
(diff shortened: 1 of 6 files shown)


src/main/cluster-manager.ts
```diff
@@ -65,6 +67,19 @@ export class ClusterManager extends Singleton {
       this.syncClustersFromCatalog(entities);
     });
 
+    reaction(() => [
+      catalogEntityRegistry.getItemsForApiKind<KubernetesCluster>("entity.k8slens.dev/v1alpha1", "KubernetesCluster"),
```
aleksfront: We can use `catalogEntityRegistry.getItemsByEntityClass(KubernetesCluster)` here.

aleksfront (commented):
While testing, suddenly hit 2 iframes showed at the same time 
2 iframes
2 clusters

aleksfront:
For some reason, one of the iframes kept visible state even if redirected to catalog. Can't reproduce more than once though. It was happened when observing spaces cluster and suddenly logged out. App redirected to catalog but the cluster view was visible behind it.

cluster view behind

Nokel81:
I cannot seem to reproduce that case...

jim-docker:
Looks like it fixes https://github.com/lensapp/lens/issues/4260, which was happening regularly on my slow windows laptop. I cannot reproduce the issue with this PR.

Nokel81:
What about when the frame just doesn't render?

jim-docker:
> What about when the frame just doesn't render?

You mean when there's just a blank UI (the user doesn't even see "Authentication proxy started")? I haven't reproduced that

Nokel81:
Yeah I mean that, thanks anyway. Do you see any reason not to merge this PR?

src/renderer/components/cluster-manager/lens-views.ts
```diff
@@ -19,82 +19,117 @@
...
+    const dispose = when(
+      () => cluster.ready,
+      () => logger.info(`[LENS-VIEW]: cluster is ready, clusterId=${clusterId}`),
+    );
+
+    when(
+      // cluster.disconnect is set to `false` when the cluster starts to connect
+      () => !cluster.disconnected,
+      () => {
+        when(
```
jim-docker, line 83: these `when()` calls don't need to be disposed?

Nokel81: No, because they would only ever be disposed on app quit

jim-docker (approved):

PR #4096, opened 2021-10-20
Fix PodDetailsList being empty over 100 pods

### LINKED ISSUE #4067: Node details pod list is empty if more than 100 pods
Probably because it switches to virtual mode if node has > 100 pods.

Nokel81:
- Add fixedHeight prop to VirtualList which bypasses AutoSizer behaviour
  when used as a child to "overflow: scroll" DOM nodes

- Minor cleanup on Table

fixes #4067

### DIFF
--- a/src/renderer/components/+workloads-pods/pod-details-list.tsx
+++ b/src/renderer/components/+workloads-pods/pod-details-list.tsx
@@ -146,6 +146,8 @@ export class PodDetailsList extends React.Component<Props> {
       return null;
     }
 
+    const virtual = pods.length > 20;
+
     return (
       <div className="PodDetailsList flex column">
         <DrawerTitle title="Pods" />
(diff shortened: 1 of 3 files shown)


src/renderer/components/virtual-list/virtual-list.tsx
```diff
@@ -98,34 +104,45 @@ export class VirtualList extends Component<Props, State> {
     this.listRef.current?.scrollToItem(index, align);
   };
 
-  render() {
-    const { width, className, items, getRow, onScroll, outerRef } = this.props;
+  renderVSL(height: number | undefined) {
```
aleksfront: `renderVSL` not a best name for the method, since `VSL` is impossible to understand at first sight. Maybe `renderList()`?

src/renderer/components/+workloads-pods/pod-details-list.tsx
```diff
@@ -144,24 +132,36 @@ export class PodDetailsList extends React.Component<Props> {
   }
 
   render() {
-    const { pods, showTitle } = this.props;
-    const virtual = pods.length > 100;
```
aleksfront, line 148: We should keep this trigger. We don't need to show empty long space if there's small amount of pods presented. Here is how it looks without `const virtual = pods.length > 100;`
long empty space

aleksfront (changes requested):

src/renderer/components/+workloads-pods/pod-details-list.tsx
Nokel81: Okay sounds good, will make it shorter (probably around 20 or so) but good point.

aleksfront (approved):

PR #4097, opened 2021-10-20
Allow port-forwards to be restarted with the same local port

jim-docker:
The current design restricts the user from restarting a port-forward with the same port. The restriction is not technically necessary, and in the event that the associated pod is terminated/restarted lifting this restriction makes it easy for the user to reconnect the port-forward. (the `kubectl port-forward` process does not stop when the associated pod is terminated).

addresses #2340

### DIFF
--- a/src/renderer/port-forward/port-forward-dialog.tsx
+++ b/src/renderer/port-forward/port-forward-dialog.tsx
@@ -49,7 +49,6 @@ const dialogState = observable.object({
 
 @observer
 export class PortForwardDialog extends Component<Props> {
-  @observable ready = false;
   @observable currentPort = 0;
   @observable desiredPort = 0;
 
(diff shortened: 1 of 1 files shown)


Nokel81:
Does this completely resolve the mentioned issue?

src/renderer/port-forward/port-forward-dialog.tsx
```diff
@@ -170,7 +166,7 @@ export class PortForwardDialog extends Component<Props> {
             contentClass="flex gaps column"
             next={this.startPortForward}
             nextLabel={this.currentPort === 0 ? "Start" : "Restart"}
-            disabledNext={!this.ready}
+            disabledNext={false}
```
Nokel81: nit: could just be removed

Nokel81 (dismissed):

jim-docker:
> Does this completely resolve the mentioned issue?

no, just wanted to make sure this PR is linked to that issue

Nokel81 (approved):

PR #4102, opened 2021-10-21
Add option for providing request options to watch requests

panuhorsmalahti:
This is needed if e.g. token for the request would expire in a subsequent watch retry. This option can be used to provide a new token.

### DIFF
--- a/src/common/k8s-api/json-api.ts
+++ b/src/common/k8s-api/json-api.ts
@@ -75,7 +75,7 @@ export class JsonApi<D = JsonApiData, P extends JsonApiParams = JsonApiParams> {
   public onData = new EventEmitter<[D, Response]>();
   public onError = new EventEmitter<[JsonApiErrorParsed, Response]>();
   
-  private getRequestOptions?: () => Promise<RequestInit>;
+  private getRequestOptions?: JsonApiConfig["getRequestOptions"];
 
   get<T = D>(path: string, params?: P, reqInit: RequestInit = {}) {
     return this.request<T>(path, params, { ...reqInit, method: "get" });
--- a/src/common/k8s-api/json-api.ts
+++ b/src/common/k8s-api/json-api.ts
@@ -52,6 +52,7 @@ export interface JsonApiConfig {
   apiBase: string;
   serverAddress: string;
   debug?: boolean;
+  getRequestOptions?: () => RequestInit;
 }
 export class JsonApi<D = JsonApiData, P extends JsonApiParams = JsonApiParams> {
   static reqInitDefault: RequestInit = {
@@ -68,10 +69,13 @@ export class JsonApi<D = JsonApiData, P extends JsonApiParams = JsonApiParams> {
     this.config = Object.assign({}, JsonApi.configDefault, config);
     this.reqInit = merge({}, JsonApi.reqInitDefault, reqInit);
     this.parseResponse = this.parseResponse.bind(this);
+    this.getRequestOptions = config.getRequestOptions;
   }
 
   public onData = new EventEmitter<[D, Response]>();
   public onError = new EventEmitter<[JsonApiErrorParsed, Response]>();
+  
+  private getRequestOptions?: () => RequestInit;
 
   get<T = D>(path: string, params?: P, reqInit: RequestInit = {}) {
     return this.request<T>(path, params, { ...reqInit, method: "get" });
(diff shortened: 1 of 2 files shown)


src/common/k8s-api/kube-api.ts
```diff
@@ -109,6 +109,7 @@ export interface IRemoteKubeApiConfig {
   }
   user: {
     token?: string;
+    getToken?: () => string;
```
jakolehm:
```suggestion
    token?: string | () => string;
```

src/common/k8s-api/json-api.ts
```diff
@@ -52,6 +52,7 @@ export interface JsonApiConfig {
   apiBase: string;
   serverAddress: string;
   debug?: boolean;
+  getRequestOptions?: () => RequestInit;
```
jakolehm: Not sure why we need this because JsonApi already has `reqInit`?

src/common/k8s-api/kube-api.ts
panuhorsmalahti: Fixed

src/common/k8s-api/json-api.ts
panuhorsmalahti: Because we need to supply a new token on retry, so we can't use a static headers value that could be set with reqInit

src/common/k8s-api/kube-api.ts
```diff
@@ -167,6 +169,13 @@ export function forRemoteCluster<T extends KubeObject>(config: IRemoteKubeApiCon
     serverAddress: config.cluster.server,
     apiBase: "",
     debug: isDevelopment,
+    ...(isFunction(token) ? {
+      getRequestOptions: async () => ({
+        headers: {
+          "Authorization": `Bearer ${await token()}`
+        }
+      })
+    } : {})
```
ixrock: Why do we check 2 times for `isFunction(token)`? Would be more readable to handle it single place..

src/common/k8s-api/json-api.ts
```diff
@@ -68,18 +69,26 @@ export class JsonApi<D = JsonApiData, P extends JsonApiParams = JsonApiParams> {
     this.config = Object.assign({}, JsonApi.configDefault, config);
     this.reqInit = merge({}, JsonApi.reqInitDefault, reqInit);
     this.parseResponse = this.parseResponse.bind(this);
+    this.getRequestOptions = config.getRequestOptions;
   }
 
   public onData = new EventEmitter<[D, Response]>();
   public onError = new EventEmitter<[JsonApiErrorParsed, Response]>();
+  
+  private getRequestOptions?: () => Promise<RequestInit>;
```
ixrock: _nit:_
```suggestion
  private getRequestOptions?: JsonApiConfig["getRequestOptions"];
```

ixrock (dismissed):
Is it required purely for extensions?

src/common/k8s-api/kube-api.ts
panuhorsmalahti: True, improved this

src/common/k8s-api/json-api.ts
panuhorsmalahti: Fixed

panuhorsmalahti:
For now it's only needed for extensions

ixrock (dismissed):

src/common/k8s-api/json-api.ts
```diff
@@ -68,18 +69,26 @@ export class JsonApi<D = JsonApiData, P extends JsonApiParams = JsonApiParams> {
...
   }
 
-  getResponse(path: string, params?: P, init: RequestInit = {}): Promise<Response> {
+  async getResponse(path: string, params?: P, init: RequestInit = {}): Promise<Response> {
     let reqUrl = `${this.config.serverAddress}${this.config.apiBase}${path}`;
-    const reqInit: RequestInit = merge({}, this.reqInit, init);
+    const reqInit: RequestInit = merge(
+      {},
+      this.reqInit,
+      this.getRequestOptions ? (await this.getRequestOptions()) : {},
```
Nokel81: I think that `getRequestOptions` should default to the function `() => ({})` and then you won't need to do this check.

panuhorsmalahti: Added default. Not sure if causes a performance hit by doing an extra await almost always though..

Nokel81 (approved):

PR #4104, opened 2021-10-21
Cluster overview css fixes

aleksfront:
Fixing annoying layout issues and move styles into css modules.

Before:
overview layout

After:

https://user-images.githubusercontent.com/9607060/138284034-93844691-c683-4b4e-891c-4a8da1edb507.mov

https://user-images.githubusercontent.com/9607060/138284201-830c9bc4-c24c-40b7-ac8b-237d0b7ccae6.mov


src/renderer/components/+cluster/cluster-issues.tsx
```diff
@@ -148,8 +148,8 @@ export class ClusterIssues extends React.Component<Props> {
 
     if (!warnings.length) {
       return (
-        <div className="no-issues flex column box grow gaps align-center justify-center">
-          <div><Icon material="check" big sticker/></div>
+        <div className={cssNames(styles.noIssues, "flex column box grow gaps align-center justify-center")}>
+          <div><Icon style={{ color: "white" }} material="check" big sticker/></div>
```
Nokel81: So what is the rule about using css in JS? I would have thought that we never should use it.

aleksfront: Fixed

Nokel81 (approved):

PR #4105, opened 2021-10-21
Fix errors in HotbarStore migrations

Nokel81:
- When the store's migration version was in between some of the
  migrations but didn't have any hotbars defined

- Add some more checks on the types of hotbars


aleksfront (approved):
LGTM

PR #4106, opened 2021-10-21
Add several more util functions and tests

Nokel81:

Just some small util functions to replace casting and what not.


src/renderer/components/dock/log-list.tsx
```diff
@@ -234,7 +234,7 @@ export class LogList extends React.Component<Props> {
   render() {
     const { isLoading } = this.props;
     const isInitLoading = isLoading && !this.logs.length;
-    const rowHeights = new Array(this.logs.length).fill(this.lineHeight);
+    const rowHeights = array.filled(this.logs.length, this.lineHeight);
```
aleksfront, line 237: Both variants almost equal and both reads easily. Do we need an `array` util if it has only `filled()` function?

src/renderer/components/+user-management/+service-accounts/secret.tsx
```diff
@@ -49,7 +49,7 @@ export class ServiceAccountsSecret extends React.Component<Props, State> {
       <>
         {!showToken && (
           <>
-            <span className="asterisks">{Array(16).fill("•").join("")}</span>
+            <span className="asterisks">{string.repeated("•", 16)}</span>
```
aleksfront: We can use lodash `repeat` https://lodash.com/docs/4.17.15#repeat here.

src/renderer/components/dock/log-list.tsx
Nokel81: The benefit of the newer version is that `rowHeights` is now typed as `number[]` instead of `any[]`

src/renderer/components/+user-management/+service-accounts/secret.tsx
Nokel81: I could switch to that too if we want. Except for debouncing and throttling (which I think are also in separate packages from the whole of lodash) I would prefer moving away from lodash since it is so big.

chenhunghan: Can we use this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat?

src/renderer/components/dock/log-list.tsx
chenhunghan: What about
```ts
const rowHeights = new Array<number>(this.logs.length).fill(this.lineHeight);
```
or it's not what you try to solve?

aleksfront (approved):
Looking good if https://github.com/lensapp/lens/pull/4106/files#r736175186 would be implemented.

src/renderer/components/dock/log-list.tsx
Nokel81: Yes that could work, but that requires the explicit type parameter. the benefit of `array.filled()` is that typescript is smart enough to infer the type parameter.

aleksfront: No harm to use `array.filled()` then.

PR #4107, opened 2021-10-21
Fix immer/merge bug

### LINKED ISSUE #4115: TypeError: Cannot assign to read only property '0' of object
Console error occurs at least in dev mode when opening a Catalog.
storage helper error

jim-docker:
This eliminates the error messages mentioned [here](https://github.com/lensapp/lens/pull/3882#issuecomment-946899952). 

But selected namespaces are always reset to just "default", and saved port-forwards are lost after disconnecting and then reconnecting the cluster.

It's a problem on loading the lens-local-storage files, because the contents are correct after disconnecting the cluster. Somewhere in the process of connecting the cluster the data is overwritten. This did work before PR #3882

### DIFF
--- a/src/renderer/utils/storageHelper.ts
+++ b/src/renderer/utils/storageHelper.ts
@@ -80,7 +80,7 @@ export class StorageHelper<T> {
     const notDefault = !this.isDefaultValue(data);
 
     if (notEmpty && notDefault) {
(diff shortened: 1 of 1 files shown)


ixrock (approved):

Nokel81 (approved):
LGTM

PR #4108, opened 2021-10-21
improved port-forward error reporting

### LINKED ISSUE #2340: Port-Forward Enhancements

jim-docker:
fixes #2340

### DIFF
--- a/src/renderer/components/+network-port-forwards/port-forward-menu.tsx
+++ b/src/renderer/components/+network-port-forwards/port-forward-menu.tsx
@@ -37,9 +37,10 @@ export class PortForwardMenu extends React.Component<Props> {
   @boundMethod
   remove() {
     const { portForward } = this.props;
+
     try {
       removePortForward(portForward);
-    } catch(eror) {
+    } catch (error) {
       Notifications.error(`Error occurred stopping the port-forward from port ${portForward.forwardPort}. The port-forward may still be active.`);
     }
   }
--- a/src/renderer/components/+network-port-forwards/port-forward-menu.tsx
+++ b/src/renderer/components/+network-port-forwards/port-forward-menu.tsx
@@ -26,6 +26,7 @@ import { MenuActions, MenuActionsProps } from "../menu/menu-actions";
 import { MenuItem } from "../menu";
 import { Icon } from "../icon";
 import { PortForwardDialog } from "../../port-forward";
+import { Notifications } from "../notifications";
 
 interface Props extends MenuActionsProps {
   portForward: PortForwardItem;
@@ -35,7 +36,12 @@ interface Props extends MenuActionsProps {
 export class PortForwardMenu extends React.Component<Props> {
   @boundMethod
   remove() {
-    return removePortForward(this.props.portForward);
+    const { portForward } = this.props;
+    try {
+      removePortForward(portForward);
+    } catch(eror) {
+      Notifications.error(`Error occurred stopping the port-forward from port ${portForward.forwardPort}. The port-forward may still be active.`);
+    }
   }
 
   renderContent() {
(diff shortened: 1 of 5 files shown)


src/renderer/components/+network-port-forwards/port-forward-menu.tsx
```diff
@@ -35,7 +36,12 @@ interface Props extends MenuActionsProps {
 export class PortForwardMenu extends React.Component<Props> {
   @boundMethod
   remove() {
-    return removePortForward(this.props.portForward);
+    const { portForward } = this.props;
+    try {
+      removePortForward(portForward);
+    } catch(eror) {
+      Notifications.error(`Error occurred stopping the port-forward from port ${portForward.forwardPort}. The port-forward may still be active.`);
```
Nokel81, line 44: can we log the error if we are not going to notify with it?

src/renderer/components/+network-services/service-port-component.tsx
```diff
@@ -71,7 +64,14 @@ export class ServicePortComponent extends React.Component<Props> {
       port: port.port,
       forwardPort: this.forwardPort,
     };
-    const activePort = await getPortForward(portForward) ?? 0;
+
+    let activePort: number;
+    try {
+      activePort = await getPortForward(portForward) ?? 0;
+    } catch(error) {
```
Nokel81: same here

src/renderer/components/+network-port-forwards/port-forward-menu.tsx
jim-docker: They are logged (and re-thrown) at a lower level (in `port-forward.store.ts`, e.g. https://github.com/lensapp/lens/blob/3c6d59645b947fb2b80c6c3cbaa089d000773036/src/renderer/port-forward/port-forward.store.ts#L128), unless I missed some

Nokel81: Wouldn't it be better to wrap the errors then instead? It seems strange to log in one place and then notify in another

Nokel81: I would do something like the following: `throw Object.assign(new Error("Failed to get port forward"), { cause: error });` until we start to use node 16 which has the following constructor for errors: `new Error("Failed to get port", { cause: error });`

jim-docker: Maybe? But it feels like the lower level (port-forward.store) telling the UI level what to report to the user...

jim-docker: I could wrap it as you suggest but I think I would still use a different message in the notification. Unless I'm misunderstanding something?

Nokel81: Sorry, I should have been more clear. I didn't mean that you should notify with the error message. But that that `logger.warn()` in `port-forward.store.ts` should be moved to where that `Notifications.error()` is.

It isn't that important, just would be nice to see both in the same place.

jim-docker: OK, but I think it might be better to log in just one place, in the function where the error is encountered, than log from every place the function is called. But maybe this is not the pattern that currently exists...

jim-docker: I'm inclined to leave it as is but if you strongly object...

Nokel81: I do not strongly

Nokel81 (approved):

PR #4109, opened 2021-10-21
Add double click to reset sidebar width

Nokel81:

Similar to how slack's sidebar has double click to reset to the default width.


aleksfront (approved):
Love that feature!

PR #4116, opened 2021-10-22
Remove iframe completely on disconnect

jakolehm:
For some reason empty iframe that is left behind on disconnect causes nodejs event loop lockups in the next iframe (for the same cluster).

### DIFF
--- a/src/renderer/components/app.tsx
+++ b/src/renderer/components/app.tsx
@@ -113,13 +113,6 @@ export class App extends React.Component {
       window.location.reload();
     });
 
-    window.addEventListener("message", (ev: MessageEvent) => {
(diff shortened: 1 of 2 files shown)


ixrock (approved):
LGTM.
Should we clean up `src/common/cluster-frames.ts` for disconnected cluster?

jakolehm:
> Should we clean up `src/common/cluster-frames.ts` for disconnected cluster?

Yes, I'll push a fix.

jakolehm:
> > Should we clean up `src/common/cluster-frames.ts` for disconnected cluster?
> 
> Yes, I'll push a fix.

@ixrock we actually already cleanup `clusterFrameMap` on disconnect/delete.

PR #4119, opened 2021-10-22
Support port forwarding https protocol

### LINKED ISSUE #3641: Port-forward should use https if port name is https

Lens should open port-forwards with https if port is named `https`.

**Why is this needed**:

Currently Lens opens browser always with `http` which will end up in an error if service is actually using https. It's common practice to name ports with `https`.

jim-docker:
fixes #3641

### DIFF
--- a/src/main/routes/port-forward-route.ts
+++ b/src/main/routes/port-forward-route.ts
@@ -34,7 +34,7 @@ interface PortForwardArgs {
   name: string;
   port: number;
   forwardPort: number;
-  protocol: string;
+  protocol?: string;
 }
 
 const internalPortRegex = /^forwarding from (?<address>.+) ->/i;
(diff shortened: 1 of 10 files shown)


src/renderer/port-forward/port-forward-dialog.tsx
```diff
@@ -60,6 +61,7 @@ export class PortForwardDialog extends Component<Props> {
   static open(portForward: ForwardedPort, options: PortForwardDialogOpenOptions = { openInBrowser: false }) {
     dialogState.isOpen = true;
     dialogState.data = portForward;
+    dialogState.useHttps = Boolean(portForward.protocol === "https");
```
Nokel81: why do you need to wrap this in `Boolean`? `===` can only produce booleans.

src/main/routes/port-forward-route.ts
```diff
@@ -119,28 +123,31 @@ export class PortForwardRoute {
...
-      }
-
       if (!portForward) {
         logger.info(`Creating a new port-forward ${namespace}/${resourceType}/${resourceName}:${port}`);
+
+        let thePort = 0;
+
+        if (forwardPort > 0 && forwardPort < 65536) {
+          thePort = forwardPort;
+        }
```
Nokel81: nit:

```suggestion
        const thePort = 0 < forwardPort && forwardPort < 65536 
          ? forwardPort 
          : 0;
```

src/renderer/port-forward/port-forward-dialog.tsx
jim-docker: oops

Nokel81 (approved):

PR #4120, opened 2021-10-22
Copy whole drawer title if no colon exists

Nokel81:

Noticed this error while playing around with port forwarding. Isn't very important


jweak (approved):
LGTM

Issue #2827, opened 2021-05-20
Prevent ability to select all texts


aleksfront:
We should go from website style to more app-like. Cursor must be set to `cursor: pointer` in most places except the details panel (drawer).

![selecting text](https://user-images.githubusercontent.com/9607060/118938906-ceef1380-b957-11eb-8ab5-33856fd3d06a.gif)

Should be like this
![not selecting text](https://user-images.githubusercontent.com/9607060/118939127-0a89dd80-b958-11eb-919c-d4f850d04708.gif)

ixrock:
Why we should disallow users to select & copy text from app views? Isn't it downgrade of UX?

jakolehm:
I think it would make sense to allow text selection from specific parts of the UI.

Nokel81:
I agree that some places it would make sense to not allow text selecting, for instance the text of buttons.

aleksfront:
App contains a lot of textual data which must have ability to be copied, so we should do otherwise issue title says: prevent to select text partially on specific areas, and not everywhere.

PR #4121, opened 2021-10-22
Add copy buttons for string DrawerItem's

### LINKED ISSUE #2200: Copying fields from overlay should not include newline

Nokel81:
- Changes <LocaleDate> to localizeDate() so that the string check passes

- Changes the rendering of booleans in CRDs to manually stringify them

resolves #2200 

This PR does remove the ability to select those `<DrawerItem>`'s that are strings (or arrays of strings) so as to be a hint that the user should go someplace else to copy the value.

@aleksfront You were right about the boolean rendering, this is a lot simpler. I do think that it is reasonable to have it as part of this PR but if you don't think so I can move it out.


src/renderer/utils/locale-date.ts
```diff
@@ -19,4 +19,13 @@
...
-export * from "./locale-date";
+import moment from "moment";
+import { UserStore } from "../../common/user-store";
+
+/**
+ * Parse `date` and reformat it based upon the settings from `UserStore`
+ * @param date The date string
+ */
+export function localizeDate(date: string): string {
+  return moment.tz(date, UserStore.getInstance().localeTimezone).format();
```
aleksfront, line 30: Can we use native `Date.prototype.toLocaleDateString()` here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString ?

src/renderer/components/drawer/drawer-item.tsx
```diff
@@ -21,30 +21,59 @@
...
 
-    if (hidden) return null;
+    if (hidden) {
+      return null;
+    }
 
-    const classNames = cssNames("DrawerItem", className, { labelsOnly });
-    const content = displayBooleans(renderBoolean, children);
+    const stringChildren = React.Children.toArray(children).filter(child => typeof child === "string");
+    const canCopy = stringChildren.length > 0;
```
aleksfront, line 62:
```suggestion
    const canCopy = React.Children.toArray(children).some(child => typeof child === "string");
```

src/renderer/components/drawer/drawer-item.scss
```diff
@@ -40,6 +51,10 @@
     max-width: 100%;
     min-width: 0;
 
+    &.noCopy {
+      user-select: none;
```
aleksfront, line 55: Though we have an issue about preventing user selection in whole app #2827, copying information from details panel might be useful. Right now, with this line, we preventing to select only some fields, which is not good.

aleksfront (changes requested):
It's a good idea to create a copy button for drawer item line 👍🏻 .

But, if we introduce it, we should show this button for every line (or at least for most of them, including container details etc). Right now it's confusing because now it's not clear why some fields contain copy button and some not.

Besides, some lines are available to select text with mouse and some not, which is also should be fixed. 

Here is example of curernt UX:

https://user-images.githubusercontent.com/9607060/138876632-a9647ce9-9d2a-4b66-8839-4eda3941f60f.mov

src/renderer/components/drawer/drawer-item.scss
Nokel81: Actually my goal here was to prevent copying on some fields. Specifically because there was a copy button provided. I did this as a hint to the user.

src/renderer/components/drawer/drawer-item.tsx
Nokel81: No, because I use `stringChildren` below. If you can think of a better name that would be great though.

src/renderer/utils/locale-date.ts
Nokel81: I don't think so because `localeTimezone` is currently a different format from what that method expects. It currently contains strings of the format `"America/Argentina/Ushuaia"`, whereas that method expects a string of the format `"en-US"`.

Nokel81:
> But, if we introduce it, we should show this button for every line (or at least for most of them, including container details etc). Right now it's confusing because now it's not clear why some fields contain copy button and some not.

So currently the rule is rather simple I admit, namely if they are directly strings copying is allowed. I guess I could support things like `<Badge>` as well which would cover more cases.

PR #4131, opened 2021-10-25
Fully support displaying all supported PodVolume types

Nokel81:

Noticed that this was missing while working on https://github.com/lensapp/lens/pull/4121. So I thought I would implement it, based the types on the GO types from `kubectl`'s repo.


aleksfront (commented):
I'm worried about large amount of newly introduced types, should we test them with appropriate Pod Volumes?

Also, I consider to move `volumeRenderers` to it's own component to avoid polluting `pod-details.tsx` since the object is huge.

Nokel81:
> I'm worried about large amount of newly introduced types, should we test them with appropriate Pod Volumes?

Probably that would be a good idea. I did do some testing but I don't think it was completely exhaustive. I tried my best to copy the types from https://github.com/kubernetes/api/blob/master/core/v1/types.go which is the canonical kubernetes types repo.

> Also, I consider to move `volumeRenderers` to it's own component to avoid polluting `pod-details.tsx` since the object is huge.

Sounds good I can do that

src/renderer/components/drawer/drawer-section.tsx
```diff
@@ -0,0 +1,47 @@
...
+      return null;
+    }
+
+    return (
+      <>
+        <DrawerSubTitle className={className} title={title} />
+        {children}
+      </>
+    );
+  }
```
aleksfront: To stick with DRY principle we can use only `DrawerSection` and pass title (which is only thing changed between these 2 components) along with `children`. For example:

```
<DrawerSection title="Some title">
  <DrawerSubTitle/>
  <div>Some content</div>
</DrawerSection>
```

src/renderer/components/drawer/drawer-title.tsx
```diff
@@ -23,3 +23,14 @@ export class DrawerTitle extends React.Component<DrawerTitleProps> {
...
+  render() {
+    const { title, children, className } = this.props;
+
+    return (
+      <div className={cssNames("DrawerSubTitle", className)}>
+        {title || children}
+      </div>
+    );
+  }
+}
```
aleksfront: Again, we shouldn't duplicate component above just for changing the class name. If we **really** need to use `DrawerSubTitle` we can use `DrawerTitle` and assign it some classname.

```
<DrawerTitle className="DrawerSubtitle"/>
```

aleksfront (commented):
Minor comments left.

src/renderer/components/+workloads-pods/details/volumes/variant.tsx
```diff
@@ -0,0 +1,291 @@
...
+import { VsphereVolume } from "./variants/vsphere-volume";
+
+const deprecatedVolumeTypes = new Set<PodVolumeKind>([
+  "flocker",
+  "gitRepo",
+  "quobyte",
+  "storageos",
+]);
+
+interface VolumeVarientProps {
```
aleksfront: nit: `VolumeVariantProps`. Same fixes for similar interfaces in this file.

aleksfront (dismissed):

aleksfront:
Just found errors while opening pod details:
ref is not a prop

aleksfront (dismissed):

src/common/k8s-api/endpoints/pods.api.ts
```diff
@@ -195,6 +195,444 @@ export interface IPodContainerStatus {
...
+  lun: number;
+  /**
+   * The type of filesystem
+   * @default "ext4"
+   */
+  fsType?: string;
+  readOnly: boolean;
+}
+
+export interface FlokerSource {
```
jim-docker:
```suggestion
export interface FlockerSource {
```

src/renderer/components/+workloads-pods/details/volumes/variant.tsx
```diff
@@ -0,0 +1,291 @@
...
+    return {
+      kind: "csi",
+      element: <ContainerStorageInterface variant={volume.csi} pod={pod} volumeName={volume.name} />,
+    };
+  }
+
+  if (volume.downwardAPI) {
+    return {
+      kind: "downwardAPI",
+      element: <DownwardApi variant={volume.downwardAPI} pod={pod} volumeName={volume.name} />,
```
jim-docker:
```suggestion
      element: <DownwardAPI variant={volume.downwardAPI} pod={pod} volumeName={volume.name} />,
```
?

src/renderer/components/+workloads-pods/details/volumes/variants/flex-volume.tsx
```diff
@@ -0,0 +1,39 @@
...
+        {readOnly.toString()}
+      </DrawerItem>
+        {
+          ...Object.entries(options)
+            .map(([key, value]) => (
+              <DrawerItem key={key} name={`Option: ${key}`}>
+                {value}
+              </DrawerItem>
+            ))
+        }
```
jim-docker: indentation off?

src/renderer/components/+workloads-pods/details/volumes/variants/empty-dir.tsx
```diff
@@ -0,0 +1,20 @@
...
+export const EmptyDir: VolumeVariantComponent<"emptyDir"> = (
+  ({ variant: { medium, sizeLimit }}) => (
+    <>
+      <DrawerItem name="Medium" hidden={!medium}>
+        {medium}
+      </DrawerItem>
+      <DrawerItem name="Size Limit" hidden={!sizeLimit}>
+        {sizeLimit}
+      </DrawerItem>
+    </>
```
jim-docker: This means nothing shows if both medium and sizeLimit are empty? Could put "Node specific" for medium, and "n/a" or "50%" for sizeLimit?

src/renderer/components/+workloads-pods/details/volumes/variants/projected.tsx
```diff
@@ -0,0 +1,82 @@
...
+        {
+          sources.map(({ secret, downwardAPI, configMap, serviceAccountToken }, index) => (
+            <React.Fragment key={index}>
+              {secret && (
+                <>
+                  <DrawerTitle size="sub-title">Secret</DrawerTitle>
+                  <DrawerItem name="Name">
+                    {secret.name}
+                  </DrawerItem>
+                  {secret.items && (
```
jim-docker: should you make sure `secret.items` is not empty too?  Or print "no items". Couple others below too.

src/renderer/components/+workloads-pods/details/volumes/variants/projected.tsx
```diff
@@ -0,0 +1,82 @@
...
+                </>
+              )}
+              {serviceAccountToken && (
+                <>
+                  <DrawerTitle size="sub-title">Service Account Token</DrawerTitle>
+                  <DrawerItem name="Audience" hidden={!serviceAccountToken.audience}>
+                    {serviceAccountToken.audience}
+                  </DrawerItem>
+                  <DrawerItem name="Expiration" hidden={!serviceAccountToken.expirationSeconds}>
+                    {serviceAccountToken.expirationSeconds}s
```
jim-docker: `toString()`?

src/renderer/components/+workloads-pods/details/volumes/variants/scale-io.tsx
```diff
@@ -0,0 +1,52 @@
...
+        title="Name"
+        kubeRef={secretRef}
+        api={secretsApi}
+      />
+      <DrawerItem name="SSL Enabled">
+        {sslEnabled.toString()}
+      </DrawerItem>
+      <DrawerItem name="SSL Enabled">
+        {sslEnabled.toString()}
+      </DrawerItem>
```
jim-docker: this is here twice

src/renderer/components/+workloads-pods/details/volumes/variants/host-path.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { DrawerItem } from "../../../../drawer";
+import type { VolumeVariantComponent } from "../variant-helpers";
+
+export const HostPath: VolumeVariantComponent<"hostPath"> = (
+  ({ variant: { path, type }}) => (
+    <>
+      <DrawerItem name="Node's Host Filesystem Path">
+        {path}
+      </DrawerItem>
+      <DrawerItem name="Check Behaviour">
```
jim-docker, line 16: Why is `type` titled with "Check Behaviour"?

src/renderer/components/+workloads-pods/details/volumes/variants/projected.tsx
```diff
@@ -0,0 +1,76 @@
...
+            <React.Fragment key={index}>
+              {secret && (
+                <>
+                  <DrawerTitle size="sub-title">Secret</DrawerTitle>
+                  <DrawerItem name="Name">
+                    {secret.name}
+                  </DrawerItem>
+                  <DrawerItem name="Items">
+                    <ul>
+                      {secret.items?.map(({ key }) => <li key={key}>{key}</li>)}
```
jim-docker: why use `key` here but `path` for downwardAPI and configMap? Other fields could be shown too? Or it's overkill?

jim-docker (dismissed):

src/renderer/components/+workloads-pods/details/volumes/variants/host-path.tsx
Nokel81: Because that is what it means, here are the docs https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

src/renderer/components/+workloads-pods/details/volumes/variants/projected.tsx
Nokel81: So for `downwardAPI` there is no `key` field so it is not possible. However, on further investigation I admit that the display for `secret` and `configMap` it could be better

src/renderer/components/+workloads-pods/details/volumes/variants/host-path.tsx
jim-docker: I read that but didn't quite understand. Doesn't seem like the best description but you're right to match the docs. Only uses the word "check" for the empty string case. Getting the word "type" in there would be better I think. "Required Type"? "Type Check Behaviour"? Or leave it.

src/renderer/components/+workloads-pods/details/volumes/variants/projected.tsx
```diff
@@ -25,7 +25,11 @@ export const Projected: VolumeVariantComponent<"projected"> = (
                   </DrawerItem>
                   <DrawerItem name="Items">
                     <ul>
-                      {secret.items?.map(({ key }) => <li key={key}>{key}</li>)}
+                      {secret.items?.map(({ key, path }) => (
+                        <li key={key}>
+                          {key} ⇢ {path}
```
jim-docker, line 30: Is the arrow appropriate? Use a colon?

jim-docker (dismissed):

src/renderer/components/+workloads-pods/details/volumes/variants/projected.tsx
Nokel81: It is a "mapping" so I think either should be fine.

jim-docker: as long as we're consistent

src/renderer/components/+workloads-pods/details/volumes/view.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  if (volumes.length === 0) {
+    return null;
+  }
+
+  return (
+    <>
+      <DrawerTitle>Volumes</DrawerTitle>
+      {volumes.map(volume => (
+        <div key={volume.name} className="volume">
```
aleksfront, line 28: Do we need setting `className` here?

src/renderer/components/+workloads-pods/pod-details-secrets.tsx
```diff
@@ -5,70 +5,49 @@
...
 
 export interface PodDetailsSecretsProps {
   pod: Pod;
 }
 
-@observer
-export class PodDetailsSecrets extends Component<PodDetailsSecretsProps> {
-  @observable secrets: Map<string, Secret> = observable.map<string, Secret>();
+export const PodDetailsSecrets = observer(({ pod }: PodDetailsSecretsProps) => {
+  const [secrets] = useState(observable.map<string, Secret>());
```
aleksfront: We should use either `useState` or `observable` map here. If getting `secrets` from state, why there is no place where we set new state?

Nokel81: Because the state tracking is done via mobx. I don't see how else we would use `observable` here. We need some way to persist the instance between rerenders which is why I am using `useState`

src/renderer/components/+workloads-pods/details/volumes/view.tsx
Nokel81: We do, it is used in the `pod-details.scss` file.

src/renderer/components/+workloads-pods/pod-details-secrets.tsx
aleksfront: I guess we don't even need to use observables here. So instead of `secrets.replace` https://github.com/lensapp/lens/pull/4131/files/6672987cd9a248bfab270604b7973ec1044cb745#diff-b4696a544159db998d3f1a489b66fb4e912e719cc420990dadd0bdfd4086b1bbR30

We can do something like this:

```
const [secrets, setSecrets] = useState(new Map());
const secrets = podSecrets.filter(Boolean).map(secret => [secret.getName(), secret])
setSecrets(new Map(secrets);
```

Nokel81: Okay I can change that

aleksfront (dismissed):

src/renderer/components/+workloads-pods/pod-details-secrets.tsx
```diff
@@ -17,18 +17,27 @@ export interface PodDetailsSecretsProps {
...
+            namespace: pod.getNs(),
+          })),
+        );
+
+        setSecrets(new Map(
+          results
+            .filter(result => result.status === "fulfilled" && result.value)
+            .map(result => (result as PromiseFulfilledResult<Secret>).value)
+            .map(secret => [secret.getName(), secret]),
+        ));
```
jim-docker, line 38: Something wrong here? The secret `storage-provisioner-token-tltgv` is not a link (it is in 5.4.4):

Screen Shot 2022-03-25 at 11 02 59 AM

Nokel81: Yes, I forgot `fireImmediately: true`

jim-docker (approved):

aleksfront (approved):

PR #4132, opened 2021-10-25
Always override PATH when syncing from shell

### LINKED ISSUE #3898: aws credential_process with gsts doesn't work
**Describe the bug**
I'm using `gsts` to auth into my AWS account via Google IDP.
It works for all my tools like kubectl, aws, etc, but Lens is complaining with:
image

My `~/aws/config` looks like:
```
[profile sts]
region = eu-west-1
credential_process =  gsts --json --idp-id xxxx --sp-id xxxx --username xxxx@xxxx.xx --aws-shared-credentials-file /home/xxxx/.aws/gsts_credentials
```

**To Reproduce**
1. Configure your aws/config just like I configured mine;
2. Launch Lens.

**Expected behavior**
I should be able to connect to my Kuberbetes/EKS cluster.

image
$ whereis gsts
gsts: /home/xxxx/.nvm/versions/node/v10.15.3/bin/gsts

Nokel81:

I cannot think of a reason not do so this merge... fixes #3898

### DIFF
--- a/src/main/shell-sync.ts
+++ b/src/main/shell-sync.ts
@@ -22,10 +22,7 @@
 import { shellEnv } from "./utils/shell-env";
 import os from "os";
 import { app } from "electron";
-
-interface Env {
-  [key: string]: string;
-}
+import logger from "./logger";
 
 /**
  * shellSync loads what would have been the environment if this application was
@@ -33,12 +30,7 @@ interface Env {
  * useful on macos where this always needs to be done.
  */
 export async function shellSync() {
-  const { shell } = os.userInfo();
-  let envVars = {};
-
-  envVars = await shellEnv(shell);
-
-  const env: Env = JSON.parse(JSON.stringify(envVars));
+  const env = await shellEnv(os.userInfo().shell);
 
   if (!env.LANG) {
     // the LANG env var expects an underscore instead of electron's dash
@@ -47,14 +39,12 @@ export async function shellSync() {
     env.LANG += ".UTF-8";
   }
 
-  // Overwrite PATH on darwin
-  if (process.env.NODE_ENV === "production" && process.platform === "darwin") {
-    process.env["PATH"] = env.PATH;
-  }
-
   // The spread operator allows joining of objects. The precedence is last to first.
   process.env = {
     ...env,
     ...process.env,
+    PATH: env.PATH,
   };
+
+  logger.debug(`[SHELL-SYNC]: Synced shell env, and updating`, env, process.env);
 }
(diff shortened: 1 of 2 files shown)


src/main/shell-sync.ts
```diff
@@ -47,14 +40,13 @@ export async function shellSync() {
     env.LANG += ".UTF-8";
   }
 
-  // Overwrite PATH on darwin
-  if (process.env.NODE_ENV === "production" && process.platform === "darwin") {
-    process.env["PATH"] = env.PATH;
-  }
+  process.env.PATH += `${isWindows ? ";" : ":"}${env.PATH}`;
```
jakolehm: Why do we append PATH now? Isn't this a bit different what we have done before on `darwin`?

Nokel81: yes it could result in some duplicate PATH entries, but this way we don't remove any from either. Maybe we could do some filtering instead and therefore a union of the two sets. WDYT?

jakolehm: Yeah, does not remove any but it also might change the priority of paths because we always keep original PATH at the beginning. Are we really sure that we need append instead of replace?

jakolehm: Also not sure if we can do this on every environment, like `snap`: https://snapcraft.io/docs/environment-variables

Nokel81: Maybe replace is good enough. I'll change it

jakolehm: @Nokel81 did you notice my comment about snap?

Nokel81: I did, but I figured that if the environment is the same everywhere then this assignment won't do anything.

Should I check for snap-ness and just skip this assignment?

Nokel81: I'll skip the assignment when in snap and on windows

Nokel81:
@jakolehm PTALA

src/main/shell-sync.ts
```diff
@@ -22,23 +22,16 @@
...
  * useful on macos where this always needs to be done.
  */
 export async function shellSync() {
-  const { shell } = os.userInfo();
-  let envVars = {};
-
-  envVars = await shellEnv(shell);
-
-  const env: Env = JSON.parse(JSON.stringify(envVars));
+  const env = await shellEnv(os.userInfo().shell);
```
jakolehm, line 34: Just wondering why we are passing always users default shell and not the current one (if current is set). Probably story for another issue/PR.

jakolehm (approved):

src/main/shell-sync.ts
Nokel81: Yeah probably for a different PR. Don't think this is wrong though, where else would they define it?

PR #4139, opened 2021-10-26
Navigate cluster context menus via broadcastMessage

### LINKED ISSUE #4088: Dropdown on cluster icon does not show "settings" and "delete"
On the Cluster view we are not adding the entries provided through extension api. Therefor some entries are missing:
Bildschirmfoto 2021-10-20 um 10 39 18

**Hint:**
- on the hotbar or cluster overview in catalog the are properly displayed.

aleksfront:
Part 1 of 2 PRs in order to fix #4088 

Extension should navigate to `Settings` no matter where this action triggered - in hotbar menu or from in-cluster view.

Should be merged along with https://github.com/lensapp/lenscloud-lens-extension/pull/812

### DIFF
--- a/src/renderer/components/layout/sidebar.tsx
+++ b/src/renderer/components/layout/sidebar.tsx
@@ -57,8 +57,8 @@ export class Sidebar extends React.Component<Props> {
   static displayName = "Sidebar";
   @observable private contextMenu: CatalogEntityContextMenuContext = {
     menuItems: [],
-    navigate: (url: string, inMainFrame = true) => {
-      if (inMainFrame) {
+    navigate: (url: string, forceMainFrame = true) => {
+      if (forceMainFrame) {
         broadcastMessage(IpcRendererNavigationEvents.NAVIGATE_IN_APP, url);
       } else {
         navigate(url);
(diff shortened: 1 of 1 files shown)


Nokel81 (changes requested):
No this is incorrect because you might what to navigate in the cluster too

aleksfront:
> No this is incorrect because you might what to navigate in the cluster too

Added `inMainFrame` flag to choose frame to navigate. @Nokel81 ptal.

src/renderer/components/layout/sidebar.tsx
```diff
@@ -55,7 +57,13 @@ export class Sidebar extends React.Component<Props> {
   static displayName = "Sidebar";
   @observable private contextMenu: CatalogEntityContextMenuContext = {
     menuItems: [],
-    navigate,
+    navigate: (url: string, inMainFrame = true) => {
```
Nokel81: I would call this `forceTopLocation` or something because `navigate("/settings", false)` will still work if called in the main frame.

aleksfront: Ended up with `forceMainFrame`.

Nokel81 (approved):
Should we do this for every context? Maybe by having a helper function that creates the context?

PR #4142, opened 2021-10-26
Close windows before trying to update

Nokel81:

Attempt #1 to try and fix auto-update.


src/main/app-updater.ts
```diff
@@ -39,7 +40,11 @@ function handleAutoUpdateBackChannel(event: Electron.IpcMainEvent, ...[arg]: Upd
   if (arg.doUpdate) {
     if (arg.now) {
       logger.info(`${AutoUpdateLogPrefix}: User chose to update now`);
-      autoUpdater.quitAndInstall(true, true);
+      setImmediate(() => {
+        app.removeAllListeners("window-all-closed");
```
jim-docker, line 44: This doesn't stop something important getting done? Like stuff getting written to file?

jim-docker (approved):

src/main/app-updater.ts
Nokel81: Of course anything can be added to event listeners, but the purpose of that event is to block the default quit actions so I doubt it.

PR #4146, opened 2021-10-26
Remove unused deps

Nokel81:

I ran the `depcheck` tool and then double checked the results and came out with these that are not used at all in the repo.


aleksfront (approved):

yarn.lock
```diff
@@ -12540,11 +12456,6 @@ serialize-javascript@^3.1.0:
   dependencies:
     randombytes "^2.1.0"
 
-serializr@^2.0.5:
```
ixrock, line 12543: Btw, this tool can do mostly similar things as `joi`. Why it was chosen over `serializr`?
More info: https://github.com/mobxjs/serializr#introduction

ixrock (approved):

yarn.lock
Nokel81: Because Jari pointed me to `Joi` when I was looking for something like that and because I didn't know we were already depending on this.

PR #4167, opened 2021-10-28
Clear up some NamespaceStore.toggleAll and NamespaceStore.hasAllContexts

Nokel81:
- Deprecate toggleAll in favour of selectAll because the other was
  dangerous, mostly broken, and didn't toggle

- Fix hasAllContexts to be true in the [] case

### DIFF
--- a/src/renderer/components/+namespaces/namespace.store.ts
+++ b/src/renderer/components/+namespaces/namespace.store.ts
@@ -148,8 +148,11 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
       .every(namespace => this.selectedNamespaces.includes(namespace));
   }
 
+  /**
+   * Is `true` if all available namespaces are selected, otherwise `false`
+   */
   @computed get hasAllContexts(): boolean {
-    return this.selectedNamespaces.length === this.allowedNamespaces.length;
+    return this.contextNamespaces.length === this.allowedNamespaces.length;
   }
 
   @action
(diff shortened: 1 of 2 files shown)


src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -148,8 +148,11 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
       .every(namespace => this.selectedNamespaces.includes(namespace));
   }
 
+  /**
+   * Is `true` if all available namespaces are selected, otherwise `false`
+   */
   @computed get hasAllContexts(): boolean {
```
jim-docker, line 154: For extensions? It's not called anywhere from Lens

src/renderer/components/+namespaces/namespace.store.ts
```diff
@@ -167,17 +170,21 @@ export class NamespaceStore extends KubeObjectStore<Namespace> {
...
-      }
-    } else {
-      this.toggleAll(!this.hasAllContexts);
-    }
+  /**
+   * Selects all available namespaces.
+   *
+   * Note: If new namespaces appear in the future those will be selected too
+   */
+  selectAll() {
```
jim-docker, line 178: no more `@action`?

jim-docker (approved):

src/renderer/components/+namespaces/namespace.store.ts
Nokel81: Action is only really required when you do several changes in one go

src/renderer/components/+namespaces/namespace.store.ts
Nokel81: Yes... Don't want too many breaking changes. This really shouldn't be exported but I don't think we can change it.

PR #4175, opened 2021-10-28
Only listen to CMD+f for <SearchInput> on macOS

### LINKED ISSUE #2423: disabling search via ctrl-f in macOS
remove ctrl-f key for search on macOS

**Why is this needed**:
In macOS, the search function works for both ctrl-f and cmd-f. As a result, when using vim in the terminal, it conflicts with the key to move to the next page.
In the case of Google Chrome, the search function in macOS only works with the cmd-f keys, but the ctrl-f keys do not.

macOS Big Sur: 11.2.3
Lens: 4.1.5

Nokel81:

resolves #2423

### DIFF
--- a/src/renderer/components/input/search-input.tsx
+++ b/src/renderer/components/input/search-input.tsx
@@ -26,6 +26,7 @@ import { observer } from "mobx-react";
 import { boundMethod, cssNames } from "../../utils";
 import { Icon } from "../icon";
 import { Input, InputProps } from "./input";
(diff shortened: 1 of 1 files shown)


aleksfront (approved):
LGTM. However, couldn't reproduce problem `ctrl+f` vim problem described in https://github.com/lensapp/lens/issues/2423

Nokel81:
Fair enough

PR #4176, opened 2021-10-28
ensure namespace placeholder reflects what will be saved to local storage

### LINKED ISSUE #4171: "All namespaces" is not the same as manually selecting all the namespaces
One small issue with https://github.com/lensapp/lens/pull/3983. If I manually select all the namespaces in the dropdown, the dropdown wrongly switches to showing "All namespaces". It should just show the list of the ones I selected. This looks to be only a UI issue because it correctly saves the list, not `[]` to local storage.

jim-docker:
If all of the namespaces have been manually selected by the user then those
namespaces are explicitly saved to the namespace store (not `[]`, representing
all namespaces). Thus, in this case, the placeholder in the selector should show this list,
not "All namespaces".

fixes #4171

### DIFF
--- a/src/renderer/components/+namespaces/namespace-select-filter.tsx
+++ b/src/renderer/components/+namespaces/namespace-select-filter.tsx
@@ -36,10 +36,10 @@ import { isMac } from "../../../common/vars";
 const Placeholder = observer((props: PlaceholderProps<any, boolean>) => {
   const getPlaceholder = (): React.ReactNode => {
     const namespaces = namespaceStore.contextNamespaces;
+    const length = namespaceStore.selectedNamespaces.length ? namespaces.length : 0;
 
-    switch (namespaces.length) {
+    switch (length) {
       case 0:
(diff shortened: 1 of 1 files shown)


src/renderer/components/+namespaces/namespace-select-filter.tsx
```diff
@@ -36,10 +36,10 @@ import { isMac } from "../../../common/vars";
 const Placeholder = observer((props: PlaceholderProps<any, boolean>) => {
   const getPlaceholder = (): React.ReactNode => {
     const namespaces = namespaceStore.contextNamespaces;
+    const length = namespaceStore.selectedNamespaces.length ? namespaces.length : 0;
```
Nokel81: why not just `namspaceStore.selectedNamespaces.length`?

jim-docker: you are right, thanks.

Nokel81 (dismissed):

jakolehm (approved):

PR #4185, opened 2021-10-29
Prevent injection of expired "DST Root CA X3" from host CA stores; Electron to 13.6

chenhunghan:
- Replace `mac-ca` module with our own logics, as `mac-ca` doesn't have api to "only get ca without injection"
- Prevent injections of expired `DST Root CA X3`
- Revert https://github.com/lensapp/lens/pull/4186

This is need because https://github.com/electron/electron/pull/31220/files removed both the `X509_V_FLAG_TRUSTED_FIRST` flag and expired `DST Root CA X3`. We shouldn't inject (again) expired `DST Root CA X3` using `mac-ca/win-ca` to prevent `https` reject because `DST Root CA X3` is expired in the certification chain.

The fix can be verified by
```
var https = require('https'); https.request("https://k8slens.dev");
```
in developer console.

Closes https://github.com/lensapp/lens/pull/4209
Closes https://github.com/lensapp/lenscloud-lens-extension/issues/839


src/common/system-ca.ts
```diff
@@ -20,22 +20,87 @@
...
+
+export const expiredCA = (cert: string) => {
```
Nokel81: This function returns true if cert is not expired. Maybe the name should be `isCertExtant` or `isCertActive`

src/common/system-ca.ts
```diff
@@ -20,22 +20,87 @@
...
+export const getMacRootCA = async () => {
+  // inspired mac-ca https://github.com/jfromaniello/mac-ca
+  const args = "find-certificate -a -p";
+  const splitPattern = /(?=-----BEGIN\sCERTIFICATE-----)/g;
+  const systemRootCertsPath = "/System/Library/Keychains/SystemRootCertificates.keychain";
+  const bin = "/usr/bin/security";
+  const trusted = (await execAsync(`${bin} ${args}`)).stdout.toString().split(splitPattern);
+  const rootCA = (await execAsync(`${bin} ${args} ${systemRootCertsPath}`)).stdout.toString().split(splitPattern);
+
+  return [...new Set([...trusted, ...rootCA])].filter(expiredCA);
```
Nokel81: Don't we want to readd the split pattern or is that what the `?=` is for in the regex?

src/common/system-ca.ts
```diff
@@ -20,22 +20,87 @@
...
+ * Only return non-expred certificates.
+ */
+export const getMacRootCA = async () => {
+  // inspired mac-ca https://github.com/jfromaniello/mac-ca
+  const args = "find-certificate -a -p";
+  const splitPattern = /(?=-----BEGIN\sCERTIFICATE-----)/g;
+  const systemRootCertsPath = "/System/Library/Keychains/SystemRootCertificates.keychain";
+  const bin = "/usr/bin/security";
+  const trusted = (await execAsync(`${bin} ${args}`)).stdout.toString().split(splitPattern);
+  const rootCA = (await execAsync(`${bin} ${args} ${systemRootCertsPath}`)).stdout.toString().split(splitPattern);
```
Nokel81: We already have `promiseExec` and `promiseExecFile` under `src/main/promise-exec.ts`. Maybe we should move that file to `src/common/utils/`

src/common/system-ca.ts
```diff
@@ -20,22 +20,87 @@
...
+    const CAs: string[] = [];
+
+    wincaAPI({
+      format: wincaAPI.der2.pem,
+      inject: false,
+      ondata: (ca: string) => {
+        CAs.push(ca);
+      },
+      onend: () => {
+        resolve(CAs.filter(expiredCA));
```
Nokel81: Why not check if they are filtered before pushing them?

src/common/system-ca.ts
```diff
@@ -20,22 +20,87 @@
...
+  });
+};
+
+/**
+ * Add (or merge) CAs to https.globalAgent.options.ca
+ */
+export const injectCAs = (CAs: Array<string>) => {
+  for (const cert of CAs) {
+    if (Array.isArray(https.globalAgent.options.ca)) {
+      !https.globalAgent.options.ca.includes(cert) && https.globalAgent.options.ca.push(cert);
```
Nokel81: Please don't use `&&` as an alias for `if` when not in `tsx` mode.

src/common/system-ca.ts
```diff
@@ -20,22 +20,87 @@
...
+    } else {
+      https.globalAgent.options.ca = [cert];
```
Nokel81, line 83: This seems strange, why does it continuously override `https.globalAgent.options.ca`? Won't this only produce the last cert in `CAs`

src/common/system-ca.ts
```diff
@@ -20,22 +20,87 @@
...
+        resolve(CAs.filter(expiredCA));
+      }
+    });
+  });
+};
+
+/**
+ * Add (or merge) CAs to https.globalAgent.options.ca
+ */
+export const injectCAs = (CAs: Array<string>) => {
```
Nokel81: nit, we don't tend to use `Array<T>` in `lensapp/lens`

src/common/system-ca.ts
```diff
@@ -20,22 +20,87 @@
...
-  try {
-    winca.inject("+"); // see: https://github.com/ukoloff/win-ca#caveats
-  } catch (error) {
-    logger.error(`[CA]: failed to force load: ${error}`);
-  }
+  getWinRootCA().then((winRootCAs) => {
+    wincaAPI.inject("+", winRootCAs);
+  }).catch((error) => {
+    console.warn(`[WIN-CA]: Error injecting root CAs from Windows. ${error?.message}`);
+  });
```
Nokel81: Would be nice if this was an exported function that is called in `src/main/index.ts` instead, or does that break stuff?

src/common/system-ca.ts
```diff
@@ -20,22 +20,87 @@
...
+
+export const expiredCA = (cert: string) => {
+  if (typeof cert === "string" && cert.includes(DSTRootCAX3)) {
+    return false;
+  }
+
+  return true;
```
Nokel81: Also it seems weird to return booleans from an `if`.

src/common/__tests__/system-ca.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  const deps = { ...dependencies, ...devDependencies };
+
+  /**
+   * The test to ensure using win-ca/api injects CAs in the same way as using 
+   * the auto injection (require('win-ca').inject('+'))
+   * 
+   * Skip the test if win-ca is not installed, or os is not win32
+   */
+  (deps["win-ca"] && os.platform().includes("win32") ? it: it.skip)("should inject the same ca as winca.inject('+')", async () => {
```
Nokel81: Why not have these checks on the `describes` themselves?

src/common/__tests__/system-ca.test.ts
```diff
@@ -0,0 +1,122 @@
...
+  (deps["mac-ca"] && os.platform().includes("darwin") ? it: it.skip)("should inject the same ca as mac-ca", async () => {
+    const osxCAs = await getMacRootCA();
+
+    injectCAs(osxCAs);
+    const injected = https.globalAgent.options.ca;
+
+    await import("mac-ca");
+    const injectedByMacCA = https.globalAgent.options.ca;
+
+    // @ts-ignore
```
Nokel81: nit I would rather see casts on lines 50 and 53 to `(string | Buffer)[]`

Nokel81 (commented):
Will test with a build on windows

Nokel81:
Also if we do this, shall we revert the reversion of 13.6.0?

src/common/system-ca.ts
chenhunghan: Because onData is called few times before promise is resolved.

Nokel81: That is not what I meant, I meant why not have the `onData` handler be:

```ts
if (!expiredCA(ca)) {
  CAs.push(ca);
}
```

src/common/system-ca.ts
chenhunghan: Sorry I don't understand what you mean..

src/common/system-ca.ts
chenhunghan: Ok, I don't think that's better, we only need to check once before promise is resolved IMHO.

src/common/system-ca.ts
chenhunghan: Ok.

src/common/system-ca.ts
chenhunghan: Ok, thanks.

src/common/system-ca.ts
chenhunghan: Ok, I think we should have some linter rules to allow/disallow though.

src/common/system-ca.ts
chenhunghan: Yeah, that should be better, hope not breaking anything...

src/common/system-ca.ts
chenhunghan: What do you suggest instead?

src/common/__tests__/system-ca.test.ts
chenhunghan: Jest's describe.skip is not so easy to use.. but I can try.

src/common/system-ca.ts
chenhunghan: It won't override but .push instead if it sees an array. but I guess this can be more readable...

src/common/system-ca.ts
Nokel81: I looked it up myself and it does. Your regex `const splitPattern = /.../;` is of the form `(?=...)` and I was asking about the `?=` there. I looked it up and it is called a "positive lookahead". Namely it matches when the `...` matches but it doesn't bind at all. So when you use it as a part of `split` the part matched is still part of the entries.

For example:

```ts
const x = "a,b,c,d,e";

x.split(/,/); // ["a", "b", "c", "d", "e"]
x.split(/(?=,)/); // ["a", ",b", ",c", ",d", ",e"]
```

Might be a good idea to add a comment about it with a link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Cheatsheet#other_assertions

src/common/system-ca.ts
Nokel81: ok

src/common/system-ca.ts
Nokel81: Fair enough, we should indeed to some more lint rule passes.

src/common/system-ca.ts
Nokel81: Oh I see what it does now, yeah slightly more readable or maybe a comment would be good.

src/common/system-ca.ts
Nokel81: `return typeof cert !== "string" || !cert.includes(DSTRootCAX3);`

src/common/system-ca.ts
chenhunghan: Thanks, it was copied from mac-ca, so honestly I don't understand what that regex means.

Nokel81: Fair enough

src/common/system-ca.ts
jweak: I like the booleans better. Maybe the best would be to put it in a descriptive variable and return that.

Nokel81: That is good to, something like?

```
const isExpired = typeof cert !== "string" || cert.includes(DSTRootCAX3);

return !isExpired;
```

src/common/system-ca.ts
```diff
@@ -20,22 +20,87 @@
  */
 
 import { isMac, isWindows } from "./vars";
-import winca from "win-ca";
-import macca from "mac-ca";
-import logger from "../main/logger";
+// @ts-expect-error winca/api module doesn't have a type definition
```
Nokel81: please just add a new declare a new module in `types/mocks.d.ts

src/common/__tests__/system-ca.test.ts
Nokel81: How is it not so easy to use? Also you can use `describeIf` and `itIf` from the integration test helpers (though those should probably be moved because we don't have to just use them in the integration code)

src/common/system-ca.ts
chenhunghan: Fixed in https://github.com/lensapp/lens/pull/4185/commits/64170cb1629f6a8cbb94694dc881af0d63ca134f

src/common/system-ca.ts
chenhunghan: Fixed in https://github.com/lensapp/lens/pull/4185/commits/64170cb1629f6a8cbb94694dc881af0d63ca134f

src/common/system-ca.ts
chenhunghan: Fixed in https://github.com/lensapp/lens/pull/4185/commits/64170cb1629f6a8cbb94694dc881af0d63ca134f

src/common/system-ca.ts
chenhunghan: Fixed in https://github.com/lensapp/lens/pull/4185/commits/64170cb1629f6a8cbb94694dc881af0d63ca134f

src/common/__tests__/system-ca.test.ts
chenhunghan: Fixed https://github.com/lensapp/lens/pull/4185/commits/78e95abb8a19749d9d432ddec833fb6c97167cc0

src/common/__tests__/system-ca.test.ts
chenhunghan: Fixed https://github.com/lensapp/lens/pull/4185/commits/78e95abb8a19749d9d432ddec833fb6c97167cc0

src/common/system-ca.ts
chenhunghan: Added https://github.com/lensapp/lens/pull/4185/commits/ab5ba848edbba2deb2b2d6c16d3b7c4b5d38557f

src/common/system-ca.ts
chenhunghan: Done https://github.com/lensapp/lens/pull/4185/commits/b9f894fafca88209d973e1b8f110f18028e3f376

src/common/system-ca.ts
chenhunghan: Done https://github.com/lensapp/lens/pull/4185/commits/65f3d12abc72a587f4aaf8149659f14a762753be

src/common/system-ca.ts
chenhunghan: Done https://github.com/lensapp/lens/pull/4185/commits/e6a178b6dab527f22857e805c2baec79ec304da0

chenhunghan:
> Also if we do this, shall we revert the reversion of 13.6.0?

Yes, we can

src/main/index.ts
```diff
@@ -121,6 +121,8 @@ app.on("second-instance", (event, argv) => {
 });
 
 app.on("ready", async () => {
+  await injectSystemCAs();
```
jakolehm: Should we call this earlier to match previous behaviour more closely?

src/main/helm/helm-chart-manager.ts
```diff
@@ -24,7 +24,7 @@ import v8 from "v8";
 import * as yaml from "js-yaml";
 import type { HelmRepo } from "./helm-repo-manager";
 import logger from "../logger";
-import { promiseExecFile } from "../promise-exec";
+import { promiseExecFile } from "../../common/utils/promise-exec";
```
Nokel81, line 27: nit: `common/utils/index.ts` should have `export * from "./promise-exec";`

src/common/system-ca.ts
```diff
@@ -20,22 +20,93 @@
...
...-----\nMIIDSjCCAjKgAwIBAgIQRK+wgNajJ7qJMDmGLvhAazANBgkqhkiG9w0BAQUFADA/\nMSQwIgYDVQQKExtEaWdpdGFsIFNpZ25hdHVyZSBUcnVzdCBDby4xFzAVBgNVBAMT\nDkRTVCBSb290IENBIFgzMB4XDTAwMDkzMDIxMTIxOVoXDTIxMDkzMDE0MDExNVow\nPzEkMCIGA1UEChMbRGlnaXRhbCBTaWduYXR1cmUgVHJ1c3QgQ28uMRcwFQYDVQQD\nEw5EU1QgUm9vdCBDQSBYMzCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEB\nAN+v6ZdQCINXtMxiZfaQguzH0yxrMMpb7NnDfcdAwRgUi+DoM3ZJKuM/IUmTrE4O\nrz5Iy2Xu/NMhD2XSKtkyj4zl93ewEnu1lcCJo6m67XMuegwGMoOifooUMM0RoOEq\nOLl5CjH9UL2AZd+3UWODyOKIYepLYYHsUmu5ouJLGiifSKOeDNoJjj4XLh7dIN9b\nxiqKqy69cK3FCxolkHRyxXtqqzTWMIn/5WgTe1QLyNau7Fqckh49ZLOMxt+/yUFw\n7BZy1SbsOFU5Q9D8/RhcQPGX69Wam40dutolucbY38EVAjqr2m7xPi71XAicPNaD\naeQQmxkqtilX4+U9m5/wAl0CAwEAAaNCMEAwDwYDVR0TAQH/BAUwAwEB/zAOBgNV\nHQ8BAf8EBAMCAQYwHQYDVR0OBBYEFMSnsaR7LHH62+FLkHX/xBVghYkQMA0GCSqG\nSIb3DQEBBQUAA4IBAQCjGiybFwBcqR7uKGY3Or+Dxz9LwwmglSBd49lZRNI+DT69\nikugdB/OEIKcdBodfpga3csTS7MgROSR6cz8faXbauX+5v3gTt23ADq1cEmv8uXr\nAvHRAosZy5Q6XkjEGB5YGV8eAlrwDPGxrancWYaLbumR9YbK+rlmM6pZW87ipxZz\nR8srzJmwN0jP41ZL9c8PDHIyh8bwRLtTcm1D9SZImlJnt1ir/md2cXjbDaJWFBM5\nJDGFoqgCWjBH4d1QB7wCCZAA62RjYJsWvIjJEubSfZGL+T0yjWW06XyxV3bqxbYo\nOb8VZRzI9neWagqNdwvYkQsEjgfbKbYK7p2CNTUQ\n-----END CERTIFICATE-----\n";
```
Nokel81, line 28: I don't know if this is a better solution in general but I just thought that it might be a better idea to parse the PEM certs using something like https://www.npmjs.com/package/@fidm/x509 and then check the `validTo` and `validFrom` fields. 

That way any future certs that become expired won't be added.

Nokel81:
FYI there is now electron@v13.6.1

src/common/system-ca.ts
```diff
@@ -20,22 +20,93 @@
...
+ * Add (or merge) CAs to https.globalAgent.options.ca
+ */
+export function injectCAs (CAs: string[]) {
+  for (const cert of CAs) {
+    if (Array.isArray(https.globalAgent.options.ca) && !https.globalAgent.options.ca.includes(cert)) {
+      https.globalAgent.options.ca.push(cert);
+    } else {
+      https.globalAgent.options.ca = [cert];
+    }
   }
```
Nokel81, line 85: I think this might be clearer if this was the solution:

```ts
  const oldCAs = [https.globalAgent.options.ca].flat();
  const injectedCAs = new Set([...oldCAs, ...CAs]);

  https.globalAgent.options.ca = [...injectedCAs];
```

src/main/index.ts
chenhunghan: Where do you suggest? Or we should use the old "require at top" of renderer.ts/main.ts instead of export a function?

src/common/system-ca.ts
chenhunghan: Ok, sounds good

src/common/system-ca.ts
chenhunghan: I am not sure, need to do some research on this.

src/main/index.ts
jakolehm: I mean before `ready` event (after imports).

chenhunghan: OK. thanks

chenhunghan: Moved at the top (as before this PR) https://github.com/lensapp/lens/pull/4185/commits/aa7645347fe5cf25d2456f9f6a6a67e6cefd71c3

src/common/system-ca.ts
chenhunghan: Changed https://github.com/lensapp/lens/pull/4185/commits/ed7d9f4541df0a7995f8e7027e9c7a82687c771a

chenhunghan:
> FYI there is now electron@v13.6.1

Updated

jakolehm (dismissed):
LGTM

src/common/system-ca.ts
chenhunghan: This doesn't work as expected, getting this, I will revert.
```
[0]   rawException: TypeError [ERR_INVALID_ARG_TYPE]: The "options.ca" property must be of type string or an instance of Buffer, TypedArray, or DataView. Received undefined
[0]       at validateKeyOrCertOption (_tls_common.js:83:11)
[0]       at Object.createSecureContext (_tls_common.js:109:9)
[0]       at Object.connect (_tls_wrap.js:1580:48)
[0]       at Agent.createConnection (https.js:129:22)
[0]       at Agent.createSocket (_http_agent.js:323:26)
[0]       at Agent.addRequest (_http_agent.js:274:10)
[0]       at new ClientRequest (_http_client.js:306:16)
[0]       at Object.request (https.js:313:10)
[0]       at y._performRequest (/Users/hchen/.k8slens/extensions/lenscloud-lens-extension/dist/main.js:15:12465)
[0]       at new y (/Users/hchen/.k8slens/extensions/lenscloud-lens-extension/dist/main.js:15:8632) {
[0]     code: 'ERR_INVALID_ARG_TYPE'
[0]   }
[0] }
```

src/main/helm/helm-chart-manager.ts
chenhunghan: Let's do this on a separate PR.

src/common/system-ca.ts
chenhunghan: I guess `https.globalAgent.options.ca` is a TypedArray, the else may have bugs though..

jim-docker (approved):
test on windows in dev console succeeded:
```
var https = require('https'); https.request("https://k8slens.dev");
```

PR #4187, opened 2021-10-29
Fix NamespaceSelectFilter

Nokel81:
- Not displaying checks when "All Namespaces" was clicked

- Not toggling namespaces that are implicitly selected

## Before

https://user-images.githubusercontent.com/8225332/139443082-4c4916e2-f454-45a3-8ad8-bc7a7dc5629d.mov

## After

https://user-images.githubusercontent.com/8225332/139441866-cfb9f437-06ea-454e-aeb4-067d9724dff9.mov

### DIFF
--- a/src/renderer/components/+namespaces/namespace-select-filter.tsx
+++ b/src/renderer/components/+namespaces/namespace-select-filter.tsx
@@ -37,7 +37,7 @@ const Placeholder = observer((props: PlaceholderProps<any, boolean>) => {
   const getPlaceholder = (): React.ReactNode => {
     const namespaces = namespaceStore.contextNamespaces;
 
-    if (!namespaceStore.selectedNamespaces.length || !namespaces.length) {
+    if (namespaceStore.areAllSelectedImplicitly || !namespaces.length) {
       return <>All namespaces</>;
     }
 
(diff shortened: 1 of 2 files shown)


jim-docker (commented):
works as described, it's very nice to now be able to select all but one by selecting "All namespaces" and then toggling one off.

I find it weird that the dropdown list is reordered while I'm holding down the cmd key. I might like to toggle one off and then back on to see how the list of pods (say) changes but that's a bit cumbersome if it jumps to the bottom when I deselect it. Is there a reason/use case for doing sorting while the cmd key is held? (I think the sorting is ok when the dropdown first appears)

src/renderer/components/+namespaces/namespace-select-filter.tsx
```diff
@@ -37,7 +37,7 @@ const Placeholder = observer((props: PlaceholderProps<any, boolean>) => {
   const getPlaceholder = (): React.ReactNode => {
     const namespaces = namespaceStore.contextNamespaces;
 
-    if (!namespaceStore.selectedNamespaces.length || !namespaces.length) {
+    if (namespaceStore.areAllSelectedImplicitly || !namespaces.length) {
```
jim-docker, line 40: that's good, makes it clearer what we're testing for

Nokel81:
Good point about the sorting, I don't think there is a good reason for that. Will fix.

jim-docker:
> Good point about the sorting, I don't think there is a good reason for that. Will fix.

It's actually also cumbersome to reselect a bunch from the bottom, the list jumps up every time

Nokel81:
Done, PTALA @jim-docker

jim-docker (approved):

PR #4188, opened 2021-10-29
Use spec.serviceName as an externalIP if nothing else

### LINKED ISSUE #4179: Doesn't display externalName for kubernetes Service
**Describe the bug**
I'm using knative and we create a service of type `ExternalName`, when using `kubectl get svc` the column "External IP" displays the value of `externalName` but in Lens is empty, I expect the value to show up in Lens

**To Reproduce**
0. create a service with the following yaml
```
apiVersion: v1
kind: Service
metadata:
  name: my-service
spec:
  externalName: kourier-internal.kourier-system.svc.cluster.local
  type: ExternalName
```
1. Go to Network
2. Click on Services
4. See "External IP" empty

**Expected behavior**
To see the value in the column `External IP` same as `kubectl` output

**Screenshots**
image

- Lens Version: 5.3.5
- OS: OSX
- Installation method (e.g. snap or AppImage in Linux): Download dmg

Nokel81:

fixes #4179

### DIFF
--- a/src/common/k8s-api/endpoints/service.api.ts
+++ b/src/common/k8s-api/endpoints/service.api.ts
@@ -108,10 +108,6 @@ export class Service extends KubeObject {
       return this.spec.externalIPs;
     }
 
-    if (this.spec.externalName) {
-      return [this.spec.externalName];
-    }
-
     return [];
   }
 
--- a/src/common/k8s-api/endpoints/service.api.ts
+++ b/src/common/k8s-api/endpoints/service.api.ts
@@ -53,7 +53,7 @@ export interface Service {
     clusterIP: string;
     clusterIPs?: string[];
     externalTrafficPolicy?: string;
-    externalName: string;
+    externalName?: string;
     loadBalancerIP?: string;
     loadBalancerSourceRanges?: string[];
     sessionAffinity: string;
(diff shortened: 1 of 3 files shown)


src/common/k8s-api/endpoints/service.api.ts
```diff
@@ -100,11 +100,19 @@ export class Service extends KubeObject {
...
+    if (lb?.ingress) {
       return lb.ingress.map(val => val.ip || val.hostname);
     }
 
-    return this.spec.externalIPs || [];
+    if (Array.isArray(this.spec?.externalIPs)) {
+      return this.spec.externalIPs;
+    }
+
+    if (this.spec.externalName) {
```
jakolehm: Why do we need to add `externalName` to `getExternalIps` method? Isn't it a bit surprising to get externalName in some cases?

Nokel81: According to the issue and to https://kubernetes.io/docs/concepts/services-networking/service/#externalname the `externalName` field is meant to be used for when the service doesn't have a selector. `kubectl` does this display rule as well.

> An ExternalName Service is a special case of Service that does not have selectors and uses DNS names instead. For more information, see the ExternalName section later in this document.

jakolehm: I'm not against for displaying `externalName`. I just mean that including externalName to `getExternalIps` method is surprising.. couldn't we do this kind of display stuff somewhere closer to the UI components (not within api).

jakolehm (approved):

PR #4189, opened 2021-10-29
Fix: Catalog's visibility columns menu -> Empty title for Icon

ixrock:
**Before:**
Screenshot 2021-10-29 at 16 43 38

**After:**
Screenshot 2021-10-29 at 16 33 13

+ removed redundant `common/utils/makeCss`


Nokel81:
Shouldn't that visibility just be tied to the name column?

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -270,12 +268,12 @@ export class Catalog extends React.Component<Props> {
           entity => entity.searchFields,
         ]}
         renderTableHeader={[
-          { title: "", className: css.iconCell, id: "icon" },
-          { title: "Name", className: css.nameCell, sortBy: sortBy.name, id: "name" },
-          !activeCategory && { title: "Kind", className: css.kindCell, sortBy: sortBy.kind, id: "kind" },
-          { title: "Source", className: css.sourceCell, sortBy: sortBy.source, id: "source" },
-          { title: "Labels", className: css.labelsCell, id: "labels" },
-          { title: "Status", className: css.statusCell, sortBy: sortBy.status, id: "status" },
+          { title: "Icon", className: styles.iconCell, id: "icon" },
```
Nokel81: I think that this should be instead covered by `showWithColumn`

src/renderer/components/+catalog/catalog.module.css
```diff
@@ -19,43 +19,45 @@
...
-      right: 0;
-      opacity: 0;
-      transition: none;
-
-      &:hover {
-        /* Drop styles defined for <Icon/> */
-        background-color: transparent;
-        box-shadow: none;
-      }
+.Catalog {}
```
aleksfront: A leftover to cleanup.

src/renderer/components/+catalog/catalog.tsx
aleksfront: Yes we can do this with `showWithColumn`. Take a look at https://github.com/lensapp/lens/blob/master/src/renderer/components/+apps-helm-charts/helm-charts.tsx#L110

src/renderer/components/+catalog/catalog.module.css
ixrock: Well, not really. I intentionally left it here and linked to component's top element.
That way we can more safely define inner styles for global class names, e.g.:

```scss
.Catalog {
 :global(.TableRow){ ... }
}
```

Moreover I would prefer to have such top class-name for every React component with a style file.

src/renderer/components/+catalog/catalog.tsx
ixrock: `showWithColumn` I guess works differently.. It hides _another_ column together with a chosen.

Here I just added a title for icon's column in the "columns visibility menu".

Nokel81: Yes but I think that the icon is conceptually the same "column" as the name which is why I suggested the `showWithColumn` alternative.

src/renderer/components/+catalog/catalog-entity-item.tsx
```diff
@@ -26,12 +26,9 @@ import type { ItemObject } from "../../../common/item.store";
 import { Badge } from "../badge";
 import { navigation } from "../../navigation";
 import { searchUrlParam } from "../input";
-import { makeCss } from "../../../common/utils/makeCss";
 import { KubeObject } from "../../../common/k8s-api/kube-object";
 import type { CatalogEntityRegistry } from "../../api/catalog-entity-registry";
 
-const css = makeCss(styles);
```
ixrock, line 33: To support css-classes auto-completion use [vscode-css-modules](https://marketplace.visualstudio.com/items?itemName=clinyong.vscode-css-modules) plugin.

ixrock:
@Nokel81 @aleksfront PTALA

https://user-images.githubusercontent.com/6377066/140787193-a4328bc2-f8e4-4c4f-a3a3-f315a8ab668e.mov

_BTW:_ it seems hovering "pin-icon" currently broken in latest release (it doesn't show up)

Nokel81:
> BTW: it seems hovering "pin-icon" currently broken in latest release (it doesn't show up)

It does on master... and on my install of `alpha.9`...

Also from the video, it looks a bit weird to have "Name" not be over the name. What I mean it is looks like the icon is part of that column, which looks a bit strange.

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -214,23 +212,7 @@ export class Catalog extends React.Component<Props> {
...
-          tooltip={isItemInHotbar ? "Remove from Hotbar" : "Add to Hotbar"}
-          onClick={prevDefault(() => isItemInHotbar ? this.removeFromHotbar(item) : this.addToHotbar(item))}
-        />
-      </div>
-    );
-  }
-
-  renderIcon(item: CatalogEntityItem<CatalogEntity>) {
-    return (
-      <RenderDelay>
```
Nokel81, line 233: I don't think you should remove the `<RenderDelay>`, it was added so that the catalogue would show up faster.

ixrock: I haven't noticed big difference and in general having `RenderDelay` for plain icon looks weird / ugly workaround.. 

_Solution:_ refactor `HotbarIcon` / figure out why it's slow? Most probably something wrong with material-ui thingy used inside..

ixrock: @aleksfront wdyt? I suggest to merge this already and optimize `HotbarIcon` (if still needed)  in separated PR, maybe you can take it over if you'd like.

ixrock:
> Also from the video, it looks a bit weird to have "Name" not be over the name. What I mean it is looks like the icon is part of that column, which looks a bit strange.

@Nokel81 I'll add spacing in the header to align with name column text.

ixrock:
https://user-images.githubusercontent.com/6377066/140899670-ede22750-3077-4777-8a5b-6937b147dadd.mov

src/renderer/components/+catalog/catalog.module.css
```diff
@@ -19,43 +19,59 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-.list :global(.TableRow) {
```
Nokel81, line 22: Instead of all these changes, wouldn't it just have been easier to use `showWithColumn`?

ixrock: No. Because otherwise it' requires other hacks to hide text "Icon" in the header column or if we leave it empty as result we have empty menu-item for the table.

aleksfront (approved):
Tested, working good.

src/renderer/components/+catalog/catalog.tsx
aleksfront: I agree both. Removing `RenderDelay` is irrelevant to current PR, but `HotbarIcon` with material component should be refactored which is on todo. Let's proceed without `RenderDelay`.

src/renderer/components/+catalog/catalog.module.css
Nokel81: `showWithColumn` hides that menu item that is the point of it.

PR #4200, opened 2021-10-29
Lint: comma-spacing (error) & comma-dangle (error)

Nokel81:

### DIFF
--- a/src/common/hotbar-store.ts
+++ b/src/common/hotbar-store.ts
@@ -99,7 +99,7 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
   toJSON(): HotbarStoreModel {
     const model: HotbarStoreModel = {
       hotbars: this.hotbars,
-      activeHotbarId: this.activeHotbarId
+      activeHotbarId: this.activeHotbarId,
     };
 
     return toJS(model);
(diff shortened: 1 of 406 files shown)


src/common/hotbar-store.ts
```diff
@@ -187,7 +187,7 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
     } else if (0 <= cellIndex && cellIndex < hotbar.items.length) {
       hotbar.items[cellIndex] = newItem;
     } else {
-      logger.error(`[HOTBAR-STORE]: cannot pin entity to hotbar outside of index range`, { entityId: uid, hotbarId: hotbar.id, cellIndex, });
+      logger.error(`[HOTBAR-STORE]: cannot pin entity to hotbar outside of index range`, { entityId: uid, hotbarId: hotbar.id, cellIndex });
```
jim-docker, line 190: what's the argument for not having a comma here? (_So what_ if the specification is all on one line, or last element is on same line as closing character). If I'm putting a comma after every element it's easier for me if I don't have to remember any exceptions.

jim-docker (dismissed):

src/common/hotbar-store.ts
Nokel81: I made the specification of requiring commas be `"always-multiline"`. This means that the commas are required on multiline objects (and arrays) and forbidden from single line ones.

The common reason to enforce dangling commas is to make diffs smaller when adding or removing entries. However, if the object or array is all on the same line anyway that benefit is moot.

Finally, I didn't think it would be a good idea to require commas in the `[ "entry" ]` example.

jim-docker: I understand that the diffs benefit is moot but on the other hand the appearance difference is minimal if there's just one more comma for the single-line object. My point is now I have to remember when not to add the trailing comma instead of just thinking of the comma as an element terminator and always using it. Not a major point, I can handle it.  Just wondered if there was more of a reason than the moot diffs benefit.

Nokel81: If you have

```
"editor.codeActionsOnSave": {
    "source.fixAll.eslint": true
},
```

turned on in vscode then you shouldn't have to remember anything.

jim-docker: perfect, I just have to remember to do that.

jim-docker (approved):

PR #4201, opened 2021-10-29
Reduce selection on triple click in Drawers

### LINKED ISSUE #2200: Copying fields from overlay should not include newline

Nokel81:

fixes #2200 

This seems to work as intend. However, I don't quite know if there are some edge cases where it won't work.


aleksfront (approved):
LGTM

Issue #3272, opened 2021-07-05
Javascript error in the main process


galalmounir:
**Describe the bug**
With a pod selected, a random popup will appear with an uncaught exception

**Screenshots**
![image](https://user-images.githubusercontent.com/2583382/124501406-7a7ce780-dd8f-11eb-801c-225b4dc4bf00.png)

**Environment:**
- Lens Version: 5.0.1+20210705.1
- OS: MacOS

Nokel81:
Do you have write permissions for `~/.kube`?

galalmounir:
Yes I do. This only started happening after the upgrade to v5

Nokel81:
Okay thanks. I am trying to come up with a way to reproduce this issue on my machine. Because we don't try and read `config.lock` anywhere in our codebase I think it might be related to `kubectl` but I don't know enough about its internals yet.

Nokel81:
Were you doing something in particular when this popup happened? For instance were you opening a local terminal instance?

galalmounir:
I don't remember doing anything on the side. I'll try to repro again, but it keeps happening randomly.
Could it be related network connection/disconnection?

Nokel81:
I have done some more reading and it could just be a race condition between two instances of `kubectl` running against the same config file.

galalmounir:
Happened again this morning, this time my laptop was in sleep mode. Once I logged in, I found the same pop-up message.
After clicking `Ok` another message appeared:
![image](https://user-images.githubusercontent.com/2583382/125076769-502c7200-e08e-11eb-8987-5bbd74a40f75.png)

Nokel81:
That might be related to kube sync thanks for the report.

gdelbergue:
Having the same issue. It happens when I keep my computer running for night. Suddenly, fan speed increases, lens CPU usage is ~ 97% and there is a lock error.

![Screenshot 2021-07-14 at 11 01 49](https://user-images.githubusercontent.com/4040959/125597899-a105f946-f8e4-4fbe-a817-624712806a93.png)
<img width="980" alt="Screenshot 2021-07-14 at 11 19 14" src="https://user-images.githubusercontent.com/4040959/125597912-c79d2ca4-c690-439e-b5f2-49dc19567f21.png">

A click on "Ok" results then in this error:

[...]

PR #4216, opened 2021-11-01
Correctly handle promises and rejections around cluster connection

Nokel81:
- Remove duplicate log in LensProxy

- Cleanup constructor of KubeAuthProxy

---

I noticed this while trying to figure out why https://github.com/lensapp/lens/issues/3272 might be happening. I don't think this is a fix but I think it should be done anyway.

### DIFF
--- a/src/main/kubeconfig-manager.ts
+++ b/src/main/kubeconfig-manager.ts
@@ -30,57 +30,60 @@ import { LensProxy } from "./lens-proxy";
 import { AppPaths } from "../common/app-paths";
 
 export class KubeconfigManager {
-  protected configDir = AppPaths.get("temp");
-  protected tempFile: string = null;
+  /**
+   * The path to the temp config file
+   *
+   * - if `string` then path
+   * - if `null` then not yet created
+   * - if `undefined` then unlinked by calling `clear()`
+   */
+  protected tempFilePath: string | null | undefined = null;
 
   constructor(protected cluster: Cluster, protected contextHandler: ContextHandler) { }
 
+  /**
+   *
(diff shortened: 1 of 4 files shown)


src/main/kubeconfig-manager.ts
```diff
@@ -30,57 +30,60 @@ import { LensProxy } from "./lens-proxy";
...
-  protected async init() {
+  protected async ensureFile() {
     try {
       await this.contextHandler.ensureServer();
-      this.tempFile = await this.createProxyKubeconfig();
-    } catch (err) {
-      logger.error(`[KUBECONFIG-MANAGER]: Failed to created temp config for auth-proxy`, err);
+      this.tempFilePath = await this.createProxyKubeconfig();
+    } catch (error) {
+      throw Object.assign(new Error("Failed to creat temp config for auth-proxy"), { cause: error });
```
aleksfront, line 86: Why it's better than `logger.error`? And can we avoid modifying `Error` object instance?

aleksfront (approved):

src/main/kubeconfig-manager.ts
Nokel81: This is more correct because where `ensureFile()` is called, it assumes that `ensureFile()` always succeeds. This way any call to the `getPath()` above will also fail.

I don't think modifying the object like this is bad. Furthermore, `cause` is already natively supported in node 16.9 https://nodejs.org/en/blog/release/v16.9.0/?utm_source=ESnextNews.com&utm_medium=Weekly%20Newsletter&utm_campaign=2021-09-28#error-cause

PR #3910, opened 2021-09-29
MonacoEditor component fixes & refactoring

ixrock:
**PR highlights:**

1. refactored react component `<MonacoEditor/>`:

- detached from `dock.store.ts`
- removed `react-monaco-editor` since buggy and provide no-real benefits instead of direct access to `monaco-editor` package apis
- removed multiple redundant imports of theme-store & user-stores from places of editor usages
- added webpack plugin `monaco-editor-webpack-plugin` to properly bundle editor with the app
- faster/more reliable handling editor's dimensions with new [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) DOM api
- improved UX with editor's focus when interacting with `<Dock/>` (e.g. resizing/open/close) + open/close animation

2. Refactored `dock-tab.store.ts` _(every dock-tab is using editor in some way, this is why it's in current PR)_:
- predictable data initialization with `DockTabStore.init(): void` and `DockTabStore.whenReady(): Promise`
- removed possibly memory leaks because of forgotten `reaction` for sync tab-data with local-storage
- added methods `DockTabStore.dispose()` or `DockTabStore.destroy()` allows to fully destroy corresponding tab's data
3. detached dock-tab views registration from `dock.tsx`
4. added method-helpers for monitoring active tab state, see also `dock.store.ts`
-> `DockStore.onTabChange(callback, options?): IReactionDisposer`
-> `DockStore.onTabClose(callback, options?): IReactionDisposer`

**TODO / known bugs:**

- [x] terminal-tab background doesn't react to switching lens-theme to light/dark (see also: #2224)

**Troubleshooting:**

Q: editor's highlighting language syntax and  theme is not applied / broken
A: Try `make clean && make dev`

**Demo:**

https://user-images.githubusercontent.com/6377066/136702954-9d370424-1209-4b11-87ed-7f9ef72e5c2e.mov

**Before** _(Lens 5.2.5 + maybe latest `master`):_

- complete loosing focus to editor and restoring state (cursor position, selection, etc.) when switching between dock tabs
- yaml validation error is shared between same kind of tabs (see video) and resets only when value changed

https://user-images.githubusercontent.com/6377066/136702946-904044d6-3489-4185-b93f-d59468f52f1c.mov

PR #4225, opened 2021-11-02
Monaco editor refactoring

### LINKED ISSUE #4239: Not possible to write unit tests for components that import Monaco editor
**Describe the bug**

We weren't able to write unit tests for components that used some component that imported Monaco editor. Screenshot attached. 

**To Reproduce**
1. Create new unit test
2. Import eg. `kube-object-menu.tsx`
3. Try rendering the component (or even console.log)

**Expected behavior**
Shouldn't crash fatally.

**Screenshots**

**Environment**
- Lens Version: from master
- OS: OSX
- Running tests using `make test -- --watch`

Screen Shot 2021-11-03 at 12 44 20

ixrock:
based on #3910


integration/__tests__/cluster-pages.tests.ts
```diff
@@ -377,10 +377,8 @@ utils.describeIf(minikubeReady(TEST_NAMESPACE))("Minikube based tests", () => {
 
     await kubeApiServerRow.click();
     await frame.waitForSelector(".Drawer", { state: "visible" });
-
-    const logButton = await frame.waitForSelector("ul.KubeObjectMenu li.MenuItem i.Icon span[data-icon-name='subject']");
-
-    await logButton.click();
+    await frame.waitForSelector(".Drawer .drawer-title .Icon >> text=subject").then(button => button.click());
```
Nokel81: nit: `await frame.click(...);`

package.json
```diff
@@ -196,8 +196,9 @@
     "electron-devtools-installer": "^3.2.0",
     "electron-updater": "^4.6.1",
     "electron-window-state": "^5.0.3",
-    "filehound": "^1.17.4",
+    "filehound": "^1.17.5",
     "fs-extra": "^9.0.1",
+    "glob": "^7.2.0",
```
Nokel81: I don't think you use this dependency in this PR.

src/extensions/registries/__tests__/page-registry.test.ts
```diff
@@ -19,19 +19,18 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
+import { jest } from "@jest/globals";
```
Nokel81, line 22: Is this needed or is this a better practice then what we do now?

src/renderer/components/dock/dock.store.ts
```diff
@@ -89,6 +88,21 @@ export interface DockStorageState {
...
 
+export interface DockTabChangeEvent {
+  tab: DockTab;
+  tabId: TabId;
+  prevTab?: DockTab;
+}
+
+export interface DockTabChangeEventOptions extends IReactionOptions {
+  tabKind?: TabKind; // filter: by dockStore.selectedTab.kind == tabKind
+  dockIsVisible?: boolean; // filter: dock and selected tab should be visible (default: true)
```
Nokel81: Would be nice if those comments were doc comments instead.

src/renderer/components/dock/dock.tsx
```diff
@@ -105,7 +105,7 @@ export class Dock extends React.Component<Props> {
     if (!isOpen || !selectedTab) return null;
 
     return (
-      <div className="tab-content" style={{ flexBasis: height }}>
+      <div className="tab-content" style={{ flexBasis: isOpen ? height : 0 }}>
```
Nokel81: what is the point of this? if `isOpen` is false then the above `if` will `return null`

src/renderer/components/input/input.scss
```diff
@@ -63,6 +63,10 @@
     }
   }
 
+  input[type=number]::-webkit-inner-spin-button {
+    -webkit-appearance: none; // hide browser controls (top/bottom arrows)
+  }
```
Nokel81, line 68: do we really need to hide these?

src/renderer/components/monaco-editor/monaco-editor.tsx
```diff
@@ -0,0 +1,290 @@
...
+import { observer } from "mobx-react";
+import { editor, Uri } from "monaco-editor";
+import { MonacoTheme, registerCustomThemes } from "./monaco-themes";
+import { MonacoValidator, monacoValidators } from "./monaco-validators";
+import { cssNames, disposer, toJS } from "../../utils";
+import { UserStore } from "../../../common/user-store";
+import { ThemeStore } from "../../theme.store";
+import debounce from "lodash/debounce";
+
+registerCustomThemes(); // setup
```
Nokel81: Please call this in `bootstrap` and not here.

src/renderer/components/monaco-editor/monaco-editor.tsx
```diff
@@ -0,0 +1,290 @@
...
+  @observable unmounting = false;
+
+  constructor(props: MonacoEditorProps) {
+    super(props);
+    makeObservable(this);
+  }
+
+  @computed get model(): editor.ITextModel {
+    const uri = MonacoEditor.createUri(this.props.id ?? this.staticId);
+    const model = editor.getModels().find(model => model.uri.toString() == uri.toString());
```
Nokel81: Having investigated the code of monaco-editor. You can safely replace this with `editor.getModel(uri)`

src/renderer/components/monaco-editor/monaco-editor.tsx
```diff
@@ -0,0 +1,290 @@
...
+    const { language, value } = this.props;
+
+    return editor.createModel(value, language, uri);
+  }
+
+  @computed get options(): editor.IStandaloneEditorConstructionOptions {
+    return toJS({
+      ...UserStore.getInstance().editorConfiguration,
+      ...(this.props.options ?? {}),
+    });
```
Nokel81: Should we instead do a deep merge? I don't know which would be less surprising down the line.

src/renderer/components/monaco-editor/monaco-editor.tsx
```diff
@@ -0,0 +1,290 @@
...
+    this.props.onModelChange?.(model, oldModel);
+    this.validateLazy();
+  };
+
+  /**
+   * Save current view-model state in the editor.
+   * This will allow restore cursor position, selected text, etc.
+   * @param {editor.ITextModel} model
+   */
+  saveViewState(model = this.model) {
```
Nokel81: I would just remove this default param since there is exactly 2 uses of this method so I think is will cause more confusion then keeping it.

src/renderer/components/monaco-editor/monaco-editor.tsx
```diff
@@ -0,0 +1,290 @@
...
+    if (!model) return;
+
+    MonacoEditor.viewStates.set(model.uri, this.editor.saveViewState());
+  }
+
+  restoreViewState(model = this.model) {
```
Nokel81: same for this.

src/renderer/components/monaco-editor/monaco-editor.tsx
```diff
@@ -0,0 +1,290 @@
...
+      model: this.model,
+      detectIndentation: false, // allow `option.tabSize` to use custom number of spaces for [Tab]
+      value: defaultValue,
+      language,
+      theme,
+      readOnly,
+      ...this.options,
+    });
+
+    this.logger?.info(`[MONACO]: editor created for language=${language}, theme=${theme}`);
```
Nokel81: I think that having `id` would be useful as well in all of these log lines. Maybe having a getter for it too.

src/renderer/components/monaco-editor/monaco-theme.clouds-midnight.json
```diff
@@ -1,4 +1,5 @@
 {
```
Nokel81: I think it would be better if we kept the folder `monaco-themes` instead of prefixing the files.

webpack.renderer.ts
```diff
@@ -87,6 +88,13 @@ export function webpackLensRenderer({ showVars = true } = {}): webpack.Configura
     },
     module: {
       rules: [
+        {
+          // Allow to load raw file contents with dynamic imports, e.g. import("../some-path.txt?raw")
+          // See also: https://webpack.js.org/guides/asset-modules/
+          use: "raw-loader",
+          resourceQuery: /raw/,
+          enforce: "post",
+        },
```
Nokel81: I don't think we use this in this PR.

src/renderer/components/monaco-editor/monaco-editor.tsx
```diff
@@ -0,0 +1,290 @@
...
+
+  focus() {
+    this.editor?.focus();
+  }
+
+  @action
+  validate = (value = this.getValue()) => {
+    const validators: MonacoValidator[] = [
+      monacoValidators[this.props.language], // parsing syntax check
+    ].filter(Boolean);
```
Nokel81, line 271: For future work, should we consider using the monaco language servers instead? That way we could get inline errors.

Nokel81 (commented):
Thanks for splitting this up. That was much easier to review. Only minor things.

src/renderer/components/monaco-editor/monaco-editor.tsx
ixrock: Do you mean something from this list?

- https://langserver.org/
- https://github.com/TypeFox/monaco-languageclient

Once we decide to support it we remove current basic validation in the editor.

webpack.renderer.ts
ixrock: This is fine.

src/renderer/components/monaco-editor/monaco-theme.clouds-midnight.json
ixrock: Okay, but in general I prefer to have some some key text in a filename to make search by files easier in project..

webpack.renderer.ts
ixrock: I don't understand your frenetic desire to restrict merging _every single line of code_  if non-relevant to PR/topic..
Why so strict? What the purpose? Clear history? Blah.. It doesn't help with bugs / don't provide any benefit to end  users..

integration/__tests__/cluster-pages.tests.ts
ixrock: Did you mean `(await frame.waitForSelector(".Drawer .drawer-title .Icon >> text=subject")).click()`?

package.json
ixrock: Removed

src/extensions/registries/__tests__/page-registry.test.ts
ixrock: Yes, it's better practice to keep it explicitly + autocomplete in editors usually works better that way.

src/renderer/components/dock/dock.tsx
ixrock: Well, previously it was done for animation when closing Dock.. Here doesn't work as intended, true.

src/renderer/components/input/input.scss
ixrock: Screenshot 2021-11-03 at 11 28 33

src/renderer/components/dock/dock.store.ts
ixrock: Updated

src/renderer/components/monaco-editor/monaco-editor.tsx
ixrock: Done

src/renderer/components/monaco-editor/monaco-editor.tsx
ixrock: Yes, probably it's a good idea.

src/renderer/components/monaco-editor/monaco-editor.tsx
ixrock: Well I don't see any confusion in having current model as default param, but okay..

src/renderer/components/dock/dock.tsx
ixrock: Removed.

src/renderer/components/monaco-editor/monaco-editor.tsx
ixrock: Added computed `logMetadata: {id, model}` to every logger's output.

webpack.renderer.ts
nevalla: I agree with @Nokel81 that we should not do include in any changes that are not necessary for the PR. Making separate PR will make it faster to get these changes in. If this PR does not work without this change, then this change should belong to this PR.

ixrock: Thanks for your opinions. Removed.

src/renderer/components/input/input.scss
Nokel81: ok

ixrock:
@Nokel81 @jim-docker PTALA

Nokel81 (approved):

jim-docker:
> @Nokel81 @jim-docker PTALA

you guys are too fast for me. I did test it out, and will test a little more, looks great, thanks @ixrock!

PR #4228, opened 2021-11-02
Display resource load errors seperately from list of items

### LINKED ISSUE #2683: Custom resources are also fetched at cluster scope when filtering by namespaces, causing "Failed to load items" to be displayed
**Describe the bug**
When opening the view for a Custom Resource and filtering by namespace, Lens tries to retrieve Custom Resources not only at namespace scope (e.g. `GET /api-kube/apis/bitnami.com/v1alpha1/namespaces/<namespace>/sealedsecrets`), but also at cluster scope (e.g. `GET /api-kube/apis/bitnami.com/v1alpha1/sealedsecrets`). Since retrieving them at cluster scope is not allowed in our Cluster, the request is answered with a 403 Forbidden and Lens displays "Failed to load items", although the request at namespace scope was processed successfully.

Furthermore, when the request at namespace scope is processed I can see the retrieved resources in the view flashing for a very short time and then being replaced by "Failed to load items", since then the request at cluster scope fails.

**To Reproduce**
1. Have Custom Resources not gettable/listable at cluster scope, but at namespace scope
2. Don't use "accessible namespaces" feature and have namespaces listable at cluster scope
3. Open Developer Tools and open the View for a custom resource, while filtering by a namespace
4. In the Network tab of the Dev Tools you can see, that one request (namespace scope) is successful and endless requests are tried at cluster scope without success

**Expected behavior**
Only retrieve custom resources at namespace scope when the namespace filter is applied.

**Screenshots**
n/a

- Lens Version: 4.2.3
- OS: Win 10 Pro Version 1909 (Build 18363.1377)
- Installation method: Setup .exe

**Logs:**
How would I find Logs on Windows? The Logs I found under `C:\Users\<user>\AppData\Roaming\Lens\Lens\logs` are almost empty and have no relevant information regarding this issue

**Kubeconfig:**
```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: <certificate-authority-data>
    server: <server>
  name: aks-test
contexts:
- context:
    cluster: aks-test
    user: clusterUser_rg-aks-test_aks-test
  name: aks-test
current-context: aks-test
kind: Config
preferences: {}
users:
- name: clusterUser_rg-aks-test_aks-test
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - get-token
      - --environment
      - AzurePublicCloud
      - --server-id
      - <server-id>
      - --client-id
      - <client-id>
      - --tenant-id
      - <tenant-id>
      command: kubelogin
      env: null
      provideClusterInfo: false

```
fy: using kubelogin: https://github.com/Azure/kubelogin

**Additional context**
n/a

### LINKED ISSUE #4192: One inaccessible namespace blocks resource list for all namespaces
If a namespace that is not actually accessible for a user is added to the accessible namespaces list then, when filtering a resource listing, if that namespace is selected the resource list shows "Failed to load items", even if other, accessible namespaces are selected as well.

https://user-images.githubusercontent.com/40840436/139463368-d95b2622-94a9-49bb-8f7b-e13a991606a5.mov

Expected behaviour: show resources for all selected namespaces that are accessible, post a notification or some error indicator that says "Failed to load items" for any namespaces that aren't accessible.

Nokel81:

fixes #4192 
fixes #2683
Screen Shot 2021-11-02 at 2 57 10 PM

### DIFF
--- a/src/common/k8s-api/kube-watch-api.ts
+++ b/src/common/k8s-api/kube-watch-api.ts
@@ -60,7 +60,7 @@ interface KubeWatchPreloadOptions {
 
 export interface KubeWatchSubscribeStoreOptions extends KubeWatchPreloadOptions {
   /**
-   * Whether to subsribe only after loading all stores
+   * Whether to subscribe only after loading all stores
    * @default true
    */
   waitUntilLoaded?: boolean;
--- a/src/common/k8s-api/kube-watch-api.ts
+++ b/src/common/k8s-api/kube-watch-api.ts
@@ -27,7 +27,7 @@ import type { ClusterContext } from "./cluster-context";
 
 import plimit from "p-limit";
 import { comparer, observable, reaction, makeObservable } from "mobx";
-import { autoBind, Disposer, noop } from "../utils";
+import { autoBind, disposer, Disposer, noop } from "../utils";
 import type { KubeApi } from "./kube-api";
 import type { KubeJsonApiData } from "./kube-json-api";
 import { isDebugging, isProduction } from "../vars";
@@ -38,11 +38,38 @@ export interface IKubeWatchEvent<T extends KubeJsonApiData> {
   object?: T;
 }
 
-export interface IKubeWatchSubscribeStoreOptions {
-  namespaces?: string[]; // default: all accessible namespaces
-  preload?: boolean; // preload store items, default: true
-  waitUntilLoaded?: boolean; // subscribe only after loading all stores, default: true
-  loadOnce?: boolean; // check store.isLoaded to skip loading if done already, default: false
+interface KubeWatchPreloadOptions {
+  /**
+   * The namespaces to watch
+   * @default all-accessible
+   */
+  namespaces?: string[];
+
+  /**
+   * Whether to skip loading if the store is already loaded
+   * @default false
+   */
+  loadOnce?: boolean;
+
+  /**
+   * A function that is called when listing fails. If set then blocks errors
+   * being rejected with
+   */
+  onLoadFailure?: (err: any) => void;
+}
+
+export interface KubeWatchSubscribeStoreOptions extends KubeWatchPreloadOptions {
+  /**
+   * Whether to subsribe only after loading all stores
+   * @default true
+   */
+  waitUntilLoaded?: boolean;
+
+  /**
+   * Whether to preload the stores before watching
+   * @default true
+   */
+  preload?: boolean;
 }
 
 export interface IKubeWatchLog {
@@ -63,15 +90,15 @@ export class KubeWatchApi {
     return Boolean(this.context?.cluster.isAllowedResource(api.kind));
   }
 
-  preloadStores(stores: KubeObjectStore<KubeObject>[], opts: { namespaces?: string[], loadOnce?: boolean } = {}) {
+  preloadStores(stores: KubeObjectStore<KubeObject>[], { loadOnce, namespaces, onLoadFailure }: KubeWatchPreloadOptions = {}) {
     const limitRequests = plimit(1); // load stores one by one to allow quick skipping when fast clicking btw pages
     const preloading: Promise<any>[] = [];
 
     for (const store of stores) {
       preloading.push(limitRequests(async () => {
-        if (store.isLoaded && opts.loadOnce) return; // skip
+        if (store.isLoaded && loadOnce) return; // skip
 
-        return store.loadAll({ namespaces: opts.namespaces });
+        return store.loadAll({ namespaces, onLoadFailure });
       }));
     }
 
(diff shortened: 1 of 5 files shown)


src/common/k8s-api/kube-watch-api.ts
```diff
@@ -38,11 +38,38 @@ export interface IKubeWatchEvent<T extends KubeJsonApiData> {
...
+  /**
+   * A function that is called when listing fails. If set then blocks errors
+   * being rejected with
+   */
+  onLoadFailure?: (err: any) => void;
+}
+
+export interface KubeWatchSubscribeStoreOptions extends KubeWatchPreloadOptions {
+  /**
+   * Whether to subsribe only after loading all stores
```
jim-docker:
```suggestion
   * Whether to subscribe only after loading all stores
```

src/renderer/components/kube-object-list-layout/kube-object-list-layout.tsx
```diff
@@ -91,6 +101,14 @@ export class KubeObjectListLayout<K extends KubeObject> extends React.Component<
               ...searchProps,
               placeholder: `Search ${placeholderString}...`,
             },
+            info: (
+              <>
+                {info}
+                {this.loadErrors.length > 0 && (
+                  <Icon material="warning" className="load-error" tooltip={this.loadErrors.join("\n")}/>
```
jim-docker: it looks like the newline is not respected when the tooltip is displayed

Screen Shot 2021-11-10 at 8 05 03 PM

jim-docker (commented):
I like the UI implementation!

Doesn't always work for All namespaces:

https://user-images.githubusercontent.com/40840436/141219343-8764e32e-fae3-48cb-8637-3b9d72a93f8d.mov

(user has access to only the default namespace)

and a more minor point, the warning icon flashes as you cmd-click to add/remove namespaces from the selection.

jim-docker (commented):
Not sure if this is due to latest commit or was previously there as an edge case:

Screen Shot 2021-11-15 at 10 50 05 AM

This is seen upon first connection to the cluster, and "All namespaces" is the setting that was loaded from local storage (user can access default namespace only, and can list namespaces. The list should show the pods in the default namespace)

jim-docker:
Workloads Overview page also suffers similar problem:

https://user-images.githubusercontent.com/40840436/141814542-48f5e96d-400d-43ec-8a26-fb34f826d544.mov

maybe this should be a separate issue/PR?

Nokel81:
> maybe this should be a separate issue/PR?

sure

jim-docker:
> > maybe this should be a separate issue/PR?
> 
> sure

#4344

jakolehm:
Warning icon placement looks a bit off to me, would close to `X item` make more sense?

Nokel81:
> Warning icon placement looks a bit off to me, would close to `X item` make more sense?

I initially, during development, had it there but thought it looked worse because it was so large and has a higher contrast level than the text.

Also, it is related to the namespace selector because that is where the user picks where to load from.

Nokel81:
> This is seen upon first connection to the cluster, and "All namespaces" is the setting that was loaded from local storage (user can access default namespace only, and can list namespaces. The list should show the pods in the default namespace)

If it has been saved to local storage then this is working as intended. Given that the user has selected "All namespaces" (explicitly) we load from cluster scope.

If you think that we should fall back to listing individually in this case, maybe. But I don't think that should be part of the PR.

jim-docker:
> > This is seen upon first connection to the cluster, and "All namespaces" is the setting that was loaded from local storage (user can access default namespace only, and can list namespaces. The list should show the pods in the default namespace)
> 
> If it has been saved to local storage then this is working as intended. Given that the user has selected "All namespaces" (explicitly) we load from cluster scope.
> 
> If you think that we should fall back to listing individually in this case, maybe. But I don't think that should be part of the PR.

But how does this work?

https://user-images.githubusercontent.com/40840436/141849682-0f5986f6-4249-4cf2-a140-3afab8c5cc75.mov

jim-docker:
Another minor point, It might be nice if the warning tooltip didn't obscure the namespace selector. Can it be lowered and still look appropriately associated with the warning icon?

Nokel81:
> Another minor point, It might be nice if the warning tooltip didn't obscure the namespace selector. Can it be lowered and still look appropriately associated with the warning icon?

We can move it to any of the corners or any of the sides surround the mount point. Do you think that left or bellow would be better?

jim-docker:
> > Another minor point, It might be nice if the warning tooltip didn't obscure the namespace selector. Can it be lowered and still look appropriately associated with the warning icon?
> 
> We can move it to any of the corners or any of the sides surround the mount point. Do you think that left or bellow would be better?

I like below, as long as it doesn't look "orphaned". Going to the left is left for when there's not enough room to the right, right?

Nokel81:
> > > This is seen upon first connection to the cluster, and "All namespaces" is the setting that was loaded from local storage (user can access default namespace only, and can list namespaces. The list should show the pods in the default namespace)
> > 
> > 
> > If it has been saved to local storage then this is working as intended. Given that the user has selected "All namespaces" (explicitly) we load from cluster scope.
> > If you think that we should fall back to listing individually in this case, maybe. But I don't think that should be part of the PR.
> 
> But how does this work?
> Screen.Recording.2021-11-15.at.3.27.37.PM.mov

That "works" because we don't clear the store on load failure, the items are still around from the last load. Do you think they should be cleared?

Nokel81:
Screen Shot 2021-11-16 at 9 21 35 AM

src/renderer/components/tooltip/tooltip.tsx
```diff
@@ -112,26 +112,21 @@ export class Tooltip extends React.Component<TooltipProps> {
...
+
+    // Then add the default ordering, these won't override the above
+    positions.add(TooltipPosition.RIGHT);
+    positions.add(TooltipPosition.BOTTOM);
+    positions.add(TooltipPosition.TOP);
+    positions.add(TooltipPosition.LEFT);
+    positions.add(TooltipPosition.TOP_RIGHT);
+    positions.add(TooltipPosition.TOP_LEFT);
+    positions.add(TooltipPosition.BOTTOM_RIGHT);
+    positions.add(TooltipPosition.BOTTOM_LEFT);
```
jim-docker: So this kind of refactoring looks like an improvement but makes your work take longer, the review take longer, testing take longer, impacts more code than this PR is intended for, and could introduce new bugs...

Nokel81: Yes you are right, I shouldn't have committed that, will revert.

jim-docker (approved):

Issue #2680, opened 2021-04-29
Number of Pods running on Node Lists


alexmontoanelli:
**What would you like to be added**:
On node lists, add a collum that show how many pods are running on that node.

**Why is this needed**:
Better view of running pods on all nodes

**Environment you are Lens application on:**

- Kubernetes distribution: default kubernetes
- Desktop OS: MAC OS 11

jim-docker:
This is desirable but will be complicated by larger clusters, which may affect performance

trombini77:
> This is desirable but will be complicated by larger clusters, which may affect performance

I see that this feature come together with the fact, por example, depending node type in AWS it has a limit of pods (see https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt)  into this menu this value makes all sense.

It's sad to hear that this feature may affect larger clusters performances, so would be an option a flag somewhere (Configurations menu or in UI Node Menu) to enable or disable this metric?

PR #4244, opened 2021-11-03
Correctly catch and notify user about NodeShell pod failures

### LINKED ISSUE #2687: Detect and notify when shell cannot be launched on node due to pod limit

Currently Lens provides an option to run a shell on a node, which it does by launching a privileged pod on the node. On my non-production clusters, this frequently fails due to the node having already reached its pod limit, so there is no room for an additional pod. Currently when this happens there is no feedback from Lens about it, Lens just waits for the pod to come up, which it never does. 

I would like Lens to detect this condition, provide a notification of some kind, and not attempt to create the pod in this case. 

It would also be helpful to have some indication of pod count and limit on the "Nodes" display, as is done with CPU, Memory, and Disk.

**Why is this needed**:

Typically I can run a shell on any node in the cluster to accomplish my task, but when they are mostly full due to pod count, it is a lot of trial and error to find an available node, and in the meantime, Lens' failing pods are somewhat disruptive to the cluster.

- Desktop OS: macOS

Nokel81:

fixes #2687

### DIFF
--- a/src/main/lens-proxy.ts
+++ b/src/main/lens-proxy.ts
@@ -80,7 +80,7 @@ export class LensProxy extends Singleton {
         const cluster = getClusterForRequest(req);
 
         if (!cluster) {
-          return void logger.error(`[LENS-PROXY]: Could not find cluster for upgrade request`);
+          return void logger.error(`[LENS-PROXY]: Could not find cluster for upgrade request from url=${req.url}`);
         }
 
         const reqHandler = isInternal ? shellApiRequest : kubeApiRequest;
--- a/src/main/lens-proxy.ts
+++ b/src/main/lens-proxy.ts
@@ -60,12 +60,10 @@ export class LensProxy extends Singleton {
 
   public port: number;
 
-  constructor(protected router: Router, functions: LensProxyFunctions) {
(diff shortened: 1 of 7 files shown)


aleksfront:
Should we create a new issue out of this (from #2687)
> It would also be helpful to have some indication of pod count and limit on the "Nodes" display, as is done with CPU, Memory, and Disk.

Because the issue actually contains 2 requests and we're fixing here only one.

Nokel81:
We already have that https://github.com/lensapp/lens/issues/2680

jim-docker:
Waiting > 10 mins:
Screen Shot 2021-11-04 at 4 42 10 PM

I have 110 running pods and I try to open a node-shell. It should timeout?

Nokel81:
It should have given you a message and then closed itself. It did when I was testing it myself, will try again...

Nokel81:
@jim-docker I think that is a different bug... weird though, because I cannot see where `this.xterm` would be `undefined` at that point will investigate.

Nokel81:
@jim-docker That should be fixed now.

jim-docker (approved):

PR #4256, opened 2021-11-04
Fix auto update not working for AppImage

### LINKED ISSUE #3767: Linux AppImage: old version restarts after update
**Describe the bug**
On linux using the app, after clicking "Yes" on the update notification, the app automatically quits and is restarted but it it is the previous version that restarts.

**To Reproduce**
1. `./Lens-5.1.3-latest.20210722.1.x86_64.AppImage`
2. wait for update notification
3. click "Yes"
4. observe old version restarts
5. note that the new AppImage file has been downloaded (and the previous version file is gone) as expected

- Lens Version: 5.2.0 (perhaps any 5.x)
- OS: Linux
- Installation method: AppImage

Nokel81:
- Emulate electron.autoUpdater's 'before-quit-for-update' event flow

fixes #3767

### DIFF
--- a/src/main/app-updater.ts
+++ b/src/main/app-updater.ts
@@ -49,6 +49,12 @@ function handleAutoUpdateBackChannel(event: Electron.IpcMainEvent, ...[arg]: Upd
            */
           electronAutoUpdater.checkForUpdates();
         } else if (isLinux) {
+          /**
+           * This is a necessary workaround until electron-updater is fixed.
+           * The problem is that because linux updating is not implemented at
+           * all via electron. Electron's autoUpdater.quitAndInstall() is never
+           * called.
+           */
           electronAutoUpdater.emit("before-quit-for-update");
         }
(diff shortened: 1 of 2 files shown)


src/main/app-updater.ts
```diff
@@ -48,6 +48,8 @@ function handleAutoUpdateBackChannel(event: Electron.IpcMainEvent, ...[arg]: Upd
            * download it from itself via electron.
            */
           electronAutoUpdater.checkForUpdates();
+        } else if (isLinux) {
+          electronAutoUpdater.emit("before-quit-for-update");
```
jim-docker: nit: I guess the above comment should apply to this too

jim-docker (dismissed):

src/main/app-updater.ts
Nokel81: Sort of, the first line yes but the reason is different.

jim-docker (approved):

PR #4258, opened 2021-11-04
Keep shell processes alive between network offline periods

### LINKED ISSUE #4250: Terminal process dies if network goes offline
**Describe the bug**

Lens kills terminal process whenever network goes offline (for example when laptop sleeps). This is really annoying because Lens resets whole terminal tab (process is killed, working directory is different etc).

**Expected behavior**

Lens should keep terminal process running and reconnect to that specific process when computer resumes from sleep (or from offline).

Nokel81:
- Allows long running processes in the lens terminal to servive the
  computer going to sleep

fixes #4250

### DIFF
--- a/src/main/shell-session/shell-session.ts
+++ b/src/main/shell-session/shell-session.ts
@@ -31,6 +31,7 @@ import { isWindows } from "../../common/vars";
 import { UserStore } from "../../common/user-store";
 import * as pty from "node-pty";
 import { appEventBus } from "../../common/event-bus";
+import logger from "../logger";
 
 export class ShellOpenError extends Error {
   constructor(message: string, public cause: Error) {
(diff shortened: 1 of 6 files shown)


src/main/shell-session/shell-session.ts
```diff
@@ -40,41 +41,140 @@ export class ShellOpenError extends Error {
   }
 }
 
+export enum WebSocketCloseEvent {
```
jakolehm, line 44: Not sure if we need all events defined (documented) here because we are just using few of them?

Nokel81: I decided to include all of them so that were could use the name and have some docs for them in our code.

I don't see the point of only including a few of them in an enumeration but if you think it would be better to just document the uses I can revert.

I prefer enums over commented (and especially uncommented) special values.

src/main/shell-session/shell-session.ts
```diff
@@ -40,41 +41,140 @@ export class ShellOpenError extends Error {
...
+
+  constructor(protected websocket: WebSocket, protected cluster: Cluster, terminalId: string) {
     this.kubectl = new Kubectl(cluster.version);
     this.kubeconfigPathP = this.cluster.getProxyKubeconfigPath();
     this.kubectlBinDirP = this.kubectl.binDir();
+    this.terminalId = `${cluster.id}:${terminalId}`;
   }
 
-  protected async open(shell: string, args: string[], env: Record<string, any>) {
+  protected async _open(shell: string, args: string[], env: Record<string, any>) {
```
jakolehm: Why this is called `_open`?

Nokel81: I renamed it because it is more of a helper function to the child classes `open()` methods. Since they both don't take any args but this one does. I felt that it made it clearer instead of overloading. 

Maybe I should add some docs to this method.

jakolehm: Ok, perhaps we could consider other name (like `openShellProcess`) for this helper.

jakolehm (dismissed):

jakolehm (approved):

PR #4266, opened 2021-11-05
Improve cluster connection visibility

### LINKED ISSUE #4260: Lens frequently stuck at "Authentication proxy started"
**Describe the bug**
When trying to connect to a cluster, sometimes Lens get stuck at the _connecting_ screen.

This bug is not deterministic (AFAICT) - sometimes it works, sometimes it gets stuck. I can confirm that when the issue arises, I'm still able to consult the cluster using `kubectl`. 

**To Reproduce**
Try to connect to a cluster.

**Expected behavior**
The cluster screen shows up.

**Screenshots**
image

- Lens Version: 5.2.5-latest.20211001.2
- OS: Windows 10
scoop.sh

Nokel81:

fixes #4260

---

This PR improves the visibility of what Lens is doing under the hood to the user so that if the connection process is taking longer than expected the user knows that Lens is still trying to do something instead of being stuck:

https://user-images.githubusercontent.com/8225332/140523405-1e322c6e-f6f3-4bc6-97eb-af84bcd71785.mov

### DIFF
--- a/src/common/cluster-store.ts
+++ b/src/common/cluster-store.ts
@@ -26,7 +26,7 @@ import { Cluster } from "../main/cluster";
 import migrations from "../migrations/cluster-store";
 import logger from "../main/logger";
 import { appEventBus } from "./event-bus";
-import { ipcMainHandle, ipcMainOn, ipcRendererOn, requestMain } from "./ipc";
+import { ipcMainHandle, requestMain } from "./ipc";
 import { disposer, toJS } from "./utils";
 import type { ClusterModel, ClusterId, ClusterState } from "./cluster-types";
 
@@ -37,8 +37,6 @@ export interface ClusterStoreModel {
 const initialStates = "cluster:states";
 
 export class ClusterStore extends BaseStore<ClusterStoreModel> {
-  private static StateChannel = "cluster:state";
-
   clusters = observable.map<ClusterId, Cluster>();
 
   protected disposer = disposer();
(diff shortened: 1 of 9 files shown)


src/common/cluster-store.ts
```diff
@@ -83,21 +81,13 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
...
+    const ipc = ipcMain ?? ipcRenderer;
 
-    if (ipcMain) {
-      this.disposer.push(ipcMainOn(ClusterStore.StateChannel, this.handleStateChange));
-    }
-
-    if (ipcRenderer) {
-      this.disposer.push(ipcRendererOn(ClusterStore.StateChannel, this.handleStateChange));
-    }
+    ipc?.on("cluster:state", (event, clusterId: ClusterId, state: ClusterState) => {
```
ixrock, line 88: Why `ipc?.`? It might not exist?

src/renderer/components/cluster-manager/cluster-status.tsx
```diff
@@ -94,58 +92,68 @@ export class ClusterStatus extends React.Component<Props> {
...
+            className="box center interactive"
+            onClick={this.manageProxySettings}
+          >
+            Manage Proxy Settings
+          </a>
+        </>
+      );
+    }
+
+    return undefined;
```
ixrock, line 145: Why it's needed? Maybe just define function to return `:void`?

ixrock (approved):

src/common/cluster-store.ts
Nokel81: It might not exist in testing. Could be wrong though

src/renderer/components/cluster-manager/cluster-status.tsx
Nokel81: Because the function doesn't just return void, it also returns a ReactNode

PR #4270, opened 2021-11-05
'About Port Forwarding' notification

jim-docker:
Guide the user to the port-forwarding "dashboard" when they've added their first port-forward

https://user-images.githubusercontent.com/40840436/140533507-012cd53c-df0c-4b57-a7fb-bf3a4bffcd22.mov

enhances #2340

### DIFF
--- a/src/renderer/components/+network-services/service-port-component.tsx
+++ b/src/renderer/components/+network-services/service-port-component.tsx
@@ -28,7 +28,7 @@ import { observable, makeObservable, reaction } from "mobx";
 import { cssNames } from "../../utils";
 import { Notifications } from "../notifications";
 import { Button } from "../button";
-import { addPortForward, getPortForward, openPortForward, PortForwardDialog, portForwardStore, predictProtocol, removePortForward } from "../../port-forward";
+import { aboutPortForwarding, addPortForward, getPortForward, getPortForwards, openPortForward, PortForwardDialog, portForwardStore, predictProtocol, removePortForward } from "../../port-forward";
 import type { ForwardedPort } from "../../port-forward";
 import { Spinner } from "../spinner";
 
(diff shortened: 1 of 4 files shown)


src/renderer/components/+network-services/service-port-component.tsx
```diff
@@ -92,13 +92,21 @@ export class ServicePortComponent extends React.Component<Props> {
 
     this.waiting = true;
 
+    // determine how many port-forwards are already active
+    const { length } = await getPortForwards();
```
Nokel81: shouldn't this be within the `try`?

Nokel81 (dismissed):
Nit: looks good. I think I agree that introducing a new endpoint just for this is a bit overkill for the time being.

src/renderer/components/+network-services/service-port-component.tsx
jim-docker: good point

src/renderer/port-forward/port-forward-notify.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+import React from "react";
+import { portForwardsURL } from "../../common/routes/port-forwards";
+import { Button } from "../components/button";
+import { Notifications, notificationsStore } from "../components/notifications";
+import { navigate } from "../navigation";
+import { getHostedClusterId } from "../utils";
+
+
+export async function aboutPortForwarding() {
```
chenhunghan: Does this need to be async?

jim-docker: I wondered that myself. Probably not, thanks

chenhunghan (dismissed):
LGTM

Nokel81 (approved):

Issue #3918, opened 2021-09-30
breadcrumb when deleting any resource


marafa-sugarcrm:
**What would you like to be added**:
when deleting a kind like a pod or a deployment, have the confirmation box show the full breadcrumb like `<cluster>/<namespace>/<kind>/<name>`. currently it only shows `<name>`

**Why is this needed**:
UX - with so many clusters and namespaces, it is useful to double check the breadcrumb in one area instead of looking all over the lens app screen to verify 

**Environment you are Lens application on:** 
all

- Kubernetes distribution: [EKS, AKS, GKE, Rancher, ...]
all

- Desktop OS: <the OS Lens app runs on in your case>
mac os big sur

marafa-sugarcrm:
@jansav thank you!

PR #4282, opened 2021-11-08
Add breadcrumbs to removal of kube objects

jansav:
Contains proposal to improve quality of codebase while implementing a trivial feature. It's a tool to make communication of difficult topics easier. Please read commits from start to finish as they describe a chain of thought.

Why?
- Because global shared state is bad.
- Because compliance to SOLID is a good way to improve design and quality overall.
- Because currently unit testing in codebase is too difficult and lacks infrastructure. Having unit tests enables faster development, fewer bugs, ability to refactor and confident deployments etc.

What was done?
- We implemented a trivial feature.
- In said feature, we replaced global shared state with dependencies.
- We implemented “local” unit tests which benefit from said dependencies that can now be stubbed.
- We implemented naive, staged, way of injecting a component with dependencies in place. This was done quick-to-dirty to better communicate a difficult topic with an example.
- We tidied up the dependencies to be minimal for the intended purpose, ie. compliance to Interface Segregation Principle.
- We introduced snapshot-testing for UI.
- We introduced library “asyncFn” to make notoriously difficult async unit testing easier.
- We introduced shared configuration for WebStorm users.

Guidelines:
- Solution must be simple and not scary.
- Solution must comply to well known patterns.
- Solution must be adaptable in small increments.

Talking points:
- Using test fakes vs. stubs vs. actuals as dependencies in unit tests.
- What is the current state of codebase?

Coders will do good to study:
- SOLID-principles and clean code.
- Team’s common agreed vocabulary for unit testing jargon.

Disclaimer:
- Our generics/TypeScript is a bit rusty.
- By decision, the unit testing as done here is *local* instead of *behavioral*. This is not something we can do forever, as coupling small unit tests with very specific technical details will in the long run hinder ability to refactor instead of enabling it. This decision was done to introduce first steps of a strategy, that can later grow to benefit from said behaviors.

Closing #3918


.idea/.gitignore
```diff
@@ -0,0 +1,6 @@
+# Default ignored files
```
jakolehm: Please split all `.idea/` files into separate PR.

extensions/kube-object-event-status/package-lock.json
```diff
@@ -16,9 +16,9 @@
       },
       "dependencies": {
         "@babel/runtime": {
-          "version": "7.14.0",
-          "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.14.0.tgz",
-          "integrity": "sha512-JELkvo/DlpNdJ7dlyw/eY7E0suy5i5GQH+Vlxaq1nsNJ+H7f4Vtv3jMeCEgRhZZQFXTjldYfQgv2qmM6M1v5wA==",
+          "version": "7.16.0",
```
jakolehm: Idk why package-lock files within extensions were changed but these changes should not be part of this PR.

src/renderer/components/kube-object-menu/inject-naive/inject-naive.tsx
```diff
@@ -0,0 +1,54 @@
...
+import type { KubeObjectMenuDependencies } from "../kube-object-menu";
+import type { KubeObject } from "../../../../common/k8s-api/kube-object";
+
+interface Props {
+  Component: React.ElementType<any>
+  [key: string]: any,
+}
+
+export const InjectNaive = ({ Component, ...props }: Props) => {
+  const kubeObjectMenuRegistry = KubeObjectMenuRegistry.getInstance();
```
jakolehm: Not sure if I understand why we need `InjectNaive`, couldn't `KubeObjectMenu` have default props which could be overridden?

extensions/kube-object-event-status/package-lock.json
jansav: True, removed.

.idea/.gitignore
jansav: #4293

src/renderer/components/kube-object-menu/inject-naive/inject-naive.tsx
jansav: `InjectNaive` is there to represent the abstraction that should be created between components and it's dependencies. We have seen that usage of default props will limit our choices in the future. Good example would be that dependency is asynchronous.

Of course, in this first step, the `InjectNaive` is just hard-coded solution to get rid of shared global state in UI component itself. In future, we need find a way to inject those dependencies from somewhere else.

.gitignore
```diff
@@ -17,3 +17,5 @@ types/extension-renderer-api.d.ts
 extensions/*/dist
 docs/extensions/api
 site/
+
```
chenhunghan: [nit] Do we need this empty line?

src/renderer/components/kube-object-menu/kube-object-menu.test.tsx
```diff
@@ -0,0 +1,295 @@
...
+          uid: "some-uid",
+          name: "some-name",
+          resourceVersion: "some-resource-version",
+          namespace: "some-namespace",
+        },
+      });
+
+      removeActionMock = asyncFn();
+
+      ({ baseElement, getByTestId, queryByTestId } = render(
```
chenhunghan: I think nowadays they recommend to use `screen` for queries (https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen) e.g. `screen.getByTestId`, so you probably don't need to have `let getByTestId` but better use `screen.getByTestId()` in each block.

src/renderer/components/kube-object-menu/kube-object-menu.test.tsx
```diff
@@ -0,0 +1,295 @@
...
+        </div>,
+      ));
+    });
+
+    it("renders", () => {
+      expect(baseElement).toMatchSnapshot();
+    });
+
+    it("does not open a confirmation dialog yet", () => {
+      expect(queryByTestId("confirmation-dialog")).toBeNull();
```
chenhunghan: See above comments.
```suggestion
      expect(screen.queryByTestId("confirmation-dialog")).toBeNull();
```

src/renderer/components/kube-object-menu/kube-object-menu.test.tsx
```diff
@@ -0,0 +1,295 @@
...
+        dependencies={dependencies}
+      />,
+    );
+
+    expect(baseElement).toMatchSnapshot();
+  });
+
+  describe("given kube object", () => {
+    let baseElement: Element;
+    let removeActionMock: any;
```
chenhunghan: [nit] Better typing
```suggestion
    let removeActionMock: ReturnType<typeof asyncFn>;
```

src/renderer/components/kube-object-menu/kube-object-menu.test.tsx
```diff
@@ -0,0 +1,295 @@
...
+  let editResourceTabStub: () => void;
+  let apiManagerStub: IGettableStore;
+  let kubeObjectMenuRegistry: KubeObjectMenuRegistry;
+  let objectStub: KubeObject | null;
+  let dependencies: KubeObjectMenuDependencies<KubeObject>;
+
+  beforeEach(() => {
+    // TODO: Remove illegal global overwrites for what should be a dependency somewhere.
+    // TODO: Remove usage of experimental browser API.
+    window.requestIdleCallback = (callback: IdleRequestCallback): number => {
```
chenhunghan: Do we need to reset this in `afterAll` or `afterEach`?

src/renderer/initializers/kube-object-detail-registry.tsx
```diff
@@ -54,7 +54,7 @@ import { ReplicaSetDetails } from "../components/+workloads-replicasets";
 import { StatefulSetDetails } from "../components/+workloads-statefulsets";
 import type { KubeObjectDetailsProps } from "../components/kube-object-details";
 
-export function intiKubeObjectDetailRegistry() {
+export function initKubeObjectDetailRegistry() {
```
nevalla: Good catch 👍  Although this could/should be in separate PR, since it's out of scope of this PR and probably would land to `master` faster.

src/renderer/components/kube-object-menu/kube-object-menu.test.tsx
jansav: Good catch. Resolved

src/renderer/components/kube-object-menu/kube-object-menu.test.tsx
jansav: Nice catch. :)

Nokel81:
Is this PR ready for review our still a draft?

jansav:
> Is this PR ready for review our still a draft?

`InjectNaive` part is not ready yet. It's just dirty hard-coded version for now. However, I suggest you to check the PR still so we can talk about the proposed changes.

Nokel81:
Will do when I have time later in the day

src/renderer/components/kube-object-menu/kube-object-menu.test.tsx
jansav: Good question. In perfect world we shouldn't need to do that but not sure at the moment. So this is something that we need to verify.

src/renderer/initializers/kube-object-detail-registry.tsx
jansav: Well, true. I don't think that it's needed here. :) It will get there eventually.

.gitignore
```diff
@@ -17,3 +17,4 @@ types/extension-renderer-api.d.ts
 extensions/*/dist
 docs/extensions/api
 site/
+.prettierrc
```
Nokel81: We have been using `eslint` as our linter and formatter in this repo, so it seems strange to add prettier

src/renderer/components/kube-object-menu/inject-naive/inject-naive.tsx
Nokel81: How could a dependency be asynchronous? React has a synchronous rendering pipeline.

src/renderer/components/kube-object-menu/inject-naive/inject-naive.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { editResourceTab } from "../../dock/edit-resource.store";
+import { hideDetails } from "../../kube-detail-params";
+import { apiManager } from "../../../../common/k8s-api/api-manager";
+import { KubeObjectMenuRegistry } from "../../../../extensions/registries";
+import { catalogEntityRegistry } from "../../../api/catalog-entity-registry";
+
+
+interface Props {
+  Component: React.ReactType<any>
+  [key: string]: any,
```
Nokel81: Also I really don't line this lack of type safety

src/common/k8s-api/api-manager.ts
```diff
@@ -27,7 +27,11 @@ import type { KubeApi } from "./kube-api";
 import type { KubeObject } from "./kube-object";
 import { IKubeObjectRef, parseKubeApi, createKubeApiURL } from "./kube-api-parse";
 
-export class ApiManager {
+export interface IGettableStore {
```
Nokel81: Personally I really dislike the `I` prefixing of interfaces. 

Is it also possible to have better names for these interfaces, currently they seem too generic.

src/common/k8s-api/api-manager.ts
```diff
@@ -27,7 +27,11 @@ import type { KubeApi } from "./kube-api";
...
+  getStore<TKubeObjectStore extends KubeObjectStore<KubeObject>>(api: string | KubeApi<KubeObject>): TKubeObjectStore | undefined;
```
Nokel81: @jakolehm I recently stumbled across the following rule https://github.com/microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md

Would it be a good idea to slowly move towards that?

.gitignore
jansav: That's the reason that it's listed in `.gitignore`.  Prettier does lot more than `eslint` and I'm trying to find a way to use it to format code without messing up the existing code so that those who want to utilize it, can do so.

src/renderer/components/kube-object-menu/inject-naive/inject-naive.tsx
jansav: Well yeah, of course. But as said, `InjectNaive` is just hard-coded solution which is not ready to be merged. :) It's here just to pass the dependencies to `KubeObjectMenu` and therefore not scalable.

src/renderer/components/kube-object-menu/inject-naive/inject-naive.tsx
jansav: Relatively easily. Of course, in React prop should always be synchronous, but with little help of MobX, we can observe the status of a promise and render based on that. You can only imagine the possibilities that it opens. 

However, since it's only one of the benefits, I think it's too deep discussion to have here :) Quick huddle would be great so that we can discuss it further. 

Our experience says, that creating abstraction here opens a lot of improvements that will make the component itself lot simpler.

Iku-turso: Indeed, historically asynchronous dependencies have been a wonderful tool to simplify some questionable code here and there. Also the separation of concerns implemented here rubs me the good way.

Let me partake in this huddle, ey guys? :)

src/renderer/initializers/kube-object-detail-registry.tsx
Iku-turso: Hmm. It could just be better to keep the overhead for trivial cleanup / boy scouting at minimum. It won't strain the PR if a reviewer simply skips/skims commits that are clearly trivial. And when it comes to clean-up, I would choose quantity over fast-tracking to `master`.

.gitignore
Iku-turso: I agree with @jansav, current `eslint` configuration is a tad suboptimal, and causes manual work to keep code reasonably formatted. If optional usage of `prettier` even complies to `eslint`, then why not go for it?

src/common/k8s-api/api-manager.ts
Iku-turso: Makes sense to me :)

src/renderer/components/kube-object-menu/inject-naive/inject-naive.tsx
Nokel81: A huddle might be a good idea. But thinking about this more I still don't see why leveraging default props isn't a better design or at least just as good. That way we don't introduce any new types

jansav: One thing that comes my mind immediately is that when we get unit testing infrastructure to correct level, we need to start writing unit tests to larger scope to support refactoring in scale. 

That being said, it would cause the issue that you need to provide huge amount of dependencies as props from top level to component levels. 

But yeah, let's schedule huddle and go through it asap.

Nokel81: That is why I think we should also leverage default props, so that the users of a component don't need to change

jansav: How would you handle following situation?

**You want to create unit test for a component which uses other components that have dependencies that need to be stubbed.**

For example:

- Render page in unit test
- One of the component in page has dependency that gets data from an API call
- You want to be able to control the result of said API call and see what effect it has on the page

Disclaimer: 
- API call is just an example and it's stupid one but could be anything that need to be stubbed in test
- jest.doMock works, but performance is horrible and it will cause issues when you have large test suite.
- Shared global state is bad and shouldn't be used here

To solve this, you would need to pass the dependency from unit test to top level (page) component to child component, but if there is an abstraction for injecting said dependency, you can just stub the dependency using the abstraction and all places that inject the dependency will have exactly same instance of it. This makes our unit test suite much more believable and robust.

Nokel81: I don't know if I buy the robustness argument since the effect is the same, and is believability really a benefit if you are just hiding where you inject the code?

jansav: Not hiding of course. As said, it was not ready when we opened the draft for discussion. We are preparing one example how it could be done :)

chenhunghan: One option to "inject dependencies" is to use custom renderer https://testing-library.com/docs/react-testing-library/setup/#custom-render

There are some real world examples in lenscloud/frontend/__test__. (Maybe not the best one...).

I think it's better to stop using defaultProps, React team seems to plan to retire it for functional components at some point in a major release.

jansav: Check the proposed solution in #4338.

jansav:
Flakiness of unit tests in Ubuntu is caused by #4354. Solving in separate PR.

src/renderer/components/kube-object-menu/inject-naive/inject-naive.tsx
Iku-turso: Type safety is now in place: #4338.

src/common/k8s-api/api-manager.ts
Iku-turso: On `I`-prefixing: Together with `T`-prefixing, this has many times proven valuable tool in making some more difficult typings understandable. Also it would be good to have the code somehow look off, when a coder is supposed to use an interface instead of concretion. With this, we have a better chance of making *Inversion of Control* and *loose coupling* happen on team level :)

On interface naming: Agreed. New interfaces added to better communicate and document intent.

Nokel81: I disagree, would you please provide some source for your claim? I find them distracting. Items should be obvious what they are from how they are used.

Iku-turso: Which claim? :) A general goal here is to make stuff more obvious. Lack of Inversion of Control makes code less obvious by being a tangle of concrete dependencies. I would say that anything that could help in fixing this is a welcome thing to be tried.

Nokel81: This claim:

> On I-prefixing: Together with T-prefixing, this has many times proven valuable tool in making some more difficult typings understandable.

panuhorsmalahti: I'd go with the commonly used convention in TypeScript projects, e.g. not using the `I`-prefix.

jweak: I tend to agree. At least have only one way throughout the codebase. Either prefix all or nothing.

Iku-turso: @panuhorsmalahti let's not make a vote out this, as best rationale should win. What is yours? Convention for sake of convention does not count :)

Nokel81: I have found that not having such prefixes limits the space of possible names so that name collisions happens more often within badly designed code.

panuhorsmalahti: For code style questions commonly used conventions do matter, but are not the only factor. Following style that is used in, say, 90% of projects, makes it easier for new developers to jump in. Using something else requires extra arguments (that have been provided here). In any case, my own opinion is that the `I`-prefix makes the code less aesthetic and makes the code slightly more awkward to read and type. Prefixes do have a (minor) cost, otherwise we could add them to every single thing freely.

Iku-turso: @Nokel81 
1. For example, we use T and I as key pieces of information when talking difficult subjects with other coders.

2. Also there was my other point:
> Also it would be good to have the code somehow look off, when a coder is supposed to use an interface instead of concretion. With this, we have a better chance of making Inversion of Control and loose coupling happen on team level :)

3. Here's a particularly dodgy piece of typing which debatably benefits from prefixes more than not:
```javascript
export interface IDependencyInjectionContainer {
    inject: <
      TInjectable extends IInjectable<TInstance, TDependencies>,
      TInstance,
      TDependencies,
    >(
      injectableKey: TInjectable,
    ) => ReturnType<TInjectable["instantiate"]>;
  }
```

jansav: I think that we have common goal here: be able to separate interfaces from actual implementations. 

That being said, for me, it doesn't matter what is the solution to achieve that, but I haven't heard any better alternatives than prefixing them with something. Interfaces and implementations are separate for a reason, so how do you think we should separate them? Currently it's not so clear.

Nokel81: We should separate them by name, but we don't need to "namespace" the interfaces with prefixes.

jansav: @Nokel81 Could you provide example? :)

Iku-turso: @Nokel81 but what about the 3 arguments? :)

Iku-turso: @panuhorsmalahti I agree, yet still find the other arguments more pressing.

After all, we are trying to make a big thing happen here, with Inversion of Control and all. Expect more eggs to be broken :)

Nokel81: 1. When taking it is quite easy to say "an interface" or "a class/implementation" called "Foo", when working in an IDE, there is hover intellisense and "goto type definition" and such.
2. I would need to an example as to why prefixing helps. The general rule of "consume interfaces, provide implementations" doesn't need any naming conventions to be realized. 
3. Because of how typescript type parameters work all those prefixes are unnecessary, `T extends F` means that `F` must be some sort of interface, and the fact that `T` is used as a parameter type means that it must be an implementation.

Iku-turso: @jweak we covered your point IRL. Nice to meet you! :)

Iku-turso: @Nokel81

> When taking it is quite easy to say "an interface" or "a class/implementation" called "Foo"

That's too much overhead to be practical.

> when working in an IDE, there is hover intellisense and "goto type definition" and such.

One shouldn't be required to hover stuff to make code easier to read in this way.

> The general rule of "consume interfaces, provide implementations" doesn't need any naming conventions to be realized.

The good rule shouldn't need the conventions, but it sure does benefit from them, as transgressions get easier to spot.

> Because of how typescript type parameters work all those prefixes are unnecessary, `T extends F` means that `F` must be some sort of interface, and the fact that `T` is used as a parameter type means that it must be an implementation.

This comes with additional cognitive load, making it a losing bargain.

Nokel81: > That's too much overhead to be practical.

I disagree, saying the prefix is more cumbersome as it won't ever flow with the rest of the name

> One shouldn't be required to hover stuff to make code easier to read in this way.

Why not, having too much information at a time can make code hard to read so having an optional way of getting more information when necessary is very useful.

> The good rule shouldn't need the conventions, but it sure does benefit from them, as transgressions get easier to spot.

Maybe but that doesn't rule out the fact that this convention in particular (adding predefined prefixes and/or suffixes) is an eyesore. Beautiful code should be readable and understandable without this crutch.

This convention naturally leads to doing what C# and Java do with `FooImpl implements IFoo` which just is terrible.

> This comes with additional cognitive load, making it a losing bargain.

How is more cognitive load then the convention? The code reader still has to understand what is going on

jakolehm: Most of the existing interfaces are without `I` prefix and we cannot really change those without breaking extension api types. So my recommendation is to use that convention for now. IMHO main thing that matters here is to have a convention.

Iku-turso: @jakolehm That is a pressing argument! I might know some work-arounds, but I don't have enough visibility at this time.

Complying to convention and moving on! :)

jansav:
Keeping this as draft until #4338 is merged.

jansav:
Up to date again. Noticed that `KubeObjectMenu` is exposed through Extension API so had to get rid of changes to components external API to preserve backwards compatibility.

src/extensions/registries/kube-object-menu-registry.ts
```diff
@@ -32,10 +32,10 @@ export interface KubeObjectMenuRegistration {
   components: KubeObjectMenuComponents;
 }
 
+
 export class KubeObjectMenuRegistry extends BaseRegistry<KubeObjectMenuRegistration> {
```
Nokel81: I don't think this change is required for this change.

src/renderer/components/kube-object-menu/index.ts
```diff
@@ -19,4 +19,5 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-export * from "./kube-object-menu";
+export type { KubeObjectMenuProps } from "./kube-object-menu";
+export * from "./kube-object-menu-container";
```
Nokel81: I think we might as well do `export { KubeObjectMenu } ...`

src/renderer/components/kube-object-menu/kube-object-menu.tsx
```diff
@@ -53,61 +65,61 @@ export class KubeObjectMenu<T extends KubeObject> extends React.Component<KubeOb
...
+    this.props.hideDetails();
     const { object, removeAction } = this.props;
 
     if (removeAction) await removeAction();
     else await this.store.remove(object);
   }
 
   @boundMethod
-  renderRemoveMessage() {
+  getRemoveMessage() {
```
Nokel81: This is technically a breaking change

src/renderer/components/kube-object-menu/kube-object-menu.tsx
```diff
@@ -53,61 +65,61 @@ export class KubeObjectMenu<T extends KubeObject> extends React.Component<KubeOb
...
-          object={object}
-          key={`menu-item-${index}`}
-          toolbar={toolbar}
-        />
-      ));
+    return this.props.kubeObjectMenuItems.map((MenuItem, index) => (
+      <MenuItem
+        object={object}
+        toolbar={toolbar}
+        // TODO: Fix misuse of index in key
```
Nokel81: Isn't this only a misuse if the items will be reordered?

src/renderer/components/menu/menu-actions.tsx
```diff
@@ -125,32 +124,31 @@ export class MenuActions extends React.Component<MenuActionsProps> {
     return (
       <>
         {this.renderTriggerIcon()}
-        <RenderDelay>
```
Nokel81, line 128: @aleksfront Wasn't this rendering delay added on purposes to help with initial load times?

jansav: We discussed with @aleksfront about the requirement of `RenderDelay` and came to decision that it's not needed. We didn't see any impact on initial load times.

src/renderer/components/kube-object-menu/kube-object-menu.tsx
jansav: I would avoid using index if UID is available as a convention but your argument is valid, so I removed the comment.

src/renderer/components/kube-object-menu/kube-object-menu.tsx
jansav: Hmm, technically maybe. 

Someone might extend the component and try accessing old method (or override it). 
Can you do it? Yes. Should you do it? No. 

Our ability to refactor anything is kinda limited, if we decide that we want to keep supporting that kind of use case. If so, I think that we should mark this component as `@deprecated` and create completely new version of `KubeObjectMenu`.

What's your opinion?

src/renderer/components/kube-object-menu/index.ts
jansav: Done

src/extensions/registries/kube-object-menu-registry.ts
jansav: Reverted changes in this file.

src/renderer/components/kube-object-menu/kube-object-menu.tsx
jansav: Side note: 

We could use some sort of mechanism to guard the Extension API from sudden changes such as this one. Currently there needs to be too much manual care and convention to be sure about conservation of the API. 

How about flurry of interfaces that communicate stake in Extension API? That way we would know which parts of code base are open for healthy refactoring.

Nokel81: I completely agree that it is currently hard to tell what the extension API both "is" and what is allowed to change.

> Our ability to refactor anything is kinda limited, if we decide that we want to keep supporting that kind of use case. If so, I think that we should mark this component as @deprecated and create completely new version of KubeObjectMenu.
> 
> What's your opinion?

I don't think we need to for this PR, looking at the change I don't actually see why you want to change the method name either. Our standard practice is to name method `render*` if they return a `React.ReactNode` which is what this one does.

jansav: The naming might be sub-optimal imho, but breaking Extension API is unacceptable. Reverted.

src/renderer/components/kube-object-menu/kube-object-menu.test.tsx
```diff
@@ -0,0 +1,299 @@
...
+import { TabKind } from "../dock/dock.store";
+import kubeObjectMenuRegistryInjectable from "./dependencies/kube-object-menu-items/kube-object-menu-registry.injectable";
+import { renderFor, DiRender } from "../test-utils/renderFor";
+import type { Cluster } from "../../../main/cluster";
+import type { ApiManager } from "../../../common/k8s-api/api-manager";
+import apiManagerInjectable from "./dependencies/api-manager.injectable";
+import { KubeObjectMenu } from "./index";
+
+describe("kube-object-menu", () => {
+  let objectStub: KubeObject | null;
```
Nokel81: nit, I don't think this needs to be here. All the uses are contained within each function call

jansav: Good catch, simplified.

Nokel81 (dismissed):

jansav:
@Nokel81 mind approving again, had to rebase branch to fix DCOs

Nokel81 (approved):

PR #4285, opened 2021-11-08
Block shell websocket connections from non-lens renderers

Nokel81:

fixes https://github.com/lensapp/lens-security/issues/1

### DIFF
--- a/src/main/proxy-functions/shell-api-request.ts
+++ b/src/main/proxy-functions/shell-api-request.ts
@@ -19,24 +19,65 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import url from "url";
 import logger from "../logger";
 import { Server as WebSocketServer } from "ws";
 import { NodeShellSession, LocalShellSession } from "../shell-session";
 import type { ProxyApiRequestArgs } from "./types";
 import { ClusterManager } from "../cluster-manager";
-import { appName, appSemVer } from "../../common/vars";
-import { LensProxy } from "../lens-proxy";
+import URLParse from "url-parse";
+import { ExtendedMap, Singleton } from "../../common/utils";
+import type { ClusterId } from "../../common/cluster-types";
+import { ipcMainHandle } from "../../common/ipc";
+import * as uuid from "uuid";
+
+export class ShellRequestAuthenticator extends Singleton {
+  private tokens = new ExtendedMap<ClusterId, Map<string, string>>();
+
+  init() {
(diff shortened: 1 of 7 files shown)


src/main/proxy-functions/shell-api-request.ts
```diff
@@ -19,29 +19,38 @@
...
 
-  ws.on("connection", ((socket: WebSocket, req: http.IncomingMessage) => {
-    const cluster = ClusterManager.getInstance().getClusterForRequest(req);
+  if (
+    socket.remoteAddress !== "127.0.0.1"
+    || !req.headers["user-agent"]?.includes(` ${appName}/${appSemVer} `)
+    || !req.headers.origin?.endsWith(`.localhost:${LensProxy.getInstance().port}`)
+    || !cluster
+  ) {
+    socket.write("Invalid shell request");
```
jim-docker, line 81: can this throw?

Nokel81: No, the documentation states that this function will return `true` if all of the buffer (or string in this case) has been flushed to kernel memory or `false` if at least some of it was buffered in user memory. 

And then it goes on to talk about not writing more to a socket after `.write()` returns false until the `"drain"` event is emitted for other security reasons, but since I am only writing once I don't think that matters.

src/main/proxy-functions/shell-api-request.ts
```diff
@@ -19,29 +19,38 @@
...
-  const ws = new WebSocket.Server({ noServer: true });
+export function shellApiRequest({ req, socket, head }: ProxyApiRequestArgs): void {
+  const cluster = ClusterManager.getInstance().getClusterForRequest(req);
 
-  ws.on("connection", ((socket: WebSocket, req: http.IncomingMessage) => {
-    const cluster = ClusterManager.getInstance().getClusterForRequest(req);
+  if (
+    socket.remoteAddress !== "127.0.0.1"
+    || !req.headers["user-agent"]?.includes(` ${appName}/${appSemVer} `)
+    || !req.headers.origin?.endsWith(`.localhost:${LensProxy.getInstance().port}`)
```
jim-docker: can't these be faked?

Nokel81: Yes it can, but at least browsers cannot be forced to modify the origin. I guess we could set up a secret requirement that is requested via IPC.

Nokel81: Going to change this to do authentication instead.

jim-docker (approved):
confirmed on mac and linux

PR #4297, opened 2021-11-09
Make ws:// authentication more timing safe

Nokel81:
- Switch to crypto.timingSafeEqual() and Uint8Array (because of the
  crypto function)

follow up to #4285 

I think this is timing safe, but I don't know if the `throw new TypeError(...)`'s are.

### DIFF
--- a/src/main/proxy-functions/shell-api-request.ts
+++ b/src/main/proxy-functions/shell-api-request.ts
@@ -28,14 +28,17 @@ import URLParse from "url-parse";
 import { ExtendedMap, Singleton } from "../../common/utils";
 import type { ClusterId } from "../../common/cluster-types";
 import { ipcMainHandle } from "../../common/ipc";
-import * as uuid from "uuid";
+import crypto from "crypto";
+import { promisify } from "util";
+
+const randomBytes = promisify(crypto.randomBytes);
 
 export class ShellRequestAuthenticator extends Singleton {
-  private tokens = new ExtendedMap<ClusterId, Map<string, string>>();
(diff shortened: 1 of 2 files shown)


src/main/proxy-functions/shell-api-request.ts
```diff
@@ -60,9 +63,17 @@ export class ShellRequestAuthenticator extends Singleton {
     }
 
     const authToken = clusterTokens.get(tabId);
+    const buf = Uint8Array.from(token.split(",").map(ar => {
+      const res = +ar;
+
+      if (isNaN(res)) {
+        throw new TypeError("Token must be a CSV list of digits");
```
jim-docker: could the token come in as base64 instead? Could be cleaner to handle?

Nokel81: Probably, let me see

Nokel81: Done

src/main/proxy-functions/shell-api-request.ts
```diff
@@ -63,15 +63,7 @@ export class ShellRequestAuthenticator extends Singleton {
...
-    const buf = Uint8Array.from(token.split(",").map(ar => {
-      const res = +ar;
-
-      if (isNaN(res)) {
-        throw new TypeError("Token must be a CSV list of digits");
-      }
-
-      return res;
-    }));
+    const buf = Uint8Array.from(Buffer.from(token, "base64"));
```
jim-docker, line 66: I guess if this fails it's not a problem. Can it throw?

jim-docker (dismissed):

jim-docker (approved):

PR #4302, opened 2021-11-09
Force scrollable behaviour for all label columns

### LINKED ISSUE #4300: Namespace labels are cut off
**Describe the bug**

If a namespace contains many labels (or long labels), they get cut off weirdly in the UI. The last label will be cut off and won't have a background. There is also a strangely placed ellipsis.

**To Reproduce**
1. Create a namespace with several labels.
2. Look at the namespace in the table in Lens.

**Screenshots**

image

Here the namespace labels are:
- app.kubernetes.io/name=ingress-nginx
- app.kubernetes.io/part-of=ingress-nginx
- istio-injection=enabled

- Lens Version: 5.2.6-latest.20211104.1
- OS: macos 11.4
- Installation method (e.g. snap or AppImage in Linux): https://docs.k8slens.dev/main/getting-started/#macos

Nokel81:

fixes #4300


src/renderer/components/badge/badge.module.css
```diff
@@ -36,6 +36,18 @@
   text-overflow: ellipsis;
 }
 
+:global(.ItemListLayout) {
```
ixrock, line 39: Seems like this is wrong place for customizing badge styles and should be moved to `item-list-layout.scss`.
`Badge` component is more generic and probably shouldn't know about `ItemListLayout` at all.

Nokel81: Fair enough, will investigate how to use a module css from within a non-module

ixrock: Or just convert `item-list-layout.scss` to css-module?

Nokel81: That could work, thanks

Nokel81: @ixrock Do you know now, even after converting to `css-module`, it would be possible to get this to work? Given that `badge.module.css` is its own local scope, I don't actually think that it is possible to do this without `.ItemListLayout` being global. 

I am having trouble finding documentation on how this might be done.

ixrock: PTAL https://github.com/lensapp/lens/pull/4358

Nokel81:
Closing due to other PR

PR #4305, opened 2021-11-09
Close issues with 'needs-information' after 60 days

Nokel81:


jim-docker (approved):
LGTM. Will test it live?

Nokel81:
I think so, there aren't that many issues that have the `needs-information` label so we shouldn't be spammed.

PR #4316, opened 2021-11-10
Cleanup Prometheus detection

Nokel81:
- Only try each type once

- Log all attempts only if they all fail


nevalla (approved):
LGTM

PR #4317, opened 2021-11-10
Use common/logger everywhere, remove main/logger export

Nokel81:
- in Renderer make it 'console' which a stub for silly level

- in Main make it a winston logger with file output as well

- in testing make it a jest mock stub

- in all other cases make it a winston logger without file output


jakolehm (commented):
Please remove (or split to another PR) all unrelated changes.

Nokel81:
@jakolehm I have undone the changes to `download_kubtctl.ts`, I think that is the only unrelated change.

jakolehm (commented):
Should this PR also add `no-console` eslint rule?

src/common/logger.ts
```diff
@@ -19,12 +19,10 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { app, ipcMain } from "electron";
+import { app, ipcRenderer } from "electron";
 import winston, { format } from "winston";
-import type Transport from "winston-transport";
 import { consoleFormat } from "winston-console-format";
 import { isDebugging, isTestEnv } from "./vars";
-import BrowserConsole from "winston-transport-browserconsole";
```
jakolehm, line 27: Could you clarify why we are removing winston completely from renderer?

src/common/logger.ts
```diff
@@ -34,10 +32,37 @@ const logLevel = process.env.LOG_LEVEL
...
+    return {
+      ...console,
+      // eslint-disable-next-line no-console
+      silly: (...args: any[]) => console.debug(...args),
+    };
+  }
 
-if (ipcMain) {
-  transports.push(
+  if (isTestEnv) {
```
jakolehm: This feels a bit odd because it relies on a global state... would it be better to have it as a function arg?

Nokel81: As a function argument to what? I could add a testing function that could be called to mock out the logger, is that what you mean?

src/common/logger.ts
Nokel81: Sure, there are several reasons:

1. Winston does not interact nicely with DevTools and the verbosity selector
2. "winston-transport-browserconsole" has a bug where any "meta-data" after the second entry is ignored (which I believe is intentional)

Nokel81:
Closing this as we probably will want two logger injectables in the future.

PR #4322, opened 2021-11-11
Replacing global themed sass-vars to css-vars

ixrock:
**PR highlights:**

- Having default css-variables in app from generated default dark theme `lens-dark.json`
- Better IDE/`vscode` css-variables support / auto-completion
- More consistent styles overall
- Possibility to get rid of `sass` in future at all (if `postcss` would be just enough)

Screenshot 2021-11-11 at 20 57 42

**Recommended `vscode` extension for css-variables auto-complete:**
https://marketplace.visualstudio.com/items?itemName=phoenisx.cssvar

And add this to `settings.json` for auto-scanning css-variables in the project from all possible files:

```json
    "cssvar.files": [
        "./src/renderer/**/*.css",
        "./src/renderer/**/*.scss",
    ],
```


src/renderer/themes/theme-vars.css
```diff
@@ -0,0 +1,136 @@
+
```
Nokel81, line 1: I don't think this file should be committed, should it?

build/build_theme_vars.ts
```diff
@@ -0,0 +1,47 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import fs from "fs-extra";
+import path from "path";
+import defaultBaseLensTheme from "../src/renderer/themes/lens-dark.json";
```
Nokel81, line 24: Why only dark?

src/renderer/themes/theme-vars.css
ixrock: Yes, it should. It's used in app as default values for themed variables + used by editor's css-vars auto-completion.

_P.S._ Most `vscode` plugins for autocompleting css-variables support only defining exact file to read for variables, so could be useful too.

build/build_theme_vars.ts
ixrock: Only? Because default could be just one as name says `defaultBaseLensTheme` _(could be default white tought)_

Nokel81: Ah so this is only for the auto-complete

jim-docker: There seem to be more colors in `lens-light.json`

jim-docker (commented):
I'm seeing this:

Screen Shot 2021-11-12 at 12 54 24 PM

should be:

Screen Shot 2021-11-12 at 12 40 55 PM

~~I thought it was a problem in master too because after testing your branch I pulled master and ran `make dev` and got the same (bad) result. But when I ran `make clean && make dev` on master it was good. This build target should be part of `make dev`? Oh, it's probably just the transition between this PR and master, once it's merged there won't be this build issue moving forward.~~ nvm

ixrock:
@lensapp/lens-maintainers  ready for re-review / PTALA

Added 2 missing theme colours to `lens-dark.json`: `--layoutTabsLineColor`, `--canvasBackground` .

@jim-docker  light theme seems to be fixed for now:

image

src/renderer/components/+nodes/nodes.scss
```diff
@@ -37,7 +37,7 @@
       align-self: center;
 
       .LineProgress {
-        color: $lensBlue;
```
jim-docker, line 40: Were `lensBlue` and `lensMagenta` defined somewhere?

jim-docker (approved):
LGTM

src/renderer/components/+nodes/nodes.scss
ixrock: Nope, they where defined in linking sass-vars to css-vars in deleted `theme-vars.scss`, e.g.

```scss
$lensBlue: var(--blue);
// ... other themed vars
```

PR #4329, opened 2021-11-12
Tweak release drafter template

jakolehm:
So that we don't get duplicate PR title on each line.


Nokel81 (approved):

jim-docker (approved):
I guess we verify what it actually looks like when it is live?

Nokel81:
I think so

PR #4337, opened 2021-11-15
Use propagationPolicy=Background by default for delete

jakolehm:
- matches better kubectl functionality
- allows to override propagationPolicy for each delete request
- fixes removal of Job not triggering removal of related pods (regression from https://github.com/lensapp/lens/pull/4325)


nevalla (approved):
LGTM

PR #4338, opened 2021-11-15
Introduce a way to use injectables

jansav:
### What:
Introduce `injectable` as solution to make dependency inversion happen.

### Why:
- To create simplicity.
- To create sustainable velocity.
- To cultivate dev-experience. This should be simple and fun to use for the Team :)
- To make unit testability happen.
- To introduce a mechanism which permits incremental shift from shared global state.
- To introduce separation of concerns for code that does stuff and code that instantiates stuff.
- To introduce a mechanism for "loose coupling".
- To make code to instantiate dependencies more DRY (Don't Repeat Yourself).
- To improve agility of architecture. E.g. with this we could have: async dependencies, lifecycles, plugins, modularity, etc. with minimal or no impact in implementation of injectables.

### How:
UI code can inject an `injectable` like this: 
`<Inject injectableKey={someInjectable} />`

... and `injectable`s themselves can inject other `injectable`s like this: 
`di => ({ someDependency: di.inject(someOtherInjectable) })`

### What is an `injectable`
- An `injectable` declares how a dependency is instantiated when injected.
- Injectable is defined in a file such as `someInjectable.injectable.ts` or `*.injectable.tsx`.
- Injectable files look something like this:

```javascript
interface ISomeDependencies {
  someDependency: { name: string };
}

const someInjectable: IInjectable<
  string,
  ISomeDependencies
> = {
  getDependencies: di => ({
    someDependency: di.inject(someOtherInjectable),
  }),

  instantiate: ({ someDependency }) => someDependency.name
};

export default someInjectable
```

Note: This PR extracts setupping of injectables from #4282. First use case is described there. (https://github.com/lensapp/lens/pull/4282/commits/9d576f199a96965e1f0dce581643021b8480859f)


package.json
```diff
@@ -185,6 +185,8 @@
     "@kubernetes/client-node": "^0.15.1",
     "@sentry/electron": "^2.5.4",
     "@sentry/integrations": "^6.13.3",
+    "@ogre-tools/injectable": "^1.0.2",
```
panuhorsmalahti: Did you look into alternatives for "injection library"? I haven't researched the options, but this dependency seems to have really low usage compared to e.g. https://www.npmjs.com/package/injection-js

ixrock: Also `injection-js` is written in typescript and has own types `*.d.ts`

Iku-turso: Me and @jansav have a fair grasp of most "injection libraries" out there by following the scene some time now. Yet, we still chose `injectable` for good reasons. But before that, full disclosure: me and @jansav are the authors of `injectable`.

Ahem. Some reasons:
1. We have extensive experience in `injectable` - and what it is supposed to do - to make it deliver, and do it fast. In fact, we are betting our rep on that :)
2. `injectable` has low usage for being released just last week. However, that doesn't mean it hasn't seen its fair bit of action.
3. `injectable` is simple, and easy to adopt. We work real hard to make all things related as *sensual* as possible.

Yeah. I suppose we'll need to "damn some torpedoes" to make a difference in the "why" part of this PR description :)

Iku-turso: @ixrock Types for `injectable` are in this codebase for demonstrational purposes, mostly. The types are about to relocate to natural `injectable` soon enough.

Nokel81: I am still not convinced that defaultProps (for class components) and object destructuring (for function components) is not enough for injection during testing.

Iku-turso: Why using default parameters/props is not good enough as pattern?

- Using default parameters means "bastard injection anti-pattern", as explained eg. here: https://joesusnick.medium.com/why-did-apple-advocate-for-bastard-injection-4b8268e91712
- Unit testability is not the only goal here. To achieve *loose coupling* and Inversion of Control would represent considerable value in codebase, right? :)

This is after using bastard injection *for years* before dropping it. I assure you we have the experience :) We are also putting our rep on the line to make all this worth it.

I feel there's a lot more to be said on this topic. We can take this to Slack if you want? :)

types/ogre-tools-injectable.d.ts
```diff
@@ -0,0 +1,71 @@
...
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+declare module "@ogre-tools/injectable" {
```
Nokel81: Also, have you run any of these types through typescript? Because none of this compiles.

src/renderer/components/getDiForUnitTesting.tsx
```diff
@@ -0,0 +1,53 @@
...
+      return {
+        id: key,
+        ...injectable,
+        aliases: [injectable, ...(injectable.aliases || [])],
+      };
+    })
+
+    .forEach(injectable => di.register(injectable));
+
+  di.preventSideEffects();
```
ixrock, line 46: What is side-effect here btw?

types/ogre-tools-injectable-react.d.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+/// <reference types="react" />
+declare module "@ogre-tools/injectable-react" {
```
ixrock: Maybe better to keep `@ogre-tools/*` types on github and consume with the npm package?

src/renderer/components/getDi.tsx
```diff
@@ -0,0 +1,31 @@
...
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import { createContainer } from "@ogre-tools/injectable";
+import type { IConfigurableDependencyInjectionContainer } from "@ogre-tools/injectable";
+
+export const getDi = () => {
```
ixrock, line 25: Can we see an example of how would look typical react component with all that stuff under the hood?
And maybe not only react components, all known use-cases would be good to see. Thanks!

src/renderer/components/getDiForUnitTesting.tsx
jansav: There is a way to mark that usage of an `Injectable` causes side-effects. So e.g.

```typescript
const someInjectable: IInjectable<ISomething> = {
  getDependencies: () => ({}),

  instantiate: () => { 
    
    // Something that shouldn't be called in unit tests (e.g. fetch)
    return something;
  } ,

  // This means that when you are injecting the `injectable` in unit tests (after
  // `di.preventSideEffects()` has been called.) it will throw a friendly error saying that 
  // you need to override it since side-effects are not welcome in tests.
  causesSideEffects: true
};
```

`causesSideEffects` property is not defined in `IInjectable` interface yet, since we don't have any requirement for it yet. But it's getting there at the latest when we move types to the library.

types/ogre-tools-injectable-react.d.ts
jansav: We added these locally: 
- because it gives as a headstart (no need to wait that library has been converted to TS)
- because we can iterate faster with interfaces until stable

Plan is to remove these asap.

types/ogre-tools-injectable.d.ts
Iku-turso: Let us help! :) Surely enough, we don't get compilation errors. Can you post them?

Nokel81: [Here is a link](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBASwHY2FAZgQwMbDgSQBFgxgkATM7AT3yQCthsYEIkBhNmTZNOAbwBQcEYgZMYALjgAeYaJEAVOo2aYARgBs8oVBQDOBFRI3aZypPu5JcAGjiLipClQTB9APlvyFFq5htgbwUlJzJKGzd9YNEPAAofUWRVbi1gAGlgamkLFNMgxIBKOABeDzgAJWAYAFcoJEVqUnNjNTSAbQAiZH8UBExUToBdDwBueQBfQXlQSFgxVAwcPHxOJHQEAHM6-LCXG1pxZlYOLh4kNETdcMMiEnCqQ5STte5eKAFEqGBNhCs0aQJEJiPJpaT4VqpMwBajlAA+BE4AFtIBcUJD8jIYV4iqVygA3CAIcjjBRgb74sgwADKxOAAFF0OgJPpAcUynBCcTxokIJSoFA6dIWkcoToQHpyLcMWkRb07HBsfY-NZcPFEiJkiYwQ4ZdoYgo+WhBZQoNIqrV6o1mrltdouj1rCwBsBhjiFOyCUSSZNppqUGgsLhERAkUi2HrgOY9hFsFFykIFMSAPzSKyCpCbUmiTbVGOudyA8gIVMEfMHSEvM7vT0Octx9zZkSaBDMmjYbSlltt6gdhlIGpI7NTGbgaDwZCLIMrSNyXx0eUFXz1qJwa4GOAQdQpUoCCYG3WWJ39FhsAAKmCgmCR1T466lm+3El3-H38nK99uyPDSFnjnu+wNp4nxJkeATOqggIanAlDOLGUQ5Cu7gHv6vQQScF5Xjeiw5Aux4DBhl7XreUAHrWKoBLgw5+muY7zGQg5wN2TC9to9IDkiIGiPoyCbNoMBsAeMBXpYbgoAe+jYBApDkIoIk8VSMQjiIAD0KlrvoLYoAAtMW+j5NpFwStpWl4DUA76MA5DaQgKLjvoKlIBA2nmTUlnWfil76NpMD6KOczwFJR5wNg3wuq85x8CUcBxAAdPFuYwFUACONQIN8ry6KyipINQ7RDJ6iSrGwGzbFeaT1k8EhVigkVQKMQA)

types/ogre-tools-injectable-react.d.ts
jakolehm: Is the plan to move types to ogre-tools before this PR is "ready"?

types/ogre-tools-injectable.d.ts
Iku-turso: Hmm. That looks outdated, eg. `IComponentInjectable` is no more. This will go away today, when we relocate the types from lens to library `injectable`.

src/renderer/components/getDi.tsx
Iku-turso: Certainly! Working on it :)

jansav: New component from scratch: https://github.com/jansav/lens/commit/e8fca8d5a4df30f7c0151c54acf743ecd6033b51
Apologies for lack of imagination. It's another TODO-list.

jansav: For more, check out detailed examples for every feature here as documentative unit tests: 
- [injectable-react library](https://github.com/ogre-works/ogre-tools/blob/master/packages/injectable-react/src/Inject/Inject.test.js)
- [injectable library](https://github.com/ogre-works/ogre-tools/blob/master/packages/injectable/src/dependency-injection-container/createContainer.test.js)

types/ogre-tools-injectable-react.d.ts
jansav: Removed

types/ogre-tools-injectable.d.ts
jansav: Types relocated to library. No compilation errors present anymore.

jansav:
PTAL

Nokel81:
So from reading the examples this is a library that allows us to write the default dependencies for using a component outside of that component, and then during testing we can inject different dependencies?

Do I have that correct?

package.json
ixrock: > Be rational about where to apply it and what parts of your system you need to decouple. Don’t try to solve problems that you don’t have yet. Maybe you will never have them or the way that you solved them now will not fit when you really face them. At the end DI is just a means to an end, like any other pattern or technology that we use. It is not a goal itself.

_From here:_ https://ilya.puchka.me/dependency-injection-in-swift/

Yes, maybe we don't need yet DI everywhere for react component  `props` / `defaultProps`..?

jansav: Absolutely! :)

Injectable should only be used when component needs dependencies, business logic or data etc.  

In most cases, where you think that you need to convert functional component to class component, you can greatly benefit from having business logic passed from outside of component and injectable is here to help with that. 

Class components are bit deeper discussion, since most times the requirement comes from the idea that you need business logic or lifecycle which should always be separated from stupid UI component. :)

jansav:
> So from reading the examples this is a library that allows us to write the default dependencies for using a component outside of that component, and then during testing we can inject different dependencies?
> Do I have that correct?

That's one advantage yes. But it also has lot of other advantages, e.g. separate concerns (UI-component vs. logic how it gets data). Also it provides ability to start writing unit tests to the larger scope, so that we don't have to start passing down dependencies from top level components to lower level components which then allows us to refactor without fear of breaking something.

jansav:
> So from reading the examples this is a library that allows us to write the default dependencies for using a component outside of that component, and then during testing we can inject different dependencies?
> 
> Do I have that correct?

There's also the other 8 things in the "Why"-list in PR description. :)

Nokel81:
Yes but most of those benefits are just the benefits of inversion of control and not deep injection, it seems.

jansav:
> Yes but most of those benefits are just the benefits of inversion of control and not deep injection, it seems.

Well yeah, it's just a tool to help you achieve those. It's an IoC-library, so it's all about how you use it. 
However, we have done research and compared it against competitors and it does what it's supposed to do with a minimal clutter. Some competitors may offer different set of features, but we value simplicity and flexibility above features that are not needed at the moment.

jakolehm (approved):
I like the simplicity of injectables. Let's merge this and then we can see how well it actually fits into different components (most likely we need few iterations to get most out of it).

src/renderer/components/getDi.tsx
chenhunghan: It's better if the examples using `react-testing-library` instead of `enzyme`, at least I don't understand what `enzymeUpdate` do in the examples.

PR #4356, opened 2021-11-16
Fix concurrent loadItems in CatalogEntityStore

panuhorsmalahti:
If `CatalogEntityStore.loadItems()` were called concurrently two or more times, the `return` inside `if (this.isLoading) {` of `ItemStore.loadItems()` would ignore the second call, which causes a bug where e.g. old filter would remain applied in the Catalog.

I decided not to modify `ItemStore` to avoid making changes to the loading behavior of that class, and instead overriding the method in `CatalogEntityStore`.

The PR also adds missing `@action` to `addCatalogFilter` and `addOnBeforeRun`.

### DIFF
--- a/src/renderer/components/+catalog/catalog-entity.store.tsx
+++ b/src/renderer/components/+catalog/catalog-entity.store.tsx
@@ -19,7 +19,7 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { computed, makeObservable, observable, reaction } from "mobx";
+import { when, computed, makeObservable, observable, reaction } from "mobx";
 import { catalogEntityRegistry, CatalogEntityRegistry } from "../../api/catalog-entity-registry";
 import type { CatalogEntity } from "../../api/catalog-entity";
 import { ItemStore } from "../../../common/item.store";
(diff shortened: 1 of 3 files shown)


src/renderer/components/+catalog/catalog-entity.store.tsx
```diff
@@ -67,4 +67,35 @@ export class CatalogEntityStore extends ItemStore<CatalogEntityItem<CatalogEntit
...
+
+  /**
+   * Override ItemStore::loadItems, the only difference is that
+   * if isLoading is true, we await for it to be false and then continue
+   * loading.
+   * This avoids a bug where if there were two or more concurrent loadItems() calls,
+   * only the first one would complete.
+   */
+  protected async loadItems(
+    request: (() => Promise<CatalogEntityItem<CatalogEntity>[] | any>) | (() => CatalogEntityItem<CatalogEntity>[] | any),
```
Nokel81: This could simplified to `() => (Promise<CatalogEntityItem<CatalogEntity>[]> | CatalogEntityItem<CatalogEntity>[])`

Nokel81: Though I would say that it would be better to add a param to `super.loadItems` to bypass that `return`, I think that would be less fragile.

panuhorsmalahti: Added parameter instead to `super.loadItems`

jweak (approved):

chenhunghan (approved):
LGTM

PR #4357, opened 2021-11-16
Mobx package upgrade

ixrock:
- "mobx": "_^6.3.7_",
- "mobx-observable-history": "^_2.0.3_", (includes @Nokel81 's fix https://github.com/ixrock/mobx-observable-history/pull/8)

closes lensapp/lens/pull/4135


src/renderer/components/delete-cluster-dialog/delete-cluster-dialog.tsx
```diff
@@ -52,15 +52,12 @@ type Props = {};
 
 @observer
 export class DeleteClusterDialog extends React.Component {
-  showContextSwitch = false;
-  newCurrentContext = "";
+  @observable showContextSwitch = false;
```
ixrock, line 55: This file changes fixes a new error from `mobx`:

```
fix: mobx.esm.js:87 Uncaught Error: [MobX] makeObservable second arg … …
e69655b
…must be nullish when using decorators. Mixing @decorator syntax with annotations is not supported.
```

So, it seems like from this PR we have to stick to single style of annotations within a class? 
Use decorators or not (_yes_).

Nokel81: I think we use decorators more often so at least for the time being we should stick with that majority.

PR looks good, just going to test it and then approve.

Nokel81 (approved):

Nokel81:
Actually please don't merge yet, I cannot seem to open drawers

Nokel81 (changes requested):
Cannot open drawers with this change

jim-docker:
I'm wondering if we should do this in 5.3.1 (i.e. don't commit to master yet). Something like this might need some time to soak/test and we may not have that time before 5.3.0 is released?

Nokel81:
@ixrock would making `mobx` a "peerDependency" of your package be a good idea, maybe this is caused by a version mismatch.

src/common/configure-packages.ts
```diff
@@ -29,7 +29,6 @@ export default function configurePackages() {
   // Docs: https://mobx.js.org/configuration.html
   Mobx.configure({
     enforceActions: "never",
-    isolateGlobalState: true,
```
ixrock, line 32: Seems like we don't need this: https://mobx.js.org/configuration.html#isolateglobalstate-boolean

ixrock:
@Nokel81 PTALA

Nokel81 (approved):

Nokel81:
> I'm wondering if we should do this in 5.3.1 (i.e. don't commit to master yet). Something like this might need some time to soak/test and we may not have that time before 5.3.0 is released?

I think I agree that we should hold off until after the release of 5.3.0

PR #4358, opened 2021-11-16
Allow to scroll table columns with labels (badges)

### LINKED ISSUE #4300: Namespace labels are cut off

ixrock:
https://user-images.githubusercontent.com/6377066/142020941-dcd55efc-debf-4e4d-b55a-95fb945ac433.mov

Alternative to #4302
Fixes #4300


src/renderer/components/badge/badge.module.css
```diff
@@ -25,6 +25,18 @@
   max-width: 100%;
 }
 
+// might be used in a scrollable area with {overflow: auto}
```
Nokel81, line 28: Do you mean "must be"?

Nokel81 (approved):
I wanted to avoid adding a new class but I do agree that not having it (along with css modules) was a dependency inversion.

src/renderer/components/badge/badge.module.css
ixrock: Well, actually not really, because this class is not required for `Badge`.
Also `overflow: auto` is not requirement for parent container and could be different.

Nokel81: okay sounds good

ixrock:
@Nokel81 tests are failed with weird unrelated error:

image

Nokel81:
I would just restart the tests in cases like that

PR #4368, opened 2021-11-17
Update built-in extensions package-lock.json

ixrock:
Motivation: each time `make dev` or `yarn dev` creates uncommitted files (similar thing as `yarn.lock`)

### DIFF
--- a/extensions/kube-object-event-status/package-lock.json
+++ b/extensions/kube-object-event-status/package-lock.json
@@ -16,9 +16,9 @@
       },
       "dependencies": {
         "@babel/runtime": {
-          "version": "7.14.0",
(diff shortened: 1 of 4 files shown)


panuhorsmalahti:
Does this solve the root cause? Maybe in few days/weeks the package-lock files start to change again?

ixrock:
> Does this solve the root cause? Maybe in few days/weeks the package-lock files start to change again?

I guess it happens only when changes some of dependencies from these packages, no?
So each time extension's dep(s) changed these lock files must be committed too AFAIK _(same as we commit `yarn.lock`)_

jim-docker (approved):

Issue #3782, opened 2021-09-12
App crash at /namespaces


hansiemithun:
**To Reproduce**
1. Go to 'Namespaces'
2. Click on 'namespace'
3. App crashes. See error below:

instrument.js:109 TypeError: namespace.getStatus is not a function

**Expected behavior**
Should show the details of the namspace without any error.

**Screenshots**
[Namespace-Crash-Error-Image-Reference](https://i.imgur.com/tUVoIci.png)

**Environment (please complete the following information):**
- Lens Version: 5.x
- OS: [e.g. OSX]: Mac OS, Catalina and Sur
- Installation method (e.g. snap or AppImage in Linux): Updated Version

**Logs:**
When you run the application executable from command line you will see some logging output. Please paste them here:
```
instrument.js:109 TypeError: namespace.getStatus is not a function
    at NamespaceDetails.render (namespace-details.tsx:85)
    at allowStateChanges (mobx.esm.js:1204)
    at mobxreact.esm.js:300
    at trackDerivedFunction (mobx.esm.js:1785)
    at Reaction.track (mobx.esm.js:2329)
    at NamespaceDetails.reactiveRender [as render] (mobxreact.esm.js:298)
    at NamespaceDetails.makeComponentReactive (mobxreact.esm.js:313)
    at NamespaceDetails.target.render (mobxreact.esm.js:226)
    at qi (react-dom.production.min.js:187)
    at pi (react-dom.production.min.js:186)
    at ck (react-dom.production.min.js:269)
    at bk (react-dom.production.min.js:250)
    at ak (react-dom.production.min.js:250)
    at Tj (react-dom.production.min.js:250)
    at Lj (react-dom.production.min.js:243)
    at react-dom.production.min.js:123
    at exports.unstable_runWithPriority (scheduler.production.min.js:18)
    at gg (react-dom.production.min.js:122)
    at jg (react-dom.production.min.js:123)
    at ig (react-dom.production.min.js:122)
    at Hb (react-dom.production.min.js:292)
    at gd (react-dom.production.min.js:73)
    at HTMLDivElement.sentryWrapped (helpers.js:72)

```

**Kubeconfig:**
Quite often the problems are caused by malformed kubeconfig which the application tries to load. Please share your kubeconfig, remember to remove any secret and sensitive information.
```
your kubeconfig here
```

**Additional context**

Nokel81:
By version 5.X do you mean that it consistently fails with 5.0, 5.1, and 5.2?

[...]

PR #4376, opened 2021-11-17
Fix crash with logs view

### LINKED ISSUE #4018: Blank Screen when accessing one (out of three) clusters

### LINKED ISSUE #4442: App crash at /
**Describe the bug**
App crash at / while opening a single cluster. 

**To Reproduce**
1. Click on connect
2. Cluster resources not showing up
3. App Crashes (See error)

**Expected behavior**
Cluster Resources should open

**Screenshots**

Screenshot 2021-11-27 at 12 04 17 PM

- Lens Version:Lens: 5.3.0-latest.20211125.
- OS: MacOS
- Installation method (e.g. snap or AppImage in Linux): .dmg

**Logs:**

```
Component stack:


    at wrappedComponent (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:10553:16)
    at div
    at Logs (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/chunks/5.js:7201:9)
    at div
    at div
    at Dock (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/chunks/5.js:7633:9)
    at div
    at div
    at MainLayout (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:286690:9)
    at ErrorBoundary (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:238000:9)
    at Router (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:79045:30)
    at ClusterFrame (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/chunks/5.js:7854:9)
    at ThemeProvider (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:604150:24)


Error stack:

TypeError: Cannot create a KubeObject from undefined
    at new KubeObject (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:32746:19)
    at new WorkloadKubeObject (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:124780:1)
    at new Pod (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:154478:9)
    at http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/chunks/5.js:7133:17
    at http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:10553:49
    at http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:10517:25
    at trackDerivedFunction (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:3114:18)
    at Reaction.track (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:3642:18)
    at useObserver (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:10515:14)
    at wrappedComponent (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:10553:16)
    at Ch (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:550643:137)
    at li (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:550666:154)
    at ki (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:550664:169)
    at ii (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:550663:178)
    at ck (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:550760:126)
    at bk (http://032c1c3b972accbf835ce6bb7e03dc78.localhost:56202/build/Lens.js:550736:347)
```

**Additional context**
This issue started after I clicked on Daemon Set YAML edit in lens. Tried updating the Lens to latest version, Restarting the IDE, Clearing up the cache at `~/Library/Application\ Support/Lens/Cache/` and reconnecting the cluster but no luck.

### LINKED ISSUE #4506: Lens crashed while viewing pod logs; now always crashes for this single cluster only
**Describe the bug**
Lens now always crashes when attempting to connect to one of my clusters. The cluster is fully reachable. The issue begn when fetching a pod log. Lens crashed (info below) and then immediately crashes after connecting to this cluster (and only this single cluster). This behavior persists after restarting. The connection seems to be successful; the crash seems to actually occur
when attempting to display the cluster.

**To Reproduce**
Simply connect to the cluster in question. Lens crashes immediately. I cannot reproduce this behavior with any other cluster, unfortunately

**Expected behavior**
Lens would connect to the cluster and display the cluster screen

**Screenshots**

- Lens Version: 5.3.2-latest.20211201.1
- OS:  OSX
- Installation method: This version was the last upgrade taken

**Logs:**
Startup is fine. Only this cluster seems to crash. The error is as follows
`App crash at /`
```
Component stack:


    at wrappedComponent (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:10557:16)
    at div
    at Logs (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/chunks/5.js:7285:9)
    at div
    at div
    at Dock (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/chunks/5.js:7717:9)
    at div
    at div
    at MainLayout (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:286596:9)
    at ErrorBoundary (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:237949:9)
    at Router (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:79054:30)
    at ClusterFrame (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/chunks/5.js:7946:9)
    at ThemeProvider (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:609860:24)

Error stack:

TypeError: Cannot create a KubeObject from undefined
    at new KubeObject (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:32750:19)
    at new WorkloadKubeObject (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:124953:1)
    at new Pod (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:154508:9)
    at http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/chunks/5.js:7217:17
    at http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:10557:49
    at http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:10521:25
    at trackDerivedFunction (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:3120:18)
    at Reaction.track (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:3649:18)
    at useObserver (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:10519:14)
    at wrappedComponent (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:10557:16)
    at Ch (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:556346:137)
    at li (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:556369:154)
    at ki (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:556367:169)
    at ii (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:556366:178)
    at ck (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:556463:126)
    at bk (http://3863dbca4b80df474930e0cfc76321c5.localhost:58504/build/Lens.js:556439:347)
```

**Kubeconfig:**
All other clusters in this file seem ok. I should note that I tried deleting this cluster from Lens; the delete attempt fails -- but with no error. Indeed, the problematic cluster was fine until the crash. I feel like if I could delete it and re-add it it would be fine

```
your kubeconfig here
```

**Additional context**

Nokel81:
- Switch to saving only IDs of pods, owners, and containers instead of
  whole objects

- Add more checks about data integrety

Fixes secondary crash reported in #3782


src/renderer/components/dock/log-list.tsx
```diff
@@ -247,7 +247,7 @@ export class LogList extends React.Component<Props> {
     if (!this.logs.length) {
       return (
         <div className="LogList flex box grow align-center justify-center">
-          There are no logs available for container
+          There are no logs available for container {logTabStore.getData(this.props.id).selectedContainer}
```
aleksfront:
```suggestion
          There are no logs available for container {logTabStore.getData(this.props.id)?.selectedContainer}
```

src/renderer/components/dock/log-resource-selector.tsx
```diff
@@ -21,93 +21,85 @@
...
-  const onPodChange = (option: SelectOption) => {
-    const selectedPod = podsStore.getByName(option.value, pod.getNs());
+function getSelected<T>(groups: GroupSelectOption<SelectOption<T>>[], value: T): SelectOption<T> | undefined {
+  for (const group of groups) {
+    for (const option of group.options) {
+      if (option.value === value) {
+        return option;
+      }
+    }
+  }
```
aleksfront: I guess previous option for setting `Select` value was shorter and more explicit?

We don't need to iterate over available group, we can just set an "arbitrary" value and the React Select will do search itself.

```
value={{ label: pod.getName(), value: pod.getName() }}
```

aleksfront (commented):
Nice improvement and overall cleaning! However, found couple of issues:

1. Do we actually need to prevent logs to show up if pod doesn't have any owners? I'm not sure how often pods can be created directly, but nothing blocks us to open logs for them? For example, here's what I see if created raw Pod from `Create Resource` tab:
pod without owner

2. If the viewed pod has suddenly disappeared, maybe it's better to show sibling pod if any? Instead of this message:
pod not found

3. For pods without (?) logs I don't see any message like "No logs available", just empty log area. Create a deployment from template dropdown to reproduce:

no message

4. This probably not affected by current PR, but it's annoying: search box is changing its width if some elements found:
jumping search

src/renderer/components/dock/log-resource-selector.tsx
Nokel81: I tried setting it to just `pod.getName()` hoping that it would pick up the `value` correctly. But that didn't work. Will revert.

Nokel81:
> Nice improvement and overall cleaning! However, found couple of issues:
> 
>     1. Do we actually need to prevent logs to show up if pod doesn't have any owners? I'm not sure how often pods can be created directly, but nothing blocks us to open logs for them? For example, here's what I see if created raw Pod from `Create Resource` tab:

Fair enough, I agree that we shouldn't have this limitation.

>     2. If the viewed pod has suddenly disappeared, maybe it's better to show sibling pod if any? Instead of this message:

Sounds good

>     3. For pods without (?) logs I don't see any message like "No logs available", just empty log area. Create a deployment from template dropdown to reproduce:

Sounds good

>     4. This probably not affected by current PR, but it's annoying: search box is changing its width if some elements found:

I will try and fix this in this PR anyway

Nokel81:
@aleksfront Resolved, can you PTALA?

src/renderer/components/dock/log-tab.store.ts
```diff
@@ -21,23 +21,28 @@
...
-  podsOwner: string;
+  podsOwner?: string;
 
   /**
    * The ID of the pod from the list of pods owned by `.podsOwner`
    */
   selectedPod: string;
 
+  /**
+   * The namespace of the pods. Used for watching those namespaces
```
aleksfront: Used for watching those pods?

src/renderer/components/dock/log-tab.store.ts
```diff
@@ -167,18 +173,34 @@ export class LogTabStore extends DockTabStore<LogTabData> {
 
     if (error) {
       logger.warn(`[LOG-TAB-STORE]: data for ${tabId} was invalid`, error);
-      this.closeTab(tabId);
+      runInAction(() => {
+        this.closeTab(tabId);
+      });
```
aleksfront: We don't need `runInAction` wrapper here since `closeTab()` method already has `@action` decorator below.

src/renderer/components/dock/log-tab.store.ts
```diff
@@ -167,18 +173,34 @@ export class LogTabStore extends DockTabStore<LogTabData> {
...
+  @action
   public closeTab(tabId: string) {
     this.clearData(tabId);
     this.dockManager.closeTab(tabId);
   }
+
+  /**
+   * Get a set of namespaces which pod tabs care about
+   */
+  public getNamespaces(): string[] {
```
aleksfront, line 180: I don't get why we should get namespaces in some way and watch them in `logs.tsx`. The pod which logs we're looking at (as well as it's "sibling" pods with same owner) could have only 1 namespace. So why we need to collect more than one? Isn't it enough to pick `namespace` attribute from the previewed pod like `pod.getNs()`?

src/common/k8s-api/kube-object.store.ts
```diff
@@ -274,10 +274,10 @@ export abstract class KubeObjectStore<T extends KubeObject> extends ItemStore<T>
...
     // update existing items
     if (merge) {
-      const namespaces = partialItems.map(item => item.getNs());
+      const uids = new Set(partialItems.map(item => item.getId()));
 
       items = [
-        ...this.items.filter(existingItem => !namespaces.includes(existingItem.getNs())),
+        ...this.items.filter(existingItem => !uids.has(existingItem.getId())),
         ...partialItems,
       ];
```
aleksfront: Not obvious why this is needed?

src/renderer/components/dock/logs.tsx
```diff
@@ -51,112 +59,178 @@ export class Logs extends React.Component<Props> {
...
+            logStore.clearLogs(this.tabId);
+          }
+
+          return when(() => this.canLoad, () => this.load(true));
+        },
+        {
+          equals: comparer.structural,
+        },
+      ),
+      disposingReaction(() => this.tabData, (data) => {
```
aleksfront: Can it be done with standard mobx reactions? Or event with good old `componentDidUpdate` props comparisons? Just thinking if we decided to use reactions less often and now we have one more reaction method. However, I have nothing against it to have one, maybe just need to add tests and first of all weight its necessity?

aleksfront:
@Nokel81 Most of the issues are gone, thanks for fixing those.

1. This is still an issue:
> If the viewed pod has suddenly disappeared, maybe it's better to show sibling pod if any? Instead of this message:
removed pod issue

2. Suggestions about tab naming. First, I doubt we should add namespace before workload name. It makes impossible to get meaningful information from tab because ns will always be long enough to cover all available space.
long space names

Maybe it's just me, but kind of strange to see titles like "ReplicaSet logs..." if you just just looking at the Pod logs:


https://user-images.githubusercontent.com/9607060/144811012-5919ea28-3330-4d3e-bdab-aba48c211627.mov

3. You've marked this PR as a candidate to fix #4018 (https://github.com/lensapp/lens/issues/4018#issuecomment-983856353). Maybe to move things forward, extract needed parts for 4018 and pull a separate PR? Or it's better to keep with current one?

Nokel81:
>  Or it's better to keep with current one?

I think it would be better to keep it here because the fix it to move out the `reloadLogs` call in `render` but that requires some other changes I think too.

Nokel81:
> 1. This is still an issue:

Oh, I thought I fixed that...

> 2. Suggestions about tab naming. First, I doubt we should add namespace before workload name. It makes impossible to get meaningful information from tab because ns will always be long enough to cover all available space.

Sounds good about the naming.

> Maybe it's just me, but kind of strange to see titles like "ReplicaSet logs..." if you just just looking at the Pod logs:

I can revert that change but let me explain why I made it. I made it because I felt that each tab was both about logs and about switching between pods and containers. And to be honest we aren't really even viewing pod logs, but instead viewing container logs.

src/renderer/components/dock/log-tab.store.ts
Nokel81: It is, however, this gets the namespaces for all tabs not just the currently open one. 

This is needed for when we reopen Lens and the `pods` view might not be open yet. So the Logs component needs to register a watch for pods in all of these namespaces.

I guess I could watch pods in a single namespace for each tab. WDYT?

src/common/k8s-api/kube-object.store.ts
Nokel81: I'll add a comment then, but the main gist is that stores cannot assume that a given call to `mergeItems` represents all namespaces being watched. 

Specifically, a Logs tab might be viewing logs from a pod in `default` while the pods view is showing only the ones in `kube-system`

src/renderer/components/dock/logs.tsx
Nokel81: For the specific call you commented on here I have replaced it with a plain `reaction`. However, with the call above I think we actually need it. Because there is an inherent race between switching of tabs or pods or containers and being able to load new logs.

src/renderer/components/dock/log-tab.store.ts
Nokel81: Actually that is incorrect, we do need the `runInAction` wrapper because otherwise mobx will try and update within the same event. And if that event is a render step then it throws an error.

However, I have refactored so that this code is no longer needed.

src/common/utils/comparer.ts
```diff
@@ -0,0 +1,49 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/**
+ * The functions of this module should be used so that
+ */
```
aleksfront: This comment should be fixed (or removed)

src/common/utils/reactions.ts
```diff
@@ -0,0 +1,42 @@
...
+import type { IReactionPublic, IReactionOptions, IReactionDisposer } from "mobx";
+import { reaction } from "mobx";
+import type { Disposer } from "./disposer";
+
+/**
+ * Similar to mobx's builtin `reaction` function but supports returning a
+ * disposer from `effect` that will be cancelled everytime a new reaction is
+ * fired and when the reaction is disposed.
+ */
+export function disposingReaction<T, FireImmediately extends boolean = false>(expression: (r: IReactionPublic) => T, effect: (arg: T, prev: FireImmediately extends true ? T | undefined : T, r: IReactionPublic) => Disposer, opts?: IReactionOptions<T, FireImmediately>): IReactionDisposer {
```
aleksfront, line 30: A wish to replace all those types if they come from mobx reaction, because it's really hard understand what is written here.

src/main/cluster.ts
```diff
@@ -34,7 +34,7 @@ import { DetectorRegistry } from "./cluster-detectors/detector-registry";
 import plimit from "p-limit";
 import type { ClusterState, ClusterRefreshOptions, ClusterMetricsResourceType, ClusterId, ClusterMetadata, ClusterModel, ClusterPreferences, ClusterPrometheusPreferences, UpdateClusterModel, KubeAuthUpdate } from "../common/cluster-types";
 import { ClusterMetadataKey, initialNodeShellImage, ClusterStatus } from "../common/cluster-types";
-import { disposer, storedKubeConfigFolder, toJS } from "../common/utils";
```
aleksfront, line 37: Why we should use custom comparer if it works same and uses exact same notation as mobx one?

src/renderer/components/cluster-manager/cluster-status.tsx
```diff
@@ -150,6 +150,10 @@ export class ClusterStatus extends React.Component<Props> {
   }
 
   render() {
+    if (!this.entity) {
+      return null;
+    }
```
aleksfront: Extra checks is good but it doesn't related to pod logs? And is there a possibility now to see empty cluster screen with one more reason? 

If we check for `entity` so it fixes some app crash issue?

src/renderer/components/dock/logs.tsx
```diff
@@ -33,15 +32,33 @@ import { logStore } from "./log.store";
...
   tab: DockTab
 }
 
+type GottenPods = {
+  pods: Pod[];
+  pod: Pod | undefined;
+} | {
+  pod: Pod | undefined;
+  pods?: undefined;
+};
```
aleksfront: Why not just return empty array [] for `pods` when needed instead of this construction? To keep predictable return value.

src/renderer/components/dock/logs.tsx
```diff
@@ -51,112 +68,176 @@ export class Logs extends React.Component<Props> {
...
+          });
+        }
+      }, {
+        fireImmediately: true,
+      }),
+    ]);
+  }
+
+  private getPods(data: LogTabData): GottenPods;
+  private getPods(data: LogTabData | undefined): GottenPods | undefined {
```
aleksfront, line 127: As mentioned above, maybe just always return same time?

src/renderer/components/dock/logs.tsx
```diff
@@ -51,112 +68,176 @@ export class Logs extends React.Component<Props> {
...
-      reaction(() => this.props.tab.id, this.reload, { fireImmediately: true }),
-    );
+    disposeOnUnmount(this, [
+      kubeWatchApi.subscribeStores([
+        podsStore,
+      ], {
+        namespaces: logTabStore.getNamespaces(),
+      }),
+      when(() => this.canLoad, () => this.load(true)),
+      disposingReaction(
```
aleksfront: I would move all of this reactions to different methods for sake of readability. Now it's too big for `componentDidMount()`?

aleksfront:
Additional observations:
1. We have blink of unstyled placeholder on cluster connect / window reload:

https://user-images.githubusercontent.com/9607060/145219825-6e97328b-ec3e-43bd-a552-6fb4dffafdac.mov

2. Lets revert tab title to have "Pod Logs" names:
renamed titles

3. Need to style this placeholder a bit, make it white and put into center, wdyt?


https://user-images.githubusercontent.com/9607060/145220460-0ffcf936-2b72-4431-8aae-440365842a3d.mov



4. For some reason now we have 2 separate spinners around:

https://user-images.githubusercontent.com/9607060/145220251-5514b2e3-fcab-4ba5-a347-1f61569b4a08.mov

src/renderer/components/cluster-manager/cluster-status.tsx
Nokel81: Ah oops, I meant to move this into a separate PR.

src/renderer/components/dock/logs.tsx
Nokel81: Because an empty array means that `workloadId` was set and because `[]` means no pods, where as `undefined` means did even try.

src/renderer/components/dock/logs.tsx
Nokel81: sounds good

src/main/cluster.ts
Nokel81: Because the mobx ones are explicitly `(a: any, b: any) => boolean` which breaks typescript type inference.

src/common/utils/reactions.ts
Nokel81: I don't quite understand what you mean...

Do you not like the types? Except for the fact that `effect` returns a `Disposer` this is the exact signature of the builtin `reaction` function.

src/renderer/components/dock/logs.tsx
Nokel81: Return same time?

src/renderer/components/dock/logs.tsx
aleksfront: Can you elaborate this a bit more? Can we convert this to 
```
pods: Pod[]
pod: Pod | undefined
```
?

Nokel81: I guess it could be:

```
pods: Pod[] | undefined;
pod: Pod | undefined;
```

I wanted to capture in the type the two different states, but I see now that it isn't that much clearer.

src/renderer/components/dock/logs.scss
```diff
@@ -0,0 +1,27 @@
...
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+.PodLogs {
+  .pod-not-found {
+    margin: auto;
+    color: var(--terminalBrightWhite);
+  }
+}
```
aleksfront, line 27: Minor thing, but it's better to stick with css modules to properly isolate styles.

aleksfront (dismissed):
Problems solved, looking good!

src/renderer/components/dock/logs.scss
Nokel81: I'll make a follow up PR to move pod logs to modules.

src/common/ipc/ipc.ts
```diff
@@ -34,7 +34,9 @@ const electronRemote = (() => {
   if (ipcRenderer) {
     try {
       return require("@electron/remote");
-    } catch {}
+    } catch {
+      // ignore temp
```
Iku-turso, line 38: I can see that there is a wild-card try-catch, but the code does not explain why. If the try/catch is indeed an absolute necessity, I would like to know why. Better comment, or preferrably lovingly named function, would be better. Future improvement.

src/common/k8s-api/kube-watch-api.ts
```diff
@@ -116,7 +117,11 @@ export class KubeWatchApi {
   #watch = new WatchCount();
 
   private subscribeStore({ store, parent, watchChanges, namespaces, onLoadFailure }: SubscribeStoreParams): Disposer {
-    if (this.#watch.inc(store) > 1) {
+    const isNamespaceFilterWatch = !namespaces;
+
+    namespaces ??= KubeWatchApi.context?.contextNamespaces ?? [];
```
Iku-turso, line 122: Global shared state, fixed elsewhere.

src/common/k8s-api/kube-watch-api.ts
```diff
@@ -116,7 +117,11 @@ export class KubeWatchApi {
...
+
+    if (isNamespaceFilterWatch && this.#watch.inc(store) > 1) {
```
Iku-turso, line 124: Named function required. Future improvement.

src/common/k8s-api/kube-watch-api.ts
```diff
@@ -167,7 +172,7 @@ export class KubeWatchApi {
       : noop; // don't watch namespaces if namespaces were provided
 
     return () => {
-      if (this.#watch.dec(store) === 0) {
+      if (!isNamespaceFilterWatch || this.#watch.dec(store) === 0) {
```
Iku-turso, line 175: Named function required. Future improvement.

src/renderer/components/dock/__test__/log-resource-selector.test.tsx
```diff
@@ -99,42 +89,46 @@ describe("<LogResourceSelector />", () => {
...
     expect(ns).toHaveTextContent("default");
   });
 
   it("renders proper selected items within dropdowns", () => {
-    const tabData = getOnePodTabData();
-    const { getByText } = render(getComponent(tabData));
+    const props = getOnePodTabProps();
+    const { getByText } = render(<LogResourceSelector {...props} />);
 
     expect(getByText("dockerExporter")).toBeInTheDocument();
```
Iku-turso, line 110: Notice: Multiple expects in a test is very bad.

src/renderer/components/dock/__test__/log-tab.store.test.ts
```diff
@@ -19,146 +19,137 @@
...
+  return {
+    renameTab: jest.fn(),
+    createTab: jest.fn(),
+    closeTab: jest.fn(),
+  };
+}
+
+describe("LogTabStore", () => {
+  describe("createPodTab()", () => {
+    it("throws if data is not an object", () => {
```
Iku-turso, line 37: Test describes technical stuff, and not reasoning behind it. Future improvement.

src/renderer/components/dock/__test__/log-tab.store.test.ts
```diff
@@ -19,146 +19,137 @@
...
+      expect(() => logTabStore.createPodTab({ selectedContainer: {} as any, selectedPod: pod })).not.toThrow();
     });
 
-    expect(logTabStore.getData(dockStore.selectedTabId)).toEqual({
-      pods: [selectedPod, siblingPod],
-      selectedPod,
-      selectedContainer,
-      showTimestamps: false,
-      previous: false,
+    it("should return a TabId if created sucessfully", () => {
```
Iku-turso, line 66: Typo.

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -19,63 +19,88 @@
...
+  storageKey?: string;
+
+  /**
+   * A function to call for validating values. It should `throw` if an error is present
+   */
+  validator?: (value: T) => void;
 }
 
-export type DockTabStorageState<T> = Record<TabId, T>;
+type PartialObject<T> = T extends object ? Partial<T> : never;
```
Iku-turso, line 46: I can see this is partial, but naming should hint why. Future improvement.

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -19,63 +19,88 @@
...
-    this.options = {
-      autoInit: true,
-      ...this.options,
-    };
+    this.validator = validator;
 
-    if (this.options.autoInit) {
-      this.init();
+    if (autoInit) {
+      this.init(storageKey);
```
Iku-turso, line 59: Side-effect in constructor is very bad. Future improvement.

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -19,63 +19,88 @@
...
-
+  protected init(storageKey: string | undefined) {
     // auto-save to local-storage
     if (storageKey) {
       this.storage = createStorage(storageKey, {});
-      this.storage.whenReady.then(() => {
-        this.data.replace(this.storage.get());
-        reaction(() => this.toJSON(), data => this.storage.set(data));
-      });
+      this.storage.whenReady.then(action(() => {
```
Iku-turso, line 67: `whenReady`, here and elsewhere, communicates uncertainty for state of what should be a dependency. A dependency is guaranteed to be initialized, and at correct state. The responsibility is somewhere else. Future improvement.

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -19,63 +19,88 @@
...
+        for (const [tabId, value] of Object.entries(this.storage.get())) {
+          try {
+            this.setData(tabId, value);
```
Iku-turso, line 70: This should be a fatal error. Using try/catch for operational logic is evil. Future improvement.

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -19,63 +19,88 @@
...
+          } catch (error) {
+            logger.warn(`[DOCK-TAB-STORE-${storageKey}]: data for ${tabId} was invalid, skipping`, error);
+            dockStore.closeTab(tabId);
+          }
+        }
+        reaction(
+          () => this.toJSON(),
```
Iku-turso, line 77: Non-explicit reaction is evil, because it may trigger unexpectedly. Future improvement.

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -19,63 +19,88 @@
...
+          data => this.storage.set(data),
```
Iku-turso, line 78: This reaction (evil in general, at most times) should be replaced with an `action`. Future improvement.

src/renderer/components/dock/dock-tab.store.ts
```diff
@@ -19,63 +19,88 @@
...
+          {
+            // fireImmediately so that invalid data is removed from the store
+            fireImmediately: true,
+          },
+        );
+      }));
     }
 
     // clear data for closed tabs
     autorun(() => {
```
Iku-turso, line 88: Notice: evil `autorun` should be an action.

src/renderer/components/dock/log-list.tsx
```diff
@@ -25,25 +25,20 @@ import React from "react";
...
-import { logStore } from "./log.store";
-import { logTabStore } from "./log-tab.store";
 import { ToBottom } from "./to-bottom";
 
 interface Props {
   logs: string[]
   isLoading: boolean
   load: () => void
-  id: string
+  selectedContainer: string
```
Iku-turso, line 41: Name does not imply string. Future improvement.

src/renderer/components/dock/log-list.tsx
```diff
@@ -64,9 +59,11 @@ export class LogList extends React.Component<Props> {
 
   componentDidMount() {
     disposeOnUnmount(this, [
-      reaction(() => this.props.logs, this.onLogsInitialLoad),
-      reaction(() => this.props.logs, this.onLogsUpdate),
-      reaction(() => this.props.logs, this.onUserScrolledUp),
+      reaction(() => this.props.logs, (logs, prevLogs) => {
```
Iku-turso, line 62: Notice: Reaction should be an action. Also this logic does not belong in a component. Future improvement.

src/renderer/components/dock/log-list.tsx
```diff
@@ -88,10 +85,14 @@ export class LogList extends React.Component<Props> {
 
   @boundMethod
   onUserScrolledUp(logs: string[], prevLogs: string[]) {
-    if (!this.virtualListDiv.current) return;
+    const { current } = this.virtualListDiv;
+
+    if (!current) {
```
Iku-turso, line 90: Notice: Code does not explain real-life reasoning of missing `current`.

src/renderer/components/dock/log-tab.store.ts
```diff
@@ -19,131 +19,172 @@
...
+
+  /**
+   * Whether to show timestamps inline with the logs
+   */
+  showTimestamps: boolean;
+
+  /**
+   * Query for getting logs of the previous container restart
+   */
+  previous: boolean;
```
Iku-turso, line 60: Naming could be better. Future improvement.

src/renderer/components/dock/log.store.ts
```diff
@@ -120,48 +121,35 @@ export class LogStore {
...
 
-  /**
-   * Converts logs into a string array
-   * @returns Length of log lines
-   */
-  @computed
-  get lines(): number {
-    return this.logs.length;
-  }
+  getLogs(tabId: TabId, { showTimestamps }: LogTabData): string[] {
```
Iku-turso, line 146: This needs to be reactive. The big-splash makes this reactive, among other things. The big-splash needs to be merged first because this.

src/renderer/components/dock/log.store.ts
```diff
@@ -179,24 +167,33 @@ export class LogStore {
...
+    if (!logs?.length) {
+      return "";
+    }
+
+    const timestamps = this.getTimestamps(logs[0]);
 
-    if (!extraction || extraction.length < 3) {
-      return ["", logs];
+    if (!timestamps) {
+      return "";
```
Iku-turso, line 185: Empty string should be null when possible. Future improvement.

src/renderer/components/dock/logs.tsx
```diff
@@ -51,112 +68,212 @@ export class Logs extends React.Component<Props> {
...
   componentDidMount() {
-    disposeOnUnmount(this,
-      reaction(() => this.props.tab.id, this.reload, { fireImmediately: true }),
+    disposeOnUnmount(this, [
+      kubeWatchApi.subscribeStores([
+        podsStore,
+      ], {
+        namespaces: logTabStore.getNamespaces(),
+      }),
+      when(() => this.canLoad, () => this.load(true)),
```
Iku-turso, line 77: Initialization in wrong place (component). State should be ready when rendering. Future improvement.

src/renderer/components/dock/logs.tsx
```diff
@@ -51,112 +68,212 @@ export class Logs extends React.Component<Props> {
...
+      },
     );
   }
 
-  get tabId() {
+  /**
+   * Only used in `componentDidMount`
+   */
+  private changeSelectedPodWhenCurrentDisappears() {
+    return reaction(() => this.getPods(this.tabData), (data) => {
```
Iku-turso, line 106: Preventable `reaction`. Use action.

src/renderer/utils/createStorage.ts
```diff
@@ -50,8 +51,11 @@ export function createStorage<T>(key: string, defaultValue: T) {
 
       try {
         storage.data = await fse.readJson(filePath);
```
Iku-turso, line 53: Notice: explicit access to filesystem! Explicit side-effect.

Iku-turso (changes requested):
Uh. To be blunt, there are too many problems with this PR to approve. We propose live session to cover all more efficiently. Also PR #4653 (ie. a big splash) makes significant improvements in code touched here. The improvements include getting rid of global shared state and involuntary re-renders which also cause side-effects.

src/renderer/components/dock/log-list.tsx
Nokel81: It does actually because the kube api has `containers` which have names which are unique to the pod.

src/renderer/components/dock/log-list.tsx
Nokel81: What do you mean by "should be an action"?

Nokel81:
Closing this PR as the commit 9d33fff90673faf8c72043f1e682acb1aa7526ce fixes a lot of the issues that this did as well.

I will recreate the changes to the storage layer in a new PR (for 5.5) once [#3738](https://github.com/lensapp/lens/pull/4738) is merged.

PR #4385, opened 2021-11-18
Reroute to landing page from previously registered extension cluster pages

### LINKED ISSUE #1247: Page not found after disabling extension
Would it be better to go somewhere else like the landing page?

extui

Nokel81:

fixes #1247


Nokel81:
The other option is to just remove the "NoFound" route like we don't have in the global scope pages.

src/renderer/cluster-frame.tsx
```diff
@@ -211,6 +211,12 @@ export class ClusterFrame extends React.Component {
               {this.renderExtensionTabLayoutRoutes()}
               {this.renderExtensionRoutes()}
               <Redirect exact from="/" to={this.startUrl}/>
+              {
+                ClusterPageRegistry
+                  .getInstance()
+                  .redirectEntries
+                  .map(url => <Redirect key={url} from={url} to={this.startUrl} />)
+              }
```
ixrock: Why not to make `this.startUrl` just default page for everything that not matched specific routes above?

So I mean something like this would work? (without saving `knownUrls`)

```tsx
        // ...
        <Route component={Apps}/>
        <Route component={NotFound}/>
        {this.renderExtensionTabLayoutRoutes()}
        {this.renderExtensionRoutes()}
        <Redirect from="*" to={this.startUrl}/>
```

Nokel81: Yeah that would work to, my only thought was that then we wouldn't be able to have the "NotFound" route which I figured we wanted. But if we don't then I completely agree that solution is better.

ixrock: Probably we don't need `NotFound` at all as it doesn't carry useful information for user (just a text _Page not found_). Better to ask others @lensapp/lens-maintainers Wdyt?

Nokel81: TBH I am fine with removing `NotFound`

jim-docker: I guess "page not found" is useful to make the user aware that something changed. But what @ixrock suggests is nice. Instead of the NotFound page can a notification be shown that says something like "Page not found, returning to start"?

Nokel81: I will investigate if that is possible

Nokel81:
@jim-docker and @ixrock PTALA

I have changed it to a notification when the route is not found.

src/renderer/cluster-frame.tsx
```diff
@@ -211,13 +210,11 @@ export class ClusterFrame extends React.Component {
...
               <Redirect exact from="/" to={this.startUrl}/>
-              {
-                ClusterPageRegistry
-                  .getInstance()
-                  .redirectEntries
-                  .map(url => <Redirect key={url} from={url} to={this.startUrl} />)
-              }
-              <Route component={NotFound}/>
+              <Route render={({ location }) => {
+                Notifications.error(`Unknown location ${location.pathname}, redirecting to main page.`);
```
jim-docker, line 214: nit: what is "main page"? What about "overview page"? Covers both cluster overview and workloads overview

jim-docker (approved):

src/renderer/cluster-frame.tsx
Nokel81: Sure

ixrock (approved):

src/renderer/cluster-frame.tsx
Nokel81: I'll do that in a follow up

PR #4387, opened 2021-11-19
Fix race condition in LensViews.handleVisibleClusterChange

Nokel81:
- Clear the previous when() disposer as soon as possible so that it
  won't fire, even in the no-cluster case


aleksfront (approved):
LGTM, didn't notice abandoned cluster iframe within this PR.

PR #4388, opened 2021-11-19
Retry watch if timeout set, but request was not retried

panuhorsmalahti:
If caller sets a timeout for kube-api watch, we expect the request to be ended and retried as before. However, if network is offline the request never ends and is stuck. This adds a safety timeout which will retry after the timeout.

### DIFF
--- a/src/common/k8s-api/kube-api.ts
+++ b/src/common/k8s-api/kube-api.ts
@@ -559,8 +559,8 @@ export class KubeApi<T extends KubeObject> {
 
     responsePromise
       .then(response => {
-        // True if the watch was retried
-        let retried = false;
+        // True if the current watch request was retried
+        let requestRetried = false;
 
         if (!response.ok) {
           logger.warn(`[KUBE-API] watch (${watchId}) error response ${watchUrl}`, { status: response.status });
@@ -573,7 +573,7 @@ export class KubeApi<T extends KubeObject> {
         if (timeout) {
           setTimeout(() => {
             // We only retry if we haven't retried, haven't aborted and haven't received k8s error
-            if (retried || abortController.signal.aborted || errorReceived) {
+            if (requestRetried || abortController.signal.aborted || errorReceived) {
               return;
             }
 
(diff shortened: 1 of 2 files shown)


src/common/k8s-api/kube-api.ts
```diff
@@ -528,35 +528,78 @@ export class KubeApi<T extends KubeObject> {
...
+              // TODO: Reduce logging here
+              logger.info(`[KUBE-API] Watch timeout set, returning due to retried: ${retried} aborted: ${abortController.signal.aborted} errorReceived: ${errorReceived}`);
+
+              return;
+            }
+
+            // Close current request, how?
+            // If we call abortController.abort();, that will abort the next request too,
+            // as abortController is passed down
+            // abortController.abort();
```
panuhorsmalahti: TODO: Need to figure out solution for this.

Nokel81: How about wrapping the abort controller and recreating it after you call abort so that the next time around a new one is passed down?

panuhorsmalahti: Thanks, I was thinking of something like that also.

panuhorsmalahti: Fixed in https://github.com/lensapp/lens/pull/4388/commits/64a9c61ffa42258f81f63991a2722f12204d3cbf

src/common/k8s-api/kube-api.ts
```diff
@@ -528,35 +530,79 @@ export class KubeApi<T extends KubeObject> {
...
+      signal: abortController.signal,
+      timeout: 600_000,
+    });
 
     logger.info(`[KUBE-API] watch (${watchId}) ${retry === true ? "retried" : "started"} ${watchUrl}`);
 
     responsePromise
       .then(response => {
+        // True if the watch was retried
+        let retried = false;
```
jakolehm: Just wondering why we need this, we are already passing `retry: true` to watch calls when we retry.

panuhorsmalahti: This variable is only about the current request, e.g. has this one been already retried. It's to prevent retrying the same watch multiple (i.e. 2) times. It's a bit confusing I agree, I modified the variable name and comment to be more clear.

jakolehm (dismissed):
LGTM

msa0311 (dismissed):

jakolehm (approved):

msa0311 (approved):

PR #4389, opened 2021-11-19
Stop retry if system suspended, resume watch when system is resume

chenhunghan:
- Stop retry if getting `"end", "close", "error"` AND system suspended.
- Resume watch if system is resume.

Screenshot 2021-11-19 at 16 37 08


src/common/k8s-api/kube-api.ts
```diff
@@ -38,6 +38,17 @@ import type { RequestInit } from "node-fetch";
...
+import electron from "electron";
+
+const onceSystemResume = () => 
+  new Promise(resolve => {
+    const handler = () => {
+      electron.powerMonitor.removeListener("resume", handler);
+      resolve(true);
+    };
+
+    electron.powerMonitor.on("resume", handler);
```
Nokel81: Why not "once"?

chenhunghan: What do you mean?

Nokel81: I mean you do `electron.powerMonitor.on("resume", handler)` but then in the handler you call `removeListener` in `handler`. I think it would be clear if this code was just:

```ts
function onceSystemResumes(): Promise<true> {
  return new Promise(resolve => electron.powerMonitor.once("resume", () => resolve(true)));
}
```

chenhunghan: Oh, ok, sounds good

src/common/k8s-api/kube-api.ts
```diff
@@ -542,6 +571,33 @@ export class KubeApi<T extends KubeObject> {
...
+    if (!this.watchingSystemStatus) {
+      this.watchSystemStatus();
+    }
+
+    onceSystemSuspend().then(async (suspended) => {
+      if (suspended) {
+        logger.info(`[KUBE-API] system suspended, abort watching of ${watchUrl}...`);
+
+        try {
+          abort();
```
chenhunghan: Got this when calling `abort()`, not sure why...
```
[0] error:   ┏ [KUBE-API] error aborting watch (devcluster-1) Expected 'this' to be an 'AbortController' object, but got undefined +0ms
[0] error:   ┃ [1] TypeError: Expected 'this' to be an 'AbortController' object, but got undefined
[0] error:   ┃ [2]     at getSignal (webpack:///./node_modules/abort-controller/dist/abort-controller.js?:104:15)
[0] error:   ┃ [3]     at abort (webpack:///./node_modules/abort-controller/dist/abort-controller.js?:91:21)
[0] error:   ┃ [4]     at KubeApi.eval (webpack-internal:///./src/common/k8s-api/kube-api.ts:450:11)
[0] error:   ┃ [5]     at Generator.next (<anonymous>)
[0] error:   ┃ [6]     at eval (webpack-internal:///./src/common/k8s-api/kube-api.ts:68:61)
[0] error:   ┃ [7]     at new Promise (<anonymous>)
[0] error:   ┃ [8]     at __async (webpack-internal:///./src/common/k8s-api/kube-api.ts:52:10)
[0] error:   ┗ [9]     at eval (webpack-internal:///./src/common/k8s-api/kube-api.ts:446:45)
```

chenhunghan: Fix it by using `node-abort-controller` in extension.

panuhorsmalahti: I think the `this` reference is wrong because the abort property is extracted from the class instance. In my related PR (https://github.com/lensapp/lens/pull/4388)  I just call the method using `abortController` so the `this` is correct.

panuhorsmalahti: I suggest fixing it in the same way in this PR too (or just wait until the other PR is merged and fix conlifcts).

PR #4393, opened 2021-11-19
Catch and display load errors for WorkloadsOverview

### LINKED ISSUE #4344: One inaccessible namespace blocks Workloads Overview for all namespaces
https://user-images.githubusercontent.com/40840436/141815807-f7d25fd1-2610-4691-89a1-9fec6330e371.mov

In this example the user can list namespaces but only has access to default namespace. If default and any other namespaces are selected in the namespace filter then the statistics all show 0

similar to #4192

Nokel81:
- Move the select filter to the top level since that is where the
  subscribe call is

fixes #4344

### DIFF
--- a/src/extensions/registries/workloads-overview-detail-registry.ts
+++ b/src/extensions/registries/workloads-overview-detail-registry.ts
@@ -19,11 +19,12 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
+import { orderBy } from "lodash";
 import type React from "react";
 import { BaseRegistry } from "./base-registry";
 
 export interface WorkloadsOverviewDetailComponents {
-  Details: React.ComponentType<any>;
+  Details: React.ComponentType<{}>;
 }
 
 export interface WorkloadsOverviewDetailRegistration {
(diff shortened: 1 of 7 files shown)


jim-docker (commented):
Something fishy:
Screen Shot 2021-11-19 at 2 42 21 PM

User does not have namespace list permission, accessible namespaces are set to `default`, `freddy`, and `garbage`. User has access only to `default`, `garbage` namespace does not exist. 

With just `default` selected I would not expect any warnings

Nokel81:
@jim-docker Good catch, thanks. I also fixed the same issue on the `<KubeObjectListLayout />` because it exists there too.

src/extensions/registries/workloads-overview-detail-registry.ts
```diff
@@ -19,22 +19,29 @@
...
   getItems() {
-    const items = super.getItems();
+    return orderBy(super.getItems(), "priority", "desc");
+  }
+
+  protected getRegisteredItem(item: WorkloadsOverviewDetailRegistration): RegisteredWorkloadsOverviewDetail {
+    const { priority = 50, ...rest } = item;
 
-    return items.sort((a, b) => (b.priority ?? 50) - (a.priority ?? 50));
+    return { priority, ...rest };
```
jim-docker, line 45: this looks unrelated to the issue, did you test that it still works?

src/renderer/components/+workloads-overview/overview.tsx
```diff
@@ -33,36 +33,77 @@ import { replicaSetStore } from "../+workloads-replicasets/replicasets.store";
...
     disposeOnUnmount(this, [
       kubeWatchApi.subscribeStores([
         podsStore, deploymentStore, daemonSetStore, statefulSetStore, replicaSetStore,
         jobStore, cronJobStore, eventStore,
       ], {
-        preload: true,
-        namespaces: clusterContext.contextNamespaces,
+        onLoadFailure: error => this.loadErrors.push(String(error)),
+      }),
+      reaction(() => clusterContext.contextNamespaces.slice(), () => {
```
jim-docker, line 64: Why can't it just be `() => clusterContext.contextNamespaces`?

jim-docker (approved):

src/extensions/registries/workloads-overview-detail-registry.ts
Nokel81: Yes I did. With the kube-resource-map extension installed (and the default overview doesn't have `priority` defined) the order is the exact same as before.

src/renderer/components/+workloads-overview/overview.tsx
Nokel81: Because then it only tracks assignments to the `clusterContext.contextNamespaces` field itself, and not internal changes.

jim-docker: Ah, right. You have to "dot" into it. Ugh.

PR #4396, opened 2021-11-19
Improve local shell CWD setting

Nokel81:

Just some quality of life improvements.
Screen Shot 2021-11-19 at 4 10 19 PM
Screen Shot 2021-11-19 at 4 10 14 PM
Screen Shot 2021-11-19 at 4 10 10 PM

### DIFF
--- a/src/renderer/components/cluster-settings/components/cluster-local-terminal-settings.tsx
+++ b/src/renderer/components/cluster-settings/components/cluster-local-terminal-settings.tsx
@@ -19,9 +19,8 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import React from "react";
-import { observable, autorun, makeObservable } from "mobx";
-import { observer, disposeOnUnmount } from "mobx-react";
+import React, { useEffect, useState } from "react";
+import { observer } from "mobx-react";
 import type { Cluster } from "../../../../main/cluster";
 import { Input } from "../../input";
 import { SubTitle } from "../../layout/sub-title";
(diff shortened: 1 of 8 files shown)


src/renderer/components/cluster-settings/components/cluster-local-terminal-settings.tsx
```diff
@@ -53,8 +59,45 @@ export class ClusterHomeDirSetting extends React.Component<Props> {
...
+          <>
+            <b>Shell Working Directory</b>
+            <p>Provided path does not exist, your changes were not saved.</p>
+          </>,
+        );
+      } else {
+        Notifications.error(
+          <>
+            <b>Shell Working Directory</b>
+            <p>Your changes were not saved due to the error bellow</p>
```
jansav: Typo: `bellow` -> `below`

src/renderer/components/cluster-settings/components/cluster-local-terminal-settings.tsx
```diff
@@ -53,8 +59,45 @@ export class ClusterHomeDirSetting extends React.Component<Props> {
...
+      }
+    } catch (error) {
+      if (error.code === "ENOENT") {
+        Notifications.error(
+          <>
+            <b>Shell Working Directory</b>
+            <p>Provided path does not exist, your changes were not saved.</p>
+          </>,
+        );
+      } else {
```
jansav, line 137: Shouldn't suppress unknown exceptions here. It hides information, prevents centralized error handling and logging etc.

jansav:
There are no unit tests. We (me and @Iku-turso) are eager to help in making this happen :)

src/renderer/components/cluster-settings/components/cluster-local-terminal-settings.tsx
Nokel81: How is this suppressing unknown errors, do you mean I should handle each FS error code separately?

Nokel81:
@jansav I have added some basic unit tests.

src/renderer/components/cluster-settings/components/cluster-local-terminal-settings.tsx
jansav: I mean that unknown exceptions are unknown for a reason. We should treat them as fatal exceptions which have to be fixed from our side.

So basically, separate fatal exceptions and non fatal exceptions from each other.

- Fatal exception is something unexpected that we need to know about. User has no way to fix it. 
- Non fatal exception is something expected that we don't need to know about. User has to fix it, since it's part of the non-happy case scenario. In this case, e.g. providing path to directory that doesn't exist is "non fatal exception", since user can simply change the directory.  (as you have done)

I can see that you have done changes to the logic handling exceptions, and it's getting better. Most important part for now, is that the unknown exceptions are reported somewhere so that we can start treating them as **bugs to be fixed**. 

In the future, I think that we should start catching the fatal exceptions from top level and using throw as a mechanism since then we have most detailed error stack available when we are trying to fix the issue. But yeah, that's long in the future, since it would mean that we need to get rid of larger scope "try/catch"-usages.

jansav:
> @jansav I have added some basic unit tests.

Great!

jansav (approved):
I don't see the big picture behind the improvements (e.g. why these changes are introduced?), so someone else should check it too.

Anyway, changes that I pointed out are done, so I'm happy.

src/renderer/components/cluster-settings/components/cluster-local-terminal-settings.tsx
Nokel81: We cannot really treat them as fatal since the application should continue to be usable, I would prefer "expected exceptions" vs "unexpected exceptions" (which I know sounds like an oxymoron). 

In this I don't really like the design of JS, and much prefer Rust's design, where errors are just data. It is slightly hard to do that in JS since the entire ecosystem uses `throw`

Nokel81:
> I don't see the big picture behind the improvements (e.g. why these changes are introduced?), so someone else should check it too.

I goal of this code was two fold:

1. Allow the user to use a folder picker for the path instead of having to type it out manually (and the clear button felt like a natural extension of that)
2. Improve the handling of CWD settings for the local shell sessions so that users never get `chdir` errors.

PR #4400, opened 2021-11-21
fix: use non-deprecated delete flag with node drain

mrkshields:
# Problem

`kubectl drain` calls a deprecated flag, `--delete-local-data`, per https://github.com/kubernetes/kubernetes/pull/95076.

# Solution

Use the designated replacement.

# Result

No warning errors about the `--delete-local-data` flag deprecation on node drain.

### DIFF
--- a/extensions/node-menu/src/node-menu.tsx
+++ b/extensions/node-menu/src/node-menu.tsx
@@ -77,7 +77,7 @@ export function NodeMenu(props: NodeMenuProps) {
   };
 
   const drain = () => {
-    const command = `${kubectlPath} drain ${nodeName} --delete-local-data --ignore-daemonsets --force`;
+    const command = `${kubectlPath} drain ${nodeName} --delete-emptydir-data --ignore-daemonsets --force`;
 
     ConfirmDialog.open({
       ok: () => sendToTerminal(command),


Nokel81 (approved):

Nokel81 (changes requested):
This might not exist on older versions of kubectl. This should be changed depending on the version.

mrkshields:
> This might not exist on older versions of kubectl. This should be changed depending on the version.

Ok. What changes are you specifically looking for, @Nokel81 ?

It's worth noting the `--delete-local-data` flag is deprecated as of v1.20 per the link I provided, and versions prior to v1.20 are not listed under the Release History heading on https://kubernetes.io/releases/.

It does, however, mention:

> Kubernetes 1.19 and newer receive approximately 1 year of patch support

Which implies support for the deprecated flag ends in less than a year.

Nokel81:
@laebshade Yes they are not supported but we download an appropriate kubectl version based on the kube version of a cluster.

So I would get the active entity in that menu, look at the kube version metadata and switch which arg is used based on that.

mrkshields:
Thanks for clarifying.

I’m unable to prioritize your requested changes.

Do you wish to close this PR, pick up ownership, or hand off your requested
changes to some one else specifically?

Nokel81:
If you are unable to spend the time now, could you open an issue describing this issue before you close the PR. That way we can track it for planning.

mrkshields:
No thank you

PR #4404, opened 2021-11-23
Add unit tests for BaseStore

jakolehm:
- adds unit tests for BaseStore
- removes misleading `async` from `onModelChange` & `saveToFile`

### DIFF
--- a/src/common/__tests__/base-store.test.ts
+++ b/src/common/__tests__/base-store.test.ts
@@ -1,0 +1,171 @@
+/**
+ * Copyright (c) 2021 OpenLens Authors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import mockFs from "mock-fs";
+import { AppPaths } from "../app-paths";
+import { BaseStore } from "../base-store";
+import { action, comparer, makeObservable, observable, toJS } from "mobx";
+import { readFileSync } from "fs";
+
+AppPaths.init();
+
+jest.mock("electron", () => ({
+  app: {
+    getVersion: () => "99.99.99",
+    getName: () => "lens",
+    setName: jest.fn(),
+    setPath: jest.fn(),
+    getPath: () => "tmp",
+    getLocale: () => "en",
+    setLoginItemSettings: jest.fn(),
+  },
+  ipcMain: {
+    handle: jest.fn(),
+    on: jest.fn(),
+    removeAllListeners: jest.fn(),
+    off: jest.fn(),
+    send: jest.fn(),
+  },
+}));
+
+interface TestStoreModel {
+  a: string;
+  b: string;
+  c: string;
+}
+
+class TestStore extends BaseStore<TestStoreModel> {
+  @observable a: string;
+  @observable b: string;
+  @observable c: string;
+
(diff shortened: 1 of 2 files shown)


msa0311 (dismissed):

src/common/__tests__/base-store.test.ts
```diff
@@ -0,0 +1,171 @@
...
+interface TestStoreModel {
+  a: string;
+  b: string;
+  c: string;
+}
+
+class TestStore extends BaseStore<TestStoreModel> {
+  @observable a: string;
+  @observable b: string;
+  @observable c: string;
```
ixrock, line 58: Shouldn't class be marked with `@observer`?

ixrock:
Might be the issue related to https://github.com/sindresorhus/conf/issues/114 ?

Currently `BaseStore.saveToFile()` looks like this:

```ts
  protected async saveToFile(model: T) {
    logger.info(`[STORE]: SAVING ${this.path}`);

    // todo: update when fixed https://github.com/sindresorhus/conf/issues/114
    if (this.storeConfig) {
      for (const [key, value] of Object.entries(model)) {
        this.storeConfig.set(key, value);
      }
    }
  }
```

I don't know how internally `conf` work, but here we have to create multiple savings to file instead of one `model` update.

src/common/base-store.ts
```diff
@@ -168,7 +168,7 @@ export abstract class BaseStore<T> extends Singleton {
 
   protected async onModelChange(model: T) {
     if (ipcMain) {
-      this.saveToFile(model); // save config file
+      await this.saveToFile(model); // save config file
```
jansav: I feel like this isn't the solution here. For me, it looks that the real issue is that `onModelChange` is called multiple times. Not sure though.

ixrock: If writing to the storage file is `sync` why it could be a an issue?

jakolehm: > If writing to the storage file is `sync` why it could be a an issue?

`onModelChange` can be triggered multiple times, if we don't `await` then the order of execution of `sync` writes might happen out of order.

ixrock: Ok. What if remove `async` from `onModelChange`? And probably from `saveToFile`.

jakolehm: > Ok. What if remove `async` from `onModelChange`? And probably from `saveToFile`.

It works as well but it modifies api behaviour (this class is exposed via extension api).

src/common/__tests__/base-store.test.ts
jakolehm: I think that's only needed for React classes.

ixrock: Yeah, right.

src/common/base-store.ts
Nokel81: > It works as well but it modifies api behaviour (this class is exposed via extension api).

No, making a previously `async` function `sync` is not a breaking API change in JS. Since you can `await` any value.

Nokel81: I would say that making both `onModelChange` and `saveToFile` sync would be the best.

ixrock: @Nokel81 I think it's still would be a breaking change if the consumer of `BaseStore` (e.g. some extension) override one of these methods ^ and start using `await` inside. If we decide to make them `sync` in addition I suggest to mark them as `private`.

Nokel81: If they override them, they can use `async` if they want:

https://www.typescriptlang.org/play?#code/MYGwhgzhAEBiD29oG8BQ1oCMwCcAUAlAFzQBu8AlgCYrobTDwB2E8IApgHQjwDmeAIgpM4iTthwCCAbjoBfVAtShIMAEJgALtHYAPTeyZUYCJGgyQAnk2BZchEgAUc8ALYUI7ADzlqAPlp6BmZWDm4+QWFoDU1xXClZDAU5IA

Furthermore, it will always be a race condition because the function is being called within a `reaction`

ixrock: Yes, you can override but in this case parent methods becomes `async` when you call inside `super.method()` (but we expect it to be `sync`)

> Furthermore, it will always be a race condition because the function is being called within a reaction

All mobx reactions by default are `sync`. If we remove `async` from `onModelChange` how it could be always race condition?

Nokel81: >  in this case parent methods becomes async when you call inside super.method() 

I don't understand what you mean by this. Do you mean that when we call the method (say in the `reaction`) the methods will still be `async` if they have been overrided as `async`?

I agree that `onModelChange` and `saveToFile` should both be `private` but that is definitely a breaking change. Though it is one that I think we can accept.

ixrock: @Nokel81 yes, probably you're right. 
Initially I thought there could be some timing issues, but maybe not..
This code produces expected result:

```tsx
class Foo {
  constructor(){
    this.bar();
  }

  bar(): void {
    console.log("in Foo.bar");
  }
}

class Bat extends Foo {
  async bar(): Promise<void> {
    await new Promise(resolve => setTimeout(resolve, 500));
    super.bar();
    console.log("in Bat.bar");
  }
}

new Bat();
```

**OUTPUT:**
```
[LOG]: "in Foo.bar" 
[LOG]: "in Bat.bar" 
```

jakolehm: Outcome of this discussion: we need to remove async from these functions so that reaction will be sync.

@ixrock @Nokel81 right?

Nokel81: That is what I believe yes.

ixrock: @jakolehm yes, it could be best solution I guess.

jakolehm: I have been testing different things and I'm quite certain now that this PR mainly just adds tests for BaseStore, nothing actually changed in the execution order. I still removed `async` because it's quite misleading in this context.

Nokel81 (approved):

PR #4406, opened 2021-11-23
Persist apiVersion when editing resources in monaco

### LINKED ISSUE #7474: Add remembering which apiVersion a resource was edited in during last save
- As a user I would like for the `apiVersion` of a kube resource to be consistent between saving an edit within Lens and then editing the same resource again

**Why is this needed**:
- If I use Lens to upgrade a resource to a newer `apiVersion` (which Lens does not use by default because it isn't the version that kube API reports as "default") I would like to continue to review that resource in that same version.

Nokel81:
- Use a new custom k8slens prefixed label

- Means that users aren't surprised when they use lens to update a
  resource to a new apiVersionWithGroup

- Doesn't touch the versions in the stores


src/renderer/components/monaco-editor/monaco-editor.tsx
```diff
@@ -186,7 +186,14 @@ export class MonacoEditor extends React.Component<MonacoEditorProps> {
     if (!this.containerElem || this.editor || this.unmounting) {
       return;
     }
-    const { language, theme, readOnly, value: defaultValue } = this.props;
+    const {
```
ixrock: How this file changes related to PR's topic?

Nokel81: Specifically the addition of the `fireImmediately` on line 220 below is needed. Without it the contents of the editor wasn't updating correctly.

src/common/k8s-api/kube-object.ts
```diff
@@ -197,25 +206,8 @@ export class KubeObject<Metadata extends KubeObjectMetadata = KubeObjectMetadata
...
-  ];
-  private static readonly nonEditablePaths = new Set([
-    "/apiVersion",
-    "/kind",
-    "/metadata/name",
-    "/metadata/selfLink",
-    "/metadata/resourceVersion",
-    "/metadata/uid",
-    ...KubeObject.nonEditablePathPrefixes,
-  ]);
```
Nokel81: With patch, this isn't necessary anymore because the kube server will enforce this.

src/common/k8s-api/kube-api.ts
```diff
@@ -270,16 +268,16 @@ export class KubeApi<T extends KubeObject> {
...
     for (const apiUrl of apiBases) {
-      // Split e.g. "/apis/extensions/v1beta1/ingresses" to parts
-      const { apiPrefix, apiGroup, apiVersionWithGroup, resource } = parseKubeApi(apiUrl);
-
-      // Request available resources
       try {
-        const response = await this.request.get<IKubeResourceList>(`${apiPrefix}/${apiVersionWithGroup}`);
+        // Split e.g. "/apis/extensions/v1beta1/ingresses" to parts
+        const { apiPrefix, apiGroup, apiVersionWithGroup, resource } = parseKubeApi(apiUrl);
+        const url = `${apiPrefix}/${apiVersionWithGroup}`;
```
Nokel81: I can move this change out to a different PR if necessary, but this fixes a bug with fallback as fallback would fail completely if the `apiUrl` is not valid.

src/renderer/components/monaco-editor/monaco-editor.tsx
ixrock: Okay, so then let's leave only `fireImmediately` here :)

Nokel81: Okay

src/renderer/components/monaco-editor/monaco-editor.tsx
```diff
@@ -225,12 +225,10 @@ class NonInjectedMonacoEditor extends React.Component<MonacoEditorProps & Depend
 
     this.dispose.push(
       reaction(() => this.model, this.onModelChange),
-      reaction(() => this.theme, theme => {
-        if (theme) {
-          editor.setTheme(theme);
-        }
+      reaction(() => this.theme, editor.setTheme),
```
ixrock: Should we run it with `{fireImmediately: true}` too?

ixrock (dismissed):
LGTM

src/renderer/components/monaco-editor/monaco-editor.tsx
Nokel81: I don't think so, because we set it above. I don't quite remember why I changed the other one to fireImmediately either.

gabriel-mirantis (approved):

PR #4411, opened 2021-11-24
Fix cluster dropdown invisible overlay

### LINKED ISSUE #4075: <Animate/> transition won't finish if window looses it's focus
Sometimes (noticed only with external monitors) dialog close animation breaks down causing `<Dialog/>` to freeze in semi-transparent state. This happens when window looses it's focus during the animation.

Most likely, the problem is located on this lines inside `animate.tsx`:

```
return React.cloneElement(contentElem, {
      className: cssNames("Animate", name, contentElem.props.className, this.statusClassName),
      children: this.isVisible ? contentElem.props.children : null,
      onTransitionEnd: this.onTransitionEnd,
});
```

`onTransitionEnd` event won't get fired, but this is responsible to closing the dialog.

### LINKED ISSUE #4398: Cluster Dashboard navigation compromised
This is how I reproduced this, there are probably other ways. ~~I think it just needs to involve a spaces cluster (not a local cluster)~~ reproduced with local cluster (that is at least shared through spaces)

1. For a spaces cluster add kube-prometheus helm chart (via Apps -> Charts)
2. go to releases tab
3. click on other tabs/drawers (in the following video Cluster was clicked on, which brings up the cluster details, not the cluster metrics overview).
4. Clicking on Nodes does nothing, then clicking workload -> overview brings up the Remove Cluster dialog (second video)

https://user-images.githubusercontent.com/40840436/142698692-ed8815b1-0182-424c-866d-9c863231fdd9.mov

https://user-images.githubusercontent.com/40840436/142698699-c665b0b3-8db8-4137-8359-9a1bc2de5ecb.mov

Actually it may be that the cluster title dropdown menu is reacting to clicks (even though the dropdown menu is not visible)??

5.3.0-beta.2, tried on Windows and Linux

aleksfront:
Replacing `onTransitionEnd` to `setTimeout` for triggering hide event inside `<Animate/>` component. The reason is unstable nature of `onTransitionEnd`: previously, if iframe looses it's focus, `onTransitionEnd` won't be fired leaving Menu in the DOM.

Fixes #4398 
Fixes #4075 

https://user-images.githubusercontent.com/9607060/143219362-f6e27017-89fd-40b2-9b1f-714c15d27bf9.mov

### DIFF
--- a/src/renderer/components/animate/animate.scss
+++ b/src/renderer/components/animate/animate.scss
@@ -21,43 +21,43 @@
 
 // Animations
 
-@mixin animate-opacity($enterDuration: 100ms, $leaveDuration: 150ms) {
+@mixin animate-opacity() {
   opacity: 0;
 
   &.enter {
     transition-property: opacity;
-    transition-duration: $enterDuration;
+    transition-duration: var(--enter-duration);
     opacity: 1;
   }
 
   &.leave {
-    transition-duration: $leaveDuration;
+    transition-duration: var(--leave-duration);
     transition-timing-function: ease-out;
     opacity: 0;
   }
 }
 
-@mixin animate-slide-right($enterDuration: 100ms, $leaveDuration: 150ms) {
+@mixin animate-slide-right() {
   transform: translateX(100%);
   will-change: transform;
 
   &.enter {
     transform: translateX(0);
-    transition: transform $enterDuration;
+    transition: transform var(--enter-duration);
     transition-timing-function: ease-in-out;
   }
 
   &.leave {
     transform: translateX(100%);
-    transition: transform $leaveDuration;
+    transition: transform var(--leave-duration);
   }
 }
 
-@mixin animate-opacity-scale($enterDuration: 250ms, $leaveDuration: 150ms) {
+@mixin animate-opacity-scale() {
   opacity: 0;
 
   &.enter {
-    transition: opacity $enterDuration;
+    transition: opacity var(--enter-duration);
     opacity: 1;
   }
 
@@ -65,7 +65,7 @@
     will-change: opacity, transform;
     opacity: 0;
     transform: scale(1.25);
-    transition: transform $leaveDuration ease-in, opacity $leaveDuration ease-out;
+    transition: transform var(--leave-duration) ease-in, opacity var(--leave-duration) ease-out;
   }
 }
 
@@ -84,6 +84,6 @@
   }
 
   &.opacity-scale {
-    @include animate-opacity-scale(100ms);
+    @include animate-opacity-scale;
   }
 }
--- a/src/renderer/components/animate/animate.scss
+++ b/src/renderer/components/animate/animate.scss
@@ -21,43 +21,43 @@
 
 // Animations
 
-@mixin animate-opacity() {
+@mixin animate-opacity($enterDuration: var(--enter-duration), $leaveDuration: var(--leave-duration)) {
   opacity: 0;
 
   &.enter {
     transition-property: opacity;
-    transition-duration: var(--enter-duration);
+    transition-duration: $enterDuration;
     opacity: 1;
   }
 
   &.leave {
-    transition-duration: var(--leave-duration);
+    transition-duration: $leaveDuration;
     transition-timing-function: ease-out;
     opacity: 0;
   }
 }
 
-@mixin animate-slide-right() {
+@mixin animate-slide-right($enterDuration: var(--enter-duration), $leaveDuration: var(--leave-duration)) {
   transform: translateX(100%);
   will-change: transform;
 
   &.enter {
     transform: translateX(0);
-    transition: transform var(--enter-duration);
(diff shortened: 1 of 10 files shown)


src/renderer/components/animate/animate.scss
```diff
@@ -21,51 +21,51 @@
 
 // Animations
 
-@mixin animate-opacity($enterDuration: 100ms, $leaveDuration: 150ms) {
+@mixin animate-opacity() {
```
ixrock: Why removed from args?

```suggestion
@mixin animate-opacity($enterDuration: var(--enter-duration), $leaveDuration: var(--leave-duration)) {
```

src/renderer/components/animate/animate.tsx
```diff
@@ -92,27 +98,21 @@ export class Animate extends React.Component<AnimateProps> {
...
-    }
-  }
-
   render() {
-    const { name } = this.props;
+    const { name, enterDuration, leaveDuration } = this.props;
     const contentElem = this.contentElem;
+    const durations = {
+      "--enter-duration": `${enterDuration}ms`,
+      "--leave-duration": `${leaveDuration}ms`,
```
ixrock, line 106: _nit:_ I would better to avoid using css-var names inside `*.js` (same as we usually avoid inline-styles)
However if you want to support custom values from props to `enter` & `leave` probably this is fine..

src/renderer/components/dialog/dialog.tsx
```diff
@@ -167,7 +167,7 @@ export class Dialog extends React.PureComponent<DialogProps, DialogState> {
 
     if (animated) {
       dialog = (
-        <Animate enter={this.isOpen} name="opacity-scale">
+        <Animate enter={this.isOpen} name="opacity-scale" enterDuration={200} leaveDuration={200}>
```
ixrock: Why `200` for dialogs? 
I'd better to use predefined values inside `*.css` for all animations and apply to element via `className`.

ixrock (dismissed):

src/renderer/components/animate/animate.tsx
aleksfront: This is done because I need duration values inside js for this line https://github.com/lensapp/lens/pull/4411/files/88d2e6667859db508963b9b48ce53cf30427e7ad#diff-7e120b4e5eff8b3b877c4f2ea13344713cec8bec98512c4d3b232383398db09bR92

We can do the opposite, trying to fetch css var from the sass file but it's complicated within `<Animate/>` since no proper refs can be assigned there.

src/renderer/components/dialog/dialog.tsx
aleksfront: Sure, let's use our basic predefined values.

src/renderer/components/animate/animate.tsx
ixrock: > since no proper refs can be assigned there.

Didn't get what you mean.. We can define something like `--animation-leave-ms: 100;` inside css and read the value in `*.js` with `utils/cssVar`. Wdyt? (better separation of concerns imho)

ixrock: You can use probably this util to find proper ref if I got you correctly:
```tsx
import {findDOMNode} from "react-dom"
```

Nokel81: Do we want to be using `findDOMNode` more?

ixrock: @Nokel81 something wrong with it? It helps in some cases.. and it's not deprecated.

ixrock: Right, it's gonna be deprecated and it's already shows warnings in strict mode: https://reactjs.org/docs/strict-mode.html#warning-about-deprecated-finddomnode-usage

ixrock (approved):

src/renderer/components/animate/animate.tsx
Nokel81: It has been deprecated in strict mode and does make unit testing harder.

jim-docker (approved):
LGTM, can't reproduce either linked issue

jim-docker:
I just encountered this while trying to reproduce #4075:

```
[0] error:   ▪ failed to request selfSubjectAccessReview: Error: connect ECONNREFUSED 127.0.0.1:51090 +2s
```

At this point Lens had no access to any kube resources in my minikube cluster (kubectl from external terminal still worked). But it's hard to imagine this PR causes that.

src/renderer/components/animate/animate.tsx
```diff
@@ -84,6 +85,11 @@ export class Animate extends React.Component<AnimateProps> {
     if (!this.isVisible) return;
     this.statusClassName.leave = true;
     this.props.onLeave();
+    this.resetAfterLeaveDuration();
+  }
+
+  resetAfterLeaveDuration() {
```
jim-docker, line 91: should something be bound here?

PR #4414, opened 2021-11-24
Fix disconnect on currently connecting cluster

Nokel81:

This PR fixes the following bugs:

1. If you use the hotbar to disconnect from a cluster while it is connecting. It now correctly navigates you back to the catalog
2. If connecting is stuck and you navigate away and back to the same cluster, the cluster is now `forced` when activating

### DIFF
--- a/src/common/cluster/cluster.ts
+++ b/src/common/cluster/cluster.ts
@@ -4,7 +4,7 @@
  */
 
 import { ipcMain } from "electron";
-import { action, comparer, computed, makeObservable, observable, reaction, when } from "mobx";
+import { action, comparer, computed, makeObservable, observable, reaction, runInAction, when } from "mobx";
 import { broadcastMessage } from "../ipc";
 import type { ContextHandler } from "../../main/context-handler/context-handler";
 import { HttpError, KubeConfig } from "@kubernetes/client-node";
@@ -287,7 +287,7 @@ export class Cluster implements ClusterModel, ClusterState {
    * @internal
    */
   protected bindEvents() {
-    logger.info(`[CLUSTER]: bind events`, this.getMeta());
+    logger.info(`[CLUSTER]: bind events`, this.getMeta(true));
     const refreshTimer = setInterval(() => !this.disconnected && this.refresh(), 30000); // every 30s
     const refreshMetadataTimer = setInterval(() => !this.disconnected && this.refreshMetadata(), 900000); // every 15 minutes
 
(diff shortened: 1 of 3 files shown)


jim-docker:
> 1. If you use the hotbar to disconnect from a cluster while it is connecting. It now correctly navigates you back to the catalog

how do you initiate disconnect while it is connecting? The hotbar context menu offers `Connect`

Nokel81:
>  The hotbar context menu offers Connect

Interesting... that must be a bug because it should offer `disconnect` let me see if I can reproduce/do about it.

Nokel81:
> how do you initiate disconnect while it is connecting? The hotbar context menu offers Connect

It is a race condition, you need to keep opening and closing the menu. I was able to reproduce this on master on the second attempt.

jim-docker (commented):
I'm seeing some strange behaviour, when I disconnect while it is connecting I find that the cluster is not always actually disconnected. `kubectl proxy` process is still running, as are any port-forwards:
```
76328 ttys006    0:00.48 /Users/jimehrismann/git/lensapp/lens/binaries/client/darwin/x64/kubectl --kubeconfig /var/folders/3d/vq292d5n71gbkjs0cnwzw3280000gq/T/kubeconfig-87bac5783b9b6bb49d42137caf351757 port-forward -n default pod/my-nginx-589f8bd8bc-drldj 55835:80
76329 ttys006    0:00.45 /Users/jimehrismann/git/lensapp/lens/binaries/client/darwin/x64/kubectl --kubeconfig /var/folders/3d/vq292d5n71gbkjs0cnwzw3280000gq/T/kubeconfig-87bac5783b9b6bb49d42137caf351757 port-forward -n nginx-ingress service/nginx-ingress 53335:80
76330 ttys006    0:00.45 /Users/jimehrismann/git/lensapp/lens/binaries/client/darwin/x64/kubectl --kubeconfig /var/folders/3d/vq292d5n71gbkjs0cnwzw3280000gq/T/kubeconfig-87bac5783b9b6bb49d42137caf351757 port-forward -n nginx-ingress service/nginx-ingress 53399:443
76337 ttys006    0:00.53 /Users/jimehrismann/Library/Application Support/OpenLensDev/binaries/kubectl/1.21.2/kubectl proxy -p 0 --kubeconfig /Users/jimehrismann/.kube/config --context minikube --accept-hosts 192.168.64.2 --reject-paths ^[^/] --api-prefix /89de87b8c8b7e
```

Also, after multiple quick connect/disconnect attempts I ended up with the connecting screen spinning "forever". And eventually got this:

Screen Shot 2021-11-29 at 5 36 54 PM 

And once the cluster's icon suddenly disappeared from the hotbar (I think it had enough of my incessant clicking).

jim-docker:
Might be easier to just not allow disconnect while it is connecting? (show Connect greyed out in the context menu?)

Nokel81:
> Might be easier to just not allow disconnect while it is connecting? (show Connect greyed out in the context menu?)

No I think it is good to support this because then there is no other way (besides starting the entire app) for retrying the connection.

jim-docker:
> > Might be easier to just not allow disconnect while it is connecting? (show Connect greyed out in the context menu?)
> 
> No I think it is good to support this because then there is no other way (besides starting the entire app) for retrying the connection.

Well, a connection attempt should timeout after a reasonable time, so that it can be retried

Nokel81:
> Well, a connection attempt should timeout after a reasonable time, so that it can be retried

Fair, but what if something else needs to be recomputed and then written to disk?

Nokel81:
Furthermore, if a connection is just failing. There is no other way to stop lens from continuing to try and connect.

Nokel81:
> I'm seeing some strange behaviour, when I disconnect while it is connecting I find that the cluster is not always actually disconnected. kubectl proxy process is still running, as are any port-forwards:

Is that on dev or prod? I was not able to see this on dev...

jim-docker:
> > I'm seeing some strange behaviour, when I disconnect while it is connecting I find that the cluster is not always actually disconnected. kubectl proxy process is still running, as are any port-forwards:
> 
> Is that on dev or prod? I was not able to see this on dev...

uhh, on dev I believe. It was very easy to reproduce. Maybe you need to have port-forwards active?

Nokel81:
> Maybe you need to have port-forwards active?

Will try with that then

Nokel81:
@jim-docker I was unable to reproduce that issue using `make dev`. I had port forwards active too.

jim-docker:
> @jim-docker I was unable to reproduce that issue using `make dev`. I had port forwards active too.

Ok I'll see if I can reproduce again

jim-docker:
I still see issues:

https://user-images.githubusercontent.com/40840436/153481296-e0439a30-e788-4cd8-a7d9-d1de075e30b7.mov

src/common/cluster/cluster.ts
```diff
@@ -320,34 +320,38 @@ export class Cluster implements ClusterModel, ClusterState {
...
+        this.ensureKubectl()
+          .catch(error => logger.warn(`[CLUSTER]: failed to download kubectl for clusterId=${this.id}`, error));
+        this.broadcastConnectUpdate("Connected, waiting for view to load ...");
+      }
 
-    this.activated = true;
-    this.pushState();
+      this.activated = true;
+      this.pushState();
+    } finally {
```
jim-docker, line 356: no catch here? I guess errors are handled higher up

Nokel81:
Will try again to fix this after the DI work (and removing of iframes) is done

PR #4415, opened 2021-11-24
Fix: display of entity name while connecting

Nokel81:

This PR fixes the display of the connecting screen to display the entity name. It also passes the cluster instance down into the comment instead of getting it from the store as it seemed cleaner that way.


ixrock (approved):
LGTM. 
Next time please provide screenshots / screencast for all UI-related PR's.

PR #4417, opened 2021-11-24
Broadcast cluster's disconnected state before stopping KubeAuthProxy

Nokel81:

Remaking the PR. After running it many times on my machine the integration tests continue to pass. Though I don't think we should merge this until after 5.3


jakolehm:
@Nokel81 integration tests seem to fail similarly as with the previous PR?

Nokel81:
@jakolehm Yes I have noticed... very strange

PR #4420, opened 2021-11-25
Use PartialDeep for patch type

panuhorsmalahti:
This makes the typing work for partial updates/patches.

### DIFF
--- a/src/common/k8s-api/kube-api.ts
+++ b/src/common/k8s-api/kube-api.ts
@@ -37,6 +37,7 @@ import type { RequestInit } from "node-fetch";
 import AbortController from "abort-controller";
 import { Agent, AgentOptions } from "https";
 import type { Patch } from "rfc6902";
+import type { PartialDeep } from "type-fest";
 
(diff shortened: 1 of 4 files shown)


src/common/k8s-api/kube-api.ts
```diff
@@ -489,7 +490,7 @@ export class KubeApi<T extends KubeObject> {
     return parsed;
   }
 
-  async patch({ name = "", namespace = "default" } = {}, data?: Partial<T> | Patch, strategy: KubeApiPatchType = "strategic"): Promise<T | null> {
+  async patch({ name = "", namespace = "default" } = {}, data?: PartialDeep<T> | Patch, strategy: KubeApiPatchType = "strategic"): Promise<T | null> {
```
jakolehm: Tests would be nice.

panuhorsmalahti: Added test

aleksfront (approved):

PR #4422, opened 2021-11-25
Allow to navigate to created resource details from notification

ixrock:
https://user-images.githubusercontent.com/6377066/143462401-d799a763-312a-4097-8f73-26018815aa86.mov

### DIFF
--- a/src/renderer/components/+extensions/extensions.tsx
+++ b/src/renderer/components/+extensions/extensions.tsx
@@ -356,7 +356,7 @@ export async function attemptInstallByInfo({ name, version, requireConfirmation 
   return attemptInstall({ fileName, dataP }, disposer);
 }
 
-async function attemptInstall(request: InstallRequest, d?: ExtendableDisposer): Promise<void | (() => void)> {
+async function attemptInstall(request: InstallRequest, d?: ExtendableDisposer): Promise<void> {
   const dispose = disposer(ExtensionInstallationStateStore.startPreInstall(), d);
   const validatedRequest = await createTempFilesAndValidate(request);
 
@@ -370,7 +370,7 @@ async function attemptInstall(request: InstallRequest, d?: ExtendableDisposer): 
   if (curState !== ExtensionInstallationState.IDLE) {
     dispose();
 
-    return Notifications.error(
+    return void Notifications.error(
       <div className="flex column gaps">
         <b>Extension Install Collision:</b>
         <p>The <em>{name}</em> extension is currently {curState.toLowerCase()}.</p>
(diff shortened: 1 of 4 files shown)


src/renderer/components/notifications/notifications.tsx
```diff
@@ -35,7 +35,7 @@ export class Notifications extends React.Component {
   public elem: HTMLElement;
 
   static ok(message: NotificationMessage) {
-    notificationsStore.add({
+    return notificationsStore.add({
```
Nokel81, line 38: might be useful to add same return to `error` and `shortInfo` and `info`

src/renderer/components/dock/create-resource.tsx
```diff
@@ -103,28 +108,40 @@ export class CreateResource extends React.Component<Props> {
...
+        });
+
+        const showDetails = () => {
+          navigate(getDetailsUrl(resourceLink));
+          close();
+        };
+
+        const close = Notifications.ok(
+          <p>
+            Resource <a onClick={prevDefault(showDetails)}>{resourceName}</a> successfully created.
```
Nokel81: maybe include the kind as well?

Nokel81 (dismissed):

src/renderer/components/dock/create-resource.tsx
ixrock: Yeah, good idea. How to format the text inside link? `%name (%kind)`?

Nokel81: Yeah that works

src/renderer/components/+extensions/extensions.tsx
```diff
@@ -356,7 +356,7 @@ export async function attemptInstallByInfo({ name, version, requireConfirmation
   return attemptInstall({ fileName, dataP }, disposer);
 }
 
-async function attemptInstall(request: InstallRequest, d?: ExtendableDisposer): Promise<void> {
```
Nokel81: What are these changes for?

src/renderer/components/dock/create-resource.tsx
ixrock: Decided to put `kind` before `name` (looks better).
Screenshot 2021-11-25 at 17 17 28

Instead of:
Screenshot 2021-11-25 at 17 16 42

src/renderer/components/+extensions/extensions.tsx
ixrock: Typescript complains about this function when `Notification.error` return function.

Nokel81: Ah I think it would be better to change line `373` below to be `return void Notifications.error(...`

ixrock: Updated

Nokel81 (approved):

PR #4424, opened 2021-11-25
Command Palette Improvements

### LINKED ISSUE #3258: Add custom resources (CRDs) to command palette
Please add custom resources (CRDs) to the command palette.

As you can see in the following screenshot **Ingresses** can be accessed from the command palette but **IngressRoutes** not.

Screen Shot 2021-07-03 at 16 29 16

If you have a lot of custom resources it might take a while to find the correct one. It would be much easier to access them from the command palette.

- Kubernetes distribution: k3s
- Desktop OS: macOS

### LINKED ISSUE #4647: CommandRegistry

Nokel81:
- Deprecate the "scope" on command registrations as it is buggy and
  generally unfixable

- Switch to mounting CommandContainer in each new frame so that there is
  no need to broadcast anything other than opening palette

- Add CRD entries to the command registry within each cluster

- Add navigate to CommandActionContext to simplify handling of root
  frame scoped URLs

fixes https://github.com/lensapp/lens/issues/3258


src/common/hotbar-store.injectable.ts
```diff
@@ -314,3 +339,10 @@ function ensureExactHotbarItemLength(hotbar: Hotbar) {
     }
   }
 }
+
+const hotbarManagerInjectable = getInjectable({
```
Iku-turso, line 24: Separate injectable and and operational code in different files. There are rare exceptions to this, but this is not the place.

src/common/hotbar-types.ts
```diff
@@ -33,14 +33,18 @@ export interface HotbarItem {
   }
 }
 
-export type Hotbar = Required<HotbarCreateOptions>;
+export type Hotbar = Required<CreateHotbarData>;
```
Iku-turso, line 36: Do we really need utility type `Required` here? If yes, name types to make the reasoning more apparent.

src/common/hotbar-store.injectable.ts
Nokel81: Okay sounds good

src/common/hotbar-types.ts
Nokel81: What you would like instead? Spelling out the field manually?

src/common/hotbar-types.ts
```diff
@@ -33,14 +33,18 @@ export interface HotbarItem {
...
 
-export interface HotbarCreateOptions {
+export interface CreateHotbarData {
   id?: string;
   name: string;
   items?: Tuple<HotbarItem | null, typeof defaultHotbarCells>;
```
Iku-turso, line 41: Not your change, but do you think we could do without this `Tuple`? The same should be doable with more simplicity.

Nokel81: Well we only want 12 items in Hotbar so I think encoding that into the type system should be good.

src/extensions/__tests__/extension-discovery.test.ts
```diff
@@ -43,6 +43,7 @@ jest.mock("../extension-installer", () => ({
     installPackage: jest.fn(),
   },
 }));
+jest.mock("fs-extra");
```
Iku-turso, line 46: FYI: instead of monkey-patching global file system, we'll soon be able to cleanly achieve the same by overriding dependencies. To the future! :)

src/common/hotbar-store.injectable.ts
Nokel81: Would you be able to explain why though?

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -78,6 +79,16 @@ export class CatalogEntityRegistry {
 
     // Make sure that we get items ASAP and not the next time one of them changes
     ipcRenderer.send(CatalogIpcEvents.INIT);
+
+    if (isMainFrame) {
```
Iku-turso, line 90: Non-env-agnostic code. As before, tooling is not here yet.

src/renderer/api/catalog-entity-registry.ts
```diff
@@ -78,6 +79,16 @@ export class CatalogEntityRegistry {
 
     // Make sure that we get items ASAP and not the next time one of them changes
     ipcRenderer.send(CatalogIpcEvents.INIT);
```
Iku-turso, line 88: Notice: This problem is solved in wrong level of abstraction.

src/renderer/bootstrap.tsx
```diff
@@ -59,6 +59,7 @@ import bindProtocolAddRouteHandlersInjectable
 import type { LensProtocolRouterRenderer } from "./protocol-handler";
 import lensProtocolRouterRendererInjectable
   from "./protocol-handler/lens-protocol-router-renderer/lens-protocol-router-renderer.injectable";
+import commandOverlayInjectable from "./components/command-palette/command-overlay.injectable";
 
 if (process.isMainFrame) {
```
Iku-turso, line 64: Notice: non-env-agnostic code.

src/renderer/bootstrap.tsx
```diff
@@ -124,7 +122,9 @@ export async function bootstrap(comp: () => Promise<AppComponent>, di: Dependenc
   initializers.initCatalogCategoryRegistryEntries();
 
   logger.info(`${logPrefix} initializing Catalog`);
-  initializers.initCatalog();
+  initializers.initCatalog({
```
Iku-turso, line 125: Notice: manual initialization of stuff here and elsewhere is something that can be done better. Soon.

src/common/hotbar-store.injectable.ts
Iku-turso: Operational code should not be entangled with the mechanism to provide it dependencies. The rare exception is if a dependency has trivial, or no operational code.

src/renderer/components/activate-entity-command/activate-entity-command.tsx
```diff
@@ -19,40 +19,49 @@
...
+      escapeClearsValue={false}
+      placeholder="Activate entity ..."
+    />
+  );
+});
+
+export const ActivateEntityCommand = withInjectables<Dependencies>(NonInjectedActivateEntityCommand, {
+  getProps: di => ({
+    closeCommandOverlay: di.inject(commandOverlayInjectable).close,
+    entities: computed(() => [...catalogEntityRegistry.items]),
```
jansav, line 65: Fine for now. In future CatalogEntityRegistry should be injectable and items already computed there

src/renderer/components/cluster-manager/active-hotbar-name.tsx
```diff
@@ -22,19 +22,31 @@
...
+    onClick={() => openCommandOverlay(<HotbarSwitchCommand />)}
+  >
+    <Icon material="bookmarks" className="mr-2" size={14} />
+    {activeHotbarName()}
+  </div>
+));
+
+export const ActiveHotbarName = withInjectables<Dependencies>(NonInjectedActiveHotbarName, {
+  getProps: (di, props) => ({
+    activeHotbarName: () => di.inject(hotbarManagerInjectable).getActive()?.name,
```
jansav, line 48: Hm. We could have injectable for `activeHotbar` which would be transient. But let's do that if this line gets duplicated.

src/renderer/components/command-palette/registered-commands/internal-commands.injectable.tsx
```diff
@@ -0,0 +1,231 @@
...
+    },
+    {
+      id: "catalog.searchEntities",
+      title: "Catalog: Activate Entity ...",
+      action: () => openCommandDialog(<ActivateEntityCommand />),
+    },
+  ];
+}
+
+const internalCommandsInjectable = getInjectable({
```
jansav, line 221: I foresee that in the future that this injectable should be splitted in a way that each command is it's own injectable. Some tooling to help solve these kind of things are already present in the DI, but not yet used. But good for now.

src/renderer/components/hotbar/hotbar-rename-command.tsx
```diff
@@ -19,81 +19,61 @@
...
+  };
+  uniqueHotbarName: InputValidator;
+}
 
-  constructor(props: {}) {
-    super(props);
-    makeObservable(this);
-  }
+const NonInjectedHotbarRenameCommand = observer(({ closeCommandOverlay, hotbarManager, uniqueHotbarName }: Dependencies) => {
+  const [hotbarId, setHotbarId] = useState("");
```
jansav, line 44: I wouldn't mix MobX and hooks when really not needed. Separating logic (state) from component should be enough and would play very nicely with MobX. Something to think about in the future.

src/renderer/initializers/catalog.tsx
```diff
@@ -64,7 +44,29 @@ async function onClusterDelete(clusterId: string) {
...
 }
 
-export function initCatalog() {
-  initWebLinks();
-  initKubernetesClusters();
+interface Dependencies {
+  openCommandDialog: (component: React.ReactElement) => void;
+}
+
+export function initCatalog({ openCommandDialog }: Dependencies) {
```
jansav, line 51: In future, we should move calls of `init` functions to `instantiate`.  In that way, we don't need to worry about timing.

jansav (approved):
No major things, let's get this merged so we can fix the conflicts with #4653 asap.

Iku-turso (approved):

PR #4431, opened 2021-11-26
Ignore .prettierrc from VCS so that it can be used locally

jansav:
Adding `.prettierrc` to `.gitignore` just so developers who wants to use it, can do with local configuration. 

ESLint remains master for code style, but prettier does so much more. If developer chooses to use prettier locally, she should always make sure that it doesn't affect existing code style where not needed and it doesn't conflict with ESLint.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ types/extension-renderer-api.d.ts
 extensions/*/dist
 docs/extensions/api
 site/
+.prettierrc


Nokel81:
Some things of note:

1. This PR shouldn't be labels as `enhancement` but instead as `chore`.
2. Would it be not better to slowly introduce prettier rules so that all PRs are consistent?

jansav:
> Some things of note:
> 
> 1. This PR shouldn't be labels as `enhancement` but instead as `chore`.
> 2. Would it be not better to slowly introduce prettier rules so that all PRs are consistent?

1. Changed. 
2. I would like to introduce prettier rules to code base as well, but I can see that's larger topic to discuss. Since it would mean that all of those would get formatted with prettier sooner or later. And that's something that we as a team need to agree. 
For the beginning, I thought that it would be easier if I just start using it locally to new code.

Nokel81:
> For the beginning, I thought that it would be easier if I just start using it locally to new code.

My concern with this is that if we don't introduce it to the repo officially, then different users might start using different styles. Which would make PR more cluttered.

jansav:
> > For the beginning, I thought that it would be easier if I just start using it locally to new code.
> 
> My concern with this is that if we don't introduce it to the repo officially, then different users might start using different styles. Which would make PR more cluttered.

Would you prefer if we add the configuration file to VCS instead of ignoring it then? Since it doesn't mean that we start enforcing it.

Nokel81:
No I would prefer if you made a PR for an initial rule (since one rule per PR is a good rule of thumb so that the team can comment on each rule change individually).

As well as adding a new CI action to make sure that all new PRs follow the code style.

jansav:
That would be enforcing it, which we haven't decided to do?

jansav:
We don't want to enforce prettier at the moment. Prettier should be allowed to used locally if developer chooses so as long as code style complies with current code style (ESLint). Future PRs must follow current code style settings.

Nokel81:
> We don't want to enforce prettier at the moment. Prettier should be allowed to used locally if developer chooses so as long as code style complies with current code style (ESLint). Future PRs must follow current code style settings.

We talked about this on the grooming meeting Monday, and the team is fine with slowing starting to enforce style on PRs.

jansav:
> > We don't want to enforce prettier at the moment. Prettier should be allowed to used locally if developer chooses so as long as code style complies with current code style (ESLint). Future PRs must follow current code style settings.
> 
> We talked about this on the grooming meeting Monday, and the team is fine with slowing starting to enforce style on PRs.

Oh, cool. I’ll make the changes then and open separate PR for it.

PR #4433, opened 2021-11-26
Fix hotbar items drag-n-drop bugs

### LINKED ISSUE #2830: Fix hotbar items dragged over iframe
Draggable items can freeze in one place if moved over cluster dashboard.

draggin over iframe

### LINKED ISSUE #4432: Jumping hotbar items onDragEnd
There are multiple mutation operations on hotbar items made and this cause jumping after dragging an item:

https://user-images.githubusercontent.com/9607060/143548766-fbfafe54-0577-4bf4-a04f-05422d4e6740.mov

aleksfront:
Fixes #2830 (Impossible to drag over iframe)
Fixes #4432 (Items are jumping onDragEnd)

https://user-images.githubusercontent.com/9607060/143549818-3ed3eb6b-6f23-438f-a368-e6b5c3974061.mov

### DIFF
--- a/src/renderer/components/hotbar/hotbar-menu.tsx
+++ b/src/renderer/components/hotbar/hotbar-menu.tsx
@@ -33,7 +33,7 @@ import { HotbarSelector } from "./hotbar-selector";
 import { HotbarCell } from "./hotbar-cell";
 import { HotbarIcon } from "./hotbar-icon";
 import { defaultHotbarCells, HotbarItem } from "../../../common/hotbar-types";
-import { makeObservable, observable } from "mobx";
+import { action, makeObservable, observable } from "mobx";
 
 interface Props {
   className?: IClassName;
@@ -62,10 +62,12 @@ export class HotbarMenu extends React.Component<Props> {
     return catalogEntityRegistry.getById(item?.entity.uid) ?? null;
   }
 
+  @action
   onDragStart() {
     this.draggingOver = true;
   }
 
+  @action
   onDragEnd(result: DropResult) {
     const { source, destination } = result;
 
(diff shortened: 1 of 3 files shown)


jweak:
Works otherwise really well, but there's some flicker when releasing the item. The item is rendered in the old place after drop and then it moves to the right place. What causes that? It seems that it's not actually regression though, since the previous version has that as well.

aleksfront:
> Works otherwise really well, but there's some flicker when releasing the item. The item is rendered in the old place after drop and then it moves to the right place. What causes that? It seems that it's not actually regression though, since the previous version has that as well.

I don't see the item is rendered in the old place. Can you show the example?

src/renderer/components/hotbar/hotbar-menu.tsx
```diff
@@ -54,9 +62,15 @@ export class HotbarMenu extends React.Component<Props> {
     return catalogEntityRegistry.getById(item?.entity.uid) ?? null;
   }
 
+  onDragStart() {
+    this.draggingOver = true;
+  }
+
   onDragEnd(result: DropResult) {
```
Nokel81, line 71: Shouldn't this be `action`?

src/renderer/components/hotbar/hotbar-menu.tsx
```diff
@@ -165,9 +179,9 @@ export class HotbarMenu extends React.Component<Props> {
     const hotbar = hotbarStore.getActive();
 
     return (
-      <div className={cssNames("HotbarMenu flex column", className)}>
+      <div className={cssNames("HotbarMenu flex column", { draggingOver: this.draggingOver }, className)}>
         <div className="HotbarItems flex column gaps">
-          <DragDropContext onDragEnd={this.onDragEnd.bind(this)}>
+          <DragDropContext onDragStart={this.onDragStart.bind(this)} onDragEnd={this.onDragEnd.bind(this)}>
```
Nokel81: Can you use properties instead of `bind`?

Nokel81 (approved):

PR #4436, opened 2021-11-26
Prefer usage of preference over autoselection

farodin91:
Before Lens 5.3, I didn't need to able to query all service in Kubernetes. 

Signed-off-by: Jan Jansen <jan.jansen@gdata.de>

### DIFF
--- a/src/main/context-handler.ts
+++ b/src/main/context-handler.ts
@@ -34,12 +34,19 @@ export interface PrometheusDetails {
   provider: PrometheusProvider;
 }
 
+interface PrometheusServicePreferences {
+  namespace: string;
+  service: string;
+  port: number;
+  prefix: string;
+}
+
 export class ContextHandler {
   public clusterUrl: UrlWithStringQuery;
   protected kubeAuthProxy?: KubeAuthProxy;
   protected apiTarget?: httpProxy.ServerOptions;
   protected prometheusProvider?: string;
-  protected prometheusPath: string | null;
+  protected prometheus?: PrometheusServicePreferences;
 
   constructor(protected cluster: Cluster) {
     this.clusterUrl = url.parse(cluster.apiUrl);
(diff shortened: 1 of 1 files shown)


src/main/context-handler.ts
```diff
@@ -58,11 +58,18 @@ export class ContextHandler {
   }
 
   public async getPrometheusDetails(): Promise<PrometheusDetails> {
-    const service = await this.getPrometheusService();
-    const prometheusPath = this.ensurePrometheusPath(service);
-    const provider = this.ensurePrometheusProvider(service);
-
-    return { prometheusPath, provider };
+    if (this.prometheusPath == null) {
+      const service = await this.getPrometheusService();
```
Nokel81: It would be better if `getPrometheusService` checked this and then returned the appropriate `PrometheusService`

farodin91: I won't be able to fix it before Monday.

src/main/context-handler.ts
```diff
@@ -91,6 +85,16 @@ export class ContextHandler {
...
     const providers = this.listPotentialProviders();
+
+    if (this.preferences.prometheus && providers.length == 1) {
+      return {
+        id: providers[0].id,
+        namespace: this.preferences.prometheus.namespace,
+        service: this.preferences.prometheus.service,
+        port: this.preferences.prometheus.port,
+      };
+    }
```
Nokel81: The `id` is the same as `this.prometheusProvider` above so if both are set you don't even need to do the `listPotentialProviders`

src/main/context-handler.ts
```diff
@@ -48,13 +48,7 @@ export class ContextHandler {
...
   public setupPrometheus(preferences: ClusterPrometheusPreferences = {}) {
     this.prometheusProvider = preferences.prometheusProvider?.type;
-    this.prometheusPath = null;
-
-    if (preferences.prometheus) {
-      const { namespace, service, port } = preferences.prometheus;
-
-      this.prometheusPath = `${namespace}/services/${service}:${port}`;
-    }
+    this.preferences = preferences;
```
Nokel81: Final comment, I don't think that we need to save the whole of this here, just like we don't for `this.prometheusProvider`.

farodin91: What do you think from this variant?

Nokel81 (dismissed):

Nokel81:
@farodin91 Please sign all commits in this PR (or just squash and sign that one)

farodin91:
@Nokel81  done

Nokel81 (approved):

PR #4441, opened 2021-11-26
Add a query param for filtering port forwards by clusterId

jim-docker:

This is a precursor to adding a disabled state to port-forwards (so they can be stopped without removing them)

### DIFF
--- a/src/renderer/port-forward/port-forward.store.ts
+++ b/src/renderer/port-forward/port-forward.store.ts
@@ -99,12 +99,13 @@ interface PortForwardsResult {
 }
 
 export async function addPortForward(portForward: ForwardedPort): Promise<number> {
+  const { port, forwardPort } = portForward;
   let response: PortForwardResult;
 
   try {
     const protocol = portForward.protocol ?? "http";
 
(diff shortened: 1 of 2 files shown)


src/renderer/port-forward/port-forward.store.ts
```diff
@@ -166,9 +164,10 @@ export async function removePortForward(portForward: ForwardedPort) {
   portForwardStore.reset();
 }
 
-export async function getPortForwards(): Promise<ForwardedPort[]> {
+export async function getPortForwards(clusterId?: string): Promise<ForwardedPort[]> {
   try {
-    const response = await apiBase.get<PortForwardsResult>(`/pods/port-forwards`);
+    const query = clusterId ? `?clusterId=${clusterId}` : "";
```
Nokel81: I would pass in the query as the second argument to `apiBase.get()`

jim-docker: ah, probably should've been done for the other `apiBase` calls too?

Nokel81: Probably

Nokel81 (dismissed):

Nokel81 (approved):

Issue #4138, opened 2021-10-26
Remove Material UI components from codebase


aleksfront:
In order to eliminate console warnings, free up bundle size and keep standard styling practice without jss.
![mui warning](https://user-images.githubusercontent.com/9607060/138826758-61e24ced-a710-4d6c-ab48-754cb723f06f.png)

PR #4452, opened 2021-11-29
Replacing <Avatar/> component by native one

aleksfront:
* Removing Material UI avatar component for sake of #4138 
* Using raw avatar instead of `<HotbarIcon/>` in most of the places
* Moving `Avatar`, `HotbarIcon` and `HotbarEntityIcon` styles into css modules

https://user-images.githubusercontent.com/9607060/143874359-f0407e0e-3932-4e4c-a5c5-3e5cec34cda2.mov


src/renderer/components/+catalog/catalog-entity-details.tsx
```diff
@@ -27,10 +27,10 @@ import type { CatalogCategory, CatalogEntity } from "../../../common/catalog";
 import { Icon } from "../icon";
 import { CatalogEntityDrawerMenu } from "./catalog-entity-drawer-menu";
 import { CatalogEntityDetailRegistry } from "../../../extensions/registries";
-import { HotbarIcon } from "../hotbar/hotbar-icon";
 import type { CatalogEntityItem } from "./catalog-entity-item";
 import { isDevelopment } from "../../../common/vars";
 import { cssNames } from "../../utils";
+import { Avatar } from "../avatar/avatar";
```
Nokel81: Shouldn't `components/avatar/index.ts` exist?

src/renderer/components/avatar/avatar.module.css
```diff
@@ -0,0 +1,51 @@
...
+  }
+}
+
+.circle {
+  border-radius: 50%;
+}
+
+.rounded {
+  border-radius: var(--border-radius);
+}
```
Nokel81: I think that the board-radius is not large enough. I liked the previous border-radius for "rounded" square more.

src/renderer/components/hotbar/hotbar-icon.tsx
```diff
@@ -75,33 +70,25 @@ export const HotbarIcon = observer(({ menuItems = [], size = 40, tooltip, ...pro
...
-          className={cssNames(active ? "active" : "default", { interactive: !!onClick })}
-          width={size}
-          height={size}
-          src={src}
-        >
-          {material && <Icon className="materialIcon" material={material} />}
-        </Avatar>
-        {children}
-      </div>
+        {material && <Icon material={material} />}
```
Nokel81, line 86: What if `material` is an svg?

aleksfront: Currently entities support only string icon notation and icons being used for General catalog entities, like here https://github.com/lensapp/lens/blob/master/src/common/catalog-entities/general.ts#L60

src/renderer/components/avatar/avatar.module.css
aleksfront: Adjusted `border-radius` with more rounded value. I agreed it looks better now:
rounded avatars

src/renderer/components/hotbar/hotbar-icon.tsx
Nokel81: What about KubernetesCluster entities https://github.com/lensapp/lens/blob/master/src/common/catalog-entities/kubernetes-cluster.ts#L49-L54?

aleksfront: This line `src?: string` relates to image URL and gets passed to `<Avatar src={src}/>` directly.

Nokel81 (approved):

PR #4462, opened 2021-11-30
Fix getLatestApiPrefixGroup to not fail when options.apiBase is not provided

### LINKED ISSUE #4443: Ingress objects not showing up
**Describe the bug**
No ingress objects are visible for any namespace selected. The yellow triangle that shows up next to the namespace dropdown shows an "undefined" tooltip when hovering over it. `kubectl get ingress -n default` from the terminal works as expected, so it doesn't seem to be a permissions issue.

**Screenshots**
image

- Lens Version: 5.3.0-latest.20211125.2
- OS: Windows 10
- Installation method: Windows x64 (.exe)

Nokel81:

fixes #4443

### DIFF
--- a/src/common/k8s-api/__tests__/kube-api.test.ts
+++ b/src/common/k8s-api/__tests__/kube-api.test.ts
@@ -25,6 +25,7 @@ import { KubeJsonApi } from "../kube-json-api";
 import { KubeObject } from "../kube-object";
 import AbortController from "abort-controller";
 import { delay } from "../../utils/delay";
+import { PassThrough } from "stream";
 
 class TestKubeObject extends KubeObject {
   static kind = "Pod";
@@ -170,8 +171,7 @@ describe("KubeApi", () => {
     const fallbackApiBase = "/apis/extensions/v1beta1/ingresses";
     const kubeApi = new KubeApi({
       request,
-      objectConstructor: KubeObject,
-      apiBase,
+      objectConstructor: Object.assign(KubeObject, { apiBase }),
       fallbackApiBases: [fallbackApiBase],
       checkPreferredVersion: true,
     });
@@ -304,19 +304,28 @@ describe("KubeApi", () => {
 
   describe("watch", () => {
     let api: TestKubeApi;
+    let stream: PassThrough;
 
     beforeEach(() => {
       api = new TestKubeApi({
         request,
         objectConstructor: TestKubeObject,
       });
+      stream = new PassThrough();
+    });
+
+    afterEach(() => {
+      stream.end();
+      stream.destroy();
     });
 
     it("sends a valid watch request", () => {
       const spy = jest.spyOn(request, "getResponse");
 
       (fetch as any).mockResponse(async () => {
-        return {};
+        return {
+          body: stream,
+        };
       });
 
       api.watch({ namespace: "kube-system" });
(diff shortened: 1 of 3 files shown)


src/common/k8s-api/__tests__/kube-api.test.ts
```diff
@@ -304,19 +304,28 @@ describe("KubeApi", () => {
...
   describe("watch", () => {
     let api: TestKubeApi;
+    let stream: PassThrough;
 
     beforeEach(() => {
       api = new TestKubeApi({
         request,
         objectConstructor: TestKubeObject,
       });
+      stream = new PassThrough();
```
Nokel81, line 314: All these `stream` changes is to get rid of those annoying logger.error calls during the unit tests.

src/common/k8s-api/__tests__/kube-api.test.ts
```diff
@@ -170,8 +171,7 @@ describe("KubeApi", () => {
     const fallbackApiBase = "/apis/extensions/v1beta1/ingresses";
     const kubeApi = new KubeApi({
       request,
-      objectConstructor: KubeObject,
-      apiBase,
+      objectConstructor: Object.assign(KubeObject, { apiBase }),
```
Nokel81, line 174: This test change validates the code change.

src/common/k8s-api/kube-api.ts
```diff
@@ -295,14 +295,18 @@ export class KubeApi<T extends KubeObject> {
    */
   private async getLatestApiPrefixGroup() {
     // Note that this.options.apiBase is the "full" url, whereas this.apiBase is parsed
-    const apiBases = [this.options.apiBase, ...this.options.fallbackApiBases];
+    const apiBases = [this.options.apiBase, this.objectConstructor.apiBase, ...this.options.fallbackApiBases];
```
Nokel81, line 298: All these `apiBase`'s are marked as optional so that is why we need to skip empty `apiUrl`'s below.

src/common/k8s-api/kube-api.ts
```diff
@@ -295,14 +295,18 @@ export class KubeApi<T extends KubeObject> {
...
-      // Split e.g. "/apis/extensions/v1beta1/ingresses" to parts
-      const { apiPrefix, apiGroup, apiVersionWithGroup, resource } = parseKubeApi(apiUrl);
+      if (!apiUrl) {
+        continue;
+      }
 
-      // Request available resources
       try {
+        // Split e.g. "/apis/extensions/v1beta1/ingresses" to parts
+        const { apiPrefix, apiGroup, apiVersionWithGroup, resource } = parseKubeApi(apiUrl);
```
Nokel81, line 307: Just because an `apiUrl` fails to be parsed as a `KubeApi` descriptor, doesn't mean that we shouldn't then try the next one

src/common/k8s-api/kube-api.ts
jweak: would it be better to filter empty urls here?

Nokel81: You mean instead of in the `for` loop below? I don't really see a difference but I can if you think it would read better.

jweak: Yeah, it's not a big deal, both are fine. 

Code looks good, don't know how to reproduce and test it. I can see ingresses on my clusters.

Nokel81: I would try it out with an older version of kube, one before 1.19. Using `minikube -p older --kubernetes-version=v1.16` for instance

jweak (approved):

PR #4463, opened 2021-11-30
Log cluster model if adding a cluster fails when syncing from catalog

Nokel81:


aleksfront (approved):
LGTM

PR #4476, opened 2021-12-01
Show active hotbar on bottom

### LINKED ISSUE #3982: Display current hotbar name

aleksfront:
Resolves #3982 

https://user-images.githubusercontent.com/9607060/144237407-f062e852-804a-401c-8998-69f73d69c82c.mov


src/renderer/components/cluster-manager/active-hotbar-name.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { HotbarSwitchCommand } from "../hotbar/hotbar-switch-command";
+
+export const ActiveHotbarName = observer(() => {
+  return (
+    <div
+      className="flex items-center"
+      data-testid="current-hotbar-name"
+      onClick={() => CommandOverlay.open(<HotbarSwitchCommand />)}
+    >
+      <Icon material="bookmarks" className="mr-2" style={{ "--size": "14px" } as React.CSSProperties}/>
```
Nokel81: why not use Icon's `size` prop?

src/renderer/components/cluster-manager/bottom-bar.module.css
```diff
@@ -19,23 +19,25 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-.BottomBar {
-  $spacing: $padding * 0.5;
-  --flex-gap: #{$spacing};
+ .BottomBar {
```
Nokel81: Did you mean to add this whitespace?

src/renderer/components/cluster-manager/bottom-bar.tsx
```diff
@@ -63,7 +64,10 @@ export class BottomBar extends React.Component {
 
   render() {
     return (
-      <div className="BottomBar flex gaps">
+      <div className={styles.BottomBar}>
+        <div className={styles.item}>
+          <ActiveHotbarName/>
+        </div>
```
Nokel81: I don't think we should special case this, why not register it? Futhermore, it probably would be a good idea for extensions to be able to choose which side they are on (defaulting to right)

Nokel81 (commented):
I like the look of it, just some comments about the impl.

aleksfront:
@Nokel81 Ptal. Move whole feature into `StatusBarRegistry` allowing to specify `left` or `right` position.
status bar

src/extensions/registries/status-bar-registry.ts
```diff
@@ -26,6 +26,7 @@ import { BaseRegistry } from "./base-registry";
 
 interface StatusBarComponents {
   Item?: React.ComponentType;
+  position?: "left" | "right";
```
Nokel81, line 34: I would add a comment about the default. Something like:

```
/**
 * The side of the bottom bar to place this component.
 *
 * @default "right"
 */
```

src/renderer/components/cluster-manager/bottom-bar.tsx
```diff
@@ -45,29 +45,36 @@ export class BottomBar extends React.Component {
...
         {items.map((registration, index) => {
           if (!registration?.item && !registration?.components?.Item) {
             return null;
           }
 
           return (
-            <div className={styles.item} key={index}>
+            <div
+              className={cssNames(styles.item, {
+                [styles.onLeft]: registration.components?.position == "left",
```
Nokel81, line 62: nit, `components` will be defined here.

aleksfront: Removed optional at first, but then reverted it back - this is more for tests to pass since we're testing different APIs in there.

Nokel81: Ah sounds good

Nokel81 (approved):

PR #4394, opened 2021-11-19
Fix not being able to remove role bindings via the UI

Nokel81:
Signed-off-by: Sebastian Malton <sebastian@malton.name>

fixes #4359

PR #4478, opened 2021-12-01
Fix KubeApi.create not populating kind and apiVersion

Nokel81:
- Add tests


panuhorsmalahti:
Regression from https://github.com/lensapp/lens/pull/4394/files I assume

panuhorsmalahti (approved):

Nokel81:
@panuhorsmalahti Yes, it seems from testing that this is required for CRDs but not for built-ins. Since you don't need to specify it for `ClusterRoleBinding`s for instance.

PR #4480, opened 2021-12-01
Remove type circularity in wizard.tsx

RyanCavanaugh:
This code has a type circularity that previously went undetected. Adding an `as Wizard` assertion here removes the circularity.

See https://github.com/microsoft/TypeScript/issues/46939 / https://github.com/microsoft/TypeScript/pull/46981 for more context


jweak:
Thank you!
@RyanCavanaugh Please sign your commit.

RyanCavanaugh:
I'm not really set up to do that; this is mostly an FYI. I'll close and whoever does the TS 4.6 upgrade can refer back to the fix. Thanks!

Nokel81:
Thanks for the heads up anyhow.

Issue #4473, opened 2021-11-30
5.3.0-latest.20211125.2 does not display windows on Windows 10


AndrewSav:
**Describe the bug**
When I run the lens executable no windows come up and no tray icon. I can see 4 lens.exe processes in the task manager.

**To Reproduce**
Can only reproduce on my laptop. On the laptop it happens 100% of the time. Start the app, the window does not show up. I also tried to reboot a few times and to uninstall / reinstall a few times. No joy.

**Expected behavior**
I expect to see the main Lens window and the tray icon.

**Screenshots**
This is the screenshot from ProcessExplorer showing Lens processes:

![2021-12-01_12-42-03](https://user-images.githubusercontent.com/27190911/144145648-bbb4671d-ab9b-4be2-9239-02b6eb128f4a.png)

**Environment (please complete the following information):**
- Lens Version: 5.3.0-latest.20211125.2
- OS: Windows 10
- Installation method:  Lens Setup 5.3.0-latest.20211125.2.exe

**Logs:**
When you run the application executable from command line you will see some logging output. Please paste them here:
```
info:    ▪ 📟 Setting Lens as protocol client for lens:// +0ms
info:    ▪ 📟 Protocol client register succeeded ✅ +2ms
info:    ▪ [APP-PATHS]: initializing +0ms
```

**Kubeconfig:**

 I have removed the entire `.kube` folder for the test, It is _not_ kubeconfig.

**Additional context**
Worked fine before that latest update.
To clarify, since there a [comment](https://github.com/lensapp/lens/issues/4473#issuecomment-983331381) talking about something else. Lens does not show _any_ UI at all. There is no window and no tray icon displayed, not even the main Lens window.

AndrewSav:
Similar report in comments: https://github.com/lensapp/lens/issues/4430#issuecomment-981433238

TIANQIEND:
![image](https://user-images.githubusercontent.com/34571669/144185295-acf4c440-c570-4843-b7ef-7f28a9042438.png)

I meet the same problem on win10,like picture
version:5.3.0-latest.20211130.1

AndrewSav:
Just to clarify, the comment about is **not** the same problem. It is unrelated for the issue reported. On the comment above there is a screenshot of Lens window displayed, in the case reported there is no window shows up. 

@TIANQIEND I suggest you open a separate issue for the problem _you_ are having because this is not it.

DavidCsaki:
[...]

PR #4481, opened 2021-12-01
add debug for locating startup issues

jim-docker:

For help diagnosing the problem in #4473

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -90,18 +90,22 @@ if (process.env.LENS_DISABLE_GPU) {
   app.disableHardwareAcceleration();
 }
 
-logger.debug("initializeRemote()");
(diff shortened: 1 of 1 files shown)


src/main/index.ts
```diff
@@ -99,6 +101,8 @@ if (app.commandLine.getSwitchValue("proxy-server") !== "") {
   process.env.HTTPS_PROXY = app.commandLine.getSwitchValue("proxy-server");
 }
 
+logger.debug("lprm stuff");
```
aleksfront: Maybe change to more descriptive message? Not quite sure what is `lprm` within context.

jim-docker: sure, this was meant to be temporary but maybe this logging should stay. I'll add a log prefix as well

Nokel81 (approved):

PR #4485, opened 2021-12-02
Fix IngressApi being registered with a wrong new apiBase

### LINKED ISSUE #4482: lens 5.3.2 ingress cant edit & cant open
environment

1. lens 5.3.2
2. Windows 21h2

i cant open a ingress rule and cant edit it
pls

Nokel81:
- Only during checkPreferredVersion()

- Add some unit tests

fixes #4482

### DIFF
--- a/src/common/k8s-api/kube-api.ts
+++ b/src/common/k8s-api/kube-api.ts
@@ -38,9 +38,14 @@ import AbortController from "abort-controller";
 import { Agent, AgentOptions } from "https";
 import type { Patch } from "rfc6902";
 
+/**
+ * The options used for creating a `KubeApi`
+ */
 export interface IKubeApiOptions<T extends KubeObject> {
   /**
    * base api-path for listing all resources, e.g. "/api/v1/pods"
+   *
+   * If not specified then will be the one on the `objectConstructor`
    */
   apiBase?: string;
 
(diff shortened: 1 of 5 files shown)


src/common/k8s-api/kube-api.ts
```diff
@@ -385,7 +402,15 @@ export class KubeApi<T extends KubeObject> {
...
-  getUrl({ name, namespace = "default" }: Partial<ResourceDescriptor> = {}, query?: Partial<IKubeApiQueryParams>) {
+  private computeApiBase(): string {
+    return createKubeApiURL({
+      apiPrefix: this.apiPrefix,
+      apiVersion: this.apiVersionWithGroup,
+      resource: this.apiResource,
+    });
+  }
+
+  getUrl({ name, namespace }: Partial<ResourceDescriptor> = {}, query?: Partial<IKubeApiQueryParams>) {
```
jim-docker, line 413: `getUrl()` could make use of `computeApiBase()`?

src/common/k8s-api/kube-api.ts
```diff
@@ -262,25 +289,20 @@ export class KubeApi<T extends KubeObject> {
...
 
-  constructor(protected options: IKubeApiOptions<T>) {
-    const {
-      objectConstructor,
-      request = apiKube,
-      kind = options.objectConstructor?.kind,
-      isNamespaced = options.objectConstructor?.namespaced,
-    } = options || {};
-
+  constructor({ objectConstructor, request, kind, isNamespaced, ...options }: IKubeApiOptions<T>) {
```
jim-docker: `options` is the right type just by virtue of the other fields being destructured here?

jim-docker (commented):
Looks good, testing...

src/common/k8s-api/kube-api.ts
Nokel81: Do you mean the right name?

src/common/k8s-api/kube-api.ts
Nokel81: Maybe, but I don't see the point since they both call the same helper function.

src/common/k8s-api/kube-api.ts
jim-docker: I mean if you didn't destructure `kind`, say, then `options` is not the expected type. Just seems odd to rely on `...`

Nokel81: But I realized that all this is a breaking change, so I will revert this options change

jim-docker (approved):

PR #4496, opened 2021-12-03
Export KubeJsonApi to extensions

aleksfront:
In order to help making requests with kubeconfig credentials within extension: https://k8slens.slack.com/archives/C01DXFK6SQ3/p1638328285035400


Nokel81:
I don't think we should export that type directly because the static method `forCluster` takes the private `Cluster` type. It think it would be better to export a function that consumes only a `ClusterId`

aleksfront:
@Nokel81 I've converted `forCluster` method to consume `clusterId` instead of `Cluster`.

Nokel81 (approved):

jweak (approved):

PR #4502, opened 2021-12-03
Fix error notifications not being displayed for CreateResource tab

Nokel81:

fixes regression from #4422


jim-docker (approved):
tested

PR #4514, opened 2021-12-06
Replace all uses of promiseExec with promiseExecFile

Nokel81:

resolves https://github.com/lensapp/lens-security/issues/7

### DIFF
--- a/src/common/system-ca.ts
+++ b/src/common/system-ca.ts
@@ -22,7 +22,7 @@
 import { isMac, isWindows } from "./vars";
 import wincaAPI from "win-ca/api";
 import https from "https";
-import { promiseExec } from "./utils/promise-exec";
+import { promiseExecFile } from "./utils/promise-exec";
 
 // DST Root CA X3, which was expired on 9.30.2021
 export const DSTRootCAX3 = "-----BEGIN CERTIFICATE-----\nMIIDSjCCAjKgAwIBAgIQRK+wgNajJ7qJMDmGLvhAazANBgkqhkiG9w0BAQUFADA/\nMSQwIgYDVQQKExtEaWdpdGFsIFNpZ25hdHVyZSBUcnVzdCBDby4xFzAVBgNVBAMT\nDkRTVCBSb290IENBIFgzMB4XDTAwMDkzMDIxMTIxOVoXDTIxMDkzMDE0MDExNVow\nPzEkMCIGA1UEChMbRGlnaXRhbCBTaWduYXR1cmUgVHJ1c3QgQ28uMRcwFQYDVQQD\nEw5EU1QgUm9vdCBDQSBYMzCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEB\nAN+v6ZdQCINXtMxiZfaQguzH0yxrMMpb7NnDfcdAwRgUi+DoM3ZJKuM/IUmTrE4O\nrz5Iy2Xu/NMhD2XSKtkyj4zl93ewEnu1lcCJo6m67XMuegwGMoOifooUMM0RoOEq\nOLl5CjH9UL2AZd+3UWODyOKIYepLYYHsUmu5ouJLGiifSKOeDNoJjj4XLh7dIN9b\nxiqKqy69cK3FCxolkHRyxXtqqzTWMIn/5WgTe1QLyNau7Fqckh49ZLOMxt+/yUFw\n7BZy1SbsOFU5Q9D8/RhcQPGX69Wam40dutolucbY38EVAjqr2m7xPi71XAicPNaD\naeQQmxkqtilX4+U9m5/wAl0CAwEAAaNCMEAwDwYDVR0TAQH/BAUwAwEB/zAOBgNV\nHQ8BAf8EBAMCAQYwHQYDVR0OBBYEFMSnsaR7LHH62+FLkHX/xBVghYkQMA0GCSqG\nSIb3DQEBBQUAA4IBAQCjGiybFwBcqR7uKGY3Or+Dxz9LwwmglSBd49lZRNI+DT69\nikugdB/OEIKcdBodfpga3csTS7MgROSR6cz8faXbauX+5v3gTt23ADq1cEmv8uXr\nAvHRAosZy5Q6XkjEGB5YGV8eAlrwDPGxrancWYaLbumR9YbK+rlmM6pZW87ipxZz\nR8srzJmwN0jP41ZL9c8PDHIyh8bwRLtTcm1D9SZImlJnt1ir/md2cXjbDaJWFBM5\nJDGFoqgCWjBH4d1QB7wCCZAA62RjYJsWvIjJEubSfZGL+T0yjWW06XyxV3bqxbYo\nOb8VZRzI9neWagqNdwvYkQsEjgfbKbYK7p2CNTUQ\n-----END CERTIFICATE-----\n";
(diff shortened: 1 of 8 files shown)


src/main/helm/helm-release-manager.ts
```diff
@@ -22,161 +22,224 @@
...
-    }
-    output.forEach((release: any, index: number) => {
-      output[index] = toCamelCase(release);
-    });
+    const { stdout } = await promiseExecFile(helmCliPath, args, options);
 
-    return output;
+    return stdout;
   } catch (error) {
     throw error?.stderr || error;
```
jim-docker, line 39: I know this is unrelated but shouldn't it just be
```suggestion
    throw error.stderr || error;
```
?

src/common/system-ca.ts
```diff
@@ -33,19 +33,25 @@ export function isCertActive(cert: string) {
   return !isExpired;
 }
 
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Cheatsheet#other_assertions
+const certSplitPattern = /(?=-----BEGIN\sCERTIFICATE-----)/g;
+
+async function execSecurity(...args: string[]): Promise<string[]> {
+  const { stdout } = await promiseExecFile("/usr/bin/security", args);
+
+  return stdout.split(certSplitPattern);
```
jim-docker, line 42: was the toString() call unnecessary?

src/main/helm/helm-release-manager.ts
```diff
@@ -22,161 +22,224 @@
...
+
+
+export async function installChart(chart: string, values: any, name: string | undefined = "", namespace: string, version: string, kubeconfigPath: string) {
+  const valuesFilePath = tempy.file({ name: "values.yaml" });
+
+  await fse.writeFile(valuesFilePath, yaml.dump(values));
+
+  const args = [
+    "install",
+    name,
```
jim-docker: name could be undefined here?

src/main/helm/helm-release-manager.ts
```diff
@@ -22,161 +22,224 @@
...
+    chart,
+    "--version", version,
+    "--values", valuesFilePath,
```
jim-docker, line 81: `--values` does the same as `-f`? Depends on Helm version?

src/main/helm/helm-release-manager.ts
```diff
@@ -22,161 +22,224 @@
...
-    throw error?.stderr || error;
-  }
+export async function rollback(name: string, namespace: string, revision: number, kubeconfigPath: string) {
+  return JSON.parse(await execHelm([
+    "rollback",
+    name,
+    "--output", "json",
+    "--namespace", namespace,
+    "--kubeconfig", kubeconfigPath,
+  ]));
```
jim-docker: missed a few changes in copy/paste from `getHistory`?

src/main/helm/helm-release-manager.ts
```diff
@@ -22,161 +22,224 @@
...
+    return new Promise((resolve, reject) => {
+      let stdout = "";
+      const kubectl = execFile(kubectlPath, kubectlArgs);
+
+      kubectl.stdout.on("data", output => stdout += output);
+      kubectl.stdin.write(helmOutput);
+      kubectl.stdin.end();
+
+      kubectl
+        .on("exit", (code, signal) => {
```
jim-docker, line 230: shouldn't this be done before writing to stdin?

src/main/helm/helm-repo-manager.ts
```diff
@@ -77,22 +90,21 @@ export class HelmRepoManager extends Singleton {
...
-      if (key && value) {
-        env[key] = value.replace(/"/g, ""); // strip quotas
-      }
-    });
+        return null;
+      })
+      .filter(Boolean);
 
-    return env;
+    return Object.fromEntries(env);
```
jim-docker: ride-along refactoring? unit tests will catch problems?

src/main/helm/helm-repo-manager.ts
```diff
@@ -125,67 +137,85 @@ export class HelmRepoManager extends Singleton {
       const { repositories } = await this.readConfig();
 
       if (!repositories.length) {
+        if (previousAttempt) {
+          throw new Error("Previous add repo called did not add repo");
+        }
+
```
jim-docker: stops infinite recursion?

src/main/helm/helm-release-manager.ts
Nokel81: I guess so. But since any value can be thrown I sort of did it to guard against null and undefined values. But I guess it doesn't guard against other non object types.

src/common/system-ca.ts
Nokel81: Yes

src/main/helm/helm-release-manager.ts
jim-docker: Then there's no point having the `||`. This gives same result? Or what am I missing?
```suggestion
    throw error?.stderr;
```

src/main/helm/helm-repo-manager.ts
Nokel81: Yes

src/main/helm/helm-repo-manager.ts
Nokel81: I'll add tests but this is more correct because values can have `=`'s.

src/main/helm/helm-release-manager.ts
Nokel81: Probably

src/main/helm/helm-release-manager.ts
Nokel81: Well if it is an error instance that doesn't have `stderr` we still want the error.

src/main/helm/helm-release-manager.ts
Nokel81: Yes thanks for the catch

src/main/helm/helm-repo-manager.ts
jim-docker: That's good, but it should be done in another PR, especially if you're adding tests(?)

Nokel81: sounds good, will do

src/main/helm/helm-release-manager.ts
jim-docker: `error` from `catch(error)` can be undefined?

Nokel81: yes, `throw;` is valid JS/TS

src/main/helm/helm-release-manager.ts
Nokel81: Nope, even in 3.0.0 it was here https://github.com/helm/helm/blob/v3.0.0/pkg/cli/values/options.go#L39

src/main/helm/helm-repo-manager.ts
```diff
@@ -65,31 +65,41 @@ async function execHelm(args: string[], options?: BaseEncodingOptions & ExecFile
...
+    if (repos.length === 0) {
+      await this.addRepo({
+        name: "bitnami",
+        url: "https://charts.bitnami.com/bitnami",
+      });
+    }
+
+    if (!this.didUpdateOnce) {
+      await this.update();
+      this.didUpdateOnce = true;
```
jim-docker: Does this have anything to do with exec => execFile?

jim-docker (dismissed):
tested on mac and linux (for regression, not security impact)

jim-docker (dismissed):
So this PR can be cherry-picked for 5.3.4 (and won't suffer this flaky test). And you can turn the test back on in https://github.com/lensapp/lens/pull/4564 for master, right? As soon as it is approved :)

Nokel81:
@jim-docker Yes that is the plan

src/main/helm/helm-repo-manager.ts
```diff
@@ -125,67 +135,87 @@ export class HelmRepoManager extends Singleton {
...
       if (!repositories.length) {
+        if (previousAttempt) {
+          throw new Error("Previous add repo called did not add repo");
+        }
+
         await HelmRepoManager.addRepo({ name: "bitnami", url: "https://charts.bitnami.com/bitnami" });
 
-        return await this.repositories();
+        return await this.repositories(true);
       }
```
jansav: Having recursion here seems bit odd. 

Could we just:

```typescript
const { repositories: storedRepositories } = await this.readConfig();

let repositories: HelmRepo[] = storedRepositories;

if (!repositories.length) {
  await HelmRepoManager.addRepo({
    name: "bitnami",
    url: "https://charts.bitnami.com/bitnami",
  });

  // Even better would be if we could trust that the HelmRepo used as argument of addRepo is the exact same that "readConfig" returns.
  ({ repositories } = (await this.readConfig()));
}
```

In bigger picture, I think that adding the default repository should be extracted away because now this function has two separate responsibilities.

aleksfront:
Probably not related to current PR, but found that we need to refresh helm chart after go from Preferences page. Right now, it might show stale repos and charts in list. If you click them, you got an app crash:

https://user-images.githubusercontent.com/9607060/146362541-6362f41f-caaa-4f3c-af38-32783539b316.mov

src/main/helm/helm-repo-manager.ts
Nokel81: Yes I agree that having a default registry should be abstracted away. I have already done this in my follow up branch
 Which I will make into a PR once this is merged. 

We felt that this PR shouldn't have too many non-necessary changes.

Nokel81:
> Probably not related to current PR, but found that we need to refresh helm chart after go from Preferences page. Right now, it might show stale repos and charts in list. If you click them, you got an app crash:
> 
> https://user-images.githubusercontent.com/9607060/146362541-6362f41f-caaa-4f3c-af38-32783539b316.mov
> 
> 
> 

I will investigate since I didn't think that this was possible before this PR

Nokel81:
@aleksfront That also currently exists in master so I don't think we should fix it in this security PR.

src/main/helm/helm-repo-manager.ts
jim-docker: Good points all, but this was a change unrelated to the security fix which has effectively delayed approval of this review.

jim-docker (approved):

jansav (approved):

Iku-turso (approved):

PR #4515, opened 2021-12-06
Validate ClusterModel before trying to update or create a Cluster

Nokel81:

This changes the error in `ClusterManager` from "EISDIR" to "invalid kubeconfigPath"


src/common/cluster-types.ts
```diff
@@ -19,6 +19,8 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
+import Joi from "joi";
```
jweak, line 6: Did you consider something like [Zod](https://github.com/colinhacks/zod) for more Typescript-first like validation library?

https://itnext.io/parse-dont-validate-incoming-data-in-typescript-d6d5bfb092c8

Nokel81: I did not since we already have `joi` in the repo. Also we don't have strict mode on currently...

> ⚠️ IMPORTANT: You must enable strict mode in your tsconfig.json. This is a best practice for all TypeScript projects.

Yes I can see the benefit of the typings working as expected but I don't think it is worth it to introduce a new library just yet.

jweak: Oh alright, I missed that `joi` is already in the project. Good points. Nevermind.

src/common/cluster-types.ts
```diff
@@ -27,6 +29,37 @@ export type ClusterId = string;
...
+ */
+export const updateClusterModelChecker = Joi.object<UpdateClusterModel>({
+  kubeConfigPath: Joi.string()
+    .required()
+    .min(1),
+  contextName: Joi.string()
+    .required()
+    .min(1),
+  workspace: Joi.string()
+    .optional(),
```
jim-docker, line 43: optional is default, right? why mark this one and not the others below?

src/common/cluster/cluster.ts
```diff
@@ -217,9 +217,16 @@ export class Cluster implements ClusterModel, ClusterState {
...
 
-  constructor(private readonly dependencies: ClusterDependencies, model: ClusterModel) {
+  constructor(private readonly dependencies: ClusterDependencies, { id, ...model }: ClusterModel) {
     makeObservable(this);
-    this.id = model.id;
+
+    const { error } = clusterModelIdChecker.validate({ id });
+
+    if (error) {
+      throw error;
```
jim-docker, line 246: I assume there's something ready to catch this?

jim-docker (dismissed):

src/common/cluster/cluster.ts
Nokel81: Yes because there are already other errors that can be thrown in this constructor. For instance on line 235 below.

src/common/cluster-types.ts
Nokel81: Fair point. I think we could probably remove this old and unused code too but I think that should be done in a separate PR.

PR #4527, opened 2021-12-08
Remove MenuRegistry in favour of dependency injection 

### LINKED ISSUE #4475: Injectables in Main Process
Now that we have a way to inject dependencies (#4338) in renderer process, it would be very nice to as well have a symmetrical solution in main process for the same benefits.

jansav:
# What we did
1. We removed a typical registry `MenuRegistry` and replaced it with an injectable `ElectronMenuItems`.  The injectable instantiates a MobX-computed observed from all applicable extensions. With this, we propose dependency injection over shared global state, and reactivity over statefulness.
2. With removal of MenuRegistry, also an instance of `Singleton`-base-class got removed. With this, we also limit shared global state.
3. Now that there is friendlier unit-testability, we added unit tests for said `ElectronMenuItems`.
4. We made sure that “Extension API” remains unmolested.
5. We made injectables usable in `Main` process similar to `Renderer`.

# Rationale
We recognized implementations of base-classes `Singleton` and `BaseRegistry` as main causes of problems like **code rigidity**, **non-testability**, **tendency to bug** etc. These classes describe shared global state, logical dependencies, and statefulness over reactivity. Now that we have infrastructure in place to inject dependencies (i.e. `injectable`), we can replace these anti-patterns with dependency injection. It is also handy that outsourcing lifecycles (e.g. singleton) to DI comes naturally, without the dependee even knowing about it.

# What’s next:
1. We will keep applying this solution to other registries for same benefits.

Closes #4475.


src/extensions/extension-loader.ts
```diff
@@ -248,25 +247,7 @@ export class ExtensionLoader extends Singleton {
...
-        if (removedExtension.id === extension.id) {
-          removeItems.forEach(remove => {
-            remove();
-          });
-        }
-      });
-
-      return removeItems;
-    });
+    this.autoInitExtensions(() => Promise.resolve([]));
```
jansav, line 250: Notice how we no longer need to dispose items from a state when extension is disabled or removed.

src/main/index.ts
```diff
@@ -236,8 +239,10 @@ app.on("ready", async () => {
   logger.info("🖥️  Starting WindowManager");
   const windowManager = WindowManager.createInstance();
 
+  const menuItems = di.inject(electronMenuItemsInjectable);
+
   onQuitCleanup.push(
-    initMenu(windowManager),
+    initMenu(windowManager, menuItems),
```
jansav, line 246: This would be cleaner if function `initMenu` was `injectable`. @panuhorsmalahti would you like to do this with us? :)

src/extensions/extension-loader.ts
```diff
@@ -248,25 +247,7 @@ export class ExtensionLoader extends Singleton {
...
   }
```
jansav, line 254: Notice how we no longer need to dispose items from a state when extension is disabled or removed.

src/main/getDiForUnitTesting.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export const getDiForUnitTesting = () => {
+  const di: ConfigurableDependencyInjectionContainer = createContainer();
+
+  getInjectableFilePaths()
+    .map(key => {
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const injectable = require(key).default;
+
+      if (!injectable) {
```
jweak: is this needed? If it is, probably a better log than just the key would be nice.

Iku-turso: Hmm. This `if` should not be here at all. Removing. Nice catch :)

src/main/getDiForUnitTesting.ts
```diff
@@ -0,0 +1,59 @@
...
+        console.log(key);
+      }
+
```
Iku-turso:
```suggestion
```

src/extensions/extension-loader.ts
Nokel81: Yes I have wanted to do this for some time, I even made https://github.com/lensapp/lens/pull/2245 about 10 months ago, but no one wanted to me to pursue it.

src/main/index.ts
```diff
@@ -61,11 +61,15 @@ import { FilesystemProvisionerStore } from "./extension-filesystem";
 import { SentryInit } from "../common/sentry";
 import { ensureDir } from "fs-extra";
 import { Router } from "./router";
-import { initMenu } from "./menu";
+import { initMenu } from "./menu/menu";
```
Nokel81, line 62: why change this import?

src/main/index.ts
```diff
@@ -61,11 +61,15 @@ import { FilesystemProvisionerStore } from "./extension-filesystem";
...
 import { initTray } from "./tray";
 import { kubeApiRequest, shellApiRequest, ShellRequestAuthenticator } from "./proxy-functions";
 import { AppPaths } from "../common/app-paths";
 import { ShellSession } from "./shell-session/shell-session";
+import { getDi } from "./getDi";
+import electronMenuItemsInjectable from "./menu/electron-menu-items.injectable";
```
Nokel81, line 68: why is this "electron" menu items?

src/extensions/extensions.injectable.ts
```diff
@@ -0,0 +1,42 @@
...
+import { computed, IComputedValue } from "mobx";
+import { ExtensionLoader } from "./extension-loader";
+import type { LensExtension } from "./lens-extension";
+
+const extensionsInjectable: Injectable<
+  IComputedValue<LensExtension[]>,
+  { extensionLoader: ExtensionLoader }
+> = {
+  getDependencies: () => ({
+    extensionLoader: ExtensionLoader.createInstance(),
```
Nokel81: No you aren't supposed to create the instances in more than once place.

src/extensions/extension-loader.ts
```diff
@@ -47,7 +46,7 @@ const logModule = "[EXTENSIONS-LOADER]";
  */
 export class ExtensionLoader extends Singleton {
   protected extensions = observable.map<LensExtensionId, InstalledExtension>();
-  protected instances = observable.map<LensExtensionId, LensExtension>();
+  instances = observable.map<LensExtensionId, LensExtension>();
```
Nokel81: I don't think this should be public. A getter for the list of instances could be useful if it doesn't already exist.

src/main/menu/electron-menu-items.test.ts
```diff
@@ -0,0 +1,132 @@
+/**
```
Nokel81, line 1: Please put test files under a `__test__` folder

src/main/menu/electron-menu-items.injectable.ts
```diff
@@ -0,0 +1,41 @@
...
+  { extensions: IComputedValue<LensMainExtension[]> }
+> = {
+  lifecycle: lifecycleEnum.singleton,
+
+  getDependencies: di => ({
+    extensions: di.inject(extensionsInjectable),
+  }),
+
+  instantiate: ({ extensions }) =>
+    computed(() => extensions.get().flatMap(extension => extension.appMenus)),
```
Nokel81, line 38: what about disabled extensions?

src/main/index.ts
Nokel81: I guess, what is the benefit of this being injectable here? Instead of just computing the entries?

src/main/menu/menu-registration.d.ts
```diff
@@ -18,15 +18,8 @@
  * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
-
-// Extensions API -> Global menu customizations
```
Nokel81, line 22: Why remove a comment?

src/main/index.ts
jansav: Moved file to separate directory. Prefer not to create `index.ts` when really not needed.

src/main/menu/menu-registration.d.ts
```diff
@@ -18,15 +18,8 @@
  * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
```
Nokel81, line 18: Also, why move this file?

src/main/menu/menu.ts
```diff
@@ -302,14 +307,17 @@ export function buildMenu(windowManager: WindowManager) {
   ]);
 
   // Modify menu from extensions-api
-  for (const { parentId, ...menuItem } of MenuRegistry.getInstance().getItems()) {
-    if (!appMenu.has(parentId)) {
-      logger.error(`[MENU]: cannot register menu item for parentId=${parentId}, parent item doesn't exist`, { menuItem });
+  for (const menuItem of electronMenuItems) {
```
Nokel81, line 310: IMO it is cleaner to not pass `menuItem.parentId` to the list of submenus.

src/main/index.ts
Nokel81: Okay I guess, I find `index` files much cleaner to read especially since the rest of this repo we use them.

src/extensions/extension-loader.ts
jansav: Glad to see that you have had same kind of changes in mind. However, I don’t see a reason not to pursue simplicity that reactive programming gives. Current state is less than optimal.

src/main/index.ts
jansav: Since `menuItems` is not enough to make it obvious. Other option was `appMenuItems` but that has same effect. App can be anything. Using Electron in name is not best, but still lot better than without it. Maybe `mainAppMenuItems`?

src/extensions/extensions.injectable.ts
jansav: Yep, this is the issue with current `Singleton`. I shouldn’t need to know how and when instance is created. 

If we change this to `.getInstance()` we need to be sure that `.createInstance()` has been called. 

This is something that we need to fix by getting rid of `Singleton`

src/main/menu/electron-menu-items.test.ts
jansav: Why? There are tests already in codebase that are not in `__tests__` directory.

src/extensions/extensions.injectable.ts
Nokel81: I completely disagree, just because something is a singleton doesn't mean there aren't dependency's on other instances, and those instances might need arguments to be created.

src/main/menu/electron-menu-items.injectable.ts
jansav: Just noticed that there is something weird happening here. Disabled extensions work already, but it seems that is not enough. We need to be sure that `.init()` has been called for extension as well. 

I have required changes done already and pushing those here tomorrow.

src/main/menu/menu-registration.d.ts
jansav: This was located in a file which was deleted.

src/main/index.ts
jansav: This file shouldn’t need to know what are the dependencies for `initMenu`. Simplicity.

src/main/menu/menu-registration.d.ts
Nokel81: Ah well the git diff implied that it was moved here instead. Then why delete that file in registries instead of just deleting the registry?

src/main/index.ts
jansav: We have seen both used in this repo. We prefer it this way so that we don’t need so much boilerplate code. 

Not that big deal though for us

src/main/index.ts
Nokel81: I meant, why is `menuItems` not just a computed without this injection

src/extensions/extensions.injectable.ts
jansav: Solution is to use DI to make instances. That’s how I don’t need to worry about creating instance and getting arguments somewhere. 

In this case, it’s clear that `ExtensionLoader` needs to be injected. Therefore `injectable` should be created.

src/main/menu/menu-registration.d.ts
jansav: Would be silly to leave empty file there?  (Or file that doesn’t do anything and doesn’t have anything to do with registries.)

Before it was the place that this interface was used first time, but since that doesn’t exist anymore, it makes sense to move it closer to useplace.

src/main/index.ts
jansav: Same argument:

`menuItems` has dependencies that I shouldn’t need to know when I use it. 

All this code should care is that the required interface is somehow fulfilled.

src/extensions/extensions.injectable.ts
jakolehm: I think @Nokel81 means that we shouldn't call `ExtensionLoader.createInstance()` in multiple places (see main/index.ts for other call). That said, it's a bit unfortunate that caller needs to know anything about this -> maybe instance should always come via injectable.

src/main/index.ts
jakolehm: To me it seems that more common practice is to use "index" in this repo but that has a downside because it requires tree-shaking (which might not work in practice too well). Without tree-shaking we are actually "loading" all dependencies from index ... of course import shouldn't have side-effects in a perfect world (we are not there yet).

src/main/index.ts
panuhorsmalahti: @jansav I'm afraid I'm a bit busy with other Lens 🔥 issues for the time being..

src/main/menu/menu-registration.d.ts
Iku-turso: It serves no purpose -> dead weight.

src/main/menu/electron-menu-items.test.ts
panuhorsmalahti: I'd argue for having tests in a tests folder (being a bit vague to not bikeshed too much). If Lens codebase is not consistent wrt, we should strive to make it so. I think file navigation becomes less convenient if test files are located next to application code.

src/main/index.ts
Iku-turso: No worries @panuhorsmalahti :)

We can share you the result, if you want.

src/extensions/extension-loader.ts
Iku-turso: Awesome :)

src/main/index.ts
panuhorsmalahti: Sure, that sounds interesting.

src/main/menu/menu-registration.d.ts
Nokel81: It is also used in `LensMainExtension` and I would argue that is its "first use" and this is merely its consumer.

src/main/index.ts
Iku-turso: Hmm. The index-files do feel like an unnecessary indirection and complication. We could consider if the benefits of encapsulation really overweigh them. As said, not a bid deal, as there's far bigger fish out there :)

...but in general, when it comes to side-effects on import, I'm afraid it's not something our team should tolerate at all. Let's make a hard line for this, right guys? :)

Nokel81: I agree, it also makes circular dependencies too easy

src/extensions/extensions.injectable.ts
jansav: I know and I agree. 

To be honest, I’m not sure why it was left as `.createInstance()` since `.getInstance()` is more than enough. 

However it works as a great example of an issue caused `Singleton` and permits this kind discussion. 

I’ll change it tomorrow anyway :)

Nokel81: Because not every place that consumes an instance needs/should know how to create it.

src/main/menu/electron-menu-items.test.ts
Iku-turso: You might be both right:

Sometimes unit tests specify real life behaviour instead of workings of certain technical entrypoint (i.e. a file). For this, `__tests__`makes sense, as the entrypoint becomes less relevant. This does not happen in current codebase yet. Some call this integration testing or functional testing, but let's not get into semantics.

At other times unit tests specify workings of a technical detail. For this, placing the test file next to entrypoint of unit makes sense, as the presence of tests (or lack thereof) is a valuable piece of information not to be swept away in a directory. When doing this, inconvenience in navigation is oftentimes sign of another problem, namely files not being organised in directories properly. Tests should not make directories feel cluttered.

src/main/menu/menu.ts
Iku-turso: Hmm. I suppose I considered the complication in typing not worth this. Do you agree? I see your point of encapsulation, but this case is trivial, and therefore simplicity is more pragmatic.

Nokel81: How is the new code more simple?

src/extensions/extensions.injectable.ts
Iku-turso: This we know ;)

src/main/menu/menu.ts
Iku-turso: I don't see this issue as good time investment. I can change it back if you feel strongly about it :)

src/extensions/extension-loader.ts
Iku-turso: Agreed, this being exposed for external mutation is scary. Good catch.

src/main/menu/electron-menu-items.injectable.ts
Iku-turso: Hmm. In general, with DI we will have more guarantees that temporal requirements, such as `.init()` this or `.createInstance()` that, will have been called when expected. Good.

src/main/index.ts
Iku-turso: I do believe the promised result will satisfy @Nokel81, as it will extract responsibilities related to instantiation away from this file, where it doesn't quite belong. @Nokel81 how about you, wanna have a go doing this with us? It may be trivial here, but having this freedom somewhere else will be a big deal. 

Or the beer's on me ;)

Nokel81: freedom?

src/main/menu/menu-registration.d.ts
jansav: I would say that `interface MenuRegistration` exists for making sure that `Menu` has items with proper interface. That being said, `Menu` is the one that defines the requirements for the interface -> logical place for the interface is next to `Menu`.

Iku-turso:
Alright! With the latest changes, this will be one of those bigger PR's we hinted the team about. The big push was getting `ExtensionLoader` to comply with dependency injection, and more importantly the design choices which made that possible.

Here's our sincerest hopes to fast-track this PR, as it will directly support the next cool stuff to come. Onwards!

Nokel81 (dismissed):

jansav:
@Nokel81, I fixed failing tests and linting errors. Needs to be approved again.

Iku-turso (approved):

Nokel81 (approved):

Issue #1734, opened 2020-12-10
CSD or native headerbars


ogarcia:
**What would you like to be added**:
Client-Side Decoration (CSD) support or native headerbars

**Why is this needed**:
Now if you open Lens in Linux you have something like this.
![lens_headbar](https://user-images.githubusercontent.com/572924/101760808-a840cd00-3adb-11eb-8c8c-70009547cb2f.png)

But if you support CSD as in [Teams](https://www.microsoft.com/en-us/microsoft-365/microsoft-teams/download-app) (for example) can use headbar to add app elements and don't waste a lot of space.
![sample_headbar](https://user-images.githubusercontent.com/3630663/36128754-0773fb80-1032-11e8-96dd-c1b83a8af226.png)

**Environment you are Lens application on:**
- Kubernetes distribution: n/a
- Desktop OS: GNOME on Linux, but n/a

Nokel81:
This is not currently possible as electron does not support it see https://github.com/electron/electron/issues/11907

rohmishra:
How would lens handle positioning of the buttons? Positioning needs to be passed from gtk to electron environment but after that the app is responsible for actually placing the buttons within the UI. 

On linux, you can have buttons on both left and right side or even both side at same time, Plus they can have different options, even non-standard ones at times.

Considering we ignore all buttons that are non standard, the app still needs to accommodate for between one to three buttons on both sides of the UI. And this is considering the app uses its own assets for the button UI and ignores the system style. This can actually be desirable as using system UI means even more complexity.

ogarcia:
[Teams](https://www.microsoft.com/en-us/microsoft-365/microsoft-teams/download-app) is an electron app and uses CSD very well. I think that Lens can handle it in same way.
![Teams CSD](https://user-images.githubusercontent.com/572924/141072495-6838dc8c-851e-42ef-8b40-eeb1a1d6f3e9.png)

rohmishra:
@ogarcia Teams has the same issue that i discussed in previous comment. It doesnt follow system in button placement. Even if you set your window decorations to be on left, teams buttons wil appear on right. This isnt an issue on macOS and windows because those settings arent configurable.

On gnome for example you need to read the following gsetting to learn user's preference: `org.gnome.desktop.wm.preferences button-layout`

There is no standardized way to read this settings. Its best if a library does the heavy lifting of querying these and the app can in turn read from the said library.

[...]

PR #4528, opened 2021-12-08
Custom title controls on Windows

### LINKED ISSUE #1437: Make windows title and menu consistent with theme
Window title and menu should be consistent with app theme.

Current windows view is too contrast
windows menu

Menu could be placed in some "sandwich" toggler that will show all of the items as a popup.

This is also could related to linux if possible.

aleksfront:
* Removing native menu and title bar since it's to contrast and spoils whole app appearance in Windows
* Moving context menu under sandwich icon
* Adding custom buttons for window minimize, maximize and close

macOS view stays untouched.

**Note**. This PR does not include custom window controls in Preferences, Add cluster, Settings and other layouts which uses full-screen components. Will be handled in ongoing contributions.

Fixes #1437 
windows controls

https://user-images.githubusercontent.com/9607060/145201266-c6b96171-344a-4de6-9df8-63f5d274f95f.mov


src/renderer/components/layout/topbar.tsx
```diff
@@ -82,25 +91,47 @@ export const TopBar = observer(({ children, ...rest }: Props) => {
...
 
   return (
-    <div className={styles.topBar} {...rest}>
-      <div className={styles.tools} onDoubleClick={windowSizeToggle}>
+    <div className={styles.topBar} onDoubleClick={windowSizeToggle} ref={elem} {...rest}>
+      <div className={styles.tools}>
+        <div className={styles.winMenu}>
+          {(isWindows || isLinux) && (
+            <div onClick={openContextMenu} data-testid="window-menu">
+              <svg width="12" height="12" viewBox="0 0 12 12" shapeRendering="crispEdges"><path fill="currentColor" d="M0,8.5h12v1H0V8.5z"/><path fill="currentColor" d="M0,5.5h12v1H0V5.5z"/><path fill="currentColor" d="M0,2.5h12v1H0V2.5z"/></svg>
```
aleksfront, line 132: Using raw svgs here instead of `<Icon/>` in order to avoid messing around with `Icon` internal styles (should rewrite many of them with `!important` declaration).

src/main/window-manager.ts
```diff
@@ -81,7 +85,7 @@ export class WindowManager extends Singleton {
         show: false,
         minWidth: 700,  // accommodate 800 x 600 display minimum
         minHeight: 500, // accommodate 800 x 600 display minimum
-        titleBarStyle: "hiddenInset",
+        titleBarStyle: isWindows ? "hidden" : "hiddenInset",
```
Nokel81: What about linux?

src/main/menu.ts
```diff
@@ -316,5 +316,10 @@ export function buildMenu(windowManager: WindowManager) {
     appMenu.delete("mac");
   }
 
-  Menu.setApplicationMenu(Menu.buildFromTemplate([...appMenu.values()]));
+  return [...appMenu.values()];
+
+}
+
+export function buildMenu(windowManager: WindowManager) {
```
Nokel81, line 323: Shouldn't we not call `initMenu` (the function that calls this function) on non-Macs?

src/renderer/components/layout/__tests__/topbar-linux.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+    TopBarRegistry.resetInstance();
+  });
+
+  it("shows menu icon", () => {
+    const { getByTestId } = render(<TopBar/>);
+
+    expect(getByTestId("window-menu")).toBeInTheDocument();
+  });
+
+  it("doesn't show windows title buttons", () => {
```
Nokel81: Why do we not want to show the title buttons?

src/main/window-manager.ts
jim-docker: No change, like on mac?

Screen Shot 2021-12-08 at 11 41 10 AM

Nokel81: I guess I meant, why not on linux

jim-docker: Yes, meant as a question to @aleksfront, if this is how it should look, wrt this PR

jim-docker:
The menu bar disappears if the window is closed and re-opened (on linux):

https://user-images.githubusercontent.com/40840436/145250595-756f4178-1c82-481d-ba25-046bb39e28a6.mov

src/main/window-manager.ts
jim-docker: ah, in an earlier DM from Alex he indicated to me this:
"Linux Lens window should have standard title bar with minimize/maximize/close, but menu should go under the sandwich"
so I guess there's a problem

src/renderer/components/layout/__tests__/topbar-linux.test.tsx
aleksfront: This means we do not want to show custom title buttons designed for windows only. Changed the test description.

src/main/menu.ts
aleksfront: No harm to run it, since I reverted linux view to show native menus also.

aleksfront:
Eventually, I reverted Linux view to use native controls. So current implementation looks like this:

**MacOS**

macos view

**Windows**

win view

**Ubuntu**

https://user-images.githubusercontent.com/9607060/145383960-e34f6873-8513-42c9-b68f-011381da6989.mov

Nokel81:
@aleksfront That ubuntu view looks strange. Why hide the menu until a click?

aleksfront:
> @aleksfront That ubuntu view looks strange. Why hide the menu until a click?

The menu is hidden by default and gets activated (showed) if you press `Alt` key. If you then just click outside the menu, it disappears.

Nokel81:
> The menu is hidden by default and gets activated (showed) if you press Alt key. If you then just click outside the menu, it disappears.

That seems suboptimal because you have to know to push ALT. There is #1734 for Linux, so I think we should do better.

jim-docker:
> > The menu is hidden by default and gets activated (showed) if you press Alt key. If you then just click outside the menu, it disappears.
> 
> That seems suboptimal because you have to know to push ALT. There is #1734 for Linux, so I think we should do better.

I agree, there needs to be some visual cue for the hidden menus on linux. I'd prefer it to be like windows, with the "sandwich", or keep the linux menu bar always visible.

jakolehm:
> > > The menu is hidden by default and gets activated (showed) if you press Alt key. If you then just click outside the menu, it disappears.
> > 
> > 
> > That seems suboptimal because you have to know to push ALT. There is #1734 for Linux, so I think we should do better.
> 
> I agree, there needs to be some visual cue for the hidden menus on linux. I'd prefer it to be like windows, with the "sandwich", or keep the linux menu bar always visible.

I'd also prefer Linux to be like Windows. Is there some deeper issue with Linux support?

aleksfront:
@jim-docker I've changed linux view to have custom window controls also. Here's current state on Win and Ubuntu 20.04. Please check linux view on your machine.
windows10
ubuntu20

src/main/index.ts
```diff
@@ -256,6 +256,12 @@ app.on("ready", async () => {
     LensProtocolRouterMain.getInstance().rendererLoaded = true;
   });
 
+  ipcMainOn(IpcMainWindowEvents.OPEN_CONTEXT_MENU, (event) => {
```
jakolehm: Should this be within `initializers.initIpcMainHandlers()` ?

src/main/window-manager.ts
```diff
@@ -81,7 +85,8 @@ export class WindowManager extends Singleton {
         show: false,
         minWidth: 700,  // accommodate 800 x 600 display minimum
         minHeight: 500, // accommodate 800 x 600 display minimum
-        titleBarStyle: "hiddenInset",
+        titleBarStyle: isMac ? "hiddenInset" : "hidden",
+        frame: isMac,
```
jakolehm, line 89: Does `frame: false` have any side-effects?

aleksfront: I suggest it doesn't. https://www.electronjs.org/docs/latest/api/frameless-window The window can be resized, minimized, maximized, closed and dragged.

aleksfront: Anyway, I can't hide title bar in linux with any other methods provided.

src/main/index.ts
aleksfront: Good catch, moved this into initializers.

jim-docker:
Looks great, works well on linux.

jim-docker (commented):
Seems a little odd that the popup menu position varies. I guess what is done in this video is correct because you don't want the mouse to be suddenly positioned over a menu item? I got it to be way off of the sandwich on one occasion (well over the nav bar) but can't reproduce. Some combination of window move/resize/click

https://user-images.githubusercontent.com/40840436/145910899-117342ff-dcda-4c44-a1f2-acb617575a8b.mov

aleksfront:
I've set menu position to be x: 20, y: 20 explicitly, it will show context menu always on icon center:

https://user-images.githubusercontent.com/9607060/145984393-2af1b32b-6027-4942-b1af-6b6ea4085775.mov

@jim-docker @Nokel81 please review this again.

src/main/initializers/ipc.ts
```diff
@@ -148,4 +149,16 @@ export function initIpcMainHandlers() {
...
+  ipcMainOn(IpcMainWindowEvents.OPEN_CONTEXT_MENU, async (event) => {
+    const menu = Menu.buildFromTemplate(getAppMenu(WindowManager.getInstance()));
+    const options = {
+      ...BrowserWindow.fromWebContents(event.sender),
+      // Center of the topbar menu icon
+      x: 20,
+      y: 20,
+    } as Electron.PopupOptions;
+
+    menu.popup(options);
```
Nokel81, line 162: Could this be done inline so as to remove the need for the cast?

src/renderer/components/layout/topbar.tsx
```diff
@@ -126,6 +158,18 @@ export const TopBar = observer(({ children, ...rest }: Props) => {
...
+          <div className={cssNames(styles.windowButtons, { [styles.linuxButtons]: isLinux })}>
+            <div className={styles.minimize} data-testid="window-minimize" onClick={minimizeWindow}>
+              <svg shapeRendering="crispEdges" viewBox="0 0 12 12"><rect fill="currentColor" width="10" height="1" x="1" y="9"></rect></svg></div>
+            <div className={styles.maximize} data-testid="window-maximize" onClick={toggleMaximize}>
+              <svg shapeRendering="crispEdges" viewBox="0 0 12 12"><rect width="9" height="9" x="1.5" y="1.5" fill="none" stroke="currentColor"></rect></svg>
+            </div>
+            <div className={styles.close} data-testid="window-close" onClick={closeWindow}>
+              <svg shapeRendering="crispEdges" viewBox="0 0 12 12"><polygon fill="currentColor" points="11 1.576 6.583 6 11 10.424 10.424 11 6 6.583 1.576 11 1 10.424 5.417 6 1 1.576 1.576 1 6 5.417 10.424 1"></polygon></svg>
+            </div>
+          </div>
```
Nokel81, line 171: Couldn't this just been "menu" icon?

Nokel81 (approved):

jim-docker (approved):

PR #4533, opened 2021-12-08
Fix crash in ClusterStatus

Nokel81:

Fixes issue reported on slack.

### DIFF
--- a/src/renderer/components/cluster-manager/cluster-status.tsx
+++ b/src/renderer/components/cluster-manager/cluster-status.tsx
@@ -153,7 +153,7 @@ export class ClusterStatus extends React.Component<Props> {
     return (
       <div className={cssNames(styles.status, "flex column box center align-center justify-center", this.props.className)}>
         <div className="flex items-center column gaps">
-          <h2>{this.entity?.getName()}</h2>
+          <h2>{this.entity?.getName() ?? this.cluster.name}</h2>
           {this.renderStatusIcon()}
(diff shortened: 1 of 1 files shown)


src/renderer/components/cluster-manager/cluster-status.tsx
```diff
@@ -153,7 +153,7 @@ export class ClusterStatus extends React.Component<Props> {
     return (
       <div className={cssNames(styles.status, "flex column box center align-center justify-center", this.props.className)}>
         <div className="flex items-center column gaps">
-          <h2>{this.entity.getName()}</h2>
+          <h2>{this.entity?.getName()}</h2>
```
aleksfront: It's a race condition or why we wouldn't have `entity` on this page?

Nokel81: Yes I think it is a race condition. I have only seen this issue when I reload the top view.

jim-docker: This will possibly set the name to `undefined`? Seems fishy that `entity` can be undefined. What was wrong with `this.cluster.preferences.clusterName`? At least as a fallback?

Nokel81: That is a good idea as a fallback since by default it is the "contextName".

aleksfront (approved):

jim-docker (approved):

PR #4535, opened 2021-12-08
Use electron.clipboard for all clipboard uses

### LINKED ISSUE #4529: Replace document.execCommand('copy') with Clipboard API
Currently, we're using `document.execCommand('copy')` command to copy text to clipboard. This is marked as `deprecated` and will be removed from the chrome soon. https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand#browser_compatibility

Instead, `Clipboard API` should be used https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API

Thanks @stefcameron for spotting this!

Nokel81:

fixes #4529


src/renderer/components/kubeconfig-dialog/kubeconfig-dialog.tsx
```diff
@@ -89,9 +89,8 @@ export class KubeConfigDialog extends React.Component<Props> {
   }
 
   copyToClipboard = () => {
-    if (this.config && copyToClipboard(this.configTextArea)) {
-      Notifications.ok("Config copied to clipboard");
-    }
+    clipboard.writeText(this.config);
```
jim-docker, line 92: No return value? Can it throw?

jim-docker (approved):

src/renderer/components/kubeconfig-dialog/kubeconfig-dialog.tsx
Nokel81: The documentation doesn't say it can. And the return value is `void`.

jim-docker:
tested on linux

jakolehm:
Issue described a solution using browser API but this PR is using Electron api, any reason for this?

Nokel81:
> Issue described a solution using browser API but this PR is using Electron api, any reason for this?

I did this for two reasons:

1. The browser API requires being in a secure context (HTTPS) according to MDN
2. The electron api was easy to use.

jweak:
> > Issue described a solution using browser API but this PR is using Electron api, any reason for this?
> 
> I did this for two reasons:
> 
> 1. The browser API requires being in a secure context (HTTPS) according to MDN
> 2. The electron api was easy to use.

We use the electron clipboard API already in a few places so I think good to use it here too.

jweak (approved):
LGTM

Tested on mac and worked on both dialogs

PR #4536, opened 2021-12-08
Fix freeze after closing terminal on windows

### LINKED ISSUE #4430: 5.3.0-latest.2021125.2 crash frequently on win10
windows version:  win10 20H2 19042.985

Nokel81:

fixes https://github.com/lensapp/lens/issues/4430

### DIFF
--- a/src/main/shell-session/shell-session.ts
+++ b/src/main/shell-session/shell-session.ts
@@ -158,6 +158,8 @@ export abstract class ShellSession {
         cwd,
         env,
         name: "xterm-256color",
+        // TODO: Something else is broken here so we need to force the use of winPty on windows
+        useConpty: false,
       }));
     }
 
--- a/src/renderer/api/terminal-api.ts
+++ b/src/renderer/api/terminal-api.ts
@@ -144,14 +144,6 @@ export class TerminalApi extends WebSocketApi<TerminalEvents> {
     this.socket.binaryType = "arraybuffer";
   }
 
-  destroy() {
-    if (!this.socket) return;
-    const controlCode = String.fromCharCode(4); // ctrl+d
-
-    this.sendMessage({ type: TerminalChannels.STDIN, data: controlCode });
-    setTimeout(() => super.destroy(), 2000);
-  }
-
   reconnect() {
     super.reconnect();
   }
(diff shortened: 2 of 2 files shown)


src/renderer/api/terminal-api.ts
```diff
@@ -144,14 +144,6 @@ export class TerminalApi extends WebSocketApi<TerminalEvents> {
     this.socket.binaryType = "arraybuffer";
   }
 
-  destroy() {
```
Nokel81: This is not needed, we kill the underlying process on normal socket close anyway

src/main/shell-session/shell-session.ts
```diff
@@ -158,6 +158,8 @@ export abstract class ShellSession {
         cwd,
         env,
         name: "xterm-256color",
+        // TODO: Something else is broken here so we need to force the use of winPty on windows
+        useConpty: false,
```
jim-docker, line 162: what does this do on mac and linux?

src/renderer/api/terminal-api.ts
jim-docker: Are you 100% sure this won't affect behaviour on all platforms? If this isn't related to the fix I'd leave it in.

src/main/shell-session/shell-session.ts
Nokel81: Nothing, this option is ignored on non-windows OS's

src/renderer/api/terminal-api.ts
Nokel81: Okay, I'll make a different PR to remove this.

src/main/shell-session/shell-session.ts
jim-docker: mention or reference the issue this is a workaround for, in the comment?

jim-docker (approved):
- The 1005 code issue persists, and the TypeError (no `url` on undefined). But no hang
- Console Window Host and winpty-agent.exe processes are getting orphaned (open a node shell, run `ls`, close the tab) Maybe this behaviour already existed.

- tested on windows
- lightly tested on mac and linux to make sure there are no side effects

If we need a fix now this looks like it will do.

jakolehm (approved):

PR #4538, opened 2021-12-08
Remove unnecesarry TerminalApi.destroy and TerminalApi.reconnect overloads

Nokel81:

### DIFF
--- a/src/renderer/api/terminal-api.ts
+++ b/src/renderer/api/terminal-api.ts
@@ -144,18 +144,6 @@ export class TerminalApi extends WebSocketApi<TerminalEvents> {
     this.socket.binaryType = "arraybuffer";
   }
 
-  destroy() {
-    if (!this.socket) return;
-    const controlCode = String.fromCharCode(4); // ctrl+d
-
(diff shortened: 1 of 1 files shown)


src/renderer/api/terminal-api.ts
```diff
@@ -144,18 +144,6 @@ export class TerminalApi extends WebSocketApi<TerminalEvents> {
     this.socket.binaryType = "arraybuffer";
   }
 
-  destroy() {
-    if (!this.socket) return;
-    const controlCode = String.fromCharCode(4); // ctrl+d
-
-    this.sendMessage({ type: TerminalChannels.STDIN, data: controlCode });
```
aleksfront, line 151: We don't need to send control code anymore?

Nokel81: This was supposed to be a followup to #4536.

We already clean up the PTY and the websocket either exits/closes. So I didn't think this was necessary. And from my testing it doesn't seem to be. But I might be missing something.

jim-docker (approved):

aleksfront (approved):

PR #4544, opened 2021-12-09
Special case goBack to navigate to catalog if no where to go back to

Nokel81:

This is really only useful in `make dev` when we restart from the settings view


jim-docker (commented):
same thing would apply for the Preferences and Extensions pages?

Nokel81:
> same thing would apply for the Preferences and Extensions pages?

Yes, and this PR fixes those too as all three don't override the default `back` function

jim-docker (approved):

aleksfront (approved):
This issue was a pain. Thanks for fixing this!

PR #4554, opened 2021-12-10
Fix extension engine range not working for some ^ ranges

### LINKED ISSUE #3404: Setting lens engine to ^5.1.0 in package.json for Lens 5.1.0 does not work
In relation to https://github.com/lensapp/lens/issues/3212, which was mostly fixed, I can set my `lens` engine, in package.json, to `^5.0.0` and Lens 5.1.0 will deem it compatible.

But if I set the engine to `^5.1.0` in Lens 5.1.0, then Lens deems it _incompatible_.

Doesn't seem right. And in the case of the https://github.com/Mirantis/lens-extension-cc extension, it was DOA in Lens until 5.1.0, so technically, I should be setting the engine to `^5.1.0` so that someone installing it in <5.1.0 isn't able to use it...

Lens version: `5.1.0-latest.20210715.1` on macOS 11.4

Nokel81:
- This is done by stripping non-API related prerelease parts from the
  version such as the "latest" prerelease part.

fixes #3404

### DIFF
--- a/src/extensions/__tests__/extension-compatibility.test.ts
+++ b/src/extensions/__tests__/extension-compatibility.test.ts
@@ -19,29 +19,24 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { isCompatibleExtension } from "../extension-compatibility";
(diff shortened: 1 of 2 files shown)


jakolehm (approved):

jweak (approved):
LGTM

PR #4556, opened 2021-12-10
Minor cleanup of Kubectl's downloadKubectl and writeInitScripts

Nokel81:

### DIFF
--- a/src/main/kubectl.ts
+++ b/src/main/kubectl.ts
@@ -27,12 +27,14 @@ import { ensureDir, pathExists } from "fs-extra";
 import * as lockFile from "proper-lockfile";
 import { helmCli } from "./helm/helm-cli";
 import { UserStore } from "../common/user-store";
-import { customRequest } from "../common/request";
 import { getBundledKubectlVersion } from "../common/utils/app-version";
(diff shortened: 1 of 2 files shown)


aleksfront (approved):

src/main/kubectl.ts
```diff
@@ -309,99 +311,86 @@ export class Kubectl {
...
-        });
-        resolve();
-      });
-      stream.pipe(file);
-    });
+    const downloadStream = got.stream({ url: this.url, decompress: true });
+    const fileWriteStream = fs.createWriteStream(this.path, { mode: 0o755 });
+    const pipeline = promisify(stream.pipeline);
+
+    await pipeline(downloadStream, fileWriteStream);
```
jim-docker, line 318: `try`/`catch`? Errors are still possible?

Nokel81: `promisify` above handles the errors and turns it into rejections

ixrock (approved):

src/main/kubectl.ts
jim-docker: So `'this.path` gets deleted in the event of an error?

Nokel81: Oh I see...

PR #4558, opened 2021-12-10
Fix ERR_UNSAFE_PORT from LensProxy

### LINKED ISSUE #4524: ERR_UNSAFE_PORT (-312) loading 'http://localhost:4045/'
**Describe the bug**
Today opening Lens I got this error:  
`ERR_UNSAFE_PORT (-312) loading 'http://localhost:4045/'`  
Seems like it is not a firewall problem, but just Lens seems scared to contact that port (like some browsers, where you should add something like `--explicitly-allowed-ports=4045`)

**To Reproduce**
1. Open Lens

**Expected behavior**
Lens should connect on its server

**Screenshots**
image
image

- Lens Version: 5.3.2
- OS: Windows

**Logs:**
```
info: [WINDOW-MANAGER]: Loading Main window from url: http://localhost:4045 ...
error: [WINDOW-MANAGER]: Failed to load Main window {"code":-312,"desc":"ERR_UNSAFE_PORT"}
error: Loading main window failed {"error":{"errno":-312,"code":"ERR_UNSAFE_PORT","url":"http://localhost:4045/"}}
info: [WINDOW-MANAGER]: Main window loaded
```

**Additional context**
Actually I didn't try to reboot my system, later I will try it but I still want report it because I think it could be useful for someone else or other reasons

Nokel81:
- Use the current list of ports from chromium as it is much easier to
  just reject using one of those instead of trying to handle the
  ERR_UNSAFE_PORT laod error from a BrowserWindow.on("did-fail-load")

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -28,7 +28,7 @@ import * as LensExtensionsCommonApi from "../extensions/common-api";
 import * as LensExtensionsMainApi from "../extensions/main-api";
 import { app, autoUpdater, dialog, powerMonitor } from "electron";
 import { appName, isIntegrationTesting, isMac, isWindows, productName } from "../common/vars";
-import { LensProxy, UnsafePortError } from "./lens-proxy";
+import { LensProxy } from "./lens-proxy";
 import { WindowManager } from "./window-manager";
 import { ClusterManager } from "./cluster-manager";
 import { shellSync } from "./shell-sync";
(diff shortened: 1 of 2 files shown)


Nokel81:
This can be tested by using the following generator to attempt to use the bad ports:

```
const portNumbers = (function*() {
  yield 1719;
  yield 1720;
  yield 1723;
  yield 2049;
  yield 3659;
  yield 4045;
  yield 5060;
  yield 5061;
  yield 6000;
  yield 6566;
  yield 6665;
  yield 6666;
  yield 6667;
  yield 6668;
  yield 6669;
  yield 6697;
  yield 10080;

  for (;;) {
    yield 0;
  }
})();
```

src/main/index.ts
```diff
@@ -184,13 +184,39 @@ app.on("ready", async () => {
...
   initializers.initClusterMetadataDetectors();
 
-  try {
-    logger.info("🔌 Starting LensProxy");
-    await lensProxy.listen();
-  } catch (error) {
-    dialog.showErrorBox("Lens Error", `Could not start proxy: ${error?.message || "unknown error"}`);
+  logger.info("🔌 Starting LensProxy");
+
+  for (let attempt = 0;; attempt += 1) {
```
jakolehm: I think retries should be hidden within `lensProxy.listen()` .. if it cannot find proper port then throw.

Nokel81: Sounds good. 

I had initially tried catching the error from the load itself but felt that that was too complicated.

jakolehm (approved):
LGTM

jweak (approved):
LGTM

PR #4564, opened 2021-12-13
Cleanup HelmRepoManager and <HelmCharts>

Nokel81:
- Makes integration tests less flaky

- Seperates the two loading states so that the UI updates faster

Follow up to #4514

### DIFF
--- a/src/renderer/components/+preferences/add-helm-repo-dialog.tsx
+++ b/src/renderer/components/+preferences/add-helm-repo-dialog.tsx
@@ -61,11 +61,11 @@ export class AddHelmRepoDialog extends React.Component<Props> {
   @observable showOptions = false;
 
   @action
-  close = () => {
-    AddHelmRepoDialog.close();
-    this.helmRepo = getEmptyRepo();
-    this.showOptions = false;
-  };
+    close = () => {
+      AddHelmRepoDialog.close();
+      this.helmRepo = getEmptyRepo();
+      this.showOptions = false;
+    };
 
   setFilepath(type: FileType, value: string) {
     this.helmRepo[type] = value;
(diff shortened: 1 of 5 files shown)


src/renderer/components/+preferences/add-helm-repo-dialog.tsx
```diff
@@ -71,12 +73,13 @@ export class AddHelmRepoDialog extends React.Component<Props> {
     dialogState.isOpen = false;
   }
 
-  @observable helmRepo : HelmRepo = this.emptyRepo;
+  @observable helmRepo = getEmptyRepo();
```
nevalla, line 60: just curious, what is the benefit of introducing this as external function instead of class method?

Nokel81: Nothing really, it could have been a private method yes. I just don't think it should be a plain object because then we are just assigning a reference which is a bit confusing IMO.

src/renderer/components/+preferences/add-helm-repo-dialog.tsx
```diff
@@ -107,7 +110,7 @@ export class AddHelmRepoDialog extends React.Component<Props> {
 
   async addCustomRepo() {
     try {
-      await HelmRepoManager.addCustomRepo(this.helmRepo);
+      await HelmRepoManager.getInstance().addRepo(this.helmRepo);
       Notifications.ok(<>Helm repository <b>{this.helmRepo.name}</b> has added</>);
```
jim-docker: noticed this while testing:
```suggestion
      Notifications.ok(<>Helm repository <b>{this.helmRepo.name}</b> has been added</>);
```

jim-docker (dismissed):

nevalla (approved):

jim-docker (approved):

PR #4570, opened 2021-12-14
Show splash window until bundled extensions have been loaded

jweak:
Change app start up so that startup waits until the bundled extensions have been loaded and enabled before splash window is hidden and main window is shown.

This is because some of the extensions want to render UI elements and they should be ready when the main screen is shown. Less flicker and better UX.

### DIFF
--- a/src/extensions/extension-loader/extension-loader.ts
+++ b/src/extensions/extension-loader/extension-loader.ts
@@ -313,7 +313,7 @@ export class ExtensionLoader {
   }
 
   protected autoInitExtensions(register: (ext: LensExtension) => Promise<Disposer[]>) {
-    const bundledLoaded: Promise<void>[] = [];
+    const loadingExtensions: { isBundled: boolean, loaded: Promise<void> }[] = [];
 
     reaction(() => this.toJSON(), installedExtensions => {
       for (const [extId, extension] of installedExtensions) {
@@ -334,9 +334,10 @@ export class ExtensionLoader {
               logger.error(`${logModule}: failed to enable`, { ext: extension, err });
             });
 
-            if (extension.isBundled) {
-              bundledLoaded.push(loaded);
-            }
+            loadingExtensions.push({
+              isBundled: extension.isBundled,
+              loaded,
+            });
             this.instances.set(extId, instance);
           } catch (err) {
             logger.error(`${logModule}: activation extension error`, { ext: extension, err });
@@ -349,7 +350,7 @@ export class ExtensionLoader {
       fireImmediately: true,
     });
 
-    return Promise.all(bundledLoaded);
+    return loadingExtensions;
   }
 
   protected requireExtension(extension: InstalledExtension): LensExtensionConstructor | null {
--- a/src/extensions/extension-loader/extension-loader.ts
+++ b/src/extensions/extension-loader/extension-loader.ts
@@ -255,7 +255,8 @@ export class ExtensionLoader {
 
   loadOnClusterManagerRenderer() {
     logger.debug(`${logModule}: load on main renderer (cluster manager)`);
-    const bundledLoaded = this.autoInitExtensions(async (extension: LensRendererExtension) => {
+
+    return this.autoInitExtensions(async (extension: LensRendererExtension) => {
       const removeItems = [
         registries.GlobalPageRegistry.getInstance().add(extension.globalPages, extension),
         registries.AppPreferenceRegistry.getInstance().add(extension.appPreferences),
@@ -278,8 +279,6 @@ export class ExtensionLoader {
 
       return removeItems;
     });
-
-    return bundledLoaded;
   }
 
   loadOnClusterRenderer() {
--- a/src/extensions/extension-loader/extension-loader.ts
+++ b/src/extensions/extension-loader/extension-loader.ts
@@ -255,7 +255,7 @@ export class ExtensionLoader {
 
   loadOnClusterManagerRenderer() {
     logger.debug(`${logModule}: load on main renderer (cluster manager)`);
-    this.autoInitExtensions(async (extension: LensRendererExtension) => {
+    const bundledLoaded = this.autoInitExtensions(async (extension: LensRendererExtension) => {
       const removeItems = [
         registries.GlobalPageRegistry.getInstance().add(extension.globalPages, extension),
         registries.AppPreferenceRegistry.getInstance().add(extension.appPreferences),
@@ -278,6 +278,8 @@ export class ExtensionLoader {
 
       return removeItems;
     });
+
+    return bundledLoaded;
   }
 
   loadOnClusterRenderer() {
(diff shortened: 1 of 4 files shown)


Nokel81:
I don't know if I agree that this is the best course of action. Would it not be possible instead to render a "dummy" component that is roughly the correct size and shape but has a loading animation on top (similar to what some websites do and what I did for the cluster sidebar icon)?

That would therefore mean that the app still loads just as fast and we wouldn't have this additional loopback dependency.

jweak:
> I don't know if I agree that this is the best course of action. Would it not be possible instead to render a "dummy" component that is roughly the correct size and shape but has a loading animation on top (similar to what some websites do and what I did for the cluster sidebar icon)?
> 
> That would therefore mean that the app still loads just as fast and we wouldn't have this additional loopback dependency.

The dummy component wouldn't be rendered until the extension has loaded. Anyway a bigger problem that dummy component wouldn't solve is that the Spaces extension wants to show a fullscreen modal that should be visible already when main window is shown and not after a while (like it does now).

src/extensions/extension-loader.ts
```diff
@@ -271,7 +271,7 @@ export class ExtensionLoader extends Singleton {
 
   loadOnClusterManagerRenderer() {
     logger.debug(`${logModule}: load on main renderer (cluster manager)`);
-    this.autoInitExtensions(async (extension: LensRendererExtension) => {
+    const bundledLoaded = this.autoInitExtensions(async (extension: LensRendererExtension) => {
```
Nokel81: nit: why not just `return this.autoInitExtensions(...`

Nokel81:
> The dummy component wouldn't be rendered until the extension has loaded. Anyway a bigger problem that dummy component wouldn't solve is that the Spaces extension wants to show a fullscreen modal that should be visible already when main window is shown and not after a while (like it does now).

Fair enough, if this is for a fullscreen modal then I don't really see any other way to do it. I wonder if it needs to be one but I don't really know what you are trying to achieve.

src/extensions/extension-loader.ts
jweak: yeah might as well. done

Nokel81 (dismissed):

jweak:
Had a unsigned commit. Fixed.

Nokel81 (dismissed):

jweak:
Rebased to fix conflicts

Nokel81 (dismissed):

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -339,6 +348,8 @@ export class ExtensionLoader {
     }, {
       fireImmediately: true,
     });
+
+    return Promise.all(bundledLoaded);
```
aleksfront: `autoInitExtensions()` method shouldn't return only specific promises with bundled extensions, it's more generic by its name? Maybe we can return all loaded extensions and then filter them some other place?

aleksfront (commented):
While testing I noticed that bundled extensions are still loading after main window is opened. Pod menu is an example:

https://user-images.githubusercontent.com/9607060/146716391-6363b421-216d-4081-bd88-c5dcff19ffec.mov

jweak:
> While testing I noticed that bundled extensions are still loading after main window is opened. Pod menu is an example:
> 
>  bundled.extensions.wait.mov

Interesting. I think it might be because extensions are loaded in 3 places, main and `root-frame` init on startup and `cluster-frame` init https://github.com/lensapp/lens/blob/f8fa33ec8cde5dcc5b56f77262b6ccb233f13115/src/renderer/cluster-frame.tsx#L105-L115

I think these menus are loaded in the third phase. This PR is currently waiting in the `root-frame` phase.  

Also this approach that I take requires the extension to override `LensExtension::onActivate`-method to do it's initial loading which many of the standard bundled extensions don't so they return immediately on the startup.

ixrock (dismissed):

src/extensions/extension-loader/extension-loader.ts
jweak: good point. fixed.

panuhorsmalahti (dismissed):

src/renderer/root-frame.tsx
```diff
@@ -54,8 +54,17 @@ export class RootFrame extends React.Component {
     lensProtocolRouterRendererInjectable: LensProtocolRouterRenderer,
   ) {
     catalogEntityRegistry.init();
-    extensionLoader.loadOnClusterManagerRenderer();
+
+    try {
+      const loadingExtensions = extensionLoader.loadOnClusterManagerRenderer();
+      const loadingBundledExtensions = loadingExtensions.filter(e => e.isBundled).map(e => e.loaded);
+
+      await Promise.all(loadingBundledExtensions);
```
jim-docker: Will we be stuck on the splash screen if a bundled extension gets stuck loading? A log message after some timeout might help with diagnosing the problem, if it happens to an end-user. Not sure if that is easy to do here, or somewhere else.

jweak: Hmm, yeah that's a good point. It shouldn't block if extension fails but if it's stuck loading will do that. I think a timeout should be doable. It's not end of the world even if we continue and the extension keeps loading. I wonder what's a good time out, 10 seconds?

jim-docker: Yes, having a slightly worse UX (what we have now) when an extension takes too long (which really shouldn't happen for bundled ones) is better than being stuck on the splash screen. I would think 10 seconds from the time the extensions start loading is adequate.

src/renderer/root-frame.tsx
```diff
@@ -54,8 +54,17 @@ export class RootFrame extends React.Component {
     lensProtocolRouterRendererInjectable: LensProtocolRouterRenderer,
   ) {
     catalogEntityRegistry.init();
-    extensionLoader.loadOnClusterManagerRenderer();
+
+    try {
+      const loadingExtensions = extensionLoader.loadOnClusterManagerRenderer();
+      const loadingBundledExtensions = loadingExtensions.filter(e => e.isBundled).map(e => e.loaded);
```
jim-docker, line 64: I made this change just so I could test what the UI looked like after waiting for all extensions. 
```suggestion
      const loadingBundledExtensions = loadingExtensions;
```
It still seems that the extensions are still loading when the UI comes up? Or maybe this is still expected?

https://user-images.githubusercontent.com/40840436/146953159-4c3ea9f0-8158-4df6-ba40-4b806d54c27c.mov

jweak: That change doesn't work since the array has objects `{extension: extension, loaded: loadingPromise}` so you'd want to have 

`const allExtensions = loadingExtensions.map(e => e.loaded)`
and
`await Promise.all(allExtensions)`

jim-docker: oops, missed the map

jim-docker: I still get the same result, are my expectations wrong?

src/renderer/root-frame.tsx
jweak: I added a 10 second timeout now. I tested it with copying one of the extensions from lens-ide repo to the `lens/extensions`, and added 20 second delay on the `onActivate`-method on `renderer.tsx`. `make build` and open the app.

src/renderer/root-frame.tsx
jweak: Interesting. I'll test it. There are couple of things why this could happen. This approach waits only on the extensions' `onActivate()` method to finish. React rendering can happen after that. I did try with Lens Spaces extension bundled and it did wait for it enough to do rendering immediately. Which kind of the point of the change anyway since we want to have the login screen ready when the app opens.

jim-docker (approved):
LGTM, there is still an issue when an extension may also be found in `~/.k8slens/extensions` but it should only impact Lens devs, and can be addressed in another PR.

panuhorsmalahti (approved):

PR #4555, opened 2021-12-10
Fix not being able to clear set cluster icon

Nokel81:
- Use `null` as a sentinal value distinct from `undefined` so that
  extensions can still set and clear icons themselves

Signed-off-by: Sebastian Malton <sebastian@malton.name>

fixes #4540

PR #4571, opened 2021-12-14
Add support for specifying an entity's shortName

### LINKED ISSUE #3498: Modify labels on cluster icons
Instead of the three letter name allow users to give it a name that is meaningful to them.

All entities.

### LINKED ISSUE #5401: Please allow 4 chars in hotbar (cluster icons bar)
Please give me a way to put a few more characters in the hotbar box name:

- a way to choose helvetica narrow for the hotbar box name
- a way to choose a smaller font for the hotbar box name
- fix a bug that allows lens to put 3 chars in a hotbar box name but i can only put 2 chars.
- enlarge the icons by 3-5 pixels on all sides to allow 1 more character to fit - up to 4 chars.

Right now, lens assigns two letters and a number in a somewhat unpredictable way to my aws arns.  For some reason, lens can assign 3-letter names such as 'AE1' but when i try to edit the arn-name (Lens > hotbar > right-click > settings > Cluster Name) and type in 'AE1' I only get 'AE' in the icon.  This is a frustrating user experience.    

**Why is this needed**:
It's difficult to distinguish 20 clusters with 2-character names (for example, we have stg1 and stg2 clusters, i am forced to use 'S1' and 'S2' because 'ST1' and 'ST2' does not work).  Similarly infra1 and infrastg, what would be my 2-char names for these?  I1 and IG?  hard to remember.  Also in my environment we have red/blue dns failover cluster names such as "a-staging1" and "b-staging1", so i potentially have 40 clusters to distinguish, a-stg1, b-stg1, etc., ast1 and bst1, ast2, bst2 just barely would work for my limited memory, without twisting our own naming schemes too much.

eks k8s 1.21
macbook pro / big sur / 11.6.5.

Nokel81:


package.json
```diff
@@ -3,7 +3,7 @@
   "productName": "OpenLens",
   "description": "OpenLens - Open Source IDE for Kubernetes",
   "homepage": "https://github.com/lensapp/lens",
-  "version": "5.3.0",
+  "version": "5.4.0-alpha.1",
```
Nokel81: I need to do this so that I can add a new migration

src/common/catalog/helpers.ts
```diff
@@ -0,0 +1,71 @@
...
+  const third = rawThird ? splitter.iterateGraphemes(rawThird) : iter.newEmpty();
+
+  return [
+    ...iter.take(first, 1),
+    ...iter.take(second, 1),
+    ...iter.take(third, 1),
+  ].filter(Boolean).join("");
+}
+
+export function getShortName(entity: CatalogEntity): string {
```
Nokel81: I didn't really think that this should be a part of `CatalogEntity`

src/main/catalog-pusher.ts
```diff
@@ -25,15 +25,25 @@ import type { CatalogEntityRegistry } from "./catalog";
...
+  }
+
+  return { metadata, spec, status, kind, apiVersion };
+}
+
+const broadcaster = debounce((items: (CatalogEntityData & CatalogEntityKindData)[]) => {
   broadcastMessage("catalog:items", items);
 }, 1_000, { leading: true, trailing: true });
 
 export function pushCatalogToRenderer(catalog: CatalogEntityRegistry) {
```
Nokel81: I made the modifications here so that we won't have to deal with issues like https://github.com/lensapp/lens/pull/4555.

That is because this acts as a pure reducer

src/renderer/components/+catalog/catalog-entity-details.tsx
```diff
@@ -27,32 +27,34 @@ import type { CatalogCategory, CatalogEntity } from "../../../common/catalog";
...
 import { cssNames } from "../../utils";
 import { Avatar } from "../avatar";
+import { getIconColourHash } from "../../../common/catalog/helpers";
+import { EntityIcon } from "../entity-icon";
+import { getLabelBadges } from "./helpers";
 
-interface Props<T extends CatalogEntity> {
-  item: CatalogEntityItem<T> | null | undefined;
+interface Props {
+  entity: CatalogEntity | null | undefined;
```
Nokel81: The main reason for removing `CatalogEntityItem` was because it was unnecessary. `CatalogItem` already implements the `ItemObject` interface so the second level of abstraction didn't gain us anything.

The only function that needed to be factored out was `getLabelBadges` but that can (and now is) easily a free function.

src/renderer/components/+catalog/catalog-entity-details.tsx
```diff
@@ -27,32 +27,34 @@ import type { CatalogCategory, CatalogEntity } from "../../../common/catalog";
...
+
+    return <Icon {...{ [key]: category.metadata.icon }} smallest />;
   }
 
-  renderContent(item: CatalogEntityItem<T>) {
-    const detailItems = CatalogEntityDetailRegistry.getInstance().getItemsForKind(item.kind, item.apiVersion);
-    const details = detailItems.map(({ components }, index) => {
-      return <components.Details entity={item.entity} key={index}/>;
-    });
+  renderContent(entity: CatalogEntity | undefined, onRun: () => void) {
```
Nokel81: The other change that was required by removing `CatalogEntityItem` was the injection of the `onRun` function here because it needs to go through the `catalogEntityRegistry` but so as to keep it injectable, I did not import that.

src/renderer/components/avatar/avatar.tsx
```diff
@@ -21,70 +21,20 @@
...
 import styles from "./avatar.module.css";
 
-import React, { HTMLAttributes, ImgHTMLAttributes } from "react";
+import React, { HTMLAttributes } from "react";
 import randomColor from "randomcolor";
-import GraphemeSplitter from "grapheme-splitter";
-import { cssNames, iter } from "../../utils";
+import { cssNames } from "../../utils";
 
 export interface AvatarProps extends HTMLAttributes<HTMLElement> {
```
Nokel81: These changes to `<Avatar />` were made to make it more reusable. Namely, I felt that there was too much business logic going on in here. Namely the computation of `shortName`'s.

Futhermore, because only `src` was a field (and not icon). There was a strange disconnect where the use of `<Avatar>` would include both passing in an optional `src` as well as optionally rendering an `<Icon />`. 

For that reason, I made `<EntityIcon>` to do all the work instead.

src/common/catalog/catalog-entity.ts
```diff
@@ -38,14 +38,38 @@ export type CatalogEntityConstructor<Entity extends CatalogEntity> = (
 );
 
 export interface CatalogCategoryVersion<Entity extends CatalogEntity> {
+  /**
+   * The specific version that the associated constructor is for. This MUST be
+   * a DNS label and SHOULD be of the form `vN`, `vNalphaY`, or `vNbetaY` where
+   * `N` and `Y` are both integers greater than 0.
+   */
```
aleksfront: This all relates to `name: string` prop? Maybe it needs some example.

src/common/catalog/catalog-entity.ts
```diff
@@ -233,6 +265,14 @@ export abstract class CatalogEntity<
     return this.metadata.name;
   }
 
+  public getSource(): string {
+    return this.metadata.source ?? "unknown";
+  }
+
+  public isEnabled(): boolean {
+    return this.status.enabled ?? true;
+  }
```
aleksfront: If no `status.enabled` then `isEnabled` returns true? Shouldn't it be otherwise, just
```
return !!this.status.enabled
```

src/common/entity-preferences-store.ts
```diff
@@ -0,0 +1,72 @@
...
+    makeObservable(this);
+    this.load();
+  }
+
+  @action
+  mergePreferences(entityId: string, preferences: PartialDeep<EntityPreferencesModel>): void {
+    if (!this.preferences.has(entityId)) {
+      this.preferences.set(entityId, preferences);
+    } else {
+      this.preferences.set(entityId, merge(this.preferences.get(entityId), preferences));
```
aleksfront: Should work fine without `if` checks above. With `undefined` value it will be merged properly. Try in repl https://lodash.com/docs/4.17.15#merge

src/extensions/registries/entity-setting-registry.ts
```diff
@@ -32,39 +33,86 @@ export interface EntitySettingComponents {
...
 }
 
-export interface RegisteredEntitySetting extends EntitySettingRegistration {
+export interface RegisteredEntitySetting extends Omit<EntitySettingRegistration, "apiVersions"> {
   id: string;
+  priority: number;
+  /**
+   * If `true` then this item matches all versions, otherwise is a set of versions it supports
+   */
+  apiVersions: Set<string> | true;
```
aleksfront: These don't relate to using entity short name?

src/common/catalog/helpers.ts
```diff
@@ -0,0 +1,71 @@
...
+  const byDashes = name.split(/[-_]+/);
+
+  if (byDashes.length > 1) {
+    return byDashes;
+  }
+
+  return name.split(/@+/);
+}
+
+export function computeDefaultShortName(name: string) {
```
aleksfront: Since all methods are in separate helper file, it's a good opportunity to have unit test for them.

aleksfront (commented):
1. Can you add a bit more description of why we need specifying short names for entities?
2. Is it possible to separate removing `CatalogEntityItem` into separate PR?

Nokel81:
> 1. Can you add a bit more description of why we need specifying short names for entities?

Sure, this is request that came through from Lee and it decently upvoted github issue. Furthermore, I think that this is a good first item for entity preferences within Lens which hopefully can be added to with custom icons down the line.

> 2. Is it possible to separate removing `CatalogEntityItem` into separate PR?

Sure, I will do that and then we can make this PR dependent on that one.

src/common/catalog/catalog-entity.ts
Nokel81: Sure I can add some examples

src/common/catalog/catalog-entity.ts
Nokel81: Nope, the documentation says that the default value is `true`

Nokel81: Plus this is what CatalogEnityitem already does

src/extensions/registries/entity-setting-registry.ts
Nokel81: They do actually because I want to add the entity setting to all entity types.

Nokel81:
Some screenshots:
Screen Shot 2022-04-01 at 5 44 34 PM
Screen Shot 2022-04-01 at 5 44 28 PM

Nokel81:
@aleksfront PTALA

src/common/hotbars/store.ts
```diff
@@ -349,3 +374,23 @@ function ensureExactHotbarItemLength(hotbar: Hotbar) {
     }
   }
 }
+
+/**
+ * This function ensures that the data coming in has the correct form
+ * @param hotbar The hotbar to modify
+ */
+function ensureNamesAndShortNames(hotbar: Hotbar) {
+  for (let i = 0; i < hotbar.items.length; i += 1) {
```
jim-docker: I can't remember the last time you used a for loop like this. no map, filter, etc.?

src/common/hotbars/types.ts
```diff
@@ -7,15 +7,16 @@ import * as uuid from "uuid";
 import type { Tuple } from "../utils";
 import { tuple } from "../utils";
 
+export interface HotbarItemEntity {
+  uid: string;
+  name: string;
+  shortName: string;
```
jim-docker: breaking change?

jim-docker (commented):
This is massive, with changes added a year later. Can the recent commits be in a separate PR?

Shouldn't the short name be limited to a small number of characters, say 5?

This doesn't look good:
Screen Shot 2023-01-18 at 2 56 09 PM

src/common/hotbars/types.ts
Nokel81: Don't think this type is exported.

src/common/hotbars/store.ts
Nokel81: I guess I could have used `map`

Nokel81:
Fair enough. Sure we can limit it to 5 graphemes

src/common/hotbars/types.ts
jim-docker: Is `HotbarItem` exported?

Nokel81: No, and neither is `Hotbar`

Nokel81:
> This is massive, with changes added a year later. Can the recent commits be in a separate PR?

I don't think the recency of a commit should matter....

Some of the more recent commits adding the necessary refactorings to get the behavioural tests (which were only added recently) to pass.

jim-docker:
> > This is massive, with changes added a year later. Can the recent commits be in a separate PR?
> 
> I don't think the recency of a commit should matter....
> 
> Some of the more recent commits adding the necessary refactorings to get the behavioural tests (which were only added recently) to pass.

I think this hasn't been approved for so long because it is so large, and it has evolved. Better to push to get it approved quickly than let it linger. And I don't think the title/description convey the scope of the work. I can approve it but it will need to soak in master/alpha, and not for a patch release if it doesn't make it in 6.4.0

Nokel81:
> I think this hasn't been approved for so long because it is so large, and it has evolved. Better to push to get it approved quickly than let it linger. And I don't think the title/description convey the scope of the work. I can approve it but it will need to soak in master/alpha, and not for a patch release if it doesn't make it in 6.4.0

Okay lets make a 6.4.0 branch first

jim-docker:
> Fair enough. Sure we can limit it to 5 graphemes

Also, it would seem more natural for the short name editing to be in the general tab, with the full name, instead of its own tab. Extensions/custom catalog entities make it difficult to do so?

Nokel81:
> Also, it would seem more natural for the short name editing to be in the general tab, with the full name, instead of its own tab. Extensions/custom catalog entities make it difficult to do so?

Yes I agree. Let me see if I can do that. Probably as a follow up PR because currently setting pages are "whole" and not partial pages, like preferences.

src/common/hotbars/store.ts
Nokel81: Messes with the types too much if we have it as a map currently.

src/common/cluster-store/cluster-store.injectable.ts
```diff
@@ -36,6 +37,7 @@ const clusterStoreInjectable = getInjectable({
     persistStateToConfig: di.inject(persistStateToConfigInjectionToken),
     enlistMessageChannelListener: di.inject(enlistMessageChannelListenerInjectionToken),
     shouldDisableSyncInListener: di.inject(shouldBaseStoreDisableSyncInIpcListenerInjectionToken),
+    sendMessageToChannel: di.inject(sendMessageToChannelInjectionToken),
```
jim-docker: There's obviously an existing pattern for this but it seems painful that this change has to be done for every store.

src/extensions/extensions-store/extensions-store.injectable.ts
```diff
@@ -27,6 +28,7 @@ const extensionsStoreInjectable = getInjectable({
     persistStateToConfig: di.inject(persistStateToConfigInjectionToken),
     enlistMessageChannelListener: di.inject(enlistMessageChannelListenerInjectionToken),
     shouldDisableSyncInListener: di.inject(shouldBaseStoreDisableSyncInIpcListenerInjectionToken),
+    sendMessageToChannel: di.inject(sendMessageToChannelInjectionToken),
```
jim-docker: Isn't this already done in the `ExtensionStore` constructor?

jim-docker (commented):
Bigger still. Maybe @jansav or @Iku-turso should have a quick look too?

src/extensions/extensions-store/extensions-store.injectable.ts
Nokel81: Haha different type all together. I get messed up too.

There is `ExtensionStore` which is exported to the extension API and then there is `ExtensionsStore` which keeps track of if an extension is enabled or not.

src/common/cluster-store/cluster-store.injectable.ts
Nokel81: Yes I agree this is quite painfull.

I think a better solution would be for `fromStore` and `toJson` to be configuration params to `BaseStore` instead of `abstract` methods.

That way we could have one `createStoreInjectable` that is referenced instead of all of the dependencies being mandatory in many places.

Namely, composition over inheritance.

src/extensions/extensions-store/extensions-store.injectable.ts
jim-docker: oops

Nokel81:
Closing, might recreate in future.

PR #4575, opened 2021-12-15
Fix namespace selector performance issues on large clusters

jakolehm:
Switches to "search only mode" if user has > 500 namespaces. This seems to help a lot with clusters that have huge number of namespaces (tested with >12k namespaces).

Note: this changes ns selector behaviour a bit. If user selects "All namespaces" it does not mark anymore every namespace as selected, instead it marks "All namespaces" as selected. I had to make this change so that logic works also with "search only mode".

https://user-images.githubusercontent.com/1446224/146182668-183eaf9b-385e-413d-92db-d77219d73413.mov


src/renderer/components/+namespaces/namespace-select.tsx
```diff
@@ -82,6 +82,24 @@ export class NamespaceSelect extends React.Component<Props> {
...
+
+    if (namespaceStore.items.length > 500 && rawInput.length < 3) {
+      return false;
+    }
+
+    if (rawInput && !option.value.includes(rawInput)) {
+      return false;
+    }
+
+    console.log("raw", rawInput);
```
Nokel81: Did you mean to leave this in?

Nokel81:
> Note: this changes ns selector behaviour a bit. If user selects "All namespaces" it does not mark anymore every namespace as selected, instead it marks "All namespaces" as selected. I had to make this change so that logic works also with "search only mode".

I am slightly concerned about this change as it means that you can no longer deselect a namespace from the "all-namespaces" set.

src/renderer/components/+namespaces/namespace-select.tsx
jakolehm: Did not mean to leave, good catch!

jakolehm:
> > Note: this changes ns selector behaviour a bit. If user selects "All namespaces" it does not mark anymore every namespace as selected, instead it marks "All namespaces" as selected. I had to make this change so that logic works also with "search only mode".
> 
> I am slightly concerned about this change as it means that you can no longer deselect a namespace from the "all-namespaces" set.

You still can, it just looks a bit different.

src/renderer/components/+namespaces/namespace-select-filter.tsx
```diff
@@ -91,27 +95,33 @@ export class NamespaceSelectFilter extends React.Component<SelectProps> {
...
-      return (
-        <div className="flex gaps align-center">
-          <Icon small material="layers" />
-          <span>{namespace}</span>
-          {isSelected && <Icon small material="check" className="box right" />}
-        </div>
-      );
+      isSelected = !namespaceStore.areAllSelectedImplicitly && namespaceStore.hasContext(namespace);
+    } else {
+      isSelected = namespaceStore.areAllSelectedImplicitly;
```
Nokel81: maybe this can be simplified? I just think it is odd to assign to `false` and then override it completely.

```ts
const isSelected = namespace
  ? !namespaceStore.areAllSelectedImplicitly && namespaceStore.hasContext(namespace)
  : namespaceStore.areAllSelectedImplicitly
```

jakolehm: Yes, will cleanup this.

jakolehm:
@Nokel81 ptal again.

Nokel81 (approved):

jim-docker:
I'm seeing issues on linux when connected to a large cluster (i.e. prod's catalog cluster):

https://user-images.githubusercontent.com/40840436/146268910-a1425da6-cbe8-42f6-8f14-f5d435edf2cf.mov

The above video shows trying to select multiple namespace search matches. The Lens UI freezes, eventually the logs show:

```
[0] error:   ▪ [CLUSTER]: Failed to connect to "Default": RequestError: Error: socket hang up +1ms
[0] [CLUSTER-CONNECT] 3 active sockets
[0] [CLUSTER-CONNECT] 3 active sockets
[0] 
[0] (electron:23882): Gtk-WARNING **: 15:46:54.138: Could not load a pixbuf from icon theme.
[0] This may indicate that pixbuf loaders or the mime database could not be found.
[0] **
[0] Gtk:ERROR:../../../../gtk/gtkiconhelper.c:494:ensure_surface_for_gicon: assertion failed (error == NULL): Failed to load /org/gtk/libgtk/icons/16x16/status/image-missing.png: Unrecognized image file format (gdk-pixbuf-error-quark, 3)
[0] /home/jim/git/lensapp/lens/node_modules/electron/dist/electron exited with signal SIGABRT
[0] [nodemon] app crashed - waiting for file changes before starting...
```
I left Lens running after that and the screen saver eventually kicked in so maybe that precipitated the crash?

5.3.3 on linux is also laggy, has connect issues, and namespace selection also makes the UI laggy or non-responsive, so this PR doesn't necessarily make things worse on linux...

jim-docker:
This is the app from a `make build` run (and I installed the Lens spaces extension manually):

https://user-images.githubusercontent.com/40840436/146278345-571de8c0-3efb-435b-8b25-9948179fbec0.mov

When the app started up the selected namespaces was initially `default` I changed it to All Namespaces, then tried to search for `space-001`. The results eventually showed up (after the end of the video above). It also froze the UI after I tried to ctrl-click one of them (like the video in my comment above). And then the core dumped:
```
[CLUSTER-CONNECT] 6 active sockets
[CLUSTER-CONNECT] 6 active sockets
Segmentation fault (core dumped)
```
Another thing seems odd, that it tried to do a watch on pods using the array of what looks like all of the namespaces. Should that have happened? The selected namespaces went from `default` to All Namespaces, which should use the equivalent of `--all-namespaces` not an array of all the namespaces, right?

src/renderer/components/+namespaces/namespace-select.tsx
```diff
@@ -82,6 +82,22 @@ export class NamespaceSelect extends React.Component<Props> {
     );
   };
 
+  filterOption = (option: SelectOption, rawInput: string) => {
+    if (option.value === "" || (!namespaceStore.areAllSelectedImplicitly && namespaceStore.selectedNames.has(option.value))) {
+      return true;
+    }
+
+    if (namespaceStore.items.length > 500 && rawInput.length < 3) {
```
ixrock: _nit:_ Maybe better to extract expression `namespaceStore.items.length > 500 && rawInput.length < 3` to some variable with meaningful name? Would add better understanding of why/when quit the function. (+similar for line:94)

ixrock (dismissed):
LGTM

Nokel81 (dismissed):
Just requesting changes so that this isn't merged before the bugs that were found are fixed.

src/renderer/components/+namespaces/namespace-select-filter.tsx
```diff
@@ -49,14 +61,22 @@ const Placeholder = observer((props: PlaceholderProps<any, boolean>) => {
   };
 
   return (
-    <components.Placeholder {...props}>
+    <components.Placeholder {...props} data-testid="namespace-select-filter-placeholder">
```
ixrock, line 64: _nit:_ Noticed in some places we're using `data-test-id` too (probably better to stick to single definition)

src/renderer/components/+namespaces/namespace-select-filter.tsx
```diff
@@ -49,14 +61,22 @@ const Placeholder = observer((props: PlaceholderProps<any, boolean>) => {
...
 });
 
+export interface NamespaceSelectFilterProps extends NamespaceSelectProps {
+  maxItems?: number;
+}
+
+interface NamespaceSelectFilterDependencies {
+  namespaceStore: NamespaceStore,
+}
+
```
ixrock, line 77: Maybe better to put `namespaceStore` inside `NamespaceSelectFilterProps`?
Or at least extend `NamespaceSelectFilterProps` with `NamespaceSelectFilterDependencies` because it's anyway  accessed via component's `props`, right?

Moreover you put it already in props interface above for placeholder:

```ts
interface NamespaceSelectFilterPlaceholderProps {
  namespaceStore: NamespaceStore
}
```

Wdyt?

src/renderer/components/+namespaces/namespace-select-filter.tsx
```diff
@@ -49,14 +61,22 @@ const Placeholder = observer((props: PlaceholderProps<any, boolean>) => {
...
 @observer
-export class NamespaceSelectFilter extends React.Component<SelectProps> {
+export class NonInjectedNamespaceSelectFilter extends React.Component<NamespaceSelectFilterProps & NamespaceSelectFilterDependencies> {
```
ixrock:
```suggestion
export class NamespaceSelectFilter extends React.Component<NamespaceSelectFilterProps> {
// ...

// export injected version as default
export default withInjectables(...);

// or export by explicit name
export const NamespaceSelectFilterInjected = withInjectables(...);

```

src/renderer/components/+user-management/+role-bindings/__tests__/dialog.test.tsx
```diff
@@ -39,6 +43,10 @@ describe("RoleBindingDialog tests", () => {
         uid: "1",
       },
     })];
+
+    const di = getDiForUnitTesting();
+
+    render = renderFor(di);
```
ixrock, line 49: Is it related to PR's somehow?

jakolehm: Only related because tests will fail without this (because NamespaceSelect now uses DI).

src/renderer/components/+namespaces/namespace-select-filter.tsx
aleksfront: It's better to stick with `data-testid` since we use it a lot more often. (102 search results vs 6 `data-test-id`).

src/renderer/components/+namespaces/namespace-select-filter.tsx
Nokel81: No, this is the correct usage so that dependencies are not exported to the extension API or the documentation.

src/renderer/components/+namespaces/namespace-select-filter.tsx
Nokel81: This currently follows the style that we have used elsewhere for injectable components.

Personally I like spitting them apart like this as it means that there is both a smaller diff and we can talk about the injected vs the injectable version of a component by name.

src/renderer/components/+namespaces/namespace-select-filter.tsx
```diff
@@ -65,69 +85,90 @@ export class NamespaceSelectFilter extends React.Component<SelectProps> {
...
       }
     } else {
-      namespaceStore.selectAll();
+      this.namespaceStore.selectAll();
     }
   };
 
+  onInputChange = (value: string, meta: InputActionMeta) => {
+    if (meta.action === "input-change") this.inputValue = value;
+    if (meta.action === "menu-close") this.inputValue = "";
```
Nokel81: Would be better if we either used a `switch` or use `else if`.

Either way, I think from a readability standpoint using brackets here would be better.

src/renderer/components/+namespaces/namespace-select.tsx
```diff
@@ -27,33 +27,53 @@ import { observer } from "mobx-react";
...
 
-const defaultProps: Partial<Props> = {
+interface NamespaceSelectDependencies {
+  namespaceStore?: NamespaceStore
+}
+
+const defaultProps: Partial<NamespaceSelectProps> = {
   showIcons: true,
+  maxItems: 500,
 };
```
Nokel81, line 49: I think that default props can probably be done via DI too, right?

jim-docker (commented):
Are we still expecting this with multi-select:

https://user-images.githubusercontent.com/40840436/148090526-2415d581-9f33-42fe-85eb-86a99c254989.mov

I pressed ctrl-click which seems to uncheck the one picked (even though it appears to be unchecked to begin with in the dropdown). It blocks for a good while but eventually recovers. Though connection failures seem to ensue.  (I think this is our current behaviour when trying to watch many namespaces, right?)

When the user starts typing and the dropdown consists of matching results, can the previous set of selected namespaces be cleared, or ignored in the context of this search? So that if I do a click or ctrl-clicks each one clicked on is added to the (initially) cleared set. If none are selected in the search dropdown after the dropdown is finally closed then restore the previous set of selected namespaces?

jim-docker:
The "Search Namespaces..." text seems to be sticky. After opening the dropdown it doesn't go away upon closing it.

https://user-images.githubusercontent.com/40840436/148097200-6be166f4-2d69-4df2-a8ec-bc6c7408c022.mov

Nokel81:
Closing because this PR is very old

PR #4557, opened 2021-12-10
Remove all 'NonBreakingSpace' characters from the repo

Nokel81:
Signed-off-by: Sebastian Malton <sebastian@malton.name>

These are part of the latest vscode version's display rules for unexpected charcters.

PR #4577, opened 2021-12-15
Add eslint:recommended lint rule

Nokel81:
~Blocked by #4576 and #4557~

### DIFF
--- a/src/common/search-store.ts
+++ b/src/common/search-store.ts
@@ -29,7 +29,7 @@ export class SearchStore {
    * @param value Unescaped string
    */
   public static escapeRegex(value?: string): string {
-    return value ? value.replace(/[\-\[\]{}()*+?.,\\\^$|#\s]/g, "\\$&") : "";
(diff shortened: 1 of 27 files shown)


src/common/search-store.ts
```diff
@@ -29,7 +29,7 @@ export class SearchStore {
    * @param value Unescaped string
    */
   public static escapeRegex(value?: string): string {
-    return value ? value.replace(/[\-\[\]{}()*+?.,\\\^$|#\s]/g, "\\$&") : "";
+    return value ? value.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, "\\$&") : "";
```
nevalla, line 32: This looks wrong to me. Should `[` be escaped?
`[-\[\]{}()*+?.,\\^$|#\s]/g`

Nokel81: No it doesn't need to be escaped because you cannot start another `one of` block within a `one of` block. This is the https://eslint.org/docs/rules/no-useless-escape rule

jim-docker (approved):

nevalla (approved):

PR #4579, opened 2021-12-16
Add ability from extension to add preferences to application preferences

jweak:

needed for this:
image

### DIFF
--- a/src/renderer/components/+preferences/application.tsx
+++ b/src/renderer/components/+preferences/application.tsx
@@ -30,6 +30,8 @@ import { isWindows } from "../../../common/vars";
 import { FormSwitch, Switcher } from "../switch";
 import moment from "moment-timezone";
 import { CONSTANTS } from "../../../common/user-store/preferences-helpers";
+import { AppPreferenceRegistry } from "../../../extensions/registries";
+import { ExtensionSettings } from "./extension-settings";
 
 const timezoneOptions: SelectOption<string>[] = moment.tz.names().map(zone => ({
   label: zone,
@@ -50,6 +52,7 @@ export const Application = observer(() => {
     );
 
   const [shell, setShell] = React.useState(UserStore.getInstance().shell || "");
+  const extensionSettings = AppPreferenceRegistry.getInstance().getItems().filter((preference) => preference.showInPreferencesTab === "application");
 
   return (
     <section id="application">
@@ -108,6 +111,10 @@ export const Application = observer(() => {
       </section>
 
       <hr />
+
+      {extensionSettings.map(setting => (
+        <ExtensionSettings key={setting.id} setting={setting} size="normal" />
+      ))}
 
       <section id="update-channel">
         <SubTitle title="Update Channel"/>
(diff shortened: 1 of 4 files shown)


src/renderer/components/+preferences/application.tsx
```diff
@@ -50,6 +52,7 @@ export const Application = observer(() => {
     );
 
   const [shell, setShell] = React.useState(UserStore.getInstance().shell || "");
+  const extensionSettings = AppPreferenceRegistry.getInstance().getItems().filter((preference) => preference.showInPreferencesTab === "application");
```
Nokel81, line 55: Maybe we could just move all preference pages to this style?

jweak: do you mean add settings from extensions to other preference pages too?

Nokel81: Yes, I don't quite see why not.

src/renderer/components/+preferences/extension-settings.tsx
```diff
@@ -0,0 +1,49 @@
...
+export function ExtensionSettings({ setting, size }: ExtensionSettingsProps) {
+  const {
+    title,
+    id,
+    components: { Hint, Input },
+  } = setting;
+
+  return (
+    <React.Fragment>
+      <section id={id} className={size === "small" ? "small" : ""}>
```
ixrock: _nit:_ 
```suggestion
      <section id={id} className={cssNames(size)}>
```

ixrock (dismissed):

src/renderer/components/+preferences/application.tsx
jim-docker: I don't know. This is meant for bundled extensions? This PR puts application extensions  after the "other" section and before "update-channel", to suit the mockup. Where do we place extension preferences on other pages? It would be better if the placements were configurable, but that's another ballgame. Otherwise implement support on other pages if the need arises?

Nokel81: We don't even have support for configuring the order of preferences on the extensions page so I don't think we have to solve that problem too.

Though if we were to do that then we probably should just follow the lead of all other orderings within Lens and use "priority"

jim-docker: For sure, I meant that I don't see the point of adding extension settings to the other pages because they don't have such ordering specifications, and without that it is all ad hoc. Add them if and when we need them to match some other mockup.

jim-docker (dismissed):
LGTM (satisfies the mockup)

src/renderer/components/+preferences/application.tsx
jweak: I agree with @jim-docker. If there will be more of these requirements then we'd like to have some ordering in place. Maybe follow YAGNI for now?

src/renderer/components/+preferences/extension-settings.tsx
jweak: good point 👍 done

ixrock (approved):

src/renderer/components/+preferences/application.tsx
Nokel81: Okay sure, I don't like this ad-hoc style of only extending parts of the extension API, but we discuss this again in the future.

Nokel81 (approved):

PR #4584, opened 2021-12-16
Change isEnabledForCluster to use KubernetesCluster instead of Cluster

### LINKED ISSUE #3790: Getting entity from cluster is unreliable in ExtensionRenderer.isEnabledForCluster()
**Describe the bug**

The following code, which makes a custom cluster page either visible or not given the active cluster, is unreliable:

```javascript
  public async isEnabledForCluster(cluster): Promise<boolean> {
    const entity = Renderer.Catalog.catalogEntities.getById(cluster.id);
    return entity?.metadata.source === "my-source";
  }
```

As I understand it, when this method is called, there should never be a case where `getById(id)` returns `undefined` because the associated entity isn't found in the Catalog.

The result is my custom page shows-up on one of the clusters added by my extension, while it doesn't on another one. It should show-up on both.

**To Reproduce**
1. Install https://github.com/Mirantis/lens-extension-cc from the `cluster-page` branch (if still available, or from `master` if not, since that would mean it was merged).
2. Add 2 clusters via the extension.
3. Open one cluster, the "Container Cloud" page might be there. Open another cluster, it might not be there.

**Expected behavior**
This test should never fail to find the associated Catalog entity.

**Screenshots**

Not there:
image

There:
image

- Lens Version: 5.2 (official)
- OS: macOS 11.5.2
- Installation method: .dmg

**Logs:**
```
There's too much sensitive stuff like usernames, emails, etc., in here to post these logs into a public repo.
```

Nokel81:

fixes #3790 

This is technically a breaking change but I think we should accept that it can be done over a minor version as this really should have happened at 5.0.0

Also I agree with the comments on the linked issue that it should be finer grained than this, but that is not for this PR.

### DIFF
--- a/src/extensions/lens-renderer-extension.ts
+++ b/src/extensions/lens-renderer-extension.ts
@@ -20,13 +20,13 @@
  */
 
 import type * as registries from "./registries";
-import type { Cluster } from "../main/cluster";
 import { Disposers, LensExtension } from "./lens-extension";
 import { getExtensionPageUrl } from "./registries/page-registry";
 import type { CatalogEntity } from "../common/catalog";
 import type { Disposer } from "../common/utils";
 import { catalogEntityRegistry, EntityFilter } from "../renderer/api/catalog-entity-registry";
 import { catalogCategoryRegistry, CategoryFilter } from "../renderer/api/catalog-category-registry";
+import type { KubernetesCluster } from "../common/catalog-entities";
 
 export class LensRendererExtension extends LensExtension {
   globalPages: registries.PageRegistration[] = [];
@@ -57,9 +57,12 @@ export class LensRendererExtension extends LensExtension {
   }
 
   /**
-   * Defines if extension is enabled for a given cluster. Defaults to `true`.
+   * Defines if extension is enabled for a given cluster. This method is only
+   * called when the extension is created within a cluster frame.
+   *
+   * The default implementation is to return `true`
    */
-  async isEnabledForCluster(cluster: Cluster): Promise<Boolean> {
+  async isEnabledForCluster(cluster: KubernetesCluster): Promise<Boolean> {
     return (void cluster) || true;
   }
 
--- a/src/renderer/cluster-frame.tsx
+++ b/src/renderer/cluster-frame.tsx
@@ -19,7 +19,7 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 import React from "react";
-import { observable, makeObservable } from "mobx";
+import { observable, makeObservable, when } from "mobx";
 import { disposeOnUnmount, observer } from "mobx-react";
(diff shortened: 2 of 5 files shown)


jakolehm (dismissed):

src/renderer/cluster-frame.tsx
```diff
@@ -101,7 +102,11 @@ export class ClusterFrame extends React.Component {
 
     catalogEntityRegistry.activeEntity = ClusterFrame.clusterId;
 
-    extensionLoader.loadOnClusterRenderer();
+    // Only load the extensions once the catalog has been populated
+    when(
+      () => Boolean(catalogEntityRegistry.activeEntity),
```
jim-docker, line 107: What's the ramification if this doesn't become true, due to a bug, say? Should there be a log after some timeout?

src/extensions/lens-renderer-extension.ts
```diff
@@ -57,9 +57,12 @@ export class LensRendererExtension extends LensExtension {
...
 
   /**
-   * Defines if extension is enabled for a given cluster. Defaults to `true`.
+   * Defines if extension is enabled for a given cluster. This method is only
+   * called when the extension is created within a cluster frame.
+   *
+   * The default implementation is to return `true`
    */
-  async isEnabledForCluster(cluster: Cluster): Promise<Boolean> {
+  async isEnabledForCluster(cluster: KubernetesCluster): Promise<Boolean> {
```
jim-docker, line 65: Is this likely going to break some existing extensions? Can you overload (keep both the old and new) to avoid breaking?
https://stackoverflow.com/a/31815015
I guess the extension developer still has to override this version for it to work.

src/renderer/cluster-frame.tsx
Nokel81: The ramification is that the extensions do not load. Sure I'll add a notification

src/extensions/lens-renderer-extension.ts
Nokel81: The point is to not expose the `Cluster` type to the extension API. So we just have to accept this breakage. 

I have checked all the extensions listed on https://github.com/lensapp/lens-extensions and none of them use this feature so I think it is probably safe to do so.

jim-docker (approved):

jakolehm (approved):

stefcameron:
> Also I agree with the comments on the linked issue that it should be finer grained than this, but that is not for this PR.

Has a new issue been created with my suggestion? I appreciate the fix to the symptom for now, but really, the existing API should be deprecated and retired, replaced with the finer grain option. It just doesn't make sense that it's an "all or nothing" thing to have or not to have custom cluster pages. An extension should be able to choose which page is displayed based on criteria it defines.

Nokel81:
@stefcameron Would you mind creating it, I completely agree about that it should be finer grained.

stefcameron:
See https://github.com/lensapp/lens/issues/4591

PR #4604, opened 2021-12-22
Best of both worlds: sass + css-modules + tailwind

### LINKED ISSUE #3701: Refactor existing scss styles to use tailwind + css modules

Refactor all existing scss styles to css modules. If we find a way to use css modules with scss like `filename.module.scss`, let's use this notation. Otherwise, use clean css (`filename.module.css`).

**Why is this needed**:

Currently we have two different ways in our codebase:
- the old way (scss with flexbox)
- the new way (tailwind + css modules)

We should refactor old -> new to keep codebase clean.

ixrock:
- using `sass` with `css-modules` and `tailwindcss`
- renaming all `*.module.css` to `*.module.scss` (though `*.module.css` still allowed extension, see `mocks.d.ts`)
- updating `tailwindcss` to latest version (3.x)

P.S. Not sure is we need to close #3701 since it requests to convert all style files to modules.
Or we could close & create new issue like "convert all possible *.scss -> *.module.scss" (or just do it incrementally or even keep part of the old styling files untouched if doesn't hurt).

_Example:_

```scss
// typical style module file: component.module.scss
.Component {
   padding: var(--padding);
   @apply mt-2, some-other, tailwind-stuff; // use anything from tailwind (!)
   @include sass-mixin, etc; // keep using sass features (!)

  .other {
   color: red;
  }
}
```

_Usage inside react component:_
```tsx
import styles from "./component.module.scss"

render(){
 return (
  <div className={styles.Component}>
     Some text is <span className={styles.other}>important</span> 
   </div>
  )
}
```


package.json
```diff
@@ -361,8 +361,7 @@
     "nodemon": "^2.0.14",
     "playwright": "^1.17.1",
     "postcss": "^8.4.5",
-    "postcss-loader": "4.3.0",
-    "postinstall-postinstall": "^2.1.0",
```
ixrock, line 365: Unused dependency.

tailwind.config.js
```diff
@@ -20,8 +20,7 @@
  */
 
 module.exports = {
-  mode: "jit",
-  purge: ["src/**/*.tsx"],
+  content: ["src/**/*.tsx"],
```
ixrock, line 23: Renamed in `tailwindcss@3.x`

types/mocks.d.ts
```diff
@@ -29,6 +29,14 @@ declare const __static: string;
 
 // Support import for custom module extensions
 // https://www.typescriptlang.org/docs/handbook/modules.html#wildcard-module-declarations
+declare module "*.module.scss" {
+  const classes: { [key: string]: string };
+  export default classes;
+}
+declare module "*.module.css" {
```
ixrock, line 36: Normal css-modules should be allowed too, but let's stick to `%component-name%.module.scss`

aleksfront (approved):
LGTM

aleksfront:
Let's keep #3701 opened until all styles are converted to css modules.

jweak (approved):
LGTM

PR #4607, opened 2021-12-23
Keep port-forward objects around when not running

### LINKED ISSUE #4580: Port-forward disabled state
As of 5.3 port-forwards can be saved across cluster connect/disconnect and app restart. However, their existence is currently tied to being active. When a cluster is connected any saved port-forwards are started, but if they fail (can't access pod, or port is not available, etc) they are removed from the saved list (and lost). Also if Lens detects that a port-forward has stopped it is removed from the saved list.

Additionally supporting a disabled state means that port-forwards can be kept whether they successfully run or not. Also, they can be stopped and started later, not just removed.

jim-docker:
fixes #4580 

The initial port-forward implementation tied the existence of port-forward objects to the existence of the underlying `kubectl port-forward` process.

Port-forwards can now be stopped and started separate from their deletion and creation.

https://user-images.githubusercontent.com/40840436/147290695-d600c94c-140c-4b04-b6d9-3cc96adae100.mov

There is a notification on cluster connection if a port-forward fails to start.
   
https://user-images.githubusercontent.com/40840436/147290756-6363a161-4834-43ee-9623-ce7feca9346a.mov


src/renderer/components/+network-port-forwards/port-forward-menu.tsx
```diff
@@ -45,21 +45,56 @@ export class PortForwardMenu extends React.Component<Props> {
...
 
     if (!portForward) return null;
 
     return (
       <>
-        <MenuItem onClick={() => openPortForward(this.props.portForward)}>
-          <Icon material="open_in_browser" interactive={toolbar} tooltip="Open in browser" />
-          <span className="title">Open</span>
-        </MenuItem>
+        { Boolean(portForward.status === "Active") &&
```
ixrock: _nit_: no need in `Boolean()` here.

src/renderer/port-forward/port-forward-dialog.tsx
```diff
@@ -27,25 +27,27 @@ import { observer } from "mobx-react";
 import { Dialog, DialogProps } from "../components/dialog";
 import { Wizard, WizardStep } from "../components/wizard";
 import { Input } from "../components/input";
-import { Notifications } from "../components/notifications";
-import { cssNames } from "../utils";
+import { cssNames, noop } from "../utils";
 import { addPortForward, getPortForwards, modifyPortForward } from "./port-forward.store";
 import type { ForwardedPort } from "./port-forward-item";
-import { aboutPortForwarding, openPortForward } from ".";
+import { aboutPortForwarding, notifyErrorPortForwarding, openPortForward } from ".";
```
ixrock: _nit:_
```suggestion
import { aboutPortForwarding, notifyErrorPortForwarding, openPortForward } from "./index";
```

src/renderer/port-forward/port-forward-dialog.tsx
```diff
@@ -84,43 +87,51 @@ export class PortForwardDialog extends Component<Props> {
     this.desiredPort = this.currentPort;
   };
 
-  onClose = () => {
+  onClose = async () => {
+    await (async () => dialogState.onClose())();
```
ixrock: Looks weird. What it's for? Maybe better to try one of these if it's kinda timing issue or so:

```   ts
    window.setTimeout(callback)
    window.requestIdleCallback(callback)
    process.nextTick(callback)
```

src/renderer/port-forward/port-forward.store.ts
```diff
@@ -94,82 +104,248 @@ interface PortForwardResult {
...
 }
 
-export async function addPortForward(portForward: ForwardedPort): Promise<number> {
-  const { port, forwardPort } = portForward;
+function findPortForward(portForward: ForwardedPort) {
+  return portForwardStore.portForwards.find(portForwardsEqual(portForward));
+
+}
+
+const setPortForward = action((portForward: ForwardedPort) => {
```
ixrock, line 121: Is there a reason why those are not direct methods of the store class?

ixrock (dismissed):

aleksfront (dismissed):
LGTM

src/renderer/port-forward/port-forward-dialog.tsx
jim-docker: I wasn't sure how to do the `onClose`, so I duplicated how the `ok` was implemented for the confirm-dialog: https://github.com/lensapp/lens/blob/4f75acf2b40b3efdae3c5b49de443dfa08459db0/src/renderer/components/confirm-dialog/confirm-dialog.tsx#L91

I just wanted to make sure the port-forward UI would update to reflect the changes after the dialog closed (and not block anything). I overcomplicated the problem. I have simplified the implementation (next commit). It is working the same.

src/renderer/port-forward/port-forward.store.ts
jim-docker: I guess not anymore, good eye. I tried to keep the store isolated (somewhat) from the functions handling the ipc calls (`apiBase.post`, etc) because the store was just shadowing the `PortForward.portForwards` array managed in the main process. (you didn't need the store to create, or get, etc, port-forwards). But now the store is the "source of truth"

I could refactor in a followup PR.

aleksfront (approved):

ixrock (approved):

PR #4610, opened 2021-12-27
Native switch component

aleksfront:
Introducing our custom `Switch` component and removing one from material ui.

Extra reasons for removing material ui switch are complicated styling and need for wrapper components.

Part of #4138

https://user-images.githubusercontent.com/9607060/147453770-1d7033dc-ed4e-4ba1-b111-fbc0ac7aad6e.mov

disabled switch

hovered switch


jweak:
Hmm what about current extensions that might be using `<FormSwitch />`?

aleksfront:
> Hmm what about current extensions that might be using `<FormSwitch />`?

Marked `<FormSwitch/>` and `<Switcher/>` as deprecated. Let's remove them in 6.x.x I guess.

ixrock:
> > Hmm what about current extensions that might be using `<FormSwitch />`?
> 
> Marked `<FormSwitch/>` and `<Switcher/>` as deprecated. Let's remove them in 6.x.x I guess.

~~Maybe that extensions just need to add `@material-ui/*` stuff to their local dependencies?~~

Just realized that `FormSwitch` is our custom component.. 
Then marking as deprecation or even delete (breaking change ofc, but not critical if we provide to extensions easy to pick up replacement component with nice API:)

src/renderer/components/+preferences/application.tsx
```diff
@@ -135,16 +128,9 @@ export const Application = observer(() => {
...
-          control={
-            <Switcher
-              checked={userStore.openAtLogin}
-              onChange={v => userStore.openAtLogin = v.target.checked}
-              name="startup"
-            />
-          }
-          label="Automatically start Lens on login"
-        />
+        <Switch checked={userStore.openAtLogin} onChange={v => userStore.openAtLogin = v.target.checked}>
```
ixrock:
```suggestion
        <Switch checked={userStore.openAtLogin} onChange={(enabled: boolean, event: MouseEvent) => userStore.openAtLogin = v.target.checked}>
```

_DRY_: i think that component should receive actual boolean value as first argument.

src/renderer/components/switch/switch.module.scss
```diff
@@ -0,0 +1,121 @@
...
+  gap: 2ch;
+  justify-content: space-between;
+  cursor: pointer;
+  user-select: none;
+  color: var(--textColorAccent);
+  font-weight: 500;
+
+  & > input {
+    --thumb-position: 0%;
+    --thumb-transition-duration: .25s;
```
ixrock: Why these variables defined not at `.Switch` level?

src/renderer/components/switch/switch.module.scss
```diff
@@ -0,0 +1,121 @@
...
+
+    flex-shrink: 0;
+    display: grid;
+    align-items: center;
+    grid: [track] 1fr / [track] 1fr;
+
+    transition: background-color .25s ease;
+
+    &::before {
+      --highlight-size: 0;
```
ixrock: This too probably better pop up to upper level of the component (with more descriptive name).

src/renderer/components/switch/switch.tsx
```diff
@@ -0,0 +1,53 @@
...
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import styles from "./switch.module.scss";
+
+import React, { DetailedHTMLProps, InputHTMLAttributes } from "react";
+import { cssNames } from "../../utils";
+
+interface Props extends DetailedHTMLProps<InputHTMLAttributes<HTMLInputElement>, HTMLInputElement> {
```
ixrock: Why not just `React.HTMLProps`? Another option to use `React.DOMAttributes` if enough to (these interfaces mostly used in all other components in codebase)

src/renderer/components/switch/switch.tsx
```diff
@@ -0,0 +1,53 @@
...
+  onClick?: () => void;
+}
+
+export function Switch({ children, onClick, ...settings }: Props) {
```
ixrock:
```suggestion
export function Switch({ children, onClick, disabled, ...props }: Props) {
```

src/renderer/components/switch/switch.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const id = `switch-${Date.now()}`;
```
ixrock: Why do we need id here? You should be able to pass one in `props` when needed..
Moreover, generating id inside `render()` with random strings will cause re-rendering every time (should be static).

ixrock (commented):
Otherwise LGTM.

src/renderer/components/+preferences/application.tsx
aleksfront: Why it needs to be added? Also, this change would collide with native `onChange` types.
```
interface Props extends React.HTMLProps<HTMLInputElement> {
}
```

src/renderer/components/switch/switch.tsx
aleksfront: Thanks for the review. It leaded to much more simpler component.

src/renderer/components/+preferences/application.tsx
ixrock: Exclude native `onChange` from interface and define yours (see: `tabs.tsx` for example)
Or you can even skip extending from `React.HTMLProps` and keep only necessary props.

ixrock: _Reason:_ to avoid repeating boilerplate / taking value from event since in most cases we just need current state of the component (not mouse event)

aleksfront: I don't agree we should juggle with built-in properties. 
1. The Switch component is basically `<input type="checkbox"/>` and should have all same properties.
2. We shouldn't rely on value from `onChange` callback at all because we're using controlled component model. Since we're passing `checked` prop explicitly, we must preserve single source of truth and get value from stores, not from the `Switch`.

I've made changes replacing this one
```
<Switch checked={userStore.terminalCopyOnSelect} onChange={v => userStore.terminalCopyOnSelect = v.target.checked}>
  Copy on select and paste on right-click
</Switch>
```

with this one

```
<Switch checked={userStore.terminalCopyOnSelect} onChange={() => userStore.terminalCopyOnSelect = !userStore.terminalCopyOnSelect}>
  Copy on select and paste on right-click
</Switch>
```

ixrock: > Since we're passing checked prop explicitly, we must preserve single source of truth and get value from stores, not from the Switch.

Didn't get what you mean.. Why it's single source of truth in stores?? when you change something it comes from user event right? So you're getting new **source of truth** always _first_ from that user event. Wdyt?

> I don't agree we should juggle with built-in properties.

Just consider `<Switch/>` as custom component, not just  plain html`<NewCheckbox/>`, it's not juggling, it's a matter of convenient API for end-users (of this component, remember this is gonna be exported to extensions-api). 
And it's  actually simple to implement: pass in callback from `props.checked`, what's the issue with that?

aleksfront: Done with
```
interface Props extends Omit<HTMLProps<HTMLInputElement>, "onChange"> {
  onChange?: (checked: boolean, event: ChangeEvent<HTMLInputElement>) => void;
}
```

ixrock (dismissed):

src/renderer/components/+preferences/editor.tsx
```diff
@@ -45,15 +45,12 @@ export const Editor = observer(() => {
...
-                  checked={editorConfiguration.minimap.enabled}
-                  onChange={(evt, checked) => editorConfiguration.minimap.enabled = checked}
-                />
-              }
-            />
+            <Switch
+              checked={editorConfiguration.minimap.enabled}
+              onChange={() => editorConfiguration.minimap.enabled = !editorConfiguration.minimap.enabled}
+            >
+              Copy on select and paste on right-click
```
jweak: This should be `Show minimap`?

aleksfront: Omg, thanks for catching this!

jweak (approved):
Looks good!

ixrock (approved):

PR #4611, opened 2021-12-27
Fix close button overflow in Preferences

aleksfront:
In small window size, close button might disappear moving behind the viewport. Current PR:
- fixes close button position so it's always visible
- adjusting `Select` dropdowns to look similar along with other inputs

Before:

hiding close button
different inputs

After:
visible close button
similar inputs

Fixes https://github.com/lensapp/lenscloud-lens-extension/issues/986

### DIFF
--- a/src/renderer/components/layout/setting-layout.tsx
+++ b/src/renderer/components/layout/setting-layout.tsx
@@ -104,11 +104,19 @@ export class SettingLayout extends React.Component<SettingLayoutProps> {
           <div className="toolsRegion">
             {
               this.props.provideBackButtonNavigation && (
-                <div className="fixedTools">
-                  <div className="closeBtn" role="button" aria-label="Close" onClick={back}>
-                    <Icon material="close" />
+                <div className="fixed top-[60px]">
+                  <div
+                    className="w-[35px] h-[35px] grid place-items-center pointer border-2 rounded-full border-solid border-textDimmed hover:bg-[#72767d25] active:translate-y-px"
+                    role="button"
+                    aria-label="Close"
+                    onClick={back}
+                  >
+                    <Icon material="close" className="text-textAccent opacity-60" />
                   </div>
-                  <div className="esc" aria-hidden="true">
+                  <div
+                    className="text-center mt-3 font-bold text-lg select-none text-textDimmed pointer-events-none"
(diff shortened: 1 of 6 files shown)


src/renderer/components/layout/setting-layout.tsx
```diff
@@ -104,11 +104,19 @@ export class SettingLayout extends React.Component<SettingLayoutProps> {
           <div className="toolsRegion">
             {
               this.props.provideBackButtonNavigation && (
-                <div className="fixedTools">
-                  <div className="closeBtn" role="button" aria-label="Close" onClick={back}>
-                    <Icon material="close" />
+                <div className="fixed top-[60px]">
+                  <div
+                    className="w-[35px] h-[35px] grid place-items-center pointer border-2 rounded-full border-solid border-textDimmed hover:bg-[#72767d25] active:translate-y-px"
```
ixrock: I would better to move all specific or custom `tailwindcss` classes to style file with `@apply` instruction.

_Reason_: if we decide to get rid of tailwind in the future it would be much easier update css-files instead of cleaning it up from react-components.

tailwind.config.js
```diff
@@ -26,7 +26,14 @@ module.exports = {
     fontFamily: {
       sans: ["Roboto", "Helvetica", "Arial", "sans-serif"],
     },
-    extend: {},
+    extend: {
+      colors: {
+        textAccent: "var(--textColorAccent)",
+        textPrimary: "var(--textColorPrimary)",
+        textTertiary: "var(--textColorTertiary)",
+        textDimmed: "var(--textColorDimmed)",
```
ixrock, line 34: Do we still need these? As we consider using plain _tailwind_ class names more as _anti-pattern_ inside react components (when could be easily replaced with plain css alternative inside module style file). 

So using `.item {color: var(--textColorAccent);}` more preferable than using `tailwind` class `text-textAccent`. Wdyt?

_Downsides of using classnames inside components:_
- classnames still **global** (extensions might not know about them == collisions/bug)
- closer binding with `tailwind` / harder to detach in future if not needed anymore
- less api is better: harder to understand in comparing with plain css rules

ixrock (approved):

jweak (approved):

PR #4612, opened 2021-12-27
Adapt to recent simplifications in injectable

jansav:
# Simplifications:

1. Type parameters are inferred instead of explicit by using `getInjectable`.
2. Boilerplate function `getDependencies` is removed. There is only `instantiate`.
3. `instantiate` can be async.
4. `Dependencies` -type does not need to be exported.


aleksfront (approved):
Less boilerplate is better, LGTM

Iku-turso (approved):

PR #3653, opened 2021-08-20
Fix prometheus operator support

marcbachmann:
Not sure why this didn't come up already, but all the group_left queries returned an empty result set because the group_left function was used wrong.

What do you think about the changes? Did i miss something?

PR #4617, opened 2021-12-28
Fix prometheus operator metrics to work out of the box

### LINKED ISSUE #4435: No metrics with 5.3.0-latest.2021125.2
**Describe the bug**
After installing the 5.3.0-latest.2021125.2 version (both on Windows 10 and Rocky Linux 8.5) the metrics for GKE based cluster are not available.

**To Reproduce**
1. Install 5.2.7 - the metrics are available
2. Install 5.3.0 - the metrics are not available
3. Uninstall 5.3.0 and install 5.2.7 again - metrics are available again

**Expected behavior**
The metric should be availabe

**Screenshots**
image

image

- Lens Version: 5.3.0-latest.2021125.2
- OS: Windows 10 and Rocky Linux 8.5 (RHEL 8.5)
- Installation method (e.g. snap or AppImage in Linux): Windows - exe; Rocky - rpm
- GKE base cluster
- Lens metrics configuration: tried Auto detect (works on 5.2.7) and Prometheus Operator (works on 5.2.7)

**Kubeconfig:**
5.2.7 works with the exact same Kubeconfig, so the latter is not an issue

**Additional context**

nevalla:
This PR brings back `group_left` joins removed by https://github.com/lensapp/lens/pull/3653. The reason why joins are not working is that `prometheus-community/kube-prometheus-stack` helm chart has invalid `kube-state-metrics` serviceMonitor custom resource and `kube_pod_info` metrics is not available (https://github.com/prometheus-community/helm-charts/issues/1631).

Metrics on Lens IDE should work without making any additional relabelings and that's why bringing these group left joins back.

Fixes: #4435

### DIFF
--- a/src/main/prometheus/operator.ts
+++ b/src/main/prometheus/operator.ts
@@ -38,7 +38,7 @@ export class PrometheusOperator extends PrometheusProvider {
       case "cluster":
         switch (queryName) {
           case "memoryUsage":
-            return `sum(node_memory_MemTotal_bytes - (node_memory_MemFree_bytes + node_memory_Buffers_bytes + node_memory_Cached_bytes))`.replace(/_bytes/g, `_bytes{node=~"${opts.nodes}"}`);
+            return `sum(node_memory_MemTotal_bytes - (node_memory_MemFree_bytes + node_memory_Buffers_bytes + node_memory_Cached_bytes))`.replace(/_bytes/g, `_bytes * on (pod,namespace) group_left(node) kube_pod_info{node=~"${opts.nodes}"}`);
           case "workloadMemoryUsage":
             return `sum(container_memory_working_set_bytes{container!="", instance=~"${opts.nodes}"}) by (component)`;
           case "memoryRequests":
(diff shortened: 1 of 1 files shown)


jweak (approved):

aleksfront (approved):

src/main/prometheus/operator.ts
```diff
@@ -50,7 +50,7 @@ export class PrometheusOperator extends PrometheusProvider {
           case "memoryAllocatableCapacity":
             return `sum(kube_node_status_allocatable{node=~"${opts.nodes}", resource="memory"})`;
           case "cpuUsage":
-            return `sum(rate(node_cpu_seconds_total{node=~"${opts.nodes}", mode=~"user|system"}[${this.rateAccuracy}]))`;
+            return `sum(rate(node_cpu_seconds_total{mode=~"user|system"}[${this.rateAccuracy}])* on (pod,namespace) group_left(node) kube_pod_info{node=~"${opts.nodes}"})`;
```
marcbachmann, line 53: The queries are now wrong again 😉 
It should be reversed, join on `node`, take the specific labels
```
on (node) group_left(pod, namespace)
```

Or does the query work for you?

Nokel81: It does not work for us. Many reports about the queries you provided as failing that is why we reverted them.

marcbachmann: I agree that they need to be reintroduced to fix the compatibilty, but they need to get applied correctly to get it to work.

marcbachmann: ahhh, so `node_cpu_seconds_total` should contain `pod` and `namespace`. Maybe a comment about that would be useful. The node exporter in my setup doesn't have those labels. That also belongs into the documentation.

Nokel81: Ah thanks for the info, would you mind making an issue to add that to the documentation?

nevalla: @marcbachmann how did you install prometheus operator?

marcbachmann: I have an external node exporter running natively on the hosts.
Therefore it only has the `node` label. I'm doing this because I have still some other services like databases outside kubernetes.

Might be an edge case and I could migrate that into the cluster.

PR #4614, opened 2021-12-28
Remove global shared state usages of ExtensionDiscovery

jansav:
This PR will:
1. Replace all usages on `ExtensionDiscovery` with `di.inject(extensionDiscoveryInjectable)`
2. Re-introduce some skipped test
3. Introduce a way to allow transitional refactoring of functions that are exposed in Extension API without breaking the contract: `asLegacyGlobalFunctionForExtensionApi`
4. Extract minimal dependencies from a god class in couple use cases

Issue #4382, opened 2021-11-18
Add automatic/sane way to update extensions


ixrock:
Currently there is no `Update` option in the extension's menu:

<img width="758" alt="Screenshot 2021-11-18 at 16 54 20" src="https://user-images.githubusercontent.com/6377066/142442528-33953d76-06ba-4bdb-9cfc-2c82e04d6b62.png">

**Why is this needed**:

Manual downloading & copy-pasting files from packages is bad UX, especially when you have a lot of extensions.

Nokel81:
Related https://github.com/lensapp/lens/issues/3579

Nokel81:
FYI @nevalla I think has mostly implemented this (at least for NPM installed packages) I think

msa0311:
This is something that is as well important for bundled extensions. I added some higher priority to this issue: https://github.com/lensapp/lens/issues/3579

Maybe this can be closed as duplicate? I think auto update is the way to go.

Nokel81:
@msa0311 Do you mean forced auto update? Also the linked issue is about `npm` installed extensions. Bundled doesn't fall under that category.

PR #4618, opened 2021-12-29
Allow to update extensions

nevalla:
image


ixrock:
Related https://github.com/lensapp/lens/issues/4382

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -249,6 +251,79 @@ export class ExtensionLoader {
...
+      }
+    }
+
+    return availableUpdates;
+  }
+
+  protected async getLatestVersionFromNpmJs(extension: InstalledExtension) {
+    const name = extension.manifest.name;
+    const registryUrl = new URLParse("https://registry.npmjs.com").set("pathname", name).toString();
+    const { promise } = downloadJson({ url: registryUrl });
```
jansav: `downloadJson` looks like a dependency to me. Minimum, could you pass it as a constructor -argument so that it can be overridden easier in future.

src/renderer/components/+extensions/attempt-install/create-temp-files-and-validate/create-temp-files-and-validate.tsx
```diff
@@ -60,6 +61,11 @@ export async function createTempFilesAndValidate({
 
     await fse.writeFile(tempFile, data);
     const manifest = await validatePackage(tempFile);
+
+    if (!isCompatibleExtension(manifest)){
+      throw new Error("Incompatible extension");
```
jansav: `throw` as a mechanism here is not the best. Makes code lot hard to read. Extract creation of notification for an error to a function and use it here?


 Scope of `try-catch` should be reduced.

src/renderer/components/+extensions/installed-extensions.tsx
```diff
@@ -47,7 +48,7 @@ function getStatus(extension: InstalledExtension) {
   return extension.isEnabled ? "Enabled" : "Disabled";
 }
 
-export const InstalledExtensions = observer(({ extensions, uninstall, enable, disable }: Props) => {
+export const InstalledExtensions = observer(({ extensions, uninstall, enable, disable, upgrade }: Props) => {
```
jansav: It seems to be that `extensions` here should be a dependency that contains objects having `menuItems` as a **computed** value. 

Then we could just write:

```typescript
extension.menuItems.map(({ disabled, onClick }) => (
  <MenuItem disabled={disabled} onClick={onClick} />
));
```

## Benefits
1. Creation of new menu item (e.g. upgrade in this PR) is easy (only one place needs to change; Open Closed Principle)
2. Business logic (e.g. which menu item should be shown, what it does etc.) is separated from UI code (Separation of Concerns)

After #4614 is merged, `InstalledExtensions` is turned to `injectable` which then can control the dependencies (e.g. map `extensions` prop to the required interface)

Maybe we could do this together as pair/mob programming? Ask any time.

src/main/index.ts
```diff
@@ -281,6 +281,7 @@ app.on("ready", async () => {
       });
 
     extensionLoader.initExtensions(extensions);
+    extensionLoader.getAvailableExtensionUpdates();
```
jansav: I'm not really sure if this is good strategy for checking the updates. 

I would prefer if we created `checkForUpdate` method which takes `extension` as an argument. Then we could control each extension separately, and do the check right after initialization (if that's the requirement)?

jansav (changes requested):

src/main/index.ts
nevalla: I will move this logic into `LensExtension` class so extension developers can extend the logic with their custom needs

jansav: Hm, is that good thing? Do we want to expose this to Extension API?

nevalla: If someone creates extension for organization internal use, they can override update check and provide an update, for example, from their own servers.

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -58,13 +62,20 @@ export class ExtensionLoader {
...
+
   @observable isLoaded = false;
+  private extensionUpdateChecker: LensExtensionUpdateChecker;
 
   get whenLoaded() {
     return when(() => this.isLoaded);
   }
 
   constructor(protected dependencies : Dependencies) {
+    this.extensionUpdateChecker = new LensExtensionUpdateChecker(this.extensionUpdateSources);
```
aleksfront, line 78: `LensExtensionUpdateChecker` is a dependency here. Can we pass it in a constructor?

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -320,9 +345,11 @@ export class ExtensionLoader {
               return null;
             }
 
+            // const instance = new LensExtensionClass(extension, this.extensionUpdateChecker);
```
aleksfront, line 348: A leftover comment.

src/extensions/github-latest-version-checker.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  protected async getJson(url: string) {
+    const headers = {
+      "user-agent": "Lens IDE",
+    };
+
+    const { promise } = this.downloadJson({ url, headers });
+    const json = await promise.catch(() => {
+      // do nothing
```
aleksfront, line 47: We should log some error in this case with `logger.error`. This will help with troubleshooting.

src/extensions/npmjs-latest-version.checker.ts
```diff
@@ -0,0 +1,55 @@
+/**
```
aleksfront, line 1: Which file should we keep and which must be deleted?

There is `npmjs-latest-version-checker.ts` as well as `npmjs-latest-version.checker.ts`.

src/extensions/extension-loader/extension-loader.ts
Nokel81: That we should

Nokel81:
Clearing from milestone 5.5 as it is not a blocker.

Nokel81:
Converting to draft

Nokel81:

PR #4619, opened 2021-12-29
Extension tray menu items

jweak:
Added API for extension to add menu items to tray menu.

https://user-images.githubusercontent.com/774344/147666598-cb848465-8c55-48de-bccf-b7037b043909.mov


src/main/tray/tray.ts
```diff
@@ -62,7 +67,7 @@ export function initTray(windowManager: WindowManager) {
   const disposers = [
     autorun(() => {
       try {
-        const menu = createTrayMenu(windowManager);
+        const menu = createTrayMenu(windowManager, toJS(trayMenuItems.get()));
```
ixrock, line 70: Why do we need `toJS()` here? As `trayMenuItems` is observable box afaik and calling `.get()` already unwraps all observability == getting plain js/json? Wdyt?

src/main/tray/tray.ts
```diff
@@ -108,6 +116,8 @@ function createTrayMenu(windowManager: WindowManager): Menu {
     });
   }
 
+  template = template.concat(extensionTrayItems);
```
ixrock: _nit:_
```suggestion
  template = [template, ...extensionTrayItems];
```

ixrock (dismissed):
LGTM

src/main/tray/tray.ts
jweak: I created an extension that had `trayMenus` as an observable array and added a menu after the extension was loaded to test if the extension can change its menus and this `autorun` then went into infinite loop after the `trayMenus` changed. This fixed it. Not sure why, I'm not a `mobx` pro but maybe something to do with nested arrays?

src/main/menu/electron-menu-items.injectable.ts
```diff
@@ -29,7 +29,7 @@ const electronMenuItemsInjectable = getInjectable({
     const extensions = di.inject(mainExtensionsInjectable);
 
     return computed(() =>
-      extensions.get().flatMap((extension) => extension.appMenus),
+      extensions.get().flatMap((extension) => extension.appMenus.slice()),
```
jansav: What is the reasoning behind `.slice()`?

jweak: It was for supporting observable arrays but it's not actually needed. good catch!

jansav (approved):
Looks good. Love the amount of new tests.

ixrock (dismissed):

src/extensions/lens-main-extension.ts
```diff
@@ -25,9 +25,11 @@ import { catalogEntityRegistry } from "../main/catalog";
 import type { CatalogEntity } from "../common/catalog";
 import type { IObservableArray } from "mobx";
 import type { MenuRegistration } from "../main/menu/menu-registration";
+import type { MenuItemConstructorOptions } from "electron";
```
jansav: I would prefer if we wouldn't expose interfaces (or anything..) from 3rd party libraries in extension API. This limits our ability to adapt to future needs. 

3rd party libraries should always be isolated behind abstractions. 

In this case, we should create own interface which exposes only the features that we want to expose and then in `createTrayMenu` we should map it to the interface that electron requires. That gives us flexibility to do what ever we want in future, even if we choose to implement this menu in some other way.

jansav (changes requested):

src/extensions/lens-main-extension.ts
jweak: Yeah I thought about. Fair enough. Added a type `TrayMenuRegistration` for that.

aleksfront (dismissed):
Tested on both linux and win, works good!

ixrock (dismissed):

docs/extensions/guides/main-extension.md
```diff
@@ -96,6 +95,23 @@ export default class SamplePageMainExtension extends Main.LensExtension {
 When the menu item is clicked the `navigate()` method looks for and displays a global page with id `"myGlobalPage"`.
 This page would be defined in your extension's `Renderer.LensExtension` implementation (See [`Renderer.LensExtension`](renderer-extension.md)).
 
+### `trayMenus`
+
+`trayMenus` is a subset of Electron's [`MenuItemConstructorOptions`](https://www.electronjs.org/docs/v14-x-y/api/menu-item). Most importantly you can define a `label` and a `click` handler. Other properties are `submenu`, `enabled`, `visible`, `toolTip`, `id` and `type`.
```
jansav: This is outdated

jansav (changes requested):

jansav (dismissed):

chenhunghan (approved):
LGTM

jansav (approved):

PR #4625, opened 2022-01-03
Upgrade to Electron 14.2.4

jakolehm:
- updates Electron to 14.2.1 (https://www.electronjs.org/releases/stable?version=14)
- removes `@electron/remote` dependency:
  > Warning! This module has many subtle pitfalls. There is almost always a better way to accomplish your task than using this module. For example, ipcRenderer.invoke can serve many common use cases.


src/common/ipc/ipc.ts
```diff
@@ -48,51 +45,52 @@ function getSubFrames(): ClusterFrameInfo[] {
...
+  args = args.map(sanitizePayload);
+
+  ipcMain?.emit(channel, ...args);
+
+  for (const view of views) {
+    let viewType = "unknown";
+
+    // There will be a uncaught exception if the view is destroyed.
+    try {
+      viewType = view.getType();
```
ixrock, line 53: Maybe better to check `view.isDestroyed()` instead of try-catch or it might catch something else?

src/renderer/components/layout/top-bar/top-bar.tsx
```diff
@@ -84,19 +81,15 @@ const NonInjectedTopBar = (({ items, children, ...rest }: Props & Dependencies)
   };
 
   const minimizeWindow = () => {
-    window.minimize();
+    requestMain(IpcMainWindowEvents.WINDOW_ACTION, "minimize");
```
ixrock: Since `requestMain(channel: string, ...args: any[])` doesn't do type-checking for view type (e.g. "minimize", "window", etc.) + in order to reduce calling `requestMain` directly we could wrap all these things into "actions" with fully type checking for arguments, e.g.:

```ts
// common/ipc/window.actions.ts

export function windowAction(type: 'backgroundPage' | 'window' | 'browserView' | 'remote' | 'webview' | 'offscreen' = "window") {
  return requestMain(IpcMainWindowEvents.WINDOW_ACTION, type);
}

```

Wdyt?

ixrock (dismissed):
LGTM

src/common/ipc/ipc.ts
```diff
@@ -48,51 +45,52 @@ function getSubFrames(): ClusterFrameInfo[] {
...
+    } catch {
+      // We can ignore the view destroyed exception as viewType is only used for logging.
+    }
+
+    // Send message to views.
+    try {
+      logger.silly(`[IPC]: broadcasting "${channel}" to ${viewType}=${view.id}`, { args });
+      view.send(channel, ...args);
+    } catch (error) {
+      logger.error(`[IPC]: failed to send IPC message "${channel}" to view "${viewType}=${view.id}"`, { error: String(error) });
```
Nokel81: I don't think we should do `{ error: String(error) }` pattern, I would say it is better to just do `, error` as then you also get the stack track (if there is one)

jakolehm: I didn't actually do any changes here... but your suggestion makes sense 👍

src/renderer/components/layout/top-bar/top-bar.tsx
jakolehm: Something like that makes sense, I really don't like how ipc stuff lacks type checks.

src/common/ipc/ipc.ts
jakolehm: Didn't change this logic, not sure if I want to change it in this PR 😅

.yarnrc
```diff
@@ -1,3 +1,3 @@
 disturl "https://atom.io/download/electron"
-target "13.6.1"
+target "14.2.1"
 runtime "electron"
```
Nokel81: I think we can actually remove this file as none of these three fields are anywhere to be seen in https://yarnpkg.com/configuration/yarnrc/ or https://classic.yarnpkg.com/lang/en/docs/yarnrc/

jakolehm: I think we cannot, native node modules (like node-pty) won't be properly built if we remove this. 

See: 
- https://www.electronjs.org/docs/latest/tutorial/using-native-node-modules
- https://github.com/microsoft/vscode/blob/main/.yarnrc

Nokel81: Okay, I don't see how that tutorial says that such a file is required but I do acknowledge that vscode has it. So if we don't actually need it then that should be done in a separate PR.

Nokel81 (dismissed):

Nokel81:
@jakolehm Please fix lint and then I say this is RTM!

jakolehm:
> @jakolehm Please fix lint and then I say this is RTM!

Fixed!

Nokel81 (dismissed):

ixrock (dismissed):

Nokel81 (dismissed):

jim-docker:
On linux and windows, after the first time opening the app menus, it takes two clicks to open them:

https://user-images.githubusercontent.com/40840436/151072624-ea13f856-85f2-4cf3-b075-2b06b883b1ca.mov

works fine on master

@Nokel81 This was introduced by the last two commits (https://github.com/lensapp/lens/pull/4625/commits/bb905bc0e56f484928f998c72c0a0c2bba0e2ca7 and https://github.com/lensapp/lens/pull/4625/commits/18ec1432627cd1ff1246dfe0897ddd2dca59f0cf)

jim-docker:
windows integration tests fail intermittently, seems to be due to an issue in playwright (something that hasn't been updated for Electron 14). Tried downgrading playwright to 1.15.2 and upgrading to 1.18.1, with same results. May be the same issue as recently reported [here](https://github.com/microsoft/playwright/issues/11095)

I have not found a workaround yet.

Nokel81 (dismissed):

Nokel81 (dismissed):

Nokel81 (approved):

jim-docker (approved):

PR #4630, opened 2022-01-05
Add links for ingresses

### LINKED ISSUE #4626: Allow opening or copying ingress URL
Please add clickable link for each ingress final URL on both list and details :)

On ingress list page there is Name, Namespace, LoadBalancers, Rules and Age columns
Rules usually contains Ingress final URL with arrow like http://foobar.com/ -> service:80

Right now there is no easy way to open that in browser, also it's hard to copy that 
For multiple rules more than one address is visible in Rules columns, still cannot open any of them. 
On click right panel is opened, and instead of final url there is table with rules divided to host, path and backend. 

Right now for links I need to open dev tools and copy link from source (list). 

Expected changes: 
1. add link to list rules urls (that won't open details page) 
2. add same url to details somewhere near rules 
3. link backend services and tls 

**Why is this needed**:

Clickable link would speed up work a lot! 

win10 + k3s, but it's true for all clusters and systems.

Nokel81:
resolves: #4626

Screen Shot 2022-01-05 at 11 34 01 AM
Screen Shot 2022-01-05 at 12 16 36 PM

### DIFF
--- a/src/common/k8s-api/endpoints/ingress.api.ts
+++ b/src/common/k8s-api/endpoints/ingress.api.ts
@@ -3,8 +3,8 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
 
-import { KubeObject } from "../kube-object";
-import { autoBind, iter } from "../../utils";
+import { KubeObject, TypedLocalObjectReference } from "../kube-object";
+import { autoBind, hasTypedProperty, isString, iter } from "../../utils";
 import { IMetrics, metricsApi } from "./metrics.api";
 import { KubeApi } from "../kube-api";
 import type { KubeJsonApiData } from "../kube-json-api";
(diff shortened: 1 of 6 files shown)


src/renderer/components/+network-ingresses/ingress-details.tsx
```diff
@@ -65,44 +65,51 @@ export class IngressDetails extends React.Component<Props> {
...
+          <Table className="paths">
+            <TableHead>
+              <TableCell className="path">Path</TableCell>
+              <TableCell className="link">Link</TableCell>
+              <TableCell className="backends">Backends</TableCell>
+            </TableHead>
+            {
+              rule.http.paths
+                .map(({ path = "/", backend }, index) => {
+                  const link = `${protocol}://${rule.host || "*"}${path}`;
```
nevalla: if host is `*`, isn't hyperlink broken then?

Nokel81: Yes it is, I guess I should then not display them as links then.

aleksfront (dismissed):

src/renderer/components/+network-ingresses/ingresses.tsx
```diff
@@ -55,7 +55,19 @@ export class Ingresses extends React.Component<IngressesProps> {
...
+            <>
+              <a
+                key={route.url}
+                href={route.url}
+                rel="noreferrer"
+                target="_blank"
+                onClick={e => e.stopPropagation()}
+              >
+                {route.url}
+              </a> ⇢ {route.service}
```
jim-docker: This should also not be a link if `route.url` contains `*`?

jim-docker (commented):
doesn't work with resource backends? Just trying the example https://raw.githubusercontent.com/kubernetes/website/main/content/en/examples/service/networking/ingress-resource-backend.yaml from https://kubernetes.io/docs/concepts/services-networking/ingress/:

https://user-images.githubusercontent.com/40840436/158704799-a5083797-cf2c-45a5-9389-d02f9d31ad7a.mov

Nokel81:
@jim-docker I have fixed the crash you have observed as well (even though I don't think it was caused by these changes). And here are some screenshots of how it looks now.

Screen Shot 2022-03-21 at 9 19 27 AM
Screen Shot 2022-03-21 at 9 19 17 AM

src/common/k8s-api/endpoints/ingress.api.ts
```diff
@@ -196,6 +198,24 @@ export class Ingress extends KubeObject {
   }
 }
 
+export function computeRuleDeclarations(ingress: Ingress, rule: IngressRule): ComputedIngressRoute[] {
+  const { host = "*", http: { paths } = { paths: [] }} = rule;
+  const protocol = (ingress.spec.tls?.length ?? 0) > 0
```
jim-docker: I'm seeing every link getting https, something fishy here?

jim-docker: If I set `tls` it gets http so this is wrong, not exactly reversed with that `?? 0` I think

Nokel81: Thanks for catching this.

jim-docker (approved):

nevalla (approved):

Issue #4551, opened 2021-12-10
Move Extensions to Preferences page


aleksfront:
Extensions page should go into app Preferences as a separate tab.

https://docs.google.com/presentation/d/1ER32kCkGaXE80cEoC-Tj5YWVGHHxLoaW8ykd7qiX38Q/edit#slide=id.g1057c8236ec_0_91

Nokel81:
Closing as old

PR #4651, opened 2022-01-10
Group custom settings by extension

aleksfront:
Problem: as part of #4551 we need to open `Extensions` page from the Preferences correspondant tab. It was impossible because we reserved `Extensions` tab for showing extension custom settings.

Solution: grouping custom preferences by extension and show them as separate tabs. It is also more meaningful now, because previously settings from all extensions were listed in one page.

custom settings

https://user-images.githubusercontent.com/9607060/148756019-35dd2a85-3d50-4ada-9623-7ead22ea35ce.mov


src/renderer/components/+preferences/extension-settings-page.tsx
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    return currentSettings.filter(e => !e.showInPreferencesTab).map((setting) =>
+      <ExtensionSettings key={setting.id} setting={setting} size="small" />,
+    );
+  };
+
+  return (
+    <section id="extensions">
+      <h2>{extensionId} settings</h2>
```
Nokel81, line 65: shouldn't we call this what it is, the extension Name?

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -48,26 +46,40 @@ import { Kubernetes } from "./kubernetes";
...
-  constructor(props: {}) {
+  constructor(props: Dependencies) {
     super(props);
     makeObservable(this);
   }
 
   renderNavigation() {
-    const extensions = AppPreferenceRegistry.getInstance().getItems();
-    const telemetryExtensions = extensions.filter(e => e.showInPreferencesTab == "telemetry");
+    const preferenceRegistries = uniqBy(AppPreferenceRegistry.getInstance().getItems(), "extensionId");
```
Nokel81, line 71: I think you should coordinate with @jim-docker, because he said that he was working to make app preferences injectable but this would be a perfect PR for them to be done in.

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -96,10 +120,19 @@ export class Preferences extends React.Component {
...
-          <Route path={extensionURL()} component={Extensions}/>
+          <Route path={extensionSettingsURL()} component={ExtensionSettingsPage}/>
           <Redirect exact from={`${preferencesURL()}/`} to={appURL()}/>
         </Switch>
       </SettingLayout>
     );
   }
 }
+
+export default withInjectables<Dependencies>(
```
Nokel81, line 131: I think we should do `export const Preferences = withInjectables<...>(...`

src/renderer/components/+preferences/preferences.tsx
Nokel81: I also think that this logic is a bit confusing and could be simplified.

Finally, why don't we just implement that extensions can add to any page in this PR too, it would be quite easy.

src/renderer/components/+preferences/__tests__/extension-settings-page.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { getDiForUnitTesting } from "../../getDiForUnitTesting";
+import { AppPreferenceRegistry } from "../../../../extensions/registries";
+import "@testing-library/jest-dom/extend-expect";
+import { LensExtension } from "../../../../extensions/lens-extension";
+import { ExtensionSettingsPage } from "../extension-settings-page";
+import { createMemoryHistory, createLocation } from "history";
+import type { match } from "react-router";
+import { extensionSettingsRoute, extensionSettingsURL } from "../../../../common/routes";
+
+const history = createMemoryHistory();
```
ixrock, line 34: I wouldn't recommend to use this since `history` module removed in `react-router@5` (which we sooner or later upgrade)

src/renderer/components/+preferences/extension-settings-page.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { AppPreferenceRegistry } from "../../../extensions/registries";
+import { Icon } from "../icon";
+import { ExtensionSettings } from "./extension-settings";
+
+interface Props extends RouteComponentProps<{ extensionId?: string }> {
+}
+
+export const ExtensionSettingsPage = (props: Props) => {
+  // https://github.com/remix-run/react-router/issues/5870#issuecomment-394194338
+  const match = matchPath<{ extensionId: string }>(props.location.pathname, {
```
ixrock, line 36: Same as comment above, these apis not exist in `react-router` v5

src/renderer/components/+preferences/kubeconfig-syncs.tsx
```diff
@@ -202,7 +200,9 @@ export class KubeconfigSyncs extends React.Component {
       <>
         {this.renderSyncButtons()}
         <SubTitle title="Synced Items" className="pt-5"/>
-        {this.renderEntries()}
+        <div className="relative">
```
ixrock, line 203: Why using global pretty-common class-name  `relative`?

src/renderer/components/+preferences/preferences.tsx
```diff
@@ -48,26 +46,40 @@ import { Kubernetes } from "./kubernetes";
...
-import { Extensions } from "./extensions";
 import { sentryDsn } from "../../../common/vars";
+import { withInjectables } from "@ogre-tools/injectable-react";
+import type { InstalledExtension } from "../../../extensions/extension-discovery";
+import userExtensionsInjectable from "../+extensions/user-extensions/user-extensions.injectable";
+import { ExtensionSettingsPage } from "./extension-settings-page";
+import { Icon } from "../icon";
+import { uniqBy } from "lodash";
+
+interface Dependencies {
```
ixrock, line 57: Why do we call it `Dependencies`? I'd better to stick to `Props` name for component props interface.

Nokel81: There is a difference, the dependencies will be injected via DI and (`withInjectables`) but the props will not.

They need to be separate types (which are `&`-ed together) so that the exported component type does not reference this type at all.

Nokel81: For example see https://github.com/lensapp/lens/pull/4653/files#diff-9439f9090e19e6ccbf3899490a1995b5bf2a63efd852dc6df1c76c1ce3aae53f from @jansav's PR

ixrock: I'm fine with  `interface Props extends Dependencies { ... }` where `Props` should be used for react-component.

Nokel81: But we don't want that in the documentation. The docs should not contain the dependencies of the components

aleksfront:
This PR will be refactored a bit after https://github.com/lensapp/lens/pull/4671 will be merged. Since it's heavily uses `AppPreferenceRegistry` and now it's converted as a dependency.

src/renderer/components/+preferences/extension-settings-page.tsx
aleksfront: Should we rather move to react-router v6? https://reactrouter.com/

aleksfront:
Closing this in favor of https://github.com/lensapp/lens/pull/5284

PR #4666, opened 2022-01-10
Allow to customize terminal theme

### LINKED ISSUE #2224: Dark edit recourse area in light color theme
**Describe the bug**
Dark edit resource area in light color theme.

**To Reproduce**
1. Go to edit any recourse 

**Expected behavior**
All elements of interface are equal.

**Screenshots**

- Lens Version: 4.1.3 (in 4.1.0 - 4.1.2 same problem, in 4.0.8 work fine)
- OS: Windows 10
- Installation method: from installer

ixrock:
fix  #2224


Nokel81:
Wouldn't it be better (and more complete) if the option was instead "Terminal Color Scheme" with a dropdown of options "Match Theme, Light, Dark"?

Also, could you make the title of the PR more descriptive. I know it is the same as the linked issue but it isn't a very good issue name.

ixrock:
> Wouldn't it be better (and more complete) if the option was instead "Terminal Color Scheme" with a dropdown of options "Match Theme, Light, Dark"?

Dunno.. since we have only 2 themes (both internal: `lens-dark` & `lens-light`) this simplicity is good enough currently imho.
With growing amount of themes we can provide more specific choice for terminal colors from other themes.

Nokel81:
My only concern with your current proposal is that it doesn't seem that a user with a dark theme picked can use a light theme on the terminal

jim-docker:
> My only concern with your current proposal is that it doesn't seem that a user with a dark theme picked can use a light theme on the terminal

I was thinking the same thing ("Match Theme, Light, Dark"). Even if we added more themes the non-theme choices for the terminal could be limited to light and dark)

aleksfront:
I agree that dropdown (or radio group) version with 3 options would be a better solution. The reason is a bit of confusion for the dark theme. Terminal stays dark despite the introduced switch.

aleksfront:
Btw noticed terminal positioning issues. That's what I get after focusing lens window after a while:
half terminal

After opening couple of more terminals I see unexpected bottom border under xterm:
terminal border

ixrock:
Screenshot 2022-01-11 at 14 30 04

aleksfront:
Maybe use notation @jim-docker has suggested: "Match theme, Light, Dark"? "Lens" part in `Dark (Lens)` and `Light (Lens)` options can create confusion.

aleksfront:
Still have weird positioning/refreshing bugs:

https://user-images.githubusercontent.com/9607060/148949648-ffeb90f2-5c3d-4f8e-92ec-20f980ca8bae.mov

ixrock:
@lensapp/lens-maintainers PTALA

Screenshot 2022-01-11 at 18 09 10

aleksfront (approved):
LGTM

Nokel81 (approved):

PR #4615, opened 2021-12-28
Replace WelcomeMenuRegistry and WelcomeBannerRegistry with reactive solution

jansav:
This PR will:
1. Remove `WelcomeMenuRegistry` and introduce computed `welcomeMenuItemsInjectable` which observes items from extensions.
2. Remove `WelcomeBannerRegistry` and introduce computed `welcomeBannerItemsInjectable` which observes items from extensions.

PR #4671, opened 2022-01-11
Refactor app-preferences-registry to use di

jim-docker:


src/renderer/components/+preferences/preferences.tsx
```diff
@@ -50,18 +49,18 @@ import { LensProxy } from "./proxy";
 import { Telemetry } from "./telemetry";
 import { Extensions } from "./extensions";
 import { sentryDsn } from "../../../common/vars";
+import { withInjectables } from "@ogre-tools/injectable-react";
+import type { AppPreferenceRegistration } from "./app-preferences/app-preference-registration";
+import appPreferencesInjectable from "./app-preferences/app-preferences.injectable";
 
-@observer
-export class Preferences extends React.Component {
-  @observable historyLength: number | undefined;
```
jim-docker, line 40: This was never used, unless I missed something

src/renderer/components/+preferences/app-preferences/get-app-preferences.ts
```diff
@@ -0,0 +1,49 @@
...
+
+
+export const getAppPreferences = ({ extensions }: Dependencies) => {
+  return computed(() => {
+    return [
+      ...extensions.get()
+        .flatMap((extension) => extension.appPreferences)
+        .map(pref => getRegisteredItem(pref)),
+    ];},
+  );
```
Nokel81: nit, there is never a need to spread a `.map()` call.

```suggestion
  return computed(() => (
    extensions.get()
      .flatMap((extension) => extension.appPreferences)
      .map(getRegisteredItem)
  ));
```

src/renderer/components/+preferences/app-preferences/app-preference-registration.d.ts
```diff
@@ -33,16 +32,3 @@ export interface AppPreferenceRegistration {
   showInPreferencesTab?: string;
   components: AppPreferenceComponents;
 }
-
-export interface RegisteredAppPreference extends AppPreferenceRegistration {
-  id: string;
-}
```
Nokel81: I don't think we should remove this type. I think that the injectable's type should be `IComputedValue<RegisteredAppPreference[]>`.

That way we communicate that `id` is always defined (which is good when we (hopefully soon) turn on `strict`)

src/renderer/components/+preferences/application.tsx
```diff
@@ -56,7 +62,7 @@ export const Application = observer(() => {
 
   const [customUrl, setCustomUrl] = React.useState(userStore.extensionRegistryUrl.customUrl || "");
   const [shell, setShell] = React.useState(userStore.shell || "");
-  const extensionSettings = AppPreferenceRegistry.getInstance().getItems().filter((preference) => preference.showInPreferencesTab === "application");
+  const extensionSettings = appPreferenceItems.get().filter((preference) => preference.showInPreferencesTab === "application");
```
Nokel81, line 49: To go even further, there could be an injectable whose type is

```
IComputedValue<Map<string, RegisteredAppPreference[]>>
```

which does the splitting once

src/renderer/components/+preferences/extensions.tsx
```diff
@@ -19,13 +19,21 @@
...
+import appPreferencesInjectable from "./app-preferences/app-preferences.injectable";
 import { ExtensionSettings } from "./extension-settings";
 
-export const Extensions = observer(() => {
-  const settings = AppPreferenceRegistry.getInstance().getItems();
+interface Dependencies {
+  appPreferenceItems: IComputedValue<AppPreferenceRegistration[]>
+}
+
+const NonInjectedExtensions: React.FC<Dependencies> = ({ appPreferenceItems }) => {
```
Nokel81, line 18: This needs to be `observer`. Which also means you can get rid of the type annotation

```
const NonInjectedExtensions = observer(({ appPreferenceItems }: Dependencies) => { ...
```

Nokel81: Ah I see that you do that below...

src/renderer/components/+preferences/app-preferences/get-app-preferences.ts
jim-docker: good catch. I was duplicating what was done in https://github.com/lensapp/lens/pull/4615, but there they have another element in the `[ ]`

src/renderer/components/+preferences/app-preferences/app-preference-registration.d.ts
jim-docker: Thanks, I wasn't sure the best way to handle this type. Internally it seems we always use `RegisteredAppPreference`

src/renderer/components/+preferences/extensions.tsx
jim-docker: again, I took this pattern from #4615

Nokel81: Fair, enough I don't think it matters that much.

src/renderer/components/+preferences/telemetry.tsx
```diff
@@ -22,13 +22,20 @@ import { observer } from "mobx-react";
...
 
-export const Telemetry = observer(() => {
-  const extensions = AppPreferenceRegistry.getInstance().getItems();
+interface Dependencies {
+  appPreferenceItems: IComputedValue<RegisteredAppPreference[]>
+}
+
+const NonInjectedTelemetry: React.FC<Dependencies> = ({ appPreferenceItems }) => {
+  const extensions = appPreferenceItems.get();
   const telemetryExtensions = extensions.filter(e => e.showInPreferencesTab == "telemetry");
```
Nokel81, line 23: I sort of think that this should be its own injectable

src/renderer/components/+preferences/application.tsx
jim-docker: I don't follow...

Nokel81: We currently have one injectable: `appPreferenceItemsInjectable`. But we could change its type to from `IComputedValue<RegisteredAppPreference[]>` to `IComputedValue<Map<string, RegisteredAppPreference[]>>` and then have an injectable for each of the pages, say called `telemetryPreferenceItemsInjectables`. That way when we come to test each component, we only need to override the single injectable and the filtering is moved out of the component.

jim-docker: Sounds good. But that may be better done in another PR? After https://github.com/lensapp/lens/pull/4651 is merged, which is waiting on this PR.

have to resolve conflicts first...

Nokel81: Okay sounds good

src/renderer/components/+preferences/telemetry.tsx
aleksfront: I can do this in #4651

aleksfront (approved):
LGTM

Nokel81 (approved):

jansav:
Merging this closed #4649

PR #4674, opened 2022-01-11
Fix Kubectl.downloadKubectl()

Nokel81:

fixes regression from #4556

### DIFF
--- a/src/main/kubectl.ts
+++ b/src/main/kubectl.ts
@@ -35,6 +35,7 @@ import { AppPaths } from "../common/app-paths";
 import got from "got/dist/source";
 import { promisify } from "util";
 import stream from "stream";
+import { noop } from "../renderer/utils";
 
 const bundledVersion = getBundledKubectlVersion();
 const kubectlMap: Map<string, string> = new Map([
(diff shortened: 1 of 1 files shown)


src/main/kubectl.ts
```diff
@@ -315,8 +316,14 @@ export class Kubectl {
     const fileWriteStream = fs.createWriteStream(this.path, { mode: 0o755 });
     const pipeline = promisify(stream.pipeline);
 
-    await pipeline(downloadStream, fileWriteStream);
-    logger.debug("kubectl binary download finished");
+    try {
+      await pipeline(downloadStream, fileWriteStream);
+      await fs.promises.chmod(this.path, 0o755);
```
jim-docker, line 321: Do you still need this given that the write stream is now created with that mode? I recall it wasn't created with that mode previously, maybe there was a reason (i.e. don't make it executable until it is successfully downloaded)?

jim-docker (approved):

ixrock (approved):

src/main/kubectl.ts
Nokel81: Maybe, but I was mostly just reverting the bug from before.

PR #4687, opened 2022-01-13
replacing the deprecated substr() with slice()

### LINKED ISSUE #4683: Replace deprecated String.substr() usage
`String.substr()` method is deprecated for now and should be replaced .

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

`String.prototype.slice()` can be an alternative.

gauravagrwal:
Fixes #4683 

**Description of changes:**
- using slice() instead of substr()


jweak (approved):
LGTM

ixrock (approved):

PR #4692, opened 2022-01-14
fix inputs not blurring when pressing enter

### LINKED ISSUE #4669: Allow saving cluster name by tapping esc key while on cluster name text box
In the current lens app when adding a cluster and trying to change its name, you must click outside the cluster name text box in order for the change to take place.
Tapping esc goes back to the previous screen without making any changes.

Everytime I add a cluster I must click outside the text box in order to save my change which means it takes a longer time to rename it and it's not quite intuitive.

OS: Ubuntu 20.04.3

Thanks.

ndrscodes:
Fixes #4669.

This fix/feature causes inputs to blur (causing the onChange event to be triggered) once the Enter key is pressed. As of now, the user has to manually click outside of the active input in order to save the value.

### DIFF
--- a/src/renderer/components/input/input.tsx
+++ b/src/renderer/components/input/input.tsx
@@ -68,6 +68,7 @@ export type InputProps = Omit<InputElementProps, "onChange" | "onSubmit"> & {
   iconRight?: IconData;
   contentRight?: string | React.ReactNode; // Any component of string goes after iconRight
   validators?: InputValidator | InputValidator[];
+  blurOnEnter?: boolean;
   onChange?(value: string, evt: React.ChangeEvent<InputElement>): void;
   onSubmit?(value: string, evt: React.KeyboardEvent<InputElement>): void;
 };
@@ -86,6 +87,7 @@ const defaultProps: Partial<InputProps> = {
   maxRows: 10000,
   showValidationLine: true,
   validators: [],
+  blurOnEnter: true,
 };
 
 export class Input extends React.Component<InputProps, State> {
@@ -284,8 +286,10 @@ export class Input extends React.Component<InputProps, State> {
         this.setDirty();
       }
 
-      //pressing enter indicates that the edit is complete, we can unfocus now
-      this.blur();
+      if(this.props.blurOnEnter){
+        //pressing enter indicates that the edit is complete, we can unfocus now
+        this.blur();
+      }
     }
   }
 
--- a/src/renderer/components/input/input.tsx
+++ b/src/renderer/components/input/input.tsx
@@ -283,6 +283,9 @@ export class Input extends React.Component<InputProps, State> {
       } else {
         this.setDirty();
       }
+
+      //pressing enter indicates that the edit is complete, we can unfocus now
+      this.blur();
     }
   }
 
--- a/src/renderer/components/editable-list/editable-list.tsx
+++ b/src/renderer/components/editable-list/editable-list.tsx
@@ -56,7 +56,6 @@ export class EditableList<T> extends React.Component<Props<T>> {
     if (val) {
       evt.preventDefault();
       this.props.add(val);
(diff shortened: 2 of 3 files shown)


src/renderer/components/input/input.tsx
```diff
@@ -283,6 +283,9 @@ export class Input extends React.Component<InputProps, State> {
       } else {
         this.setDirty();
       }
+
+      //pressing enter indicates that the edit is complete, we can unfocus now
+      this.blur();
```
Nokel81: Shouldn't this happen a few lines up, after the `this.setState()` call?

ndrscodes: It depends. doing it like this, the user can unfocus the input even if it's not valid. If we blurred the input right after calling setState, the user would lose the ability to do so (even if trying to submit invalid data doesn't make much sense).. Idk which solution is better.

Nokel81: How does this work with components like `EditableList`? Have you tried that?

makes sense

ndrscodes: i've tested EditableLists using the accessible namespaces. The namespace is added and after that, the input blurs. Should we keep the input focussed in those cases?

Nokel81: I think we should

ndrscodes: What's the cleanest way of doing this in this scenario? Checking the type of Input in the onKeyDown method seems pretty unclean, but that would mean each class would have to decide if the input needs to be blurred itself, or needs to  focus the input again. All of these solutions are not optimal imho

Nokel81: Since I think `EditableList` is the only place where this matters I would make a prop called something like `blurOnEnter` with default value `true` and then `EditableList` can set it to `false`.

editable-list.tsx
```diff
@@ -0,0 +1,95 @@
+/**
```
Nokel81: Did you mean to add files at the root of the repo?

ndrscodes: errare humanum est. I amended the commit and moved the files to their correct places. sorry!

Nokel81 (dismissed):

ndrscodes:
sorry for causing trouble again. I removed an unnecessary call to stopPropagation(). @Nokel81

Nokel81 (dismissed):

jim-docker (dismissed):

Nokel81:
@ndrscodes Can you please investigate the namespace selector test that is failing?

ndrscodes:
> @ndrscodes Can you please investigate the namespace selector test that is failing?

I've checked it, but i do not know what is causing this issue. It seems like the event is not being propagated to the Wizard/WizardStep. As soon as i implement this.blur() (even this.input.blur()), the WizardStep stops executing the next() function. Do you know what could be causing this? @Nokel81

ndrscodes:
> > @ndrscodes Can you please investigate the namespace selector test that is failing?
> 
> I've checked it, but i do not know what is causing this issue. It seems like the event is not being propagated to the Wizard/WizardStep. As soon as i implement this.blur() (even this.input.blur()), the WizardStep stops executing the next() function. Do you know what could be causing this? @Nokel81 

What we could do is giving an onKeyDown handler to the `WizardStep` in order to execute `submit` when pressing enter. That would also likely lead to cleaner implementation than trusting in onSubmit to fire if Enter is pressed.

Nokel81:
> What we could do is giving an onKeyDown handler to the WizardStep in order to execute submit when pressing enter. That would also likely lead to cleaner implementation than trusting in onSubmit to fire if Enter is pressed.

That is a good idea

Nokel81 (approved):
Looks great. Thanks!

ixrock (approved):

PR #4693, opened 2022-01-14
Add auto-detect support for Victoria Metrics Single

jakolehm:
Adds auto-detect support for Victoria Metrics Single version.

### DIFF
--- a/src/main/prometheus/victoria-metrics-single.ts
+++ b/src/main/prometheus/victoria-metrics-single.ts
@@ -19,19 +19,17 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { PrometheusProviderRegistry } from "../prometheus";
-import { PrometheusHelm } from "../prometheus/helm";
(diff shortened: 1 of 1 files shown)


stevejr (approved):
LGTM

If the user changes the Chart name by setting `values.nameOverride` then this will fail but at least that is known.

nevalla (approved):
LGTM

jim-docker (commented):
I can't get Lens to use vm. I followed [these instructions](https://docs.victoriametrics.com/guides/k8s-monitoring-via-vm-single.html?highlight=helm#1-victoriametrics-helm-repository), verified with grafana. The labels/selectors are correct for `getFirstNamespacedService()`. What am I missing? The Prometheus service address on the cluster settings page is not relevant, right? Or the name is inaccurate now. Setting it explicitly (to `default/vmsingle-victoria-metrics-single-server:8428`) didn't help anyway. Also tried using the name same as the chart (`victoria-metrics-single`).

jakolehm:
> I can't get Lens to use vm.

@jim-docker my cluster seems to have a custom config, will investigate this more when I have a bit of time.

Issue #642, opened 2020-08-05
[macOS] Add support for changing both font-type and font-size


conradwt:
**What would you like to be added**:

I would like the ability to change both the font and its size because I didn't see a way within your Lens -> Preferences.

**Why is this needed**:

I need it because I would like to reduce overall eye-strain.

**Environment you are Lens application on:**

- Kubernetes distribution: [EKS, AKS, GKE, Rancher, ...]

Docker for Desktop Mac, Digital Ocean, Linode, and GKE

- Desktop OS: macOS 10.15.6

kdx2:
I would like to ask about the exact same enhancement. Not only for Mac, it should be for all platforms you support.
I just installed Lens and gave it a go. The default terminal font loaded was like 7-8 on my 17'' laptop. The rest of the panels were readable though.
My current fix is zoom in but it's global. Even though that fixes the terminal font, all the rest scales up, too.

I suggest you to build-in the capability to set the font size of the terminal/editor independently of the rest of the application. Maybe **Ctl**+**+** would do or a setting in preferences.

karibertils:
I need this as well on linux. The fonts are barely readable. If i zoom out to make fonts readable the UI looks gigantic.

mebays:
I would like to second @kdx2 option of the capability to set the font of the terminal/editor independently. I use powerline in my terminal which usually involves selecting a powerline font. The current terminal font does ok, but I would prefer to set the terminal font to what I would normally use.

erhhung:
I'd say even separate the preferences for **terminal font** and **editor font** because, while both likely use a fixed width font, terminal would often use a **PowerLine** font with special symbols while the editor would use a font like **Fira Code** with special ligatures.

I often use Lens to **edit script file contents in ConfigMaps**, and the lack of a fixed width font makes editing difficult.

domainname:
Is there any hack to change font size, before this feature is available?

Nokel81:
related to #1166

w1593950:
The terminal front is really ugly for me 👎  Please include 'ubuntu' front as terminal default front or at lest changeable front. Please.

Nokel81:
We are in the process of allowing custom themes

colegatron:
> We are in the process of allowing custom themes

my eyes will thank you!! 
2500x1440 with that small font in a 27" is a pain for the eyes

ianjs:
+1

Font size in the terminal and editor are barely readable in Ubuntu. 

[...]

PR #4694, opened 2022-01-14
Font settings for editor and ternimal

DmitriyNoa:
https://github.com/lensapp/lens/issues/642

This introduces font settings for the editor and terminal.

As many have special types of fonts for terminals (Powerlevel, Nerdfonts, etc) font settings are separated for terminal and editor.

As there are more settings for the terminal now the change suggests moving terminal settings to its own tab.

Font size and font family settings for the terminal.
Screenshot 2022-01-14 at 15 41 40.

Font size and font family settings for the editor.
Screenshot 2022-01-14 at 15 41 34


src/common/user-store/preferences-helpers.ts
```diff
@@ -194,6 +195,32 @@ const downloadBinariesPath: PreferenceDescription<string | undefined> = {
   },
 };
 
+const terminalFontSize: PreferenceDescription<number | undefined> = {
+  fromStore(val) {
+    return val;
```
Nokel81: The benefit of having these descriptors is so that if the value is the default, we can return `undefined` (and thus not save them to disk). That way we can change the defaults in the future.

src/common/user-store/preferences-helpers.ts
```diff
@@ -36,17 +36,18 @@ export interface KubeconfigSyncValue {
 }
 
 export type EditorConfiguration = Pick<editor.IStandaloneEditorConstructionOptions,
-  "minimap" | "tabSize" | "lineNumbers">;
+  "minimap" | "tabSize" | "lineNumbers" | "fontSize" | "fontFamily">;
 
 export const defaultEditorConfig: EditorConfiguration = {
   tabSize: 2,
   lineNumbers: "on",
+  fontSize: 12,
```
Nokel81: why are you not using `defaultFontSize` here?

src/common/user-store/preferences-helpers.ts
```diff
@@ -36,17 +36,18 @@ export interface KubeconfigSyncValue {
...
+  fontFamily: defaultFontFamily,
```
Nokel81: should this var be called `defaultTerminalFontFamily` instead?

Also, given that it really should only be used here, I would move those vars into this file and not export them

src/renderer/components/dock/terminal.ts
```diff
@@ -49,8 +49,8 @@ export class Terminal {
   private xterm: XTerm | null = new XTerm({
     cursorBlink: true,
     cursorStyle: "bar",
-    fontSize: 13,
-    fontFamily: "RobotoMono",
+    fontSize: UserStore.getInstance().terminalFontSize || defaultFontSize,
```
Nokel81: `UserStore.getInstance().terminalFontSize` should always be set, even to the default value.

Nokel81: Also, I would say this is a good oportunity to do some DI changes, but I have already done some of it, so lets do that later.

src/renderer/components/dock/terminal.ts
```diff
@@ -49,8 +49,8 @@ export class Terminal {
   private xterm: XTerm | null = new XTerm({
```
Nokel81: Something should also change the font configs for already existing tabs if the preferences change.

src/renderer/components/+preferences/terminal.tsx
```diff
@@ -0,0 +1,99 @@
...
+    <section>
+      <SubTitle title="Font family"/>
+      <Input
+        theme="round-black"
+        type="text"
+        value={terminalSettings.terminalFontFamily}
+        onChange={(value) => setTerminalSettings({
+          ...terminalSettings,
+          terminalFontFamily: value.toString(),
+        })}
```
ixrock: Why not to update storage right-away? Using internal react-state looks redundant. Why to save on-blur?

```suggestion
        onChange={() => userStore.terminalConfig.fontFamily = value}
```

Note: `userStore.terminalConfig` is observable object, no need in `...spread` for update.

src/renderer/components/+preferences/editor.tsx
```diff
@@ -69,6 +69,27 @@ export const Editor = observer(() => {
...
       </section>
+      <section>
+        <SubTitle title="Font size"/>
+        <Input
+          theme="round-black"
+          type="number"
+          min={10}
+          validators={InputValidators.isNumber}
+          value={editorConfiguration.fontSize.toString()}
+          onChange={value => editorConfiguration.fontSize = Number(value)}
```
ixrock, line 80: With `<input type="number">` using `Number()` probably redundant.

DmitriyNoa: TS complains when you remove the toString() method from the `value` property so we have to use it there.
And this is causing further need to cast it to number onChange. We might need to refactor the Input component.

src/renderer/components/+preferences/terminal.tsx
DmitriyNoa: Save on blur was used in other places so I followed the conversion. We can move it to onChange. WDYT?

ixrock: Yes, I'd moved this to `onChange` and get rid of local react state.

src/renderer/components/+preferences/editor.tsx
Nokel81: Good idea, though I think we should instead start writing a new Input component from the ground up as the current one is very massive.

DmitriyNoa: This is even better. But probably out of scope of this PR

Nokel81: Definitely

src/renderer/components/+preferences/terminal.tsx
DmitriyNoa: Fixed

src/renderer/components/dock/terminal.ts
DmitriyNoa: Fixed

src/common/user-store/preferences-helpers.ts
```diff
@@ -273,6 +283,15 @@ const editorConfiguration: PreferenceDescription<EditorConfiguration, EditorConf
   },
 };
 
+const terminalConfig: PreferenceDescription<TerminalConfig, TerminalConfig> = {
+  fromStore(val) {
+    return merge(defaultTerminalConfig, val);
+  },
+  toStore(val) {
+    return val;
```
Nokel81, line 291: nit, I would change this to remove all the default settings as well. So that we can change the defaults in the future.

Nokel81 (approved):
Looking great

ixrock (approved):

PR #4702, opened 2022-01-17
Support extending KubernetesCluster in extensions

panuhorsmalahti:
This PR allows extension to specify a new KubernetesCluster by extending it.

Example use in extension:
```ts
const kind = "ManagedDevCluster";

export class ManagedDevCluster extends Common.Catalog.KubernetesCluster {
  public static readonly kind = kind;

  public readonly kind = kind;
}
```

This is required for an extension to add a new category to the catalog for specific type of cluster.

Three issues are fixed:
* the`kind` of `KubernetesCluster` is typed `string` so that the child class can change it
* `getItemsByEntityClass` is changed to use `instanceof` instead of `apiVersion`/`kind`. This way the child class (e.g. `ManagedDevCluster`)  will be true for `instanceof KubernetesCluster`, and Lens has hardcoded using `KubernetesCluster` in a few places.
* A `when` is changed to use  `catalogEntityRegistry.items` to check for "catalog has been populated". `.activeEntity` doesn't work, since that depends on extension being loaded (which will add the custom category to the catalog)
* Refactor `autoInitExtensions` to be async and return the array properly

### DIFF
--- a/src/common/catalog-entities/kubernetes-cluster.ts
+++ b/src/common/catalog-entities/kubernetes-cluster.ts
@@ -59,7 +59,7 @@ export interface KubernetesClusterStatus extends CatalogEntityStatus {
 }
 
 export class KubernetesCluster extends CatalogEntity<KubernetesClusterMetadata, KubernetesClusterStatus, KubernetesClusterSpec> {
-  public static readonly apiVersion = "entity.k8slens.dev/v1alpha1";
+  public static readonly apiVersion: string = "entity.k8slens.dev/v1alpha1";
   public static readonly kind: string = "KubernetesCluster";
 
   public readonly apiVersion = KubernetesCluster.apiVersion;
--- a/src/common/catalog-entities/kubernetes-cluster.ts
+++ b/src/common/catalog-entities/kubernetes-cluster.ts
@@ -60,7 +60,7 @@ export interface KubernetesClusterStatus extends CatalogEntityStatus {
 
 export class KubernetesCluster extends CatalogEntity<KubernetesClusterMetadata, KubernetesClusterStatus, KubernetesClusterSpec> {
   public static readonly apiVersion = "entity.k8slens.dev/v1alpha1";
-  public static readonly kind = "KubernetesCluster";
+  public static readonly kind: string = "KubernetesCluster";
 
   public readonly apiVersion = KubernetesCluster.apiVersion;
   public readonly kind = KubernetesCluster.kind;
--- a/src/extensions/extension-loader/extension-loader.ts
+++ b/src/extensions/extension-loader/extension-loader.ts
@@ -270,7 +270,7 @@ export class ExtensionLoader {
     });
   };
 
-  loadOnClusterRenderer(getCluster: () => KubernetesCluster) {
+  loadOnClusterRenderer = (getCluster: () => KubernetesCluster) => {
     logger.debug(`${logModule}: load on cluster renderer (dashboard)`);
 
     this.autoInitExtensions(async (extension: LensRendererExtension) => {
@@ -298,7 +298,7 @@ export class ExtensionLoader {
 
       return removeItems;
     });
-  }
+  };
 
   protected async loadExtensions(installedExtensions: Map<string, InstalledExtension>, register: (ext: LensExtension) => Promise<Disposer[]>) {
     // Steps of the function:
@@ -364,13 +364,13 @@ export class ExtensionLoader {
     });
   }
 
-  protected autoInitExtensions = (register: (ext: LensExtension) => Promise<Disposer[]>) => {
+  protected autoInitExtensions(register: (ext: LensExtension) => Promise<Disposer[]>) {
     // Setup reaction to load extensions on JSON changes
     reaction(() => this.toJSON(), installedExtensions => this.loadExtensions(installedExtensions, register));
 
     // Load initial extensions
     return this.loadExtensions(this.toJSON(), register);
-  };
+  }
 
   protected requireExtension(extension: InstalledExtension): LensExtensionConstructor | null {
     const entryPointName = ipcRenderer ? "renderer" : "main";
--- a/src/extensions/extension-loader/extension-loader.ts
+++ b/src/extensions/extension-loader/extension-loader.ts
@@ -270,11 +270,12 @@ export class ExtensionLoader {
     });
   };
 
-  loadOnClusterRenderer = (entity: KubernetesCluster) => {
+  loadOnClusterRenderer(getCluster: () => KubernetesCluster) {
     logger.debug(`${logModule}: load on cluster renderer (dashboard)`);
 
     this.autoInitExtensions(async (extension: LensRendererExtension) => {
-      if ((await extension.isEnabledForCluster(entity)) === false) {
+      // getCluster must be a callback, as the entity might be available only after an extension has been loaded
+      if ((await extension.isEnabledForCluster(getCluster())) === false) {
         return [];
       }
 
--- a/src/extensions/extension-loader/extension-loader.ts
+++ b/src/extensions/extension-loader/extension-loader.ts
@@ -364,7 +364,7 @@ export class ExtensionLoader {
     });
   }
 
-  protected autoInitExtensions(register: (ext: LensExtension) => Promise<Disposer[]>) {
+  protected autoInitExtensions = (register: (ext: LensExtension) => Promise<Disposer[]>) => {
     // Setup reaction to load extensions on JSON changes
     reaction(() => this.toJSON(), installedExtensions => this.loadExtensions(installedExtensions, register));
 
--- a/src/extensions/extension-loader/extension-loader.ts
+++ b/src/extensions/extension-loader/extension-loader.ts
@@ -298,7 +298,7 @@ export class ExtensionLoader {
 
       return removeItems;
     });
-  };
+  }
 
   protected autoInitExtensions(register: (ext: LensExtension) => Promise<Disposer[]>) {
     const loadingExtensions: ExtensionLoading[] = [];
--- a/src/extensions/extension-loader/extension-loader.ts
+++ b/src/extensions/extension-loader/extension-loader.ts
@@ -300,11 +300,16 @@ export class ExtensionLoader {
     });
   }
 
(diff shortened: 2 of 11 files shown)


src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -76,7 +76,7 @@ export interface KubernetesClusterStatus extends CatalogEntityStatus {
 
 export class KubernetesCluster extends CatalogEntity<KubernetesClusterMetadata, KubernetesClusterStatus, KubernetesClusterSpec> {
   public static readonly apiVersion = "entity.k8slens.dev/v1alpha1";
```
Nokel81: can you make this also string.

Nokel81 (dismissed):

src/renderer/cluster-frame.tsx
```diff
@@ -103,7 +103,8 @@ export class ClusterFrame extends React.Component {
 
     // Only load the extensions once the catalog has been populated
     when(
-      () => Boolean(catalogEntityRegistry.activeEntity),
+      // watch for .items, as .activeEntity might be populated only after extensions are loaded (if using custom Catalog Category)
```
jakolehm: This sounds a bit wrong, shouldn't `.activeEntity` work also in this case?

Nokel81: Especially since we still need `activeEntity` to be truthy for the next line.

panuhorsmalahti: activeEntity doesn't work if the entity happens to be of custom type, e.g. in my case ManagedDevCluster, and that type is added by an extension. In that case `getEntityForData` returns null, because it can't find the category. But we load the extensions only after the when resolves, and that will add the custom category.

@Nokel81's point seems correct though, not sure why the UI still seems to work. Probably I need to refactor this to have 3 steps: 
1. wait for catalog items
2. load extensions
3. Do the stuff in `loadOnClusterRenderer` with a known `entity`

src/common/catalog-entities/kubernetes-cluster.ts
panuhorsmalahti: Makes sense, thanks, fixed

src/renderer/cluster-frame.tsx
panuhorsmalahti: Improved this part. It still a bit.. dangerous because of this; we can have 10 extensions, only one of which will register the new class, so we should first  load the correct extension, and then the others that might use isEnabledForCluster.

src/extensions/lens-extension.ts
```diff
@@ -113,6 +112,11 @@ export class LensExtension {
     }
   }
 
+  @action
+  activate() {
+    return this.onActivate();
+  }
```
Nokel81, line 118: This is unnecessary, why not just call `onActivate`?

src/renderer/frames/cluster-frame/init-cluster-frame/init-cluster-frame.injectable.ts
```diff
@@ -12,16 +12,17 @@ import appEventBusInjectable from "../../../../common/app-event-bus/app-event-bu
 import clusterFrameContextInjectable from "../../../cluster-frame-context/cluster-frame-context.injectable";
 
 const initClusterFrameInjectable = getInjectable({
-  instantiate: (di) =>
-    initClusterFrame({
+  instantiate: (di) => {
+    return initClusterFrame({
```
Nokel81: I don't think you need to touch this in this PR

src/renderer/cluster-frame.tsx
panuhorsmalahti: Improved more, should be fixed now.

src/extensions/lens-extension.ts
panuhorsmalahti: That's protected, and I didn't want to change the API interface, I thought that could affect extensions. But I could just make it a public method if that's fine.

src/renderer/frames/cluster-frame/init-cluster-frame/init-cluster-frame.injectable.ts
panuhorsmalahti: Yes, forgot to change this back. Reverting change.

src/extensions/lens-extension.ts
Nokel81: You just did make change the API interface though...

Nokel81: Also, it seems strange that `enable` no longer calls `onActivate`. Could you also change its name? Because I find it a tad confusing why something would change `_enabled` while not calling `onActivate`

panuhorsmalahti: I mean making `onActive` from protected to public is a breaking change, since extensions need to implement that interface, right? But here we're adding a new method, which is a non-breaking change.

I'll try to figure out better names tomorrow. Basically we need two steps: 1. "load" and 2. "enable".

Nokel81: You could also just check the extensions and see if anyone also marks their overloads as protected too.

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -270,11 +270,12 @@ export class ExtensionLoader {
     });
   };
 
-  loadOnClusterRenderer = (entity: KubernetesCluster) => {
+  loadOnClusterRenderer(getCluster: () => KubernetesCluster) {
```
jansav: @panuhorsmalahti notice that arrow function here is important.

panuhorsmalahti: Fixed (i.e. reverted change, which was due to rebasing error).

src/extensions/lens-extension.ts
panuhorsmalahti: That seems risky.. we don't necessarily know about all the extensions, some might be private. Breaking change is still a breaking change, and adding a new method seems fine to me.

panuhorsmalahti: I think name `enable` is fine for that, and `activate` too. I can change if you can come up with something better.

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -312,35 +317,58 @@ export class ExtensionLoader {
...
-
-            loadingExtensions.push({
+            return {
+              extId,
+              instance,
               isBundled: extension.isBundled,
-              loaded,
-            });
-            this.instances.set(extId, instance);
+              activated: instance.activate(),
```
ixrock, line 333: Shouldn't be awaited?
```suggestion
              activated: await instance.activate(),
```

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -312,35 +317,58 @@ export class ExtensionLoader {
...
+    // as this might register new catalog categories. Afterwards we can safely .enable the extension.
+    await Promise.all(extensions.map(extension => extension.activated));
+
+    // Return ExtensionLoading[]
+    return extensions.map(extension => {
+      const loaded = extension.instance.enable(register).catch((err) => {
+        logger.error(`${logModule}: failed to enable`, { ext: extension, err });
+      });
+
+      this.instances.set(extension.extId, extension.instance);
```
ixrock, line 357: So we gonna save instance anyway in case of error?

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -312,35 +317,58 @@ export class ExtensionLoader {
...
+      return {
+        isBundled: extension.isBundled,
+        loaded,
+      };
     });
+  }
+
+  protected autoInitExtensions(register: (ext: LensExtension) => Promise<Disposer[]>) {
+    // Setup reaction to load extensions on JSON changes
+    reaction(() => this.toJSON(), installedExtensions => this.loadExtensions(installedExtensions, register));
```
ixrock, line 368: `this.toJSON()` seems like always return new object/value, should we limit amount of reactions maybe with 
`{fireImmediately: true, equals: comparer.structural}`?

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -312,35 +317,58 @@ export class ExtensionLoader {
...
-      fireImmediately: true,
+
+        return null;
+      })
+      // Remove null values
+      .filter(extension => Boolean(extension));
+
+    // We first need to wait until each extension's `onActivate` is resolved,
+    // as this might register new catalog categories. Afterwards we can safely .enable the extension.
+    await Promise.all(extensions.map(extension => extension.activated));
```
ixrock, line 349: `extension.activated` expected to be a boolean? If so we should wrap them into `when(() => extension.activated)` in order to use with `await Promise.all()`?

ixrock (dismissed):
LGTM

jweak:
Is this something that should updated to the Extension documentation (`docs/extensions`)?

src/extensions/extension-loader/extension-loader.ts
panuhorsmalahti: activated is a promise, so this is fine

src/extensions/extension-loader/extension-loader.ts
panuhorsmalahti: I removed `fireImmediately: true`, since I want to load extensions once when this is called, but return a promise, which is resolved after extensions have been loaded.

The reaction tracks one observable:  `this.extensions` of `ExtensionLoader`. It seems that `this.extensions` is only updated if there is some change, so this should work correctly even without `comparer.structural`.

src/extensions/extension-loader/extension-loader.ts
panuhorsmalahti: This mirrors the existing behavior.. not sure how this should be changed.

src/extensions/extension-loader/extension-loader.ts
panuhorsmalahti: That's not needed, as I want to store a promise to the object, then await them all, and go to the next step.

If I add `await` here, I need to add `async` to the `map` callback, which makes it an array of promises.

panuhorsmalahti:
@jweak good point, added documentation

ixrock (approved):

jweak (approved):
LGTM

PR #3310, opened 2021-07-07
Remove ExtenedMap, ToggleSet, and observable versions

Nokel81:
- Switch to helper functions instead. They don't impact the prototype
  tree and it means that we don't need to introduce versions that are
  both observable and "normal"

- Tidy <ItemListLayout> and hiddenTableColumns to not need
  pre-initialization

Signed-off-by: Sebastian Malton <sebastian@malton.name>

PR #4708, opened 2022-01-18
Add support for custom columns on catalog categories

### LINKED ISSUE #4686: Add support for custom category columns
Support for registering custom columns for categories. These should only be displayed when in a specific category view (not "Browse").

The api SHOULD be an OPTIONAL `priority: number` field for sort order (with some default, probably 50).

The api SHOULD be an OPTIONAL `defaultVisibility: boolean` field (default `true`) for determining if the column is by default visible (or else needs to be enabled by using the hidden columns feature).

The api SHOULD have a REQUIRED `id: string` field which needs to be unique to the extension. We should prepend the extension name to it to make them globally unique.

The api SHOULD have a REQUIRED `renderCell: (entity: CatalogEntity) => React.ReactNode` field for generating the cell contents.

The api SHOULD have a REQUIRED `renderColumnTitle: () => React.ReactNode` field for generating the table column title cell.

The api SHOULD have an OPTIONAL `sortCallback: (entity: CatalogEntity) => string | number | (string | number)[]` field, which when present determines the sort order when sorting by that column.

The api SHOULD have an OPTIONAL `searchFilter: (entity: CatalogEntity) => string | string[]` field which when present determines if that column should be searchable.

Nokel81:

resolves #4686

### DIFF
--- a/src/common/catalog/catalog-entity.ts
+++ b/src/common/catalog/catalog-entity.ts
@@ -84,35 +84,91 @@ export interface CatalogCategorySpec {
 }
 
 /**
- * If the filter returns true, the menu item is displayed
+ * If the filter return a thruthy value, the menu item is displayed
  */
 export type AddMenuFilter = (menu: CatalogEntityAddMenu) => any;
 
 export interface CatalogCategoryEvents {
+  /**
+   * This event will be emitted when the category is loaded in the catalog
+   * view.
+   */
   load: () => void;
+
+  /**
+   * This event will be emitted when the catalog add menu is opened and is the
+   * way to added entries to that menu.
+   */
   catalogAddMenu: (context: CatalogEntityAddMenuContext) => void;
+
+  /**
+   * This event will be emitted when the context menu for an entity is declared
+   * by this category is opened.
+   */
   contextMenuOpen: (entity: CatalogEntity, context: CatalogEntityContextMenuContext) => void;
 }
 
 export abstract class CatalogCategory extends (EventEmitter as new () => TypedEmitter<CatalogCategoryEvents>) {
+  /**
+   * The version of category that you are wanting to declare.
+   *
+   * Currently supported values:
+   *
+   * - `"catalog.k8slens.dev/v1alpha1"`
+   */
   abstract readonly apiVersion: string;
+
+  /**
+   * The kind of item you wish to declare.
+   *
+   * Currently supported values:
+   *
+   * - `"CatalogCategory"`
+   */
   abstract readonly kind: string;
-  abstract metadata: {
+
+  /**
+   * The data about the category itself
+   */
+  abstract readonly metadata: {
+    /**
+     * The name of your category. The category can be searched for by this
+     * value. This will also be used for the catalog menu.
+     */
     name: string;
+
+    /**
+     * Either an `<svg>` or the name of an icon from {@link IconProps}
+     */
     icon: string;
   };
+
+  /**
+   * The most important part of a category, as it is where entity versions are declared.
+   */
   abstract spec: CatalogCategorySpec;
 
+  /**
+   * @internal
+   */
   protected filters = observable.set<AddMenuFilter>([], {
     deep: false,
   });
 
-  static parseId(id = ""): { group?: string, kind?: string } {
+  /**
+   * Parse a category ID into parts.
+   * @param id The id of a category is parse
+   * @returns The group and kind parts of the ID
+   */
+  public static parseId(id: string): { group?: string, kind?: string } {
     const [group, kind] = id.split("/") ?? [];
 
     return { group, kind };
   }
 
+  /**
+   * Get the ID of this category
+   */
   public getId(): string {
     return `${this.spec.group}/${this.spec.names.kind}`;
   }
--- a/src/common/catalog/catalog-entity.ts
+++ b/src/common/catalog/catalog-entity.ts
@@ -8,6 +8,7 @@ import type TypedEmitter from "typed-emitter";
 import { observable, makeObservable } from "mobx";
 import { once } from "lodash";
 import { iter, Disposer } from "../utils";
+import type { CategoryColumnRegistration } from "../../renderer/components/+catalog/custom-category-columns";
 
 type ExtractEntityMetadataType<Entity> = Entity extends CatalogEntity<infer Metadata> ? Metadata : never;
 type ExtractEntityStatusType<Entity> = Entity extends CatalogEntity<any, infer Status> ? Status : never;
@@ -46,6 +47,7 @@ export interface CatalogCategorySpec {
    * The grouping for for the category. This MUST be a DNS label.
    */
   group: string;
+
   /**
    * The specific versions of the constructors.
    *
@@ -54,6 +56,10 @@ export interface CatalogCategorySpec {
    * `name = "v1alpha1"` then the resulting `.apiVersion` MUST be `entity.k8slens.dev/v1alpha1`
    */
   versions: CatalogCategoryVersion<CatalogEntity>[];
+
+  /**
+   * This is the concerning the category
+   */
   names: {
     /**
      * The kind of entity that this category is for. This value MUST be a DNS
@@ -62,6 +68,19 @@ export interface CatalogCategorySpec {
      */
     kind: string;
   };
+
+  /**
+   * These are the columns used for displaying entities when in the catalog.
+   *
+   * If this is not provided then some default columns will be used, similar in
+   * scope to the columns in the "Browse" view.
+   *
+   * Even if you provide columns, a "Name" column will be provided as well with
+   * `priority: 0`.
+   *
+   * These columns will not be used in the "Browse" view.
+   */
+  displayColumns?: CategoryColumnRegistration[];
 }
 
 /**
--- a/src/common/utils/collection-functions.ts
+++ b/src/common/utils/collection-functions.ts
@@ -1,0 +1,91 @@
+/**
+ * Copyright (c) 2021 OpenLens Authors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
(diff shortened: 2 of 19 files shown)


src/common/utils/collection-functions.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Get the value behind `key`. If it was not pressent, first insert the value returned by `getVal`
+ *
+ * This function should be used over `getInsertWith` if it is expensive to build a `V`
+ * @param map The map to interact with
+ * @param key The key to insert into the map with
+ * @param builder A function that returns a new instance of `V`.
+ * @returns The value in the map
+ */
+export function getOrInsertWith<K, V>(map: Map<K, V>, key: K, builder: () => V): V {
```
jakolehm: This is not really used anywhere, why would we add "dead code"?

src/common/utils/collection-functions.ts
```diff
@@ -0,0 +1,91 @@
...
+  return map.set(key, builder()).get(key);
+}
+
+/**
+ * Set the value associated with `key` iff there was not a previous value
+ * @param map The map to interact with
+ * @throws if `key` already in map
+ * @returns `this` so that `strictSet` can be chained
+ */
+export function strictSet<K, V>(map: Map<K, V>, key: K, val: V): typeof map {
```
jakolehm: Same here, not used anywhere?

src/common/utils/collection-functions.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  return map.set(key, val);
+}
+
+/**
+ * Get the value associated with `key`
+ * @param map The map to interact with
+ * @throws if `key` did not a value associated with it
+ */
+export function strictGet<K, V>(map: Map<K, V>, key: K): V {
```
jakolehm: Same here, not used anywhere?

src/common/utils/collection-functions.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  return map.get(key);
+}
+
+/**
+ * Toggles the "has" in `set` of `value`
+ * @param set The set to interact with
+ * @param value The value to either add if not present or remove if present from `set`
+ */
+export function toggle<T>(set: Set<T>, value: T): void {
```
jakolehm: Same here, not used anywhere?

src/common/utils/index.ts
```diff
@@ -26,11 +26,19 @@ export function noop<T extends any[]>(...args: T): void {
   return void args;
 }
 
+/**
+ * A typecorrect version of <function>.bind()
+ */
+export function bind<BoundArgs extends any[], NonBoundArgs extends any[], ReturnType>(fn: (...args: [...BoundArgs, ...NonBoundArgs]) => ReturnType, thisArg: any, ...boundArgs: BoundArgs): (...args: NonBoundArgs) => ReturnType {
```
jakolehm, line 16: Shouldn't this have tests? (not very easy to read so hard to say if it works)

src/renderer/components/+catalog/custom-category-columns.injectable.tsx
```diff
@@ -0,0 +1,70 @@
...
+import type { LensRendererExtension } from "../../../extensions/lens-renderer-extension";
+import rendererExtensionsInjectable from "../../../extensions/renderer-extensions.injectable";
+import { getOrInsert } from "../../utils";
+import type { RegisteredAdditionalCategoryColumn } from "./custom-category-columns";
+
+interface Dependencies {
+  extensions: IComputedValue<LensRendererExtension[]>;
+}
+
+function getAdditionCategoryColumns({ extensions }: Dependencies): IComputedValue<Map<string, Map<string, RegisteredAdditionalCategoryColumn[]>>> {
```
jakolehm, line 16: Should this be `getAdditionalCategoryColumns`?

src/renderer/components/+catalog/get-category-columns.injectable.ts
```diff
@@ -0,0 +1,96 @@
...
+  extensionColumns: IComputedValue<Map<string, Map<string, RegisteredAdditionalCategoryColumn[]>>>;
+}
+
+export interface GetCategoryColumnsParams {
+  activeCategory: CatalogCategory | null | undefined;
+}
+
+export type CategoryColumns = Required<Pick<ItemListLayoutProps<CatalogEntity>, "sortingCallbacks" | "searchFilters" | "renderTableContents" | "renderTableHeader">>;
+
+function getNonGlobalColums(activeCategory: CatalogCategory | null | undefined, extensionColumns: IComputedValue<Map<string, Map<string, RegisteredAdditionalCategoryColumn[]>>>) {
```
jakolehm: Could we improve function name, at least to me it's not clear how "non global" relates to the fact that it can return active category columns or browse all columns 🤔 Maybe it could be `getActiveCategoryOrBrowseAllColumns` ?

jakolehm (commented):
Few things still unclear / missing (after first round of code review):
- not yet clear how a category can define its own columns
- not yet clear how an extension can register additional columns for a category
- I would like to see tests that verify behaviour

src/common/utils/collection-functions.ts
Nokel81: I did a `git checkout` of this file from #3310 which is why the whole file was added. Will remove dead code.

src/renderer/components/+catalog/get-category-columns.injectable.ts
Nokel81: Sure, that might be a bit long. But I'll try and make this better

Nokel81:
> not yet clear how a category can define its own columns

From the discussion, I did not think that was what we wanted....

> not yet clear how an extension can register additional columns for a category

Here https://github.com/lensapp/lens/pull/4708/files#diff-27aabe91821c7d8aa0dc60e534727135220298aa56d098180f883c06f8b1d5c2R52

> I would like to see tests that verify behaviour

sounds good

jweak:
Lens Extension documentation should updated have this and maybe some example code as well.

src/common/catalog/catalog-entity.ts
```diff
@@ -62,6 +68,19 @@ export interface CatalogCategorySpec {
      */
     kind: string;
   };
+
+  /**
+   * These are the columns used for displaying entities when in the catalog.
+   *
+   * If this is not provided then some default columns will be used, similar in
```
panuhorsmalahti, line 75: Can we document what these "some" are? Maybe point to another file where they're specified.

src/renderer/components/+catalog/custom-category-columns.ts
```diff
@@ -0,0 +1,85 @@
...
+   */
+  kind: string;
+
+  /**
+   * The catalog entity group that is declared by the category for this registration
+   *
+   * e.g.
+   * - `"entity.k8slens.dev"`
+   */
+  group: string;
```
jweak, line 75: Could we make `entity.k8slens.dev` default and override it if group is defined?

src/common/utils/collection-functions.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+/**
+ * Get the value behind `key`. If it was not pressent, first insert `value`
```
panuhorsmalahti: typo: pressent

src/common/utils/collection-functions.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @param map The map to interact with
+ * @param key The key to insert into the map with
+ * @param value The value to optional add to the map
+ * @returns The value in the map
+ */
+export function getOrInsert<K, V>(map: Map<K, V>, key: K, value: V): V {
+  if (map.has(key)) {
```
panuhorsmalahti: Code golf:
```
if (!map.has(key)) {
    map.set(key, value);
}

return value;
```

jweak:
smoke tested with extension, seems to work pretty well



https://user-images.githubusercontent.com/774344/150125660-5292d959-635c-4dcf-9146-0cded1c12961.mov

src/common/catalog/catalog-entity.ts
```diff
@@ -62,6 +68,19 @@ export interface CatalogCategorySpec {
...
+   * scope to the columns in the "Browse" view.
+   *
+   * Even if you provide columns, a "Name" column will be provided as well with
+   * `priority: 0`.
```
jweak, line 79: I wonder if we should let priority 0 take over the name. Now it is impossible to put a column on the left side of name.

src/renderer/components/+catalog/custom-category-columns.ts
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * This is the type used to declare new catalog category columns
+ */
+export interface AdditionalCategoryColumnRegistration extends CategoryColumnRegistration {
+  /**
+   * The catalog entity kind that is declared by the category for this registration
+   *
+   * e.g.
+   * - `"KubernetesCluster"`
```
jweak, line 65: We should document these different `kind`s of default categories.

src/common/catalog/catalog-entity.ts
Nokel81: No there is a way, negative numbers.

src/renderer/components/+catalog/custom-category-columns.ts
Nokel81: I don't think that it should be done here but will soon export the category class types. So it will be included in the documentation

src/common/catalog/catalog-entity.ts
jweak: oh alright, got it 👍

src/common/utils/collection-functions.ts
Nokel81: No those have the wrong semantics. We want to get the value in the map always. The value provided is just the default value for the key.

src/renderer/components/+catalog/custom-category-columns.ts
Nokel81: I'd rather not, since it complicates the impl and doesn't gain us that much. It also makes internal categories "more important" than others.

src/common/catalog/catalog-entity.ts
Nokel81: I don't really want to because I don't want to lock in what they are. And from an API perspective I think it is fine that the users just have to try it out to find out.

src/common/utils/collection-functions.ts
panuhorsmalahti: Ah, thanks.

panuhorsmalahti (dismissed):

panuhorsmalahti (dismissed):

panuhorsmalahti (approved):

jweak (approved):
Looks good!

Issue #4703, opened 2022-01-17
Namespace Selector is closed when trying to multiselect


jansav:
**Describe the bug**
Any page using KubeObjectListLayout is rendered too often causing e.g. `NamespaceSelector` to close when not supposed to.

**To Reproduce**
1. Go to e.g. Pods
2. Click on namespace selector
3. Ctrl/Cmd + Click value to multiselect
4. See that selector is closed

**Expected behavior**
Selector shouldn't close

**Additional context**
Regression of #4653

jansav:
Relates also to https://github.com/lensapp/lens/pull/4653#issuecomment-1012457220

Also any KubeObjectListLayout usages have same issue that if you select row, it scrolls to top automatically. (re-renders wrong components)

PR #4710, opened 2022-01-18
Fix infinite render loop in release details

### LINKED ISSUE #4704: ReleaseDetails doesn't open
**Describe the bug**
Opening ReleaseDetails causes endless loading.

**To Reproduce**
1. Go to Releases
2. Click on any release
3. See endless loading

**Expected behavior**
Should open details

**Additional context**
Regression of #4653 

Fix almost done.

jansav:
Fix infinite render loop in release details by replacing stateful, UI-triggered releaseStore with reactive async computed

Significant bit here is usage of `asyncComputed` which permits a reactive pattern superior to stateful stores.

We propose to start considering `autorun` and `reaction` as new red-flag-trigger-words as they are very likely to implement a stateful anti-pattern where ever used.

Closes #4704 

> ## Example from codebase for why side-effects on render are universally bad-bad.
> 
> ### Current non-ideal behaviour
> 
> 1. release.tsx is rendered (uses releaseStore.items which is not currently populated)
> 2. Renders ItemListLayout and passes releaseStore as a prop
> 3. ItemListLayout calls releaseStore.loadItems() on mount. This is the criminal side-effect to populate the items.
> 4. releaseStore.loadItems() sets releaseStore.items
> 5. release.tsx is re-rendered -> back to 1 = infinite loop
> 
> Note: The reason why our team has been able to cope so far is very brittle stale UI state for which there is lot of hacky code (e.g. local duplicated state in component, autorun/reactions to update state, stateful instead of reactive solutions, etc.)
> 
> ### Proposed ideal behaviour
> 
> 1. release.tsx wants to be rendered
> 2. release.tsx injects releases as `asyncComputed`
> 3. observation of said `asyncComputed` triggers population of releases
> 4. release.tsx is rendered without releases but instead with observable knowledge of `asyncComputed` being pending
> 5. data for `asyncComputed` arrives and release.tsx observes the change and re-renders
> 
> ### Proposal:
> 1. We should have no requirement for anything but functional components in the future
> 2. We should have very little use for hooks
> 3. New red-flag-trigger words: reaction, autorun, init (outside of injectable.)
>


src/renderer/components/+apps-releases/releases.tsx
```diff
@@ -101,14 +101,57 @@ class NonInjectedHelmReleases extends Component<Dependencies & Props> {
...
 
   render() {
+    if (this.props.releasesArePending.get()) {
+      // TODO: Make Spinner "center" work properly
+      return <div className="flex center" style={{ height: "100%" }}><Spinner /></div>;
+    }
+
+    const releases = this.props.releases;
+
+    // TODO: Implement ItemListLayout without stateful stores
```
jansav, line 99: Notice: This TODO will be fixed in separate PR which tackles #4703.

src/renderer/components/+apps-releases/releases.injectable.ts
```diff
@@ -0,0 +1,40 @@
...
+import { listReleases } from "../../../common/k8s-api/endpoints/helm-releases.api";
+
+const releasesInjectable = getInjectable({
+  instantiate: (di) => {
+    const namespaceStore = di.inject(namespaceStoreInjectable);
+
+    // TODO: Inject clusterContext directly instead of accessing dependency of a dependency
+    const clusterContext = namespaceStore.context;
+
+    return asyncComputed(async () => {
```
jansav, line 17: Notice usage of `asyncComputed` as an advanced pattern.

src/renderer/components/+apps-releases/create-release/create-release.injectable.ts
```diff
@@ -0,0 +1,28 @@
...
+import releasesInjectable from "../releases.injectable";
+
+const createReleaseInjectable = getInjectable({
+  instantiate: (di) => {
+    const releases = di.inject(releasesInjectable);
+
+    return async (payload: IReleaseCreatePayload) => {
+      const release = await createRelease(payload);
+
+      releases.invalidate();
```
jansav, line 19: Notice how we have full control for when `asyncComputed` is to be invalidated/refreshed.

src/renderer/components/+apps-releases/create-release/create-release.injectable.ts
```diff
@@ -0,0 +1,28 @@
...
+  createRelease,
+  IReleaseCreatePayload,
+} from "../../../../common/k8s-api/endpoints/helm-releases.api";
+import releasesInjectable from "../releases.injectable";
+
+const createReleaseInjectable = getInjectable({
+  instantiate: (di) => {
+    const releases = di.inject(releasesInjectable);
+
+    return async (payload: IReleaseCreatePayload) => {
```
panuhorsmalahti, line 16: I haven't followed the discussion, but have we now decided to use the `I` prefix for interfaces? (And no, lets not start a long discussion about this :) ).

src/renderer/components/+apps-releases/get-release-secret/get-release-secret.ts
```diff
@@ -0,0 +1,21 @@
...
+import type { SecretsStore } from "../../+config-secrets/secrets.store";
+
+interface Dependencies {
+  secretsStore: SecretsStore;
+}
+
+export const getReleaseSecret =
+  ({ secretsStore }: Dependencies) =>
+    (release: HelmRelease) => {
+      return secretsStore
```
panuhorsmalahti: [nit] return can be left out for consistency

src/renderer/components/+apps-releases/create-release/create-release.injectable.ts
jansav: Existing code, so don’t know why it’s there. :)

src/renderer/components/+apps-releases/get-release-secret/get-release-secret.ts
jansav: Nice catch. Resolved

src/common/k8s-api/endpoints/helm-releases.api.ts
```diff
@@ -152,7 +152,7 @@ export async function rollbackRelease(name: string, namespace: string, revision:
   return apiBase.put(path, { data });
 }
 
-export interface HelmRelease {
+interface HelmReleaseDto {
```
Nokel81, line 155: what does DTO mean?

src/common/k8s-api/endpoints/helm-releases.api.ts
```diff
@@ -162,27 +162,30 @@ export interface HelmRelease {
...
+  getStatus: () => string
+  getVersion: () => string
+  getUpdated: (humanize?: boolean, compact?: boolean) => string | number
+  getRepo: () => Promise<string>
+}
 
-  static create(data: any) {
-    return new HelmRelease(data);
-  }
+const toHelmRelease = (release: HelmReleaseDto) : HelmRelease => ({
```
Nokel81, line 175: Why is this better than a class?

src/renderer/components/+apps-releases/create-release/create-release.injectable.ts
Nokel81: Ah so calling this means that `releases` should be refreshed?

src/renderer/components/+apps-releases/removable-releases.ts
```diff
@@ -0,0 +1,52 @@
...
+        },
+
+        delete: async () => {
+          await deleteRelease(release);
+        },
+      }),
+    ),
+  );
+};
+
```
Nokel81: nit: lots of blank lines

src/common/k8s-api/endpoints/helm-releases.api.ts
jansav: It’s better because now we can map it to use case specific models  

In this case we have removableReleases which have methods for selecting and removing release. 

If it’s class, we have no way to add functionality to it where needed without exposing functionality to everyone.

src/common/k8s-api/endpoints/helm-releases.api.ts
jansav: Data Transfer Object. 

So basically the model that is stored, retrieved etc. No getters, setters, methods etc. Only the data.

src/renderer/components/+apps-releases/create-release/create-release.injectable.ts
jansav: It means that releases are reloaded when observed.

src/common/k8s-api/endpoints/helm-releases.api.ts
Nokel81: I don't understand why you think that classes cannot have functionality added to them?

jansav: Of course you could, but it’s a lot more complex than it should be. 

Why do you think that this should be a class instead of a object? And if so, how would you implement removableReleases?

Nokel81: Why not a child class that takes the original as a constructor?

jansav: Because current design of `ItemListLayout` expects `items` to extend `ItemObject`. That means that child class would need to duplicate all properties and methods of a parent. 

What does class give us here that object doesn't? Optimization?

Usage of classes allow developer to more easily attach god-class anti-pattern and does not make **composition** easy. 

Objects in the other hand makes **composition** very easy and simple. (e.g. in this case we want to compose HelmRelease, "Selectable" (_which has nothing to do with HelmReleases_) and add custom delete -method. 

**Note: "Selectable" is inlined here, because it's only used once. If we see that it's duplicated a lot of times and think that it's not wrong abstraction, then we will create function which makes any object `Selectable`.**)

In a **hypothesis** where `Selectable` is a abstraction `removableReleases` could be written:

``` typescript
const removableReleases = ({
  releases,
}: {
  releases: IAsyncComputed<HelmRelease[]>;
}): IComputedValue<RemovableHelmRelease[]> => {
  return computed(() =>
    releases.map(toSelectable).map((release) => ({
      ...release,

      delete: () => {
        // Something
      },
    }))
  );
};


// Notice how we can make any object selectable
// ----------------

interface Selectable {
  isSelected: boolean;
  toggleSelection: () => void
}

interface Item {
  someProperty: string
}

interface SelectableItem extends Item, Selectable {}

const selectableItems = ({ items }: { items: Item[] }): SelectableItem =>
  items.map(toSelectable);
```

src/renderer/components/+apps-releases/removable-releases.ts
jansav: Good catch

src/common/k8s-api/endpoints/helm-releases.api.ts
nevalla: just curious, how would you add tests for this object that implements `HelmRelease` interface?

Nokel81: Maybe I should ask a different question, why does this need to happen at all. Why not just have `removeRelease` be a dependency?

jansav: > just curious, how would you add tests for this object that implements `HelmRelease` interface?

I don't see why you would want to test that directly? These should be tested on in some broader scope which uses it.

Iku-turso: @Nokel81 That would make `removeRelease` require parameters and thus add avoidable coupling and complexity.

Iku-turso: @nevalla this is a technical detail, and should probably not be unit-tested independently, but instead as part of bigger system. But for a good reason, I see no problem creating a technical unit-test for this. What is the difficulty in testing you see here?

jim-docker:
The infinite loop is gone but there's something else wrong:

https://user-images.githubusercontent.com/40840436/150235593-e7bb916c-7120-4e49-badf-99f39e66e965.mov

I let it spin for a long time. It did not load the release details. Performing the same actions on the 5.3.4 (prerelease) version worked fine.

If I then close and reopen the details page, the details do appear after a moment:

https://user-images.githubusercontent.com/40840436/150235819-90d36d55-ac47-444c-8a17-4660490fafbf.mov

jim-docker:
Actually, I left Lens running at the details page (at the end of the second video above) and after about 15 minutes it started spinning again (blank page with spinner)

src/renderer/components/+apps-releases/releases.tsx
jansav: Nope, didn't happen in that one.

src/common/k8s-api/endpoints/helm-releases.api.ts
nevalla: so we have HelmReleases API that have `listReleases` method. That method returns HelmRelease type of objects. This HelmRelease type of object has for example public `getVersion()` method that returns on some cases version from chart and on some cases empty string. Obviously this is something we want to add a test case that method is working as expected. So what is the right place to ensure the functionality of this public method?

jansav:
Nice catch @jim-docker. Reason for this is same old: side effects on rendering and autoruns to get data. Latest commits made `ReleaseDetails` have data as dependency to get rid of this problem.

src/common/k8s-api/endpoints/helm-releases.api.ts
jansav: @nevalla Well it doesn't matter at all if `HelmRelease` is object or class for that. If you really want to write that kind of black-box-test then you could do it like this:

```typescript
describe("someTest", () => {
  let apiBaseGetMock: AsyncFnMock<() => HelmReleaseDto[]>;
  let listReleases: () => Promise<HelmRelease[]>

  beforeEach(async () => {
    const di = getDiForUnitTesting();

    apiBaseGetMock = asyncFn();

    di.override(apiBaseInjectable, { get: apiBaseGetMock });

    await di.runSetups();

    listReleases = di.inject(listReleasesInjectable);
  });

  describe("when called", () => {
    let listReleasesPromise: Promise<HelmRelease[]>;

    beforeEach(() => {
      listReleasesPromise = listReleases();
    });

    it("calls API", () => {
      expect(apiBaseGetMock).toHaveBeenCalledWith({
        // Correct arguments
      });
    });

    describe("when API resolves", () => {
      beforeEach(async () => {
        const helmReleasesDtoStub = [
          {
            appVersion: "some-app-version",
            name: "some-name",
            namespace: "some-namespace",
            chart: "some-chart",
            status: "some-status",
            updated: "some-updated",
            revision: "some-revision",
          },

          {
            appVersion: "some-other-app-version",
            name: "some-other-name",
            namespace: "some-other-namespace",
            chart: "some-other-chart",
            status: "some-other-status",
            updated: "some-other-updated",
            revision: "some-other-revision",
          },
        ];

        await apiBaseGetMock.resolve(helmReleasesDtoStub);
      });

      it("knows version", async () => {
        const releases = await listReleasesPromise;

        const actual = releases.find(matches({ name: "some-name" }));

        expect(actual.getVersion()).toBe("Some special version");
      });
    });
  });
})
```

Iku-turso: @nevalla 
> So what is the right place to ensure the functionality of this public method?

The final right place does not exist yet, as codebase hostility prevents proper scope for unit tests, at least for some time still. We are working on this. While waiting for that, we can write technical unit tests as described by @jansav.

Also: instead of coupling unit tests systematically to technical details such as a "public method", we could do a lot better by starting to describe behavior in unit tests. But again, we are working in this.

Iku-turso: There's a lot to be said on this subject. Let's have a chat on your convenience. @nevalla? :)

nevalla: to me this pseudo test would make more sense here:
```
List helm releases
  when called
    it calls API
    it returns helm releases
```
and somewhere else 
```
Helm release
  knows version
  returns update as time diff
  etc
```

Or am I thinking this totally wrong?

> if you really want to write that kind of black-box-test then you could do it like this

~~do you mean that we should not write these kind of tests?~~ just noticed @Iku-turso comment above

nevalla: but yes...this is not the case we need to resolve on this PR

jansav: Hm. I don't see why we shouldn't? This is the only place that uses `listReleases` therefore only place that uses `static create()` and it benefits `listReleases` returning object instead of classes. We could even remove `listReleases` function.

Iku-turso: Hmm. We suppose we need to stop using PR-scope as excuse to not fix and touch stuff. It is simply something we cannot afford. If someone has the energy to make something better, cherishing that should be an overriding priority.

jim-docker:
Much better, but it still reloads the details page after leaving it for ~15 minutes. I see the spinner for a second first then the details. Maybe this pre-existed? Anyway maybe it is not that serious, it could be addressed in another PR.

An observation, in general it seems to take longer to load the releases and details (I see the spinner for longer) than on 5.3.3

jansav:
> Much better, but it still reloads the details page after leaving it for ~15 minutes. I see the spinner for a second first then the details. Maybe this pre-existed? Anyway maybe it is not that serious, it could be addressed in another PR.

It seems that there is somewhere else that updates `namespaces` after 15 minutes. Which then causes `releases` to be refreshed. It has been there before, not sure if stale UI state has prevented re-rendering though. However, I would say it's something to figure out in upcoming PR's. 

> An observation, in general it seems to take longer to load the releases and details (I see the spinner for longer) than on 5.3.3

Hmm, I made some research on this and it looks that `await getRelease(..)` is just taking long. So no changes to there. 

One thing that has changed, is that `getReleaseValues()` is called only when needed. 

Before it was called simultaneously with `getReleaseDetails` but now it's called after. That being said, it won't affect rendering time because Spinner is only present when `releaseDetails` has not arrived yet.

nevalla (approved):
LGTM

jakolehm (approved):
LGTM

PR #4653, opened 2022-01-10
Eliminate Global Shared State from app paths and relatives

jansav:
Massive PR for removing global shared state from AppPaths and all of its use places. 

Fast live reviews needed to get this PR merged asap for limit amount of merge conflicts.

Closes #4659
Closes #4665
Closes #4664 
Closes #4663 
Closes #4661 
Closes #4660 
Closes #4658 
Closes #4645 
Closes #4657 
Closes #4656 
Closes #4655 
Closes #4654 
Closes #4637 
Closes #4648 
Closes #4636 
Closes #4640 
Closes #4662 
Closes #4667
Closes #4641
Closes #4642
Closes #4643
Closes #4646
Closes #4639
Closes #4644

PR #4720, opened 2022-01-19
Fix involuntary rerenders in item list layout

### LINKED ISSUE #4703: Namespace Selector is closed when trying to multiselect

jansav:
Main attraction here is that `ItemListLayout` has now been split to multiple components to make re-rendering happen more locally instead of full-renders. It's also nice to have a smaller components. 

**The changes here highlight that ItemListLayout is wrong abstraction and should be re-designed.**

Closes #4703. 

Opening separate issue for proper love for `ItemListLayout`.

Note: This PR is built on top of #4710 which should be merged first.


src/renderer/components/item-object-list/item-list-layout-content/item-list-layout-content.tsx
```diff
@@ -0,0 +1,271 @@
...
+            if (!headCell || this.showColumn(headCell)) {
+              return <TableCell key={index} {...cellProps} />;
+            }
+
+            return null;
+          })
+        }
+        {renderItemMenu && (
+          <TableCell className="menu">
+            <div onClick={stopPropagation}>
```
ixrock: What does `stopPropagation` here and why do we need it?

src/renderer/components/item-object-list/item-list-layout-content/item-list-layout-content.tsx
```diff
@@ -0,0 +1,271 @@
...
+  removeItemsDialog() {
+    const { customizeRemoveDialog, store } = this.props;
+    const { selectedItems, removeSelectedItems } = store;
+    const visibleMaxNamesCount = 5;
+    const selectedNames = selectedItems.map(ns => ns.getName()).slice(0, visibleMaxNamesCount).join(", ");
+    const dialogCustomProps = customizeRemoveDialog ? customizeRemoveDialog(selectedItems) : {};
+    const selectedCount = selectedItems.length;
+    const tailCount = selectedCount > visibleMaxNamesCount ? selectedCount - visibleMaxNamesCount : 0;
+    const tail = tailCount > 0 ? <>, and <b>{tailCount}</b> more</> : null;
+    const message = selectedCount <= 1 ? <p>Remove item <b>{selectedNames}</b>?</p> : <p>Remove <b>{selectedCount}</b> items <b>{selectedNames}</b>{tail}?</p>;
```
ixrock: _nit:_




```suggestion
    const message = selectedCount === 1 
      ? <p>Remove item <b>{selectedNames}</b>?</p> 
      : <p>Remove <b>{selectedCount}</b> items <b>{selectedNames}</b>{tail}?</p>;
```

src/renderer/components/item-object-list/item-list-layout-content/item-list-layout-content.tsx
```diff
@@ -0,0 +1,271 @@
...
+    if (!this.props.getIsReady()) {
+      return <Spinner center />;
+    }
+
+    if (this.props.getFilters().length > 0) {
+      return (
+        <NoItems>
+          No items found.
+          <p>
+            <a onClick={() => pageFilters.reset()} className="contrast">
```
ixrock: Probably we should stop/ban using global css class names like `contrast` (except all from `tailwindcss` for the time being) since currenlt plan to have (s)css-modules only in whole codebase (e.g. `item-list-layout.module.scss`)

src/renderer/components/item-object-list/item-list-layout-header/item-list-layout-header-info/item-list-layout-header-info.tsx
```diff
@@ -0,0 +1,55 @@
...
+  <I extends ItemObject>({
+    headerPlaceholders,
+    getItems,
+    getFilters,
+    store,
+    toggleFilters,
+  }: ItemListLayoutHeaderInfoProps<I>) => {
+    const renderInfo = () => {
+      const allItemsCount = store.getTotalCount();
+      const itemsCount = getItems().length;
```
ixrock: _nit:_ I'd better to move destructing input params inside constructor (more readable):

```suggestion
export const ItemListLayoutHeaderInfo = observer(
  <I extends ItemObject>(props: ItemListLayoutHeaderInfoProps<I>) => {
    const { headerPlaceholders, getItems, getFilters, store, toggleFilters, } = props;
```

src/renderer/components/item-object-list/item-list-layout-header/item-list-layout-header-title/item-list-layout-header-title.tsx
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { observer } from "mobx-react";
+import React from "react";
+import type { HeaderPlaceholders } from "../../item-list-layout";
+
+interface ItemListLayoutHeaderTitleProps {
+  renderHeaderTitle: React.ReactNode | (() => React.ReactNode); // MIKKO: "this" parameter
```
ixrock: What means comment `// MIKKO: "this" parameter`?

src/renderer/components/item-object-list/item-list-layout-content/item-list-layout-content.tsx
```diff
@@ -0,0 +1,271 @@
...
+            <div onClick={stopPropagation}>
+              {renderItemMenu(item, store)}
+            </div>
+          </TableCell>
+        )}
+      </TableRow>
+    );
+  }
+
+  @boundMethod
```
ixrock: I'd like to remove `@boundMethod` in nearest future since we currently have 2 libraries for that:


```ts
// lens/src/common/utils/autobind.ts
import { boundMethod, boundClass } from "autobind-decorator";
import autoBindClass, { Options } from "auto-bind";...
```

For most cases to bind all class methods we use this:
```ts
constructor(private dependencies: Dependencies)
    makeObservable(this);
    autoBind(this); // note: this is important to call after `makeObservable`
```

For binding single methods I'm suggesting stick to arrow-functions or to use re-assign class fields with `.bind(this)` inside constructor.

ixrock (dismissed):
LGTM

src/renderer/components/item-object-list/item-list-layout-content/item-list-layout-content.tsx
jansav: Old code, no idea.

src/renderer/components/item-object-list/item-list-layout-content/item-list-layout-content.tsx
jansav: Yep, old code. This PR just splits these to separate components and make them observer where needed.

src/renderer/components/item-object-list/item-list-layout-header/item-list-layout-header-title/item-list-layout-header-title.tsx
jansav: Wow, good catch. Shouldn't be there. Just a reminder to have discussion with Mikko.

src/renderer/components/item-object-list/item-list-layout-content/item-list-layout-content.tsx
jansav: I'd like to remove all these binds. But, this is old code and not touching this on this PR

ixrock:
One other thing what I don't really like is too much nesting in codebase (besides too much bloated amount of files): `src/renderer/components/item-object-list/item-list-layout-content/item-list-layout-content.tsx`

Could we keep more flat in `src/renderer/components/item-object-list/item-*.tsx`?
Also would be nice to have whole bundle of inner components available at `src/renderer/components/item-object-list/index.ts`

src/renderer/components/item-object-list/item-list-layout-header/item-list-layout-header-info/item-list-layout-header-info.tsx
jansav: Updated

src/renderer/components/item-object-list/item-list-layout-header/item-list-layout-header-title/item-list-layout-header-title.tsx
jansav: Updated

src/renderer/components/item-object-list/item-list-layout-content/item-list-layout-content.tsx
jansav: Again, old code. Not on this PR. But I agree with your comment.

Iku-turso (dismissed):

jim-docker:
Just noticed this, not sure if this PR is the source of this bug:

https://user-images.githubusercontent.com/40840436/150566869-175f9a3f-5672-416f-a6b0-c1d6fd97a837.mov

It should select all/select none. Other uses of item-list-layout are ok.

**** confirmed this is happening on master too. Made issue #4740 and PR #4742  ****

jim-docker:
item/filtered count is missing (between Pods title and namespace selector):
Screen Shot 2022-01-21 at 12 07 10 PM

jim-docker:
The list scrolls to top as it rerenders due to changes in the list items (I'm trying to keep it near the bottom, which works on 5.3.x):

https://user-images.githubusercontent.com/40840436/150576354-aac7a94c-eda0-47d6-a8b5-81b165fd1805.mov


**** happens on master too, suspect it is due to #4653. This similar issue is still occurring as well: https://github.com/lensapp/lens/pull/4653#issuecomment-1012457220

Nokel81 (dismissed):

jim-docker (dismissed):
will fix https://github.com/lensapp/lens/pull/4720#issuecomment-1018698807 in separate PR

src/renderer/components/item-object-list/header-info.tsx
```diff
@@ -0,0 +1,54 @@
...
+        </>
+      );
+    }
+
+    return allItemsCount === 1
+      ? `${allItemsCount} item`
+      : `${allItemsCount} items`;
+  };
+
+  const info = headerPlaceholders.info ?? renderInfo();
```
jim-docker: This looks like the wrong check. It uses `headerPlaceholders.info` but it is simply an empty `</>`

Nokel81 (approved):

PR #4723, opened 2022-01-19
Convert CloseButton styles out from css modules

aleksfront:
Extensions webpack config doesn't allow to export components using scss modules for styling.

This PR converts styles to "regular" scss instead of scss modules as a workaround. 

esc button

### DIFF
--- a/src/renderer/components/layout/close-button.scss
+++ b/src/renderer/components/layout/close-button.scss
@@ -20,25 +20,27 @@
  */
 
 .SettingsCloseButton {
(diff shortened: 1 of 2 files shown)


Nokel81 (approved):
Wouldn't it be better to fix the webpack config instead?

aleksfront:
> Wouldn't it be better to fix the webpack config instead?

Ofc it will be better, will fix it in other PR.

jim-docker (approved):

PR #4724, opened 2022-01-20
Fix race condition with extension loading.

panuhorsmalahti:
Regression from https://github.com/lensapp/lens/pull/4702.

If `loadExtensions` is called multiple times before an extension is activated, `this.instances.set(extId, instance);` was not called fast enough, which caused the extension to be activated multiple times.


nevalla (approved):
LGTM

jweak (approved):

PR #4725, opened 2022-01-20
Upgrade to webpack@5

ixrock:
This also fixing nasty caching bug with UI files after combination(s) of `make clean && make dev` / page reload / `git checkout` branches / etc.

**TODO:**

- [x] fix unit tests
- [x] fix integration tests
- [x] fix warnings in compilation of `main` process _(types must be imported separately in files participating in extension-api re-exports)_ 
- [x] fix incorrect import warnings in compilation of `renderer` process
- [x] use `@pmmmwh/react-refresh-webpack-plugin` with `react-refresh-typescript` to support HMR in most cases


Nokel81:
Shouldn't think be marked as a chore?

ixrock:
> Shouldn't think be marked as a chore?

whatever :)

webpack.renderer.ts
```diff
@@ -46,7 +57,6 @@ export function webpackLensRenderer({ showVars = true } = {}): webpack.Configura
     },
     output: {
       libraryTarget: "global",
-      library: "",
```
ixrock, line 49: Not supported in `webpack@5`. What is was used for? @jakolehm maybe you know?

src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -134,7 +135,7 @@ class KubernetesClusterCategory extends CatalogCategory {
   public readonly kind = "CatalogCategory";
   public metadata = {
     name: "Clusters",
-    icon: require("./icons/kubernetes.svg?raw"),
+    icon: KubeClusterCategoryIcon,
```
Nokel81, line 138: Woot! 🎉

src/main/index.ts
```diff
@@ -43,18 +43,26 @@ import { initTray } from "./tray/tray";
 import { ShellSession } from "./shell-session/shell-session";
 import { getDi } from "./getDi";
 import extensionLoaderInjectable from "../extensions/extension-loader/extension-loader.injectable";
-import lensProtocolRouterMainInjectable from "./protocol-handler/lens-protocol-router-main/lens-protocol-router-main.injectable";
-import extensionDiscoveryInjectable from "../extensions/extension-discovery/extension-discovery.injectable";
-import directoryForExesInjectable from "../common/app-paths/directory-for-exes/directory-for-exes.injectable";
-import initIpcMainHandlersInjectable from "./initializers/init-ipc-main-handlers/init-ipc-main-handlers.injectable";
+import lensProtocolRouterMainInjectable
```
Nokel81: Can we not have these all on dual lines?

src/main/window-manager.ts
```diff
@@ -34,17 +34,14 @@ export class WindowManager extends Singleton {
   protected disposers: Record<string, Function> = {};
 
   @observable activeClusterId: ClusterId;
+  @observable mainContentUrl = `http://localhost:${LensProxy.getInstance().port}`;
```
Nokel81: I don't think this needs to be `@observable`

src/renderer/components/dock/terminal/terminal.ts
```diff
@@ -26,11 +27,14 @@ export class Terminal {
   }
 
   static async preloadFonts() {
-    const fontPath = require("../../fonts/roboto-mono-nerd.ttf").default; // eslint-disable-line @typescript-eslint/no-var-requires
-    const fontFace = new FontFace(defaultTerminalFontFamily, `url(${fontPath})`);
+    try {
+      const fontFace = new FontFace(defaultTerminalFontFamily, `url(${RobotoMonoFontUrl})`);
```
Nokel81: Is it possible to move this to the `HTML` file?

webpack.main.ts
```diff
@@ -6,23 +6,23 @@
...
 configs.push((): webpack.Configuration => {
   console.info("WEBPACK:main", vars);
+  const { mainDir, buildDir, isDevelopment } = vars;
 
   return {
     context: __dirname,
     target: "electron-main",
-    mode: isProduction ? "production" : "development",
-    devtool: isProduction ? "source-map" : "cheap-eval-source-map",
+    mode: isDevelopment ? "development" : "production",
```
Nokel81, line 24: Why was this switched?

src/main/index.ts
ixrock: Changes reverted. It was probably auto-formatting files via [.idea](https://github.com/lensapp/lens/tree/master/.idea) settings. Why do we even need to commit them to the repo? @jansav any clue?

webpack.main.ts
ixrock: Because it is the only place where `isProduction` was used when `isDevelopment` used in multiple places.
Also it's done to streamline code with other `webpack` configurations.

src/renderer/components/dock/terminal/terminal.ts
ixrock: If we have settings to change terminal's font - than not. (if it's not always bundled with the app)
But I have to double check this..

webpack.main.ts
Nokel81: Sounds good, I didn't notice that you also removed `isProduction` as a variable but reducing that set is good.

src/renderer/components/dock/terminal/terminal.ts
ixrock: Preloading fonts moved to `template.html` and `fonts.scss`

src/main/window-manager.ts
ixrock: Removed.

src/main/menu/menu.ts
```diff
@@ -9,13 +9,11 @@ import { appName, isMac, isWindows, docsUrl, supportUrl, productName } from "../
 import logger from "../logger";
 import { exitApp } from "../exit-app";
 import { broadcastMessage } from "../../common/ipc";
-import * as packageJson from "../../../package.json";
+import packageJson from "../../../package.json";
 import { preferencesURL, extensionsURL, addClusterURL, catalogURL, welcomeURL } from "../../common/routes";
 import { checkForUpdates, isAutoUpdateEnabled } from "../app-updater";
 import type { MenuRegistration } from "./menu-registration";
 
-export type MenuTopId = "mac" | "file" | "edit" | "view" | "help";
```
ixrock: Not used anywhere.

src/common/catalog-entities/kubernetes-cluster.ts
ixrock: By default imported `*.svg` files contain data-url (string) that can be consumed normally by `` or with `<Icon svg/>` as inline/raw xml icon.

src/main/menu/menu.ts
Nokel81: It should be, for the registration of menu items since we don't support adding new menu items.

ixrock: Well, it was used before. Dunno who was refactored menu that it now has some leftovers. 
Anyway, this is out of scope.

Nokel81: I think it is out of scope to remove it.

ixrock: Nope. This is just like removing a comment since it's not used / doesn't touch any other code.

Nokel81: Disagree, an upgrade PR like this should be only doing the minimal work to get it working, this should be a separate PR.

ixrock: Okay, I'll revert it, but just want to say.. With that frenetic attitude in PR's you probably gonna work alone on this project ~(maybe this is your goal)~.. Just see how other projects has quick PR reviews without PITA..

ixrock:
@lensapp/lens-maintainers PTAL, ready for testing & reviews.

_HMR demo:_

https://user-images.githubusercontent.com/6377066/153879443-bea48fa6-70aa-490f-bb65-77d3834a32e2.mov

extensions/kube-object-event-status/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kube-object-event-status",
-  "version": "0.0.1",
+  "version": "5.4.0-beta.1.1644584997435",
```
Nokel81: Can you revert this change?

extensions/kube-object-event-status/package.json
```diff
@@ -18,8 +18,8 @@
   "dependencies": {},
   "devDependencies": {
     "@k8slens/extensions": "file:../../src/extensions/npm/extensions",
-    "ts-loader": "^8.0.4",
-    "typescript": "^4.3.2",
-    "webpack": "^4.46.0"
+    "ts-loader": "latest",
```
Nokel81, line 21: We shouldn't use `latest` here because it will break if there is a mismatch of versions between this file and `../../package.json`

src/renderer/components/+network-ingresses/ingress-details.tsx
```diff
@@ -15,7 +15,7 @@ import { ResourceMetrics } from "../resource-metrics";
 import type { KubeObjectDetailsProps } from "../kube-object-details";
 import { IngressCharts } from "./ingress-charts";
 import { KubeObjectMeta } from "../kube-object-meta";
-import { getBackendServiceNamePort, getMetricsForIngress, IIngressMetrics } from "../../../common/k8s-api/endpoints/ingress.api";
+import { getBackendServiceNamePort, getMetricsForIngress, type IIngressMetrics } from "../../../common/k8s-api/endpoints/ingress.api";
```
Nokel81, line 18: Wondering out loud, is this down automatically? (Because it would be great if it was)

webpack.extensions.ts
```diff
@@ -6,9 +6,11 @@
 
 import path from "path";
 import type webpack from "webpack";
-import { sassCommonVars, isDevelopment, isProduction } from "./src/common/vars";
+import * as vars from "./src/common/vars";
+import { cssModulesWebpackRule, fontsLoaderWebpackRules, iconsAndImagesWebpackRules } from "./webpack.renderer";
 
 export default function generateExtensionTypes(): webpack.Configuration {
+  const { isDevelopment, isProduction } = vars;
```
Nokel81: Shouldn't we switch to just using `isDevelopment` here too?

src/renderer/components/+network-ingresses/ingress-details.tsx
ixrock: No idea.. but those warnings just in compilation process and probably could be ignored.

webpack.extensions.ts
ixrock: Yes, why not.

extensions/kube-object-event-status/package.json
ixrock: Why it will break when it's not used in "production" / inside packed extension?
I think only `ts-loader` must be compatible with local `webpack` package.

extensions/kube-object-event-status/package.json
ixrock: Reverted.

extensions/kube-object-event-status/package.json
Nokel81: We do run `npm install` on extensions, even bundled extensions.

ixrock: When we do? So maybe we should move it to `peerDependencies` then?
Anyway I think it should be handled from parent's package version, otherwise it's just copy-pasting all around. Can we use actual versions from app's `package.json` on extensions install?

Nokel81: https://github.com/lensapp/lens/blob/master/src/extensions/extension-discovery/extension-discovery.ts#L383-L406

This is where we do the install calls

ixrock: Well, it's still unclear how this installing dependency handle unpacking extension:

```ts
await this.dependencies.installExtension(extension.absolutePath);
```

In general it feels wrong if compiled extension require matching own development dependencies with the  app (in case of `react`, `mobx` and some other packages we're sharing them via global env because version matching there is important for app's work. @jakolehm @nevalla @panuhorsmalahti wdyt?

jakolehm: @Nokel81 @ixrock Lens installs extension npm package without development dependencies -> extension dev dependencies don't need to match same versions on Lens side.

jakolehm: Note: using `latest` version might still be a bit risky (in general).

jakolehm (approved):
LGTM

extensions/kube-object-event-status/package.json
ixrock: I think it's not more risky than just having outdated specific version == in both cases I assume it's easy to spot/fix/use what is needed per extension package.

src/common/catalog-entities/kubernetes-cluster.ts
aleksfront: Amazing!

aleksfront:
Found 2 terminal-related issues:
1. Fonts don't preload properly in terminal window.
broken fonts

2. An error occured on every terminal tab close:
```
Uncaught TypeError: Cannot read properties of null (reading 'url')
```

https://user-images.githubusercontent.com/9607060/154052621-d1aeeeb4-ccc5-4793-98f5-d13ec72b8076.mov

Nokel81:
@aleksfront that second issue is a known issue that exists on master. I don't think it should block this PR

Nokel81 (approved):
Except for the font issue. This looks good to me.

ixrock:
> Except for the font issue. This looks good to me.

fix is already done (gonna create new PR very soon)

PR #4733, opened 2022-01-21
Add extension API for registering custom category views

### LINKED ISSUE #4685: Add support for custom category views

An extension API for registering custom category views. Once registered all the views for a specific category should be displayed only on the category view (not "Browse") and should be displayed in line.

The API should have a `priority` field (for sort order).

The API should just have a component taking no props each of which should be rendered.

Nokel81:

fixes #4685 

- [x] Write tests

### DIFF
--- a/src/renderer/components/+catalog/__tests__/custom-views.test.ts
+++ b/src/renderer/components/+catalog/__tests__/custom-views.test.ts
@@ -1,0 +1,96 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import type { ConfigurableDependencyInjectionContainer } from "@ogre-tools/injectable";
+import { computed } from "mobx";
+import type React from "react";
+import type { LensRendererExtension } from "../../../../extensions/lens-renderer-extension";
+import rendererExtensionsInjectable from "../../../../extensions/renderer-extensions.injectable";
+import { getDiForUnitTesting } from "../../../getDiForUnitTesting";
+import type { CustomCategoryViewRegistration } from "../custom-views";
+import customCategoryViewsInjectable from "../custom-views.injectable";
+
+describe("Custom Category Views", () => {
+  let di: ConfigurableDependencyInjectionContainer;
+
+  beforeEach(() => {
+    di = getDiForUnitTesting();
+  });
+
+  it("should order items correctly over all extensions", () => {
+    const component1 = (): React.ReactElement => null;
+    const component2 = (): React.ReactElement => null;
+
+    di.override(rendererExtensionsInjectable, () => computed(() => [
+      {
+        customCategoryViews: [
+          {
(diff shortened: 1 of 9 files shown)


jweak (approved):
LGTM

jim-docker:
Looks weird when the view has priority higher than 50. It is cropped at the bottom. You have to scroll to see the rest of the content:

https://user-images.githubusercontent.com/40840436/150596484-fa799555-0fd4-451c-a0d1-58e66c54b774.mov

Can the extension dev control the size of the list?

Nokel81:
> Can the extension dev control the size of the list?

No, I have already created #4735 to cover the work for making the list more reactive.

It was discussed and scrolling is the desired behaviour if enough items are registered.

src/renderer/components/+catalog/__tests__/custom-views.test.ts
```diff
@@ -0,0 +1,96 @@
...
+          } as CustomCategoryViewRegistration,
+        ],
+      },
+    ] as LensRendererExtension[]));
+
+    const customCategoryViews = di.inject(customCategoryViewsInjectable);
+    const { after } = customCategoryViews.get().get("foo").get("bar");
+
+    expect(after[0].View).toBe(component2);
+    expect(after[1].View).toBe(component1);
```
jim-docker, line 57:
```suggestion
    expect(after[0]?.View).toBe(component2);
    expect(after[1]?.View).toBe(component1);
```

src/renderer/components/+catalog/__tests__/custom-views.test.ts
```diff
@@ -0,0 +1,96 @@
...
+            priority: 95,
+          } as CustomCategoryViewRegistration,
+        ],
+      },
+    ] as LensRendererExtension[]));
+
+    const customCategoryViews = di.inject(customCategoryViewsInjectable);
+    const { before } = customCategoryViews.get().get("foo").get("bar");
+
+    expect(before[0].View).toBe(component1);
```
jim-docker, line 94:
```suggestion
    expect(before[0]?.View).toBe(component1);
```

src/renderer/components/+catalog/__tests__/custom-views.test.ts
Nokel81: I don't quite see the benefit here because the tests will still fail if those are out of bounds.

jim-docker: won't they throw an error?

Nokel81: Yes, which is what an `expect().toBe()` does as well.

jim-docker: I think when looking at the test results seeing that the expectation failed is more informative than "can't access View of undefined". But I'll approve...

jim-docker (approved):

PR #4737, opened 2022-01-24
Fix: keep ItemListLayout scrollable with custom catalog category views

### LINKED ISSUE #4735: Custom Category Views CSS
**Describe the bug**
When a custom category view is registered for a specific category and it is only a small height. The default list view should shrink to make it so that the views are scrollable always.

However, if enough views are added then the whole page should be scrollable.

Follow up to #4733

aleksfront:
When custom category views added using #4733 whole content view started to be scrollable (which is bad). Current PR fixes styles allowing list area to shrink until some minimal height while keeping it scrollable.

The problem:
unnecessary content scroll

The fix:

https://user-images.githubusercontent.com/9607060/150793422-db124c33-4de5-4b1c-aeeb-2d742b8f0f1e.mov

Fixes #4735 

**Note** It's better to set `max-height` property on any of custom views if they should have fixed height. Otherwise, they will have flex height adapting viewport one.

### DIFF
--- a/src/renderer/components/+catalog/catalog.module.scss
+++ b/src/renderer/components/+catalog/catalog.module.scss
@@ -12,6 +12,8 @@
   :global(.TableRow):hover .pinIcon {
     opacity: 1;
   }
+
+  flex-grow: 1;
 }
 
 .entityName {
@@ -127,4 +129,11 @@
 
 .catalogAvatar {
   font-size: 1.2ch;
+}
+
+.views {
+  padding: calc(var(--padding) * 2);
+  display: flex;
+  flex-direction: column;
+  height: 100%;
 }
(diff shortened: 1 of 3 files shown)


Nokel81:
> Note It's better to set max-height property on any of custom views if they should have fixed height. Otherwise, they will have flex height adapting viewport one.

Could you add this to the documentation of the guide for custom catalog views?

ixrock (approved):

src/renderer/components/+catalog/catalog.module.scss
```diff
@@ -128,3 +130,10 @@
 .catalogAvatar {
   font-size: 1.2ch;
 }
+
+.views {
+  padding: calc(var(--padding) * 2);
```
DmitriyNoa, line 135: Why multiply by 2 specifically?

aleksfront: This basically needs to get `16px` value which is equal visually to previous one (`1.5rem`).

DmitriyNoa (approved):

PR #4738, opened 2022-01-24
Fix pod logs dock tab

Nokel81:
- Move all dependencies into a transient LogsViewModel

- Remove dependencies on dockStore, and logTabStore

- Make all bindings as late as possible, as per mobx rules

This fixes an integration test regression that is pressent on #4720


src/renderer/components/dock/logs/store.ts
```diff
@@ -138,21 +135,36 @@ export class LogStore {
...
    */
-  @computed
-  get logs() {
-    return this.podLogs.get(this.dependencies.dockStore.selectedTabId) ?? [];
+  get logs(): string[] {
+    return [];
   }
 
   /**
+   * @deprecated This now only returns the empty array
```
jim-docker, line 163: either the comment is wrong or you forgot to change the method implementation?

src/renderer/components/dock/logs/get-log-tab-data.injectable.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable, lifecycleEnum } from "@ogre-tools/injectable";
+import logTabStoreInjectable from "./tab-store.injectable";
+
+const getLogTabDataInjectable = getInjectable({
+  instantiate: (di) => di.inject(logTabStoreInjectable).getData,
```
jim-docker, line 9: So the logTabStore is not directly a dependency but its methods are? (just seeing if I understand)

jim-docker (approved):

src/renderer/components/dock/logs/store.ts
Nokel81: Nope, it is correct because `this.logs` is a getter above on line 159 which all it does is `return [];`

src/renderer/components/dock/logs/get-log-tab-data.injectable.ts
Nokel81: Yes that is correct. This is one of the benefits of DI since it makes testing much easier as each of these dependencies can not be overridden individually.

You can observe this in the modified tests for `log-resource-selector`

jweak (approved):

PR #4754, opened 2022-01-25
Improve dock tabs UX

### LINKED ISSUE #4459: Improve Dock tabs UX
Currently dock tabs provide poor user experience, make it hard to work with large number of tabs. Among the problems:

* Tabs are always fixed sized despite how many of them opened
* Type icons merge with names make it harder to separate each tab from one another
* Tabs don't have visual vertical separators
* Impossible to say if tab area is scrollable or not (e. g. if more tabs opened on the right)

Browsers can be a good example of how tabs should be implemented.

Acceptance criteria:

* Selected tab should be clearly outlined from the rest
* Tabs should be switched easily with `Cmd+Tab` and other keyboard shortcuts
* Tab type should be clearly visible (make icons colorful?)
* Tabs should shrink their width to contain as much as possible in available space.
dock tabs

DmitriyNoa:
Closes: https://github.com/lensapp/lens/issues/4459

The current solution is inspired by VSCode and Webstrom tabs management solutions.

When the number of tabs is getting too high the area will be scrollable.

When the number of tabs is getting too high the scroll control are presented (scroll left, and scroll right).

Scroll controls are displayed based on the scroll position.

Tabs are now clearly separated from each other.

The active tab is highlighted and visible.

The double click handler was removed as it sometimes caused accidental close of the terminal area.

The active tab can be changed with keypress:
Next tab ctrl + .
Previous tab ctrl + ,

test2

**Some comments:**

I decided to avoid the shrinking of the tabs as they are added because it makes navigation and searches pretty difficult (titles are getting cropped and the size is too small). Looks like it's common to approach among IDEs to scroll tabs.

VSCode approach with scroll:
VSCode

Webstrom approach with arrows:
Screenshot 2022-01-25 at 16 37 40

E.g. browser scenario with too many tabs:
Screenshot 2022-01-25 at 16 31 39

**Additional suggestions to add:**

Drag and drop to change the order of the tabs?

**UPDATE**
Latest UX changes can be viewed below -> https://github.com/lensapp/lens/pull/4754#issuecomment-1068872388


src/renderer/components/dock/dock.scss
```diff
@@ -76,4 +80,47 @@
...
+    overflow: auto;
+    position: relative;
+
+    &::-webkit-scrollbar {
+      visibility: hidden;
+      height: 4px;
+      background:transparent;
+      position: absolute;
+      bottom: 0;
+    }
```
ixrock: Maybe we should use `@include hidden-scrollbar` from common `mixins.scss`?

ixrock (dismissed):
LGTM

src/renderer/components/dock/dock.scss
DmitriyNoa: Tried importing hidden-scrollbar but unfortunately, that hides scroll bars completely. We need custom implementation here.

ixrock: That was the point (to hide it completely) imho.. Scrolling still should work + if we indicating it with arrows on left/right sides when available.

DmitriyNoa: My idea was also to show it when a user hovers over the tabs, so they can scroll quickly by grabbing the handle, the same way it's done in VSCode.

src/renderer/components/dock/dock-tabs.tsx
```diff
@@ -93,17 +100,18 @@ export const DockTabs = ({ tabs, autoFocus, selectedTab, onChangeTab }: Props) =
...
+    updateStateValues();
 
     // update values in store on scroll
     elem.current.addEventListener("scroll", updateScrollPosition);
 
     // update current values on resize to show/hide scroll
     window.addEventListener("resize", onWindowResize);
+
+    // update scroll state if tabs numbers has changed
+    dockStore.onTabsNumberChange(() => {
```
ixrock: I guess `useEffect` should return cleanup function for unsubscribing for scroll/resize/tabs-change events.

src/renderer/components/dock/dock-store/dock.store.ts
```diff
@@ -224,6 +233,15 @@ export class DockStore implements DockStorageState {
     }), reactionOpts);
   }
 
+  onTabsNumberChange(callback: (evt: DockTabsNumberEvent) => void) {
+    return reaction(() => this.tabsNumber, ((currentTabsNumber, previousTabsNumber) => {
```
ixrock: This reaction is quite simple and I don't see any benefits of abstraction unless `tabsNumber` is `private` or `protected` field.

DmitriyNoa: Makes sense. Cleaned up.

src/renderer/components/dock/dock-tabs.tsx
```diff
@@ -109,9 +110,12 @@ export const DockTabs = ({ tabs, autoFocus, selectedTab, onChangeTab, dockStore
     window.addEventListener("resize", onWindowResize);
 
     // update scroll state if tabs numbers has changed
-    dockStore.onTabsNumberChange(() => {
-      updateStateValues();
-    });
+    reaction(() => dockStore.tabsNumber, updateStateValues);
```
ixrock: Mobx reaction returns `IReactionDisposer` (function), so it must be also included inside returning callback.

src/renderer/components/dock/dock-store/dock.store.ts
```diff
@@ -97,6 +97,11 @@ export interface DockTabCloseEvent {
   tabId: TabId; // closed tab id
 }
 
+export interface DockTabsNumberEvent {
```
ixrock: Some leftovers from removed reaction event? Or is  it used somewhere? 🧐

src/renderer/components/dock/dock-tabs.tsx
```diff
@@ -39,14 +99,56 @@ export const DockTabs = ({ tabs, autoFocus, selectedTab, onChangeTab }: Props) =
     }
   };
 
+
+  useEffect(() => {
+    updateStateValues();
```
ixrock: Instead of this you could also add `{fireImmediately: true}` for your reaction (to apply initial values + watch changes at the same time)

src/renderer/components/dock/dock.tsx
```diff
@@ -64,6 +70,14 @@ class NonInjectedDock extends React.Component<Props & Dependencies> {
       closeTab(selectedTab.id);
       this.element?.current.focus();  // Avoid loosing focus when closing tab
     }
+
+    if(ctrlKey && code === "Period") {
+      this.nextTab(Direction.next);
+    }
+
+    if(ctrlKey && code === "Comma") {
+      this.nextTab(Direction.prev);
```
ixrock: _nit:_ probably having `Direction` as enum only for that purpose is redundant and something like `this.nextTab(-1)` should be enough.

ixrock (dismissed):

src/renderer/components/dock/dock-store/dock.store.ts
DmitriyNoa: Good catch

src/renderer/components/dock/dock-tabs.tsx
```diff
@@ -39,14 +99,54 @@ export const DockTabs = ({ tabs, autoFocus, selectedTab, onChangeTab }: Props) =
...
+
+  useEffect(() => {
+    // update values in store on scroll
+    elem.current.addEventListener("scroll", updateScrollPosition);
+
+    // update current values on resize to show/hide scroll
+    window.addEventListener("resize", onWindowResize);
+
+    // update scroll state if tabs numbers has changed
+    const tabsNumberChangeDisposer = reaction(() => dockStore.tabsNumber, updateStateValues, { fireImmediately: true });
```
Nokel81: This doesn't need to be from `tabsNumber`. It could just be a `useEffect` that is dependent on `tabs.length`

Nokel81: That way you don't need to make `dockStore` a prop for this component.

src/renderer/components/dock/dock.tsx
```diff
@@ -37,6 +37,8 @@ interface Dependencies {
   dockStore: DockStore
 }
 
+type Direction = 1 | -1;
```
Nokel81: nit: would prefer if this was named:

```
enum Direction {
  RIGHT = 1,
  LEFT = -1,
```

src/renderer/components/dock/dock.tsx
```diff
@@ -75,6 +86,22 @@ class NonInjectedDock extends React.Component<Props & Dependencies> {
     this.element?.current.focus();
   };
 
+  nextTab = (direction: Direction = 1) => {
```
Nokel81: I don't think this parameter should be defaulted.

src/renderer/components/dock/dock.tsx
```diff
@@ -75,6 +86,22 @@ class NonInjectedDock extends React.Component<Props & Dependencies> {
...
+    const { tabs, selectedTab } = this.props.dockStore;
+    const currentIndex = tabs.indexOf(selectedTab);
+    const nextIndex = currentIndex + direction;
+
+    // check if moving to the next tab is possible.
+    if (direction === 1 && currentIndex!== -1 && nextIndex >= tabs.length) return;
+
+    // check if moving to the previous tab is possible
+    if (direction === -1 && currentIndex!== -1 && nextIndex < 0) return;
```
Nokel81: Wouldn't either of these checks be sufficiant, that way they aren't dependent on the argument.

```suggestion
if (currentIndex >= 0 && !(0 <= nextIndex && nextIndex < tabs.length)) {
  return;
}
```

src/renderer/components/dock/dock.tsx
Nokel81: nit: I also think this function should have a better name, something like `switchToNextTab`

src/renderer/components/dock/dock.tsx
DmitriyNoa: This was enum before, we decided to remove it.

Nokel81: Oh strange okay...

I find enums more readable because they give names to the special values.

src/renderer/components/dock/dock-tabs.tsx
DmitriyNoa: Makes sense. Changing

src/renderer/components/dock/dock.tsx
DmitriyNoa: Changed

src/renderer/components/dock/dock.tsx
DmitriyNoa: I moved to enum back and also simplifies the solution a bit

src/renderer/components/dock/dock.tsx
DmitriyNoa: Made the check simpler.

Nokel81 (dismissed):
Looks good, please do not merge until after 5.4

ixrock (dismissed):

Nokel81 (dismissed):

nevalla:
Is it possible to make tab separator be 100% height? Now there's a small gap.

aleksfront:
Converting this to draft since we're adding more UI/UX changes as discussed with @DmitriyNoa

aleksfront:
Updated dock tabs behavior includes:
1. Shrinking tabs to some extent to show maximum info available (this is common behavior across code editors).
2. Separating tabs with vertical borders
3. Adding visual left/right "shadows" to indicate if area is scrollable
4. Hide "close" button until tab is hovered/active (again, to save some more space for a label)
5. Updated material icons font allowing to use consistent icons for tabs
6. Mouse wheel can be used now for scrolling

Also, removed left/right arrow buttons introduced in first iteration because:
1. They take room which may be used for tabs.
2. There are some caveats appeared and tuning them could take time.

Before:

https://user-images.githubusercontent.com/9607060/158550064-dabc3cf2-bead-4c14-bfc9-5588f417e97b.mov

After:


https://user-images.githubusercontent.com/9607060/158550398-47adb781-ec0a-4cf3-a1f2-ae6b8469cbfb.mov

tabs light 2
tabs light 1

src/renderer/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,26 @@
...
+
+      observer.observe(element);
+
+      return () => {
+        observer.disconnect();
+      };
+    }
+
+    return undefined;
+  }, [callback]);
```
Nokel81: shouldn't this also be dependent on `element` instance changing?

src/renderer/components/dock/dock-tabs.tsx
```diff
@@ -20,6 +22,14 @@ export interface DockTabsProps {
 }
 
 export const DockTabs = ({ tabs, autoFocus, selectedTab, onChangeTab }: DockTabsProps) => {
+  const elem = useRef<HTMLDivElement>();
+  const minTabSize = useRef<number>(0);
```
Nokel81, line 26: Why is this a ref and not `state`?

src/renderer/components/dock/dock-tabs.tsx
```diff
@@ -20,6 +22,14 @@ export interface DockTabsProps {
...
+  const [showScrollbar, setShowScrollbar] = useState<boolean>(false);
```
Nokel81: nit: there shouldn't be a need to specify these type params, they should be inferred by the default value.

src/renderer/components/dock/dock-tabs.tsx
```diff
@@ -31,22 +41,61 @@ export const DockTabs = ({ tabs, autoFocus, selectedTab, onChangeTab }: DockTabs
...
+
+  useEffect(() => {
+    const cssVars = cssVar(elem.current);
+
+    minTabSize.current = cssVars.get("--min-tab-width").valueOf();
+  });
+
+  useEffect(() => {
+    updateScrollbarVisibility();
+  }, [tabs]);
```
Nokel81: Is this enough? I thought that react dependencies were "shallow" so this would only update if the tabs parameter changed instances, what happens if the instance stays the same and just new tabs are pushed to it?

Nokel81 (commented):
Looks great, though I have found some oddities:

1. The new tab button is oddly moved to the right when there are no tabs:
Screen Shot 2022-03-16 at 8 57 11 AM
2. The close button covers text of the current tab:
Screen Shot 2022-03-16 at 8 57 26 AM
3. The current tab should be given more space so that you can still read it, even if the other tabs need to be shrunk:
Screen Shot 2022-03-16 at 8 58 06 AM

jim-docker:
I'm seeing some odd behaviour, sometimes when I click on a tab (one in the middle, not clipped) the tabs shift but the tab I clicked on is not selected.

https://user-images.githubusercontent.com/40840436/158628572-4264c588-5fab-41bf-a41e-4e0f2facc6e6.mov

Seems to only happen the first time you try to select a tab this way. I had to restart the app to reproduce.

jim-docker:
Seems to open (and scroll if necessary) to the last tab when you close a tab:

https://user-images.githubusercontent.com/40840436/158630973-2973df96-49d7-4df6-b59f-acff3e68dc2d.mov

Better to open a neighbouring tab?

src/renderer/components/dock/dock-tabs.tsx
aleksfront: Updating it only once and basically using it as instance variable - keeping across rerenders. https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables

src/renderer/components/dock/dock-tabs.tsx
aleksfront: As it turned out, there's not need for this part at all. Resize event triggered when tabs added or removed.

aleksfront:
@Nokel81 @jim-docker Fixed all addressed issues except one described below. Please review again.

https://user-images.githubusercontent.com/9607060/158988852-f9b4a4c9-82ac-414c-9986-aa260ecbaa59.mov

single plus button
single tab
tabs as a whole

aleksfront:
Related this one
> The current tab should be given more space so that you can still read it, even if the other tabs need to be shrunk: 

I don't think it's a good decision since this will lead us to layout jumps since switching active tab will cause all sibling tabs to change their sizes instantly. Also, can't find similar behavior across other apps.

Nokel81:
Okay I just know that in vscode the default tab size is such that for most file names it doesn't happen.

Nokel81 (dismissed):

DmitriyNoa:
LGTM

src/renderer/components/dock/dock-tabs.tsx
```diff
@@ -31,22 +41,57 @@ export const DockTabs = ({ tabs, autoFocus, selectedTab, onChangeTab }: DockTabs
...
   };
 
+  const scrollActiveTabIntoView = () => {
+    const tab = elem.current?.querySelector(".Tab.active");
+
+    tab?.scrollIntoView();
+  };
+
+  const updateScrollbarVisibility = () => {
+    const allTabsShrinked = getTabElements().every(tab => tab.offsetWidth == minTabSize.current);
```
jim-docker: nit: 
```suggestion
    const allTabsShrunk = getTabElements().every(tab => tab.offsetWidth == minTabSize.current);
```

src/renderer/components/dock/dock.tsx
```diff
@@ -75,6 +87,19 @@ class NonInjectedDock extends React.Component<DockProps & Dependencies> {
     this.element?.current.focus();
   };
 
+  switchToNextTab = (direction: Direction = Direction.NEXT) => {
+    const { tabs, selectedTab } = this.props.dockStore;
+    const currentIndex = tabs.indexOf(selectedTab);
+    const nextIndex = currentIndex + direction;
+
+    // check if moving to the next or previous tab is possible.
+    if (currentIndex!== -1 && (nextIndex >= tabs.length || nextIndex < 0)) return;
```
jim-docker: Should also return if `currentIndex === -1`?

```suggestion
    if (currentIndex === -1 || (nextIndex >= tabs.length || nextIndex < 0)) return;
```

jim-docker (commented):
Got this when I clicked "Cancel" on an upgrade chart tab (doesn't happen on 5.4.3):

Screen Shot 2022-03-18 at 2 12 45 PM

aleksfront:
@jim-docker Created an issue with the bug you found. Happens in `master` too. https://github.com/lensapp/lens/issues/5050

Nokel81 (approved):

jim-docker (approved):

PR #4755, opened 2022-01-25
Upgrade to latest for all linting deps

Nokel81:

I have also condenced the eslint rules to join together the rules for `.tsx` and `.ts` files since they should always have the same rules and `jsx: true` does nothing according to the eslint documentation for `.tsx` and `.jsx` files anyway.

I have also changed all the `.d.ts` files into `.ts` files as vscode doesn't do auto complete within `.d.ts` files.

This PR removes some mobx decorators, namely when they were used on function fields as the indent eslint rule is broken for function fields with annotations. They were replaced by the higher order function calls instead.


src/renderer/components/icon/icon.tsx
```diff
@@ -59,10 +60,7 @@ export class Icon extends React.PureComponent<IconProps> {
 
         // fallthrough
       case "Enter": {
-        // eslint-disable-next-line react/no-find-dom-node
-        const icon = findDOMNode(this) as HTMLElement;
-
-        setTimeout(() => icon.click());
+        setImmediate(() => this.ref?.click());
```
aleksfront: We should omit `setImmediate`. It works without it and `setImmediate` doesn't have enough browser support https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate

src/renderer/components/icon/icon.tsx
```diff
@@ -31,6 +30,8 @@ export interface IconProps extends React.HTMLAttributes<any>, TooltipDecoratorPr
 
 @withTooltip
 export class Icon extends React.PureComponent<IconProps> {
+  public ref?: HTMLElement;
```
aleksfront: We can use `React.creareRef` here https://reactjs.org/docs/refs-and-the-dom.html

Hovewer, I don't see any behavior difference.

aleksfront:
Also, "removing mobx decorations" should be mentioned in description.

src/renderer/components/icon/icon.tsx
Nokel81: I'll use createRef

Nokel81:
@aleksfront PTALA

aleksfront (approved):
LGTM

src/extensions/extensions-store/extensions-store.ts
```diff
@@ -42,10 +42,9 @@ export class ExtensionsStore extends BaseStore<LensExtensionsStoreModel> {
...
   }
 
-  @action
-  mergeState = (extensionsState: Record<LensExtensionId, LensExtensionState>) => {
+  mergeState = action((extensionsState: Record<LensExtensionId, LensExtensionState>) => {
     this.state.merge(extensionsState);
-  };
+  });
 
   @action
```
jim-docker, line 49: Is it weird now to to use decorator sometimes? Guess it depends on context

src/renderer/components/+workloads-pods/pod-details-secrets.tsx
```diff
@@ -20,19 +20,22 @@ interface Props {
...
+            namespace: pod.getNs(),
+          })),
+        );
 
-    secrets.forEach(secret => secret && this.secrets.set(secret.getName(), secret));
-  });
+        secrets.forEach(secret => secret && this.secrets.set(secret.getName(), secret));
+      }),
+    ]);
+  }
```
jim-docker, line 38: Isn't it acceptable to just add the indentation that linting demands, i.e.
```
  @disposeOnUnmount
    secretsLoader = autorun(async () => {
      const { pod } = this.props;

      const secrets = await Promise.all(
        pod.getSecrets().map(secretName => secretsApi.get({
          name: secretName,
          namespace: pod.getNs(),
        })),
      );

      secrets.forEach(secret => secret && this.secrets.set(secret.getName(), secret));
    });
```

It feels odd to be able to use decorators in some instances but not in others. 

Is this just a bug with eslint?

yarn.lock
```diff
@@ -12,13 +12,6 @@
   resolved "https://registry.yarnpkg.com/@async-fn/jest/-/jest-1.5.3.tgz#42be6c0e8ba5ccd737e006ca600e7e319fe2a591"
   integrity sha512-iQ9gAPZFW5U6TNcFS99ffwYYsB9LNecTnvG73BaDc/zAD0qOWctY1imEACC1pLymmm/xaf/OUq9I9QenfkatTA==
```
jim-docker, line 14: is this large diff expected?

jim-docker (approved):

src/extensions/extensions-store/extensions-store.ts
Nokel81: The decorators are useful for methods as opposed to field functions

yarn.lock
Nokel81: Yes because I updated dependencies

PR #4761, opened 2022-01-27
Fix: remove catalog first column padding

aleksfront:
Left padding on `.entityName` column shifted all other table head columns.

Before:
shifted columns

After:
fixed columns

It's okay to have `Name` column above avatars.


chenhunghan (approved):

ixrock (approved):
> It's okay to have Name column above avatars.

lol. i remember totally opposite opinion of someone about this at some point, but everything changes :D

PR #4776, opened 2022-01-31
Fix: increase catalog entity name col width

aleksfront:
1. Increase catalog entity name column width to see more of it (temporary solution before we have table with adjustable columns).
2. Fixes catalog navigation paddings. Before:
broken paddings
3. Adjust table border.

bigger name column

Fixes https://github.com/lensapp/lenscloud-lens-extension/issues/1235


ixrock (approved):

jweak (approved):
LGTM

Issue #4589, opened 2021-12-17
Release rollback not updated in Releases list


jim-docker:
When applying a rollback from the Apps -> Releases tab the release in the item-list-layout is not updated

https://user-images.githubusercontent.com/40840436/146572849-69bcb836-3d85-4ab7-89ad-6abb6fa8e946.mov

5.3.3 (linux)

jansav:
UI flow is:

```typescript
await listReleases(...); // current revision = 5 

await rollbackRelease(...); // to revision -> 2

await listReleases(...); // current revision = 5, should be 6

// Note: Just to verify that listReleases eventually return right revision
setTimeout(() => {
  await listReleases(...); // current revision = 6
}, 1500)

```

Back-end issue?

jansav:
> Back-end issue?

Confirmed. Working on fix now.

jansav:
Fixed by #4785

PR #4785, opened 2022-02-01
Fix bugs in helm releases

jansav:
Bugs:

1. Helm releases not updating after upgrade
2. Helm releases not updating after rollback #4589 
3. Rollback to specific revision
4. Opening of upgrade release tab

### DIFF
--- a/src/main/helm/helm-release-manager.ts
+++ b/src/main/helm/helm-release-manager.ts
@@ -182,6 +182,7 @@ export async function rollback(name: string, namespace: string, revision: number
   return JSON.parse(await execHelm([
     "rollback",
     name,
+    `${revision}`,
     "--namespace", namespace,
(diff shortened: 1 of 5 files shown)


Iku-turso (approved):

Nokel81 (approved):

src/main/helm/helm-release-manager.ts
```diff
@@ -179,12 +179,13 @@ export async function getHistory(name: string, namespace: string, kubeconfigPath
 }
 
 export async function rollback(name: string, namespace: string, revision: number, kubeconfigPath: string) {
-  return JSON.parse(await execHelm([
+  await execHelm([
```
jansav, line 182: Note: Result is not used. And actually it has never worked. Since `execHelm` returns `string` and `JSON.parse("some string")` throws

PR #4788, opened 2022-02-01
Fix extensions not being able to be installed in some cases

### LINKED ISSUE #4766: Extension Installation Issue
Lens gets stuck on the extension installation 
While trying to install the extension installation lens stucks after pasting the .tgz file or @mirantis/lens-extension-cc 

I had to go to Lens status bar icon and quit app and then restart the app and the extension was there. This is missing from the installation guide.

Environment
- Lens Version: 5.3.4
- OS: Windows 64 
- Installation method (e.g. snap or AppImage in Linux):  npm pack @mirantis/lens-extension-cc

Nokel81:
To make testing easier, I took the opportunity to factor out and slightly change how ipc handling is overridden (including the ipc bridge).

I also moved around the ipc injectables to more suitable locations since for some reason they were deep within `app-paths` 

---

The fix within this PR has been moved to https://github.com/lensapp/lens/pull/4794. This PR should be reviewed after 5.4 is released.


src/renderer/components/+extensions/attempt-install/attempt-install.tsx
```diff
@@ -130,5 +101,9 @@ export const attemptInstall =
             onClose: dispose,
           },
         );
+      } else {
+        // Remove the old dir because it isn't a valid extension anyway
+        await removeDir(extensionFolder);
+        await unpackExtension(validatedRequest, dispose);
```
Nokel81: This is the main substance of the bug fix. Namely, instead of assuming that the folder existing is the same as the extension being installed, we check if there is an installed extension and if not we delete all the old data just in case.

jim-docker: Can this be 1 PR that fixes the bug for 5.4 and the rest be another PR for 5.5?

Nokel81: Sure can do, will do that tomorrow

jim-docker:
I uninstalled the lenscloud extension (which was installed in ~/.k8slens/extensions) and then installed it from tgz file I had downloaded. It installed and worked but when I returned to the extensions page the Install button was still showing busy.

https://user-images.githubusercontent.com/40840436/152064036-1b29d949-8927-4608-9bfc-ed54e934fa4a.mov

Nokel81:
Note: Jim installed the extension via clicking on the folder icon.

Nokel81:
Closing this as the problem has already been solved and other PRs will so some cleanup here as well.

PR #4800, opened 2022-02-02
Use system-node-critical for the node shell pod

### LINKED ISSUE #4765: node shell should have system-cluster-critical priorityclass
similar to: https://github.com/lensapp/lens/issues/4751

hard to debug a node where it won't fit

jakolehm:
Fixes #4765

### DIFF
--- a/src/main/shell-session/node-shell-session/node-shell-session.ts
+++ b/src/main/shell-session/node-shell-session/node-shell-session.ts
@@ -94,6 +94,7 @@ export class NodeShellSession extends ShellSession {
           tolerations: [{
             operator: "Exists",
           }],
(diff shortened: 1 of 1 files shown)


Nokel81 (approved):

jim-docker (approved):
Can this cause a node shell to fail on some clusters (under specific RBAC say) where it otherwise wouldn't have failed?

jakolehm:
> Can this cause a node shell to fail on some clusters (under specific RBAC say) where it otherwise wouldn't have failed?

Maybe if a cluster has custom admission controller that denies this specific thing. For majority of users this should work better.

PR #4802, opened 2022-02-03
ShellEnv extension api

### LINKED ISSUE #4596: Allow to set terminal environment variables via extension api

To set environment variables (per cluster) for the terminal instances within Lens (Dock) via extension api.

**Why is this needed**:

Some extensions need to configure cli tools.

jim-docker:
In the extension:

```
export default class ExampleMainExtension extends Main.LensExtension {
  .
  .
  .
  terminalShellEnvModifier = (ctx: ShellEnvContext, env: Record<string, string>) => {
    const { catalogEntity: entity } = ctx;

    if (entity.kind !== "KubernetesCluster" ) {
      return env;
    }

    return ({
      ...env, 
      MYCONTEXT: entity.spec.kubeconfigContext,
      MYKUBECONFIGPATH: entity.spec.kubeconfigPath,
      MYID: entity.getId(),
      MYKIND: entity.kind,
    });
  }

  .
  .
  .
}
```

result:

Screen Shot 2022-02-03 at 6 24 00 PM


src/main/shell-session/shell-session.ts
```diff
@@ -358,6 +361,8 @@ export abstract class ShellSession {
       .filter(Boolean)
       .join();
 
+    this.shellEnvModifiers.map(modifier => env = { ...env, ...modifier({ context: this.cluster.contextName,  kubeconfigPath, id: this.cluster.id }, env) });
```
jim-docker: Probably not what we want to do if multiple extensions modify the env. Not sure what we should do. Allow user to enable/disable some? Select the order they are applied?

src/main/shell-session/shell-env-modifier/shell-env-modifier-registration.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import type { noop } from "../../../common/utils";
+
+export interface ShellEnvClusterData {
+  context: string;
+  kubeconfigPath: string;
+  id: string;
```
jim-docker: would it be enough to just provide the cluster id and let the extension get other things it wants from there?

src/main/shell-session/shell-env-modifier/shell-env-modifier-registration.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import type { noop } from "../../../common/utils";
+
+export interface ShellEnvClusterData {
```
jakolehm: For future proof reasons this could be just `ShellEnvContext` (terminal can be also in other contexts within Lens). It would have just one property, CatalogEntity.

src/extensions/lens-main-extension.ts
```diff
@@ -10,10 +10,15 @@ import type { CatalogEntity } from "../common/catalog";
...
 import type { MenuRegistration } from "../main/menu/menu-registration";
 import type { TrayMenuRegistration } from "../main/tray/tray-menu-registration";
+import type { ShellEnvModifier } from "../main/shell-session/shell-env-modifier/shell-env-modifier-registration";
+import { noop } from "../common/utils";
+
 export class LensMainExtension extends LensExtension {
   appMenus: MenuRegistration[] = [];
   trayMenus: TrayMenuRegistration[] = [];
 
+  shellEnvModifier: ShellEnvModifier = noop;
```
Nokel81: I don't think we want `noop` here, it would be better to have this be an optional field.

src/main/shell-session/shell-env-modifier/shell-env-modifier-registration.ts
```diff
@@ -0,0 +1,14 @@
...
+  context: string;
+  kubeconfigPath: string;
+  id: string;
+}
+
+export type ShellEnvModifier = ((cluster: ShellEnvClusterData, env: Record<string, string | undefined>) => Record<string, string | undefined>) | typeof noop;
```
Nokel81: I think that having `typeof noop` is wrong here, I would just make the field in `LensMainExtension` be optional.

src/main/shell-session/shell-env-modifier/shell-env-modifier-registration.ts
Nokel81: That is probably a good idea because the other info should be present if it is a KubernetesCluster

src/main/shell-session/shell-env-modifier/shell-env-modifiers.ts
```diff
@@ -0,0 +1,18 @@
...
+import type { LensMainExtension } from "../../../extensions/lens-main-extension";
+ 
+interface Dependencies {
+  extensions: IComputedValue<LensMainExtension[]>;
+}
+ 
+export const shellEnvModifiers = ({ extensions }: Dependencies) => {
+  return computed(() => (
+    extensions.get()
+      .flatMap((extension) => extension.shellEnvModifier)
```
Nokel81: And thus a `.filter(Boolean)` here too. Also this can be just `.map` because `shellEnvModifier` is not an array.

src/main/shell-session/shell-session.ts
Nokel81: This should be a reduce call instead. I think leaving the order be undefined is fine.

Something like:
```
const env = clearKubeconfigEnvVars(JSON.parse(JSON.stringify(await shellEnv())));
...
const ctx = { ... };
const modifiedEnv = this.shellEnvModifiers.reduce((env, modifier) => modifier(ctx, env), env);
```

jim-docker: >  I think leaving the order be undefined is fine.

Really? The order can vary from time to time, right (across restarts)? Extensions might clobber each other's vars/values.

jim-docker: I also wonder if it's a good idea to offer the whole default shell env to the extension, to be freely modified.

Nokel81: They might, but if we declare that the order is unspecified then the extensions need to be defensive in their naming.

Plus I don't see a way to do it otherwise, since we need an order and some extension has to be first.

jim-docker: That is true, especially if this is meant to be seamless for the user. Otherwise, we could offer the user a list of Shell Envs in the dock UI and let them decide how they are applied when opening a terminal. This of course assumes the user knows what they're doing.

Nokel81: That is true (about users). I think we should start with the order unspecified and if absolutely necessary add in some mechanism in the future.

src/main/shell-session/node-shell-session/node-shell-session.ts
```diff
@@ -25,7 +25,7 @@ export class NodeShellSession extends ShellSession {
   protected readonly cwd: string | undefined = undefined;
 
   constructor(protected nodeName: string, kubectl: Kubectl, socket: WebSocket, cluster: Cluster, terminalId: string) {
-    super(kubectl, socket, cluster, terminalId);
+    super(kubectl, [], socket, cluster, terminalId);
```
jim-docker: Is this correct, that we don't offer modifying the shell env for node shells?

Nokel81: I think we should allow it because an extension might need to set some things up for the node to work maybe. but we should pass into the modifiers the type of shell too.

jim-docker: This would let an extension set env vars in the terminal that Lens uses solely to launch the `kubectl exec` into the node pod. The terminal and the pod are under Lens's control and are just a means to a shell on the node. Is there something that can be added by an extension to the terminal env that could  positively affect the behaviour of the `kubectl exec` to get the shell?

Being able to set env vars within the node shell might be useful? @jakolehm?

Also, instead of another parameter to the modifiers we could add another method to the main lens extension (i.e. `nodeShellEnvModifier()`) which is of the same modifier type

jakolehm: I would focus on "local" terminal envs in this PR. We just need to design the api so that we can introduce node shell envs later (if needed) without breaking the existing api.

jim-docker: Ok. As for the terminal that Lens uses solely to launch the `kubectl exec` into the node pod, I don't see a reason to let an extension set its env, which is how this PR is written. Does that make sense?

jakolehm (dismissed):
LGTM

src/main/shell-session/local-shell-session/local-shell-session.ts
```diff
@@ -20,7 +29,14 @@ export class LocalShellSession extends ShellSession {
   }
 
   public async open() {
-    const env = await this.getCachedShellEnv();
+    let env = await this.getCachedShellEnv();
+  
+    if (this.entity) {
+      const ctx = { catalogEntity: this.entity };
+
+      env = JSON.parse(JSON.stringify(this.shellEnvModifiers.reduce((env, modifier) => modifier(ctx, env), env)));
```
Nokel81: I meant this actually:

```suggestion
      env = this.shellEnvModifiers.reduce((env, modifier) => modifier(ctx, env), JSON.parse(JSON.stringify(env)));
```

src/main/shell-session/shell-session.ts
```diff
@@ -298,17 +294,16 @@ export abstract class ShellSession {
 
   protected async getCachedShellEnv() {
     const { id: clusterId } = this.cluster;
-    const { ShellType } = this;
 
-    let env = ShellSession.shellEnvs.get({ clusterId, ShellType });
+    let env = ShellSession.shellEnvs.get(clusterId);
```
Nokel81: Changing the caching in a different PR?

jim-docker: hmm, no, maybe I missed saving a file? will fix

src/main/shell-session/local-shell-session/local-shell-session.ts
jim-docker: right!

src/main/shell-session/shell-session.ts
jim-docker: that file is now untouched by this PR, are you looking at all commits?

Nokel81: I was not, I meant the change to only having temporal validity as the caching technique. Which could very well be its own PR.

jim-docker: right, definitely its own PR. If we even need it? It doesn't impact the ShellEnvModifiers now

src/extensions/lens-main-extension.ts
```diff
@@ -10,10 +10,14 @@ import type { CatalogEntity } from "../common/catalog";
 import type { IObservableArray } from "mobx";
 import type { MenuRegistration } from "../main/menu/menu-registration";
 import type { TrayMenuRegistration } from "../main/tray/tray-menu-registration";
+import type { ShellEnvModifier } from "../main/shell-session/shell-env-modifier/shell-env-modifier-registration";
+
 export class LensMainExtension extends LensExtension {
   appMenus: MenuRegistration[] = [];
   trayMenus: TrayMenuRegistration[] = [];
 
+  terminalShellEnvModifier?: ShellEnvModifier;
```
Nokel81, line 32: It would be great if we could get some documentation this. Could be a follow up PR.

src/main/shell-session/local-shell-session/local-shell-session.injectable.ts
```diff
@@ -17,10 +19,12 @@ interface InstantiationParameter {
 const localShellSessionInjectable = getInjectable({
   instantiate: (di, { cluster, tabId, webSocket }: InstantiationParameter) => {
     const createKubectl = di.inject(createKubectlInjectable);
+    const localShellEnvModifiers = di.inject(terminalShellEnvModifiersInjectable);
 
     const kubectl = createKubectl(cluster.version);
+    const entity = catalogEntityRegistry.getById(cluster.id);
 
-    return new LocalShellSession(kubectl, webSocket, cluster, tabId);
+    return new LocalShellSession(localShellEnvModifiers.get(), entity, kubectl, webSocket, cluster, tabId);
```
Nokel81: It would be even better if `terminalShellEnvModifiersInjectable` returned a function of the signature `(clusterId: ClusterId, env: Record<string, string>) => Record<string, string>`.

That way this code doesn't need to know about the type `ShellEnvContext`.

src/extensions/lens-main-extension.ts
jim-docker: Is this the place to put the documentation?

Nokel81: It is one of the two places to do so. Any doc comments (of the form `/** * */`) above items such as fields, methods, functions, types, ... will be rendered as part of https://api-docs.k8slens.dev/v5.3.4/extensions/api/

The other option is of course guides.

src/main/shell-session/local-shell-session/local-shell-session.injectable.ts
jim-docker: Where does it need to know about `ShellEnvContext`? Do you mean `catalogEntityRegistry`?

Nokel81: Yes I do mean `catalogEntityRegistry`. Also, it would make the use of it within `open` simpler since we could just call it unconditionally.

src/extensions/lens-main-extension.ts
jim-docker: None of the other fields are documented here, so I wondered if they're documented elsewhere, like where they're (or the lens main extension is) exported to the extension api

Nokel81: They are not documented at all, which is sort of a problem. I mentioned it because I think we should try to improve the state here but I don't think it should block this PR.

src/main/shell-session/local-shell-session/local-shell-session.injectable.ts
jim-docker: I pushed another commit. Is it what you were thinking here?

jim-docker:
This implementation allows an extension to completely overwrite our default terminal shell environment. Is that ok? Also, would there be any security concerns with an extension setting/overwriting env vars for the terminal?

Also, are there any other contextual details we might want to provide to the extension, apart from the catalog entity? Dock terminal information? General Lens information?

jakolehm (approved):

Nokel81 (approved):
Yes this is what I was basically thinking

PR #4805, opened 2022-02-03
Don't block other extensions from loading if one activation fails

### LINKED ISSUE #4804: Extension activation error blocks other extensions
If one extension has an error in the `onActivate` method, this will block all other extensions from loading. Instead we should load the other extensions and reject only the failed extension from being loaded.

panuhorsmalahti:
* Fixes https://github.com/lensapp/lens/issues/4804
* Adjust error message to be correct (error was with loading, not activation)
* The issue was that Promise.all would throw if just one extension's `onActivate` method would throw.

### DIFF
--- a/src/extensions/extension-loader/extension-loader.ts
+++ b/src/extensions/extension-loader/extension-loader.ts
@@ -349,7 +349,6 @@ export class ExtensionLoader {
         extension.activated.catch((error) => {
           extension.activationError = error;
           logger.error(`${logModule}: activation extension error`, { ext: extension.installedExtension, error });
-          this.instances.delete(extension.extId);
         }),
       ),
     );
--- a/src/extensions/extension-loader/extension-loader.ts
+++ b/src/extensions/extension-loader/extension-loader.ts
@@ -324,11 +324,13 @@ export class ExtensionLoader {
             return {
               extId,
               instance,
+              installedExtension: extension,
               isBundled: extension.isBundled,
               activated: instance.activate(),
+              activationError: null,
             };
           } catch (err) {
-            logger.error(`${logModule}: activation extension error`, { ext: extension, err });
+            logger.error(`${logModule}: error loading extension`, { ext: extension, err });
           }
         } else if (!extension.isEnabled && alreadyInit) {
           this.removeInstance(extId);
@@ -339,21 +341,32 @@ export class ExtensionLoader {
       // Remove null values
       .filter(extension => Boolean(extension));
 
-    // We first need to wait until each extension's `onActivate` is resolved,
+    // We first need to wait until each extension's `onActivate` is resolved or rejected,
     // as this might register new catalog categories. Afterwards we can safely .enable the extension.
-    await Promise.all(extensions.map(extension => extension.activated));
+    await Promise.all(
+      extensions.map(extension =>
+        // If extension activation fails, log error
+        extension.activated.catch((error) => {
+          extension.activationError = error;
+          logger.error(`${logModule}: activation extension error`, { ext: extension.installedExtension, error });
+          this.instances.delete(extension.extId);
+        }),
+      ),
+    );
 
     // Return ExtensionLoading[]
-    return extensions.map(extension => {
-      const loaded = extension.instance.enable(register).catch((err) => {
-        logger.error(`${logModule}: failed to enable`, { ext: extension, err });
-      });
+    return extensions
+      .filter(extension =>!extension.activationError)
+      .map(extension => {
+        const loaded = extension.instance.enable(register).catch((err) => {
+          logger.error(`${logModule}: failed to enable`, { ext: extension, err });
+        });
 
-      return {
-        isBundled: extension.isBundled,
-        loaded,
-      };
-    });
+        return {
+          isBundled: extension.isBundled,
+          loaded,
+        };
+      });
   }
 
   protected autoInitExtensions(register: (ext: LensExtension) => Promise<Disposer[]>) {
--- a/src/extensions/extension-loader/extension-loader.ts
+++ b/src/extensions/extension-loader/extension-loader.ts
@@ -322,10 +322,8 @@ export class ExtensionLoader {
             this.instances.set(extId, instance);
 
             return {
-              extId,
               instance,
               installedExtension: extension,
-              isBundled: extension.isBundled,
               activated: instance.activate(),
             };
           } catch (err) {
@@ -358,7 +356,7 @@ export class ExtensionLoader {
       });
 
       return {
-        isBundled: extension.isBundled,
+        isBundled: extension.installedExtension.isBundled,
         loaded,
       };
     });
--- a/src/extensions/extension-loader/extension-loader.ts
+++ b/src/extensions/extension-loader/extension-loader.ts
@@ -327,7 +327,6 @@ export class ExtensionLoader {
               installedExtension: extension,
               isBundled: extension.isBundled,
               activated: instance.activate(),
-              activationError: null,
             };
           } catch (err) {
             logger.error(`${logModule}: error loading extension`, { ext: extension, err });
@@ -347,7 +346,6 @@ export class ExtensionLoader {
       extensions.map(extension =>
         // If extension activation fails, log error
         extension.activated.catch((error) => {
-          extension.activationError = error;
           logger.error(`${logModule}: activation extension error`, { ext: extension.installedExtension, error });
         }),
       ),
@@ -355,7 +353,6 @@ export class ExtensionLoader {
 
     // Return ExtensionLoading[]
     return extensions
-      .filter(extension =>!extension.activationError)
       .map(extension => {
         const loaded = extension.instance.enable(register).catch((err) => {
           logger.error(`${logModule}: failed to enable`, { ext: extension, err });
--- a/src/extensions/extension-loader/extension-loader.ts
+++ b/src/extensions/extension-loader/extension-loader.ts
@@ -352,17 +352,16 @@ export class ExtensionLoader {
     );
 
     // Return ExtensionLoading[]
-    return extensions
-      .map(extension => {
-        const loaded = extension.instance.enable(register).catch((err) => {
-          logger.error(`${logModule}: failed to enable`, { ext: extension, err });
-        });
-
-        return {
-          isBundled: extension.isBundled,
-          loaded,
-        };
+    return extensions.map(extension => {
+      const loaded = extension.instance.enable(register).catch((err) => {
+        logger.error(`${logModule}: failed to enable`, { ext: extension, err });
       });
+
+      return {
+        isBundled: extension.isBundled,
+        loaded,
+      };
+    });
   }
 
   protected autoInitExtensions(register: (ext: LensExtension) => Promise<Disposer[]>) {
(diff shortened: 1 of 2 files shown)


src/extensions/lens-extension.ts
```diff
@@ -114,7 +114,7 @@ export class LensExtension {
 
   @action
   activate() {
-    return this.onActivate();
+    return Promise.resolve(this.onActivate());
```
panuhorsmalahti: `onActive` might not return a Promise. This converts non-promises to promises.

Nokel81: Does it? If it throws I don't think a promise will be returned

panuhorsmalahti: Good point, changed this to handle sync errors

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -324,11 +324,13 @@ export class ExtensionLoader {
             return {
               extId,
               instance,
+              installedExtension: extension,
```
Nokel81, line 326: I don't think we need this, we can always just do `instance.manifest`

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -324,11 +324,13 @@ export class ExtensionLoader {
...
               isBundled: extension.isBundled,
               activated: instance.activate(),
+              activationError: null,
```
Nokel81: I don't think this is necessary here, read below

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -339,21 +341,32 @@ export class ExtensionLoader {
       // Remove null values
       .filter(extension => Boolean(extension));
 
-    // We first need to wait until each extension's `onActivate` is resolved,
+    // We first need to wait until each extension's `onActivate` is resolved or rejected,
     // as this might register new catalog categories. Afterwards we can safely .enable the extension.
-    await Promise.all(extensions.map(extension => extension.activated));
+    await Promise.all(
```
Nokel81, line 343: Given that we have already started all the activations above, why not just use a for loop and `try...catch`, then we don't need another field, or the filter, or the map. Just an array to `push` to.

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -339,21 +341,32 @@ export class ExtensionLoader {
...
+      extensions.map(extension =>
+        // If extension activation fails, log error
+        extension.activated.catch((error) => {
+          extension.activationError = error;
+          logger.error(`${logModule}: activation extension error`, { ext: extension.installedExtension, error });
+          this.instances.delete(extension.extId);
```
Nokel81: Also, why delete the instances?

src/extensions/extension-loader/extension-loader.ts
panuhorsmalahti: That's slightly different type, and has a bit less information. I think we should keep installedExtension in the error log, and that would match the original functionality too.

src/extensions/extension-loader/extension-loader.ts
panuhorsmalahti: You're right, we can keep extensions that can't be enabled in `.instances`. That mirrors the original functionality too. Fixed

src/extensions/extension-loader/extension-loader.ts
panuhorsmalahti: I checked the original code, and it seems we want to return even extensions that fail to be activated from `loadExtensions`. I removed the filter and the field to simplify it.

src/extensions/lens-extension.ts
```diff
@@ -113,7 +113,7 @@ export class LensExtension {
   }
 
   @action
```
Nokel81: This does not need this annotation, we should remove it.

src/extensions/extension-loader/extension-loader.ts
Nokel81: Maybe, but this mapping is getting rather cluttered IMO since the extension's ID and the `isBundled` fields could be removed too.

src/extensions/extension-loader/extension-loader.ts
Nokel81: I don't think we should really be mixing `await` and `.catch` here since we can easily avoid doing so.

I would suggest something like the following:

```
for (const extension of extensions) {
  try {
    await extension.activated;
  } catch (error) {
    logger.error(`${logModule}: activation extension error`, { ext: extension.installedExtension, error });
  }
}
```

Furthermore I think the use of `Promise.all` here is a case of using the wrong tool, since we don't actually care about the result. Only the errors (if there are any)

Nokel81: So looking at the code around this, we could easily make `activate` call `enable`, have the logging within `LensExtension` and then pass those promises directly to the `return extensions.map(...)` bellow, because we don't actually need this awaits here.

panuhorsmalahti: `Promise.all` is valid use here, because we want to wait until all the Promises have been resolved.

Your try / catch / await is okay.. but has a bit different semantics. It would wait until each extension has been activated, and only print the error in order (e.g. loading extensions A, B and C, if C fails, it would log error from C only after A and B are done). With Promise.all each error from activation will be logged immediately, which is slightly beneficial.

Nokel81: No it would print the errors as they come in.

Technically, we would would `allSettled`, but anyway, I don't see why we need that there in the first place

src/extensions/extension-loader/extension-loader.ts
panuhorsmalahti: Good point, removed extId and isBudled from the object.

src/extensions/lens-extension.ts
panuhorsmalahti: Removed

src/extensions/extension-loader/extension-loader.ts
panuhorsmalahti: They wouldn't print as they come, as await inside the for-of loop waits until each iteration has been completed, it doesn't run "in parallel".

Nokel81: Oh they will run concurrently because you call the `activate` function above before the loop. They just won't be reported concurrently.

panuhorsmalahti: Yeah, I was talking about the code block you suggested, not earlier / other code.

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -339,9 +338,16 @@ export class ExtensionLoader {
...
       .filter(extension => Boolean(extension));
 
-    // We first need to wait until each extension's `onActivate` is resolved,
+    // We first need to wait until each extension's `onActivate` is resolved or rejected,
     // as this might register new catalog categories. Afterwards we can safely .enable the extension.
-    await Promise.all(extensions.map(extension => extension.activated));
+    await Promise.all(
+      extensions.map(extension =>
+        // If extension activation fails, log error
+        extension.activated.catch((error) => {
```
ixrock, line 346: Should we return promise in order to make `Promise.all` working properly?

```suggestion
        return extension.activated.catch((error) => {
```

ixrock (approved):

src/extensions/extension-loader/extension-loader.ts
Nokel81: Completely agree!

src/extensions/extension-loader/extension-loader.ts
Nokel81: Yes, however you are touching this code. And I think that the current logic is very complex and could be simplified without harming our debugging experience.

src/extensions/extension-loader/extension-loader.ts
ixrock: Oh, actually this is fine already, there is no extra brackets from top `{}`.

src/extensions/extension-loader/extension-loader.ts
panuhorsmalahti: What do you suggest to change? Promise.all is fine for this use case, and I explained why using your try/catch is inferior.

The 'we could easily make activate call enable, have the logging within LensExtension and then pass those promises directly to the return extensions.map(...)' is a refactoring that I don't think is needed for this fix.

panuhorsmalahti: And in fact having activate call enable wouldn't work because of the following reason:
```
    // We first need to wait until each extension's `onActivate` is resolved,
    // as this might register new catalog categories. Afterwards we can safely .enable the extension.
```

Nokel81: I would do the following:

```ts
protected async loadExtensions(installedExtensions: Map<string, InstalledExtension>, register: (ext: LensExtension) => Promise<Disposer[]>) {
  const extensions: ExtensionLoading[] = [];

  for (const [extId, ext] of installedExtensions) {
    const alreadyInit = this.instances.has(extId) || this.nonInstancesByName.has(ext.manifest.name);

    if (ext.isCompatible && ext.isEnabled && !alreadyInit) {
      try {
        const LensExtensionClass = this.requireExtension(ext);

        if (!LensExtensionClass) {
          this.nonInstancesByName.add(ext.manifest.name);

          continue;
        }

        const instance = this.dependencies.createExtensionInstance(
          LensExtensionClass,
          ext,
        );

        this.instances.set(extId, instance);

        extensions.push({
          isBundled: ext.isBundled,
          loaded: (async (): Promise<void> => {
            try {
              await instance.activate();
            } catch (error) {
              return void logger.error(`${logModule}: extension activation error`, { ext, error });
            }

            try {
              await instance.enable(register);
            } catch (error) {
              return void logger.error(`${logModule}: extension enabling error`, { ext, error });
            }
          })(),
        });
      } catch (error) {
        logger.error(`${logModule}: extension loading error`, { ext, error });
      }
    } else if (!ext.isEnabled && alreadyInit) {
      this.removeInstance(extId);
    }
  }

  return extensions;
}
```

panuhorsmalahti: That doesn't work for this reason:
```
    // We first need to wait until each extension's `onActivate` is resolved,
    // as this might register new catalog categories. Afterwards we can safely .enable the extension.
```

Your code example might call .enable before another extension is activated fully, thus causing error.

See earlier PR for some discussion:
https://github.com/lensapp/lens/pull/4702

Nokel81: If that parameter becomes `IComputedValue<CatalogEntity | undefined>` and you do an `await when(() => Boolean(entity.get());` before that if, then the race condition would be fixed as well and it would also be fixed for non-bundled extensions.

panuhorsmalahti: I propose we fix that in another PR as that's becoming not related to the issue at hand.

Nokel81: Okay sounds good

Nokel81 (approved):

PR #4792, opened 2022-02-02
Fix bugs in ItemListLayouts/KubeObjectListLayouts

jansav:
**Bugs:**

1. Scrolls to top when selecting a row - Closes #4793 
2. Scrolls to top when selecting all rows
3. Rendering of contents flicker
4. Getting release values sometimes throw
5. Render-storm in Events - Closes #4769 

**Other changes:**

1. Replace avoidable `asyncComputed` with normal `computed`
2. Remove noisy debugging from `getDetailsUrl`

**Still to be fixed:**

1. KubeObjectListLayout pages are re-rendered involuntarely losing scroll position

PR #4808, opened 2022-02-03
Fix ItemListLayouts scrolling to top involuntarely

### LINKED ISSUE #4796: ItemListLayout scrolls to top involuntarely
1. Clicking a row (say in pods) to open up the drawer still makes the list bounce around
2. Closing that drawer makes the list scroll to the top
3. The lists still scroll to the top when items are added to the list (Create a cronjob from the template under Create resource - this creates a new pod every minute)

_Originally posted by @Nokel81 in https://github.com/lensapp/lens/issues/4792#issuecomment-1027951457_

jansav:
The culprit was wrapper components using type parameters that created new "dynamic components" using `withInjectables` on every render making the components indistinguishable for React. This causes the dynamic components to re-mount instead of re-render causing all the trouble.

Closes #4796

### DIFF
--- a/src/renderer/components/kube-object-list-layout/kube-object-list-layout.tsx
+++ b/src/renderer/components/kube-object-list-layout/kube-object-list-layout.tsx
@@ -138,23 +138,25 @@ class NonInjectedKubeObjectListLayout<K extends KubeObject> extends React.Compon
   }
 }
 
+const InjectedKubeObjectListLayout = withInjectables<
+  Dependencies,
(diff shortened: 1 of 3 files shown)


jakolehm (approved):
LGTM

Iku-turso:
Worth noting: this fix closer to the now-realized root problem would have fixed most (but not all) bugs already attended in #4792. Good.

Iku-turso (approved):

PR #4813, opened 2022-02-03
Enforce some naming conventions

Nokel81:
- No type may be called "Props" or "State" (must be more descriptive). This is necessary for better documentation
- Enforce the majority preference to `interface` declarations over `type` for object types. 
- Require that interfaces and type declarations must be PascalCase.


ixrock (approved):

src/renderer/components/dock/info-panel.tsx
```diff
@@ -17,12 +17,12 @@ import { Notifications } from "../notifications";
...
 import dockStoreInjectable from "./dock/store.injectable";
 
-interface Props extends OptionalProps {
+export interface InfoPanelProps extends OptionalProps {
   tabId: TabId;
   submit?: () => Promise<ReactNode | string | void>;
 }
 
-interface OptionalProps {
+export interface OptionalProps {
```
jim-docker, line 25:
```suggestion
export interface InfoPanelOptionalProps {
```
?

src/renderer/components/+custom-resources/crd-list.tsx
```diff
@@ -32,7 +32,7 @@ enum columnId {
 }
 
 @observer
-export class CrdList extends React.Component {
+export class CustomResourceDefinitions extends React.Component {
```
jim-docker, line 35: is this a breaking change?

Nokel81: No, because it is not exported

src/renderer/components/dock/info-panel.tsx
Nokel81: I'll do that in a follow up PR.

jim-docker (approved):

PR #4817, opened 2022-02-04
Replace WorkloadsOverviewDetailRegistry with reactive solution

jansav:
Signed-off-by: Janne Savolainen <janne.savolainen@live.fi>

Closes #4816

PR #4820, opened 2022-02-04
Replace CatalogEntityDetailRegistry with reactive solution

### LINKED ISSUE #4819: CatalogEntityDetailRegistry

jansav:

Closes #4819


src/renderer/components/+catalog/catalog-entity-details.tsx
```diff
@@ -109,3 +114,18 @@ export class CatalogEntityDetails<T extends CatalogEntity> extends Component<Pro
     );
   }
 }
+
+const InjectedCatalogEntityDetails = withInjectables<Dependencies<CatalogEntity>, Props<CatalogEntity>>(
```
Nokel81: nit: does this really need to be spread out over so many lines?

src/renderer/components/+catalog/details-for-catalog-entity.injectable.ts
```diff
@@ -0,0 +1,35 @@
...
+import type { CatalogEntity } from "../../../common/catalog";
+import detailItemRegistrationsInjectable from "./detail-item-registrations.injectable";
+import { conforms, eq, includes } from "lodash/fp";
+
+const detailsForCatalogEntityInjectable = getInjectable({
+  instantiate: (di, catalogEntity: CatalogEntity) => {
+    const allRegistrations = di.inject(detailItemRegistrationsInjectable);
+
+    const registrationsForEntity = allRegistrations.get().filter(
+      conforms({
```
Nokel81: Do we really want to increase our dependence on lodash?

src/renderer/components/+catalog/detail-item-registrations.injectable.ts
```diff
@@ -0,0 +1,39 @@
...
+export default detailItemRegistrationsInjectable;
+
+const orderByPriority = (
+  registrations: CatalogEntityDetailRegistration<CatalogEntity>[],
+) =>
+  orderBy(
+    "priority",
+    "desc",
+    registrations.map(({ priority = 50, ...rest }) => ({ priority, ...rest })),
+  );
```
Nokel81: This is something we do a lot within the extension API, can we try and factor it out into a common utility?

jansav: Nothing to be shared here. `orderBy` would be it, but it's already reusable. Extracted to separate function just to make use-place cleaner.

Nokel81: No it is shared, we have a bunch of types that extend `{ priority?: number }`, so I could imagine the following function

```ts
interface PrioritySortable {
  priority?: number;
}

export function orderByPriority<T extends PrioritySortable>(objs: T[]) {
  const withDefaultPriority = objs.map(({ priority = 50, ...rest }) => ({ priority, ...rest }));

  return orderBy("priority", "desc", withDefaultPriority);
}
```

src/renderer/components/+catalog/details-for-catalog-entity.injectable.ts
jansav: TBH `lodash` is only library that I think we should depend without creating abstractions. That's because it provides helper functions which can be replaced relatively easily if ever needed. `lodash` provides a lot simplicity and ability to get rid nested code when combined with functional programming. (which we don't do at the moment)

src/renderer/components/+catalog/detail-item-registrations.injectable.ts
jansav: Sure, as long as there is no other usages which require default priority other than `50`. 

And when we start sharing functions it becomes even more important that function does only one thing and name doesn't lie. In example you provided function lies and does two things.

1. Adds default priority
2. Orders by priority

Just to keep in mind in future, since it feels acceptable in this case.

I'll do the change.

jansav: Done. Keep in mind that base of this branch changed. Do not merge until #4817 has been merged.

src/renderer/utils/order-by-priority.ts
```diff
@@ -0,0 +1,17 @@
+/**
```
ixrock: Maybe this file should be located in `common/utils`?

src/renderer/components/+catalog/detail-item-registrations.injectable.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable, lifecycleEnum } from "@ogre-tools/injectable";
+import { computed } from "mobx";
+import rendererExtensionsInjectable from "../../../extensions/renderer-extensions.injectable";
+import coreDetailItemRegistrationsInjectable from "./core-detail-item-registrations.injectable";
+import { orderByPriority } from "../../utils/order-by-priority";
+
+const detailItemRegistrationsInjectable = getInjectable({
```
ixrock: When first time see this code/file it's hard to predict what kind of type it returns/accepts? (without extra navigating in codebase).
Maybe we should start using definition with type? Kinda `const detailItemRegistrationsInjectable: Injectable<string[]> = ...` so `detailItemRegistrations` expected to be strings array. Wdyt?

src/renderer/components/+catalog/details-for-catalog-entity.injectable.ts
```diff
@@ -0,0 +1,35 @@
...
+        (registration) => registration.priority > 999,
+      ),
+
+      components: registrationsForEntity.map(
+        (registration) => registration.components.Details,
+      ),
+    };
+  },
+
+  lifecycle: lifecycleEnum.transient,
```
ixrock: Random question: if i replace lifecycle to `lifecycleEnum.singleton` how it would it break this injectable usages?

ixrock (dismissed):

src/renderer/components/+catalog/details-for-catalog-entity.injectable.ts
jansav: Two things:

1. Injectable would throw because we would be using `singleton` with `instantiationParameter`. (see below)
2. (if 1. defensive code would not be present) CatalogEntity used would be same for every time when injected.

However, this is something that we are actually working on, because we see this bad design. 

This should be `singleton` but only in scope of `instantiationParameter`. So if you give `"some-string"` and `"some-other-string"` as `instantiationParameter`, you will get two different instances, but if you inject `"some-string"` twice, you will get same instance. This allows us to get rid of these type of transient lifecycle usages and start using `computed` as a mechanism for achieving reactive UI.

src/renderer/components/+catalog/detail-item-registrations.injectable.ts
jansav: Hm. I think it would help a lot if I just extracted implementation to separate function.

src/renderer/utils/order-by-priority.ts
jansav: Hm, maybe. It's used only in renderer though and that's why it's there.

src/renderer/components/+catalog/detail-item-registrations.injectable.ts
Nokel81: Why was the branch changed if that PR doesn't also use the function?

jansav: This PR makes the change for both.

Nokel81: That is weird, why have it so that we merge a PR and then immediately merge another PR that changes it?

jansav: Because both of these need the same code. One has to be merged first so other can use it. I don’t think PR process should cripple refactoring. 

End result is same and both of these are relatively small PRs.

Nokel81: I would have just thought that adding the function in that PR and then just using it in this PR would make more sense and have a cleaner history.

src/renderer/utils/order-by-priority.ts
panuhorsmalahti: I think code that can be used in both should be placed in common.

jansav:
@panuhorsmalahti @Nokel81 @ixrock: This branch has been completely reworked with modern tools. PTAL again.

src/renderer/components/+catalog/catalog-entity-detail-items/extension-catalog-entity-detail-items-registrator.injectable.ts
```diff
@@ -0,0 +1,45 @@
...
+
+            instantiate: () => ({
+              apiVersions: registration.apiVersions,
+              kind: registration.kind,
+
+              components: {
+                Details: registration.components.Details,
+              },
+
+              orderNumber: -registration.priority || -50,
```
Nokel81, line 30: is the priority negative?

src/renderer/components/+catalog/catalog-entity-detail-items/extension-catalog-entity-detail-items-registrator.injectable.ts
```diff
@@ -0,0 +1,45 @@
...
+              extension,
+            }),
+
+            injectionToken: catalogEntityDetailItemInjectionToken,
+          }),
+        ),
+
+        forEach(di.register),
```
Nokel81, line 38: Why is this better than using `computed` directly? Now we have a whole bunch of instantiates every time any single extension gets added or removed or adds an item.

src/renderer/components/+catalog/catalog-entity-detail-items/catalog-entity-detail-item-injection-token.ts
```diff
@@ -0,0 +1,25 @@
...
+  apiVersions: string[];
+  kind: string;
+
+  components: {
+    Details: React.ComponentType<CatalogEntityDetailItemComponentProps<T>>;
+  };
+
+  orderNumber: number;
+
+  extension?: LensRendererExtension;
```
Nokel81, line 20: Why is this optional?

jansav: It's optional because there is no need to limit `catalog-entity-detail-item` for extensions. Same interface is used for detail-items that live in OpenLens. Notice that this relates to other comment.

src/renderer/components/+catalog/catalog-entity-detail-items/extension-catalog-entity-detail-items-registrator.injectable.ts
jansav: Because we changed it to `orderNumber`. Priority has double meaning; it doesn't tell whether items should be sorted in ascending or descending order. `orderNumber` does so. It also allows us to use `sortBy` from `lodash/fp`

src/renderer/components/+catalog/catalog-entity-detail-items/extension-catalog-entity-detail-items-registrator.injectable.ts
jansav: `computed` has limitations. We noticed those during refactoring of routes. `computed` is especially problematic when there is `React` components inside. It's very easy to create new instance of a component every time.

`registrator`-pattern is something that we are trying for couple reasons: 
1. Extensions can just register more implementations of an injection token (use places don't need to prepare for it)
2. We have full control of given registrator in single place. Which means that we can keep use places flexible and create adapters for Extension API whenever needed without introducing breaking changes.

There is still some downsides in `registator`-pattern, most critical ones are the usage of installation counter and that we need to filter results of `injectMany` based on enabled extensions. This is however something that we are actively solving and the end goal is that you only need to `injectMany` and get results as reactive.

jansav: Note: This is still something that we are trying out to see if it carries us over the finish line or not.

src/renderer/components/+catalog/catalog-entity-detail-items/extension-catalog-entity-detail-items-registrator.injectable.ts
Nokel81: But then we have to remember to make them negative, could sort by increasing by default?

src/renderer/components/+catalog/catalog-entity-detail-items/extension-catalog-entity-detail-items-registrator.injectable.ts
Nokel81: > computed is especially problematic when there is React components inside. It's very easy to create new instance of a component every time.

What do you mean by this? Wouldn't some form of shallow equality or reference marker be the solution then?

jansav: For example: 

In case of pages we have logic that makes page component observable if page params are defined. It also provides page params as props for the component. This requires that we introduce higher-order component for Page component if page params are present. This higher-order component can't be created inside `computed` since that would cause React to treat it as a new component (re-mount) every time that the value of computed recomputes.

```typescript
computed(() => {
  return pipeline({
    extensions,
    flatMap(extension => extension.clusterPages),

    map(pageRegistration => ({
      ...pageRegistration,
      
      components: {
        ...pageRegistration.components,
        
        ...(pageRegistration.params ? {
          Page: observer(() => (<pageRegistration.components.Page />))
        } : {})
      }
    });
  });
});
```

Notice that I don't remember the exact PageRegistration API so property names could be wrong.

src/renderer/components/+catalog/catalog-entity-detail-items/extension-catalog-entity-detail-items-registrator.injectable.ts
jansav: No need, in extension API property name is still `priority` and it's positive, but in OpenLens (and hopefully in Extension API 2.0) property name is `orderNumber` which is also positive. Code that we are discussing about is the adapter which converts Extension API 1.0 to internal API.

src/renderer/components/+catalog/catalog-entity-detail-items/extension-catalog-entity-detail-items-registrator.injectable.ts
Nokel81: Why do we need to do that? I would argue that since the page dereferences computed values, if it is not marked as observable itself, that is a bug or at least wrong and should be fixed. Locality of use is key here.

jansav: I completely agree with you on that. We shouldn't make the component `observer` in behind the scenes. However, that's how current implementation works and is exposed in Extension API so we can't make the change without introducing breaking change.  

Anyway, that's just one requirement. There might be other requirements for higher-order components which cause the same exact issue. E.g. we add `TabLayout` higher order component to page components if they have siblings.

jansav: We are currently testing the hypothesis that Extension register new implementations of an `injectionToken` which then allows us to treat extensions as first-class citizens and therefore introduce a lot simplicity. Imagine that everything is built in a way that if we that we want to expose something to Extension API, we just expose the interface for injection token. In that world, we have:

1. One place that registers all implementations of `injectionToken` from extension (extension is enabled)
2. One place that deregisters all injectables that was registered because of extension was enabled (extension is disabled)

In that way we can use `di.injectMany` (or something similar) safely everywhere we want to have all implementations and we don't need to worry about enabled extensions etc. Also `registrator` will be consolidated to one place. 

Treat `registrator` as an adapter for Extension API 1.0 -> Extension API 2.0, which allows us to experiment with requirements of an interface before we actually expose the interface to extensions. At the same time it allows us to start working on the changes required (if any) in use places of said implementations.

PR #4826, opened 2022-02-04
Fix Cluster.refresh sometimes taking upwards of 20s

Nokel81:
- Only read and parse the proxy config once

- Reuse the AuthorizationV1Api instance for the entire refresh instead
  of recreating it between 32 and 302 times, this should allow for
  reusing sockets

### DIFF
--- a/src/common/cluster/cluster.ts
+++ b/src/common/cluster/cluster.ts
@@ -7,7 +7,7 @@ import { ipcMain } from "electron";
 import { action, comparer, computed, makeObservable, observable, reaction, when } from "mobx";
 import { broadcastMessage } from "../ipc";
 import type { ContextHandler } from "../../main/context-handler/context-handler";
-import { AuthorizationV1Api, CoreV1Api, HttpError, KubeConfig, V1ResourceAttributes } from "@kubernetes/client-node";
+import { HttpError, KubeConfig } from "@kubernetes/client-node";
 import type { Kubectl } from "../../main/kubectl/kubectl";
 import type { KubeconfigManager } from "../../main/kubeconfig-manager/kubeconfig-manager";
 import { loadConfigFromFile, loadConfigFromFileSync, validateKubeConfig } from "../kube-helpers";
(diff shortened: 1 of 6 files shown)


jakolehm (approved):
This does not solve the issue of slow requests when network resources are completely exhausted but still an improvement.

src/common/cluster/cluster.ts
```diff
@@ -507,50 +521,6 @@ export class Cluster implements ClusterModel, ClusterState {
...
-  }
-
-  /**
-   * @internal
-   */
-  async canUseWatchApi(customizeResource: V1ResourceAttributes = {}): Promise<boolean> {
-    return this.canI({
-      verb: "watch",
-      resource: "*",
-      ...customizeResource,
```
jim-docker, line 550: This option is not used?

src/common/cluster/cluster.ts
```diff
@@ -425,10 +429,20 @@ export class Cluster implements ClusterModel, ClusterState {
    * @internal
    */
   private async refreshAccessibility(): Promise<void> {
-    this.isAdmin = await this.isClusterAdmin();
-    this.isGlobalWatchEnabled = await this.canUseWatchApi({ resource: "*" });
-    this.allowedNamespaces = await this.getAllowedNamespaces();
-    this.allowedResources = await this.getAllowedResources();
+    const proxyConfig = await this.getProxyKubeconfig();
+    const canI = this.dependencies.createAuthorizationReview(proxyConfig);
```
jim-docker, line 433: Is this where the optimization is?

src/common/cluster/cluster.ts
Nokel81: No, that function was only used in one place and that was within the `refreshAccessibility` method above.

src/common/cluster/cluster.ts
Nokel81: Yes, it creates the API object only once instead of dozens or hundred of times.

jim-docker: I can see here where it creates it once instead of twice like the old code. Where else is the creation avoided?

Nokel81: The big win is from `getAllowedResources` which calls `canI` in a loop

jim-docker: ah, right, the class member version of `canI()` was creating the api each time it was called.

Nokel81: Exactly

jim-docker (approved):

PR #4830, opened 2022-02-07
Fix CustomResourceDefinitions not being navigable

Nokel81:
- Make buildURL more type safe
- Removed `buildURLPositional` as it is only used in one place and was incompatible with the type safety above
- 


src/common/routes/config.ts
```diff
@@ -12,15 +12,15 @@ import { pdbRoute } from "./pod-disruption-budgets";
...
-    hpaRoute,
-    pdbRoute,
-  ].map(route => route.path.toString()),
+    configMapsRoute.path,
+    secretsRoute.path,
+    resourceQuotaRoute.path,
+    limitRangesRoute.path,
+    hpaRoute.path,
+    pdbRoute.path,
+  ],
```
jim-docker, line 23: Why change this? And did you mean to change all such instances? (like `networkRoute`)

jim-docker (commented):
Where exactly is the bug fix? And are you sure you want to risk introducing new bugs right before release with all these changes? I don't know if it applies to lines changed but "they" say there are 15-50 bugs introduced per 1000 lines of code

jim-docker:
Clicking on Helm (Charts, Releases) and Access Control fields routes to Custom Resources

https://user-images.githubusercontent.com/40840436/152876185-149b4c6e-63da-487c-9364-e712057ce898.mov

Nokel81:
@jim-docker Yes I have found that too. I am going to close this PR as I have found a lot of issues with it and will go back to the drawing board.

PR #4837, opened 2022-02-08
Bundle kubectl 1.23.3

jakolehm:
- bundle kubectl v1.23.3
- add 1.22 - 1.23 to kubectl version matrix


Nokel81 (dismissed):

Nokel81:
I think we should eventually make it so that we pick the latest patch version automatically for all non-bundled kubectl versions, but that is future work.

aleksfront (dismissed):

Nokel81 (approved):

jim-docker (approved):

PR #4679, opened 2022-01-12
Fix: remove hotbar name from status bar

aleksfront:
* Reverting part of the changes from https://github.com/lensapp/lens/pull/4476 by removing hotbar name from status bar
* Showing active hotbar name in tooltip when switching hotbars

https://user-images.githubusercontent.com/9607060/149136446-7ef8efcd-0412-4a65-a2ab-e4ff7da62843.mov

PR #4838, opened 2022-02-08
Fix StatusBar extension API

Nokel81:
- Make sure that items are still on right side if no items are
  registered on left side

- Make more injectable

- Switch the naming to `StatusBar` to be more consistent within the code base


jim-docker (commented):
Is this a refactor or did you just rename some files?

src/renderer/components/status-bar/status-bar.tsx
```diff
@@ -0,0 +1,49 @@
...
+        {right.map((Item, index) => (
+          <div className={styles.item} key={index}>
+            <Item />
+          </div>
+        ))}
+      </div>
+    </div>
+  );
+
+});
```
Nokel81, line 42: @jim-docker I mostly just renamed some files, but the main part of the bug fix is here and in `src/renderer/components/status-bar/registered-status-bar-items.injectable.tsx`.

The bug was occurring because the old implementation assumed that there was always going to be an item on the left hand side (the active hotbar name). However, that was removed in #4679 which caused this bug to be noticed.

The solution that I have done here is to actually separate the left and the right sides in the DOM (that is what the next `div`'s are for) and use CSS to spread out those. Instead of spreading out the items themselves.

src/renderer/components/status-bar/status-bar.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  beforeEach(async () => {
+    di = getDiForUnitTesting({ doGeneralOverrides: true });
+    render = renderFor(di);
+
+    await di.runSetups();
+  });
+
+  it("renders w/o errors", () => {
+    di.override(statusBarItemsInjectable, () => computed(() => [] as StatusBarRegistration[]));
```
jansav: By overriding statusBarItems you made test less credible. Could we change it back so that it overrides extensions?

Nokel81: Okay, would an intermediate dependency be sufficient as well? One whose only job is to `flatMap` from the extensions?

jansav: There is no need for that. `extensions.get().flatMap(...)` is duplicated yes, but abstracting it would just make things more complicated.

Nokel81: Okay, I'll revert.

Nokel81: Done

jim-docker (dismissed):
minor issue, may have pre-existed, the right-most item must fight for focus with the window-resize operation:

https://user-images.githubusercontent.com/40840436/153279668-c75be8a4-d23d-43fa-ae2a-bc3b11172baf.mov

this is on mac. On linux it is OK.

aleksfront:
Fixed styling issues, @jim-docker ptal once more.
link with icon
status bar line

aleksfront (approved):

jim-docker:
> Fixed styling issues, @jim-docker ptal once more.

Were you trying to fix the minor issue I mentioned? It's still there but I don't think it is a big deal. Looks good otherwise

jim-docker (approved):

PR #4842, opened 2022-02-09
Update injectable library

jansav:
Changes are:

1. Require `id` for injectables to permit better error messages
2. Remove uses of `lifecycle: lifecycleEnum.singleton` as it is the default
3. Make injections happening in setup async. 

Motivation:

1. Allow setuppable to inject another setuppable
2. Introduce detection of circular injections
3. Introduce detection of circular setuppables
4. Improve error messages for injecting non-registered injectables
5. Introduce injectMany as new pattern
6. Introduce keyed singletons as new lifecycle
7. Allow instantiation parameter of any type
8. Introduce more concise naming of types


src/renderer/app-paths/app-paths.injectable.ts
```diff
@@ -2,15 +2,17 @@
...
 import getValueFromRegisteredChannelInjectable from "./get-value-from-registered-channel/get-value-from-registered-channel.injectable";
 
 let syncAppPaths: AppPaths;
 
 const appPathsInjectable = getInjectable({
+  id: "app-paths",
+
   setup: async (di) => {
-    const getValueFromRegisteredChannel = di.inject(
+    const getValueFromRegisteredChannel = await di.inject(
```
jansav, line 15: Notice: `di.inject` which happens in `setup` is always async because if the injected injectable is also setuppable, the setup will be ran first asynchronously before instantiation.

Nokel81:
> Remove `lifecycle: lifecycleEnum.singleton` being default

Do you mean that singleton is now the default so we don't have to specify it anymore?

Nokel81:
> Add id for injectables to permit better error messages

Why make these explicit and not based on the file path?

jansav:
> > Remove `lifecycle: lifecycleEnum.singleton` being default
> 
> Do you mean that singleton is now the default so we don't have to specify it anymore?

Yes

jansav:
> > Add id for injectables to permit better error messages
> 
> Why make these explicit and not based on the file path?

Hmm, good thinking. Maybe that could be done.

jansav:
> > > Add id for injectables to permit better error messages
> > 
> > 
> > Why make these explicit and not based on the file path?
> 
> Hmm, good thinking. Maybe that could be done.

Seems like we don't have easy, reliable way to access filename. 

```typescript
console.log(module.filename); // undefined when running app.
console.log(__filename); // always ./index.js when running app.
```

Anything more complicated seems unwelcome.

Nokel81:
I mean you could add the filename while building the container from the globs. Or you could use `__filename` from Node

jansav:
> I mean you could add the filename while building the container from the globs. Or you could use `__filename` from Node

ID needs to be present in the `injectable` and cannot be added in register time. Because `di.inject` you need to provide injectable or injection token containing the ID.

`__filename` works in unit tests but not in app it self as explained above.

Nokel81:
Is the idea to move to ID strings instead of the object references?

Because this is JS, could the registration not modify the injectables directly?

jansav:
> Is the idea to move to ID strings instead of the object references?
> 
> Because this is JS, could the registration not modify the injectables directly?

No. ID strings are used for error messages and creating context (who injects what) at the moment. Mutating is very dangerous and shouldn't be used here.

Nokel81 (approved):

jim-docker (approved):

PR #4843, opened 2022-02-09
Allow better typings for extending type of KubernetesCluster

Nokel81:

### DIFF
--- a/src/main/cluster-manager.ts
+++ b/src/main/cluster-manager.ts
@@ -48,7 +48,7 @@ export class ClusterManager extends Singleton {
     );
 
     reaction(
-      () => catalogEntityRegistry.getItemsByEntityClass(KubernetesCluster),
+      () => catalogEntityRegistry.getItemsByEntityClass(KubernetesCluster) as KubernetesCluster[],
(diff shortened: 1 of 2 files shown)


src/main/cluster-manager.ts
```diff
@@ -48,7 +48,7 @@ export class ClusterManager extends Singleton {
     );
 
     reaction(
-      () => catalogEntityRegistry.getItemsByEntityClass(KubernetesCluster),
+      () => catalogEntityRegistry.getItemsByEntityClass(KubernetesCluster) as KubernetesCluster[],
```
jakolehm, line 51: Why do we need this?

jakolehm (approved):

jim-docker (approved):

src/main/cluster-manager.ts
Nokel81: Because the change to definition of `KubernetesCluster` broke a type inference in the `getItemsByEntityClass` method.

Which TBH we should not even have that type parameter for that method but I didn't want to remove it in this PR.

PR #4844, opened 2022-02-10
Expose Electron powerMonitor API to extensions

jweak:

Electron API lets extension to react on `suspend` and `resume` events.

### DIFF
--- a/src/extensions/main-api/power.ts
+++ b/src/extensions/main-api/power.ts
@@ -4,7 +4,7 @@
  */
 import { powerMonitor } from "electron";
 import type { Disposer } from "../../common/utils/disposer";
-type PowerEventListener = () => void;
+export type PowerEventListener = () => void;
 
 export const onSuspend = (listener: PowerEventListener): Disposer => {
   powerMonitor.on("suspend", listener);
--- a/src/extensions/main-api/power.ts
+++ b/src/extensions/main-api/power.ts
@@ -1,0 +1,6 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+export { powerMonitor } from "electron";
+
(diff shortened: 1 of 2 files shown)


src/extensions/main-api/power.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+export { powerMonitor } from "electron";
```
Nokel81: I would like to see an abstraction here instead of directly exposing this.

src/extensions/main-api/power.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { powerMonitor } from "electron";
+import type { Disposer } from "../../common/utils/disposer";
+type PowerEventListener = () => void;
```
Nokel81: Please export this type since it will be required to soon anyway for documentation purporses

src/extensions/main-api/power.ts
jweak: Abstracted `powerMonitor` implementation with `onSuspend`, `onResume` and `onShutdown` functions

src/extensions/main-api/power.ts
jweak: Done.

src/extensions/main-api/power.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { powerMonitor } from "electron";
+import type { Disposer } from "../../common/utils/disposer";
+export type PowerEventListener = () => void;
+
+export const onSuspend = (listener: PowerEventListener): Disposer => {
```
Nokel81, line 18: Note: in the future we will need to do some documentation for these functions, it might be useful to do that now. Just a simple one liner doc comment explaining what each of these four items (the 3 functions and the type) "do"

Nokel81 (dismissed):

Nokel81 (approved):

jim-docker (approved):

PR #4852, opened 2022-02-11
Convert catalog entity sync to websocket

Nokel81:
- Send events about individual entities instead of the whole set

- Do basic diffing for updates


Nokel81:
This PR has served its purpose of being a place to discuss things like this. I will close it for the time being.

PR #4854, opened 2022-02-11
Rename KubernetesCluster MenuItem from Delete to Remove

### LINKED ISSUE #4846: "Delete" action for a cluster in Lens should be renamed to "Remove"
How important is it that Lens use the term "Delete" in the context menu for a cluster, to remove that cluster from Lens?

I ask because "Delete" has a much stronger connotation than "Remove" does, and since deleting the cluster from the Lens Catalog just removes it from Lens (without actually deleting it from the cloud it's being hosted in), "Remove" would seem more appropriate.

This would then solve a naming problem we're having with the new version of the [Lens MCC Extension](https://github.com/Mirantis/lens-extension-cc) where we'll enable users to delete (actually delete from the MCC management cluster) a cluster they have synced into Lens (or created from Lens in the MCC mgmt cluster).

Otherwise, we have the Lens version of "Delete" which really just _removes_ from Lens itself (leaving the physical cluster untouched), and then the extension's version of "Delete" which is like "destroy and remove from Lens".

We'll have two "Delete" items in the context menu for a Cluster in the Catalog -- or we'll have to find another even stronger term than "Delete" for our purposes.

Nokel81:

fixes https://github.com/lensapp/lens/issues/4846

### DIFF
--- a/integration/__tests__/cluster-pages.tests.ts
+++ b/integration/__tests__/cluster-pages.tests.ts
@@ -320,7 +320,7 @@ utils.describeIf(minikubeReady(TEST_NAMESPACE))("Minikube based tests", () => {
     await frame.waitForSelector(`.Menu >> text="Add to Hotbar"`);
     await frame.waitForSelector(`.Menu >> text="Settings"`);
     await frame.waitForSelector(`.Menu >> text="Disconnect"`);
-    await frame.waitForSelector(`.Menu >> text="Delete"`);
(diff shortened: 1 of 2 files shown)


stefcameron (dismissed):
Thank you!

stefcameron (approved):
👍

jansav (approved):

jweak (approved):

PR #4859, opened 2022-02-14
Make routes in back-end comply to Open Closed Principle

jansav:
## Motivation:

- I should be able to add route with minimal modifications (none)
- Route should have it's own dependencies
- Route shouldn't be responsible of e.g. error handling which can be abstracted to higher level.

## Solution:

Introduce `routeInjectionToken` which any back-end route can implement by adding `injectionToken: routeInjectionToken`. Any injectable implementing said `injectionToken` will be registered automatically.

Note: base-branch is PR #4842 so let's get that merged first.

## TBD later:
1. Get rid of `IncomingMessage` and `http.ServerResponse` interface from router so that new routes can be unit tested easily even when using router in tests.


src/main/router.ts
```diff
@@ -4,18 +4,16 @@
...
+import { toPairs } from "lodash/fp";
 import path from "path";
-import { readFile } from "fs-extra";
 import type { Cluster } from "../common/cluster/cluster";
-import { apiPrefix, appName, publicPath, isDevelopment, webpackDevServerPort } from "../common/vars";
-import { HelmApiRoute, KubeconfigRoute, MetricsRoute, PortForwardRoute, ResourceApplierApiRoute, VersionRoute } from "./routes";
-import logger from "./logger";
+import type { LensApiResultContentType } from "./router-content-types";
+import { contentTypes } from "./router-content-types";
+import "../common/vars";
```
ixrock: Why do we need import whole package without using anything locally from it?

src/main/routes/static-file-route.injectable.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    try {
+      const filename = path.basename(req.url);
+
+      // redirect requests to [appName].js, [appName].html /sockjs-node/ to webpack-dev-server (for hot-reload support)
+      const toWebpackDevServer =
+        filename.includes(appName) ||
+        filename.includes("hot-update") ||
+        req.url.includes("sockjs-node");
+
```
ixrock: Lines related to `webpack-dev-server` were removed from `router.ts` in https://github.com/lensapp/lens/pull/4894 and here probably should be too.

package.json
```diff
@@ -326,6 +326,7 @@
     "@typescript-eslint/eslint-plugin": "^5.10.1",
     "@typescript-eslint/parser": "^5.10.1",
     "ansi_up": "^5.1.0",
+    "mock-http": "^1.1.0",
```
ixrock, line 337: Why do we need this package / extra dependency? Can we live without it?

ixrock (dismissed):
Didn't run the app, but code LGTM.

src/main/router.ts
jansav: My question exactly... There is side-effect in `common/vars` which defines `__static` and therefore is required. Something that is of course wrong, but out of scope in this PR.

src/main/routes/static-file-route.injectable.ts
jansav: Let's figure this out when we start merging 5.5.0 stuff to master. Good catch.

package.json
jansav: At the moment, no.

> Get rid of IncomingMessage and http.ServerResponse interface from router so that new routes can be unit tested easily even when using router in tests.

However, we can after this.

src/main/helm/helm-release-manager.ts
```diff
@@ -108,7 +108,7 @@ export async function upgradeRelease(name: string, chart: string, values: any, n
 
     return {
       log: output,
-      release: getRelease(name, namespace, kubeconfigPath, kubectlPath),
+      release: await getRelease(name, namespace, kubeconfigPath, kubectlPath),
```
chenhunghan, line 97: Noticed no `catch` in this `try...finally`, will `await getRelease` throw? Or error is caught and handled somewhere else?

src/main/routes/static-file-route.injectable.ts
jansav: Merge conflict resolved.

src/main/helm/helm-release-manager.ts
jansav: Key difference here is that all back-end API responds go through same place which already catches fatal errors. That being said, if it fails fatally, API call is resolved with status code `422` and error message. And in this case, there is no non-fatal error to be handled. Goal is to get rid of try-catches in operational code for good (strategic places that handle errors for everything to allow error monitoring etc). Throwing is mechanism that is reserved for fatal errors which should prevent code to be executed and for non fatal errors we have better alternatives.

Iku-turso (dismissed):

src/main/router-content-types.ts
```diff
@@ -0,0 +1,86 @@
+/**
```
Nokel81: I think that all of these router files could be in there own folder:

```
src/
  main/
    router/
      content-types.ts
      router.ts
      router.injectable.ts
      ...
```

src/main/router-content-types.ts
```diff
@@ -0,0 +1,86 @@
...
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import type { LensApiResult } from "./router";
+
+export interface LensApiResultContentType {
+  resultMapper: (result: LensApiResult<any>) => ({
+    statusCode: number;
+    content: any;
+    headers: { [ name: string ]: string };
```
Nokel81: nit: `Record<string, string>` or even `Partial<Record<string, string>>`

src/main/router-content-types.ts
```diff
@@ -0,0 +1,86 @@
...
+      const resultMapper = resultMapperFor("application/json");
+
+      const mappedResult = resultMapper(result);
+
+      const contentIsObject = typeof mappedResult.content === "object";
+      const contentIsBuffer = mappedResult.content instanceof Buffer;
+      const contentShouldBeStringified = contentIsObject && !contentIsBuffer;
+
+      const content = contentShouldBeStringified
+        ? JSON.stringify(mappedResult.content)
```
Nokel81: Do we really want to stringify a buffer?

src/main/router.test.ts
```diff
@@ -0,0 +1,295 @@
...
+    beforeEach(() => {
+      requestStub = new Request({
+        url: "/some-path",
+        method: "get",
+        headers: {
+          "content-type": "application/json",
+        },
+      });
+
+      // @ts-ignore
```
Nokel81: Lets not use ignore comments, ever.

src/main/router.ts
```diff
@@ -36,40 +34,24 @@ export interface LensApiRequest<P = any> {
...
-    woff2: "font/woff2",
-    ttf: "font/ttf",
-  };
-
-  return mimeTypes[path.extname(filename).slice(1)] || "text/plain";
-}
-
 interface Dependencies {
-  routePortForward: (request: LensApiRequest) => Promise<void>;
+  parseRequest: (request: http.IncomingMessage, _: null, options: { parse: boolean; output: string }) => Promise<{ payload: any }>;
```
Nokel81: why even have the second param?

src/main/router.ts
```diff
@@ -103,78 +86,101 @@ export class Router {
...
-      }
+export interface Route<TResponse> {
+  path: string;
+  method: "get" | "post" | "put" | "patch" | "delete";
+  handler: RouteHandler<TResponse>;
+}
 
-      try {
-        const data = await readFile(asset);
+const handleRoute = (route: Route<any>) => async (request: LensApiRequest, response: http.ServerResponse) => {
```
Nokel81: Shouldn't it be `Route<unknown>`?

src/main/router.ts
```diff
@@ -103,78 +86,101 @@ export class Router {
...
-}
+
+  const contentType = result.contentType || contentTypes.json;
+
+  const mappedResult = contentType.resultMapper(result);
+
+  writeServerResponse(mappedResult);
+};
+
+const writeServerResponseFor =
```
Nokel81: nit, all this extra whitespace makes it harder to read. Could this be a normal function or at least add some `()`?

src/main/routes/helm/charts/get-chart-route.injectable.ts
```diff
@@ -0,0 +1,36 @@
...
+  readme: string;
+  versions: RawHelmChart[];
+}
+
+const getChartRouteInjectable = getInjectable({
+  id: "get-chart-route",
+
+  instantiate: (): Route<GetChartResponse> => ({
+    method: "get",
+    path: `${apiPrefix}/v2/charts/{repo}/{chart}`,
```
Nokel81, line 22: I have (what I think) is a pretty cool idea of how to make sure that this string is "correctly typed" with regards to the interface that `Params` expects below but we can do that later.

src/main/router.ts
jansav: It's third-party library.

src/main/router.test.ts
jansav: Good catch.

src/main/router.ts
jansav: Changed

src/main/router-content-types.ts
jansav: Changed

src/main/router.ts
Nokel81: Yeah but we can wrap it if we want

src/main/router-content-types.ts
jansav: We don't. `const contentShouldBeStringified = contentIsObject && !contentIsBuffer;`

src/main/router.ts
jakolehm: Maybe add comment/todo so that next reader knows why it's there?

src/main/router.ts
```diff
@@ -103,78 +86,101 @@ export class Router {
...
-          response.statusCode = 404;
+  const writeServerResponse = writeServerResponseFor(response);
 
-          return response.end();
-        }
+  try {
+    result = await route.handler(request);
+  } catch(error) {
+    const mappedResult = contentTypes.txt.resultMapper({
+      statusCode: 422,
```
jakolehm: Should this be `500`?

src/main/router-content-types.ts
jansav: Done

src/main/router.ts
jansav: > Should this be `500`?

Previously these were 422. But yeah, seems wrong. Changed.

jakolehm (approved):
LGTM

Iku-turso (approved):

PR #4861, opened 2022-02-14
Stop using HelmCli from Renderer

### LINKED ISSUE #4862: HelmRepoManager

jansav:
Noteworthy:
---
1. Add behaviours for Helm Repository preferences
2. Consolidate all usages of `$ helm` -cli to `main`

This PR looks big, but that's mostly snapshots which should be pretty uninteresting from perspective of review.

Closes #4862


jansav:
This PR was reworked to introduce complete behaviours. PTAL.

src/behaviours/helm-charts/deactivate-helm-repository-from-list-of-active-repository-in-preferences.test.ts
```diff
@@ -0,0 +1,126 @@
...
+            { name: "some-active-repository", url: "some-url" },
+          ],
+        });
+      });
+
+      it("renders", () => {
+        expect(rendered.baseElement).toMatchSnapshot();
+      });
+
+      describe("when deactivating repository", () => {
```
Nokel81: By deactive, do you mean "remove from local"?

src/common/utils/async-result.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+export type AsyncResult<Response, Error = string> =
```
Nokel81, line 5: Why is this specifically called "Async"?

src/main/helm/get-helm-env/get-helm-env.injectable.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import execHelmInjectable from "../exec-helm/exec-helm.injectable";
+import type { AsyncResult } from "../../../common/utils/async-result";
+
+export type HelmEnv = Record<string, string> & {
```
Nokel81, line 9: nit: `Partial<Record<string, string>> & ...`

src/main/helm/helm-binary-path.injectable.ts
```diff
@@ -0,0 +1,35 @@
...
+  instantiate: (di) => {
+    const getAbsolutePath = di.inject(getAbsolutePathInjectable);
+    const lensResourcesDir = di.inject(lensResourcesDirInjectable);
+    const normalizedPlatform = di.inject(normalizedPlatformInjectable);
+    const normalizedPlatformArchitecture = di.inject(normalizedPlatformArchitectureInjectable);
+    const isProduction = di.inject(isProductionInjectable);
+
+    const resourcesDir = isProduction
+      ? lensResourcesDir
+      : getAbsolutePath(lensResourcesDir, "binaries", "client", normalizedPlatform);
```
Nokel81: Shouldn't this be an injectable on its own? Don't we already have this for `kubectl`?

src/main/helm/helm-service/delete-helm-release.injectable.ts
```diff
@@ -0,0 +1,28 @@
...
+const deleteHelmReleaseInjectable = getInjectable({
+  id: "delete-helm-release",
+
+  instantiate: (di) => {
+    const logger = di.inject(loggerInjectable);
+
+    return async (cluster: Cluster, releaseName: string, namespace: string) => {
+      const proxyKubeconfig = await cluster.getProxyKubeconfigPath();
+
+      logger.debug("Delete release");
```
Nokel81, line 19: nit: a `[HELM]:` prefix might make this log line easier to find.

src/main/helm/helm-service/list-helm-charts.injectable.ts
```diff
@@ -0,0 +1,41 @@
...
+const listHelmChartsInjectable = getInjectable({
+  id: "list-helm-charts",
+
+  instantiate: (di) => {
+    const getActiveHelmRepositories = di.inject(getActiveHelmRepositoriesInjectable);
+
+    return async () => {
+      const result = await getActiveHelmRepositories();
+
+      assert(result.callWasSuccessful);
```
Nokel81, line 20: why not return `[]`? Or is this match previous behaviour?

src/main/helm/helm-service/update-helm-release.injectable.ts
```diff
@@ -0,0 +1,45 @@
...
+
+      return upgradeRelease(
+        releaseName,
+        data.chart,
+        data.values,
+        namespace,
+        data.version,
+        proxyKubeconfig,
+        kubectlPath,
+      );
```
Nokel81, line 38: For such large number of parameters should we use object for named parameters?

src/renderer/components/+preferences/kubernetes/helm-charts/activation-of-custom-helm-repository/custom-helm-repo.injectable.ts
```diff
@@ -0,0 +1,25 @@
...
+    insecureSkipTlsVerify: false,
+    caFile: "",
+    keyFile: "",
+    certFile: "",
+  }),
+
+  lifecycle: lifecycleEnum.transient,
+});
+
+export default customHelmRepoInjectable;
```
Nokel81: Why should this be an injectable and not just a function? When would we want to override this?

src/renderer/components/+preferences/kubernetes/helm-charts/activation-of-custom-helm-repository/helm-file-input/get-file-paths.injectable.ts
```diff
@@ -0,0 +1,24 @@
...
+import { getInjectable } from "@ogre-tools/injectable";
+import type { FileFilter } from "electron";
+import { requestOpenFilePickingDialog } from "../../../../../../ipc";
+
+const getFilePathsInjectable = getInjectable({
+  id: "get-file-paths",
+
+  instantiate: () => async (fileFilter: FileFilter) =>
+    await requestOpenFilePickingDialog({
+      // defaultPath: this.getFilePath(type),
```
Nokel81: left over?

src/renderer/components/input/validators/is-path.injectable.ts
```diff
@@ -0,0 +1,32 @@
...
+
+      validate: async (value) => {
+        try {
+          await pathExists(value);
+        } catch {
+          throw new AsyncInputValidationError(
+            `${value} is not a valid file path`,
+          );
+        }
+      },
```
Nokel81, line 27: This is wrong, as seen in #5403 `pathExists` never throws.

Nokel81 (commented):
Looks mostly good

src/main/helm/helm-service/delete-helm-release.injectable.ts
jansav: Old implementation just extracted to own file since we had to modify couple of them.

src/main/helm/helm-service/list-helm-charts.injectable.ts
jansav: Old implementation just extracted to own file since we had to modify couple of them.

src/main/helm/helm-service/update-helm-release.injectable.ts
jansav: Old implementation just extracted to own file since we had to modify couple of them.

src/main/helm/helm-service/delete-helm-release.injectable.ts
Nokel81: Ah I see

src/renderer/components/+preferences/kubernetes/helm-charts/activation-of-custom-helm-repository/custom-helm-repo.injectable.ts
jansav: Key here is the lifecycle. You will get new instance every time when injected. 

This removes requirement for all cleanup after the dialog is closed, since next time you open it you will get new instance. :)

src/renderer/components/+preferences/kubernetes/helm-charts/activation-of-custom-helm-repository/helm-file-input/get-file-paths.injectable.ts
jansav: Yep, good catch. Removing and resolving after done.

src/renderer/components/input/validators/is-path.injectable.ts
jansav: Hm, we've just made it injectable to get rid of usage of global shared state.

src/common/utils/async-result.ts
jansav: All known usages are async. We can rename later if the requirement emerges.

src/renderer/components/+preferences/kubernetes/helm-charts/activation-of-custom-helm-repository/custom-helm-repo.injectable.ts
Nokel81: No I mean, why not have the following function (yes I saw it is transient).

```ts
function newBlankCustomHelmRepo() {
  return observable({
    name: "",
    url: "",
    username: "",
    password: "",
    insecureSkipTlsVerify: false,
    caFile: "",
    keyFile: "",
    certFile: "",
  });
}
```

And then just call that?

src/behaviours/helm-charts/deactivate-helm-repository-from-list-of-active-repository-in-preferences.test.ts
jansav: We tried to come up with naming which is less confusing. 

However, it might be better use the ones that Helm. (Add, Remove, List)

src/main/helm/helm-binary-path.injectable.ts
jansav: Tried to understand the logic here, but without more context it was hard give good name for those. Also if I'm not mistaken, there is no other usages. It can be split when requirement emerges.

src/common/utils/async-result.ts
Nokel81: Okay, it just seems strange that it would say `Async` when it doesn't involve `Promises`.

src/main/helm/helm-binary-path.injectable.ts
Nokel81: There already is uses, just slightly more hidden I guess because of all the DI work lol.

We have this value already as part of `src/common/vars/bundled-resources-dir.injectable.ts` and `src/common/vars/base-bundled-binaries-dir.injectable.ts`

src/renderer/components/input/validators/is-path.injectable.ts
Nokel81: Fair enough, maybe once we merge that PR (which should be soon) this can just be updated to reflect the proper implementation.

src/main/helm/helm-binary-path.injectable.ts
jansav: I'll figure out this tomorrow. Good catch

src/behaviours/helm-charts/deactivate-helm-repository-from-list-of-active-repository-in-preferences.test.ts
jansav: Consolidated naming of Activate and Deactivate to Add and Remove

src/renderer/components/+preferences/kubernetes/helm-charts/activation-of-custom-helm-repository/custom-helm-repo.injectable.ts
jansav: Hmm. Keeping state outside of the component was the thing. Could have been done by using function as you suggested, but just to make it consistent, we decided to make injectable for it. It has other benefits too, since now we can control the lifecycle if we so decide. But here the choice was made just because of consistency.

src/renderer/components/input/validators/is-path.injectable.ts
jansav: Sure. Async validators are also causing trouble in unit tests so there is lot to improve. But, as you said we can tackle that in separate PR. :)

Nokel81 (approved):

Iku-turso (approved):

PR #4867, opened 2022-02-15
Fix: preloading fonts for terminal

ixrock:
_Before:_
Screenshot 2022-02-15 at 17 12 38

_After:_
Screenshot 2022-02-15 at 17 13 43

Note: this behaviour happened only with initial tab opening after page loaded.

follow up #4725


Nokel81 (approved):

ixrock:
Will be recreated with initial #4725 cause of revert in #4872

PR #4868, opened 2022-02-15
release v5.4.0-beta.2

jim-docker:
## 🚀 Features

ShellEnv extension api (**#4802**) @jim-docker  
Allow better typings for extending type of KubernetesCluster (**#4843**) @Nokel81  
Expose Electron powerMonitor API to extensions (**#4844**) @jweak  
Helm v3.7.2 (**#4836**) @jakolehm  

## 🐛 Bug Fixes

Fix: closing terminal tab error (**#4874**) @aleksfront  
Don't block other extensions from loading if one activation fails (**#4805**) @panuhorsmalahti  
Fixes the Catalog ItemListLayout that rendered outdated Entities (**#4845**) @msa0311  
Fix StatusBar extension API (**#4838**) @Nokel81  
Restrict title name to be only string (**#4839**) @Nokel81  
Fix extension sidebar items sometimes being expandable (**#4835**) @Nokel81  
Fix custom-resource navigation (**#4831**) @Nokel81  
Fix Cluster.refresh sometimes taking upwards of 20s (**#4826**) @Nokel81  
Only group delete visible items (**#4798**) @Nokel81  
Fix not being able to install extensions sometimes (**#4794**) @Nokel81  
Fix observation of asyncComputed when invalidated before ever observed (**#4828**) @jansav  

## 🧰 Maintenance

Revert "Upgrade to webpack@5 (#4725)" (**#4872**) @jim-docker  
Upgrade to webpack@5 (**#4725**) @ixrock  
Rename KubernetesCluster MenuItem from Delete to Remove (**#4854**) @Nokel81  
Consolidate way of passing items to ItemListLayout (**#4850**) @jansav  
Bundle kubectl 1.23.3 (**#4837**) @jakolehm

PR #4872, opened 2022-02-15
Revert "Upgrade to webpack@5 (#4725)"

jim-docker:
This reverts commit d656a9e289c87ca7b56c6a79ae84b1f8c5706590.

This is blocking the release build (#4868) since the bundled (not open source) extensions are failing to build against the generated extension api. They may need changes similar to those made to the in-tree extensions in this reverted PR. Maybe the fix is simple but I could not see it.

I think this PR should be reverted until the release can be built (fix the bundled extensions).

I am also not sure that we want to introduce #4725 right before releasing 5.4.0


jansav (approved):

jweak (approved):

aleksfront:
I prefer to stick with webpack 5, it fixes a lot of pain points, now HMR works well and speed up the work in overall. @ixrock Can we fix build issues with small changes to keep v5?

ixrock:
> I prefere to stick with webpack 5, it fixes a lot of pain points, now HMR works well and speed up the work in overall. @ixrock Can we fix build issues with small changes to keep v5?

I think fix should be pretty easy since build doesn't work only with https://github.com/lensapp/lens-ide but works fine for  https://github.com/lensapp/lens (most probably `lens-ide` needs to upgrade some packages)

PR #4874, opened 2022-02-16
Fix: closing terminal tab error

aleksfront:
Adding one extra defence check.

Before:

https://user-images.githubusercontent.com/9607060/154215479-0b7b15a6-fa5d-4c3f-9ecd-581c64afe73a.mov

After:

https://user-images.githubusercontent.com/9607060/154215495-e871a484-c5c5-4764-8d6f-0e856c1d258c.mov


src/renderer/api/websocket-api.ts
```diff
@@ -181,7 +181,7 @@ export class WebSocketApi<Events extends WebSocketEvents> extends (EventEmitter
     if (error) {
       const { reconnectDelay } = this.params;
 
-      if (reconnectDelay) {
+      if (reconnectDelay && this.socket) {
```
panuhorsmalahti, line 184: Shouldn't we change the type to be something like `protected socket?: WebSocket | null;` to catch these errors with the compiler?

Nokel81 (approved):

ixrock (approved):

PR #4876, opened 2022-02-16
Introduce and use "openBrowser"

stevenjohnstone:
"openExternal" can result in security issues e.g.
https://nvd.nist.gov/vuln/detail/CVE-2022-0484. Added
a new function "openBrowser" and replaced some usage
of "openExternal". "openBrowser" when the intention is
to open a link in the default browser. Usage of
"openExternal" should be rare.

Signed-off-by: Steven Johnstone <sjohntone@mirantis.com>


src/common/utils/openExternal.ts
```diff
@@ -3,7 +3,12 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
 
-// Opens a link in external browser
+// Opens a link in an external program according to the URL scheme e.g.
+// "file" scheme may open arbitrary programs on Windows and Mac.
+//
+// SECURITY: This can open a wide range of programs with security impacts which
+// differ across platforms. If the intention is to open a link in the
+// default browser, use openBrowser
```
Nokel81: We could even do the opposite check here, where if the protocol is `http:` or `https:` then throw an error?

src/common/utils/openBrowser.ts
```diff
@@ -0,0 +1,23 @@
...
+ * http or https to prevent programs other than the default browser
+ * running.
+ *
+ * @param url The URL to open
+ * @returns Promise<void>
+ */
+import { shell } from "electron";
+
+export function openBrowser(url: string) {
+  if (!["http:", "https:"].includes(new URL(url).protocol)) {
```
Nokel81: Nit, I would factor this constant out and make it a set,

something like:

```
const allowedProtocols = new Set(["http:", "https:"]);
```

src/common/utils/openExternal.ts
stevenjohnstone: I think this would break current usage in extensions e.g. the MCC extension?

Nokel81: We could then issue a warning then and export `openBrowser` as well, couldn't we?

src/common/utils/openBrowser.ts
stevenjohnstone: Done, thanks.

Nokel81 (dismissed):

stefcameron:
@Nokel81 As-is, I don't see any obvious evidence in this PR that suggests this new function will be exported via the Lens Extensions package, so just checking that it will, since I'm assuming this is something we'd prefer extension authors to use (when appropriate)?

stevenjohnstone:
> @Nokel81 As-is, I don't see any obvious evidence in this PR that suggests this new function will be exported via the Lens Extensions package, so just checking that it will, since I'm assuming this is something we'd prefer extension authors to use (when appropriate)?

Yeah, I missed that, thanks.

src/common/utils/openBrowser.ts
```diff
@@ -0,0 +1,30 @@
...
+const allowedProtocols =  new Set(["http:", "https:"]);
+
+/**
+ * Opens a link using the program configured as the default browser
+ * on the target platform. Will reject URLs with a scheme other than
+ * http or https to prevent programs other than the default browser
+ * running.
+ *
+ * @param url The URL to open
+ * @returns Promise<void>
```
Nokel81: Given that we are using typescript, this sort of annotation isn't really necessary here.

If you want to mark it as such then I would add it to the end of line 19

```
...(url: string): Promise<void> {
```

Nokel81 (dismissed):

Nokel81 (approved):

ixrock (approved):
LGTM

PR #4899, opened 2022-02-18
Catalog striped lines placeholder

aleksfront:
* Adding catalog empty placeholder regarding to wireframes.
* Adding ability to set custom no-items-message for extensions.

https://user-images.githubusercontent.com/9607060/154911941-80204b93-5a38-49e3-83d6-7494c0ff82ef.mov


src/renderer/components/+catalog/catalog.tsx
```diff
@@ -263,6 +263,8 @@ class NonInjectedCatalog extends React.Component<Props & Dependencies> {
         isConfigurable={true}
         store={catalogEntityStore}
         getItems={() => catalogEntityStore.entities}
+        showEmptyTablePlaceholder
+        noItemsMessage="There are no Dev Clusters in this space"
```
jweak: default should not mention dev clusters

aleksfront: Good catch! Fixed.

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -250,6 +250,9 @@ class NonInjectedCatalog extends React.Component<Props & Dependencies> {
       ? `catalog-items-${activeCategory.metadata.name.replace(" ", "")}`
       : "catalog-items";
 
+    const noItemsMessage = activeCategory?.metadata.name == "Dev Clusters" ? "There are no Dev Clusters in this space"
```
jweak: Can we add this to the `CatalogCategory` (same as `icon` and `name`) metadata? So the category could define it's own `noItemsMessage`?

@panuhorsmalahti @Nokel81

Nokel81: Oh yes, this most certainly should not be special cased like this.

Nokel81: I would make it a "string" field and put it as part of the `spec` near `displayColumns`

jweak (dismissed):
Looks good!

msa0311:
Lets discuss before merging

src/renderer/components/item-object-list/content.tsx
```diff
@@ -41,6 +42,18 @@ interface ItemListLayoutContentProps<I extends ItemObject> {
...
+   * Message to display when no items available
+   *
+   * @default "Item list is empty"
+   */
+  noItemsMessage?: React.ReactNode;
+
+  /**
+   * Flag indicating striped lines should be displayed when no items provided
+   */
+  showEmptyTablePlaceholder?: boolean;
```
jakolehm, line 55: Could this be opt-out instead of opt-in? Then it would be consistent for all uses (I don't see why catalog should be any different from other lists that use this component).

src/renderer/components/item-object-list/content.tsx
```diff
@@ -201,7 +201,7 @@ export class ItemListLayoutContent<I extends ItemObject> extends React.Component
     if (this.props.showEmptyTablePlaceholder && this.props.renderTableHeader && this.props.tableId) {
       return (
         <>
-          <Placeholder renderTableHeader={this.props.renderTableHeader} tableId={this.props.tableId}/>
+          <Placeholder renderTableHeader={this.props.renderTableHeader} tableId={this.props.tableId} showCheckColumn={this.props.isSelectable}/>
```
Nokel81, line 204: nit, with so many props now I think they should each be on there own line.

Though maybe we should just enable the default of https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-first-prop-new-line.md

src/renderer/components/item-object-list/placeholder.tsx
```diff
@@ -6,24 +6,25 @@
...
 interface Props {
   renderTableHeader: TableCellProps[];
-  showExtraColumn?: boolean;
+  showActionsColumn?: boolean;
+  showCheckColumn?: boolean;
   tableId: string;
 }
 
-export function Placeholder({ renderTableHeader, showExtraColumn = true, tableId }: Props) {
+export function Placeholder({ renderTableHeader, showActionsColumn: showExtraColumn = true, showCheckColumn, tableId }: Props) {
```
Nokel81, line 19: nit, I think it would be better if we did the renaming on both sides.

src/common/catalog/catalog-entity.ts
```diff
@@ -81,6 +81,11 @@ export interface CatalogCategorySpec {
    * These columns will not be used in the "Browse" view.
    */
   displayColumns?: CategoryColumnRegistration[];
+
+  /**
+   * Custom message to display when no items available inside current category
```
Nokel81, line 86: I think we could improve this doc comment.

Something like:

```
This message will be displayed in the "single category view" when this category has no entities.
```

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -263,6 +265,8 @@ class NonInjectedCatalog extends React.Component<Props & Dependencies> {
         isConfigurable={true}
         store={catalogEntityStore}
         getItems={() => catalogEntityStore.entities}
+        showEmptyTablePlaceholder
```
Nokel81, line 268: Isn't showing by default true?

aleksfront:
Closing this because of:
1. The controversial nature of this PR.
2. New tables WIP which will conflict this changes.

PR #4920, opened 2022-02-22
Fix HelmChart validator requiring digest field

### LINKED ISSUE #4909: Issue when adding helm chart repositories. (helm 3.8.0)
Validation fails at lens and it is not possible to load charts. Digest was empty (null) and I manually added the value in order to make it visible in the charts list.

image

Is this an issue with helm?

Nokel81:

fixes #4909 

It does not seem that this field is actually required.


src/common/k8s-api/endpoints/helm-charts.api.ts
```diff
@@ -324,7 +324,11 @@ export class HelmChart {
   }
 
   getId(): string {
-    return `${this.repo}:${this.apiVersion}/${this.name}@${this.getAppVersion()}+${this.digest}`;
+    const digestPart = this.digest
+      ? `+${this.digest}`
+      : "";
+
+    return `${this.repo}:${this.apiVersion}/${this.name}@${this.getAppVersion()}${digestPart}`;
```
jim-docker, line 331: Does this risk the id not being unique (and would that be a problem)?

jim-docker (approved):

src/common/k8s-api/endpoints/helm-charts.api.ts
Nokel81: Given that we also have the `repo` in there I don't think that the IDs won't be unique anymore.

aleksfront (approved):
LGTM

PR #4924, opened 2022-02-23
Replace kubectl-proxy with lens-k8s-proxy

### LINKED ISSUE #230: Lens doesn't use MacOs DNS resolver
**Describe the bug**
Lens can't resolve DNS records when MacOs can.
We are connecting to development/production environments in AWS through VPN.
We are using private Route53 zones for internal DNS, so they are not visible worldwide.
Looks like Lens tries to resolve DNS using nameserver from /etc/resolv.conf which is usually router which not connected to VPN. 
At the same time, MacOs DNS resolver can resolve records through VPN.

**To Reproduce**
1. Connect to VPN with private DNS on Route53
2. Add a cluster by private DNS name

**Expected behavior**
The cluster should be connected successfully

**Screenshots**
Screen Shot 2020-04-08 at 12 16 50
78767216-f7ee5800-7992-11ea-8781-1bb4076e81ca

- Lens Version: 3.1.0
- OS: MacOs Catalina 10.15.3 (19D76)
- Installation method: dmg
 
**Logs:**
```
```

**Kubeconfig:**
```
apiVersion: v1
kind: Config
clusters:
- name: "dc1"
  cluster:
    server: "https://rancher.preprod.voicebase.com/k8s/clusters/c-56vd7"
    certificate-authority-data: ""
users:
- name: "dc1"
  user:
    token: ""

contexts:
- name: "dc1"
  context:
    user: "dc1"
    cluster: "dc1"
current-context: "dc1"
```

**Additional context**

### LINKED ISSUE #716: Cannot connect to k8s behind VPN on Mac OS
**Describe the bug**
The issue described: https://github.com/kubernetes/kubectl/issues/48
And solution already implemented in kubectl: https://github.com/kubernetes/release/issues/469

Kubectl works fine, but lens doesnt

On macOS, if a Go program is compiled with CGO_ENABLED=0 then Go uses its own internal network name resolver. This resolver only knows about /etc/resolv.conf and doesn’t know about the libSystem (macOS libc) library and its name resolution functions.

macOS (like most modern OSes) has smarter DNS lookups than just using an /etc/resolv.conf which allows it to smoothly handle switching networks.

When a VPN is being used then DNS lookups will be split between the VPN’s DNS servers and your ISP’s DNS servers, depending on the hostname.

You can set up custom name servers for certain domains by creating resolv.conf style files in /etc/resolver/<domain>. This is what your VPN software is doing behind the covers.

This issue can be reproduced by creating /etc/resolver/<domain> resolver for a particular cluster. This resolver must use different DNS servers than /etc/resolv.conf has. And only these different DNS servers have to resolve cluster fqdn.

**To Reproduce**
1. Install Lens on Mac OS
2. Connect to VPN network, such as corporate network
3. Try to connect to cluster behind VPN

Or

1. Install Lens on Mac OS
2. create /etc/resolver/<domain> resolver for a particular cluster
3. Try to connect to the cluster 

**Expected behavior**
The same behavior as for kubectl - I can work with cluster.

**Screenshots**

- Lens Version: 3.5.3
- OS: Mac OS
- Installation method (e.g. snap or AppImage in Linux): brew cask install lens
 
**Logs:**
```
Connecting ...
Authentication proxy started
2020/08/19 18:47:39 http: proxy error: dial tcp: lookup api.cluster.example.com on 192.168.1.1:53: no such host
```

### LINKED ISSUE #3101: Lens does not respect custom DNS resolvers on MacOS
**Describe the bug**
On MacOS you can config domain specific dns resolving via `/etc/resolver`. When creating a file in this directory with the name `my.domain.com` and `nameserver 15.15.15.15` for example, MacOS will DNS resolve only to this specific domain via nameserver 15.15.15.15.

**Expected behavior**
I have never had an isue with GUI's or CLI's with this setup before, but Lens seems to not respect this resolving, and will try to resolve targets (kubernetes api endpoint) via a public DNS. Might be hardcoded?

jakolehm:
[Lens K8s proxy](https://github.com/lensapp/lens-k8s-proxy) is purpose built proxy that is based on kubectl-proxy but can be used in more secure manner in Lens.

- options can be passed as environment variables (won't be visible in process list)
- supports TLS (not enabled in this PR)
- lens-k8s-proxy is not statically built, should fix few DNS problems in enterprise environments

Fixes #3101
Fixes #716
Fixes #230
Fixes https://github.com/lensapp/lens-security/issues/230


src/main/kube-auth-proxy/kube-auth-proxy.ts
```diff
@@ -45,23 +45,15 @@ export class KubeAuthProxy {
...
-    logger.debug(`spawning kubectl proxy with args: ${args}`);
-
-    this.proxyProcess = spawn(proxyBin, args, { env: this.env });
+    const proxyBin = this.dependencies.proxyBinPath;
+    
+    this.proxyProcess = spawn(proxyBin, [], { 
+      env: {
+        ...this.env,
+        KUBECONFIG: this.cluster.kubeConfigPath,
+        API_PREFIX: this.apiPrefix,
```
jakolehm, line 51: Still missing `--context` equivalent.

build/download_k8s_proxy.ts
```diff
@@ -0,0 +1,98 @@
...
+  public async download() {
+    if (await this.checkBinary()) {
+      return console.log("Already exists");
+    }
+
+    await ensureDir(path.dirname(this.path), 0o755);
+
+    const file = fs.createWriteStream(this.path);
+
+    console.log(`Downloading lens-k8s-version ${this.version} from ${this.url} to ${this.path}`);
```
jim-docker:
```suggestion
    console.log(`Downloading lens-k8s-proxy ${this.version} from ${this.url} to ${this.path}`);
```
?
few other places too

src/common/app-paths/directory-for-bundled-binaries/directory-for-bundled-binaries.injectable.ts
```diff
@@ -0,0 +1,21 @@
...
+import isDevelopmentInjectable from "../../vars/is-development.injectable";
+import contextDirInjectable from "../../vars/context-dir.injectable";
+
+const directoryForBundledBinariesInjectable = getInjectable({
+  instantiate: (di) => {
+    if (di.inject(isDevelopmentInjectable)) {
+      return path.join(di.inject(contextDirInjectable), "binaries");
+    }
+
+    return path.join(process.resourcesPath);
```
jim-docker:
```suggestion
    return path.join(process.resourcesPath, "binaries");
```
?

jakolehm: We don't want to have `"binaries"` here, removed unnecessary `path.join`.

build/download_k8s_proxy.ts
jakolehm: Good catch! Fixed.

build/download_k8s_proxy.ts
```diff
@@ -0,0 +1,98 @@
...
+      console.log(error);
+      fs.unlink(this.path, noop);
+      throw error;
+    });
+
+    return new Promise<void>((resolve, reject) => {
+      file.on("close", () => {
+        console.log("lens-k8s-proxy binary download closed");
+        fs.chmod(this.path, 0o755, (err) => {
+          if (err) reject(err);
```
chenhunghan, line 68: Do we want to `resolve` in the `else` block, to ensure `fs.chmod` is done before resolving this Promise?

Nokel81: Also wouldn't it be better to use `node-fetch` here instead of `request`?

jakolehm: Not very good excuse but this file is almost direct c&p from kubectl download. Maybe I should rewrite these using node-fetch (I think it should be relatively easy to do).

Nokel81: We could also just do that in a different PR too

jim-docker:
This PR appears to break port forwarding, investigating...
node and pod shells also don't work (seem to hang). Tried this on local clusters (minikube, docker-desktop)

jakolehm:
> This PR appears to break port forwarding, investigating... node and pod shells also don't work (seem to hang). Tried this on local clusters (minikube, docker-desktop)

@jim-docker noticed the same now (I'm pretty sure it worked originally), seeing this in the logs:
```
[0] error:   ▪ [LENS-PROXY]: Could not find cluster for upgrade request from url=/7d5380eef02a13f6/api/v1/namespaces/default/pods/nginx-1624377707-6c6444cb9f-6dxw9/exec?command=sh&command=-c&command=clear%3B+%28bash+%7C%7C+ash+%7C%7C+sh%29&container=nginx&stdin=true&stdout=true&tty=true +4s
```

Idk how this PR can affect resolving of a cluster object from a request URL... will investigate more.

jakolehm:
@jim-docker all functionality that requires http request upgrade should be now fixed (port-forward, exec to pod/node).

jakolehm:
Finally tests are green 🎉

nevalla (approved):
LGTM

src/main/kube-auth-proxy/create-kube-auth-proxy.injectable.ts
```diff
@@ -5,16 +5,18 @@
...
+import { isDevelopment, isWindows } from "../../common/vars";
+import directoryForBundledBinariesInjectable from "../../common/app-paths/directory-for-bundled-binaries/directory-for-bundled-binaries.injectable";
 
 const createKubeAuthProxyInjectable = getInjectable({
   id: "create-kube-auth-proxy",
 
   instantiate: (di) => {
-    const bundledKubectl = di.inject(bundledKubectlInjectable);
-
+    const binaryName = isWindows ? "lens-k8s-proxy.exe" : "lens-k8s-proxy";
```
Nokel81, line 16: Nit: should change changed to be dependencies (and thus injected) but we can do that work in the future.

Nokel81 (approved):
Other issues can be fixed in follow up

PR #4941, opened 2022-03-01
Enable TLS on lens-k8s-proxy

jakolehm:
Enables secure TLS connections from clients (mainly `LensProxy`) to `lens-k8s-proxy`. Follow-up PR will enable TLS also from clients (renderer etc.) to `LensProxy`.


src/main/kube-auth-proxy/kube-auth-proxy-cert.injectable.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import directoryForUserDataInjectable from "../../common/app-paths/directory-for-user-data/directory-for-user-data.injectable";
+import { createKubeAuthProxyCertificateFiles, getKubeAuthProxyCertificatePath } from "./kube-auth-proxy-cert";
+
+const getKubeAuthProxyCertDirInjectable = getInjectable({
+  id: "get-kube-auth-proxy-cert-dir",
+
+  setup: async (di) => {
+    await createKubeAuthProxyCertificateFiles(getKubeAuthProxyCertificatePath(await di.inject(directoryForUserDataInjectable)));
```
Iku-turso: Hmm. Add variables? One line of code should do one thing. Also code should be left-to-right readable :)

src/main/kube-auth-proxy/kube-auth-proxy-cert.injectable.ts
```diff
@@ -0,0 +1,19 @@
...
+  },
+
+  instantiate: (di) => getKubeAuthProxyCertificatePath(di.inject(directoryForUserDataInjectable)),
```
Iku-turso: Add variables.

src/main/kube-auth-proxy/kube-auth-proxy-cert.ts
```diff
@@ -0,0 +1,47 @@
...
+export async function createKubeAuthProxyCertificateFiles(dir: string): Promise<string> {
+  const cert = getKubeAuthProxyCertificate();
+
+  try {
+    await access(dir);
+  } catch {
+    await mkdir(dir);
+  }
+
+  await writeFile(path.join(dir, "proxy.key"), cert.private);
```
Iku-turso: Explicit side-effect `writeFile` should be a dependency instead.

src/main/kube-auth-proxy/kube-auth-proxy.ts
```diff
@@ -42,13 +43,15 @@ export class KubeAuthProxy {
     }
 
     const proxyBin = this.dependencies.proxyBinPath;
-    
-    this.proxyProcess = spawn(proxyBin, [], { 
+    const certPath = await this.dependencies.proxyCertPath;
+
+    this.proxyProcess = spawn(proxyBin, [], {
```
Iku-turso: Explicit side-effect `spawn` should be a dependency instead.

src/main/kube-auth-proxy/kube-auth-proxy-cert.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export function getKubeAuthProxyCertificatePath(baseDir: string) {
+  return path.join(baseDir, "kube-auth-proxy");
+}
+
+export async function createKubeAuthProxyCertificateFiles(dir: string): Promise<string> {
+  const cert = getKubeAuthProxyCertificate();
+
+  try {
```
Iku-turso: Avoidable try/catch as part of business-logic, which is a code-smell. Try/catch is best reserved for fatal errors, and using it for something else damages credibility of eg. centralized error handling.

src/main/kube-auth-proxy/kube-auth-proxy-cert.ts
```diff
@@ -0,0 +1,47 @@
...
+    await access(dir);
```
Iku-turso: Explicit side-effect `access` should be a dependency instead.

src/main/kube-auth-proxy/kube-auth-proxy-cert.ts
```diff
@@ -0,0 +1,47 @@
...
+  } catch {
+    await mkdir(dir);
```
Iku-turso: Explicit side-effect `mkdir` should be a dependency instead.

src/main/kube-auth-proxy/kube-auth-proxy-cert.ts
```diff
@@ -0,0 +1,47 @@
...
+export function getKubeAuthProxyCertificatePath(baseDir: string) {
+  return path.join(baseDir, "kube-auth-proxy");
+}
+
+export async function createKubeAuthProxyCertificateFiles(dir: string): Promise<string> {
+  const cert = getKubeAuthProxyCertificate();
+
+  try {
+    await access(dir);
+  } catch {
```
Iku-turso: "catch all" could be "catch specific" if decision is to use `catch` here.

src/main/context-handler/context-handler.ts
```diff
@@ -132,10 +135,21 @@ export class ContextHandler {
   }
 
   protected async newApiTarget(timeout: number): Promise<httpProxy.ServerOptions> {
+    await this.ensureServer();
```
Iku-turso, line 141: Explicit side-effect `ensureServer` should be a dependency instead.

src/main/kube-auth-proxy/kube-auth-proxy-cert.ts
```diff
@@ -0,0 +1,47 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import { access, mkdir, writeFile } from "fs/promises";
+import path from "path";
+import * as selfsigned from "selfsigned";
+
+export function getKubeAuthProxyCertificate(): selfsigned.SelfSignedCert {
```
Iku-turso: Does not require to be a function?

src/main/kube-auth-proxy/kube-auth-proxy-cert.ts
```diff
@@ -0,0 +1,47 @@
...
+  const opts = [
+    { name: "commonName", value: "Lens Certificate Authority" },
+    { name: "organizationName", value: "Lens" },
+  ];
+
+  return selfsigned.generate(opts, {
```
Iku-turso: Is `selfSigned.generate` deterministic? If not, it is an explicit side-effect, and should be a dependency instead.

src/main/kube-auth-proxy/create-kube-auth-proxy.injectable.ts
```diff
@@ -17,6 +18,7 @@ const createKubeAuthProxyInjectable = getInjectable({
     const proxyPath = isDevelopment ? path.join("client", process.platform, process.arch) : process.arch;
     const dependencies = {
       proxyBinPath: path.join(di.inject(directoryForBundledBinariesInjectable), proxyPath, binaryName),
+      proxyCertPath: di.inject(getKubeAuthProxyCertDirInjectable),
```
Iku-turso: Name `getKubeAuthProxyCertDirInjectable` implies function, but it is used as a value here.

src/main/kube-auth-proxy/kube-auth-proxy-cert.ts
```diff
@@ -0,0 +1,47 @@
...
+      { name: "basicConstraints", cA: true },
+      { name: "subjectAltName", altNames: [
+        { type: 2, value: "localhost" },
+        { type: 7, ip: "127.0.0.1" },
+      ] },
+    ],
+  });
+}
+
+export function getKubeAuthProxyCertificatePath(baseDir: string) {
```
Iku-turso: Thought: a convention is to use `function` to advertise that `this` has special significance. Here it does not, and lambda could be used to advertise that instead.

src/main/kube-auth-proxy/kube-auth-proxy-cert.injectable.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import directoryForUserDataInjectable from "../../common/app-paths/directory-for-user-data/directory-for-user-data.injectable";
+import { createKubeAuthProxyCertificateFiles, getKubeAuthProxyCertificatePath } from "./kube-auth-proxy-cert";
+
+const getKubeAuthProxyCertDirInjectable = getInjectable({
```
Iku-turso: Name of `injectable` does not document unrelated stuff done in `setup`.

Iku-turso: This could be wonderfully unit-testable either locally, or preferably as part of a larger system ;)

src/main/context-handler/context-handler.ts
```diff
@@ -27,6 +29,7 @@ interface PrometheusServicePreferences {
 
 interface Dependencies {
   createKubeAuthProxy: (cluster: Cluster, environmentVariables: NodeJS.ProcessEnv) => KubeAuthProxy;
+  certPath: string;
 }
 
 export class ContextHandler {
```
Iku-turso, line 34: Note: some unrelated janky code around these parts needs a lovin' champion at some.

src/main/kube-auth-proxy/create-kube-auth-proxy.injectable.ts
jakolehm: What would be better naming convention for values? `createKubeAuthProxyCertDirInjectable`?

src/main/context-handler/context-handler.ts
jakolehm: Might be out-of-scope for this PR (not sure yet).

src/main/kube-auth-proxy/kube-auth-proxy-cert.injectable.ts
jakolehm: What would be better naming convention for injectables that have setup?

src/main/kube-auth-proxy/kube-auth-proxy-cert.injectable.ts
jakolehm: Yes, will cleanup this PR before removing "draft".

src/main/kube-auth-proxy/kube-auth-proxy-cert.ts
jakolehm: Not 100% deterministic because cert creation always creates an unique certificate. That said, it has deterministic fields... not sure where we should draw the line here.

src/main/kube-auth-proxy/kube-auth-proxy-cert.ts
Nokel81: I would disagree, this codebase only uses `class`-es for `this` significant code and lambdas can sometimes make the code harder to read.

src/main/index.ts
```diff
@@ -62,9 +62,11 @@ const di = getDi();
 
 app.setName(appName);
 
-di.runSetups().then(() => {
-  injectSystemCAs();
+app.on("ready", async () => {
```
jakolehm, line 65: App start order was previously wrong... any async setups caused `app.on("ready")` registration happen too late -> we missed it completely.

src/main/proxy-functions/kube-api-upgrade-request.ts
```diff
@@ -4,21 +4,25 @@
...
-import net from "net";
+import tls from "tls";
 import url from "url";
 import { apiKubePrefix } from "../../common/vars";
 import type { ProxyApiRequestArgs } from "./types";
 
 const skipRawHeaders = new Set(["Host", "Authorization"]);
 
-export async function kubeApiRequest({ req, socket, head, cluster }: ProxyApiRequestArgs) {
+export async function kubeApiUpgradeRequest({ req, socket, head, cluster }: ProxyApiRequestArgs) {
```
jakolehm, line 14: Renamed because name was misleading, only upgrade requests are handled here.

src/main/context-handler/context-handler.ts
```diff
@@ -12,6 +12,9 @@ import url, { UrlWithStringQuery } from "url";
 import { CoreV1Api } from "@kubernetes/client-node";
 import logger from "../logger";
 import type { KubeAuthProxy } from "../kube-auth-proxy/kube-auth-proxy";
+import path from "path";
+import { readFile } from "fs/promises";
```
Nokel81: This should probably be a dependency

src/main/context-handler/context-handler.ts
```diff
@@ -118,7 +122,11 @@ export class ContextHandler {
     await this.ensureServer();
     const path = this.clusterUrl.path !== "/" ? this.clusterUrl.path : "";
 
-    return `http://127.0.0.1:${this.kubeAuthProxy.port}${this.kubeAuthProxy.apiPrefix}${path}`;
+    return `https://127.0.0.1:${this.kubeAuthProxy.port}${this.kubeAuthProxy.apiPrefix}${path}`;
+  }
+
+  async resolveAuthProxyCa() {
+    return readFile(path.join(this.dependencies.certPath, "proxy.crt"));
```
Nokel81: If this file never changes I think it should be a setup-able dependency.

src/main/index.ts
Nokel81: So that we can catch these async errors I would suggest a slightly different shape:

```
async function main() {
  const di = getDi();

  await app.whenReady();
  await di.runSetups();

  ...
}

main().catch(error => {
  console.error(error);
  dialog.showErrorBox("Lens Startup Failed", String(error));
  app.exit(1);
});
```

src/main/kube-auth-proxy/create-kube-auth-proxy-cert-files.injectable.ts
```diff
@@ -0,0 +1,28 @@
...
+import { getKubeAuthProxyCertificatePath } from "./kube-auth-proxy-cert";
+import * as selfsigned from "selfsigned";
+import { createKubeAuthProxyCertFiles } from "./create-kube-auth-proxy-cert-files";
+
+const createKubeAuthProxyCertFilesInjectable = getInjectable({
+  id: "create-kube-auth-proxy-cert-files",
+
+  instantiate: async (di) => {
+    const userData = di.inject(directoryForUserDataInjectable);
+    const certPath = getKubeAuthProxyCertificatePath(userData);
```
Nokel81: This sounds more like a dependency which should depend on `directoryForUserDataInjectable`.

src/main/kube-auth-proxy/create-kube-auth-proxy-cert-files.injectable.ts
```diff
@@ -0,0 +1,28 @@
...
+    
+    return createKubeAuthProxyCertFiles(certPath, {
```
Nokel81, line 19: Given that this creates the files and then returns `certPath` (which coincendentally should probably be called something like `directoryForProxyCertificates`. I think that this injectable should actually be mostly just `setup` and then instantiate should return `undefined` or maybe a new `SetupOnly` marker type.

Then whatever depends on `createKubeAuthProxyCertFilesInjectable` could instead just depend on `directoryForProxyCertificatesInjectable` and know that the setup and creation of the files will already have been done.

src/main/kube-auth-proxy/create-kube-auth-proxy.injectable.ts
```diff
@@ -4,18 +4,24 @@
...
 const createKubeAuthProxyInjectable = getInjectable({
   id: "create-kube-auth-proxy",
 
-  instantiate: (di) => {
+  instantiate: (di): CreateKubeAuthProxy => {
     const binaryName = getBinaryName("lens-k8s-proxy");
     const dependencies: KubeAuthProxyDependencies = {
       proxyBinPath: path.join(di.inject(directoryForBundledBinariesInjectable), binaryName),
+      proxyCertPath: di.inject(getKubeAuthProxyCertDirInjectable),
+      spawn,
```
Nokel81: This should be its own injectable, maybe within a folder `src/main/child-process` since we probably shouldn't have any of those calls within `renderer`

types/selfsigned.d.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+ declare module "selfsigned" {
```
Nokel81, line 6: I think it would be better if these were hosted into `@types`. The PR process is actually really simple, I have done it a couple of times.

src/main/context-handler/context-handler.ts
jakolehm: Refactored. Now `authProxyCa` is a dependency (but still a promise because of `createContextHandlerInjectable` instantiate has to be non-async for now). This way `ContextHandler` does not need to know `readFile` at all.

src/main/context-handler/context-handler.ts
jakolehm: It is out of scope, topic for an another PR.

src/main/kube-auth-proxy/create-kube-auth-proxy-cert-files.injectable.ts
jakolehm: Indeed, there was already an injectable for this. Refactored.

src/main/kube-auth-proxy/create-kube-auth-proxy.injectable.ts
jakolehm: Done.

types/selfsigned.d.ts
jakolehm: It would be better yes but I think it can be separately from this PR. Note: this declaration is probably not complete, it only touches parts this PR needs.

jakolehm:
@Nokel81 PTAL again.

src/main/lens-proxy.ts
```diff
@@ -61,7 +61,7 @@ export class LensProxy extends Singleton {
 
   public port: number;
 
-  constructor(protected router: Router, protected proxy: httpProxy, { shellApiRequest, kubeApiRequest, getClusterForRequest }: LensProxyFunctions) {
+  constructor(protected router: Router, protected proxy: httpProxy, { shellApiRequest, kubeApiUpgradeRequest, getClusterForRequest }: LensProxyFunctions) {
```
Nokel81, line 64: nit: all of these feel like dependencies. Should we consolidate them?

src/renderer/components/delete-cluster-dialog/__tests__/delete-cluster-dialog.test.tsx
```diff
@@ -91,6 +92,8 @@ describe("<DeleteClusterDialog />", () => {
   beforeEach(async () => {
     const { mainDi, runSetups } = getDisForUnitTesting({ doGeneralOverrides: true });
 
+    mainDi.override(createContextHandlerInjectable, () => () => undefined);
```
Nokel81, line 95: Should this be done in `getDiForUnitTesting` so that it is always run by default?

src/main/lens-proxy.ts
jakolehm: Probably yes but sounds like out of scope for this PR?

src/renderer/components/delete-cluster-dialog/__tests__/delete-cluster-dialog.test.tsx
jakolehm: It cannot be there because some tests do require the real implementation.

src/main/lens-proxy.ts
Nokel81: Sure

src/renderer/components/delete-cluster-dialog/__tests__/delete-cluster-dialog.test.tsx
Nokel81: Oh yes, that is why I would put it behind some sort of flag (like `doGeneralOverrides` is). I suggested moving it because I think that the vast majority of tests would want it overridden.

Nokel81 (approved):
We can change how certain injectables are overridden by default in a future PR

nevalla (approved):

PR #4942, opened 2022-03-01
Consolidate downloading binaries into one script

### LINKED ISSUE #4957: lens-k8s-proxy binary not found on Windows
Screen Shot 2022-03-02 at 5 37 57 PM

Binary is in `C:\Users\jiim\git\lensapp\lens\binaries\client\windows\x64` on my system (`windows` not `win32`)

master branch, dev build

Nokel81:
- Consolidate downloading bundled binaries into one script
- Remove `LensBinary` as it was only used by the `HelmCli` code which has now also been removed.
- Introduce a temporary `lazyInitialized` to bypass global state problems and the fact that `webpack.*.ts` files are importing application code


package.json
```diff
@@ -48,7 +45,7 @@
     "postversion": "git push --set-upstream ${GIT_REMOTE:-origin} release/v$npm_package_version"
   },
   "config": {
-    "k8sProxyVersion": "0.1.2",
+    "k8sProxyVersion": "0.1.5",
```
jakolehm: This change probably should not be part of this PR?

src/main/helm/helm-cli.ts
```diff
@@ -3,47 +3,20 @@
...
-
-if (!isProduction) {
-  baseDir = path.join(process.cwd(), "binaries", "client", process.arch);
-}
-
-export const helmCli = new HelmCli(baseDir, helmVersion);
+/**
+ * @deprecated use `helmBinaryPath` or its injection equivalent instead
+ */
+export const helmCli = {
```
jakolehm: Any reason to leave deprecated code here? I assume this is not exposed via extension api...

Nokel81: No and fair enough, I wanted to show that I could make the API the same for the rest of the repo, but I can change this.

package.json
Nokel81: I can make that a separate PR yes.

Nokel81: PR is now up https://github.com/lensapp/lens/pull/4946

jakolehm: Idk why this change is still visible here? Maybe rebase with master?

src/common/vars.ts
```diff
@@ -29,6 +30,54 @@ export const defaultTheme = "lens-dark" as string;
...
+
+export function getBinaryName(name: string, { forPlatform = normalizedPlatform } = {}): string {
+  if (forPlatform === "windows") {
+    return `${name}.exe`;
+  }
+
+  return name;
+}
+
+export const resourcesDir = onceCell(() => (
```
jakolehm: Why do we need `onceCell`?

Nokel81: Because `process` is not defined when the `vars` are imported by the `webpack.*.ts` files and therefore throw. This allows for the code of instantiation to happen at first use not at definition time.

package.json
Nokel81: Yes I should rebase with master.

build/download_binaries.ts
```diff
@@ -0,0 +1,252 @@
...
+import { pipeline as _pipeline, Transform, Writable } from "stream";
+import { MultiBar, SingleBar } from "cli-progress";
+import AbortController from "abort-controller";
+import { extract } from "tar-stream";
+import gunzip from "gunzip-maybe";
+import { getBinaryName, normalizedPlatform } from "../src/common/vars";
+
+const pipeline = promisify(_pipeline);
+
+interface DownloadK8sProxyArgs {
```
jim-docker: why this name? Seems confusing

jim-docker (commented):
Something fishy:
Screen Shot 2022-03-04 at 5 28 32 PM

src/common/utils/once-cell.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  return {
+    get() {
+      if (called) {
+        return value;
+      }
+
+      called = true;
+
+      return value = builder();
```
jim-docker: what if `builder()` throws? (set `called` to true after?)

build/download_binaries.ts
Nokel81: Yes, good catch. Shows the history of the implementation.

Nokel81:
Weird, I tested listing helm repos which did work

src/common/utils/once-cell.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+/**
+ * A OnceCell is an object that wraps some function that produces a value.
+ *
+ * It then only calls the function on the first call to `get()` and returns the
+ * same instance/value on every subsequent call.
+ */
+export interface OnceCell<T> {
```
jim-docker: feels like deferred should be in the name. `DeferredGet`? `LazyGet`?

src/common/vars.ts
jim-docker: Global state problem? @jansav

src/common/vars.ts
```diff
@@ -29,6 +30,54 @@ export const defaultTheme = "lens-dark" as string;
...
+  isProduction
+    ? process.resourcesPath
+    : path.join(process.cwd(), "binaries", "client", normalizedPlatform)
+));
+
+export const baseBinariesDir = onceCell(() => path.join(resourcesDir.get(), normalizedArch));
+
+export const helmBinaryName = getBinaryName("helm");
+export const helmBinaryPath = onceCell(() => path.join(resourcesDir.get(), helmBinaryName));
```
jim-docker:
```suggestion
export const helmBinaryPath = onceCell(() => path.join(baseBinariesDir.get(), helmBinaryName));
```
?

jim-docker:
> Weird, I tested listing helm repos which dog work

You probably made a change after testing that?

Nokel81:
Yes it is a global state issue. But also it is an issue with the `webpack.*.ts` files importing program code.

I decided against changing them to DI in this PR because I know it is coming and felt this was a lower impact fix.

src/common/vars.ts
Nokel81: Yes global state, but mainly we should not be importing application code in our build files like this.

I think a PR to remove the imports of `common/vars` from our webpack files would be a good thing.

jakolehm (dismissed):

jakolehm (dismissed):

jim-docker (commented):
rename `once-cell.ts`?

jakolehm (approved):

jim-docker (approved):

PR #4956, opened 2022-03-02
Fix reset select value

DmitriyNoa:
The select value should be set to null if no value is found. Currently select will return undefined if no values are selected. This causes the reach select to store old selected value even after it was cleared from props.


Nokel81:
Would you mind adding a test to show why this is a problem?

DmitriyNoa:
@Nokel81 Added basic tests including the test case

src/renderer/components/select/select.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+    const onChange = jest.fn();
+
+    const { container, rerender } = render(<Select value={options[0].value} onChange={onChange} options={options} />);
+    const selectedValueContainer = container.querySelector(".Select__single-value");
+
+    expect(selectedValueContainer.textContent).toBe(options[0].label);
+
+    rerender(<Select value={null} onChange={onChange} options={options} />);
+
+    expect(selectedValueContainer).not.toBeInTheDocument();
```
Nokel81: nit, I would explicitly call `querySelector` again here because it might be that a different container was produced.

Nokel81 (dismissed):

Nokel81 (dismissed):

DmitriyNoa:
Closing in favor of https://github.com/lensapp/lens/pull/4961

PR #4959, opened 2022-03-03
Fix api-manager kubeobject types

jakolehm:
Currently it's not possible to have a `MyClass extends KubeObject` with additional methods because `ApiManager` requires objects to be exactly `KubeObject`. ~`ApiManager` should not care about `KubeObject` types within `KubeApi` / `KubeObjectStore` because it's not possible to create those classes with invalid types.~

Needs a separate backport PR to `release/5.4` branch once merged.

### DIFF
--- a/src/common/k8s-api/api-manager.ts
+++ b/src/common/k8s-api/api-manager.ts
@@ -79,7 +79,7 @@ export class ApiManager {
     return this.stores.get(this.resolveApi(api)?.apiBase) as S;
   }
 
-  lookupApiLink<T extends KubeObject = KubeObject>(ref: IKubeObjectRef, parentObject?: T): string {
+  lookupApiLink(ref: IKubeObjectRef, parentObject?: KubeObject): string {
     const {
       kind, apiVersion, name,
       namespace = parentObject?.getNs(),
--- a/src/common/k8s-api/api-manager.ts
+++ b/src/common/k8s-api/api-manager.ts
@@ -69,7 +69,7 @@ export class ApiManager<TKubeObject extends KubeObject = any> {
   }
 
   @action
-  registerStore(store: KubeObjectStore<any>, apis: KubeApi<any>[] = [store.api]) {
+  registerStore(store: KubeObjectStore<TKubeObject>, apis: KubeApi<TKubeObject>[] = [store.api]) {
     apis.filter(Boolean).forEach(api => {
       if (api.apiBase) this.stores.set(api.apiBase, store);
     });
--- a/src/common/k8s-api/api-manager.ts
+++ b/src/common/k8s-api/api-manager.ts
@@ -11,16 +11,16 @@ import type { KubeApi } from "./kube-api";
 import type { KubeObject } from "./kube-object";
 import { IKubeObjectRef, parseKubeApi, createKubeApiURL } from "./kube-api-parse";
 
-export class ApiManager {
-  private apis = observable.map<string, KubeApi<any>>();
-  private stores = observable.map<string, KubeObjectStore<any>>();
+export class ApiManager<TKubeObject extends KubeObject = any> {
+  private apis = observable.map<string, KubeApi<TKubeObject>>();
+  private stores = observable.map<string, KubeObjectStore<TKubeObject>>();
 
   constructor() {
     makeObservable(this);
     autoBind(this);
   }
 
-  getApi(pathOrCallback: string | ((api: KubeApi<any>) => boolean)) {
+  getApi(pathOrCallback: string | ((api: KubeApi<TKubeObject>) => boolean)) {
     if (typeof pathOrCallback === "string") {
       return this.apis.get(pathOrCallback) || this.apis.get(parseKubeApi(pathOrCallback).apiBase);
     }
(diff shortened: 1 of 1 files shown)


nevalla (dismissed):

samitiilikainen (dismissed):

src/common/k8s-api/api-manager.ts
```diff
@@ -11,16 +11,16 @@ import type { KubeApi } from "./kube-api";
 import type { KubeObject } from "./kube-object";
 import { IKubeObjectRef, parseKubeApi, createKubeApiURL } from "./kube-api-parse";
 
-export class ApiManager {
-  private apis = observable.map<string, KubeApi<KubeObject>>();
-  private stores = observable.map<string, KubeObjectStore<KubeObject>>();
+export class ApiManager<TKubeObject extends KubeObject = any> {
```
Nokel81: Please no defaulting to `any`, or even better we should try and remove all uses of `any` since it basically is no better than JS.

src/common/k8s-api/api-manager.ts
```diff
@@ -69,17 +69,17 @@ export class ApiManager {
...
     });
   }
 
-  getStore<S extends KubeObjectStore<KubeObject>>(api: string | KubeApi<KubeObject>): S | undefined {
+  getStore<S extends KubeObjectStore<TKubeObject>>(api: string | KubeApi<TKubeObject>): S | undefined {
     return this.stores.get(this.resolveApi(api)?.apiBase) as S;
   }
 
-  lookupApiLink(ref: IKubeObjectRef, parentObject?: KubeObject): string {
+  lookupApiLink<T extends KubeObject = KubeObject>(ref: IKubeObjectRef, parentObject?: T): string {
```
Nokel81: This is an "unused type param" and therefore shouldn't be used. 

See: https://github.com/microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md

src/common/k8s-api/api-manager.ts
```diff
@@ -46,19 +46,19 @@ export class ApiManager {
...
     if (typeof api === "string") {
-      return this.getApi(api) as KubeApi<K>;
+      return this.getApi(api) as KubeApi<TKubeObject>;
     }
 
     return api;
   }
 
-  unregisterApi(api: string | KubeApi<KubeObject>) {
+  unregisterApi(api: string | KubeApi<any>) {
```
Nokel81: nit: remove any

src/common/k8s-api/api-manager.ts
```diff
@@ -46,19 +46,19 @@ export class ApiManager {
...
 
-  protected resolveApi<K extends KubeObject>(api?: string | KubeApi<K>): KubeApi<K> | undefined {
+  protected resolveApi(api?: string | KubeApi<TKubeObject>): KubeApi<TKubeObject> | undefined {
     if (!api) {
       return undefined;
     }
 
     if (typeof api === "string") {
-      return this.getApi(api) as KubeApi<K>;
+      return this.getApi(api) as KubeApi<TKubeObject>;
```
Nokel81: Shouldn't need a cast here if `getApi` has the correct typings.

src/common/k8s-api/api-manager.ts
```diff
@@ -69,17 +69,17 @@ export class ApiManager {
...
   @action
-  registerStore(store: KubeObjectStore<KubeObject>, apis: KubeApi<KubeObject>[] = [store.api]) {
+  registerStore(store: KubeObjectStore<TKubeObject>, apis: KubeApi<TKubeObject>[] = [store.api]) {
     apis.filter(Boolean).forEach(api => {
       if (api.apiBase) this.stores.set(api.apiBase, store);
     });
   }
 
-  getStore<S extends KubeObjectStore<KubeObject>>(api: string | KubeApi<KubeObject>): S | undefined {
+  getStore<S extends KubeObjectStore<TKubeObject>>(api: string | KubeApi<TKubeObject>): S | undefined {
```
Nokel81: So is this, but that would require marking one of the versions with `@deprecated` which should be fine.

See: https://github.com/microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md

src/common/k8s-api/api-manager.ts
jakolehm: After discussing via Slack I completely agree, will change this PR to have proper types 👍

jakolehm: Fixed 🎉

src/common/k8s-api/api-manager.ts
```diff
@@ -46,13 +46,13 @@ export class ApiManager {
...
 
-  protected resolveApi<K extends KubeObject>(api?: string | KubeApi<K>): KubeApi<K> | undefined {
+  protected resolveApi(api?: string | KubeApi<KubeObject>): KubeApi<KubeObject> | undefined {
     if (!api) {
       return undefined;
     }
 
     if (typeof api === "string") {
-      return this.getApi(api) as KubeApi<K>;
+      return this.getApi(api) as KubeApi<KubeObject>;
```
Nokel81, line 55: nit: we can remove this cast

Nokel81 (approved):

nevalla (approved):

PR #4961, opened 2022-03-03
Fix unselect. Add unit tests

DmitriyNoa:
- Fix Select component not resetting value when null is passed.
- Add unit tests

### DIFF
--- a/src/renderer/components/select/select.test.tsx
+++ b/src/renderer/components/select/select.test.tsx
@@ -1,0 +1,137 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import React from "react";
+import "@testing-library/jest-dom/extend-expect";
+import { Select } from "./select";
+import { UserStore } from "../../../common/user-store";
+import { ThemeStore } from "../../theme.store";
+import { getDiForUnitTesting } from "../../getDiForUnitTesting";
+import type { DiContainer } from "@ogre-tools/injectable";
+import { DiRender, renderFor } from "../test-utils/renderFor";
+import mockFs from "mock-fs";
+import directoryForUserDataInjectable
+  from "../../../common/app-paths/directory-for-user-data/directory-for-user-data.injectable";
+import rendererExtensionsInjectable from "../../../extensions/renderer-extensions.injectable";
+import { computed } from "mobx";
+import type { LensRendererExtension } from "../../../extensions/lens-renderer-extension";
+
+
+describe("<Select />", () => {
+  let di: DiContainer;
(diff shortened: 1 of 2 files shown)


Nokel81 (dismissed):

src/renderer/components/select/select.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+    rerender(<Select value={options[1].value} onChange={onChange} options={options} />);
+
+    expect(container.querySelector(".Select__single-value").textContent).toBe(options[1].label);
+
+    rerender(<Select value={null} onChange={onChange} options={options} />);
+
+    expect(container.querySelector(".Select__single-value")).not.toBeInTheDocument();
+  });
+
+  it("should unselect value if no value is passed", async () => {
```
jim-docker: how is this test different from what's done on lines 109-111?

src/renderer/components/select/select.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+    ];
+
+    const onChange = jest.fn();
+
+    const { container, rerender } = render(<Select value={options[0].value} onChange={onChange} options={options} />);
+    const selectedValueContainer = container.querySelector(".Select__single-value");
+
+    expect(selectedValueContainer.textContent).toBe(options[0].label);
+
+    rerender(<Select value={null} onChange={onChange} options={options} />);
```
jim-docker: value of `null` here means nothing is selected? What was happening when `selectedOption` was `undefined`?

jim-docker (dismissed):

src/renderer/components/select/select.test.tsx
DmitriyNoa: Yes. If an undefined was passed the old selected value was still selected, at least the label wasn't cleared.

Nokel81: Maybe it should, since both values are not in the list of options.

DmitriyNoa: Then it means there is no way to clear the selection programmatically?

Nokel81: How so? I am just proposing to add a new test case to cover `value={undefined}`. Or do you think that shouldn't also work to clear the value?

src/renderer/components/select/select.test.tsx
DmitriyNoa: Cleaned that

src/renderer/components/select/select.test.tsx
DmitriyNoa: Sorry, got it wrong. Added also test for undefined

Nokel81 (approved):

jim-docker (approved):

PR #4970, opened 2022-03-04
Remove listener for online event

### LINKED ISSUE #2434: Blank dashboard after sleep
Leaving Lens running/open when computer sleeps results in blank dashboard for the cluster that was connected, upon awakening the computer.

https://user-images.githubusercontent.com/40840436/113300835-5a3f1780-92cc-11eb-9c31-02be3934c960.mov

Dashboard is recovered for the the cluster by disconnecting it and then reconnecting it (Or restarting Lens)

Lens 4.2.0, observed on Windows 10

jim-docker:
On Windows only, as a result of the system sleeping, `window.onbeforeunload()` is called for the visible cluster dashboard (not sure why). The behaviour before this PR was to simply remove the root element of the cluster frame. This leaves Lens in limbo as the cluster is still connected but the dashboard is gone. This PR initiates a cluster disconnect instead of removing the root element. So when the windows system is awakened the user is greeted with the catalog rather than a blank dashboard.

Fixes #2434 

Might still want to investigate why `window.onbeforeunload()` is getting called during sleep.

UPDATE:
- restored the `window.onbeforeunload()` behaviour as it was relied upon during user reloads (View menu -> Reload)
- removed the listener for online event, verifying that
  - it doesn't impact cluster loading behaviour (doesn't appear to be necessary given other listeners)
  - it did get triggered on windows during sleep/resume and is responsible for triggering the problematic call to `window.onbeforeunload()`

This should soak in case I missed any reload behaviour being impacted.

### DIFF
--- a/src/renderer/frames/cluster-frame/init-cluster-frame/init-cluster-frame.ts
+++ b/src/renderer/frames/cluster-frame/init-cluster-frame/init-cluster-frame.ts
@@ -10,10 +10,9 @@ import { Notifications } from "../../../components/notifications";
 import type { AppEvent } from "../../../../common/app-event-bus/event-bus";
 import type { CatalogEntity } from "../../../../common/catalog";
 import { when } from "mobx";
-import { unmountComponentAtNode } from "react-dom";
 import type { ClusterFrameContext } from "../../../cluster-frame-context/cluster-frame-context";
 import { KubeObjectStore } from "../../../../common/k8s-api/kube-object.store";
-import { requestSetClusterFrameId } from "../../../ipc";
+import { requestClusterDisconnection, requestSetClusterFrameId } from "../../../ipc";
 
 interface Dependencies {
   hostedCluster: Cluster;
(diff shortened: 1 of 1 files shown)


Nokel81 (dismissed):

src/renderer/frames/cluster-frame/init-cluster-frame/init-cluster-frame.ts
```diff
@@ -84,7 +83,7 @@ export const initClusterFrame =
           `${logPrefix} Unload dashboard, clusterId=${(hostedCluster.id)}, frameId=${frameRoutingId}`,
         );
 
-        unmountComponentAtNode(rootElem);
+        requestClusterDisconnection(hostedCluster.id);
```
jakolehm: Unmount was here on purpose, without unmount Lens might leak watch etc requests because they happen within nodejs runtime (not within browser runtime).

jakolehm: Also: I think `onbeforeunload` happens when a page is reloaded.. we probably don't want to disconnect on every reload.

jim-docker: Good catch, I will find another way to fix

Nokel81 (approved):

jakolehm (approved):

PR #4973, opened 2022-03-05
Add 'Sync with computer' theme option

### LINKED ISSUE #2478: Option for theme to match system theme

An option to allow the Lens theme to match the system theme (on systems where there is such a concept).

**Why is this needed**:

It's a fairly trivial request, but I have my system configured to switch automatically between the light theme (during the day) and dark theme (during the night), which I find more restful for my eyes. (Especially when needing to fix something in the small hours, when a sudden burst of light theme can be... disconcerting, to say the least.) Given that this is automated, it would be most convenient for me if all applications allowing light/dark theming had the option to follow along with these changes to the system theme.

- Kubernetes distribution: k3s 1.20.4_+k3s1
- Desktop OS: Windows 10

aleksfront:
Now theme can be synced with OS settings and switched automatically.

https://user-images.githubusercontent.com/9607060/156878187-3888b2c3-4caf-4d0e-9929-ae036f9a8f81.mov

Fixes #2478


Nokel81 (approved):
Looks good, I have code to rework `ThemeStore` to use DI so no need to do that here.

aleksfront:
> Looks good, I have code to rework `ThemeStore` to use DI so no need to do that here.

Yes, I agree. Ideally, we should pass `ipcRenderer` in a `ThemeStore` constructor.

src/renderer/theme.store.ts
```diff
@@ -34,6 +36,8 @@ export class ThemeStore extends Singleton {
     "lens-light": lensLightThemeJson as Theme,
   });
 
+  @observable osNativeTheme: "dark" | "light" | undefined;
```
DmitriyNoa, line 39: We are using "dark" | "light" many times. Maybe move to a type?

DmitriyNoa (approved):

src/renderer/theme.store.ts
```diff
@@ -43,7 +47,7 @@ export class ThemeStore extends Singleton {
   }
 
   @computed get activeTheme(): Theme {
-    return this.themes.get(this.activeThemeId) ?? this.themes.get(defaultTheme);
+    return this.systemTheme ?? this.themes.get(this.activeThemeId) ?? this.themes.get(defaultTheme);
```
DmitriyNoa, line 50: Are you sure about using nullish coalescing here? Why not simple || ?

Nokel81: I would argue that `??` is more correct because the type of `this.systemTheme` is `Theme | undefined`

PR #4980, opened 2022-03-08
Webpack quality of life improvements for dev mode

jakolehm:
- removes `webpack-dev-server` dependency from the application code (it was runtime import in `src/main/index.ts`), webpack-dev-server is now running as a separate process
- proxies asset requests to webpack-dev-server in development mode
- adds filesystem cache for faster boot time
- removes unnecessary `devDependencies` from bundled extensions
- simplifies Makefile


src/main/router.ts
```diff
@@ -140,12 +143,26 @@ export class Router {
...
       }
     }
-
   }
 
   protected addRoutes() {
     // Static assets
-    this.router.add({ method: "get", path: "/{path*}" }, Router.handleStaticFile);
+    if (this.dependencies.httpProxy) {
+      this.router.add({ method: "get", path: "/{path*}" }, (apiReq: LensApiRequest) => {
```
jakolehm, line 151: This is ugly but I didn't want to polish further because it will be refactored anyway in #4859 .

nevalla (dismissed):

nevalla (approved):

extensions/kube-object-event-status/package.json
```diff
@@ -8,18 +8,15 @@
...
   "files": [
     "dist/**/*"
   ],
-  "dependencies": {},
   "devDependencies": {
     "@k8slens/extensions": "file:../../src/extensions/npm/extensions",
-    "ts-loader": "latest",
-    "typescript": "latest",
-    "webpack": "latest"
+    "npm": "^8.5.3"
```
jakolehm, line 20: npx in npm6 is buggy, that's why bundled extensions now have npm@8

Nokel81: Do we need `npx`?

package.json
```diff
@@ -15,10 +15,9 @@
     "dev": "concurrently -i -k \"yarn run dev-run -C\" yarn:dev:*",
     "dev-build": "concurrently yarn:compile:*",
     "debug-build": "concurrently yarn:compile:main yarn:compile:extension-types",
-    "dev-run": "nodemon --watch static/build/main.js --exec \"electron --remote-debugging-port=9223 --inspect .\"",
+    "dev-run": "nodemon --watch ./static/build/main.js --exec \"electron --remote-debugging-port=9223 --inspect .\"",
     "dev:main": "yarn run compile:main --watch --progress",
-    "dev:renderer": "yarn run compile:renderer --watch --progress",
-    "dev:extension-types": "yarn run compile:extension-types --watch --progress",
```
Nokel81, line 21: Why do we not run this anymore?

extensions/kube-object-event-status/package.json
jakolehm: `npx` is the thing that allows us to remove webpack/typescript/ts-loader etc.

Nokel81 (approved):

PR #4990, opened 2022-03-10
Fix all Age, LastSeen, and FirstSeen displays

### LINKED ISSUE #4901: Sorting pod disruption budgets by Age use alphabetic sort
**Describe the bug**
When sorting other entities, sort by Age does the logical thing. For pod disruption budgets, though, the sort appears to be based on the alphabetic order of the convenient value displayed. For example, I see:

100d
108d
10d
121d
2d16h
2d22h
2d2h
73d

**To Reproduce**
1. Go to 'Configuration -> Pod Disruption Budgets'
2. Click on 'Age'
3. Scroll down to observe ages not logically sorted

**Expected behavior**
PDBs should be sorted logically. i.e. for the above list, you should see

2d2h
2d16h
2d22h
10d
73d
100d
108d
121d

**Screenshots**
image

- Lens Version: 5.3.4-latest.20220120.1
- OS: macOS Monterey 12.1

### LINKED ISSUE #4989: Custom Resource age sort not working
**Describe the bug**

**To Reproduce**
1. Go to 'Custom Resources > Definitions'
2. Click on 'Age' column

**Expected behavior**
Expecting a sort

**Screenshots**
image
- Lens Version:  5.4.1
- OS: macOS

Nokel81:
- Sorting based on non-changing values without using Date.now()

- Displaying base on "mobx-utils".now() with some optimization for
  updates of very long durations

- Deprecating KubeObject.getTimeDiffFromNow and KubeObject.getAge
  methods as they are not reactive

- Export `ReactiveDuration` to the extension API.

### DIFF
--- a/src/renderer/components/+custom-resources/crd-resources.tsx
+++ b/src/renderer/components/+custom-resources/crd-resources.tsx
@@ -64,7 +64,7 @@ export class CustomResourceDefinitionResources extends React.Component<CustomRes
         sortingCallbacks={{
           [columnId.name]: customResource => customResource.getName(),
           [columnId.namespace]: customResource => customResource.getNs(),
-          [columnId.age]: customResource => customResource.getCreationTimestamp(),
+          [columnId.age]: customResource => -customResource.getCreationTimestamp(),
           ...Object.fromEntries(extraColumns.map(({ name, jsonPath }) => [
             name,
             customResource => value(customResource, parseJsonPath(jsonPath.slice(1))),
--- a/src/renderer/components/+custom-resources/crd-resources.tsx
+++ b/src/renderer/components/+custom-resources/crd-resources.tsx
@@ -6,17 +6,16 @@
 import "./crd-resources.scss";
 
 import React from "react";
-import jsonPath from "jsonpath";
+import { value } from "jsonpath";
 import { observer } from "mobx-react";
 import type { RouteComponentProps } from "react-router";
 import { KubeObjectListLayout } from "../kube-object-list-layout";
-import type { KubeObject } from "../../../common/k8s-api/kube-object";
 import { computed, makeObservable } from "mobx";
 import { crdStore } from "./crd.store";
-import type { TableSortCallbacks } from "../table";
 import { apiManager } from "../../../common/k8s-api/api-manager";
 import { parseJsonPath } from "../../utils/jsonPath";
 import type { CRDRouteParams } from "../../../common/routes";
+import { KubeObjectAge } from "../kube-object/age";
 
 export interface CustomResourceDefinitionResourcesProps extends RouteComponentProps<CRDRouteParams> {
 }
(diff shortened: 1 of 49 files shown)


jim-docker:
Looks like it's too reactive now:

https://user-images.githubusercontent.com/40840436/158455769-d269227a-0a88-490a-94e5-63402e9a6e6b.mov



Also noticed Last seen time for newly created pod of 52y, but might be an old bug.

Nokel81:
> newly created pod of 52y

Sounds like the unix epoch

> Looks like it's too reactive now:

Do you think it should be slower then? Because those updates are all synchronized (namely they all react against the same observable clock instance)

jim-docker:
> > newly created pod of 52y
> 
> Sounds like the unix epoch
> 
> > Looks like it's too reactive now:
> 
> Do you think it should be slower then? Because those updates are all synchronized (namely they all react against the same observable clock instance)

Updating the item list every second seems too much(?) I guess if it's not making kube api calls every update it's good.

Nokel81:
> Updating the item list every second seems too much(?) I guess if it's not making kube api calls every update it's good.

For something like age I think updating every second (while seconds are displayed) and then only updating every minute is fine.

And yes this is a completely local calculation. It is based on the `metadata.creationTimestamp` field which should never change.

src/renderer/components/+custom-resources/crd-resources.tsx
```diff
@@ -93,36 +84,39 @@ export class CustomResourceDefinitionResources extends React.Component<CustomRes
...
           }),
-          crdInstance.getAge(),
+          <KubeObjectAge key="age" object={crdInstance} />,
         ]}
-        failedToLoadMessage={failedToLoadMessage}
+        failedToLoadMessage={(
+          <>
+            <p>Failed to load {crd.getPluralName()}</p>
+            {!version.served && (
+              <p>Prefered version ({crd.getGroup()}/{version.name}) is not served</p>
```
jim-docker, line 116:
```suggestion
              <p>Preferred version ({crd.getGroup()}/{version.name}) is not served</p>
```

jim-docker (dismissed):
Nice work.
The sort direction for Age seems backwards to me:
Screen Shot 2022-03-15 at 6 12 56 PM
Shouldn't the values go from small to large reading down the list for this indicator?

Nokel81:
> Shouldn't the values go from small to large reading down the list for this indicator?

I guess it depends on what you think the default sorting means. I took it as the chevron points in the direction of sort from smallest to largest.

So an up chevron would mean that the largest age is near the top of the table.

src/renderer/components/duration/reactive-duration.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return 60 * 1000;
+}
+
+export const ReactiveDuration = observer(({ timestamp, compact = true }: ReactiveDurationProps) => {
+  const creationTimestamp = new Date(timestamp).getTime();
+
+  return (
+    <>
+      {formatDuration(now(computeUpdateInterval(creationTimestamp)) - creationTimestamp, compact)}
```
jakolehm, line 41: IMHO we need to tweak `formatDuration` (or use other formatter) now that we update/compute the duration. Updating things like `2m3s` every second will be annoying. My suggestion is to drop seconds completely when over `1m` and then update duration only once per minute.

jakolehm: Just realized that current `formatDuration` mimics kubectl output. Maybe it's ok to keep using that.

jakolehm (dismissed):
Not sure if updating age/duration every second for 10 minutes is a good idea but we can tweak it after we get some real usage.

src/renderer/components/duration/reactive-duration.tsx
Nokel81: I can tweak it if you like...

I could slow down the updates to every 5s if you think it is too fast, and thus distracting.

jim-docker:
> > Shouldn't the values go from small to large reading down the list for this indicator?
> 
> I guess it depends on what you think the default sorting means. I took it as the chevron points in the direction of sort from smallest to largest.
> 
> So an up chevron would mean that the largest age is near the top of the table.

I agree with how you interpret what it should mean but I don't think that is consistent with other fields. For names the chevron pointing up has the "smallest" name at the top.

I assumed how it works is chevron up means ascending, and since the natural direction for a list is downward, then ascending means the values go up as you read the list (down). But I agree it seems more natural for the chevron to point in the direction of the ordering in the list.

Anyway I think what you have here for age is inconsistent with other columns (like restarts on pods)

Nokel81:
> Anyway I think what you have here for age is inconsistent with other columns (like restarts on pods)

That is very convincing, we should be consistent with out UI. Will change.

jim-docker (approved):
Still want it for 5.4.3? I guess we can decide when we go to release 5.4.3. I think it should be 5.5, since it changes behaviour (updates every second), which may be a bit surprising to see in a patch.

Nokel81:
There seems to be more consensus for it being 5.5 so let's do that.

jakolehm (approved):

PR #4992, opened 2022-03-10
Remove webpack.*.ts importing values from src/

Nokel81:
- Remove global modifying __static


tsconfig.json
```diff
@@ -32,7 +32,11 @@
         "types/*"
       ]
     },
-    "plugins": [{ "name": "typescript-plugin-css-modules" }]
+    "plugins": [
+      {
+        "name": "typescript-plugin-css-modules"
+      }
+    ]
```
ixrock, line 39: This probably could be removed (css-modules supported by webpack's `css-loader`)

webpack/extensions.ts
```diff
@@ -6,25 +6,21 @@
...
     target: "electron-renderer",
-    entry,
+    entry: extensionEntry,
     // this is the default mode, so we should make it explicit to silence the warning
     mode: isDevelopment ? "development" : "production",
     output: {
       filename: "extension-api.js",
       // need to be an absolute path
-      path: path.resolve(__dirname, `${outDir}/src/extensions`),
+      path: path.resolve(extensionOutDir, "src", "extensions"),
```
ixrock, line 23: Is there any difference in `path.resolve("path", "sub/path")` and `path.resolve("path", "sub", "path")`?
_(I was thinking it should be covered in both cases no matter which OS/path separator)._

webpack/main.ts
```diff
@@ -54,8 +58,8 @@ configs.push((): webpack.Configuration => {
         cwd: __dirname,
         exclude: /node_modules/,
         failOnError: true,
-      }),
-    ].filter(Boolean),
+      }) as unknown as WebpackPluginInstance,
```
ixrock, line 61: Why need to cast as `WebpackPluginInstance`? _(looks weird within whole config)_

webpack/renderer.ts
```diff
@@ -143,15 +151,19 @@ export function fontsLoaderWebpackRules(): webpack.RuleSetRule[] {
...
-  {
-    styleLoader = vars.isDevelopment ? "style-loader" : MiniCssExtractPlugin.loader,
-  } = {}): webpack.RuleSetRule {
-  const { isDevelopment, sassCommonVars } = vars;
+export function cssModulesWebpackRule({ styleLoader }: CssModulesWebpackRuleOptions = {}): webpack.RuleSetRule {
+  styleLoader ??= (
+    isDevelopment
+      ? "style-loader"
+      : MiniCssExtractPlugin.loader
+  );
```
ixrock: Why so many extra lines? _(harder to read relatively short whole expression than just in a single line)_

webpack/tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "../tsconfig.json"
```
ixrock, line 2: Why to extend base `../tsconfig.json`? Do we use it somewhere else except `webpack`?

webpack/vars.ts
```diff
@@ -0,0 +1,27 @@
...
+export const assetsFolderName = "assets";
+export const rendererDir = path.join(process.cwd(), "src", "renderer");
+export const appName = isDevelopment
+  ? `${packageInfo.productName}Dev`
+  : packageInfo.productName;
+export const htmlTemplate = path.resolve(rendererDir, "template.html");
+export const publicPath = "/build/";
+export const sassCommonVars = path.resolve(rendererDir, "components/vars.scss");
+
+export function ignoreIf<T>(check: boolean, src: T[]): T[] {
```
ixrock: I highly recommend don't bring everywhere such simple utilities as they more adds cognitive load than do some usefulness..

ixrock (dismissed):

webpack/main.ts
Nokel81: Because the typing of `CircularDependencyPlugin` is incorrect and seems to be for an older version of webpack. It still works though...

webpack/tsconfig.json
Nokel81: Because `tsconfig.json` files are only about files at the same level or within folders that are added under `include`. Since I moved the webpack files to their own folder I either had to make another `tsconfig.json` (which is what I did) or add `webpack/` to the top level `tsconfig.json` (which felt wrong because it seems to break the abstraction of which code we care about).

webpack/renderer.ts
Nokel81: I prefer this style for the `?:` operator because it makes it easier to notice both options instead of having to search along a single line for the `:`. This style has both options clearly demarked but I do admit this is a style thing.

webpack/extensions.ts
Nokel81: I don't think so and having checked `path.resolve` will replace `/` with `\` on windows.

tsconfig.json
Nokel81: I believe this is here to support intellisense for css modules.

webpack/vars.ts
Nokel81: Can remove.

webpack/renderer.ts
ixrock: Ok, maybe then just remove wrapping in brackets?

tsconfig.json
ixrock: Nope, it doesn't. For `VScode` you still have to install some plugin.
Recommended extension: https://marketplace.visualstudio.com/items?itemName=phoenisx.cssvar

webpack/main.ts
```diff
@@ -7,16 +7,20 @@ import path from "path";
 import type webpack from "webpack";
 import ForkTsCheckerPlugin from "fork-ts-checker-webpack-plugin";
 import nodeExternals from "webpack-node-externals";
-import * as vars from "./src/common/vars";
-import getTSLoader from "./src/common/getTSLoader";
+import getTypeScriptLoader from "./get-typescript-loader";
 import CircularDependencyPlugin from "circular-dependency-plugin";
-import { iconsAndImagesWebpackRules } from "./webpack.renderer";
+import { iconsAndImagesWebpackRules } from "./renderer";
+import type { WebpackPluginInstance } from "webpack";
```
ixrock, line 13: We have exported types from `webpack` above btw. 
Moreover I would better to remove this at all and mock every package with wrong or missing types in `mocks.d.ts`

webpack/renderer.ts
Nokel81: Sure I can, I prefer having it on separate lines but it isn't that big of a deal.

webpack/main.ts
Nokel81: If https://github.com/import-js/eslint-plugin-import/pull/2419 gets merged then we can switch to using that instead so this isn't really that important.

I also don't know if it is worth it making the types in `mocks.d.ts`. Here it is explicit that something is strange.

tsconfig.json
Nokel81: According to https://www.npmjs.com/package/typescript-plugin-css-modules#visual-studio-code it seems that this is enough.

jim-docker (commented):
Is this just coincidence? Lens won't start (`make clean && make dev`) for this branch but does start on `master`. Gets stuck here:
Screen Shot 2022-04-04 at 7 37 19 PM

Tried thrice with this branch (no start), thrice with master (all start)

webpack/renderer.ts
```diff
@@ -103,14 +109,18 @@ export function webpackLensRenderer({ showVars = true } = {}): webpack.Configura
...
+      }) as unknown as WebpackPluginInstance,
 
       new MiniCssExtractPlugin({
         filename: "[name].css",
       }),
 
-      isDevelopment && new ReactRefreshWebpackPlugin(),
-    ].filter(Boolean),
+      ...(
+        isDevelopment
```
Iku-turso, line 119: Is this ternary the wrong way around? `ReactRefreshWebpackPlugin` should be present if `isDevelopment`?

Iku-turso (changes requested):
https://github.com/lensapp/lens/pull/4992/files#r849236015

webpack/renderer.ts
Nokel81: Good catch

Iku-turso (approved):

ixrock (approved):

jim-docker (approved):

PR #5034, opened 2022-03-17
Fix KubeJsonApi.forCluster throwing on renderer

### LINKED ISSUE #5028: lens_proxy_LensProxy error with newly exported KubeJsonApi in lens extension from Lens 5.4.0 +
**Describe the bug**
While using newly exported KubeJsonApi in lens extension from Lens 5.4.0 +, it is throwing below error:
error:
`TypeError: instance of lens_proxy_LensProxy is not created`

**To Reproduce**

run the below code snippet:

```
export const authPath: string =
  'apis/authorization.k8s.io/v1/selfsubjectaccessreviews';
export const authPayload = {
  data: {
    apiVersion: 'authorization.k8s.io/v1',
    kind: 'SelfSubjectAccessReview',
    spec: {
      resourceAttributes: {
        group: '*',
        user: '*',
        name: '*',
        namespace: '*',
        resource: '*',
        verb: '*',
      },
    },
  },
};

export const canI = async (resource: string, verb: string): Promise<any> => {
  authPayload.data.spec.resourceAttributes.resource = resource;
  authPayload.data.spec.resourceAttributes.verb = verb;

  const clusterId=Renderer.Catalog.catalogEntities.activeEntity.getId()
  const a = Renderer.K8sApi.KubeJsonApi.forCluster(clusterId);
  const res = await a.post(authPath,authPayload);
  
  return res;
};
```

**Expected behavior**
It should return the response without error

**Screenshots**
attaching below in comment.

- Lens Version: Lens: 5.4.2-latest.20220314.1
- OS: OSX
- Installation method (e.g. snap or AppImage in Linux): dmg

**Logs:**
```
Your logs go here...
```

**Kubeconfig:**
```
your kubeconfig here
```

**Additional context**

Nokel81:

fixes #5028


src/common/k8s-api/api-kube.ts
```diff
@@ -0,0 +1,20 @@
...
+  ? new KubeJsonApi({
+    serverAddress: `http://127.0.0.1:${window.location.port}`,
+    apiBase: apiKubePrefix,
+    debug: isDevelopment,
+  }, {
+    headers: {
+      "Host": window.location.host,
+    },
+  })
+  : null;
```
jim-docker: You changed this from `undefined` to `null`. Is this part of the solution, or could there be subtle problems now with this change?

Nokel81: No that is not part of the solution. I can change it back. There shouldn't be except that the error would now be `Cannot read property X of null` instead of `Cannot read property X of undefined`

jim-docker: Could break an extension using `if ( apiKube !== undefined )`? Probably not a big deal (they should use `if ( apiKube )`) but why change it unless `null` is somehow better?

jim-docker (dismissed):

src/common/k8s-api/api-kube.ts
Nokel81: It certainly isn't better, though some people I think argue that `null` should represent a deliberate case of not present but I will revert it.

aleksfront (approved):

jim-docker (approved):

PR #5071, opened 2022-03-22
Emit AppEvent when opening Catalog and changing Catalog Category

panuhorsmalahti:


src/renderer/components/+catalog/catalog.tsx
```diff
@@ -146,6 +154,14 @@ class NonInjectedCatalog extends React.Component<CatalogProps & Dependencies> {
   onTabChange = action((tabId: string | null) => {
     const activeCategory = this.categories.find(category => category.getId() === tabId);
 
+    this.props.emitEvent({
+      name: "catalog",
+      action: "change-category",
+      params: {
+        category: activeCategory ? activeCategory?.metadata?.name : "Browse",
```
Nokel81: nit:
```
category: activeCategory?.getName() ?? "Browse All",
```

Though this might get confused is someone names a category "Browse All", should it really be two fields?

panuhorsmalahti: Why "Browse All" instead of "Browse"? The label/name is "Browse" in the UI.

It seems unlikely that anyone would name a custom category as "Browse", the same label as an existing menu item.

panuhorsmalahti: getName doesn't exist for category (it does for entity). Should I add similar method to the CatalogCategory class?

Nokel81: Fair enough, yeah I would add such a method.

panuhorsmalahti: Added the method now

Nokel81 (approved):

DmitriyNoa (approved):

PR #5082, opened 2022-03-23
Get DI cleaner

panuhorsmalahti:
We only need the di inside the callback, so it's cleaner to get it there. This function also has side effect (which might be removed later, though).

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -58,11 +58,11 @@ import userStoreInjectable from "../common/user-store/user-store.injectable";
 import trayMenuItemsInjectable from "./tray/tray-menu-items.injectable";
 import { broadcastNativeThemeOnUpdate } from "./native-theme";
 
-const di = getDi();
-
 app.setName(appName);
 
 app.on("ready", async () => {
+  const di = getDi();
+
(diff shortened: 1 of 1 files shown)


src/main/index.ts
```diff
@@ -58,11 +58,11 @@ import userStoreInjectable from "../common/user-store/user-store.injectable";
 import trayMenuItemsInjectable from "./tray/tray-menu-items.injectable";
 import { broadcastNativeThemeOnUpdate } from "./native-theme";
 
-const di = getDi();
-
 app.setName(appName);
 
 app.on("ready", async () => {
+  const di = getDi();
```
Iku-turso, line 64: Hmm. Running `getDi()` later is risky for still some time, as there yet exists some vile import-time side-effects and global shared states which require `setLegacyGlobalDiForExtensionApi()` (in `getDi()`) to have run as soon as possible as a temporal dependency.

I do see how normally this change makes sense, but this change cannot be done right now out of unfair fear :(

Abandon PR?

panuhorsmalahti: Okay, thanks for the explanation

PR #5083, opened 2022-03-23
Bump package version to 5.5.0-alpha.0

Nokel81:


jim-docker (approved):
will this be automated in the release process?

Nokel81:
We might be able to. I think it should be part of the release doc at least.

aleksfront (approved):

PR #5084, opened 2022-03-23
Technical requirements for behavioural unit tests

Iku-turso:
Implement a lot of technical requirements for behavioural unit tests

With this:
1. Routes and parameters related are strongly typed
2. Routes exist in all environments making usage transparent
3. Routes and related phenomena comply to Open Closed Principle (eg. what it means to be a new route does not require modification of existing code, but new code)
4. Behavioural unit tests are possible (see #5085)

Note: the crux of this was to make routing env-agnostic, and not based on URLs as magic strings, but instead something type-enforced.

Note: extension-based routes comply to same exact interface by "late-registering" their routes when installed. Routes are injectables which implement a shared injection token.

Note: another chunk of global shared state is no more.

Note: a lot of stuff has become reactive as part if this.

Note: a lot of explicit side effects have been cornered to injectables with `causesSideEffects: true`.

Note: old unit tests which want to test what is `causesSideEffects: true` can now do so with `di.permitSideEffects(someInjectable);`.


package.json
```diff
@@ -200,8 +200,9 @@
     "@hapi/call": "^8.0.1",
     "@hapi/subtext": "^7.0.3",
     "@kubernetes/client-node": "^0.16.1",
-    "@ogre-tools/injectable": "5.1.2",
-    "@ogre-tools/injectable-react": "5.1.2",
+    "@ogre-tools/injectable": "5.2.0",
+    "@ogre-tools/injectable-react": "5.2.0",
+    "@ogre-tools/fp": "5.2.0",
```
Nokel81, line 207: I would like to raise notice against using this. I find that even with types `pipeline` is very hard to read. I would like to motion that chaining is better ergonomically.

integration/__tests__/cluster-pages.tests.ts
```diff
@@ -323,148 +39,390 @@ utils.describeIf(minikubeReady(TEST_NAMESPACE))("Minikube based tests", () => {
     await frame.waitForSelector(`.Menu >> text="Remove"`);
   });
 
-  it("should navigate around common cluster pages", async () => {
-    for (const test of commonPageTests) {
-      if (isTopPageTest(test)) {
-        const { href, expectedText, expectedSelector } = test.page;
-        const menuButton = await frame.waitForSelector(`a[href^="${href}"]`);
+  it(
```
Nokel81, line 42: Can we also standardize our formatting rules ASAP. All these changes which your prettier config is introducing is making diffs much larger and is not consistent with the majority of the code base.

package.json
Iku-turso: Opinion noted! :) 

Given a bit of time and familiarity, the pattern of `pipeline` will once again present itself as **very easy** to read, instead of anything else. This will become more apparent in more complex algorithms.

As with most design patterns, truly realized ergonomy comes from repetition ;)

Noteworthy: `pipeline` has multiple advantages over the proposed **Collection chaining** -pattern:

1. Better support for functional composition, and other functional patterns, such as `flow` or ternary functions.
2. Better refactorability, especially for nested or recursed scenarios.
3. Power to control flow and aspect: eg. if required, `pipeline` can abstract promises and even generators.
4. Full control of vocabulary (and not just `map`, `reduce`, `filter`, etc.) leads to more expressive code.
5. It can be used for other than collections (yet, so can eg. `lodash.chain`, but then again [this](https://medium.com/bootstart/why-using-chain-is-a-mistake-9bc1f80d51ba)).
6. Etc.

We used **Collection chaining** for years before growing out of it. As with everything, we do not use something like `pipeline` to flex, but instead for the necessity of the  _Greater good_ ;)

integration/__tests__/cluster-pages.tests.ts
Iku-turso: You have my axe!

Diffs and inconsistency sure are something too, but not being able to properly use one-truth formatting in development is a real downer.

src/common/hotbar-store.ts
```diff
@@ -3,326 +3,10 @@
...
-    const lastNull = hotbar.items.lastIndexOf(null);
-
-    if (lastNull >= 0) {
-      hotbar.items.splice(lastNull, 1);
-    } else {
-      hotbar.items.length = defaultHotbarCells;
-    }
-  }
-}
+export const HotbarStore = asLegacyGlobalSingletonForExtensionApi(hotbarStoreInjectable);
```
Nokel81: nit: mark as deprecated

src/common/logger.ts
```diff
@@ -10,6 +10,14 @@ import { consoleFormat } from "winston-console-format";
 import { isDebugging, isTestEnv } from "./vars";
 import BrowserConsole from "winston-transport-browserconsole";
 
+export interface Logger {
+  info: (...args: any) => void;
+  error: (...args: any) => void;
+  debug: (...args: any) => void;
+  log: (...args: any) => void;
+  warn: (...args: any) => void;
```
Nokel81: 1. I don't think we should have `log` here
2. I think it would be better to enforce `(message: string, ...args: any[]) => void`

src/common/user-store/user-store.injectable.ts
```diff
@@ -8,6 +8,7 @@ import { UserStore } from "./user-store";
 const userStoreInjectable = getInjectable({
   id: "user-store",
   instantiate: () => UserStore.createInstance(),
```
Nokel81: Shouldn't this have resetInstance too?

src/common/utils/buildUrl.ts
```diff
@@ -11,25 +11,21 @@ export interface URLParams<P extends object = {}, Q extends object = {}> {
   fragment?: string;
 }
 
-export function buildURL<P extends object = {}, Q extends object = {}>(path: string | any) {
+export function buildURL<P extends object = {}, Q extends object = {}>(path: string | any, { params, query, fragment }: URLParams<P, Q> = {}) {
```
Nokel81: can you also remove the `| any` from `path`?

src/common/utils/is-allowed-resource.injectable.ts
```diff
@@ -2,21 +2,25 @@
...
-  instantiate: (di) => {
+  instantiate: (di, resourceName: KubeResource) => {
     const allowedResources = di.inject(allowedResourcesInjectable);
 
-    return (resource: KubeResource) => allowedResources.get().has(resource);
+    return computed(() => allowedResources.get().has(resourceName));
   },
+
+  lifecycle: lifecycleEnum.keyedSingleton({
+    getInstanceKey: (_, resource: KubeResource) => resource,
```
Nokel81: nit: is this type annotation required?

src/extensions/as-legacy-globals-for-extension-api/legacy-global-di-for-extension-api.ts
```diff
@@ -4,10 +4,29 @@
...
+) => {
+  legacyGlobalDis.set(environment, di);
+};
+
+export const getLegacyGlobalDiForExtensionApi = () => {
+  const globalDis = [...legacyGlobalDis.values()];
+
+  if (globalDis.length > 1) {
+    throw new Error("Tried to get DI container using legacy globals where there is multiple containers available.");
+  }
```
Nokel81, line 26: When would this even occur? Testing?

jansav (dismissed):

package.json
panuhorsmalahti: I think the pipeline syntax is good, but I'm a bit wary of adding this particular dependency, which appears to be quite a thin wrapper over lodash. This dependency has really small usage. Perhaps there's another alternative that is more commonly used in the ecosystem.

panuhorsmalahti: I think the pipeline syntax is good, but I'm a bit wary of adding this particular dependency, which appears to be quite a thin wrapper over lodash. This dependency has really small usage. Perhaps there's another alternative that is more commonly used in the ecosystem.

integration/__tests__/cluster-pages.tests.ts
panuhorsmalahti: I agree with @Nokel81. Until prettier (or another tool) is integrated I think developers should disable automatic formatting changes like these.

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -317,6 +316,12 @@ export class ExtensionLoader {
               extension,
             );
 
+            const installationCount = (this.dependencies.extensionInstallationCounter.get(instance.sanitizedExtensionId) | 0) + 1;
+
+            this.dependencies.extensionInstallationCounter.set(instance.sanitizedExtensionId, installationCount);
+
+            await Promise.all(this.dependencies.extensionRegistrators.map(x => x(instance, installationCount)));
```
panuhorsmalahti: `registrator` would be more readable than `x`

src/common/front-end-routing/route-injection-token.ts
```diff
@@ -0,0 +1,20 @@
...
+import type { LensRendererExtension } from "../../extensions/lens-renderer-extension";
+
+export const routeInjectionToken = getInjectionToken<Route<unknown>>({
+  id: "route-injection-token",
+});
+
+export interface Route<TParameter = void> {
+  path: string;
+  clusterFrame: boolean;
+  isEnabled: IComputedValue<boolean>;
```
panuhorsmalahti, line 16: How about we type this as `boolean | IComputedValue<boolean>` to avoid the `computed(() => true)`?

src/common/rbac.ts
```diff
@@ -10,6 +10,16 @@ export type KubeResource =
...
 
+export enum KubeResourceEnum {
+  pods = "pods",
+  deployments = "deployments",
+  cronJobs = "cronjobs",
+  jobs = "jobs",
+  replicasets = "replicasets",
+  daemonsets = "daemonsets",
+  statefulsets = "statefulsets",
+}
```
Nokel81: Do we need this?

integration/__tests__/cluster-pages.tests.ts
jansav: Yep, we've removed these type of changes previously. Couple slipped in, so we'll revert those.

package.json
Iku-turso: @panuhorsmalahti hmm, interesting point. Here's some thoughts: 
1. We control this library ourselves, and for that, it is flexible and without magic. 

2. We created the library only recently, and therefore there is yet little adoption in the ecosystem.

3. We could have implemented this function in Lens as well, but it is pretty general to be local. Eg., also `injectable` uses it.

4. To look for an alternative full foreign library for this "thin wrapper" is a bit of an overkill, as the proposed pattern is universal, and does not require certain library to work properly with. Instead, any function with arity of one can participate in composition. Therefore a full alternate library would add extra weight with all the other bells and whistles. I mean, we already have `lodash`.

src/common/front-end-routing/route-injection-token.ts
Iku-turso: Hmm. A conditional type here would come with non-insignificant cost in complexity, as it introduces an alternate way of using `isEnabled`. There is no actual value in minimal implementation of just `true` over `computed(() => true)`. Instead, having to use value of `isEnabled` somewhere without certainty of type is of negative value.

Nokel81: The other option could be for this to be optional and if not provided it defaults to `computed(() => true)`

panuhorsmalahti: I think it it wouldn't come with non-insignificant cost in complexity. There is actual value of just providing `true`, e.g. simplicity and readability. Even in this PR most uses seem to use `computed(() => true)`. There might be cost when "using isEnabled".. but what exactly is that cost here?

panuhorsmalahti: So the cost is that caller needs to call `.get()` for standalone computed value, and since there are multiple uses of this `isEnabled` we can go with enforcing the computed type.

src/common/utils/is-allowed-resource.injectable.ts
jansav: Improved variable naming.

src/common/utils/buildUrl.ts
jansav: Done.

src/common/user-store/user-store.injectable.ts
jansav: Added to this and all other similar places.

src/common/logger.ts
jansav: Done.

src/extensions/as-legacy-globals-for-extension-api/legacy-global-di-for-extension-api.ts
jansav: Yes in unit tests, this is there to prevent hard to find "why is this test green or red" -moments. Can't occur in production since both  there is always only one DI, but in unit tests this can happen very easily at the moment. 

Of course, in long run we are removing these "legacy-global" -functions for good, so it will get removed as well.

src/common/rbac.ts
jansav: At the moment yes. Union type above doesn't do the same. 

The real solution would be to make `KubeResource`s comply to Open Closed Principle. But that's out of scope here, but when done, we can safely remove whole file and all contents from it.

integration/__tests__/cluster-pages.tests.ts
jansav: Done

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -317,6 +316,16 @@ export class ExtensionLoader {
...
 
+            const installationCount = (this.dependencies.extensionInstallationCounter.get(instance.sanitizedExtensionId) | 0) + 1;
+
+            this.dependencies.extensionInstallationCounter.set(instance.sanitizedExtensionId, installationCount);
+
+            await Promise.all(
+              this.dependencies.extensionRegistrators.map((register) =>
+                register(instance, installationCount),
+              ),
+            );
```
Nokel81: The fact that this is `async` is highly suspicious.

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -317,6 +316,16 @@ export class ExtensionLoader {
               extension,
             );
 
+            const installationCount = (this.dependencies.extensionInstallationCounter.get(instance.sanitizedExtensionId) | 0) + 1;
+
+            this.dependencies.extensionInstallationCounter.set(instance.sanitizedExtensionId, installationCount);
```
Nokel81, line 321: I find this state very suspicious and would like to see it removed. I think that separating the extension sourced items from internal sources like with other DI solutions would be a good solution.

Having discussed this with Mikko, I am sort of fine with this being left in this PR but with being marked "suspicious".

src/common/catalog-entities/general-catalog-entities/implementations/catalog-catalog-entity.injectable.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { generalCatalogEntityInjectionToken } from "../general-catalog-entity-injection-token";
+import { GeneralEntity } from "../../index";
+import { buildURL } from "../../../utils/buildUrl";
+import catalogRouteInjectable from "../../../front-end-routing/routes/catalog/catalog-route.injectable";
+
+const catalogCatalogEntityInjectable = getInjectable({
+  id: "general-catalog-entity-for-catalog",
```
jim-docker, line 12: Are these the right names?

src/common/front-end-routing/routes/catalog/catalog-route.injectable.ts
```diff
@@ -0,0 +1,26 @@
...
+export interface CatalogPathParameters {
+  group?: string;
+  kind?: string;
+}
+
+const catalogRouteInjectable = getInjectable({
+  id: "catalog-route",
+
+  instantiate: (): Route<CatalogPathParameters> => ({
+    path: "/catalog/:group?/:kind?",
```
jim-docker, line 18: couldn't this be auto-generated given the typing now (`CatalogPathParameters`)? Seems open to programmer error.

src/common/front-end-routing/routes/cluster/overview/cluster-overview-route.injectable.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import isAllowedResourceInjectable from "../../../../utils/is-allowed-resource.injectable";
+import { routeInjectionToken } from "../../../route-injection-token";
+
+const clusterOverviewRouteInjectable = getInjectable({
+  id: "cluster-overview-route",
+
+  instantiate: (di) => {
+    const isAllowedResource = di.inject(isAllowedResourceInjectable, "nodes");
```
jim-docker, line 13: should `isEnabled` just be `() => true`? I guess it's all info per the nodes? Was a criterion previously?

src/common/front-end-routing/routes/preferences/navigate-to-preferences.injectable.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import navigateToAppPreferencesInjectable from "./app/navigate-to-app-preferences.injectable";
+
+const navigateToPreferencesInjectable = getInjectable({
```
jim-docker, line 8: why do we need this? Can't `navigateToAppPreferencesInjectable` be used directly?

jim-docker (commented):
still going...

src/common/front-end-routing/routes/cluster/overview/cluster-overview-route.injectable.ts
jansav: It's exactly same check that was before in sidebar item for cluster-overview. Computed because it's based of an observable which can change.

src/common/front-end-routing/routes/preferences/navigate-to-preferences.injectable.ts
jansav: If you want to navigate to first page of preferences you should use `navigateToPreferences`. If you want to navigate to specific tab in preferences, you should use e.g. `navigateToAppPreferences`. 

Motivation is to allow change of mind in future without need for multiple modifications.

src/common/front-end-routing/routes/catalog/catalog-route.injectable.ts
jansav: Maybe. I'd call this topic for of the future.

src/common/catalog-entities/general-catalog-entities/implementations/catalog-catalog-entity.injectable.ts
jansav: It's there to highlight the issue with naming. 

Interface is `GeneralCatalogEntity` and implementations are `PreferencesGeneralCatalogEntity`, `WelcomeGeneralCatalogEntity` and `CatalogGeneralCatalogEntity`.

1. "General" is redundant duplication -> `PreferencesCatalogEntity`, `WelcomeCatalogEntity` and `CatalogCatalogEntity`
2. `CatalogCatalogEntity` feels wrong BUT `CatalogEntity` would not be sufficient either. 

We are joggling between:
1. `PreferencesGeneralCatalogEntity`, `WelcomeGeneralCatalogEntity`, `CatalogGeneralCatalogEntity`
2. `PreferencesCatalogEntity`, `WelcomeCatalogEntity`, `CatalogCatalogEntity`

src/common/rbac.ts
Nokel81: What do you mean by "do the same"? And it is impossible for a type to be OCP in TS so that point is moot.

Nokel81 (commented):
Will do more reviewing today.

src/extensions/extension-loader/extension-loader.ts
jansav: Good catch. No need to be async anymore.

jansav:
> What do you mean by "do the same"? And it is impossible for a type to be OCP in TS so that point is moot.

These enum/union-types are not needed if the implementation complies to OCP. That's the point.

Union types require you to write "magical strings" and which should be considered as bad practise. E.g. those are harder to find since you cannot rely on IDE's "find usages" -tool.

src/extensions/extension-loader/extension-loader.ts
jansav: Resolved

Nokel81:
While I admit that "magical strings" are generally bad practise. The fact that TS (with intellisense) auto completes them and will emit a compile time error if it is incorrect removes that instance.

Finally, you are incorrect that "find references" does not work with union types:
Screen Shot 2022-03-30 at 9 07 36 AM

jansav:
> While I admit that "magical strings" are generally bad practise. The fact that TS (with intellisense) auto completes them and will emit a compile time error if it is incorrect removes that instance.
> 
> Finally, you are incorrect that "find references" does not work with union types: Screen Shot 2022-03-30 at 9 07 36 AM

Hm, weird. Just tried similar in webstorm and it didn’t work.

Nokel81:
> Hm, weird. Just tried similar in webstorm and it didn’t work.

Does webstorm use the official typescript lang server?

src/common/front-end-routing/routes/catalog/catalog-route.injectable.ts
jim-docker: probably not so easy as I realized the params can be arbitrarily placed in the path.

src/common/rbac.ts
Iku-turso: Here we have a list of magic strings for entities which also relate to functionalities which are spread-and-lost around the codebase, making it not OCP. If what it means to be an entity of "KubeResource" was specified in a centralized place, it would be OCP.

I propose not investing a lot of effort in this before OCP happens.

Nokel81: Then just make it `string`? I don't see the point of adding this enum.

src/common/front-end-routing/routes/catalog/catalog-route.injectable.ts
Iku-turso: For this, we could take a note from [these guys](https://github.com/kruschid/typesafe-routes). (Props to @Nokel81 for the link). But later? :)

Nokel81: I think later

src/common/front-end-routing/routes/preferences/navigate-to-preferences.injectable.ts
jim-docker: You mean so one day `navigateToAppPreferences` may take optional param(s) to select the tab?

src/common/app-paths/directory-for-binaries/directory-for-binaries.injectable.ts
```diff
@@ -3,14 +3,18 @@
...
 import { getInjectable } from "@ogre-tools/injectable";
-import path from "path";
 import directoryForUserDataInjectable from "../directory-for-user-data/directory-for-user-data.injectable";
+import getAbsolutePathInjectable from "../../path/get-absolute-path.injectable";
 
 const directoryForBinariesInjectable = getInjectable({
   id: "directory-for-binaries",
 
-  instantiate: (di) =>
-    path.join(di.inject(directoryForUserDataInjectable), "binaries"),
```
Nokel81, line 13: Given such simple injectables I think that having them so deeply nested hurts readability on the use of them side.

Namely, the import paths get needlessly long.

I would like to propose that these get move to `src/common/app-paths/directory-for-binaries.injectable.ts` instead, especially since I doubt that they will ever have suplimental injectables. And even then those could be in a folder while this one is left in the parent folder.

src/common/path/get-absolute-path.injectable.ts
```diff
@@ -0,0 +1,20 @@
...
+export type GetAbsolutePath = (...args: string[]) => string;
+
+const getAbsolutePathInjectable = getInjectable({
+  id: "get-absolute-path",
+
+  instantiate: (): GetAbsolutePath => path.resolve,
+
+  // This causes side effect e.g. Windows creates different kinds of
+  // absolute paths than linux
+  causesSideEffects: true,
```
Nokel81, line 17: I disagree that this sort of thing should be overridden in tests, instead the tests should be written in an agnositc fasion.

src/common/catalog-entities/general-catalog-entities/implementations/catalog-catalog-entity.injectable.ts
Nokel81: I think this name is fine too

src/common/catalog-entities/general-catalog-entities/general-catalog-entities.injectable.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { observable } from "mobx";
+import { generalCatalogEntityInjectionToken } from "./general-catalog-entity-injection-token";
+
+const generalCatalogEntitiesInjectable = getInjectable({
+  id: "general-catalog-entities",
+
+  instantiate: (di) =>
+    observable(di.injectMany(generalCatalogEntityInjectionToken)),
```
Nokel81: This does not need to be observable. Only computed.

src/common/k8s-api/endpoints/crd.api.ts
```diff
@@ -99,7 +101,11 @@ export class CustomResourceDefinition extends KubeObject {
   }
 
   getResourceUrl() {
-    return crdResourcesURL({
+    const di = getLegacyGlobalDiForExtensionApi();
+
+    const customResourcesRoute = di.inject(customResourcesRouteInjectable);
```
Nokel81, line 106: Why not inject the `navigateToCustomResourcesInjectable` instead?

src/common/hotbar-store.injectable.ts
```diff
@@ -2,12 +2,377 @@
...
+export interface HotbarStoreModel {
+  hotbars: Hotbar[];
+  activeHotbarId: string;
+}
+
+interface Dependencies {
+  catalogCatalogEntity: GeneralEntity;
+}
+
+export class HotbarStore extends BaseStore<HotbarStoreModel> {
```
Nokel81: I know that having the implementation beside the injectable is good, but for such a large implementation as this I don't think it is. I think that this should stay in `hotbar-store.ts`

src/main/catalog/catalog-entity-registry.injectable.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { catalogEntityRegistry } from "./catalog-entity-registry";
+
+const catalogEntityRegistryInjectable = getInjectable({
+  id: "catalog-entity-registry",
+  instantiate: () => catalogEntityRegistry,
```
Nokel81, line 10: Could we actually do it the other way around so that this isn't `causesSideEffects`.

Namely, actually create a new instance here and then make `catalogEntityRegistry` a `legacyGlobal...`?

src/main/initializers/init-ipc-main-handlers/init-ipc-main-handlers.ts
```diff
@@ -14,25 +14,23 @@ import { catalogEntityRegistry } from "../../catalog";
...
 import { showOpenDialog } from "../../ipc/dialog";
 import { windowActionHandleChannel, windowLocationChangedChannel, windowOpenAppMenuAsContextMenuChannel } from "../../../common/ipc/window";
 import { getNativeColorTheme } from "../../native-theme";
 import { getNativeThemeChannel } from "../../../common/ipc/native-theme";
+import type { MenuItemsOpts } from "../../menu/get-app-menu-items.injectable";
+import type { GetAbsolutePath } from "../../../common/path/get-absolute-path.injectable";
 
 interface Dependencies {
-  electronMenuItems: IComputedValue<MenuRegistration[]>;
+  getAppMenuItems: () => MenuItemsOpts[];
```
Nokel81: Shouldn't this be `IComputedValue<MenuItemsOpts[]>`?

src/main/menu/get-app-menu-items.injectable.ts
```diff
@@ -0,0 +1,333 @@
...
+import navigateToCatalogInjectable from "../../common/front-end-routing/routes/catalog/navigate-to-catalog.injectable";
+import navigateToWelcomeInjectable from "../../common/front-end-routing/routes/welcome/navigate-to-welcome.injectable";
+import navigateToAddClusterInjectable from "../../common/front-end-routing/routes/add-cluster/navigate-to-add-cluster.injectable";
+import isMacInjectable from "../../common/vars/is-mac.injectable";
+
+function ignoreIf(check: boolean, menuItems: MenuItemConstructorOptions[]) {
+  return check ? [] : menuItems;
+}
+
+export interface MenuItemsOpts extends MenuItemConstructorOptions {
```
Nokel81: nit: the name probably should be `MenuItemOpts`

src/main/window-manager.ts
```diff
@@ -187,7 +187,7 @@ export class WindowManager extends Singleton {
     return this.mainWindow;
   }
 
-  private sendToView({ channel, frameInfo, data = [] }: SendToViewArgs) {
+  sendToView({ channel, frameInfo, data = [] }: SendToViewArgs) {
```
Nokel81, line 190: why is this public now?

src/renderer/app-paths/get-value-from-registered-channel/register-ipc-channel-listener.injectable.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const registerIpcChannelListenerInjectable = getInjectable({
+  id: "register-ipc-channel-listener",
+
+  instantiate: (di) => {
+    const ipc = di.inject(ipcRendererInjectable);
+
+    return ({ channel, handle }: IpcChannelListener) => {
+      ipc.on(channel.name, (_, data) => {
+        handle(data);
```
Nokel81, line 19: I would call this `listener` instead because `handle` (and `handler`) implies waiting (an `ipc.invoke(...)`) instead of just an event handler (an `ipc.send(...)`)

src/renderer/routes/current-route.injectable.ts
```diff
@@ -0,0 +1,27 @@
...
+const currentRouteInjectable = getInjectable({
+  id: "current-route",
+
+  instantiate: (di) => {
+    const matchingRoute = di.inject(matchingRouteInjectable);
+
+    return computed(() => {
+      const match = matchingRoute.get();
+
+      if (match && match.route.isEnabled.get()) {
```
Nokel81, line 18: nit: `match?.route.isEnabled.get()`

src/renderer/routes/current-path-parameters.injectable.ts
```diff
@@ -0,0 +1,23 @@
...
+import matchingRouteInjectable from "./matching-route.injectable";
+
+const currentPathParametersInjectable = getInjectable({
+  id: "current-path-parameters",
+
+  instantiate: (di) => {
+    const matchingRoute = di.inject(matchingRouteInjectable);
+
+    return computed((): Record<string, string> => {
+      const match = matchingRoute.get();
```
Nokel81, line 16: nit, I don't think we should have this injectable at all. Because either you are in the case where you already know you have a defined matching route or you know that you are in the case where you don't in which case the `pathParameters` don't matter.

src/renderer/routes/routes.injectable.ts
```diff
@@ -0,0 +1,36 @@
...
+    const currentlyInClusterFrame = di.inject(
+      currentlyInClusterFrameInjectable,
+    );
+
+    return computed(() =>
+      allRoutes
+        .get()
+        .filter(matches({ clusterFrame: currentlyInClusterFrame }))
+        .filter((route) => {
+          if (route.isEnabled) {
```
Nokel81: I thought that this was required, so why the defensive code here?

src/renderer/routes/route-path-parameters.injectable.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const routePathParametersInjectable = getInjectable({
+  id: "route-path-parameters",
+
+  instantiate: (di, route: Route<unknown>) => {
+    const currentPath = di.inject(currentPathInjectable);
+
+    // TODO: Reuse typing from route for accuracy
+    return computed((): Record<string, string> => {
+      const match = matchPath(currentPath.get(), {
```
Nokel81, line 19: I don't like this design, too defensive in nature. Instead I think it would be better to have these types cascade once matched.

src/renderer/bootstrap.tsx
```diff
@@ -58,6 +63,9 @@ async function attachChromeDebugger() {
 export async function bootstrap(di: DiContainer) {
   await di.runSetups();
 
+  // TODO: Consolidate import time side-effect to setup time
+  bindEvents();
```
Nokel81, line 69: Wouldn't this be the PR to do this in?

yarn.lock
```diff
@@ -956,28 +956,28 @@
...
+  resolved "https://registry.yarnpkg.com/@ogre-tools/injectable-react/-/injectable-react-5.2.0.tgz#ec55cd88d3b7e778952d77e6fcb4771f315430d2"
+  integrity sha512-V1aMIKGIXrhZe6OwhcZ5xm4PLfE9F21KXOXhhz7dXcECWKh0+tHUJaG3MaMPzJqloRSE9SNbMz5nCOv172xRvg==
   dependencies:
-    "@ogre-tools/fp" "^5.1.2"
-    "@ogre-tools/injectable" "^5.1.2"
+    "@ogre-tools/fp" "^5.2.0"
+    "@ogre-tools/injectable" "^5.2.0"
     lodash "^4.17.21"
 
-"@ogre-tools/injectable@5.1.2", "@ogre-tools/injectable@^5.1.2":
```
Nokel81, line 975: Should this package's typings of `lifecycleEnum.singleton` be fixed and then brought in with this PR?

src/renderer/bootstrap.tsx
jansav: Had to remove import time side-effect to make behavioural unit tests possible.

yarn.lock
jansav: What do you mean? How is it broken?

Nokel81: Just try using it, the typings are broken. The fix is really simple too.

src/renderer/routes/current-path-parameters.injectable.ts
jansav: There is cases where you don’t want to know which route you are on, but instead you want to do something with path parameters. e.g. all cluster routes after iframe has been removed.

src/common/k8s-api/endpoints/crd.api.ts
jansav: Too big change for time being.

src/common/catalog-entities/general-catalog-entities/general-catalog-entities.injectable.ts
jansav: It was observable before. It looks like something that shouldn’t be reactive at all. However we kept it observable simply to make sure that we don’t break anything. Should be changed in different PR to figure out what is really needed and what is not.

src/common/path/get-absolute-path.injectable.ts
jansav: Unit tests shouldn’t have side effects. All OS specific business logic should be unit tested without relying side-effects (here OS). 

Integration tests are something different, and those shouldn’t override this type of stuff. 

(keep in mind that current integration tests are not really integration tests, those are e2e smoke tests.)

1. It is enough to run unit tests with one operating system

2. Integration tests should be ran with all known operating systems

3. E2E tests should be ran with all known operating systems.

src/common/app-paths/directory-for-binaries/directory-for-binaries.injectable.ts
jansav: Maybe. Could even be that we only have one injectable for all app paths. Let’s figure it out in separate PR?

src/main/catalog/catalog-entity-registry.injectable.ts
jansav: Sure, we could but do we want to do it here? We have tried to avoid expanding scope of this PR as much as possible. Changing that will introduce more files that change.

src/main/window-manager.ts
jansav: Good catch, not needed anymore.

yarn.lock
jansav: Last I tried it didn’t allow it at all. And that’s fine because it’s default. Need to try it tomorrow.

src/renderer/routes/current-path-parameters.injectable.ts
Nokel81: I mean the path parameters should be passed into the component by whatever mounts it. That way it is still agnostic and unit testable but we don't have this defensive code.

src/common/k8s-api/endpoints/crd.api.ts
Nokel81: okay, bit confused but okay.

src/main/catalog/catalog-entity-registry.injectable.ts
Nokel81: alright

src/renderer/routes/current-path-parameters.injectable.ts
jansav: If we kept doing that only the components would have access to path parameters. However, with this you can access path parameters anywhere. Which is requirement when we get rid of iframes.

src/common/app-paths/directory-for-binaries/directory-for-binaries.injectable.ts
Nokel81: I sort of like having them separate especially since we have already started to compose upon them. And this sort of change should be simple to do we auto-refactorings.

src/common/catalog-entities/general-catalog-entities/general-catalog-entities.injectable.ts
Nokel81: okay sure, I know for a fact that it doesn't need to be. But it can be a follow up PR if you like.

src/common/path/get-absolute-path.injectable.ts
Nokel81: I don't see how having this overriden doesn't diminish our trust in the unit test.

src/renderer/routes/current-path-parameters.injectable.ts
Nokel81: I disagree that it is a requirement to get rid of iframes. The top level component could instead do any of the required passing of data.

src/common/path/get-absolute-path.injectable.ts
Iku-turso: A side-effect has no place in a unit test. As background, this particular side-effect already caused damage by only revealing its nasty flakiness once down in CI.

If trust in modules with side effects is diminished, one can write integration tests for them. This case however does not warrant such a thing, as there is no reasonable doubt that `path.resolve` would betray us.

src/renderer/routes/current-path-parameters.injectable.ts
Iku-turso: > If we kept doing that only the components would have access to path parameters. However, with this you can access path parameters anywhere.

I think this is the most important bit in what @jansav said.

src/renderer/routes/routes.injectable.ts
Iku-turso: I don't understand.

src/renderer/routes/route-path-parameters.injectable.ts
Iku-turso: I don't quite understand what you mean here, but after this PR, ideas like this will sure start living in increasingly more fertile grounds, as they will be protected by the behavioural unit tests, and thus be free to be refactored and improved on over and over again.

yarn.lock
Iku-turso: This is fixed for `@ogre-tools/injectable@6.0.1`, but not brought in this PR because of the major version bump. There is already another PR for this update in the works.

src/renderer/routes/routes.injectable.ts
Nokel81: The interface `FrontEndRoute` has a field `isEnabled` of type `IComputedValue<boolean>` which is a required field (not an optional one) so this sort of check shouldn't be needed.

src/common/path/get-absolute-path.injectable.ts
Nokel81: If there is no reasonable doubt that it would betray us then isn't unit test just wrong?

src/renderer/routes/routes.injectable.ts
jansav: You are right. Changing it tomorrow. Background: it wasn't required before, we changed our mind on that.

src/common/front-end-routing/routes/preferences/navigate-to-preferences.injectable.ts
jansav: No. `navigateToAppPreferences` doesn't need any parameters. However we might change our mind that `navigateToPreferences` doesn't go to app preferences. One day it might go to editor preferences and for that change we would need to change lot of file instead of one.

Implementation for `navigateToPreferences` could even be "go to first tab in preferences" and it shouldn't need to know which tab is first. In that case `navigateToPreferences` doesn't have to change at all if someday we change order of tabs in preferences.

src/common/catalog-entities/general-catalog-entities/general-catalog-entities.injectable.ts
jansav: I'll do the change tomorrow since you are absolutely sure. :)

Iku-turso: Nah just fixed it.

src/common/path/get-absolute-path.injectable.ts
Iku-turso: If there is no reasonable doubt of operation for a minimal module with side-effects, we are free and right to stub it in unit testing.

src/renderer/routes/routes.injectable.ts
Iku-turso: Fixed it already.

src/common/front-end-routing/verify-that-all-routes-have-route-component.test.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+import { getDiForUnitTesting } from "../../renderer/getDiForUnitTesting";
+import { routeSpecificComponentInjectionToken } from "../../renderer/routes/route-specific-component-injection-token";
+import { routeInjectionToken } from "./route-injection-token";
+import { filter, map, matches } from "lodash/fp";
+import clusterStoreInjectable from "../cluster-store/cluster-store.injectable";
+import type { ClusterStore } from "../cluster-store/cluster-store";
+import { pipeline } from "@ogre-tools/fp";
+
+describe("verify-that-all-routes-have-component", () => {
```
jim-docker, line 13: A better test makes sure each route has the right component? I guess we have an integration test that sort of does that.

src/common/get-configuration-file-model/app-version/app-version.injectable.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import packageInfo from "../../../../package.json";
+
+const appVersionInjectable = getInjectable({
+  id: "app-version",
+  instantiate: () => packageInfo.version,
+  causesSideEffects: true,
```
jim-docker, line 11: what side effect(s) are caused?

src/renderer/routes/current-route-component.injectable.ts
```diff
@@ -0,0 +1,33 @@
...
+    return computed(() => {
+      const currentRoute = route.get();
+
+      if (!currentRoute) {
+        return null;
+      }
+
+      const routeSpecificComponent = di
+        .injectMany(routeSpecificComponentInjectionToken)
+        .find(matches({ route: currentRoute }));
```
jim-docker, line 26: I feel that running this every time the currentRouteComponent is required will accelerate global warming. But it's javascript so...

jim-docker (commented):
still going...

src/renderer/routes/current-route-component.injectable.ts
jansav: Shouldn't be a problem since `computed` caches most of the usages.

src/common/front-end-routing/verify-that-all-routes-have-route-component.test.ts
jansav: Yes, definitely. Motivation of this test is to be friendly reminder that you need to add component for route.
Each route should have own set of tests which cover this too.

src/common/catalog-entities/general-catalog-entities/general-catalog-entities.injectable.ts
jansav: @Iku-turso fix didn't do it. Replaced it with: 

**sync-general-catalog-entities.injectable.ts**
```typescript
const generalCatalogEntities = di.injectMany(generalCatalogEntityInjectionToken);

// TODO: This shouldn't be reactive at all but catalogEntityRegistry accepts only reactive sources
const reactiveGeneralCatalogEntities = computed(() => generalCatalogEntities);

return () => {
  catalogEntityRegistry.addComputedSource(
    "lens:general",
    reactiveGeneralCatalogEntities,
  );
};
```

Notice that the requirement for being reactive comes from `catalogEntityRegistry` and in this PR we don't want to dig deeper. Sounds like there should be a way to add source which isn't reactive at all, but as said; lets figure out that in separate PR.

src/main/initializers/init-ipc-main-handlers/init-ipc-main-handlers.ts
jansav: That's better. Changed.

src/main/menu/get-app-menu-items.injectable.ts
jansav: True. Changed.

src/common/hotbar-store.injectable.ts
jansav: Removed all global shared state usages of HotbarStore since it wasn't that big thing. Implementation is in separate file again.

src/common/rbac.ts
jansav: Done.

src/common/k8s-api/endpoints/crd.api.ts
jansav: `navigateToCustomResourcesInjectable` provides function which navigates. This place requires URL, or we need to change all use places. Latter is bigger task which we scoped out based on checking use places.

jansav:
I'd say all critical review-comments have been taken care of. What do you think @Nokel81 @jim-docker? Are we ready to get this merged?

src/common/catalog-entities/general-catalog-entities/general-catalog-entities.injectable.ts
Nokel81: @jansav I don't think we should support that use case because a user might accidentally add it and then try to push to that array (or even modify the entities). 

I actually want to get rid of `observableSource` since that complicates things too, but certainly not in this PR.

src/common/get-configuration-file-model/app-version/app-version.injectable.ts
Nokel81: Mostly that tests could fail if this was not overridden. That field is only about testing (unit and integration but not e2e smoke). 

Because a test might do a `.execpt(....version...)` in a static string and if this injectable isn't overridden then that test will fail in the future.

jansav: Simply because we don't want to update unit tests whenever version is updated in `package.json`.

Nokel81 (approved):
My qualms about some specific design aspects should be all non-breaking to fix later.

jansav (approved):

src/common/get-configuration-file-model/app-version/app-version.injectable.ts
jim-docker: Would be clearer if `causesSideEffects`  was `mustOverrideForTests`. I was assuming there was more to this, guess I should've known this already :)

src/renderer/routes/current-route-component.injectable.ts
jim-docker: Does `computed` cache results for multiple state snapshots, or just the last?

Nokel81: Just the last, however I think that this will be cleaned up once we remove the state handling from installing and reinstalling extensions.

PR #5085, opened 2022-03-23
Behavioural unit tests for navigation and reference

Iku-turso:
As motivation, behavioural unit tests:
1. Test orchestration of main, renderer, etc., ie. the whole application, as if done by a user.
2. Work also as documentation. This documentation is always up to date by virtue of tests being green.
3. Cover a lot of production code with reasonably few lines of test code.
4. Are fast, and thus support more efficient development (eg. TDD).

See [conversation](https://miracloud.slack.com/archives/C014H7TELTW/p1649151978863679) for more info.

Not to be merged before #5084.

### DIFF
--- a/src/behaviours/add-cluster/navigation-using-application-menu.test.ts
+++ b/src/behaviours/add-cluster/navigation-using-application-menu.test.ts
@@ -1,0 +1,51 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import type { RenderResult } from "@testing-library/react";
+import { ApplicationBuilder, getApplicationBuilder } from "../../renderer/components/test-utils/get-application-builder";
+import isAutoUpdateEnabledInjectable from "../../main/is-auto-update-enabled.injectable";
+
+// TODO: Make components free of side effects by making them deterministic
+jest.mock("../../renderer/components/tooltip");
+jest.mock("../../renderer/components/monaco-editor/monaco-editor");
+
+describe("add-cluster - navigation using application menu", () => {
+  let applicationBuilder: ApplicationBuilder;
+  let rendered: RenderResult;
+
+  beforeEach(async () => {
+    applicationBuilder = getApplicationBuilder().beforeSetups(({ mainDi }) => {
+      mainDi.override(isAutoUpdateEnabledInjectable, () => () => false);
+    });
+
+    rendered = await applicationBuilder.render();
+  });
+
+  it("renders", () => {
+    expect(rendered.container).toMatchSnapshot();
+  });
+
+  it("does not show add cluster page yet", () => {
+    const actual = rendered.queryByTestId("add-cluster-page");
+
+    expect(actual).toBeNull();
+  });
+
+  describe("when navigating to add cluster using application menu", () => {
+    beforeEach(() => {
+      applicationBuilder.applicationMenu.click("file.add-cluster");
+    });
+
+    it("renders", () => {
+      expect(rendered.container).toMatchSnapshot();
+    });
+
+    it("shows add cluster page", () => {
+      const actual = rendered.getByTestId("add-cluster-page");
+
+      expect(actual).not.toBeNull();
+    });
+  });
+});
--- a/src/behaviours/cluster/__snapshots__/order-of-sidebar-items.test.tsx.snap
+++ b/src/behaviours/cluster/__snapshots__/order-of-sidebar-items.test.tsx.snap
@@ -1,0 +1,719 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`cluster - order of sidebar items when rendered renders 1`] = `
+<div>
+  <div
+    class="flex flex-col"
+    data-testid="cluster-sidebar"
+  >
+    <div>
+      <div
+        class=""
+        style="width: 40px; height: 40px;"
+      >
+        ??
+      </div>
+      <div />
+    </div>
+    <div
+      class="undefined sidebar-active-status"
+    >
+      <div
+        class="SidebarItem"
+        data-id-test="workloads"
+        data-is-active-test="false"
+        data-test-id="workloads"
+        data-testid="sidebar-item"
+      >
+        <a
+          class="nav-item flex gaps align-center expandable"
+          data-testid="sidebar-item-link-for-workloads"
+          href="/"
+        >
+          <i
+            class="Icon svg focusable"
+          >
+            <span
+              class="icon"
+            />
+          </i>
+          <span
+            class="link-text box grow"
+          >
+            Workloads
+          </span>
+          <i
+            class="Icon expand-icon box right material focusable"
+          >
+            <span
+              class="icon"
+              data-icon-name="keyboard_arrow_down"
+            >
+              keyboard_arrow_down
+            </span>
+          </i>
+        </a>
+      </div>
+      <div
+        class="SidebarItem"
+        data-id-test="config"
+        data-is-active-test="false"
+        data-test-id="config"
+        data-testid="sidebar-item"
+      >
+        <a
+          class="nav-item flex gaps align-center"
+          data-testid="sidebar-item-link-for-config"
+          href="/"
+        >
+          <i
+            class="Icon material focusable"
+          >
+            <span
+              class="icon"
+              data-icon-name="list"
+            >
+              list
+            </span>
+          </i>
+          <span
+            class="link-text box grow"
+          >
+            Config
+          </span>
+        </a>
+      </div>
+      <div
+        class="SidebarItem"
+        data-id-test="some-parent-id"
+        data-is-active-test="false"
+        data-test-id="some-parent-id"
+        data-testid="sidebar-item"
+      >
+        <a
+          class="nav-item flex gaps align-center expandable"
+          data-testid="sidebar-item-link-for-some-parent-id"
+          href="/"
+        >
+          <span
+            class="link-text box grow"
+          >
+            Some parent
+          </span>
+          <i
+            class="Icon expand-icon box right material focusable"
+          >
+            <span
+              class="icon"
+              data-icon-name="keyboard_arrow_down"
+            >
+              keyboard_arrow_down
+            </span>
+          </i>
+        </a>
+      </div>
+      <div
+        class="SidebarItem"
+        data-id-test="network"
+        data-is-active-test="false"
+        data-test-id="network"
+        data-testid="sidebar-item"
+      >
+        <a
+          class="nav-item flex gaps align-center expandable"
+          data-testid="sidebar-item-link-for-network"
+          href="/"
+        >
+          <i
+            class="Icon material focusable"
+          >
+            <span
+              class="icon"
+              data-icon-name="device_hub"
+            >
+              device_hub
+            </span>
+          </i>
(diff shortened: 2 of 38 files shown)


jansav:
Note: Updated base to make diff easier to read.

src/behaviours/cluster/__snapshots__/order-of-sidebar-items.test.tsx.snap
```diff
@@ -0,0 +1,719 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
panuhorsmalahti, line 1: I think large snapshots are not a good thing. It makes it difficult to review and understand diffs, which can hide regressions, and in particular exposes rather technical details like classnames to the diff. I think snapshots, if used, should be kept as small as semantically reasonable. Alternatively, the snapshot assertion could be replaced with a manully defined assertion (about the order, for example for this test).

jansav: These snapshots are real lifesavers some times (e.g. unexpected change in snapshot). Snapshots are used only as additions and functionality should be tested separately. That being said, it's nice to know what are the changes in DOM.

panuhorsmalahti: I think large snapshots have more costs than benefits and I wouldn't approve this PR with these.

src/behaviours/navigate-to-extension-page.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+      expect(currentPath.get()).toBe("/extension/some-extension-id");
+    });
+
+    it("query parameters is empty", () => {
+      expect(queryParameters.get()).toEqual({});
+    });
+
+    describe("when changing page parameters", () => {
+      beforeEach(() => {
+        const button = rendered.getByTestId("button-to-change-page-parameters");
```
panuhorsmalahti, line 61: I think getByTestId should be avoided and getByText etc. should be preferred.

See best practice in the official docs: https://testing-library.com/docs/queries/bytestid/

src/behaviours/cluster/__snapshots__/order-of-sidebar-items.test.tsx.snap
Nokel81: I also don't like that snapshot based testing locks in the DOM, I think that the user doesn't care at all about it and more cares about the text then if an extra `<div>` is added somewhere.

jansav: Again: Reasoning for snapshots is different.  Snapshots are very important for detecting unexpected changes in UI caused by some change in somewhere else.  

Please, provide more than you opinions. Concreate examples how you would solve the issue described here and in slack. Snapshots provide easy solution without minimal downsides (if any).

src/behaviours/add-cluster/navigation-using-application-menu.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    });
+
+    rendered = await applicationBuilder.render();
+  });
+
+  it("renders", () => {
+    expect(rendered.container).toMatchSnapshot();
+  });
+
+  it("does not show welcome page yet", () => {
```
jim-docker:
```suggestion
  it("does not show add cluster page yet", () => {
```

src/behaviours/add-cluster/navigation-using-application-menu.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    const actual = rendered.queryByTestId("add-cluster-page");
+
+    expect(actual).toBeNull();
+  });
+
+  describe("when navigating to add cluster using application menu", () => {
```
jim-docker, line 36: Do the two describes here both match the same snapshot? How do they get different results when looking for "add-cluster-page"?

src/behaviours/navigate-to-extension-page.test.tsx
jansav: Read the post linked in the page you provided. https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change

src/behaviours/add-cluster/navigation-using-application-menu.test.ts
jansav: Good catch.

src/behaviours/add-cluster/navigation-using-application-menu.test.ts
jansav: First snapshot test in this suite is snapshot that will match to "front page" of the application. However, currently it's just component that doesn't render any page component. Second snapshot matches to "add-cluster-page". 

- `it("shows add cluster page", () => {` tests that we are actually on that page and   
- `it("renders", () => {` tests how it's rendered at the moment. Covers a lot, since all of the sub components are rendered too and if there is change, it will break the snapshot and make sure you are attempting to change functionality in this page.

This is just a behaviour for navigating to that page. In future, there will be a lot more behaviours for "add-cluster-page" which will cover all of the functionalities. Most of these will be covered in other behaviours that actually do some action.

Examples of behaviours:
- Add single cluster
- Add multiple clusters
- Cancel 
- Navigate using application menu
- Navigate from catalog (most likely covered with all other behaviours)

src/behaviours/cluster/__snapshots__/order-of-sidebar-items.test.tsx.snap
panuhorsmalahti: I don't agree that this should be a requirement for testing. We want tests for UI components to test from a "user perspective", not asserting a specific DOM structure. If we have a good test coverage of unit tests of UI components we can be confident that the UI components work correctly, making snapshots unnecessary. I think that it's a rare edge case where these snapshots would be actionable for developers, thus making the downsides exceed the benefits.

The downsides of large snapshots are
a) difficult to review snapshot diffs, making them mostly useless
b) creating noise to diffs
c) having to update the snapshots when DOM changes
d) creating an incentive for developers to rely on a snapshot for asserting a semantic meaning of the DOM tree

src/behaviours/navigate-to-extension-page.test.tsx
panuhorsmalahti: I read it, and it agrees with my review comment..

To be more specific: using test ids is okay if other queries are not reliable / possible.

Quote from the post:
"Sometimes you can't reliably select an element by any of the other queries. For those, it's recommended to use data-testid"

Longer quote from the post:
"This is why [Testing Library](https://testing-library.com/) has the queries that it does. The queries help you to find elements in the same way that users will find them. These queries allow you to find elements by their [role](https://testing-library.com/docs/dom-testing-library/api-queries#byrole), [label](https://testing-library.com/docs/dom-testing-library/api-queries#bylabeltext), [placeholder](https://testing-library.com/docs/dom-testing-library/api-queries#byplaceholdertext), [text contents](https://testing-library.com/docs/dom-testing-library/api-queries#bytext), [display value](https://testing-library.com/docs/dom-testing-library/api-queries#bydisplayvalue), [alt text](https://testing-library.com/docs/dom-testing-library/api-queries#byalttext), [title](https://testing-library.com/docs/dom-testing-library/api-queries#bytitle), [test ID](https://testing-library.com/docs/dom-testing-library/api-queries#bytestid).

That's actually in the order of [recommendation](https://testing-library.com/docs/guide-which-query)."

The priority is even documented here:
https://testing-library.com/docs/queries/about/#priority

That's why I said that others should be preferred.

I'm fine with using test ids if others are not possible / practical for some reason.

src/behaviours/cluster/__snapshots__/order-of-sidebar-items.test.tsx.snap
jansav: > If we have a good test coverage of unit tests

We don't have. And we won't have in near future. And we won't have ever, as long as whole team doesn't write code in TDD.

-- 
a) Large snapshot diffs are not supposed to be reviewed. That doesn't mean they are useless.
b) Read a)

c) It's not downside that you get instant feedback from unit tests that something has changed. Current situation in code base is that components are re-used in so many places that unexpected places need to be caught. And this doesn't affect only components, even though components are the ones that are rendered. How many bugs we have seen caused by some change in somewhere else? I can answer that: **too many**. If there is a way to give instant feedback to developer, we should definitely use it.

d) This is where PR review comes to place. No functionality should be tested with snapshots.

panuhorsmalahti: Yes, the goal is to have good coverage.

The downside is having to do a manual step to update the snapshots (even if that step is pretty small).

src/behaviours/navigate-to-extension-page.test.tsx
jansav: There is no way to reliably select an element using `title`, `label`, `role`, `placeholder`, `text contents`, `display value` or `alt text`. Role is only one of those that is not changed too often. But role alone isn't enough. Therefore, only option is to select with `test ID`. I would even use plain old data-attributes if `testing-library` wouldn't limit us. 

Needs to be said again, I'm not interested in the tools and how we use them. Tool (here. testing-library) doesn't affect my mind in anyway. If the tool is not solving the problem, then we change the tool. I'm more interested about the problem that we are trying to solve. In this case it is: 

**How do we test something that user can act with, in a way that:**
1. we can minimize amount of unit tests that need to adapt for unrelated change
2. we can change our mind in the content
3. we can even change our mind on the element (e.g. could be button now, but in the future we decide that it's a link)
4. we can be sure enough that user find's the element (doesn't have to be same way, since it's just a burden. As long as abstraction levels are correct in the unit tests, we should be fine)

panuhorsmalahti: Clearly it's possible to reliably select an element using title, label, role, placeholder, text contents, display value or alt text.

Usually it's good to follow library's best practises.

jansav: How? Could you elaborate how you would solve the problem that I introduced in previous message?

src/behaviours/add-cluster/navigation-using-application-menu.test.ts
jim-docker: ah, sorry, missed that there are two snapshots in the snap file.

src/behaviours/navigate-to-extension-page.test.tsx
jim-docker: One could argue that changing the title is not an unrelated change. I wouldn't find it unreasonable to have to update the test if the title changed. I'd even expect to have to change the test if the element changed. The title and element shouldn't be changing that often that it is a huge burden to have to update the test too. In fact, having to update tests in this situation has the advantage of keeping you aware of them, their implementations, and how they drive development...

jim-docker (approved):
Certainly much careful hard work has gone into this and I'm confident that it can lead to a more stable and robust product. Although I still have some things to figure out before I'm comfortable with this di-based test framework. Lots to work by example from.

I think test names are suitable as documentation written as a user story...for users. I don't find the test implementations themselves that well documented, or easy to understand (I assume it comes with time). Also, test names are susceptible to staleness like code comments, they don't always get changed when the code changes.

Good work @jansav and @Iku-turso!

jansav (approved):

PR #5087, opened 2022-03-23
Yarn scripts for quality of life in development

Iku-turso:
Run and watch targetable unit- or integration-tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "build:mac": "yarn run compile && electron-builder --mac --dir",
     "build:win": "yarn run compile && electron-builder --win --dir",
     "integration": "jest --runInBand --detectOpenHandles --forceExit integration",
+    "test:unit": "jest --watch --testPathIgnorePatterns integration",
     "dist": "yarn run compile && electron-builder --publish onTag",
     "dist:dir": "yarn run dist --dir -c.compression=store -c.mac.identity=null",
     "download:binaries": "yarn run ts-node build/download_binaries.ts",
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "build:win": "yarn run compile && electron-builder --win --dir",
     "integration": "jest --runInBand --detectOpenHandles --forceExit integration",
     "test:unit": "jest --watch --testPathIgnorePatterns integration",
+    "test:integration": "func() { jest ${1:-xyz} --watch --runInBand --detectOpenHandles --forceExit --modulePaths=[\"<rootDir>/integration/\"]; }; func",
     "dist": "yarn run compile && electron-builder --publish onTag",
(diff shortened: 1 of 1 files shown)


Nokel81 (approved):

package.json
```diff
@@ -28,6 +28,8 @@
     "build:mac": "yarn run compile && electron-builder --mac --dir",
     "build:win": "yarn run compile && electron-builder --win --dir",
     "integration": "jest --runInBand --detectOpenHandles --forceExit integration",
+    "test:unit": "jest --watch --testPathIgnorePatterns integration",
```
jim-docker, line 31: I discovered you need to make sure you have an up-to-date build before running this

package.json
```diff
@@ -28,6 +28,8 @@
...
+    "test:integration": "func() { jest ${1:-xyz} --watch --runInBand --detectOpenHandles --forceExit --modulePaths=[\"<rootDir>/integration/\"]; }; func",
```
jim-docker, line 32: What file changes will cause tests to rerun? I edited/saved some files but no tests were run ("No tests found, exiting with code 0").

package.json
Iku-turso: Yes, this is intentional. The script is for development, and for that, it is optimized for quick start-up.

package.json
Iku-turso: For `unit tests`, watch triggers after a change in most production or test code (and also some configuration files). The initial behavior of `jest --watch` is to use git to find uncommitted production and test-code, and run all tests directly or indirectly affected by those changes. 

In your case: 
1. The changed files need to be related to a test for anything to happen.
2. Our current unrelated configuration expects `*.test`-files to contain at least one test in order to not be perceived as failure. This explains `"No tests found, exiting with code 0"`.

Deviating from this, for `integration tests` the watch triggers only for changes in test code, as the tests happen over already built application. Still, this makes tweaking the slow integration tests a bit more friendly :)

FYI: In watch mode, you can also use `w` to see additional controls. Eg. `p` or `t` will make you able to specify explicitly which tests to run, `a` will make you run all tests despite git, and `f` will make you re-run only failing tests. Cool stuff, right? :)

package.json
jim-docker: ok, bit messy if it's already running when you checkout a new branch

package.json
jim-docker: pretty sure I edited an integration test file (edited some string, maybe the test name). I'll try again

jim-docker: I added a test to `integration/__tests__/cluster-pages.tests.ts` and saved. No tests found. Pressing `o` runs the tests. Is this how to use it?

jim-docker (approved):

package.json
Iku-turso: We could have a sesh? Or all this should be documented somewhere within jest. Hmm.

PR #5098, opened 2022-03-24
Proposal: Add ContextProviders support to the Extension API

panuhorsmalahti:
* Draft PR for discussion of adding a new Extension API which allows extensions to register a React Context Provider
* At the moment this is not feasible as an Extension exports/configures multiple "entry points" that are rendered by the Lens
* The use case here is to support using React Context in extensions, which will reduce boilerplate / prop drilling
* Draft PR, missing documentation and tests
* More information about Context and it's use cases: https://reactjs.org/docs/context.html
* If the value of the Provider changes, only the components under the Consumer(s) are re-rendered
* Performance consideration: the whole app's root might be re-rendered during loading multiple extensions when the application is being loaded
* Note: each extension can provide a single component that contains multiple Providers

Example code in an extension (in renderer):
```
  getContextProviders = async () => {
    return [{
      components: {
        Provider: LensPlatformStoreProvider
      }
    }];
  };
```

The interface is similar to entity settings for consistency.


panuhorsmalahti:
It was discussed that we could make this addition to the new Extension API instead of the legacy (=current) API.

PR #5107, opened 2022-03-25
Add ability to hide KubeObjectMenu Edit and Remove buttons in extensions

### LINKED ISSUE #5105: Allow extensions to hide 'Edit' and 'Remove' while using KubeObjectDetails component
Now we check for store methods in https://github.com/lensapp/lens/blob/master/src/renderer/components/kube-object-menu/kube-object-menu.tsx#L43 which is not always the best approach.
edit remove buttons

Nokel81:
- The core of the solution is to add event listener handlers for the `onContextMenuOpen` event for `KubeObject` in a similar fashion to that of catalog entities. This allows any extension to add/remove items and means that the producers don't have to worry about things like `<Menu>` and `<MenuItem>` which the old `kubeObjectMenuItems` registration pattern on `LensRendererExtension` requires.

fixes #5105


Nokel81:
fixes https://github.com/lensapp/lens/issues/5105

aleksfront (commented):
Should we set default `true` values for brand new `showEditButton` & `showRemoveButton` fields to prevent breaking changes? I guess all working extensions showing those button by default?

Nokel81:
@aleksfront I don't understand the stores have `true` as default. I just default to`false` if there is neither a store nor a provided `remove` function.

src/renderer/components/kube-object-menu/kube-object-menu.tsx
```diff
@@ -41,11 +41,11 @@ class NonInjectedKubeObjectMenu<TKubeObject extends KubeObject> extends React.Co
   }
 
   get isEditable() {
-    return this.props.editable ?? Boolean(this.store?.patch);
+    return this.props.editable ?? this.store?.showEditButton ?? false;
```
ixrock: Having `showEditButton` and other field on `KubeObjectStore` just weird as heck.
Why only `props.editable` not enough to control this items?

Nokel81: I agree it is extremely weird.

Nokel81: It is not enough because the problem here is the fact that the extension API does not have a mechanism to remove these buttons for any kube objects, let alone for new kube objects which an extension might actually control.

Another solution would be to require extension developers to add these buttons explicitly but that is a breaking change.

src/common/k8s-api/kube-object.store.ts
```diff
@@ -64,6 +64,24 @@ export abstract class KubeObjectStore<T extends KubeObject> extends ItemStore<T>
...
   @observable private loadedNamespaces?: string[];
 
+  /**
+   * The field is used for configuring whether the edit button is displayed by default from
+   * {@link KubeObjectMenu} for detail views of instances of `T`.
+   *
+   * This should only be set once, in the constructor of base classes since the views will
+   * not be reactive to any changes to this value.
+   */
+  public readonly showEditButton = true;
```
jakolehm: IMHO it's not good practice to mix UI logic to store/api level. Maybe these could be called as `isEditable` & `isRemovable` ?

Nokel81: I agree that we should not mix them. But I don't see exactly how changing the name would help.

Maybe we could just extract this into some sort of extension API call which can be called to turn it off for any store. WDYT?

jakolehm: `showEditButton` says that it affects the edit button visibility. `isEditable` is much more generic, edit button can use this information but so can other functionality. Small detail but I think we should use more generic names if possible.

Nokel81: Okay will change.

jakolehm (commented):
Quite a big diff but probably all changes have a good reason. Please update the PR description to match the implementation (and maybe add why so many changes were required).

src/common/catalog/catalog-entity.ts
```diff
@@ -278,11 +278,22 @@ export interface CatalogEntitySettingsMenu {
   };
 }
 
+export interface CatalogEntityContextMenuNavigate {
```
Nokel81, line 281: This just clarifies that this is actually (previous to this PR) supported here witch some additional documentation. 

This probably could be moved into a separate PR if desired.

src/extensions/lens-renderer-extension.ts
```diff
@@ -49,6 +50,7 @@ export class LensRendererExtension extends LensExtension {
   topBarItems: TopBarRegistration[] = [];
   additionalCategoryColumns: AdditionalCategoryColumnRegistration[] = [];
   customCategoryViews: CustomCategoryViewRegistration[] = [];
+  kubeObjectHandlers: KubeObjectHandlerRegistration[] = [];
```
Nokel81, line 53: This is how the event listeners are registered

src/extensions/renderer-api/components.ts
```diff
@@ -43,14 +47,23 @@ export type {
 } from "../../renderer/components/+catalog/custom-category-columns";
 
 // other components
+export type {
+  ConfirmDialogBooleanParams,
+  ConfirmDialogParams,
+  ConfirmDialogProps,
+};
+export const ConfirmDialog = Object.assign(_ConfirmDialog, {
```
Nokel81, line 55: It was required to transform `ConfirmDialog` to use dependency injection. This was required because confirm dialog was not cooperating with behaviour unit tests (it had global shared state).

Because this transformation was required, I needed to move the `open` and `confirm` functions away from being statics on the component to be there own injectables. This rectifies that for the extension API only.

src/renderer/catalog/normalize-menu-item.injectable.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export interface NormalizedCatalogEntityContextMenu {
+  title: string;
+  icon?: string;
+  onClick: () => void;
+}
+
+export type NormalizeCatalogEntityContextMenu = (menuItem: CatalogEntityContextMenu) => NormalizedCatalogEntityContextMenu;
+
+const normalizeCatalogEntityContextMenuInjectable = getInjectable({
```
Nokel81, line 17: Because `ConfirmDialog` needed to be DI-ed, I took the opportunity to make this code more DRY. It was repeated 3 or 4 times throughout the codebase previously.

src/renderer/components/+extensions/attempt-install-by-info.injectable.tsx
```diff
@@ -0,0 +1,138 @@
+/**
```
Nokel81, line 1: Even git makes this code look like it is completely new, it is not. It is mostly just the renaming of the file previously at `+extensions/attempt-install-by-info/attempt-install-by-info.tsx`, just with the `getInjectable` call added, the factory not exported, and `confirm` (which is not injectable) added as a dependency.

src/renderer/components/+extensions/confirm-uninstall-extension.injectable.tsx
```diff
@@ -0,0 +1,54 @@
+/**
```
Nokel81, line 1: This file's diff is similar to that of the `attempt-install-by-info` ones

src/renderer/components/animate/animate.tsx
```diff
@@ -46,15 +46,24 @@ export class Animate extends React.Component<AnimateProps> {
     return React.Children.only(this.props.children) as React.ReactElement<React.HTMLAttributes<any>>;
   }
 
+  private toggle(enter: boolean) {
```
Nokel81, line 49: I made this change while debugging why `Menu` and `Drawer` were being non-deterministic in behavioural unit test. This change doesn't change the behaviour and ultimately I just decided to remove the use of `<Animate>` from the tests.

The other option would be to make `requestAnimateFrame` (which has side effects) a dependency and override it in tests. Could be very useful for future work.

src/renderer/components/confirm-dialog/with-confirm.injectable.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import type { ConfirmDialogParams } from "./confirm-dialog";
+import { getInjectable } from "@ogre-tools/injectable";
+import openConfirmDialogInjectable from "./open.injectable";
+
+export type WithConfirmation = (params: ConfirmDialogParams) => () => void;
```
Nokel81, line 10: This is a useful higher order function which can better express what the intent is. Namely, that I want to confirm something as a callback instead of immediately.

jakolehm (approved):

src/renderer/components/drawer/drawer.tsx
```diff
@@ -199,7 +199,7 @@ class NonInjectedDrawer extends React.Component<DrawerProps & Dependencies, Stat
                 )}
               </div>
               {toolbar}
-              <Icon material="close" onClick={this.close}/>
+              <Icon material="close" tooltip="Close" onClick={this.close}/>
```
Nokel81, line 202: I noticed this bug while testing, it is the only icon for drawers that doesn't have a tooltip.

This could be moved to a separate PR.

src/renderer/components/icon/icon.tsx
```diff
@@ -15,22 +15,69 @@ import { withTooltip } from "../tooltip";
...
-  link?: LocationDescriptor;   // render icon as NavLink from react-router-dom
-  href?: string;              // render icon as hyperlink
-  size?: string | number;     // icon-size
-  small?: boolean;            // pre-defined icon-size
-  smallest?: boolean;            // pre-defined icon-size
-  big?: boolean;              // pre-defined icon-size
-  active?: boolean;           // apply active-state styles
-  interactive?: boolean;      // indicates that icon is interactive and highlight it on focus/hover
-  focusable?: boolean;        // allow focus to the icon + show .active styles (default: "true", when icon is interactive)
+export interface BaseIconProps {
```
Nokel81, line 18: I separated out these props so that I could export this subset as part of the `KubeObjectHandlerRegistration` type.

The bonus of using doc comments it that our API docs should be better documented now.

src/renderer/components/kube-object-menu/dependencies/static-kube-object-menu-items.ts
```diff
@@ -37,11 +36,4 @@ export const staticKubeObjectMenuItems = [
       MenuItem: ReplicaSetMenu,
     },
   },
-  {
```
Nokel81, line 40: I changed this single `KubeObjectMenu` registration into a `KubeObjectHandler` registration to show in this PR how it can be done.

src/renderer/components/layout/sidebar-cluster.tsx
```diff
@@ -86,8 +64,17 @@ const NonInjectedSidebarCluster = observer(({ clusterEntity, hotbarStore }: Depe
       ? () => hotbarStore.removeFromHotbar(clusterEntity.getId())
       : () => hotbarStore.addToHotbar(clusterEntity);
 
-    contextMenu.menuItems = [{ title, onClick }];
-    clusterEntity.onContextMenuOpen(contextMenu);
+    menuItems.replace([{ title, onClick }]);
+    clusterEntity.onContextMenuOpen({
+      menuItems,
+      navigate: (url, forceMainFrame = true) => {
```
Nokel81, line 70: This is copied from line 25 in the left hand side of the diff. And is why I updated the typings above.

src/renderer/components/+extensions/attempt-install-by-info.injectable.tsx
```diff
@@ -0,0 +1,138 @@
...
+  attemptInstall,
+  getBaseRegistryUrl,
+  extensionInstallationStateStore,
+  confirm,
+}: Dependencies): AttemptInstallByInfo => (
+  async (info) => {
+    const { name, version, requireConfirmation = false } = info;
+    const disposer = extensionInstallationStateStore.startPreInstall();
+    const baseUrl = await getBaseRegistryUrl();
+    const registryUrl = new URLParse(baseUrl).set("pathname", name).toString();
```
ixrock, line 47: Why not using standard `new URL()`? (it's also available in `nodejs`)

```ts
const registryUrl = new URL(baseUrl);
registryUrl.pathname = name;
registryUrl.toString(); // result
```

src/common/catalog/catalog-entity.ts
```diff
@@ -278,11 +278,22 @@ export interface CatalogEntitySettingsMenu {
...
+  /**
+   * @param pathname The location to navigate to in the main iframe
+   */
+  (pathname: string, forceMainFrame?: boolean): void;
+  /**
+   * @param pathname The location to navigate to in the current iframe. Useful for when called within the cluster frame
+   */
+  (pathname: string, forceMainFrame: false): void;
```
ixrock, line 289: A bit weird definition. Would be more logical imho if by default with `false` or `undefined` second argument it will use current iframe, and force main iframe only with optional `forceMainFrame: true`. Wdyt?

src/renderer/components/confirm-dialog/confirm-dialog.tsx
```diff
@@ -124,10 +114,10 @@ export class ConfirmDialog extends React.Component<ConfirmDialogProps> {
       <Dialog
         {...dialogProps}
         className={cssNames("ConfirmDialog", className)}
-        isOpen={dialogState.isOpen}
+        isOpen={isOpen}
         onClose={this.onClose}
         close={this.close}
-        {...(dialogState.isOpen ? { "data-testid":"confirmation-dialog" } : {})}
+        {...(isOpen ? { "data-testid": "confirmation-dialog" } : {})}
```
ixrock, line 120: Why `data-testid` passed only when dialog is open?

src/renderer/components/layout/sidebar-cluster.tsx
```diff
@@ -7,74 +7,52 @@ import styles from "./sidebar-cluster.module.scss";
...
+  const [menuItems] = useState(observable.array<CatalogEntityContextMenu>());
 
   if (!clusterEntity) {
-    return renderLoadingSidebarCluster();
+    // render a Loading version of the SidebarCluster
+    return (
+      <div className={styles.SidebarCluster}>
+        <Avatar
+          title="??"
+          background="var(--halfGray)"
```
ixrock, line 49: Why not adding background directly to `styles.loadingAvatar`?

src/renderer/components/menu/menu.tsx
```diff
@@ -319,28 +321,40 @@ export class Menu extends React.Component<MenuProps, State> {
...
-    const menu = (
-      <MenuContext.Provider value={this}>
+    let menu = (
+      <ul
+        id={id}
+        ref={this.bindRef}
+        className={className}
+        style={{
+          left: this.state?.menuStyle?.left,
+          top: this.state?.menuStyle?.top,
```
ixrock, line 331: `this.state` might not exists?

ixrock (approved):

src/renderer/components/confirm-dialog/confirm-dialog.tsx
Nokel81: I questioned this myself, it is so that while the dialog is closed `findByTestId` won't find it.

src/renderer/components/+extensions/attempt-install-by-info.injectable.tsx
Nokel81: Mostly habit, but fair point, and given that they now recommend doing that we probably should as well. I will make a follow up PR to remove the `url-parse` library from our repo.

src/common/catalog/catalog-entity.ts
Nokel81: This documentation was done to match the current implementation, yeah the default is a bit confusing.

src/renderer/components/layout/sidebar-cluster.tsx
Nokel81: Because inline styles (which is what `<Avatar>` uses for `background`, `width`, and `height`) overrides class based styling, and background is something is always set by code.

src/renderer/components/menu/menu.tsx
Nokel81: Nope, just copy pasted it the old code. Probably should be cleaned up in the future.

src/renderer/components/confirm-dialog/confirm-dialog.tsx
ixrock: In general we should not render dialog itself, instead of hiding reachability in tests. 
But this is fine for now I guess.

src/renderer/components/layout/sidebar-cluster.tsx
ixrock: We could use `background: var(--halfGray) !important;` (let's try next time avoid using css-variables inside components as possible == easier refactoring at later point / separation of concerns)

PR #5108, opened 2022-03-25
Fix slackLink in catalog becoming out of date

### LINKED ISSUE #5025: catalog link to slack workspace is "no longer active"
**Describe the bug**

go to the catalog. click on the link called "lens community slack" . be greeted with "this link is no longer active"

**To Reproduce**

**Expected behavior**

**Screenshots**

- Lens Version:
- OS: [e.g. OSX]

**Logs:**
```
Your logs go here...
```

**Kubeconfig:**
```
your kubeconfig here
```

**Additional context**

Nokel81:
- Add migration to make default weblinks' ids not their URLs

- Add migration to always update the slack weblink if still present

### DIFF
--- a/src/migrations/weblinks-store/5.1.4.ts
+++ b/src/migrations/weblinks-store/5.1.4.ts
@@ -7,6 +7,13 @@ import { docsUrl, slackUrl } from "../../common/vars";
 import type { WeblinkData } from "../../common/weblink-store";
 import type { MigrationDeclaration } from "../helpers";
 
+export const lensWebsiteLinkName = "Lens Website";
+export const lensDocumentationWeblinkName = "Lens Documentation";
+export const lensSlackWeblinkName = "Lens Community Slack";
+export const lensTwitterWeblinkName = "Lens on Twitter";
+export const lensBlogWeblinkName = "Lens Official Blog";
+export const kubernetesDocumentationWeblinkName = "Kubernetes Documentation";
+
 export default {
   version: "5.1.4",
   run(store) {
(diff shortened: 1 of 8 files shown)


src/migrations/weblinks-store/5.4.5-beta.1.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import { kubernetesDocumentationWeblinkId, lensBlogWeblinkId, lensDocumentationWeblinkId, lensSlackWeblinkId, lensTwitterWeblinkId, lensWebsiteWeblinkId } from "../../common/vars";
+import type { WeblinkData } from "../../common/weblink-store";
+import type { MigrationDeclaration } from "../helpers";
+import { kubernetesDocumentationWeblinkName, lensBlogWeblinkName, lensDocumentationWeblinkName, lensSlackWeblinkName, lensTwitterWeblinkName, lensWebsiteLinkName } from "./5.1.4";
+
+export default {
+  version: "5.4.5-beta.1 || >=5.5.0-alpha.0",
```
Nokel81, line 12: This is here so that users who get `5.4.5` will have this update as well as those running `master`

aleksfront:
Please describe how this will prevent slack link becoming out of date.

If I understand it right, we rewrite slack link url after each app upgrade from the value located in 
```
src/common/vars.ts
```

Does it mean we should rewrite link inside `vars.ts` from time to time manually with separate PRs?

Also, as far as I remember, the link expiration problem was related to Lens channel restrictions which doesn't allow to use links for more than 2000 users or so?

Nokel81:
This doesn't prevent the link from becoming out of date but that is not actually the problem that the user is seeing.

The problem that the user is seeing is that their Weblink in the catalog called `Slack` as the value of the old slack link because that entry was added to the WeblinkStore before the update to the slack link var was done.

This PR makes it so that any update to the slackLink variable will also appear in the catalog

aleksfront (approved):

src/migrations/weblinks-store/5.1.4.ts
```diff
@@ -7,19 +7,26 @@ import { docsUrl, slackUrl } from "../../common/vars";
...
-      { id: slackUrl, name: "Lens Community Slack", url: slackUrl },
-      { id: "https://kubernetes.io/docs/home/", name: "Kubernetes Documentation", url: "https://kubernetes.io/docs/home/" },
-      { id: "https://twitter.com/k8slens", name: "Lens on Twitter", url: "https://twitter.com/k8slens" },
-      { id: "https://medium.com/k8slens", name: "Lens Official Blog", url: "https://medium.com/k8slens" },
+      { id: "https://k8slens.dev", name: lensWebsiteLinkName, url: "https://k8slens.dev" },
+      { id: docsUrl, name: lensDocumentationWeblinkName, url: docsUrl },
+      { id: slackUrl, name: lensSlackWeblinkName, url: slackUrl },
+      { id: "https://twitter.com/k8slens", name: lensTwitterWeblinkName, url: "https://twitter.com/k8slens" },
+      { id: "https://medium.com/k8slens", name: lensBlogWeblinkName, url: "https://medium.com/k8slens" },
+      { id: "https://kubernetes.io/docs/home/", name: kubernetesDocumentationWeblinkName, url: "https://kubernetes.io/docs/home/" },
```
ixrock, line 29: Why to duplicate `id` & `url`?

ixrock (approved):
This fix looks a bit over engineered or maybe I missing something.. 

But question is: why do even need to save static links (I assume from `vars.ts`) to `WebLinkStore` (which saves data to `.json`) if those might get outdated at some point?

src/migrations/weblinks-store/5.1.4.ts
Nokel81: the type `Weblink` has the following definition:

```ts
interface Weblink {
  id: string;
  name: string;
  url: string;
}
```

And in most cases (namely when a user adds a new weblink) the `id` will be a `uuid.v4`.

When these were added, whoever added them didn't do that and instead used the URLs as the IDs.

The change in this file is to have it so that the names of all of the default weblinks are constants declared in `vars.ts`. (Users cannot edit the names of weblinks).

---

As for why this is all necessary. Users are able to delete weblinks, even the default ones. So the other migrations are present to change the IDs (to some constants which are not UUIDs but also not URLs). And then the third one is run on every new version to make sure that the stored weblink is the current value from `vars.ts`.

---

Yes this is pretty complex, however, given that users can delete weblinks, and that I didn't want to have to keep all previous slack links in the repo, this is how I think it needs to be done.

Nokel81:
Will make a new PR updating that one migration value (since now we are targeting 5.4.6 instead of 5.4.5)

Issue #2243, opened 2021-02-26
Lens becomes unusably laggy in a large cluster


Double0negative:
**Describe the bug**
Lens becomes unusably laggy with a large cluster, to the point where it becomes completely frozen if the size grows enough. I am attempting to use it on a cluster with ~20k pods (many hundreds starting/stopping per minute) and the app completely hangs and never recovers when attempting to do anything (especially the pods tab). 

**To Reproduce**
1. Have significantly large cluster and connect
2. open pods tab
3. Lens will freeze after a few seconds

**Expected behavior**
Lens doesn't freeze.

**Environment (please complete the following information):**
- Lens Version: 4.1.3
- OS: [e.g. OSX]

jim-docker:
Are these pods all in a single namespace?

Double0negative:
No, but a vast majority are. However it still lags just as much if I have the namespace hidden. If theres anything I can do to help get you info let me know

Nokel81:
The solution for this is probably related to streaming results to stores. Though the fact that you said that the pods are in the 100s starting and stopping per minute is certainly the bigger problem.

Are you using pods for individual requests? I would like to have some understanding behind why so many pods are being stopped and started. That should help us prioritize this issue.

Double0negative:
We are a game server hosting, so we have lots user requested servers spinning up and down constantly.

Nokel81:
Thanks for the information. I guess we need some feedback.

How "up to date" do you want your information to be if the number of pods is constantly changing?

Double0negative:
I think at that size/rate, the most important thing is being able to find the pod when searching for it. So as long as the search is able to find new pods, it doesn't matter as much if the full list is "live". For smaller, less active namespaces, its far more important for the list to be "live"

Nokel81:
Thanks for the information. I think I will flesh this out into a feature request tomorrow and then close this then.

Nokel81:
@Double0negative Would you be fine if you were required to mark a cluster like this as "large" (or something similar, ignore bikeshedding) and have the system instead of doing streaming we do slow polling and fall back to querying the kube server for an item if we don't find it?

Double0negative:
Yea that would be fine. If possible it would be preferable to do this per namespace, but if not on the whole cluster would be ok

Nokel81:
[...]

PR #5114, opened 2022-03-28
Optimise table list rendering

aleksfront:
Part of #2243 

* Use `React.memo` for `getRow` table method to avoid rerender for all table rows
* Clean up DOM from Menus and Tooltips. Create specific html elements only on hover/click
* Increase events buffer delay from 1s to 4s, which triggers table render less often

https://user-images.githubusercontent.com/9607060/160387267-52e64720-56d1-4a6b-8d48-4bf264cd3746.mov

CPU consumption is still high, the next step might be `bindWatchEventsUpdater` method optimisation https://github.com/lensapp/lens/compare/master...fix-big-cluster-table-freeze#diff-ac63bb8652e3d96654d31988d5d2fd31c10d780581a5d3852d19d7260d102c28R379. Currently it triggers bunch of array calculations each 4s. We can:
* Convert `ItemStore` items to Maps instead of Arrays
* Skip items manipulations for workloads not visible at current moment
* Use separate CPU thread for these using Web Workers https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers


src/common/k8s-api/kube-object.store.ts
```diff
@@ -457,6 +457,9 @@ export abstract class KubeObjectStore<T extends KubeObject> extends ItemStore<T>
     const items = this.getItems();
 
     for (const { type, object } of this.eventsBuffer.clear()) {
+      // TODO: Convert items from Array to Map to avoid doing findIndex many times per second (events delay)
+      // Also, its worth to use Web Worker for such operations to prevent blocking UI
```
Nokel81: Web worker for the watching would be great! Though updating the state probably would have to be done in the UI thread.

ixrock: Is this main bottleneck? _(converting `Array -> Map` o_O)_ 
Why not keep then data always in `Map`?

Nokel81: No I think the main bottleneck is the fact that data is currently only an `Array` so to update an entry it is an O(n) operation. I think that is what Alex is talking about.

aleksfront: Yes, I believe converting store items to maps will speed things up.

aleksfront:
Renamed PR since it's not fully fixing #2243

src/common/k8s-api/kube-object.store.ts
```diff
@@ -377,7 +377,7 @@ export abstract class KubeObjectStore<T extends KubeObject> extends ItemStore<T>
   // collect items from watch-api events to avoid UI blowing up with huge streams of data
   protected eventsBuffer = observable.array<IKubeWatchEvent<KubeJsonApiData>>([], { deep: false });
 
-  protected bindWatchEventsUpdater(delay = 1000) {
+  protected bindWatchEventsUpdater(delay = 4000) {
```
jansav, line 380: Arbitrary delays are bad. Could you explain why it is needed?

src/renderer/components/item-object-list/content.tsx
```diff
@@ -202,7 +180,7 @@ export class ItemListLayoutContent<I extends ItemObject> extends React.Component
       return null;
     }
 
-    return this.props.getItems().map(item => this.getRow(item.getId()));
+    return this.props.getItems().map(item => this.getRow(item.getId(), item));
```
jansav, line 183: Why two separate parameters? `item` should be enough.

src/renderer/components/+catalog/internal-category-columns.tsx
```diff
@@ -119,3 +86,37 @@ export const defaultCategoryColumns: RegisteredAdditionalCategoryColumn[] = [
     searchFilter: entity => entity.status.phase,
   },
 ];
+
+const EntityName = observer(({ entity }: { entity: CatalogEntity }) => {
+  const hotbarStore = HotbarStore.getInstance();
```
jansav: Usage of shared global state. Pass hotbar store as prop.

jansav (changes requested):

src/common/k8s-api/kube-object.store.ts
aleksfront: Increased delay because triggering `bindWatchEventsUpdater` every second results very costly table re-render process as well as bunch of Array manipulations (currently trying to optimise them in another branch).

src/renderer/components/+catalog/internal-category-columns.tsx
Nokel81: And this is now easier because in `master` branch `HotbarStore` is fully injectable

src/renderer/components/item-object-list/content.tsx
Nokel81: Unfortunately not because `getRow` must be match the `(id: string) => React.ReactNode` signature (at least I thought it did).

src/common/k8s-api/kube-object.store.ts
Nokel81: Wouldn't it be fine if this was an `@action` so that all the updates were clumped together, maybe that would also help with the performance issues.

aleksfront: The down-level method already has `@action` decorator on it.

src/renderer/components/item-object-list/content.tsx
aleksfront: Yes, it all goes from the legacy part. Virtual list rows will use following notation for ` getRow` prop:
```
getRow?: (uid: string | number, item?: ItemObject) => React.ReactElement<TableRowProps>;
```

All existing extensions using virtual tables are waiting a `uid` arguments for their internal `getRow` functions. But obviously passing an item itself is much faster than later looking for it in the `items` array. So I decided to use second optional argument for that.

src/renderer/components/+catalog/internal-category-columns.tsx
aleksfront: Done

src/common/k8s-api/kube-object.store.ts
Nokel81: True that should be good enough for grouping rerenders

aleksfront:
Considering to close this PR since it has changes from the different areas. Will create more precise PRs instead of this one. For instance - https://github.com/lensapp/lens/pull/5168

PR #5117, opened 2022-03-28
Always preload app fonts at earliest stages

### LINKED ISSUE #5019: RobotoMono font not loaded when terminal opened first time
To reproduce:
1. If located in cluster view, choose "Cluster" -> "Disconnect" from the sidebar menu
2. Open cluster again.
3. Open terminal tab.

fonts not loaded

Firstly, this issue occurred within webpack 5 update: 
* https://github.com/lensapp/lens/pull/4867
* https://github.com/lensapp/lens/pull/4725#issuecomment-1040158630

ixrock:
Terminal font wasn't loaded in some specific conditions of app working.
Reproduced: after disconnect  & connect to active cluster with opened terminal tab.
From now all fonts used in app must be manually declared in `template.html` and enabled for preloading in `fonts.scss`

fix #5019

### DIFF
--- a/src/renderer/components/mixins.scss
+++ b/src/renderer/components/mixins.scss
@@ -60,14 +60,19 @@
   }
 }
 
-// Makes custom font available at earlier stages (start preloading)
-// Element must be in DOM and contain some text of loading font.
-@mixin font-preload() {
+// Makes custom @font-family available at earlier stages.
+// Element must exist in DOM as soon as possible to initiate preloading.
+@mixin font-preload($fontFamily: "RobotoMono") {
+  position: absolute;
+  visibility: hidden;
+  height: 0;
+
   &:before {
     width: 0;
     display: block;
     overflow: hidden;
-    content: "x"; // required
-    @content; // must contain `font-family`
+    content: "x"; // some text required to start applying font in document
+    font-family: $fontFamily; // imported name in @font-face declaration
+    @content;
   }
 }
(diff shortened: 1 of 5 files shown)


src/renderer/components/mixins.scss
```diff
@@ -60,14 +60,19 @@
   }
 }
 
-// Makes custom font available at earlier stages (start preloading)
-// Element must be in DOM and contain some text of loading font.
-@mixin font-preload() {
+// Makes custom @font-family available at earlier stages.
+// Element must exist in DOM as soon as possible to initiate preloading.
+@mixin font-preload($fontFamily: "RobotoMono") {
```
Nokel81: Do we need to specify the default param value here?

Nokel81 (dismissed):

src/renderer/components/mixins.scss
ixrock: Nope, provided just as an example.

Nokel81: Would it be better to use doc comments?

 https://sass-lang.com/documentation/syntax/comments#documentation-comments

ixrock: Nope, it wouldn't better.. 
I can remove the default value but it really doesn't matter _(code should be self-documented as possible imho)._

Nokel81 (approved):

jim-docker (approved):
LGTM. Found 2 bugs, but they were pre-exisiting, will open issues

PR #5142, opened 2022-03-30
Fix: remove HotbarIcon rounded corner artifacts

aleksfront:
Follows #5131 

Before:
artifacts

After:
hotbar icons
no artifacts

### DIFF
--- a/src/renderer/components/avatar/avatar.tsx
+++ b/src/renderer/components/avatar/avatar.tsx
@@ -57,14 +57,7 @@ function getLabelFromTitle(title: string) {
 
 export function Avatar(props: AvatarProps) {
   const { title, variant = "rounded", size = 32, colorHash, children, background, imgProps, src, className, disabled, ...rest } = props;
-
-  const getBackgroundColor = () => {
-    if (src) {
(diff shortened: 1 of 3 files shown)


src/renderer/components/avatar/avatar.tsx
```diff
@@ -81,7 +74,11 @@ export function Avatar(props: AvatarProps) {
         [styles.rounded]: variant == "rounded",
         [styles.disabled]: disabled,
       }, className)}
-      style={{ width: `${size}px`, height: `${size}px`, backgroundColor: getBackgroundColor() }}
+      style={{
+        width: `${size}px`,
+        height: `${size}px`,
```
ixrock, line 79: Pixels are default numeric values in React's inline style prop.

```suggestion
        width: size,
        height: size,
```

ixrock (approved):

Nokel81 (approved):

src/renderer/components/avatar/avatar.tsx
aleksfront: Didn't know that. Thanks.

PR #5149, opened 2022-03-30
Reset ClusterStatus state when switching clusters

### LINKED ISSUE #5134: Connecting screen doesn't refresh after failed connection to a different cluster
If a cluster connection fails, and an attempt is then made to connect to a different cluster, the failure message from the first cluster connection attempt is not cleared:

https://user-images.githubusercontent.com/40840436/160646997-ea168447-d97f-4043-8413-4495d550b445.mov

happens on 5.4.4

Nokel81:

fixes #5134

### DIFF
--- a/src/renderer/components/cluster-manager/cluster-status.tsx
+++ b/src/renderer/components/cluster-manager/cluster-status.tsx
@@ -55,6 +55,13 @@ export class ClusterStatus extends React.Component<ClusterStatusProps> {
     ]);
   }
 
+  componentDidUpdate(prevProps: Readonly<ClusterStatusProps>): void {
+    if (prevProps.cluster.id !== this.props.cluster.id) {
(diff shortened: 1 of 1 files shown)


aleksfront (approved):

jim-docker (approved):

jim-docker:
One ramification of this change is if you try to connect cluster A (via hotbar), and it shows an error, then try to connect to cluster B (via hotbar), then click cluster A hotbar icon, while it is still trying to connect, the previously encountered error message is cleared. Though maybe this is intended/acceptable behaviour?

Nokel81:
I guess that is sort of a problem, it was sort of intended though because we don't want to show previous errors when connecting and switching between clusters while connecting is a bit strange. We could fix it though I guess.

PR #5156, opened 2022-03-31
Add optional badge to Catalog Category label

### LINKED ISSUE #5146: Extension API for Catalog Category badge
Allow extension to specify a badge component to be displayed after the Category label in the Catalog Category menu.

Example usage in extension:
```
export default class extends Renderer.LensExtension {
  customCategoryViews = [
    {
      group: "entity.k8slens.dev",
      kind: "KubernetesCluster",
      priority: 10,
      components: {
        View: MyKubernetesClusterView,
        Badge: MyBadge
      },
    },
  ];
}
```

panuhorsmalahti:
* Fixes https://github.com/lensapp/lens/issues/5146
* Makes it possible for extensions to specify a badge for a custom Catalog Category
* Add `badgeBackgroundColor` CSS variable for extension use

Example use by an extension:
Screenshot 2022-03-31 at 16 17 35

### DIFF
--- a/src/renderer/components/+catalog/catalog-category-label.tsx
+++ b/src/renderer/components/+catalog/catalog-category-label.tsx
@@ -3,19 +3,22 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
 import React from "react";
+import type { CatalogCategory } from "../../../common/catalog/catalog-entity";
 
 interface CatalogCategoryLabelProps {
-  label: string | React.ReactNode;
-  badge?: React.ReactNode;
+  category: CatalogCategory;
 }
 
 /**
  * Display label for Catalog Category for the Catalog menu
  */
-export const CatalogCategoryLabel = ({ label, badge }: CatalogCategoryLabelProps) =>
-  (
+export const CatalogCategoryLabel = ({ category }: CatalogCategoryLabelProps) => {
+  const badge = category.getBadge();
+
+  return (
     <div className="flex">
-      <div>{label}</div>
+      <div>{category.metadata.name}</div>
       {badge ? (<div className="flex items-center">{badge}</div>) : null}
     </div>
   );
+};
--- a/src/renderer/components/+catalog/catalog-category-label.tsx
+++ b/src/renderer/components/+catalog/catalog-category-label.tsx
@@ -1,0 +1,21 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import React from "react";
+import type { CatalogCategory } from "../../../common/catalog/catalog-entity";
+
+interface CatalogCategoryLabelProps {
+  category: CatalogCategory;
+}
+
+/**
+ * Display label for Catalog Category for the Catalog menu
+ */
+export const CatalogCategoryLabel = ({ category }: CatalogCategoryLabelProps) =>
+  (
+    <div className="flex">
+      <div>{category.metadata.name}</div>
+      <div className="flex items-center">{category.getBadge()}</div>
+    </div>
+  );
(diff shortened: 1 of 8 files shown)


src/renderer/components/+catalog/catalog-category-label.tsx
```diff
@@ -0,0 +1,21 @@
...
+}
+
+/**
+ * Display label for Catalog Category for the Catalog menu
+ */
+export const CatalogCategoryLabel = ({ category }: CatalogCategoryLabelProps) =>
+  (
+    <div className="flex">
+      <div>{category.metadata.name}</div>
+      <div className="flex items-center">{category.getBadge()}</div>
```
jweak: I think it's better not to render container div if there's no badge.

We could also have props like below, so this UI component wouldn't need to know about category interface.
```react.jsx
{
  label: string | React.ReactNode,
  badge?: React.ReactNode
}
```

panuhorsmalahti: Good points, changed

jweak (dismissed):

src/renderer/components/+catalog/catalog-category-label.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+interface CatalogCategoryLabelProps {
+  label: string | React.ReactNode;
+  badge?: React.ReactNode;
+}
+
+/**
+ * Display label for Catalog Category for the Catalog menu
+ */
+export const CatalogCategoryLabel = ({ label, badge }: CatalogCategoryLabelProps) =>
```
Nokel81: I think that this should take only a `Category` as its input.

panuhorsmalahti: This was discussed here before it was changed to take in just label and badge: https://github.com/lensapp/lens/pull/5156#discussion_r839652665

Nokel81: Interesting, I agree with the first part of that comment, but not really the second since this component is specifically about doing all that logic, I don't see the benefit of making the caller provide it.

Nokel81: @jweak Thoughts?

jweak: I think a presentational component like Label should not do that logic.

Nokel81: Interesting, may I garner some insight then. 

 would have said that a presentation component was the best place to do it because it encapsulates the behaviour. The user of this component doesn't care if it displays or not.

jweak: I can see both implementations have merit. My point of would be to have a numbered components like `<Label>foo</Label>` and `<Header size={2}>catalog</Header>` and not create a new Label or Header component for each object we would like to display something.  

Anyway, if you feel strongly that it's better the other way, I'm fine with changing it back. It was just a suggestion.

Nokel81: I don't feel terribly strong about it, I just wanted to know you thought pattern.

I agree 100% that there should be some base component for doing the some rendering but I also think that there should be a separate component for doing the logic (and not having it be done in line).

jim-docker: I think the name `CatalogCategoryLabel` implies it is specific to a catalog category and it seems appropriate to pass just a `Category` to it. If it were called something generic like `LabelWithBadge` and could be used more generally then passing the `label` and `badge` are appropriate

jim-docker (dismissed):
tested from extension

src/renderer/components/+catalog/catalog-category-label.tsx
panuhorsmalahti: Waiting for 1 more approval.

jweak: I guess it's two against one for changing it back to taking the category so maybe change it and let's approve this and get it merged.

panuhorsmalahti: It's changed back now

jweak (approved):

Nokel81 (approved):

PR #5157, opened 2022-03-31
Compute Tray Icon on the fly instead of at build 

Nokel81:
After this PR is removed we can remove the icon artifacts from Lens-IDE as well as they are no longer needed.


src/main/tray/tray.ts
```diff
@@ -3,43 +3,86 @@
...
+interface ComputeTrayIconArgs {
+  shouldUseDarkColors: boolean;
+  size: number;
+  sourceSvg: string;
 }
 
-export function initTray(
+const trayIcons = new Map<boolean, NativeImage>();
+
+async function computeTrayIcon({ shouldUseDarkColors, size, sourceSvg }: ComputeTrayIconArgs): Promise<NativeImage> {
```
ixrock:
```suggestion
async function createTrayIcon({ shouldUseDarkColors, size, sourceSvg }: ComputeTrayIconArgs): Promise<NativeImage> {
```

src/main/tray/tray.ts
```diff
@@ -3,43 +3,86 @@
...
 import { checkForUpdates, isAutoUpdateEnabled } from "../app-updater";
 import type { WindowManager } from "../window-manager";
 import logger from "../logger";
-import { isDevelopment, isWindows, productName } from "../../common/vars";
+import { isWindows, productName } from "../../common/vars";
 import { exitApp } from "../exit-app";
-import { toJS } from "../../common/utils";
+import { getOrInsertWithAsync, toJS } from "../../common/utils";
 import type { TrayMenuRegistration } from "./tray-menu-registration";
+import parseDataURL from "data-urls";
```
ixrock: Do we really need one more dependency?
If you're not going to use `.body` from this package (which is not required actually in this PR) it would be just enough to use this for getting UTF-8 string:

```ts
common/utils/base64.decode(url.split("base64,")[1])
```

Nokel81: Good point will remove dependency.

Nokel81:
@ixrock PTALA?

package.json
```diff
@@ -202,12 +202,13 @@
     "@hapi/call": "^8.0.1",
     "@hapi/subtext": "^7.0.3",
     "@kubernetes/client-node": "^0.16.3",
+    "@ogre-tools/fp": "5.2.0",
```
ixrock: How this related to PR?

package.json
```diff
@@ -202,12 +202,13 @@
...
     "@ogre-tools/injectable": "5.2.0",
     "@ogre-tools/injectable-react": "5.2.0",
-    "@ogre-tools/fp": "5.2.0",
     "@sentry/electron": "^3.0.6",
     "@sentry/integrations": "^6.19.3",
     "@types/circular-dependency-plugin": "5.0.5",
+    "@types/data-urls": "^2.0.1",
```
ixrock: Leftovers

src/main/tray/tray.ts
```diff
@@ -3,43 +3,84 @@
...
+    const iconBuffer = await sharp(Buffer.from(svgRoot.outerHTML))
+      .resize({ width: size, height: size })
+      .png()
+      .toBuffer();
+
+    return nativeImage.createFromBuffer(iconBuffer);
+  });
+}
+
+function computeCurrentTrayIcon() {
```
ixrock: Similarly as above:

```suggestion
function createCurrentTrayIcon() {
```

src/main/tray/tray.ts
```diff
@@ -3,43 +3,84 @@
...
   trayMenuItems: IComputedValue<TrayMenuRegistration[]>,
   navigateToPreferences: () => void,
 ) {
-  const icon = getTrayIcon();
+  const icon = await computeCurrentTrayIcon();
 
   tray = new Tray(icon);
   tray.setToolTip(packageInfo.description);
   tray.setIgnoreDoubleClickEvents(true);
+  watchShouldUseDarkColors(tray);
```
ixrock: Shoudn't we dispose this in `disposers` below?

package.json
Nokel81: It isn't but I think @jansav added this value manually (instead of using `npm i --save` or `yarn add` and so added it in the wrong place. Both of those CLIs will sort these two objects whenever they touch them.

Nokel81:
@ixrock Done, thanks for the reviews

ixrock (dismissed):
LGTM

ixrock (dismissed):

Nokel81:
@ixrock I fixed the linux issue. I thought I had fixed that before, but I guess I didn't push it

ixrock (approved):

jakolehm (approved):
LGTM

Issue #4512, opened 2021-12-06
Optimise usage of getDetailsUrl() from params.ts


aleksfront:
While browsing `Pods` page as well as other part of the app a lot of `getDetailsUrl()` gets triggered. This leads to thousands `console.log()` messages in dev mode and about 7% of page rendering time.

Probably we should convert `getDetailsUrl()` to be used on click, and not generate links on render.

https://user-images.githubusercontent.com/9607060/144821533-723b5705-0f95-4a2d-b975-aa65f46fcadc.mov

![seven percent of render](https://user-images.githubusercontent.com/9607060/144821575-c0ea4b86-e9e8-49fc-b2e2-1d2a7cf31e59.png)

aleksfront:
I don't observe bunch of `console.log` messages anymore as well as any visible CPU impact over rendering table. Testing in `master`.

PR #5168, opened 2022-04-01
Do not render Tooltip and Menu elements until needed

aleksfront:
Part of #2243

Before this PR the DOM was flooded with html elements produces by each `Menu` and `Tooltip` components visible on the screen. Every table scroll caused bunch of these elements removed and created in a crazy speed.

Before:

https://user-images.githubusercontent.com/9607060/161263049-980de65e-f743-4dac-9094-31c12ea43fef.mov

After:

https://user-images.githubusercontent.com/9607060/161263081-fb70ee99-8425-41ec-bc99-dd2e53871b87.mov

### DIFF
--- a/src/renderer/components/animate/animate.tsx
+++ b/src/renderer/components/animate/animate.tsx
@@ -90,9 +90,13 @@ export class Animate extends React.Component<AnimateProps> {
       "--leave-duration": `${leaveDuration}ms`,
     } as React.CSSProperties;
 
+    if (!this.isVisible) {
+      return null;
+    }
+
     return React.cloneElement(contentElem, {
       className: cssNames("Animate", name, contentElem.props.className, this.statusClassName),
-      children: this.isVisible ? contentElem.props.children : null,
+      children: contentElem.props.children,
       style: {
         ...contentElem.props.style,
         ...durations,
--- a/src/renderer/components/animate/animate.tsx
+++ b/src/renderer/components/animate/animate.tsx
@@ -83,23 +83,23 @@ export class Animate extends React.Component<AnimateProps> {
   }
 
   render() {
-    const { name, enterDuration, leaveDuration } = this.props;
-    const contentElem = this.contentElem;
-    const durations = {
-      "--enter-duration": `${enterDuration}ms`,
-      "--leave-duration": `${leaveDuration}ms`,
-    } as React.CSSProperties;
-
     if (!this.isVisible) {
       return null;
     }
+
+    const { name, enterDuration, leaveDuration } = this.props;
+    const contentElem = this.contentElem;
+    const cssVarsForAnimation = {
+      "--enter-duration": `${enterDuration}ms`,
+      "--leave-duration": `${leaveDuration}ms`,
+    } as React.CSSProperties;
 
     return React.cloneElement(contentElem, {
       className: cssNames("Animate", name, contentElem.props.className, this.statusClassName),
       children: contentElem.props.children,
       style: {
         ...contentElem.props.style,
-        ...durations,
+        ...cssVarsForAnimation,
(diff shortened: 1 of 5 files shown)


src/renderer/components/menu/menu.tsx
```diff
@@ -218,7 +216,7 @@ export class Menu extends React.Component<MenuProps, State> {
     }
   }
 
-  onKeyDown(evt: KeyboardEvent) {
+  onKeyDown(evt: React.KeyboardEvent | KeyboardEvent) {
```
Nokel81, line 219: Doesn't this need to be `@autobind`?

src/renderer/components/animate/animate.tsx
```diff
@@ -90,9 +90,13 @@ export class Animate extends React.Component<AnimateProps> {
       "--leave-duration": `${leaveDuration}ms`,
     } as React.CSSProperties;
 
+    if (!this.isVisible) {
+      return null;
```
ixrock: Does animation works fine with that approach? Afaik this is the reason why keeping lot of html elements in DOM was done for css animation to kick-start..

Nokel81:
Should we also late bind the `<Link>`'s too so that we don't call `apiManager.lookupLink()` so many times unless needed?

Or probably in a different PR.

src/renderer/components/menu/menu.tsx
ixrock: @Nokel81 menu's constructor has `autoBind(this)`

src/renderer/components/animate/animate.tsx
ixrock: `<Animate>` also used in `notifications`, `drawer` & `dialog` components. Does all of them works as before with that change?

aleksfront:
> Should we also late bind the <Link>'s too so that we don't call apiManager.lookupLink() so many times unless needed?

Good catch. Probably this might be related https://github.com/lensapp/lens/issues/4512. Let's do this in separate PR.

src/renderer/components/animate/animate.tsx
aleksfront: Compared all this elements with same in `master`. They work without changes.

https://user-images.githubusercontent.com/9607060/161486987-2da4d0f4-f5b4-4a87-ab3f-fed16e65dd24.mov


https://user-images.githubusercontent.com/9607060/161487007-872b3505-9679-4510-8aa0-cd82acbf2118.mov


https://user-images.githubusercontent.com/9607060/161487013-129ddfe1-5dcc-49cc-bf8b-08c598f16efe.mov

src/renderer/components/tooltip/tooltip.tsx
```diff
@@ -54,13 +54,18 @@ export class Tooltip extends React.Component<TooltipProps> {
 
   @observable.ref elem: HTMLElement;
   @observable activePosition: TooltipPosition;
+  @observable isDomNodeRendered = false;
```
ixrock: Why do we really need this extra field? 
Can we live only `this.isVisible`? (since you have extra `get visible()` in readonly)

ixrock (dismissed):

Nokel81 (dismissed):

src/renderer/components/tooltip/tooltip.tsx
aleksfront: This is the way to preserve tooltip animation.

1. We show DOM node first which is invisible by styles.
2. We add `visible` class on it which starts the animation.

That's why we using 2 fields in here.

Is there a way to start the css-only animation right from the element appearing?

ixrock (approved):

Nokel81 (approved):

PR #5172, opened 2022-04-01
Upgrade to React@18

ixrock:
_Read more:_ https://reactjs.org/blog/2022/03/29/react-v18.html


package.json
```diff
@@ -202,9 +202,9 @@
     "@hapi/call": "^8.0.1",
     "@hapi/subtext": "^7.0.3",
     "@kubernetes/client-node": "^0.16.3",
+    "@ogre-tools/fp": "5.2.0",
```
panuhorsmalahti, line 205: Offtopic, but I wonder why this is moving around all the time in PRs

panuhorsmalahti:
Does this mean all extensions must upgrade to React 18 for 5.5.0?

package.json
ixrock: It happens all the time because when it first came into `master` someone added it manually to `package.json` instead of with `yarn add *` or `npm install -D` (this commands sorts packages by name before writing to `package.json`).
So whenever you add or update _any_ package with that commands this one becomes sorted.

ixrock:
> Does this mean all extensions must upgrade to React 18 for 5.5.0?

No. Most of extensions should work as before afaik.
Also app shares same instance of `react` and `react-dom` to extensions so there is nothing to update for them..
Extensions might follow update recommendations if will see some warnings from `react` packages, but nothing critical expected.

package.json
Nokel81: FYI @jansav

chenhunghan (approved):
Changes looks good. Wonder if we want to enable React strict mode (or are we already?)

ixrock:
> Changes looks good. Wonder if we want to enable React strict mode (or are we already?)

Probably not enabled yet. But we can give it a try later on!

Nokel81 (approved):

src/renderer/components/menu/menu.tsx
```diff
@@ -144,7 +144,7 @@ export class Menu extends React.Component<MenuProps, State> {
     }
   }
 
-  refreshPosition = () => {
+  refreshPosition = () => requestAnimationFrame(() => {
```
aleksfront, line 147: I guess it may be marked as `startTransition` (https://reactjs.org/docs/hooks-reference.html#usetransition) if component would use functional form.

aleksfront (approved):

package.json
ixrock: Since now it's merged to `master` it should disappears from other PRs diffs..

PR #5174, opened 2022-04-01
Fix crash when switching OS theme type

### LINKED ISSUE #5171: Crash when switching OS theme
I think this is a regression on master since it doesn't happen on `5.4.4` release.

https://user-images.githubusercontent.com/8225332/161295350-b3fe048d-e60c-47b8-900c-be83f0cd827e.mov

Nokel81:
- Changed the representation of theme preferences to no longer use
  sentinal values and instead use flags

- This new format is forward looking to when support for user provided
  themes is added, which is now quite easy.

- Added unit tests for the migration code

- Made Theme's more type safe my having the lens themes be .ts files

- Moved ThemeStore into renderer/themes as it is a more fitting place

fixes #5171


aleksfront (commented):
This PR contains more than what it states in the subject. I suggest spread it into 2 if possible: one for the fix, one for the type improvements.

Regarding to second part the improvement should follow requirements (what I personally find useful):
1. Remove intermedia `theme-vars.css` file along with its generators
2. Add ability to add custom theme as an extension
3. Add ability to see color values right into the ide if possible

PR #5175, opened 2022-04-01
Fix Pod Container sorting not correlating to visuals

### LINKED ISSUE #5165: Pod list Containers sort doesn't work
https://user-images.githubusercontent.com/9607060/161237511-f8088df4-1152-46a5-b18b-c12287bf3ba0.mov

Nokel81:
- Add some specific tuple.zip types

- Remove the `string` part of `TableOrderBy` to simplify implementation
  and because it removes any sort of intellisense

fixes #5165

### DIFF
--- a/src/renderer/components/+workloads-pods/pods.tsx
+++ b/src/renderer/components/+workloads-pods/pods.tsx
@@ -113,7 +113,7 @@ class NonInjectedPods extends React.Component<Dependencies> {
           sortingCallbacks={{
             [columnId.name]: pod => getConvertedParts(pod.getName()),
             [columnId.namespace]: pod => pod.getNs(),
-            [columnId.containers]: pod => pod.getContainers().length,
+            [columnId.containers]: pod => pod.getContainerStatuses().length,
             [columnId.restarts]: pod => pod.getRestartsCount(),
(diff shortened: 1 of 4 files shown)


src/renderer/components/+workloads-pods/pods.tsx
```diff
@@ -82,7 +82,7 @@ export class Pods extends React.Component {
           sortingCallbacks={{
             [columnId.name]: pod => getConvertedParts(pod.getName()),
             [columnId.namespace]: pod => pod.getNs(),
-            [columnId.containers]: pod => pod.getContainers().length,
+            [columnId.containers]: pod => pod.getContainerStatuses().length,
```
aleksfront, line 116: I find it more intuitive to keep container amount as the sorting criteria. Sorting by the statuses is not clear enough.

aleksfront (changes requested):

src/renderer/components/+workloads-pods/pods.tsx
Nokel81: Well the bug was that you found the previous sorting mechanism unintuitive. With this the sorting matches the visuals.

aleksfront (approved):

ixrock (approved):

PR #5182, opened 2022-04-04
Fix broken hotbar switcher

### LINKED ISSUE #5167: Broken hotbar switcher
```
Cannot read properties of undefined (reading 'activeHotbarIndex')
```
error happens when trying to change Hotbar from the Hotbar switcher.

https://user-images.githubusercontent.com/9607060/161257710-098ff8ab-a564-45f0-940c-a949551cda23.mov

aleksfront:
Fixes #5167 

https://user-images.githubusercontent.com/9607060/161502976-f6298b00-95cc-4ffc-aabd-f4cab3ff0b24.mov

### DIFF
--- a/src/common/hotbar-store.ts
+++ b/src/common/hotbar-store.ts
@@ -285,7 +285,7 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
     }
   }
 
-  switchToPrevious = () => {
+  switchToPrevious() {
     let index = this.activeHotbarIndex - 1;
 
     if (index < 0) {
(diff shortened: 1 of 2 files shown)


DmitriyNoa (dismissed):

aleksfront:
@Nokel81 Set 5.5.0 milestone since it's fixed regression from #5084 which won't go into 5.4.5

Nokel81 (dismissed):

Nokel81 (commented):
Actually, is this really the best solution? Wouldn't it be better to just not treat methods as free functions at the use?

aleksfront:
> Actually, is this really the best solution? Wouldn't it be better to just not treat methods as free functions at the use?

Can you elaborate this a bit?

Nokel81:
> Can you elaborate this a bit?

Sure, I meant to ask "would it be better to switch the uses of `hotbarStore.switchToPrevious` in `src/renderer/components/hotbar/hotbar-selector.tsx` to `() => hotbarStore.switchToPrevious()` instead of changing `switchToPrevious` from a method to a field function?"

aleksfront:
@Nokel81 Made suggested changes

DmitriyNoa (approved):

Nokel81 (approved):

PR #5184, opened 2022-04-04
Change notification when extension is not found

### LINKED ISSUE #5141: Misleading error if extension can't be downloaded
**Describe the bug**
Misleading error is displayed for the user if extension can't be downloaded in the Extension page.

**To Reproduce**
1. Open Extension page
2. Add extension url that will return 404 error

**Expected behavior**
Error notification should be displayed, saying something like "Extension can't be downloaded"

**Actual behavior**
Error notification tells users that package.json is missing.

- Lens Version: 5.4.4
- OS: Mac

Nokel81:
- A notification will be displayed if the URL provided results in a
  status of 404

fixes #5141 
Screen Shot 2022-04-04 at 11 06 14 AM


src/renderer/components/+extensions/attempt-install-by-info/attempt-install-by-info.tsx
```diff
@@ -23,20 +24,34 @@ interface Dependencies {
...
   const baseUrl = await getBaseRegistryUrl();
   const registryUrl = new URLParse(baseUrl).set("pathname", name).toString();
-  let json: any;
+  let json;
 
   try {
-    json = await downloadJson({ url: registryUrl }).promise;
+    const request = await fetch(registryUrl);
+
+    if (request.status === 404) {
```
panuhorsmalahti: It would be even better if we check that if the status code is not between 200 and 299

Nokel81:
@panuhorsmalahti Done
Screen Shot 2022-04-05 at 11 04 09 AM

panuhorsmalahti (dismissed):

package.json
```diff
@@ -219,6 +219,7 @@
     "electron-devtools-installer": "^3.2.0",
     "electron-updater": "^4.6.5",
     "electron-window-state": "^5.0.3",
+    "fetch-mock": "^9.11.0",
```
jansav: Maybe good for time being, but in the future we should wrap fetch with our own functions, which do one thing and limit the amount of combinations. That will make `fetch-mock` redundant.

Mocking of those functions is enough in behavioural unit tests, the functions themself should have unit tests that makes sure that they work against fetch.

E.g. (details irrelevant)

```typescript
const callJsonApi = async ({ method, url, body }) => {
  const httpResponse = await fetch(url, {
    method,
    body: JSON.stringify(body),
    headers: { "Content-Type": "application/json", Accept: "application/json" },
  });

  return {
    callWasSuccessful: true, // ...whatever it means to get status of the call
    response: await httpResponse.json(),
  };
};
```

This will give us:
1. Ability to keep interfaces simple and clean
2. Ability to change our mind in future in single place
3. Ability to extend in single place (e.g. add error monitoring for all calls)
4. Ability to stop using try-catch in operational code being wrong tool for the job
5. Ability to override our own implementation in behavioural unit tests.

src/renderer/components/+extensions/__tests__/extensions.test.tsx
```diff
@@ -20,36 +20,27 @@ import { DiRender, renderFor } from "../../test-utils/renderFor";
...
-import getConfigurationFileModelInjectable
-  from "../../../../common/get-configuration-file-model/get-configuration-file-model.injectable";
-import appVersionInjectable
-  from "../../../../common/get-configuration-file-model/app-version/app-version.injectable";
+import getConfigurationFileModelInjectable from "../../../../common/get-configuration-file-model/get-configuration-file-model.injectable";
+import appVersionInjectable from "../../../../common/get-configuration-file-model/app-version/app-version.injectable";
+import fetchInjectable from "../../../../common/fetch/fetch.injectable";
+import fetchMock, { type FetchMockSandbox } from "fetch-mock";
+
+console.log("This is here as a reminder that mockFs breaks things and needs to be removed");
```
jansav: I don't think we should bloat the unit tests with this. Makes tests harder to read.

src/renderer/components/+extensions/attempt-install-by-info/attempt-install-by-info.tsx
```diff
@@ -109,7 +124,13 @@ export const attemptInstallByInfo = ({ attemptInstall, getBaseRegistryUrl, exten
 
   const url = json.versions[version].dist.tarball;
   const fileName = path.basename(url);
-  const { promise: dataP } = downloadFile({ url, timeout: 10 * 60 * 1000 });
+  const request = await fetch(url, { timeout: 10 * 60 * 1000 });
```
jansav: `downloadFile` is closer to abstraction level that we should have here. Problem here of course is that everything else is in low level of abstraction making code very hard to read. Guiding factor should be that function should only contain code that is in same level of abstraction thus easier to read.

src/renderer/components/+extensions/__tests__/extensions.test.tsx
Iku-turso: The "availability" of this reminder is such that it should be fixed right away :)

src/renderer/components/+extensions/__tests__/extensions.test.tsx
```diff
@@ -60,6 +51,9 @@ describe("Extensions", () => {
     di.permitSideEffects(getConfigurationFileModelInjectable);
     di.permitSideEffects(appVersionInjectable);
 
+    mockFetch = fetchMock.sandbox();
```
Iku-turso: Let's consider if we need a library to mock a low-level and technical abstraction such as `fetch`.

Unit tests would be just as reliable, yet simpler to implement, if instead of `fetch` we had a higher level API, eg. `download` as required here.

Stubbing complex (yet elsewhere unit tested) API's can be a good thing when said complexity is irrelevant in the scope of the unit test.

src/renderer/components/+extensions/__tests__/extensions.test.tsx
```diff
@@ -103,39 +97,47 @@ describe("Extensions", () => {
...
-    await waitFor(() => {
+    await waitFor(async () => {
       expect(extensionDiscovery.uninstallExtension).toHaveBeenCalled();
       fireEvent.click(menuTrigger);
-      expect(res.getByText("Disable")).toHaveAttribute("aria-disabled", "true");
-      expect(res.getByText("Uninstall")).toHaveAttribute("aria-disabled", "true");
+      expect(await res.findByText("Disable")).toHaveAttribute("aria-disabled", "true");
+      expect(await res.findByText("Uninstall")).toHaveAttribute("aria-disabled", "true");
     }, {
       timeout: 30000,
```
Iku-turso: Unacceptable timeout in unit test.

src/renderer/components/+extensions/__tests__/extensions.test.tsx
```diff
@@ -103,39 +97,47 @@ describe("Extensions", () => {
...
-    fireEvent.click(res.getByText("Install"));
-    expect(res.getByText("Install").closest("button")).toBeDisabled();
+    let resolveBuffer: (buf: Buffer) => void;
+
+    mockFetch.getOnce(url, new Promise<Buffer>(resolve => {
+      resolveBuffer = resolve;
+    }));
+
+    fireEvent.click(await res.findByText("Install"));
+    expect((await res.findByText("Install")).closest("button")).toBeDisabled();
```
Iku-turso: Brittle selectors such as this are likely to cause test failure from irrelevant reasons, such as volatile UI code being volatile. Use eg. a dedicated data-attribute for a more reliable access to elements of UI?

I acknowledge there's an aggravated discussion on this topic somewhere else.

src/renderer/components/+extensions/__tests__/extensions.test.tsx
```diff
@@ -20,36 +20,27 @@ import { DiRender, renderFor } from "../../test-utils/renderFor";
...
+import getConfigurationFileModelInjectable from "../../../../common/get-configuration-file-model/get-configuration-file-model.injectable";
+import appVersionInjectable from "../../../../common/get-configuration-file-model/app-version/app-version.injectable";
+import fetchInjectable from "../../../../common/fetch/fetch.injectable";
+import fetchMock, { type FetchMockSandbox } from "fetch-mock";
+
+console.log("This is here as a reminder that mockFs breaks things and needs to be removed");
 
 mockWindow();
 
 jest.setTimeout(30000);
```
Iku-turso: Unacceptably long timeout in unit test

src/renderer/components/+extensions/__tests__/extensions.test.tsx
```diff
@@ -20,36 +20,27 @@ import { DiRender, renderFor } from "../../test-utils/renderFor";
...
-    cancel: () => {},
-  })),
-  downloadJson: jest.fn(({ url }) => ({
-    promise: Promise.resolve({}),
-    url,
-    cancel: () => { },
-  })),
+jest.mock("../../../../common/utils/tar", () => ({
+  listTarEntries: () => [] as string[],
 }));
```
Iku-turso: New code should `di.override` instead of monkey-patch. I suspect there's a reason for this temptation. Let's take that discussion offline?

src/renderer/components/+extensions/__tests__/extensions.test.tsx
```diff
@@ -103,39 +97,47 @@ describe("Extensions", () => {
 
     fireEvent.click(menuTrigger);
 
-    expect(res.getByText("Disable")).toHaveAttribute("aria-disabled", "false");
-    expect(res.getByText("Uninstall")).toHaveAttribute("aria-disabled", "false");
+    expect(await res.findByText("Disable")).toHaveAttribute("aria-disabled", "false");
```
Iku-turso: A unit test would benefit of only having one expectation (assertion) to eg. make the nake of test alone sufficient in pointing out why a test fails.

src/renderer/components/+extensions/__tests__/extensions.test.tsx
```diff
@@ -103,39 +97,47 @@ describe("Extensions", () => {
...
-    expect(res.getByText("Install").closest("button")).toBeDisabled();
+    let resolveBuffer: (buf: Buffer) => void;
+
+    mockFetch.getOnce(url, new Promise<Buffer>(resolve => {
+      resolveBuffer = resolve;
+    }));
+
+    fireEvent.click(await res.findByText("Install"));
+    expect((await res.findByText("Install")).closest("button")).toBeDisabled();
+    resolveBuffer(Buffer.from([]));
```
Iku-turso: The effect of this is async, and makes no difference as sync in this test?

Iku-turso (dismissed):
Requires some conversations offline :)

src/renderer/components/+extensions/__tests__/extensions.test.tsx
```diff
@@ -103,39 +97,47 @@ describe("Extensions", () => {
...
-    expect(res.getByText("Uninstall")).toHaveAttribute("aria-disabled", "false");
+    expect(await res.findByText("Disable")).toHaveAttribute("aria-disabled", "false");
+    expect(await res.findByText("Uninstall")).toHaveAttribute("aria-disabled", "false");
 
-    fireEvent.click(res.getByText("Uninstall"));
+    fireEvent.click(await res.findByText("Uninstall"));
 
     // Approve confirm dialog
-    fireEvent.click(res.getByText("Yes"));
+    fireEvent.click(await res.findByText("Yes"));
```
Iku-turso: Brittle, ambiguous selector.

src/renderer/components/+extensions/__tests__/extensions.test.tsx
Nokel81: I agree, just not in this PR

Nokel81: This is required to use console.log within tests.

src/renderer/components/+extensions/__tests__/extensions.test.tsx
Nokel81: Not wanting to change everything in this PR

src/renderer/components/+extensions/__tests__/extensions.test.tsx
Nokel81: Maybe, but that doesn't mean it must be fixed in this PR

src/renderer/components/+extensions/__tests__/extensions.test.tsx
Nokel81: If you like I can revert all of these `get*` -> `find*` changes so that this diff is smaller.

src/renderer/components/+extensions/__tests__/extensions.test.tsx
Nokel81: True, but out of scope

src/renderer/components/+extensions/attempt-install-by-info/attempt-install-by-info.tsx
Nokel81: okay I will make it so that fetch is hierarchy of injectables instead.

yarn.lock
```diff
@@ -7,6 +7,13 @@
   resolved "https://registry.yarnpkg.com/7zip-bin/-/7zip-bin-5.1.1.tgz#9274ec7460652f9c632c59addf24efb1684ef876"
   integrity sha512-sAP4LldeWNz0lNzmTird3uWfFDWWTeg6V/MsmyyLR9X1idwKBWIgt/ZvinqQldJm3LecKEs1emkbquO6PCiLVQ==
 
+"@ampproject/remapping@^2.1.0":
```
jim-docker: just checking, is this big set of changes to yarn.lock expected?

jim-docker (commented):
My usual concern, this seems a big change to fix a bug for a patch release.

yarn.lock
Nokel81: It is because currently `fetch-mock` is being brought in. But as @jansav mentioned that is less than ideal and it would be better if the injectables were more focused. So this should go away.

package.json
Nokel81: Okay will remove

src/renderer/components/+extensions/__tests__/extensions.test.tsx
Nokel81: I don't understand what you mean.

src/renderer/components/+extensions/__tests__/extensions.test.tsx
Nokel81: Removed this mock as it is not needed anymore

src/renderer/components/+extensions/__tests__/extensions.test.tsx
Nokel81: Removed anyway

src/renderer/components/+extensions/__tests__/extensions.test.tsx
Nokel81: Done

src/renderer/components/+extensions/attempt-install-by-info/attempt-install-by-info.tsx
Nokel81: Done

src/common/fetch/download-binary.injectable.ts
```diff
@@ -0,0 +1,56 @@
...
+    const fetch = di.inject(fetchInjectable);
+
+    return async (url, opts) => {
+      let result: Response;
+
+      try {
+        // TODO: upgrade node-fetch once we switch to ESM
+        result = await fetch(url, opts as RequestInit);
+      } catch (error) {
+        return {
```
panuhorsmalahti, line 28: Why is the promise not rejected instead of returning an object with this metadata?

src/common/fetch/download-binary.injectable.ts
```diff
@@ -0,0 +1,56 @@
...
+          callWasSuccessful: false,
+          error: String(error),
```
panuhorsmalahti, line 30: `errorMessage` would make it more clear that we're not returning an Error object.

src/common/fetch/timeout-controller.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+/**
+ * Creates an AbortController with an associated timeout
+ * @param timeout The number of miliseconds before this controller will auto abort
```
panuhorsmalahti: [nit] typo: miliseconds

src/renderer/components/+extensions/attempt-install-by-info.injectable.tsx
```diff
@@ -31,84 +49,131 @@ const attemptInstallByInfoInjectable = getInjectable({
...
-      let json: any;
-      let finalVersion = version;
+      let json: NpmRegistryPackageDescriptor;
 
       try {
-        json = await downloadJson({ url: registryUrl }).promise;
+        const result = await downloadJson(registryUrl);
+
+        if (!result.callWasSuccessful) {
+          showErrorNotification(`Failed to get registry information for that extension: ${result.error}`);
```
panuhorsmalahti: "that" sounds superfluous

src/common/fetch/download-binary.injectable.ts
Nokel81: To answer both this and your other review comment, this style of programming is called "Errors as data" instead of "Errors as exceptions". That means we have started using the more functional programming approach.

This allows us to better encode the error states into the program, to explicitly change the errors we expect into just data, and remove the need to use try/catch in business logic.

Furthermore, we have moved to using a single type definition `AsyncResult` 

```ts
export type AsyncResult<Response, Error = string> =
  | (
    Response extends void
    ? { callWasSuccessful: true; response?: undefined }
    : { callWasSuccessful: true; response: Response }
  )
  | { callWasSuccessful: false; error: Error };
```

So that the error handling code is consistent throughout.

panuhorsmalahti (dismissed):

panuhorsmalahti (approved):

PR #5185, opened 2022-04-04
Add copy button on 'About Lens' dialog

### LINKED ISSUE #5177: Information in "About Lens" is not selectable for copy to clipboard
**Describe the bug**
Information in "About Lens" popup window is not selectable for copy to clipboard. Please make it selectable by mouse or at least add button "Copy to clipboard".

This info is needed at least for every new issue, to fill the "Environment" section, and posting screenshot or retyping the version (with build id) is too stressful :)
image

- Lens Version: 5.4.4-latest-many-following-numbers
- OS: Ubuntu 21.04 Linux
- Installation method: snap

Nokel81:

fixes #5177 
Screen Shot 2022-04-04 at 1 47 19 PM


src/main/menu/application-menu-items.injectable.ts
```diff
@@ -69,7 +69,8 @@ const applicationMenuItemsInjectable = getInjectable({
             label: `About ${productName}`,
             id: "about",
             click(menuItem: MenuItem, browserWindow: BrowserWindow) {
-              showAbout(browserWindow);
+              showAbout(browserWindow)
```
jansav: Is there a reason why it's now promise which is not `await`ed? It's not good to left promises in the air. Those invite unwelcome complexity and should be avoided.

src/main/menu/menu.ts
```diff
@@ -18,22 +18,28 @@ export function initMenu(
...
+    buttons: ["Close", "Copy"],
     message: productName,
-    detail: appInfo.join("\r\n"),
+    detail: appInfo,
+    cancelId: 0,
+    defaultId: 0,
   });
+
+  if (result.response === 0) {
+    clipboard.writeText(appInfo);
```
jansav: Side-effect which should be a dependency.

src/main/menu/application-menu-items.injectable.ts
```diff
@@ -69,7 +69,8 @@ const applicationMenuItemsInjectable = getInjectable({
...
+                .catch(error => logger.error(`[MENU]: Failed to show Lens About view`, { error }));
```
Iku-turso: This logic to catch, log and suppress errors (!) is duplicated in many places.

Iku-turso: It would be better to only suppress specific errors, instead of all of them. The specific error may not be fatal, but actual fatal errors should be rethrown.

src/main/menu/menu.ts
Iku-turso: Explicit side effect.

src/main/menu/menu.ts
```diff
@@ -18,22 +18,28 @@ export function initMenu(
...
-    buttons: ["Close"],
+    buttons: ["Close", "Copy"],
     message: productName,
-    detail: appInfo.join("\r\n"),
+    detail: appInfo,
+    cancelId: 0,
+    defaultId: 0,
   });
+
+  if (result.response === 0) {
```
Iku-turso: Nit: Code could make effort to explain what `response === 0` means.

src/main/menu/application-menu-items.injectable.ts
Iku-turso: Orphan promise could be awaited instead.

Iku-turso (changes requested):
Some non-nit comments exist.

PR #5193, opened 2022-04-05
Fix cluster connect statuses being intertwined

### LINKED ISSUE #5173: Cluster connecting page shows wrong errors
When trying to connect to 2 clusters simultaneously, where both clusters are failing to connect, the connecting page sometimes shows the errors for the other cluster (docker-desktop page, mydevcluster errors):

Screen Shot 2022-04-01 at 12 46 16 PM

not quite sure how to reproduce. May require a 3rd cluster: on hotbar click cluster A, click cluster B, click cluster C, click cluster B (where A and B fail to connect)

Observed on master

Nokel81:
- Broadcast on a single channel and add `ClusterId` as data argument

- Make everything else computed based on the current props.cluster value

fixes #5173


Iku-turso:
Nit: a bit better PR description would be appreciated :) The underlying problem, and how it is solved is not that apparent.

src/main/__test__/kube-auth-proxy.test.ts
```diff
@@ -223,34 +223,34 @@ describe("kube auth proxy tests", () => {
       await proxy.run();
       listeners.emit("error", { message: "foobarbat" });
 
-      expect(mockBroadcastIpc).toBeCalledWith("cluster:foobar:connection-update", { message: "foobarbat", isError: true });
+      expect(mockBroadcastIpc).toBeCalledWith("cluster:connection-update", "foobar", { message: "foobarbat", isError: true });
```
Iku-turso: Nit: we could prefer something more informative than "foobar" for naming in testing in general. 

Eg. here instead of "foobar", we could have "some-cluster-id". This can make a big difference when the scenarios grow less trivial.

src/renderer/components/cluster-manager/cluster-status.tsx
```diff
@@ -32,100 +32,92 @@ interface Dependencies {
...
-    return this.authOutput.some(({ isError }) => isError);
+  private get clusterId() {
+    return this.props.cluster.id;
   }
 
   componentDidMount() {
     disposeOnUnmount(this, [
-      ipcRendererOn(`cluster:${this.cluster.id}:connection-update`, (evt, res: KubeAuthUpdate) => {
-        this.authOutput.push(res);
+      ipcRendererOn(`cluster:connection-update`, (evt, clusterId: ClusterId, update: KubeAuthUpdate) => {
```
Iku-turso: Explicit side-effect in touched (if not seen) code should be consolidated to a dependency

src/renderer/components/cluster-manager/cluster-status.tsx
```diff
@@ -32,100 +32,92 @@ interface Dependencies {
...
+        getOrInsert(this.authOutputs, clusterId, []).push(update);
```
Iku-turso: Nit: Hmm. Using something called `getOrInsert` here is a bit convoluted, as nothing is never "got". Could here be a function called eg. `setAuthSuccess` to document proper intent in this level of abstraction?

src/renderer/components/cluster-manager/cluster-status.tsx
```diff
@@ -32,100 +32,92 @@ interface Dependencies {
...
+  private reconnect = async () => {
+    this.authOutputs.delete(this.clusterId);
+    this.reconnecting.add(this.clusterId);
 
     try {
-      await requestClusterActivation(this.cluster.id, true);
+      await requestClusterActivation(this.clusterId, true);
     } catch (error) {
-      this.authOutput.push({
+      getOrInsert(this.authOutputs, this.clusterId, []).push({
```
Iku-turso: Nit: Hmm. Using something called `getOrInsert` here is a bit convoluted, as nothing is never "got". Could here be a function called eg. `setAuthFailure` to document proper intent in this level of abstraction?

src/renderer/components/cluster-manager/cluster-status.tsx
```diff
@@ -32,100 +32,92 @@ interface Dependencies {
...
     return (
       <pre>
         {
-          this.authOutput.map(({ message, isError }, index) => (
-            <p key={index} className={cssNames({ error: isError })}>
-              {message.trim()}
-            </p>
-          ))
+          this.authOutput
+            .get()
```
Iku-turso, line 79: Nit/idea/opinion: keeping act of MobX de-referencing and the `observer` at same level of abstraction can lead to simpler code.

Eg. here `.get()` is pretty far from `observer` causing uncertainty in reactive context and its granularity.

Iku-turso (dismissed):

src/renderer/components/cluster-manager/cluster-status.tsx
Nokel81: 1. Something is "got", it is the list of updates for `clusterId`.
2. This does not "set" anything, it "appends" which is obvious from the `.push` call.

src/renderer/components/cluster-manager/cluster-status.tsx
Nokel81: How is is far? It is in the same component...

Iku-turso: It's a big component :) There may be some benefits to be had if observer and whatever dereferences would be right next to each other. Eg. Intent of code could be more clear, like "ah, so this mutation causes this specific area in UI to update", and not "ah, so this entire component re-renders for a large amount of reasons that are unknown to me".

We've seen this as worthwhile somewhere else, could be here too :)

src/renderer/components/cluster-manager/cluster-status.tsx
Iku-turso: Here I can see how the code works, but more effort in code expressiveness could save lives somewhere else :) The code could convey intent over low-level technicalities.

src/renderer/components/cluster-manager/cluster-status.tsx
jim-docker: Like:
```
  appendAuthUpdate = (update: KubeAuthUpdate) => {
    getOrInsert(this.authOutputs, this.clusterId, []).push(update);
};
```
?

jim-docker (dismissed):
can't reproduce the bug anymore 👍

jim-docker:
I think this problem pre-exists this PR (I could reproduce it once on 5.4.5) but it is much easier to reproduce here. I can make a separate issue if you prefer.

Tries to connect after selecting disconnect (instead of going to the catalog, like in 5.4.5):

https://user-images.githubusercontent.com/40840436/162276868-a50eda25-d234-438d-a5ff-c123a270b2ed.mov

Nokel81:
>  I can make a separate issue if you prefer.

No I think it is reasonable to fix that here.

Nokel81:
@jim-docker Fixed, it was nice and simple too.

jim-docker:
> @jim-docker Fixed, it was nice and simple too.

nice, but there is a small residual problem:

https://user-images.githubusercontent.com/40840436/162312613-aa222487-d1ad-4228-8fef-61af98e38e9c.mov

It goes to the catalog when trying to connect a second time

src/renderer/components/cluster-manager/cluster-status.tsx
Iku-turso: May I ask you to please stop resolving conversation threads which are not resolved :)

src/renderer/components/cluster-manager/cluster-status.tsx
Iku-turso: More like `setAuthFailure(error);`, because `appendAuthUpdate` is just something technical without clear intent or motivation.

src/renderer/components/cluster-manager/cluster-status.tsx
Nokel81: How is it not resolved? I have moved it into a dependency.

Nokel81:
@jim-docker Fixed

src/renderer/routes/route-path-parameters.injectable.ts
```diff
@@ -2,32 +2,30 @@
...
-      return match ? match.params : {};
-    });
+    return <Param>(route: Route<Param>): IComputedValue<Param | null> => (
+      getOrInsertWith(pathParametersCache, route, () => computed(() => (
+        matchPath<Param>(currentPath.get(), {
+          path: route.path,
+          exact: true,
+        })?.params ?? null
+      ))) as IComputedValue<Param | null>
+    );
```
jansav, line 27: What is the motivation here? This caching is something that `keyedSingleton` does already. Typing?

src/renderer/utils/channel/channel-listeners/enlist-message-channel-listener.injectable.ts
```diff
@@ -24,9 +24,11 @@ const enlistMessageChannelListenerInjectable = getInjectable({
         );
       };
 
+      console.debug(`[IPC]: listening on ${channel.id}`);
```
jansav: Use `logger` if you really want to add these. I don't see reason.

src/renderer/components/cluster-manager/cluster-view.tsx
```diff
@@ -68,19 +94,43 @@ class NonInjectedClusterView extends React.Component<Dependencies> {
...
+              ));
+            },
+          ),
+        );
+
+        this.navigateToClusterDisposer = navigateToClusterDisposer;
+
+        disposeOnUnmount(this, [
+          navigateToClusterDisposer,
+        ]);
```
jansav: What is the thing that we are trying to do here? It's just so complex that it requires explanation. There must be simpler solution which doesn't require `reaction`, `when` or even `bindEvents` at all.

src/renderer/routes/route-path-parameters.injectable.ts
Nokel81: Yes typing.

src/renderer/utils/channel/channel-listeners/enlist-message-channel-listener.injectable.ts
Nokel81: Will remove, I think this was me trying to debug something.

src/renderer/components/cluster-manager/cluster-view.tsx
Nokel81: 1. `bindEvents` is just a helper that is called in `componentDidMount`.
2. Yes I would very much like to see something simpler.
3. It does several things
    1. It requests main to start the connection and activation process for a given cluster (when the cluster being view changes)
    2. Sets up a handler that navigates to the catalogue if the current cluster becomes disconnected

Nokel81:
Closing as I can split the work done here into smaller PRs.

PR #5201, opened 2022-04-06
Fix returning to specific catalog tab and removing a hotbar item

### LINKED ISSUE #5192: Clicking Catalog Hotbar item goes to Browse instead of selected Catalog category
**Describe the bug**
Clicking Catalog Hotbar item goes to Browse instead of selected Catalog category

**To Reproduce**
1. Click on `Clusters` category
2. Connect to a cluster
3. Click *Catalog* hotbar item
4. See that it goes to `Browse`

**To Reproduce part 2**
1. Click on `Clusters` category
2. Click *Catalog* hotbar item
3. See that it goes to `Browse`

**Expected behavior**
It should navigate to the `Clusters` category

**Screenshots**

https://user-images.githubusercontent.com/774344/161769716-16de74a7-00f6-43b7-8550-d94ea56c285a.mov

- Lens Version master 85f2ce7504602b5270fff068ed8772534e62e853:

jansav:
This a hotfix which imitates the previous stateful solution.

Proper fix would mean redesign of hotbar items which operate at wrong level of abstraction and inherit a lot instead of composition.

Fixes #5192 

Also fixes removing of hotbar items which caused an error because of unexpected `this`.

### DIFF
--- a/src/renderer/components/+catalog/catalog.tsx
+++ b/src/renderer/components/+catalog/catalog.tsx
@@ -42,7 +42,7 @@ import hotbarStoreInjectable from "../../../common/hotbar-store.injectable";
 import type { HotbarStore } from "../../../common/hotbar-store";
 
 interface Dependencies {
-  catalogPreviousActiveTabStorage: { set: (value: string ) => void };
+  catalogPreviousActiveTabStorage: { set: (value: string ) => void; get: () => string };
   catalogEntityStore: CatalogEntityStore;
   getCategoryColumns: (params: GetCategoryColumnsParams) => CategoryColumns;
   customCategoryViews: IComputedValue<Map<string, Map<string, RegisteredCustomCategoryViewDecl>>>;
@@ -78,6 +78,12 @@ class NonInjectedCatalog extends React.Component<Dependencies> {
       return `${dereferencedGroup}/${dereferencedKind}`;
     }
 
+    const previousTab = this.props.catalogPreviousActiveTabStorage.get();
+
+    if (previousTab) {
+      return previousTab;
+    }
+
     return browseCatalogTab;
   }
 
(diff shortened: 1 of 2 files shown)


Iku-turso (approved):

Nokel81:
There is already a PR that fixes the hotbar crash so please back out the fix from this PR.

jansav:
> There is already a PR that fixes the hotbar crash so please back out the fix from this PR.

#5167 is not related.

Iku-turso (approved):

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -42,7 +42,7 @@ import hotbarStoreInjectable from "../../../common/hotbar-store.injectable";
 import type { HotbarStore } from "../../../common/hotbar-store";
 
 interface Dependencies {
-  catalogPreviousActiveTabStorage: { set: (value: string ) => void };
+  catalogPreviousActiveTabStorage: { set: (value: string ) => void; get: () => string };
```
Nokel81, line 45: Shouldn't this be `IObservableBox<string>`?

src/renderer/components/+catalog/catalog.tsx
```diff
@@ -78,6 +78,12 @@ class NonInjectedCatalog extends React.Component<Dependencies> {
       return `${dereferencedGroup}/${dereferencedKind}`;
     }
 
+    const previousTab = this.props.catalogPreviousActiveTabStorage.get();
+
+    if (previousTab) {
+      return previousTab;
+    }
+
     return browseCatalogTab;
```
Nokel81, line 87: nit:

```suggestion
    return this.props.catalogPreviousActiveTabStorage.get() || browseCatalogTab;
```

src/renderer/components/hotbar/hotbar-menu.tsx
```diff
@@ -73,17 +73,17 @@ class NonInjectedHotbarMenu extends React.Component<Dependencies & HotbarMenuPro
     this.props.hotbarStore.restackItems(from, to);
   }
 
-  removeItem(uid: string) {
+  removeItem = (uid: string) => {
```
Nokel81, line 76: I don't like that these are no longer methods, if you want to pretend that JS allows referencing to methods like this I would recommend just calling `autoBind(this)` in the constructor instead.

ixrock (approved):

PR #5218, opened 2022-04-07
Remove "autobind-decorator" and @boundMethod

ixrock:
In favour of `auto-bind` (used in class constructors only) + arrow-functions when applicable:

**Rule of thumb for `autoBind(this)`:** 

- 2 or more class methods needs to be bound with `this`
- class methods participating in inheritance and/or by performance reasons you want to keep it in prototype chain

```ts
constructor(){ autoBind(this) }
```

**In other cases for single custom bindings:**

```ts
class MyClass {
  constructor() {
    this.method = this.method.bind(this);
  }

  method() {
  }

  // as class-field arrow-function (recommended)
  method2 = () => {};
}
```


aleksfront (approved):
LGTM

Nokel81 (approved):

PR #5224, opened 2022-04-08
Move devserver to different port

DmitriyNoa:
Port 8080 is pretty common and often used. Currently, if something is running on 8080, ``make dev`` fails with unavailable port error.

``
Exception:
Error: connect ECONNREFUSED 127.0.0.1:8080
at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1146:16)
``

### DIFF
--- a/src/common/vars.ts
+++ b/src/common/vars.ts
@@ -136,4 +136,4 @@ export const docsUrl = "https://docs.k8slens.dev/main/" as string;
 
 export const sentryDsn = packageInfo.config?.sentryDsn ?? "";
 
-export const webpackDevServerPort: number = +process.env.WEBPACK_DEV_SERVER_PORT || 9191;
+export const webpackDevServerPort: number = Number(process.env.WEBPACK_DEV_SERVER_PORT) || 9191;
--- a/src/common/vars.ts
+++ b/src/common/vars.ts
@@ -135,3 +135,5 @@ export const appSemVer = new SemVer(packageInfo.version);
 export const docsUrl = "https://docs.k8slens.dev/main/" as string;
 
 export const sentryDsn = packageInfo.config?.sentryDsn ?? "";
+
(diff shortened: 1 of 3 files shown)


panuhorsmalahti (dismissed):

webpack.dev-server.ts
```diff
@@ -26,6 +26,7 @@ function createDevServer(): WebpackDevServer {
     },
     allowedHosts: "all",
     host: "localhost",
+    port: 9191,
```
ixrock: How about moving webpack-server port to common `vars.ts`, e.g.?

```ts
export const webpackDevServerPort: number = +process.env.WEBPACK_DEV_SERVER_PORT || 9191;
```

ixrock (dismissed):

webpack.dev-server.ts
Nokel81: I would have suggested letting it decide and then parsing out the value would be best but this also is a good idea.

Nokel81 (dismissed):

webpack.dev-server.ts
DmitriyNoa: Move to common

src/common/vars.ts
```diff
@@ -135,3 +135,5 @@ export const appSemVer = new SemVer(packageInfo.version);
 export const docsUrl = "https://docs.k8slens.dev/main/" as string;
 
 export const sentryDsn = packageInfo.config?.sentryDsn ?? "";
+
+export const webpackDevServerPort: number = +process.env.WEBPACK_DEV_SERVER_PORT || 9191;
```
panuhorsmalahti: Sorry to bikeshed, but `Number(process.env.WEBPACK_DEV_SERVER_PORT)` is more readable.

Nokel81 (approved):

ixrock (approved):

PR #5230, opened 2022-04-11
Add optional destination property to app event

jweak:
Motivation for this is to app event bus add for support other extensions send telemetry events to their own services.

This makes telemetry extension able to send telemetry to other services than the property belonging to `Lens IDE`.

Other extensions wouldn't need to know about the user telemetry settings since the extension wouldn't send the event if user has not allowed it.

### DIFF
--- a/src/common/app-event-bus/event-bus.ts
+++ b/src/common/app-event-bus/event-bus.ts
@@ -5,15 +5,11 @@
 
 import { EventEmitter } from "../event-emitter";
 
-interface AppEventParams {
-  [key: string]: any;
-}
-
 export interface AppEvent {
   name: string;
   action: string;
   destination?: string;
-  params?: AppEventParams;
+  params?: Record<string, any>;
 }
 
 export const appEventBus = new EventEmitter<[AppEvent]>();
--- a/src/common/app-event-bus/event-bus.ts
+++ b/src/common/app-event-bus/event-bus.ts
@@ -5,30 +5,15 @@
 
 import { EventEmitter } from "../event-emitter";
 
-export interface Keyable { [key: string]: any }
-export interface LensAppEvent {
-  type: "APP_EVENT";
+interface AppEventParams {
+  [key: string]: any;
+}
+
+export interface AppEvent {
   name: string;
   action: string;
-  params?: Keyable;
+  destination?: string;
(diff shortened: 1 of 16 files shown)


src/common/app-event-bus/event-bus.ts
```diff
@@ -5,10 +5,15 @@
...
+interface AppEventParams {
+  [key: string]: any;
+}
+
 export interface AppEvent {
   name: string;
   action: string;
-  params?: object;
+  destination?: string;
+  params?: AppEventParams;
```
Nokel81: nit: I would just use `Record<string, any>` inline.

Nokel81 (dismissed):

Nokel81:
I don't think this is a breaking change anymore

src/common/app-event-bus/event-bus.ts
jweak: sure why not 👍

Nokel81 (approved):

src/common/app-event-bus/event-bus.ts
```diff
@@ -8,7 +8,8 @@ import { EventEmitter } from "../event-emitter";
 export interface AppEvent {
   name: string;
   action: string;
-  params?: object;
+  destination?: string;
```
jim-docker, line 11: could `destination` just be kept in `params`? Or are `params` "url params"?

jim-docker (approved):

src/common/app-event-bus/event-bus.ts
Nokel81: Params is more of "payload" or "metadata"

jim-docker: Who is expected to assign `destination`?

Nokel81: The author of the event. The main implementation will be in lens-ide and the telemetry extension.

jweak: You can see it here https://github.com/lensapp/lens-ide/pull/108

jweak: And we would add the `destination: "Segment"` from the extension side.

PR #5234, opened 2022-04-12
Fix: replace all list items on store.loadAll()

### LINKED ISSUE #4897: Resource removal not always detected
Resource deletion outside of Lens not always observed by Lens

1. Open Create resource
2. pick Deployment from templates
3. Create
4. go to Deployments tab, observe nginx-deployment was created
5. go to another tab (like Namespaces)
6. In the terminal run `kubectl delete deployment/nginx-deployment`
7. go to Deployments tab, observe nginx-deployment is still there

https://user-images.githubusercontent.com/40840436/154582906-6b96dd92-bad5-4f01-9d8f-2ea7f9d7593e.mov

The deleted deployed pods also remain in the pods list in some situations

happens on 5.4.0-beta.2 and 5.3.4

reload or disconnect/reconnect fixes the issue. Also if the resource watch detects another change (i.e. create another deployment) the deleted resource disappears from the list.

aleksfront:
Adding `merge: false` rule on KubeObjectListLayout mount causing all store items to be replaces with loaded ones.

https://user-images.githubusercontent.com/9607060/162903929-73dae1bb-4419-465d-ba05-df4e163afb83.mov

Fixes #4897 

### DIFF
--- a/src/renderer/kube-watch-api/kube-watch-api.ts
+++ b/src/renderer/kube-watch-api/kube-watch-api.ts
@@ -96,7 +96,7 @@ export class KubeWatchApi {
 
     const loadThenSubscribe = async (namespaces: string[]) => {
       try {
-        await store.loadAll({ namespaces, reqInit: { signal: childController.signal }, onLoadFailure });
+        await store.loadAll({ namespaces, merge: false, reqInit: { signal: childController.signal }, onLoadFailure });
         unsubscribe.push(store.subscribe({ onLoadFailure, abortController: childController }));
       } catch (error) {
         if (!(error instanceof DOMException)) {


src/renderer/kube-watch-api/kube-watch-api.ts
```diff
@@ -96,7 +96,7 @@ export class KubeWatchApi {
 
     const loadThenSubscribe = async (namespaces: string[]) => {
       try {
-        await store.loadAll({ namespaces, reqInit: { signal: childController.signal }, onLoadFailure });
```
Nokel81, line 99: Could we somehow have a unit test to cover this case?

I am think something along the lines of:
1. Add items to store manually
2. call `.subscribe()`
3. Ensure than old items no longer in store.

Nokel81:
I have a concern with this, what happens if multiple different subscribe calls are done with non-equal namespace lists?

Nokel81:
FYI I have made https://github.com/lensapp/lens/pull/5309 which doesn't have (IMO) the issue about multiple non-equal namespace lists being passed into separate calls of `loadAll`.

PR #5240, opened 2022-04-13
Fix misleading no metrics message

### LINKED ISSUE #5200: "Metrics not available" message is misleading
**Describe the bug**

"Right click cluster icon" does not work unless cluster is added to a hotbar.

image

aleksfront:
Change no-metrics message replacing `Right click cluster icon to open cluster settings` with `Open cluster settings` link.

https://user-images.githubusercontent.com/9607060/163190777-9b92ca4d-18ed-4c9d-8ef8-0b0edb77dbe6.mov

Fix #5200


src/renderer/components/+cluster/cluster-no-metrics.tsx
```diff
@@ -3,20 +3,33 @@
...
   className: string;
 }
 
 export function ClusterNoMetrics({ className }: ClusterNoMetricsProps) {
+  function getClusterId() {
+    return catalogEntityRegistry.activeEntity;
+  }
+
+  function openSettingsPage() {
+    broadcastMessage(IpcRendererNavigationEvents.NAVIGATE_IN_APP, `/entity/${getClusterId()?.getId()}/settings`);
```
Nokel81: Could this be a dependency? @Iku-turso does your work currently support this sort of navigation?

Nokel81: If this is a dependency, then we could have a unit test as well as the e2e test.

jim-docker: Could it go straight to the Metrics tab? Maybe not necessary

src/renderer/components/+cluster/cluster-no-metrics.module.scss
```diff
@@ -0,0 +1,18 @@
...
+.ClusterNoMetrics {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  line-height: 1.7;
+  flex-grow: 1;
+
+  .link {
+    color: var(--blue);
```
jim-docker, line 15: should this be a theme-friendly colour? I guess blue is used elsewhere too, and is already friendly to our 2 themes.

ixrock: Currently we have the same color for `--blue` for both themes defined in `lens-dark.json` and `lens-light.json`.

Anyway I would suggest to add some prefix to all themed variables to distinguish them from other css-variables and for easier lookup the roots, e.g. `--theme-blue` or `--lens-blue` (`blue` btw wrong name for themed variable, better to name for this case would be `--theme-link-color`)

Nokel81 (dismissed):

src/renderer/components/+cluster/cluster-no-metrics.tsx
ixrock: If `getClusterId()` will be `undefined` would it be still a valid route? Maybe we should not send anything if cluster not exists?

ixrock (dismissed):

aleksfront:
Using `navigateToEntitySettingsInjectable` instead of `broadcastMessage`.

https://user-images.githubusercontent.com/9607060/165473112-2fa73dee-c3c5-4239-ad7a-50957448e127.mov

ixrock (approved):

Nokel81 (approved):

PR #5249, opened 2022-04-19
Fix: do not use paddings for empty top bar items

aleksfront:
Fixing styles a bit to hide any paddings for empty top bar elements.

#### Why this is needed?

Extension sometimes can add empty invisible elements in TopBar area such as dialogs, modals etc which doesn't actually gets viewed in top bar area. This elements, however, create their own `div`s with unnecessary paddings.

Example:
topbar unnecessary padding

### DIFF
--- a/src/renderer/components/layout/top-bar/top-bar.module.scss
+++ b/src/renderer/components/layout/top-bar/top-bar.module.scss
@@ -26,7 +26,7 @@
   &:first-of-type {
     padding-left: 1.5rem;
 
-    & > * {
+    & > *:not(:empty) {
       margin-right: 1rem;
     }
   }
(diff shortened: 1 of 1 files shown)


jweak (approved):

src/renderer/components/layout/top-bar/top-bar.module.scss
```diff
@@ -26,15 +26,15 @@
   &:first-of-type {
     padding-left: 1.5rem;
 
-    & > * {
+    & > *:not(:empty) {
```
ixrock, line 29: _nit:_
```suggestion
    & > :not(:empty) {
```

ixrock (approved):

PR #5264, opened 2022-04-21
Expose Lens Monaco Editor component to extensions-api

### LINKED ISSUE #5094: Expose Lens Monaco Editor component to public
At present we are not able to use the Monaco Editor used by Lens. We were trying to use this since long time back but unfortunately with lens when we try to add this dependency and use, syntax highlighting was not working. So we have `js-yaml` 

W.r.to the latest lens CRD definitions, we see Monaco Editor used inside side panel. Can this be exposed for easy integration for us? This helps everyone.

ixrock:
Screenshot 2022-04-21 at 15 49 41

**Breaking change in extension's development process:**

- typescript version 4.5+ required to support type on import names, see: [typescript release notes](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#type-on-import-names)

**How to fix:** _(to avoid compilation errors/warnings)_

`npm install typescript -D` or `yarn add typescript` _(latest version TS recommended)_

close #5094

### DIFF
--- a/src/renderer/components/monaco-editor/monaco-editor.tsx
+++ b/src/renderer/components/monaco-editor/monaco-editor.tsx
@@ -9,12 +9,12 @@ import { observer } from "mobx-react";
 import { action, computed, makeObservable, observable, reaction } from "mobx";
 import { editor, Uri } from "monaco-editor";
 import type { MonacoTheme } from "./monaco-themes";
-import type { MonacoValidator } from "./monaco-validators";
-import { monacoValidators } from "./monaco-validators";
+import { type MonacoValidator, monacoValidators } from "./monaco-validators";
 import { debounce, merge } from "lodash";
-import { cssNames, disposer } from "../../utils";
+import { autoBind, cssNames, disposer } from "../../utils";
 import { UserStore } from "../../../common/user-store";
 import { ThemeStore } from "../../theme.store";
+import logger from "../../../main/logger";
 
 export type MonacoEditorId = string;
 
@@ -45,8 +45,8 @@ export const defaultEditorProps: Partial<MonacoEditorProps> = {
 
 @observer
 export class MonacoEditor extends React.Component<MonacoEditorProps> {
-  static defaultProps = defaultEditorProps as object;
-  static viewStates = new WeakMap<Uri, editor.ICodeEditorViewState>();
+  static readonly defaultProps = defaultEditorProps as object;
+  static readonly viewStates = new WeakMap<Uri, editor.ICodeEditorViewState>();
 
   static createUri(id: MonacoEditorId): Uri {
     return Uri.file(`/monaco-editor/${id}`);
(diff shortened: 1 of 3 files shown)


aleksfront (commented):
Got this errors trying to compile extension with new `k8slens/extensions` types:
Screenshot 2022-04-22 at 11 51 40

It points to the following line
new monaco line

ixrock:
This happens because some extensions are using old typescript < 4.5 see: [typescript release notes](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#type-on-import-names)
So, since it affects only development process, runtime should work as before. 
Moreover we agreed with @jakolehm that we should add our Lens release notes about it to help developers fix this issue faster.

aleksfront (dismissed):
LGTM, works with fresh (at least 4.5.0) typescript version.

src/renderer/components/monaco-editor/monaco-editor.tsx
```diff
@@ -55,22 +55,18 @@ export class MonacoEditor extends React.Component<MonacoEditorProps> {
   public staticId = `editor-id#${Math.round(1e7 * Math.random())}`;
   public dispose = disposer();
```
Nokel81: Shouldn't this also be private?

src/renderer/components/monaco-editor/monaco-editor.tsx
```diff
@@ -55,22 +55,18 @@ export class MonacoEditor extends React.Component<MonacoEditorProps> {
...
 
-  // TODO: investigate how to replace with "common/logger"
-  //  currently leads for stucking UI forever & infinite loop.
-  //  e.g. happens on tab change/create, maybe some other cases too.
-  logger = console;
-
   @observable.ref containerElem: HTMLElement;
   @observable.ref editor: editor.IStandaloneCodeEditor;
```
Nokel81, line 59: Shouldn't these also be private?

Nokel81 (dismissed):

src/renderer/components/monaco-editor/monaco-editor.tsx
ixrock: Maybe `protected`, but not sure.. @jakolehm any suggestions?

src/renderer/components/monaco-editor/monaco-editor.tsx
ixrock: Fixed

src/renderer/components/monaco-editor/monaco-editor.tsx
ixrock: I think these ones should be public as if developer will decide to make a `ref={monaco => monaco.editor}` those would be available.

Nokel81 (approved):

aleksfront (approved):

PR #5273, opened 2022-04-25
adding https_proxy to execHelm env

### LINKED ISSUE #3787: Unable to add helm repos (from pre-populated list or manual)
**Describe the bug**
Errors encountered when trying to add helm repos. Seems there's communication issue somewhere.
* I am on a corporate network using proxy for non-cluster communication.
* Proxy settings are validated and allow list of repositories to load. 
* Error is after selecting any repo.

**To Reproduce**
1. Go to File > Preferences
2. Click on Kubernetes
3. Select 'prometheus-community' from list of Repositories under Helm Charts

**Expected behavior**
Add helm repos without any communication issues.
Install and manage helm charts without any issues.
I would expect the behavior to be very similar to how I am able to install lens-metrics on cluster using Lens app. Point and click and it just works!

**Screenshots**

- Lens Version: 5.2.0-latest.20210908.1
- OS: Windows 10
- Installation method (e.g. snap or AppImage in Linux): Windows installer (auto-update from prior version)

**Logs:**
```
Adding helm branch prometheus-community has failed: Error: looks like "https://prometheus-community.github.io/helm-charts" is not a valid chart repository or cannot be reached: Get "https://prometheus-community.github.io/helm-charts/index.yaml": read tcp 135.70.245.39:50417->185.199.109.153:443: wsarecv: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.
```

**Kubeconfig:**
```
Not applicable - cluster communication works fine
```

**Additional context**
Also tried to manually add repos after installing helm CLI on my local machine.
* That at least adds the repo locally
* Still unable to perform any action on charts from Apps section in cluster
* Can select available charts after manually adding repo, but the detail screen that pops in from right just shows loading and never does anything else.

jim-docker:
applies the https_proxy env var to all `execHelm` calls, for simplicity and to be certain it is set when needed for helm (could instead set it for specific invocations where http access is certain)

This code needs conversion to DI, but beyond the scope of this bug fix(?)

Fixes #3787

### DIFF
--- a/src/main/helm/exec.ts
+++ b/src/main/helm/exec.ts
@@ -17,9 +17,7 @@ export async function execHelm(args: string[], options?: BaseEncodingOptions & E
   try {
     const opts = { ...options };
 
-    if (!opts.env) {
-      opts.env = process.env
-    }
+    opts.env ??= process.env;
 
     if (!opts.env.HTTPS_PROXY && UserStore.getInstance().httpsProxy) {
       opts.env.HTTPS_PROXY = UserStore.getInstance().httpsProxy;
--- a/src/main/helm/exec.ts
+++ b/src/main/helm/exec.ts
@@ -7,6 +7,7 @@ import { promiseExecFile } from "../../common/utils/promise-exec";
 import type { BaseEncodingOptions } from "fs";
 import type { ExecFileOptions } from "child_process";
 import { helmBinaryPath } from "../../common/vars";
+import { UserStore } from "../../common/user-store";
 
 /**
  * ExecFile the bundled helm CLI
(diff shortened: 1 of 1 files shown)


Nokel81 (approved):

src/main/helm/exec.ts
```diff
@@ -7,14 +7,23 @@ import { promiseExecFile } from "../../common/utils/promise-exec";
...
  * @returns STDOUT
  */
 export async function execHelm(args: string[], options?: BaseEncodingOptions & ExecFileOptions): Promise<string> {
   try {
-    const { stdout } = await promiseExecFile(helmBinaryPath.get(), args, options);
+    const opts = { ...options };
+
+    opts.env ??= process.env;
+
+    if (!opts.env.HTTPS_PROXY && UserStore.getInstance().httpsProxy) {
```
jakolehm, line 22: Imho doing `!opts.env.HTTPS_PROXY` here will lead to duplicate code (or different logic in some other place). This should be abstracted away so that we don't copy if to every place where proxy is needed.

jakolehm (approved):

src/main/helm/exec.ts
Nokel81: Agreed, though I think that be done when this code is made DI complaint.

jim-docker: Not sure I understand. You mean if someone copies this code to add proxy support elsewhere (possibly unrelated to helm)? It is flexibility in case the caller of `execHelm` actually wants different behaviour. Maybe not necessary.

Nokel81: I think that Jari means is that we copy over `HTTPS_PROXY` env var in many places. It would be great if we have a helper function injectable to do it instead. Basically, more DRY

PR #5277, opened 2022-04-26
Upgrade ReactSelect to latest version

ixrock:
fix: `prop.styles` was previously ignored before passing to final `react-select` component for rendering
fix: dev-mode error in `@pmmmwh/react-refresh-webpack-plugin`, close #5282


Nokel81 (dismissed):

Nokel81 (dismissed):

Nokel81 (dismissed):

panuhorsmalahti (dismissed):

Nokel81 (approved):

ixrock:
@panuhorsmalahti PTALA, all tests fixed.

Nokel81:
@ixrock FYI There might be a conflict with #5282 given that it seems you have also fixed the dev mode in this PR. We probably could just merge both of these since you have already gotten the approval once. WDYT?

ixrock:
> @ixrock FYI There might be a conflict with #5282 given that it seems you have also fixed the dev mode in this PR. We probably could just merge both of these since you have already gotten the approval once. WDYT?

It should be easy to fix or even auto-fixable by Github? Do we still need both?
I assume conflict may be because `@pmmmwh/react-refresh-webpack-plugin`, right?

Nokel81:
> I assume conflict may be because @pmmmwh/react-refresh-webpack-plugin, right?

Yes.

Should be able to be done on github maybe. We need the other one because it includes a CI check to try and catch things like this in the future.

panuhorsmalahti (approved):

PR #5284, opened 2022-04-27
Show extension preferences in separate page

aleksfront:
Grouping custom preferences by extension and showing them in separate tabs. Previously settings from all extensions were listed in one page.

extension custom settings

This PR is rewritten from https://github.com/lensapp/lens/pull/4651 using recently introduced routing mechanism.


src/renderer/components/+preferences/extensions.tsx
```diff
@@ -11,18 +11,21 @@ import type { RegisteredAppPreference } from "./app-preferences/app-preference-r
...
+  extensionName: string;
 }
 
-const NonInjectedExtensions = ({ preferenceItems }: Dependencies) => (
+const NonInjectedExtensions = ({ preferenceItems, extensionName }: Dependencies) => (
   <Preferences data-testid="extension-preferences-page">
     <section id="extensions">
-      <h2>Extensions</h2>
-      {preferenceItems.get().map((preferenceItem) => (
+      <h2>{extensionName} settings</h2>
```
jakolehm: I think we should use consistent naming (settings vs preferences)

```suggestion
      <h2>{extensionName} preferences</h2>
```

src/renderer/components/+preferences/preferences-navigation/preferences-navigation.tsx
```diff
@@ -13,30 +13,52 @@ import type {
...
+    >
+      <div className="header">Preferences</div>
+
+      {generalNavItems.map(renderTab)}
+
+      {extensionNavItems.length > 0 && (
+        <div data-testid="extension-settings">
+          <hr/>
+          <div className="header flex items-center">
+            <Icon material="extension" smallest className="mr-3"/> Custom Settings
```
jakolehm: "Custom Settings" does not sound right when we are in "Preferences". How about just:

```suggestion
            <Icon material="extension" smallest className="mr-3"/> Extensions
```

jakolehm (commented):
I think we should handle bundled extensions in a bit different way (vs 3rd party extensions). Maybe those bundled extensions should not be under "Extensions" at all?

aleksfront:
@jakolehm Added `AppPreferenceTabRegistration` allowing to create "native" preference tabs for the extensions. It will work with following code in `renderer.tsx`:

```
appPreferenceTabs = [{
  title: "Specific tab",
  id: "specific-extension-tab",
  orderNumber: 100,
}];

appPreferences = [
  {
    title: "Hello world Preferences",
    showInPreferencesTab: "specific-extension-tab",
    components: {
      Input: () => <Renderer.Component.Input />,
      Hint: () => <p>Hello world preferences</p>
    }
  }
];
```

Note that appPreferenceTabs `id` and `showInPreferencesTab` param from `appPreferences` should be the same.

https://user-images.githubusercontent.com/9607060/166891090-be8bd053-4bc3-42bc-be3d-b01b5a6eecf2.mov

src/behaviours/preferences/navigation-to-extension-specific-preferences.test.tsx
```diff
@@ -134,3 +234,82 @@ const extensionStubWithExtensionSpecificPreferenceItems: Partial<LensRendererExt
   ],
 };
 
+const someOtherExtensionStubWithExtensionSpecificPreferenceItems: Partial<LensRendererExtension> = {
```
Nokel81: Can we have a test where two extensions register a `appPreferencesMenu` with the same id and maybe also with different titles?

Nokel81:
Looks good except for a few unit tests

src/behaviours/preferences/navigation-to-extension-specific-preferences.test.tsx
aleksfront: Added test coverage for cases when different extensions registered same `ids` for their tabs. Application will show each of such tabs. This is done by adding unique `extensionId` prefix.

src/behaviours/preferences/navigation-to-extension-specific-preferences.test.tsx
```diff
@@ -102,6 +118,12 @@ describe("preferences - navigation to extension specific preferences", () => {
         expect(actual).not.toBeNull();
       });
 
+      it("link doesn't have 'active' class", () => {
```
Iku-turso: The name of test describes a technicality, and not a behaviour.

src/renderer/components/+preferences/preferences-navigation/navigate-to-preference-tab.injectable.ts
```diff
@@ -12,7 +12,7 @@ const navigateToPreferenceTabInjectable = getInjectable({
   instantiate: (di) => {
     const navigateToRoute = di.inject(navigateToRouteInjectionToken);
 
-    return (route: Route) => () => {
+    return (route: Route<unknown>) => () => {
```
Iku-turso: Should this really be unknown?

src/behaviours/preferences/navigation-to-extension-specific-preferences.test.tsx
```diff
@@ -85,6 +86,21 @@ describe("preferences - navigation to extension specific preferences", () => {
       });
     });
 
+    describe("given multiple extensions with and without specific preferences, when navigating to extension specific preferences page", () => {
```
Iku-turso: Test claims to do...
> when navigating to extension specific preferences page

...but it doesn't do that. Consider real hard to practise the [Cycle of TDD](https://blog.cleancoder.com/uncle-bob/2014/12/17/TheCyclesOfTDD.html) to first see the test fail for a good reason before making it pass by adding implementation.

src/behaviours/preferences/navigation-to-extension-specific-preferences.test.tsx
```diff
@@ -128,6 +150,12 @@ describe("preferences - navigation to extension specific preferences", () => {
 
           expect(actual).toBeNull();
         });
+
+        it("link does have 'active' class", () => {
```
Iku-turso: The name of test describes a technicality, and not a behaviour.

src/renderer/components/+preferences/preferences-navigation/extensions-preferences-navigation-item.injectable.ts
```diff
@@ -25,7 +26,7 @@ const extensionPreferencesNavigationItemRegistratorInjectable = getInjectable({
           label: `${extension.name}`,
           navigate: () => navigateToExtensionPreferences(extension.sanitizedExtensionId),
           isActive: computed(() => false),
-          isVisible: computed(() => true),
+          isVisible: computed(() => isVisible),
```
Iku-turso: Should the de-referencing happen within this computed? Looks brittle this way.

Iku-turso: It would be nice to see a unit test within this commit to see that this change is tested, and also study the test to further understand the motivation behind this change.

src/renderer/components/+preferences/preferences-navigation/extensions-preferences-navigation-item.injectable.ts
```diff
@@ -3,10 +3,14 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
```
Iku-turso: Commit name explains technicality and not motivation.

src/behaviours/preferences/navigation-to-extension-specific-preferences.test.tsx
```diff
@@ -90,15 +90,22 @@ describe("preferences - navigation to extension specific preferences", () => {
...
+        await applicationBuilder.addExtensions(someTestExtension, extensionWithoutPreferences, extensionWithSpecificTab);
       });
 
       it("doesn't show link for extension without preferences", () => {
         const actual = rendered.queryByTestId("tab-link-for-extension-without-preferences-id");
 
         expect(actual).toBeNull();
       });
+
+      it("doesn't show link for preferences with 'showInPreferencesTab' param", () => {
```
Iku-turso: Name of a behaviour cannot be technical.

src/behaviours/preferences/navigation-to-extension-specific-preferences.test.tsx
```diff
@@ -90,15 +90,22 @@ describe("preferences - navigation to extension specific preferences", () => {
       beforeEach(async () => {
```
Iku-turso, line 79: This is a commit with tests and implementation together! Nice! :)

src/renderer/components/+preferences/extension-preference-items.injectable.ts
```diff
@@ -2,49 +2,35 @@
...
+
+      return extension.appPreferences
+        .filter(preference => !preference.showInPreferencesTab)
+        .map(preference => ({
+          id: preference.id,
+          ...preference,
+        }));
     });
   },
+  lifecycle: lifecycleEnum.transient,
```
Iku-turso: This injectable returns already a computed, and therefore it should not be `transient`.

src/renderer/components/+preferences/extension-preference-items.injectable.ts
```diff
@@ -2,14 +2,23 @@
...
 import rendererExtensionsInjectable from "../../../extensions/renderer-extensions.injectable";
 
+import type { LensRendererExtension } from "../../../extensions/lens-renderer-extension";
 import type { RegisteredAppPreference } from "./app-preferences/app-preference-registration";
 
+interface ExtensionPreferenceItem extends RegisteredAppPreference {
+  extension: LensRendererExtension;
+}
+
+export const extensionPreferenceItemInjectionToken = getInjectionToken<ExtensionPreferenceItem>({
```
Iku-turso: This should be in a separate file :)

src/renderer/components/+preferences/preferences-navigation/preference-navigation-items.injectable.ts
```diff
@@ -20,6 +20,7 @@ export interface PreferenceNavigationItem {
   isVisible: IComputedValue<boolean>;
   navigate: () => void;
   orderNumber: number;
+  fromExtension?: boolean;
```
Iku-turso: Note: Ideally, the contract of a `PreferenceNavigationItem` should require no knowledge of extensions. There is a change in the works to make something like this easier to implement, and in a reactive way.

Iku-turso: Does this work, if an extension is enabled, disabled, and then re-enabled?

src/renderer/components/+preferences/preferences-navigation/__tests__/preferences-navigation.test.tsx
```diff
@@ -0,0 +1,182 @@
...
+      expect(container).toHaveTextContent("Custom Settings");
+    });
+
+    it("renders extension navigation items", () => {
+      const { container } = render(
+        <PreferencesNavigation/>,
+      );
+
+      expect(container).toHaveTextContent("lensapp-node-menu");
+      expect(container).toHaveTextContent("lensapp-pod-menu");
```
Iku-turso: Multiple expects in test prevent accurate name of test.

src/renderer/components/+preferences/preferences-navigation/extensions-preferences-navigation-item.injectable.ts
aleksfront: I would not change commit message because I find it a bit risky for commit history, but will keep this in mind for future reference.

src/renderer/components/+preferences/preferences-navigation/extensions-preferences-navigation-item.injectable.ts
aleksfront: The motivations here is quite simple: to show extension tab links only when conditions met, instead of always. This change is actually already tested in test cases like https://github.com/lensapp/lens/pull/5284/files#diff-024668245f9688497c5a846e68c643b1ae07015a2b255dd2ee550d8f6ae9981cR84-R91

which shows and hide preferences links in different conditions.

src/renderer/components/+preferences/preferences-navigation/preference-navigation-items.injectable.ts
aleksfront: Most likely this kind of testing functionality (disable/enable extensions after adding to app builder) would require changes in `ApplicationBuilder`. I suggest to provide this test cases on follow up PR.

src/renderer/components/+preferences/extension-preference-items.injectable.ts
```diff
@@ -2,49 +2,40 @@
...
-        ),
-      );
+      if (tabId) {
+        return preferences.filter(preference => preference.showInPreferencesTab == tabId);
+      }
+
+      return preferences.filter(preference => !preference.showInPreferencesTab);
     });
   },
+  lifecycle: lifecycleEnum.singleton,
```
Nokel81: nit: this is the default

src/renderer/components/+preferences/extensions.tsx
```diff
@@ -36,8 +43,20 @@ export const Extensions = withInjectables<Dependencies>(
...
-    }),
+    getProps: (di) => {
+      const pathParameters = di.inject(currentPathParametersInjectable);
+      const extensionId = pathParameters.get().extensionId;
+      const extensions = di.inject(rendererExtensionsInjectable);
+      const extension = extensions.get().find((extension) => extension.sanitizedExtensionId === extensionId);
+
+      if (!extension) {
+        throw new Error(`Extension ${extensionId} not found`);
+      }
```
Nokel81: Wouldn't it be better to display some error on the screen?

src/behaviours/preferences/navigation-to-extension-specific-preferences.test.tsx
```diff
@@ -87,14 +144,143 @@ describe("preferences - navigation to extension specific preferences", () => {
...
+        expect(actual).toBeInTheDocument();
+      });
+
+      it("shows tab from the second extension", () => {
+        const actual = rendered.getByTestId("tab-link-for-extension-duplicated-tab-page-id-nav-item-metrics-extension-tab");
+
+        expect(actual).toBeInTheDocument();
+      });
+
+      describe("when navigating to first extension tab", () => {
```
Nokel81, line 272: I think that we should also specify the behaviour when navigating to a specific settings tab from outside the settings page, especially when there are multiple extensions that register the same ID.

Furthermore, I would like to see some tests about two extensions where extensions A registers a new tab and extension B uses that tab.

Nokel81 (commented):
Looking very good, missing a few behaviour test cases though they might be able to be moved to a future PR if the time is needed.

src/renderer/components/+preferences/app-preference-tab/app-preference-tab-registration.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+export interface AppPreferenceTabRegistration {
```
ixrock, line 6: Why this needs to have separate file?

src/renderer/components/+preferences/preferences-navigation/extensions-preferences-navigation-item.injectable.ts
```diff
@@ -3,49 +3,54 @@
...
-      navigate: navigateToPreferenceTab(route),
-      isActive: routeIsActive,
-
-      isVisible: computed(
-        () => preferenceItems.get().length > 0,
-      ),
-
-      orderNumber: 70,
+    return (ext) => {
+      const extension = ext as LensRendererExtension;
```
ixrock: Why not like this? Is there something wrong with types in args?

```suggestion
    return (extension: LensRendererExtension) => {
```

src/renderer/components/+preferences/preferences-navigation/extensions-preferences-navigation-item.injectable.ts
```diff
@@ -3,49 +3,54 @@
...
+          navigate: () => navigateToExtensionPreferences(extension.sanitizedExtensionId),
+          isActive,
+          isVisible,
+          orderNumber: 20,
+          fromExtension: true,
+        }),
+        injectionToken: preferenceNavigationItemInjectionToken,
+      });
+
+      di.register(extensionInjectable);
```
ixrock: Haven't seen before. If we not register the injectable it won't work or what's for?

ixrock (dismissed):

src/renderer/components/+preferences/preferences-navigation/extensions-preferences-navigation-item.injectable.ts
Nokel81: Unfortunately that doesn't work because the typing required are `(extension: LensExtension, extensionInstallationCount: number) => void`. This cast was previously done in `ExtensionLoader` but for now needs to be done in these functions.

src/renderer/components/+preferences/preferences-navigation/extensions-preferences-navigation-item.injectable.ts
aleksfront: I can only suggest but yes it doesn't work if we just return `getInjectable()` without `di.register`. @jansav Can you bring more clarification there?

src/renderer/components/+preferences/app-preference-tab/app-preference-tab-registration.ts
aleksfront: No specific reason to keep it in separate file.

src/renderer/components/+preferences/extension-preference-items.injectable.ts
aleksfront: Yes, but typescript won't allow to leave injectable without this line:
errors without lifecycle

Nokel81: Ah I see now, this is because you are using an injection parameter (pathParams). @jansav Should specifying singleton with an injection parameter be even allowed?

Why don't you just use `di.inject(currentPathParametersInjectable)` within this injectable?

src/behaviours/preferences/navigation-to-extension-specific-preferences.test.tsx
aleksfront: Add some external test cases navigating directly to extension specific tabs. 

> Furthermore, I would like to see some tests about two extensions where extensions A registers a new tab and extension B uses that tab.

Added such test case too, and tabs registered in different extension don't affect each other, because specific tab route "located" inside extension route, for example
`path: "/preferences/extension/:extensionId/:tabId?"`

src/renderer/components/+preferences/extensions.tsx
```diff
@@ -49,13 +52,13 @@ export const Extensions = withInjectables<Dependencies>(
       const extensions = di.inject(rendererExtensionsInjectable);
       const extension = extensions.get().find((extension) => extension.sanitizedExtensionId === extensionId);
 
-      if (!extension) {
-        throw new Error(`Extension ${extensionId} not found`);
-      }
+      // if (!extension) {
+      //   throw new Error(`Extension ${extensionId} not found`);
+      // }
```
ixrock: _nit:_ some leftovers/ dead code?

src/renderer/components/+preferences/preferences-navigation/extensions-preferences-navigation-item.injectable.ts
jansav: Sure. Calling `getInjectable` doesn't have any side-effects. It just gets the injectable as it says. Here we are late-registering more implementations for `preferenceNavigationItemInjectionToken`. 

This code is registrator which adapts current extension API to the implementation we are using in Open Lens. This is to allow us do the required refactorings OpenLens without breaking the extension API. In the future we will be removing all of these registrators and export direct interfaces.

Naming of this file should mention "registrator".

src/renderer/components/+preferences/extensions.tsx
```diff
@@ -16,7 +16,7 @@ import rendererExtensionsInjectable from "../../../extensions/renderer-extension
 
 interface Dependencies {
   preferenceItems: IComputedValue<RegisteredAppPreference[]>;
-  extensionName: string;
+  extensionName: string | undefined;
```
ixrock: _nit:_ 

```suggestion
extensionName?: string;
```

ixrock (dismissed):

src/renderer/components/+preferences/extension-preference-items.injectable.ts
jansav: No. `singleton` lifecycle should not have parameters. Reasons: 
- What happens when you inject with different parameters?
- You would need to specify the parameter where you inject

src/renderer/components/test-utils/get-application-builder.tsx
```diff
@@ -181,7 +185,7 @@ export const getApplicationBuilder = () => {
     setMenuItems: (items) => {
       trayMenuItemsStateFake = items;
     },
-    setIconPath: (path) => {
+    setIconPath: (path: string) => {
```
Nokel81: nit: adding this type here is not necessary because it is declared above on line 73

src/renderer/components/+preferences/extensions.tsx
```diff
@@ -36,8 +46,16 @@ export const Extensions = withInjectables<Dependencies>(
...
 
   {
-    getProps: (di) => ({
-      preferenceItems: di.inject(extensionsPreferenceItemsInjectable),
-    }),
+    getProps: (di) => {
+      const pathParameters = di.inject(currentPathParametersInjectable);
+      const extensionId = pathParameters.get().extensionId;
+      const extensions = di.inject(rendererExtensionsInjectable);
+      const extension = extensions.get().find((extension) => extension.sanitizedExtensionId === extensionId);
```
Nokel81: Because these won't be done within an observer component I don't think this will be reacted to. Therefore it should be an injectable of its own (resolving to some computed).

src/renderer/components/+preferences/extension-preference-items.injectable.ts
Nokel81: That is what I thought. I had thought that the typings I did prevented this, but I guess not.

src/renderer/components/+preferences/extensions.tsx
jansav: Indeed dereferencing is in wrong place. All deferencing should happen inside reactive context (e.g. in observer). 

Here it could be:


**extensionPreferencesModelInjectable**
```
const extensionPreferencesModel = getInjectable({
  id: 'extension-preferences-model',

  instantiate: (di) => {
    const pathParameters = di.inject(currentPathParametersInjectable);
    const extensions = di.inject(rendererExtensionsInjectable);
      
    return computed(() => {
      const extensionId = pathParameters.get().extensionId;
      const targetExtension = extensions.get().find((extension) => extension.sanitizedExtensionId === extensionId);
      
      if (!targetExtension) {
        throw new Error('Tried to get extension preferences for extension that does not exist.')
      }
      
      return {
        extensionName: targetExtension.manifest.name,
        preferenceItems: getExtensionPreferenceItems(targetExtension)
      }
    });
  }
});
```


Then `getProps` in component could be. 

```
return {
  model: di.inject(extensionPreferencesModelInjectable)
};
```

In that way, we have one dependency for the page. Which is page model which composes stuff from other places. And rereferencing reactive values happen inside reactive context.

src/renderer/components/+preferences/preferences-navigation/extension-tab-preferences-navigation-item.injectable.ts
```diff
@@ -0,0 +1,63 @@
...
+      );
+      const route = di.inject(extensionPreferencesRouteInjectable);
+      const routeIsActive = di.inject(routeIsActiveInjectable, route);
+      const pathParameters = di.inject(currentPathParametersInjectable);
+
+      const injectables = pipeline(
+        extension.appPreferenceTabs,
+
+        map((tab) => {
+          const id = `extension-${extension.sanitizedExtensionId}-nav-item-${tab.id}`;
```
jansav, line 37: For now, need to know stuff. In line 24. there is second argument which is `extensionInstallationCount` Let's add that to the `id` in line 37. This is to allow enabling, disabling and re-enabling extension. 

This is something that we'll get rid of soon being absurd technical complexity. :)

src/renderer/components/+preferences/preferences-navigation/extensions-preferences-navigation-item.injectable.ts
```diff
@@ -3,49 +3,54 @@
...
+      );
+      const isVisible = computed(() => extensionHasPreferences && extensionHasGeneralPreferences);
+      const extensionRoute = di.inject(extensionPreferencesRouteInjectable);
+      const pathParameters = di.inject(currentPathParametersInjectable);
+      const routeIsActive = di.inject(routeIsActiveInjectable, extensionRoute);
+
+      const isActive = computed(() => routeIsActive.get() && pathParameters.get().extensionId === extension.sanitizedExtensionId);
+
+      const extensionInjectable = getInjectable({
+        id: `extension-preferences-navigation-item-${extension.sanitizedExtensionId}`,
```
jansav: For now, need to know stuff. In line 19. there is second argument which is `extensionInstallationCount` Let's add that to the `id` in line 36. This is to allow enabling, disabling and re-enabling extension. 

This is something that we'll get rid of soon being absurd technical complexity. :)

src/renderer/components/+preferences/preferences-navigation/extensions-preferences-navigation-item.injectable.ts
aleksfront: Renamed the file.

src/renderer/components/+preferences/extension-preference-items.injectable.ts
aleksfront: Changed lifecycle to `transient`.

src/renderer/components/+preferences/preferences-navigation/preference-navigation-items.injectable.ts
jansav: #5609 adds this functionality to `ApplicationBuilder`

src/renderer/components/+preferences/preferences-navigation/extension-tab-preferences-navigation-item.injectable.ts
aleksfront: Yes, and my additional question is do we really need such long ids? 
```
const id = `extension-${extension.sanitizedExtensionId}-${extensionInstallationCount}-nav-item-${tab.id}`;
```
They really stand out from the "native" ids on the preferences sidebar list.

jansav: https://github.com/lensapp/lens/pull/5662 tackles the thing I mentioned in earlier comment. 

Value of ID doesn't really matter as long as it's unique. What makes it unique here, is `extension.sanitizedExtensionId`, `tab.id` and `extensionInstallationCount` (which is removed completely in the PR, so that you don't have to add it anymore)

src/renderer/components/+preferences/preferences-navigation/extensions-preferences-navigation-item.injectable.ts
jansav: https://github.com/lensapp/lens/pull/5662 This PR removes the `extensionInstallationCount` for good.

src/renderer/components/+preferences/extensions.tsx
```diff
@@ -4,25 +4,29 @@
...
-      {preferenceItems.get().map((preferenceItem) => (
+      <h2>
+        {extensionName}
+        {" "}
+        preferences
+      </h2>
+      {preferenceItems.map((preferenceItem, index) => (
         <ExtensionSettings
-          key={preferenceItem.id}
+          key={`${preferenceItem.id}-${index}`}
```
Nokel81: Is the preference item id not unique enough?

src/renderer/components/+preferences/preferences-navigation/extension-tab-preferences-navigation-item.injectable.ts
```diff
@@ -0,0 +1,63 @@
...
+              isVisible: computed(() => true),
+              isActive,
+            }),
+          });
+        }),
+      );
+
+      di.register(...injectables);
+
+      return;
```
Nokel81: nit: I don't think this is necessary

Nokel81 (dismissed):

src/behaviours/preferences/navigation-to-extension-specific-preferences.test.tsx
```diff
@@ -87,14 +147,202 @@ describe("preferences - navigation to extension specific preferences", () => {
...
+
+        await applicationBuilder.extensions.renderer.enable(extension);
+      });
+
+      it("shows each of registered tabs in general area", () => {
+        const helloTab = rendered.getByTestId("tab-link-for-extension-hello-world-tab-page-id-1-nav-item-hello-extension-tab");
+        const logsTab = rendered.getByTestId("tab-link-for-extension-hello-world-tab-page-id-1-nav-item-logs-extension-tab");
+
+        expect(helloTab).toBeInTheDocument();
+        expect(logsTab).toBeInTheDocument();
```
jansav: These should be split to two separate tests for documentation and better error message in case of failure.

src/behaviours/preferences/navigation-to-extension-specific-preferences.test.tsx
```diff
@@ -6,8 +6,12 @@ import type { RenderResult } from "@testing-library/react";
 import type { ApplicationBuilder } from "../../renderer/components/test-utils/get-application-builder";
```
jansav, line 6: Love the new behaviours. :)

src/renderer/components/+preferences/__tests__/extensions.test.tsx
```diff
@@ -0,0 +1,135 @@
+/**
```
jansav: All of these could and should be covered in behaviours.

src/renderer/components/+preferences/extension-preference-model.injectable.ts
```diff
@@ -0,0 +1,34 @@
...
+    const pathParameters = di.inject(currentPathParametersInjectable);
+    const extensions = di.inject(rendererExtensionsInjectable);
+      
+    return computed(() => {
+      const { extensionId, tabId } = pathParameters.get();
+      const targetExtension = extensions.get().find((extension) => extension.sanitizedExtensionId === extensionId);
+      
+      if (!targetExtension) {
+        throw new Error("Tried to get extension preferences for extension that does not exist.");
+      }
```
jansav: This defensive code shouldn't be needed at all. There is bigger problem, if some would try to observe this in a page where extension is not there.

If this is required now, we can get rid of it by rendering error message in UI if path parameter doesn't lead to extension.

src/renderer/components/+preferences/extensions.tsx
jansav: ID should always be unique.

src/renderer/components/+preferences/app-preference-tab/get-app-preference-tab.ts
```diff
@@ -0,0 +1,18 @@
...
+import { computed } from "mobx";
+import type { LensRendererExtension } from "../../../../extensions/lens-renderer-extension";
+
+interface Dependencies {
+  extensions: IComputedValue<LensRendererExtension[]>;
+}
+
+export const getAppPreferenceTabs = ({ extensions }: Dependencies) => {
+  return computed(() => (
+    extensions.get().flatMap((extension) => extension.appPreferenceTabs)
```
jansav: InjectMany, InjectionToken and registrator should be used. Benefit is that then it will work automatically for OpenLens items as well.

src/renderer/components/+preferences/app-preference-tab/app-preference-tab.injectable.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { getInjectable } from "@ogre-tools/injectable";
+import rendererExtensionsInjectable from "../../../../extensions/renderer-extensions.injectable";
+import { getAppPreferenceTabs } from "./get-app-preference-tab";
+
+const appPreferenceTabInjectable = getInjectable({
+  id: "app-preference-tabs",
+
+  instantiate: (di) =>
+    getAppPreferenceTabs({
```
jansav: Nit: implementation is so simple, so could be inlined.

src/renderer/components/+preferences/preferences-navigation/__tests__/preferences-navigation.test.tsx
```diff
@@ -0,0 +1,164 @@
+/**
```
jansav, line 1: Should be behaviour.

src/renderer/components/+preferences/preferences-navigation/__tests__/preferences-navigation.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+        // Extension navigation items
+        {
+          id: "extension-preferences-navigation-item-lensapp-node-menu",
+          label: "lensapp-node-menu",
+          isActive: computed(() => false),
+          isVisible: computed(() => true),
+          navigate: () => noop,
+          orderNumber: 0,
+          fromExtension: true,
+        },
```
jansav: !! We should write code in a way that whether item is from Open Lens or extension shouldn't matter. Therefore stuff like `fromExtension` is not welcome to navigation items.

src/renderer/components/+preferences/preferences-navigation/extensions-preferences-navigation-item-registrator.injectable.ts
```diff
@@ -0,0 +1,56 @@
...
+      const extensionInjectable = getInjectable({
+        id: `extension-preferences-navigation-item-${extension.sanitizedExtensionId}-${extensionInstallationCount}`,
+        instantiate: () => ({
+          id: `extension-${extension.sanitizedExtensionId}`,
+          label: `${extension.name}`,
+          navigate: () => navigateToExtensionPreferences(extension.sanitizedExtensionId),
+          isActive,
+          isVisible,
+          orderNumber: 20,
+          fromExtension: true,
```
jansav: `fromExtension` is not welcome.

src/renderer/components/+preferences/preferences-navigation/preferences-navigation.tsx
```diff
@@ -13,30 +13,58 @@ import type {
...
+            Extensions
+          </div>
+          <div>
+            {extensionNavItems.map(renderTab)}
+          </div>
+        </div>
+      )}
+    </Tabs>
+  );
+};
```
jansav: This is code that should be refactored in the future to way that we just loop items through. This is because there is no need to hard-code stuff like "Preferences", "Extensions" and order of them. If I want to add third one, I shouldn't need to change this file at all.

src/renderer/components/+preferences/preferences-navigation/preferences-navigation.tsx
```diff
@@ -13,30 +13,58 @@ import type {
...
-    className="flex column"
-    scrollable={false}
-    onChange={(item: PreferenceNavigationItem) => item.navigate()}
-  >
-    <div className="header">Preferences</div>
-
-    {navigationItems.get().map((item) => (
+}: Dependencies) => {
+  const generalNavItems = navigationItems.get().filter(item => !item.fromExtension);
+  const extensionNavItems = navigationItems.get().filter(item => item.fromExtension);
```
jansav: At least, split to two separate injectables. However, in the future real dependency is the array that contains hierarchy of navigation items (and the headers, icons etc.)

src/renderer/components/layout/setting-layout.scss
```diff
@@ -34,7 +34,7 @@
 
     .sidebar {
       width: 218px;
-      padding: 60px 0 60px 20px;
+      padding: 60px 10px 60px 20px;
```
jansav, line 37: Might sound small, but we really should start moving to direction of [8pt-grid](https://medium.com/swlh/the-comprehensive-8pt-grid-guide-aa16ff402179)

src/renderer/components/+preferences/preferences-navigation/__tests__/preferences-navigation.test.tsx
aleksfront: We need this flag to understand where the navigation item comes from. Look at this file `src/renderer/components/+preferences/preferences-navigation/preferences-navigation.tsx` where we decided where to visually show navigation item regarding its source ("generic" or "from extension"). Is there any other way to do this without explicitly marking navigation item?

aleksfront: All navigation items from extensions are renderer within separate block.

jansav: Then, should the "block" be created with a loop as well? In that way we would have support for multiple menu "blocks" immediately. Then menu items wouldn't known anything about extensions. Menu item would know label, "block" / parent so that we can create the menu hierarchy with a loop.


Imagine you have something like this:

```
interface MenuItem {
  id: string;
  parentId: string | null;
  label: string;
  onClick: string;
}

const extensions: MenuItem = {
  id: 'extensions',
  parentId: null,
  label: 'Extensions',
  onClick: () => {}
}

const someExtension: MenuItem = {
  id: 'some-extension-id',
  parentId: 'extensions',
  label: 'Some extension',
  onClick: () => navigateToExtensionPreferences(extension)
}
```

Now you can simply loop through the items and create the UI based on that.

aleksfront: Made refactoring and replaces `fromExtensions` prop to `parent`.

src/renderer/components/+preferences/preferences-navigation/extensions-preferences-navigation-item-registrator.injectable.ts
aleksfront: Removed from the navigation item type.

src/renderer/components/+preferences/preferences-navigation/preferences-navigation.tsx
aleksfront: Refactored to much simpler version of it. Created one more level of abstraction by introducing `navigation groups` which contain related preference links.

src/renderer/components/+preferences/__tests__/extensions.test.tsx
aleksfront: Removed this file and moved some of the test cases from it to behavioural `src/behaviours/preferences/navigation-to-extension-specific-preferences.test.tsx`

src/renderer/components/+preferences/preferences-navigation/__tests__/preferences-navigation.test.tsx
aleksfront: Rewritten with behaviour style.

jansav (approved):

Iku-turso (approved):

PR #5285, opened 2022-04-27
Fix: avoid jumping scroll to the right due default search-input auto-focus

ixrock:
Applicable when clicking between pages:

**Before:**

https://user-images.githubusercontent.com/6377066/165496238-d2c2fef9-04bf-4b0e-8995-608fbaa365f2.mov

**After:**

https://user-images.githubusercontent.com/6377066/165496127-0c0c0a58-4364-485c-9808-b7a9d718ec35.mov

### DIFF
--- a/src/renderer/components/input/search-input-url.tsx
+++ b/src/renderer/components/input/search-input-url.tsx
@@ -5,65 +5,62 @@
 
 import React from "react";
 import debounce from "lodash/debounce";
-import { autorun, observable, makeObservable } from "mobx";
+import { action, autorun, makeObservable, observable } from "mobx";
 import { disposeOnUnmount, observer } from "mobx-react";
-import type { InputProps } from "./input";
-import { SearchInput } from "./search-input";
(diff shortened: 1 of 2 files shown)


aleksfront:
Not sure we should remove search field auto-focus behavior. It may be the feature users already stick with. Instead, maybe we should shrink inputs width?

ixrock:
> Not sure we should remove search field auto-focus behavior. It may be the feature users already stick with. Instead, maybe we should shrink inputs width?

You can easily focus to that search by `Cmd+F` or just by click.

Anyway, when you coming first time to all that pages answer the question: search is the first thing you need to use? How often? And why search-area is taking such priority? _(focus)_ 

Maybe `<CommandPalette/>` should auto-focus/open instead? _(no, just kidding, but it could be more useful in terms of handling bigger datas in one place)_

Nokel81:
Another possible solution would be to reflow the header for very narrow window sizes. Maybe stack the namespace selector and the search bar.

Issue #3872, opened 2021-09-22
No longer seeing a blue 'LED' on connected clusters in the Hotbar


alex-hempel:
**Describe the bug**
Since 5.2.2 I am no longer seeing a blue 'LED' in the Hotbar on clusters that I have connected to.

**To Reproduce**
1. Open Lens
2. Click a pinned cluster to connect
3. Observe that the black dot in the top left of the cluster icon remains black

**Expected behavior**
Previously the black dot would turn blue when a cluster was connected.

**Screenshots**
![image](https://user-images.githubusercontent.com/6581307/134427866-57ce0d2f-5f60-4c7e-92b2-7f6e4875f329.png)

**Environment (please complete the following information):**
- Lens Version: 5.2.3 (observed since 5.2.2)
- OS: Windows 10 21H1

Nokel81:
If you go to the catalog does that entity's status show as `"connected"`?

alex-hempel:
No, all clusters show as disconnected, even the ones I just connected to.

Nokel81:
Would you be able to provide some logs? Please fully stop lens, clear out all the old log files, and then fire up lens. 

https://docs.k8slens.dev/main/faq/?h=logs#where-can-i-find-application-logs

alex-hempel:
Please see log attached. Includes two clusters being connected to.
[lens.log](https://github.com/lensapp/lens/files/7248562/lens.log)

Nokel81:
Can you do the same with the env var `LOG_LEVEL` set to `silly`?

alex-hempel:
Here are all the logs created from a single app start and cluster connect.
[logs.zip](https://github.com/lensapp/lens/files/7255312/logs.zip)

Laurens-W:
Continuing from @Nokel81 's mention above:  Good info to know that docker-desktop does work. Both my rancher clusters don't.

Nokel81:
@Laurens-W What version are you running?

agomez-infoblox:
@Nokel81 I am adding my logs as I am not seeing the proxy error
[lens.log](https://github.com/lensapp/lens/files/8514541/lens.log)

Nokel81:
I think I have tracked down the source of this issue. Do you have multiple kubeconfig sync options (under preferences)?

starudream:
@Nokel81

starudream:
@Nokel81 
[...]

PR #5288, opened 2022-04-27
Add some debug logging to help debug LED issue

Nokel81:

### DIFF
--- a/src/main/cluster-manager.ts
+++ b/src/main/cluster-manager.ts
@@ -168,7 +168,7 @@ export class ClusterManager extends Singleton {
           return entity.status.phase;
         }
 
-        logger.debug(`${logPrefix} not clearing entity ${entity.getName()} DISCONNECTED, reason=""`);
+        logger.debug(`${logPrefix} setting entity ${entity.getName()} to DISCONNECTED, reason="fallthrough"`);
 
         return LensKubernetesClusterStatus.DISCONNECTED;
       })();
--- a/src/main/cluster-manager.ts
+++ b/src/main/cluster-manager.ts
@@ -5,7 +5,7 @@
 
 import "../common/ipc/cluster";
 import type http from "http";
-import { action, makeObservable, observable, observe, reaction, runInAction, toJS } from "mobx";
+import { action, makeObservable, observable, observe, reaction, toJS } from "mobx";
 import type { Cluster } from "../common/cluster/cluster";
 import logger from "./logger";
 import { apiKubePrefix } from "../common/vars";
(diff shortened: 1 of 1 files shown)


src/main/cluster-manager.ts
```diff
@@ -68,7 +68,12 @@ export class ClusterManager extends Singleton {
 
     observe(this.deleting, change => {
       if (change.type === "add") {
-        this.updateEntityStatus(catalogEntityRegistry.getById(change.newValue));
+        runInAction(() => {
+          const entity = catalogEntityRegistry.getById(change.newValue);
+
+          entity.status.phase = LensKubernetesClusterStatus.DELETING;
+          entity.status.enabled = false;
+        });
```
jim-docker: Why break this out of `updateEntityStatus()`? (this PR does more than add debug logging)

Nokel81: Fair enough. And after some more consideration I shouldn't have moved it out since this is wrong now.

src/main/cluster-manager.ts
```diff
@@ -144,22 +144,32 @@ export class ClusterManager extends Singleton {
     } else {
```
jim-docker, line 144: Should you log for the deleting case above too?

src/main/cluster-manager.ts
```diff
@@ -144,22 +144,32 @@ export class ClusterManager extends Singleton {
...
         }
 
         // Extensions are not allowed to use the Lens specific status phases
         if (!lensSpecificClusterStatuses.has(entity?.status?.phase)) {
+          logger.debug(`${logPrefix} not clearing entity ${entity.getName()} status, reason="custom string"`);
+
           return entity.status.phase;
         }
 
+        logger.debug(`${logPrefix} not clearing entity ${entity.getName()} DISCONNECTED, reason=""`);
```
jim-docker: "not clearing entity" again?

jim-docker (dismissed):
much easier to review now :)

src/main/cluster-manager.ts
Nokel81: Maybe, but that doesn't seem to be part of the symptoms that #3872 reports.

src/main/cluster-manager.ts
Nokel81: Good catch

Nokel81:
I am targeting this to 5.4.7 so that we can do a release quicker with this in it to help with debugging.

jim-docker (approved):

ixrock (approved):

PR #5290, opened 2022-04-28
Change engines.lens signature to only respect minimums

### LINKED ISSUE #5212: Change `engines.lens` signature to only respect minimums
Only `^MAJOR.MINOR` and `^MAJOR` will be supported, other values will be accepted and converted into this style.

ixrock:
- removed `src/extensions/extension-compatibility.ts` as unused/partly duplicated code

close #5212


src/extensions/extension-discovery/is-compatible-extension/is-compatible-extension.ts
```diff
@@ -2,51 +2,25 @@
...
-      });
+    if (manifest?.engines.lens) {
+      const [extMajor, extMinor] = manifest.engines.lens.split(".");
+
+      return semver.gte(`${extMajor}.${extMinor}`, `${appMajor}.0`);
     }
 
-    return false;
+    // all extensions by default should be compatible with any Lens-engine (if not specified)
+    return true;
```
Nokel81: specifying was only a requirement after 5.0.0, so if the extension hasn't specified it then aren't compatible.

src/extensions/__tests__/extension-compatibility.test.ts
```diff
@@ -1,128 +0,0 @@
-/**
```
Nokel81, line 1: Why remove all the tests?

ixrock: As mentioned in description it was a test for deleted and not used `src/extensions/extension-compatibility.ts`

src/extensions/extension-discovery/is-compatible-extension/is-compatible-extension.ts
ixrock: Reverted to default `return false`

Nokel81 (dismissed):

Nokel81 (dismissed):

src/extensions/extension-discovery/is-compatible-extension/is-compatible-extension.ts
```diff
@@ -9,44 +9,18 @@ interface Dependencies {
...
-       * compatibility is not limited by it
-       */
-      return semver.satisfies(strippedVersion, manifest.engines.lens, {
-        includePrerelease: true,
+    if (manifest?.engines.lens) {
+      const { major: extMajor, minor: extMinor } = new SemVer(manifest.engines.lens, {
+        loose: true,
       });
+
+      return semver.gte(`${extMajor}.${extMinor}`, `${appVersion}.0`);
```
jim-docker: If I understand this correctly it just requires the major version specified by the extension to be the same or higher than the app major version. The extension's minor version is irrelevant (because it's always >= 0). Is this what we want? So extension written for 5.4 is compatible with all 5.x apps? That's consistent with limiting breaking changes to major version jumps, but are we really enforcing that?

Nokel81: Oh good catch. I missed that

ixrock: So how we should calculate compatibility between extension version and app?

src/extensions/extension-discovery/is-compatible-extension/is-compatible-extension.ts
```diff
@@ -9,44 +9,18 @@ interface Dependencies {
...
     }
 
-    return false;
+    return false; // not compatible by default
```
ixrock: Btw, if this is `false` by default should we mark `LensExtensionManifest` with required `engines.lens` field? Or how extensions know that and not ignored during compatibility check on install/etc?

src/extensions/extension-discovery/is-compatible-extension/is-compatible-extension.ts
ixrock: So `lens-app-extension@5.4.1` should be compatible only with lens-app `>=5.4.1`, but not `5.5.0` and `5.3.10`, right?

Nokel81: I think what you want is to make sure that `^${extMajor}.${extMinor}` is satisfied by `${appMajor}.${appMimor}` (or maybe the `^` needs to be on the app version side, I don't remember).

That way of we make a version 6 of Lens, extensions are not automatically compatible.

src/extensions/extension-discovery/is-compatible-extension/is-compatible-extension.ts
Nokel81: We certainly should make it required elsewhere. In our extension development documentation (https://api-docs.k8slens.dev/v5.4.6/extensions/get-started/anatomy/#extension-manifest) it doesn't imply that it is optional (though we really should update our docs)

Nokel81 (dismissed):

ixrock:
@Nokel81 PTALA

src/extensions/extension-discovery/is-compatible-extension/is-compatible-extension.ts
```diff
@@ -9,44 +9,14 @@ interface Dependencies {
...
-    if (manifest.engines?.lens) {
-      /**
-       * include Lens's prerelease tag in the matching so the extension's
-       * compatibility is not limited by it
-       */
-      return semver.satisfies(strippedVersion, manifest.engines.lens, {
-        includePrerelease: true,
-      });
-    }
+    const { major: extMajor, minor: extMinor } = new SemVer(manifest.engines.lens, {
```
Nokel81: It would be really good to have some tests for this code nonetheless. For instance what happens if an extension has the string `>=5.5.0`? I think this code might throw in that case.

src/extensions/extension-discovery/is-compatible-extension/is-compatible-extension.ts
```diff
@@ -9,44 +9,14 @@ interface Dependencies {
...
+      loose: true,
+    });
 
-    return false;
+    return semver.intersects(`>=${appMajor}.${appMinor}`, `<${extMajor}.${extMinor + 1}`);
```
Nokel81: So if the extension has the value of `lens: "5.2"` and the current version is `5.6.1` this would return `false`, right?

ixrock: _Yes:_
Screenshot 2022-05-05 at 18 44 41

But.. using raw version as `5.2` throws error and some version normalization needed before creating instance.
image

Nokel81: But we want `5.2` and `^5.2` to both be valid ways of specifying the range (and to mean the same thing). We want it to return `true` in that case.

Furthermore, we currently accept `>=5.2` and other such semver ranges. So we need to convert them as well.

ixrock: _Test in chrome's devtools:_

```ts
// test
var {major: appMajor, minor: appMinor} = new semver.SemVer("5.6.1");
var {major: extMajor, minor: extMinor} = new semver.SemVer("5.2.0");
semver.intersects(`>=${appMajor}.${appMinor}`, `<${extMajor}.${extMinor + 1}`)
```

```ts
// bootstrap.tsx
import semver from "semver";
export {semver};
```

src/extensions/extension-discovery/is-compatible-extension/is-compatible-extension.ts
ixrock: Right! Already noticed versions starting with `^` throws in semver constructor, so something need to be done before passing raw version from extension/app (normalize version to not throwable or maybe `Semver` has some built-in options).
Anyway I'll try to add some unit tests to check that different ext/app version range formats wont break this function.

Nokel81: Something else I just noticed while doing some testing myself. `"5.6"` is also invalid.

src/extensions/__tests__/is-compatible-extension.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    }));
+
+    expect(isCompatible).toBeTruthy();
+  });
+
+  it("supporting `manifest.engines.lens='*'` to match any base-app version", () => {
+    expect(isCompatibleExtension({
+      appSemVer: semver.coerce("1.0.0"),
+    })(getExtensionManifestMock({
+      lensEngine: "*",
```
Nokel81: I don't think we want to support `*`.

src/extensions/__tests__/is-compatible-extension.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    }))).toBeTruthy();
+  });
+
+  it("is compatible with higher patch-versions of main app", () => {
+    expect(isCompatibleExtension({
+      appSemVer: semver.coerce("5.5.5"), // for patch-versions
+    })(getExtensionManifestMock({
+      lensEngine: "^5.5.0",
+    }))).toBeTruthy();
+  });
```
Nokel81: Can we have a test to like this but to make sure that specifying `5.5.0` is still compatible with an app version of `5.6.0`?

src/extensions/__tests__/is-compatible-extension.test.ts
```diff
@@ -0,0 +1,65 @@
...
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import semver from "semver";
+import {
+  isCompatibleExtension,
+} from "../extension-discovery/is-compatible-extension/is-compatible-extension";
+import type { LensExtensionManifest } from "../lens-extension";
+
+describe("Extension/App versions compatibility check", () => {
```
Nokel81, line 12: I like we should force extensions to not be compatible if they try and use a non-`^` range, can we have a few tests to that affect?

src/extensions/__tests__/is-compatible-extension.test.ts
jim-docker: There should be some "is not compatible with..." tests too.

Nokel81: Probably, though I don't know how we should handle that. I looked at all the extensions I know of in the wild and they all use `^` but what should we do about extensions that use a different format? Just say that they aren't compatible? (probably I think).

src/extensions/__tests__/is-compatible-extension.test.ts
ixrock: Why not? If extension does really some simple things and doesn't really use lot of `@k8slens/extensions`-apis == rarely updated, so then using star-value could be preferred probably. And why to restrict that anyway? It's responsibility for authors of extensions to keep them in sync with latest supported manifest versions for _lens-extension_.

src/extensions/__tests__/is-compatible-extension.test.ts
ixrock: > try and use a non-^ range

do you mean this range format `>=2.1.0` should not be supported?

src/extensions/__tests__/is-compatible-extension.test.ts
Nokel81: The main reason is that once auto update is turned on we might auto update over to a major version and then this will still say "compatible"

src/extensions/__tests__/is-compatible-extension.test.ts
Nokel81: Yeah we don't really want an extension to be able to specify that they support multiple major versions at a time. Plus we don't want to support extensions saying `~` ranges

Nokel81 (dismissed):

src/extensions/extension-discovery/is-compatible-extension/is-compatible-extension.ts
```diff
@@ -12,7 +12,24 @@ interface Dependencies {
...
   return (manifest: LensExtensionManifest): boolean => {
     const { raw: appVersion } = appSemVer;
-    const supportedVersionsByExtension: string = semver.validRange(manifest.engines.lens);
+    const manifestLensEngine = manifest.engines.lens;
+    const validVersion = manifestLensEngine.match(/^[\^0-9]\d*\.\d+\b/); // must start from ^ or number
+
+    if (!validVersion) {
+      const errorInfo = [
+        `Invalid format for "manifest.engines.lens"="${manifestLensEngine}"`,
+        `Range versions could be specified only starting with '^'.`,
```
jim-docker:
```suggestion
        `Range versions can only be specified starting with '^'.`,
```

jim-docker (commented):
Something fishy:
Screen Shot 2022-05-09 at 4 42 53 PM

ran `make dev` on this branch, they are all compatible on master branch. helloworld-sample has 
```
  "engines": {
    "lens": "^5.1.0"
  },
```

I think it is the prerelease tag gumming things up (again). Lens `master` has `"version": "5.5.0-alpha.0",` If I change it to `"version": "5.5.0",` it works.

ixrock:
Couple of local tests seems to works as expected (even with lens-version with `-alpha.0`):

Screenshot 2022-05-10 at 13 49 16

Screenshot 2022-05-10 at 13 48 58

Nokel81:
I found a test case that fails @ixrock :

```ts
  it("is compatible with lensEngine with prerelease", () => {
    expect(isCompatibleExtension({
      appSemVer: semver.parse("5.5.0-alpha.0"),
    })(getExtensionManifestMock({
      lensEngine: "^5.4.0-alpha.0",
    }))).toBeTruthy();
  });
```

ixrock:
> I found a test case that fails @ixrock :
> 
> ```ts
>   it("is compatible with lensEngine with prerelease", () => {
>     expect(isCompatibleExtension({
>       appSemVer: semver.parse("5.5.0-alpha.0"),
>     })(getExtensionManifestMock({
>       lensEngine: "^5.4.0-alpha.0",
>     }))).toBeTruthy();
>   });
> ```

You should use `appSemVer: semver.coerce("5.5.0-alpha.0")` for creating new `SemVer` instance:

image

Nokel81:
> You should use `appSemVer: semver.coerce("5.5.0-alpha.0")` for creating new SemVer instance:

No, because that is not what we do to produce the actual `appSemVer`, we use `new Semver()` (which is equivalent to `semver.parse`. I think that `semver.coerce` drops the `prerelease` sections.

Nokel81:
Also, I believe that the test case I provided covers what @jim-docker saw (which I could reproduce as well).

ixrock:
> > You should use `appSemVer: semver.coerce("5.5.0-alpha.0")` for creating new SemVer instance:
> 
> No, because that is not what we do to produce the actual `appSemVer`, we use `new Semver()` (which is equivalent to `semver.parse`. I think that `semver.coerce` drops the `prerelease` sections.

To support various formats for `isCompatibleExtension()`-check we have to use `semver.coerce` since for example `semver.parse("5.2")` produces `null` (but it's allowed for extension to specify such values for `engines.lens`)

Nokel81:
Yes I know, but this `parse` is only for the `appSemVer` not the version that the extension provides. I agree that `coerce` is the better choice there.

ixrock:
@Nokel81 now your test-case should work fine. PTALA.

Nokel81 (dismissed):
Checked the extensions page to verify that this works as expected.

Nokel81 (approved):

integration/__tests__/cluster-pages.tests.ts
```diff
@@ -39,7 +39,8 @@ utils.describeIf(minikubeReady(TEST_NAMESPACE))("Minikube based tests", () => {
     await frame.waitForSelector(`.Menu >> text="Remove"`);
   });
 
-  it("opens cluster settings by clicking link in no-metrics area", async () => {
+  // FIXME: failed locally since metrics might already exist, cc @aleksfront
```
jansav, line 42: Should be fixed?

Nokel81: Yes probably, but I don't think so in this PR. I think it would be better if this was a behavioural test instead of an e2e test.

src/extensions/__tests__/is-compatible-extension.test.ts
```diff
@@ -0,0 +1,121 @@
...
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import semver from "semver";
+import {
+  isCompatibleExtension,
+} from "../extension-discovery/is-compatible-extension/is-compatible-extension";
+import type { LensExtensionManifest } from "../lens-extension";
+
```
jansav, line 11: Multiple `expect` in single test prevents accurate name of test.

src/extensions/__tests__/is-compatible-extension.test.ts
```diff
@@ -0,0 +1,121 @@
...
+describe("Extension/App versions compatibility check", () => {
+  it("is compatible with exact version matching", () => {
+    expect(isCompatibleExtension({
+      appSemVer: semver.coerce("5.5.0"),
+    })(getExtensionManifestMock({
+      lensEngine: "5.5.0",
+    }))).toBeTruthy();
```
jansav, line 18: Most tests duplicate this test logic. How about:

`expect(versionsAreCompatible("5.5.0", "5.6.0")).toBe(false)`;

Nokel81: I would prefer an object as the params. I nice thing about the current tests is that I can easily read the test to figure out which version string means which.

jim-docker (commented):
I noticed that if something like `"~5.5.0"` or `">=5.5.0"` is specified by my extension's `engine.lens` it is rightly rejected as invalid  but my extension no longer shows up in the installed extensions list on the Extensions page. Shouldn't it still be there, just marked as incompatible?

jim-docker:
> I noticed that if something like `"~5.5.0"` or `">=5.5.0"` is specified by my extension's `engine.lens` it is rightly rejected as invalid but my extension no longer shows up in the installed extensions list on the Extensions page. Shouldn't it still be there, just marked as incompatible?

I guess it's a developer issue, as long as they see the error in the logs, and/or are made aware of the new range spec restriction in the "what's new for extensions" docs they can correct the issue before they "ship" their extension. Might impact existing extensions though.

jim-docker (approved):

Nokel81:
> I guess it's a developer issue, as long as they see the error in the logs, and/or are made aware of the new range spec restriction in the "what's new for extensions" docs they can correct the issue before they "ship" their extension. Might impact existing extensions though.

I think we can change this in the future, even before 5.6 is released if we want.

src/extensions/__tests__/is-compatible-extension.test.ts
jansav: Then more readable solution would be

`expect(versionsAreCompatible({ versionOfExtension: "5.5.0",  versionOfLens: "5.6.0" })).toBe(false);`

ixrock: @jansav sounds good. I'll create follow-up with that test update.
But we compare not extension's version ( `versionOfExtension`), but it's `engines.lens` version.

PR #5309, opened 2022-05-02
Fix last item not being removed on initial loadAll

### LINKED ISSUE #4897: Resource removal not always detected

Nokel81:
- Add a unit test to cover this case

Fixes https://github.com/lensapp/lens/issues/4897

close https://github.com/lensapp/lens/pull/5234

### DIFF
--- a/src/common/k8s-api/kube-object.store.ts
+++ b/src/common/k8s-api/kube-object.store.ts
@@ -60,6 +60,14 @@ export interface KubeObjectStoreSubscribeParams {
   abortController?: AbortController;
 }
 
+export interface MergeItemsOptions {
+  merge?: boolean;
+  updateStore?: boolean;
+  sort?: boolean;
+  filter?: boolean;
+  namespaces: string[];
+}
+
 export abstract class KubeObjectStore<T extends KubeObject> extends ItemStore<T> {
   static defaultContext = observable.box<ClusterContext>(); // TODO: support multiple cluster contexts
 
@@ -227,7 +235,7 @@ export abstract class KubeObjectStore<T extends KubeObject> extends ItemStore<T>
   }
 
   @action
-  async loadAll({ namespaces, merge = true, reqInit, onLoadFailure }: KubeObjectStoreLoadAllParams = {}): Promise<void | T[]> {
+  async loadAll({ namespaces, merge = true, reqInit, onLoadFailure }: KubeObjectStoreLoadAllParams = {}): Promise<undefined | T[]> {
     await this.contextReady;
     namespaces ??= this.context.contextNamespaces;
     this.isLoading = true;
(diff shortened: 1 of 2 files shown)


src/common/k8s-api/kube-object.store.ts
```diff
@@ -227,15 +235,15 @@ export abstract class KubeObjectStore<T extends KubeObject> extends ItemStore<T>
...
+  async loadAll({ namespaces, merge = true, reqInit, onLoadFailure }: KubeObjectStoreLoadAllParams = {}): Promise<undefined | T[]> {
     await this.contextReady;
     namespaces ??= this.context.contextNamespaces;
     this.isLoading = true;
 
     try {
       const items = await this.loadItems({ namespaces, reqInit, onLoadFailure });
 
-      this.mergeItems(items, { merge });
+      this.mergeItems(items, { merge, namespaces });
```
jakolehm, line 246: So the fix is to pass all requested namespaces to `mergeItems` (`items` might not contain all if a namespace is empty)?

jakolehm (dismissed):

src/common/k8s-api/kube-object.store.ts
Nokel81: Yes that is the fix.

src/common/k8s-api/kube-object.store.ts
```diff
@@ -248,29 +256,31 @@ export abstract class KubeObjectStore<T extends KubeObject> extends ItemStore<T>
...
 
   @action
-  protected mergeItems(partialItems: T[], { merge = true, updateStore = true, sort = true, filter = true } = {}): T[] {
+  protected mergeItems(partialItems: T[], { merge = true, updateStore = true, sort = true, filter = true, namespaces }: MergeItemsOptions): T[] {
     let items = partialItems;
 
     // update existing items
-    if (merge) {
-      const namespaces = partialItems.map(item => item.getNs());
+    if (merge && this.api.isNamespaced) {
```
Nokel81, line 279: This check is here because we shouldn't even attempt to merge cluster scoped items.

jakolehm (approved):

jim-docker (approved):

PR #5195, opened 2022-04-05
Turn on strict mode in tsconfig.json, some helpful lints, and required cleanup where strictness necessitates it

Nokel81:
- Turn on the following new lints:
  - react/jsx-first-prop-new-line
  - react/jsx-wrap-multilines
  - react/jsx-one-expression-per-line
  - react/jsx-max-props-per-line
  - react/jsx-indent
  - react/jsx-indent-props
  - no-restricted-imports

All this work is done so that we can have greater confidence in our codebase. The main part of `strict: true` is treating `null` and `undefined` as separate and non-assignable types to all the other types. This will help catch bugs at compile time.

Signed-off-by: Sebastian Malton <sebastian@malton.name>

PR #5317, opened 2022-05-03
Change tray icon when update is available to be applied

### LINKED ISSUE #1909: Change tray icon when update is available.
When the app-updater finds that updating the app is possible, the tray icon should be updated to show some notification. Similar to the docker tray icon:

Screen Shot 2021-01-07 at 8 53 41 AM

**Why is this needed**:
A nice to have from a users perspective.

Nokel81:

- make initTray injectable
- Add computational support for a notice on the tray icon
- Rework the image updating injectables to use the updater state
  - Move cache up a level and add a new option to createTryIcon
  - Make useDarkColors and updateAvailable fully observable, computed, and injectable
  - Add locking to remove race conditions within async code

fixes #1909 

screenshot:
Screen Shot 2022-05-03 at 12 24 58 PM


src/common/utils/hash-map.ts
```diff
@@ -0,0 +1,104 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+function makeIterableIterator<T>(iterator: Iterator<T>): IterableIterator<T> {
```
aleksfront, line 6: Can you describe why we need this hash map methods over the native `Map`? Seems like it reintroduces same methods js `Map` has

src/main/electron/use-dark-colors.injectable.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const useDarkColorsInjectable = getInjectable({
+  id: "use-dark-colors",
+  instantiate: (di) => {
+    const nativeTheme = di.inject(nativeThemeInjectable);
+    const state = observable.box(nativeTheme.shouldUseDarkColors);
+
+    nativeTheme.on("updated", () => state.set(nativeTheme.shouldUseDarkColors));
+
+    return state;
```
aleksfront, line 17: @jansav Doesn't look right to have side effect within `instantiate`. It should only return some result? Also, it breaks single responsibility principle.

src/main/app-updater/state.injectable.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { observable } from "mobx";
+
+export type AppUpdaterState = {
+  status: "idle";
+} | {
+  status: "update-install-ready";
+};
```
aleksfront, line 12:
```suggestion
export type AppUpdaterState = {
  status: "idle" | "update-install-ready";
};
```

aleksfront (commented):
What is the best way to test it in production? I've compiled lens ide from this branch `issue-1909`, but it seems to use latest version so it doesn't trigger 'update-available' signal, nor changes a tray icon.

src/main/electron/use-dark-colors.injectable.ts
Nokel81: How does it break SRP? Its single responsibility is to be an up to date, observable, version of `nativeTheme.shouldUseDarkColors`.

src/common/utils/hash-map.ts
Nokel81: `HashMap` is an abstraction over `Map` in that the keys are compared as equal if there hashes are equal. This is to get around the fact that JS doesn't yet have record types (https://github.com/tc39/proposal-record-tuple is only stage 2) so `Map` requires instance equality and not field equality for `get` and `set`

src/main/app-updater/state.injectable.ts
Nokel81: I used the style above to show that the desire is this will be a tagged union type. Where in the future there can be tags with extra data. For instance there could be a variant `"downloading-update"` which has an additional field `progress`.

src/main/tray/create-tray-icon.injectable.ts
```diff
@@ -0,0 +1,92 @@
...
+export type CreateTrayIcon = (args: CreateTrayIconArgs) => Promise<NativeImage>;
+
+const createTrayIconInjectable = getInjectable({
+  id: "create-tray-icon",
+  instantiate: (di): CreateTrayIcon => {
+    const produceTemplateImage = di.inject(isMacInjectable);
+
+    return async ({ useDarkColors, updateAvailable }) => {
+      const size = 32; // 2x zoom
+      const trayIconColor = useDarkColors ? "white" : "black"; // Invert to show contrast
```
aleksfront, line 29: Necessity of this line should be double checked because tray icon in latests macOS version should **always** be dark no matter what native theme is used.

For example, here is one of the installed apps changed tray color to white for no good reason.
dark tray icons

Nokel81: Yes I know, that is what `produceTemplateImage` is supposed to cover.

src/main/electron/use-dark-colors.injectable.ts
Iku-turso: Thoughts:

1. Something called `useDarkColorsInjectable`, should return a function. This returns a state, and therefore it should be called `darkColorsAreUsedStateInjectable`. However, this should probably not return state because...
2. Is it required to expose a way to set this state, if there already is an event (`nativeTheme.on("updated"`) which sets it? These settings of state will compete with each other causing confusion. As a fix, `darkColorsAreUsedInjectable` should be `IComputedValue<boolean>` instead of `IObservableValue<boolean>`.
3. As an improvement, if rigging of `nativeTheme.on("updated")` is yet made a separate responsibility (e.g. `startNativeThemeSyncInjectable`), we could also have another responsibility to clean it up (e.g. `stopNativeThemeSyncInjectable`), which currently does not get done at all. Important: #5324 already does this (see `operatingSystemThemeInjectable`), however the PR is not merged yet, because it is being blocked by #5195 (@Nokel81).

Nokel81: I'll clean this up after we merge #5324  then

Nokel81:
I will reopen this PR instead of trying to fix it up as it is going to be very different.

PR #5324, opened 2022-05-04
Make starting of application modular and unit testable

jansav:
# Highlights:

## 1. `runMany` and `runManySync`
- Runs all implementations of a runnable injection token
- Replacement of `setuppable` in `injectable`
- Granular and maintainable
- Reusable
- Exact in timing
- Orderable for temporal dependencies (see: `runAfter`)
- Permits parallelism for faster application start
- Compliant with Open Closed Principle (e.g. one can add new runnables for a token without modifying existing code)
- See technical unit tests for detailed documentation ([run-many-for.test.ts](https://github.com/lensapp/lens/blob/consolidate-running-of-setups-on-application-start/src/common/runnable/run-many-for.test.ts))

## 2. Current injection tokens for `runMany`:
- `beforeElectronIsReadyInjectionToken`
- `beforeApplicationIsLoadingInjectionToken`
- `onLoadOfApplicationInjectionToken`
- `afterApplicationIsLoadedInjectionToken`
- `afterWindowIsOpenedInjectionToken`
- `beforeFrameStartsInjectionToken`
- `afterRootFrameIsReadyInjectionToken`
- `beforeQuitOfFrontEndInjectionToken`
- `beforeQuitOfBackEndInjectionToken`

## 3. `startableStoppable`
- Reusable way to govern things that start and stop, like a disposer
- Synergizes well with `runMany` e.g. `catalogSync` is startable stoppable. It runs its start at `afterRootFrameIsReadyInjectionToken` and runs its stop at `beforeQuitOfFrontEndInjectionToken`
- See technical unit tests for detailed documentation ([get-startable-stoppable.test.ts](https://github.com/lensapp/lens/blob/consolidate-running-of-setups-on-application-start/src/common/utils/get-startable-stoppable.test.ts))

## 4. WindowManager is no more
- Windows in Lens are now instances instead of explicit
- see: `ApplicationWindow` and `SplashWindow`

## 5. `index.ts` for main is no longer bloated
- Responsibilities have been extracted to runnables
- Timing of said runnables is controlled is centrally in `startMainApplication`


src/main/index.ts
```diff
@@ -5,362 +5,19 @@
...
-  }
-
-  setTimeout(() => {
-    appEventBus.emit({ name: "service", action: "start" });
-  }, 1000);
-}
-
-main(getDi());
+(async () => {
+  await startApplication();
```
Nokel81, line 19: This await is useless. Why not just have injecting `startMainApplicationInjectable` do all the work immediately?

src/main/electron-app/features/get-electron-theme.injectable.ts
```diff
@@ -0,0 +1,18 @@
...
+import { getInjectable } from "@ogre-tools/injectable";
+import nativeThemeInjectable from "./native-theme.injectable";
+
+const getElectronThemeInjectable = getInjectable({
+  id: "get-electron-theme",
+
+  instantiate: (di) => {
+    const nativeTheme = di.inject(nativeThemeInjectable);
+
+    return () => nativeTheme.shouldUseDarkColors ? "dark" : "light";
```
Nokel81, line 14: This is not reactive, see https://github.com/lensapp/lens/pull/5317

src/main/electron-app/features/sync-theme-from-operating-system.injectable.ts
```diff
@@ -0,0 +1,35 @@
...
+  id: "sync-theme-from-operating-system",
+
+  instantiate: (di) => {
+    const currentThemeState = di.inject(operatingSystemThemeStateInjectable);
+    const nativeTheme = di.inject(nativeThemeInjectable);
+    const getElectronTheme = di.inject(getElectronThemeInjectable);
+
+    return getStartableStoppable("sync-theme-from-operating-system", () => {
+      const updateThemeState = () => {
+        const newTheme = getElectronTheme();
```
Nokel81, line 21: If this is the only use of `getElectronTheme` it seems strange to have it as an `injectable`

src/main/theme/operating-system-theme.injectable.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { computed } from "mobx";
+import operatingSystemThemeStateInjectable from "./operating-system-theme-state.injectable";
+
+const operatingSystemThemeInjectable = getInjectable({
```
Nokel81, line 9: What is the benefit of this injectable? `currentThemeStateInjectable` already has ` .get()` method.

src/main/theme/operating-system-theme-state.injectable.ts
```diff
@@ -0,0 +1,24 @@
...
+export type Theme = "dark" | "light";
+
+const operatingSystemThemeStateInjectable = getInjectable({
+  id: "operating-system-theme-state",
+
+  instantiate: (di) => {
+    const getElectronTheme = di.inject(getElectronThemeInjectable);
+    const defaultTheme = getElectronTheme();
+
+    return observable.box<Theme>(
```
Nokel81, line 18: We need to standardize formatting ASAP after auto update, I find this hard to read and `prettier` is not the format that the repo tends to use so we probably shouldn't be using it at all.

src/main/theme/sync-theme-from-os/stop-syncing-theme-from-operating-system.injectable.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const stopSyncingThemeFromOperatingSystemInjectable = getInjectable({
+  id: "stop-syncing-theme-from-operating-system",
+
+  instantiate: (di) => {
+    const syncTheme = di.inject(syncThemeFromOperatingSystemInjectable);
+
+    return {
+      run: async () => {
+        await syncTheme.stop();
```
Nokel81, line 17: Again, why with the wrapping? Style?

src/common/utils/get-startable-stoppable.ts
```diff
@@ -0,0 +1,52 @@
...
+    start: async () => {
+      if (started) {
+        throw new Error(`Tried to start "${id}", but it has already started.`);
+      }
+
+      startingPromise = startAndGetStopCallback();
+      stop = await startingPromise;
+
+      stopped = false;
+      started = true;
```
Nokel81, line 39: This has a race condition, calling `start()` twice before `startAndGetStopCallback` resolves means that it is called twice.

src/common/utils/environment-variables.injectable.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+
+const environmentVariablesInjectable = getInjectable({
+  id: "environment-variables",
+
+  instantiate: () => {
+    // IMPORTANT: The syntax needs to be exactly this in order to make environment variable values
```
Nokel81, line 11: By this comment do you mean that object destructuring won't work?

src/common/test-utils/flush-promises.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+export const flushPromises = () => new Promise(setImmediate);
```
Nokel81, line 5: This function is going to need quite a bit of documentation.

src/common/runnable/run-many-for.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { pipeline } from "@ogre-tools/fp";
+import type {
+  DiContainerForInjection,
+  InjectionToken,
+} from "@ogre-tools/injectable";
+import { filter, map } from "lodash/fp";
+
+export interface Runnable<TParameter = void> {
```
Nokel81, line 13: Why does this type param need a default?

src/common/runnable/run-many-for.ts
```diff
@@ -0,0 +1,44 @@
...
+  run: (parameter: TParameter) => Promise<void> | void;
+  runAfter?: this;
+}
+
+export const runManyFor =
+  (di: DiContainerForInjection) =>
+  <TRunnable extends Runnable<unknown>>(
```
Nokel81: I think that this really should be generic over `TParameter` and then you won't need the `unknown` here or even the `Paramaters<>` call on line 22 bellow.

src/common/cluster/cluster.ts
```diff
@@ -33,6 +33,8 @@ export interface ClusterDependencies {
   createKubectl: (clusterVersion: string) => Kubectl;
   createAuthorizationReview: (config: KubeConfig) => CanI;
   createListNamespaces: (config: KubeConfig) => ListNamespaces;
+  detectorRegistry: DetectorRegistry;
```
Nokel81: nit: this should be marked as `readonly`

src/common/app-paths/app-paths.test.ts
```diff
@@ -40,30 +40,32 @@ describe("app-paths", () => {
...
-    mainDi.override(
-      setElectronAppPathInjectable,
-      () =>
-        (key: PathName, path: string): void => {
-          defaultAppPathsStub[key] = path;
-        },
-    );
-
-    mainDi.override(appNameInjectable, () => "some-app-name");
+    applicationBuilder.beforeSetups(({ mainDi }) => {
```
Nokel81: Should `beforeSetups`'s name be changed because `setup`'s aren't a thing anymore?

src/common/vars/is-linux.injectable.ts
```diff
@@ -3,11 +3,10 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
 import { getInjectable } from "@ogre-tools/injectable";
-import { isLinux } from "../vars";
 
 const isLinuxInjectable = getInjectable({
   id: "is-linux",
-  instantiate: () => isLinux,
+  instantiate: () => process.platform === "linux",
```
Nokel81: I think that `process.platform` should be an `injectable` itself so that these are consistent while in testing while only having to override a single injectable.

src/common/hotbar-store.ts
```diff
@@ -44,8 +44,8 @@ export class HotbarStore extends BaseStore<HotbarStoreModel> {
       },
       migrations,
     });
+
     makeObservable(this);
-    this.load();
```
Nokel81: Why was this removed? I cannot find where you guys readded this. So this is certainly a bug

src/extensions/lens-main-extension.ts
```diff
@@ -4,42 +4,55 @@
...
-   * 
+   *
    * @returns the modified shell environment that the terminal will be opened with. The implementation must return env as passed in, if it
    * does not modify the shell environment
    */
   terminalShellEnvModifier?: ShellEnvModifier;
 
   async navigate(pageId?: string, params?: Record<string, any>, frameId?: number) {
-    return WindowManager.getInstance().navigateExtension(this.id, pageId, params, frameId);
+    const di = getEnvironmentSpecificLegacyGlobalDiForExtensionApi(Environments.main);
```
Nokel81: Why do it this way instead of changing the dependencies?

src/main/app-paths/app-name/app-name.injectable.ts
```diff
@@ -3,11 +3,19 @@
...
   id: "app-name",
-  instantiate: (di) => di.inject(electronAppInjectable).name,
+
+  instantiate: (di) => {
+    const isDevelopment = di.inject(isDevelopmentInjectable);
+
+    return `${packageInfo.productName}${isDevelopment ? "Dev" : ""}`;
+  },
+
+  causesSideEffects: true,
```
Nokel81, line 18: Why is this marked as `causesSideEffects`? How is it causing any more side effects then those caused by `isDevelopmentInjectable`?

src/main/catalog-sources/weblinks.ts
```diff
@@ -28,9 +28,12 @@ async function validateLink(link: WebLink) {
   }
 }
 
+interface Dependencies {
+  weblinkStore: WeblinkStore;
+  catalogEntityRegistry: CatalogEntityRegistry;
+}
 
-export function syncWeblinks() {
-  const weblinkStore = WeblinkStore.getInstance();
```
Nokel81, line 33: Why not just make this fully injectable?

src/main/cluster-detectors/base-cluster-detector.ts
```diff
@@ -15,14 +15,14 @@ export interface ClusterDetectionResult {
 export class BaseClusterDetector {
   key: string;
 
-  constructor(public cluster: Cluster) {
+  constructor(public cluster: Cluster, private _k8sRequest: K8sRequest) {
```
Nokel81: Why not just have this as a dependency? We haven't exported these at all anywhere so we can break the constructor all we like.

src/main/cluster-detectors/detector-registry.ts
```diff
@@ -5,11 +5,17 @@
...
 import type { Cluster } from "../../common/cluster/cluster";
 import type { BaseClusterDetector, ClusterDetectionResult } from "./base-cluster-detector";
+import type { K8sRequest } from "../k8s-request.injectable";
+
+interface Dependencies {
+  k8sRequest: K8sRequest;
+}
+
+export class DetectorRegistry {
+  constructor(private dependencies: Dependencies) {}
```
Nokel81, line 19: nit: dependencies should always be marked as `readonly` and basically always be `protected` and not `private` (if not for consistency but also so that child classes can extend the dependencies if need be).

src/main/cluster-detectors/create-version-detector.injectable.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { VersionDetector } from "./version-detector";
+import k8sRequestInjectable from "../k8s-request.injectable";
+import type { Cluster } from "../../common/cluster/cluster";
+
+const createVersionDetectorInjectable = getInjectable({
```
Nokel81, line 10: Wouldn't it be better to just request that the registry detects a specific kind of metadata?

src/main/context-handler/context-handler.ts
```diff
@@ -30,6 +29,7 @@ interface PrometheusServicePreferences {
 interface Dependencies {
   createKubeAuthProxy: CreateKubeAuthProxy;
   authProxyCa: string;
+  prometheusProviderRegistry: PrometheusProviderRegistry;
```
Nokel81: nit: these should be `readonly`

src/main/electron-app/runnables/setup-ipc-main-handlers/setup-ipc-main-handlers.injectable.ts
```diff
@@ -0,0 +1,54 @@
...
+import loggerInjectable from "../../../../common/logger.injectable";
+import clusterManagerInjectable from "../../../cluster-manager.injectable";
+import applicationMenuItemsInjectable from "../../../menu/application-menu-items.injectable";
+import getAbsolutePathInjectable from "../../../../common/path/get-absolute-path.injectable";
+import catalogEntityRegistryInjectable from "../../../catalog/catalog-entity-registry.injectable";
+import clusterStoreInjectable from "../../../../common/cluster-store/cluster-store.injectable";
+import { whenApplicationIsLoadingInjectionToken } from "../../../start-main-application/runnable-tokens/when-application-is-loading-injection-token";
+import operatingSystemThemeInjectable from "../../../theme/operating-system-theme.injectable";
+
+const setupIpcMainHandlersInjectable = getInjectable({
```
Nokel81, line 18: Shouldn't this be marked as `causesSideEffects` because it accesses global shared state?

src/main/electron-app/runnables/clean-up-deep-linking.injectable.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { beforeQuitOfBackEndInjectionToken } from "../../start-main-application/runnable-tokens/before-quit-of-back-end-injection-token";
+import lensProtocolRouterMainInjectable from "../../protocol-handler/lens-protocol-router-main/lens-protocol-router-main.injectable";
+
+const cleanUpDeepLinkingInjectable = getInjectable({
```
Nokel81, line 9: why do you call it "deep linking"?

src/main/start-main-application/lens-window/application-window/lens-window-injection-token.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectionToken } from "@ogre-tools/injectable";
+import type { ClusterFrameInfo } from "../../../../common/cluster-frames";
+
+export interface SendToViewArgs {
+  channel: string;
+  frameInfo?: ClusterFrameInfo;
+  data?: any[];
```
Nokel81: nit: this should be `unknown[]` as we should be trying to move away from using `any` as much as possible.

src/main/start-main-application/lens-window/application-window/lens-window-injection-token.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export interface LensWindow {
+  show: () => Promise<void>;
+  close: () => void;
+  send: (args: SendToViewArgs) => Promise<void>;
+  visible: boolean;
```
Nokel81, line 18: nit: this should be `readonly`

src/main/start-main-application/lens-window/hide-all-windows/close-all-windows.injectable.ts
```diff
@@ -0,0 +1,20 @@
...
+import { getInjectable } from "@ogre-tools/injectable";
+import { lensWindowInjectionToken } from "../application-window/lens-window-injection-token";
+
+const closeAllWindowsInjectable = getInjectable({
+  id: "close-all-windows",
+
+  instantiate: (di) => () => {
+    const lensWindows = di.injectMany(lensWindowInjectionToken);
+
+    lensWindows.forEach((lensWindow) => {
```
Nokel81, line 14: question: Why the preference of this over `for...of`?

src/main/electron-app/runnables/setup-update-checking.injectable.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const setupUpdateCheckingInjectable = getInjectable({
+  id: "setup-update-checking",
+
+  instantiate: (di) => {
+    const isAutoUpdateEnabled = di.inject(isAutoUpdateEnabledInjectable);
+
+    return {
+      run: () => {
+        startUpdateChecking(isAutoUpdateEnabled)();
```
Nokel81: Shouldn't `startUpdateChecking` have `isAutoUpdateEnabled` as its own dependency?

src/main/start-main-application/runnables/flag-renderer/flag-renderer-as-loaded.injectable.ts
```diff
@@ -0,0 +1,29 @@
...
+const flagRendererAsLoadedInjectable = getInjectable({
+  id: "flag-renderer-as-loaded",
+
+  instantiate: (di) => {
+    const lensProtocolRouterMain = di.inject(lensProtocolRouterMainInjectable);
+
+    return {
+      run: () => {
+        runInAction(() => {
+          // Todo: remove this kludge which enables out-of-place temporal dependency.
```
Nokel81, line 19: What would you recommend instead?

src/main/start-main-application/runnables/emit-close-to-command-bus.injectable.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import appEventBusInjectable from "../../../common/app-event-bus/app-event-bus.injectable";
+import { beforeQuitOfFrontEndInjectionToken } from "../runnable-tokens/before-quit-of-front-end-injection-token";
+
+const emitCloseToCommandBusInjectable = getInjectable({
```
Nokel81: do you mean "event bus"?

src/main/start-main-application/start-main-application.injectable.ts
```diff
@@ -0,0 +1,76 @@
...
+
+      await waitForElectronToBeReady();
+
+      await beforeApplicationIsLoading();
+
+      if (!shouldStartHidden) {
+        await applicationIsLoadingWindow.show();
+      }
+
+      await whenApplicationIsLoading();
```
Nokel81: I found this name to be confusing since it was non-obvious that adding something to `applicationIsLoading` causes the application to take longer while loading.

I think we should change the nominclature a bit.

src/main/start-main-application/start-main-application.injectable.ts
```diff
@@ -0,0 +1,76 @@
...
+
+      if (!shouldStartHidden) {
+        const deepLinkUrl = getDeepLinkUrl(commandLineArguments);
+
+        if (deepLinkUrl) {
```
Nokel81, line 59: still don't think deep linking is the correct term here either.

src/main/start-main-application/runnables/setup-hotbar-store.injectable.ts
```diff
@@ -0,0 +1,26 @@
...
+import { whenApplicationIsLoadingInjectionToken } from "../runnable-tokens/when-application-is-loading-injection-token";
+
+const setupHotbarStoreInjectable = getInjectable({
+  id: "setup-hotbar-store",
+
+  instantiate: (di) => ({
+    run: () => {
+      const hotbarStore = di.inject(hotbarStoreInjectable);
+
+      hotbarStore.load();
```
Nokel81, line 17: Ah now I see, ignore my above comment.

src/main/routes/metrics/get-metric-providers-route.injectable.ts
```diff
@@ -7,26 +7,34 @@ import { getInjectable } from "@ogre-tools/injectable";
...
-      for (const { name, id, isConfigurable } of PrometheusProviderRegistry.getInstance().providers.values()) {
-        providers.push({ name, id, isConfigurable });
-      }
+      handler: () => {
+        const providers: MetricProviderInfo[] = [];
 
-      return { response: providers };
-    },
-  }),
+        for (const {
```
Nokel81: Why all the new lines?

src/main/protocol-handler/lens-protocol-router-main/lens-protocol-router-main.ts
```diff
@@ -40,11 +39,15 @@ function checkHost<Query>(url: URLParse<Query>): boolean {
...
   extensionLoader: ExtensionLoader;
   extensionsStore: ExtensionsStore;
+  showApplicationWindow: () => Promise<void>;
 }
 
 export class LensProtocolRouterMain extends proto.LensProtocolRouter {
   private missingExtensionHandlers: FallbackHandler[] = [];
 
+  /**
+   * @deprecated This is used to solve out-of-place temporal dependency. Remove, and solve otherwise.
```
Nokel81: I don't think this should be marked as deprecated if it is merely considered bad. The needs to be a replacement immediately usable when something is deprecated.

src/main/protocol-handler/lens-protocol-router-main/open-deep-link-for-url/open-deep-link.injectable.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import lensProtocolRouterMainInjectable from "../lens-protocol-router-main.injectable";
+
+const openDeepLinkInjectable = getInjectable({
```
Nokel81, line 8: Still don't think this nomincature is correct.

src/main/prometheus/prometheus-provider-registry.injectable.ts
```diff
@@ -0,0 +1,13 @@
+/**
```
Nokel81, line 1: Why the really long file name? When reading the whole file path I trip up because the word "prometheus" is repeated twice.

src/main/menu/show-about.injectable.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { showAbout } from "./menu";
+import showMessagePopupInjectable from "../electron-app/features/show-message-popup.injectable";
+
+const showAboutInjectable = getInjectable({
+  id: "show-about",
+
+  instantiate: (di) =>
```
Nokel81, line 12: Wouldn't it be easier to read if it was the following:

```ts
instantiate: (di) => showAbout({
  showMessagePopup: di.inject(showMessagePopupInjectable),
}),
```

src/main/menu/application-menu.injectable.ts
```diff
@@ -0,0 +1,25 @@
...
+import applicationMenuItemsInjectable from "./application-menu-items.injectable";
+import { getStartableStoppable } from "../../common/utils/get-startable-stoppable";
+
+const applicationMenuInjectable = getInjectable({
+  id: "application-menu",
+
+  instantiate: (di) => {
+    const applicationMenuItems = di.inject(applicationMenuItemsInjectable);
+
+    return getStartableStoppable("build-of-application-menu", () =>
```
Nokel81, line 17: When would we want to stop building this?

src/main/kubeconfig-manager/create-kubeconfig-manager.injectable.ts
```diff
@@ -17,6 +18,7 @@ const createKubeconfigManagerInjectable = getInjectable({
   instantiate: (di) => {
     const dependencies = {
       directoryForTemp: di.inject(directoryForTempInjectable),
+      lensProxyPortNumberState: di.inject(lensProxyPortNumberStateInjectable),
```
Nokel81: haha, this is also very wordy. Wouldn't something like `lensProxyPort` by itself convey just enough information as well without being so long?

src/main/helm/__tests__/helm-service.test.ts
```diff
@@ -17,14 +17,11 @@ describe("Helm Service tests", () => {
...
-          { name: "stable", url: "stableurl" },
-          { name: "experiment", url: "experimenturl" },
-        ];
-      }),
-    });
+      repositories: jest.fn().mockImplementation(async () => [
+        { name: "stable", url: "stableurl" },
+        { name: "experiment", url: "experimenturl" },
+      ]),
+    } as any);
```
Nokel81, line 24: Why this cast?

Nokel81: Wouldn't it be better (and possible in this PR to make `HelmRepoManager` not `Singleton`)?

Nokel81: Which I see  you guys did anyway....

src/main/lens-proxy-port-number-state.injectable.ts
```diff
@@ -0,0 +1,36 @@
+/**
```
Nokel81: nit: I think having all the code concerning "lens-proxy" should be in their own folder.

src/main/k8s-request.injectable.ts
```diff
@@ -0,0 +1,37 @@
+/**
```
Nokel81, line 1: nit: this should probably be under `k8s` instead of out in the open.

src/main/app-updater.ts
```diff
@@ -44,9 +44,8 @@ autoUpdater.logger = {
 
 /**
  * starts the automatic update checking
- * @param interval milliseconds between interval to check on, defaults to 24h
  */
-export const startUpdateChecking = once(function (interval = 1000 * 60 * 60 * 24): void {
+export const startUpdateChecking = (isAutoUpdateEnabled : () => boolean) => once(function (interval = 1000 * 60 * 60 * 24): void {
```
Nokel81: Why is this `isAutoUpdateEnabled` a function?

src/main/tray/stop-tray.injectable.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import trayInjectable from "./tray.injectable";
+import { beforeQuitOfBackEndInjectionToken } from "../start-main-application/runnable-tokens/before-quit-of-back-end-injection-token";
+
+const stopTrayInjectable = getInjectable({
```
Nokel81: "stopping" the tray is counter intuitive. I think a better name would be `removeTray`

src/renderer/getDi.tsx
```diff
@@ -4,25 +4,25 @@
...
 
 export const getDi = () => {
-  const di = createContainer(
-    getRequireContextForRendererCode,
-    getRequireContextForCommonExtensionCode,
-    getRequireContextForCommonCode,
-  );
+  const di = createContainer();
+
+  registerMobX(di);
```
Nokel81: Can you explain what this does and why it is needed?

src/renderer/app-paths/setup-app-paths.injectable.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import type { AppPaths } from "../../common/app-paths/app-path-injection-token";
+import { appPathsIpcChannel } from "../../common/app-paths/app-path-injection-token";
+import getValueFromRegisteredChannelInjectable from "./get-value-from-registered-channel/get-value-from-registered-channel.injectable";
+import appPathsStateInjectable from "../../common/app-paths/app-paths-state.injectable";
+import { beforeFrameStartsInjectionToken } from "../before-frame-starts/before-frame-starts-injection-token";
+
+let syncAppPaths: AppPaths;
```
Nokel81: why have this state outside the injectable, I do not see the point.

src/renderer/components/+helm-charts/helm-charts.tsx
```diff
@@ -74,6 +74,8 @@ class NonInjectedHelmCharts extends Component<Dependencies> {
   render() {
     return (
       <SiblingsInTabLayout>
+        <div data-testid="page-for-helm-charts" style={{ display: "none" }}/>
```
Nokel81, line 82: This seems hacky. Why not have `ItemListLayout` or maybe `SiblingsInTabLayout` take an optional testid?

src/renderer/components/+preferences/extension-preference-item-registrator.injectable.ts
```diff
@@ -43,7 +43,7 @@ const extensionPreferenceItemRegistratorInjectable = getInjectable({
           }),
         );
 
-        injectables.forEach(di.register);
+        di.register(...injectables);
```
Nokel81, line 47: I still don't think that forcing these sorts of reactive temporary source items into a static like `injectMany` call is a good idea. As for the components not being the same, `computed` can mark specific child values as "not deeply observable"

src/renderer/components/layout/tab-layout.tsx
```diff
@@ -7,7 +7,7 @@ import "./tab-layout.scss";
 
 import type { ReactNode } from "react";
 import React from "react";
-import { matchPath, Redirect, Route, Switch } from "react-router";
+import { matchPath, Redirect, Route, Switch } from "react-router-dom";
```
Nokel81: What is the difference between the ones from `react-router` and the ones from `react-router-dom`? Why the switch?

Nokel81:
It seems most good.

src/common/utils/environment-variables.injectable.ts
jansav: Yes. Webpack doesn't replace values when they are destructured.

src/common/utils/get-startable-stoppable.ts
jansav: Good catch.

src/main/theme/sync-theme-from-os/stop-syncing-theme-from-operating-system.injectable.ts
jansav: Prettier + ESlint. We don't do any reformatting by hand. It's job for a computer. :)

src/main/theme/operating-system-theme-state.injectable.ts
jansav: Agreed that we need to do something for it. However I don't think there is better alternatives available than Prettier.

src/main/electron-app/features/sync-theme-from-operating-system.injectable.ts
jansav: Motivation is that Electron is the one we override in unit test. That's how we make sure we override dependencies in architectural boundaries and not before. That's to keep most business logic in scope of behavioural tests.

src/main/electron-app/features/get-electron-theme.injectable.ts
jansav: Sure, but the injectable using it is. This is just a function that we override in behavioural unit test to keep reactive part working as is.

src/main/theme/operating-system-theme.injectable.ts
jansav: Because we shouldn't expose the `.set(` at all, since there is no way to actually set it. 

In future, we could introduce possibility to add "private" injectables which cannot be injected outside of given scope (e.g file) for robust encapsulation. This would remove requirement for these type of injectables. But that's for the future. :)

src/common/test-utils/flush-promises.ts
jansav: This is something that is very deep in the core of javascript. 

Promises are run in microtask queue, which means that sometimes we need to wait till "next tick". This is where flushPromises comes to play. 

Simple example would be:

`setTimeout(() => Promise.resolve(), 0)`

VS.

`Promise.resolve()`

src/common/app-paths/app-paths.test.ts
jansav: Good catch.

src/common/vars/is-linux.injectable.ts
jansav: Sure. We've been done that when we have multiple usages for e.g `process`.

src/common/hotbar-store.ts
jansav: As you noticed, it's moved to controlled place. This is because of the temporal dependencies between syncing general entities and hotbar store.

src/main/app-paths/app-name/app-name.injectable.ts
jansav: package.json. :)

src/main/cluster-detectors/base-cluster-detector.ts
jansav: This is part of codebase which we are not that familiar yet. Therefore we did only bare minimum to limit cascading of the PR.

src/main/electron-app/runnables/setup-ipc-main-handlers/setup-ipc-main-handlers.injectable.ts
jansav: Yes. Good catch

src/main/electron-app/runnables/clean-up-deep-linking.injectable.ts
jansav: Because the feature it tries to achieve is commonly known as deep linking. 

https://www.electronjs.org/docs/latest/tutorial/launch-app-from-url-in-another-app

ProtocolRouterMain etc. is technical, deep linking is behaviour we want to do. Naming should always describe to behaviour if possible.

src/main/electron-app/runnables/setup-update-checking.injectable.ts
jansav: Absolutely.

src/main/start-main-application/runnables/flag-renderer/flag-renderer-as-loaded.injectable.ts
jansav: Topic for the future. We just wanted to highlight that the current implementation is not acceptable and we need to figure out better alternative way.

src/main/start-main-application/runnables/emit-close-to-command-bus.injectable.ts
jansav: Good catch.

src/main/start-main-application/start-main-application.injectable.ts
jansav: Any suggestions?

src/main/start-main-application/start-main-application.injectable.ts
jansav: Read earlier comment related to this. Let's keep the discussion regarding the term "deep linking" there.

src/main/protocol-handler/lens-protocol-router-main/open-deep-link-for-url/open-deep-link.injectable.ts
jansav: Same as above.

src/main/menu/application-menu.injectable.ts
jansav: Do we want to rely on electron not having memory leaks and actually removing all traces as it should? In perfect world, sure. 

Our code should always clean-up these type of things before we quit the application. 

This is exact same behaviour than before.

src/main/app-updater.ts
jansav: It was before too. If I remember correctly, it has some nasty side-effect. Agreed, it shouldn't be function

src/main/tray/stop-tray.injectable.ts
jansav: Agreed

src/renderer/getDi.tsx
jansav: This is extension for injectable which permits us to injectMany in a reactive way. No usages in this PR yet, but used e.g. to get rid of all `extensionInstallationCounter`

src/renderer/components/+helm-charts/helm-charts.tsx
jansav: Agreed, that's what we would have wanted to do. But seems bit too much since actual DOM is far in the structure of components. 

Would be great if any of our components took test-id as prop by default, so that it wouldn't feel a burden.

src/renderer/components/layout/tab-layout.tsx
jansav: This was suckerpunch from React update PR which shouldn't have anything do with this PR. 

If we mix `react-router` and `react-router-dom` they do not have shared instance of `history`. This worked before, but not anymore. 

Seems like a bug.

src/renderer/app-paths/setup-app-paths.injectable.ts
jansav: Good catch. Copy-paste from old implementation which required it.

src/common/vars/is-linux.injectable.ts
Nokel81: We already have multiple uses of `process`, including of `process.env`

src/main/app-paths/app-name/app-name.injectable.ts
Nokel81: Eh I don't think accessing the package json's produce name is within scope for unit tests to need to code around.

src/main/electron-app/runnables/clean-up-deep-linking.injectable.ts
Nokel81: Okay

src/common/vars/is-linux.injectable.ts
jansav: Sure. :)

src/main/start-main-application/start-main-application.injectable.ts
Nokel81: How about:

1. `beforeElectronIsReady` (no change)
2. `afterElectronIsReady`
4. `beforeMainWindowVisible`
5. `afterMainWindowVisibile`

src/extensions/lens-main-extension.ts
jansav: Bigger thing to reorganize it in a way that we have full control over constructor of these abstract classes. Easy way to introduce unwanted breaking changes. Could be done, but we decided to keep it out from this PR.

src/main/app-paths/app-name/app-name.injectable.ts
jansav: That doesn't make it free from side-effects still. There is no point not to globally override it. It only makes tests more stable without any downsides.

src/main/catalog-sources/weblinks.ts
jansav: `syncWebLinks` part done. WeblinkStore cascades a bit too much, so not including it to this PR.

src/main/cluster-detectors/detector-registry.ts
jansav: I'd like to argue that we shouldn't use inheritance as we have before. Composition over inheritance, right?

Also with keywords such as `private` and `protected` we should go with strictest one required and only change when something less strict is required case by case. This is for clean encapsulation.

src/main/cluster-detectors/create-version-detector.injectable.ts
jansav: This is part of codebase which we are not that familiar yet. Therefore we did only bare minimum to limit cascading of the PR.

src/main/start-main-application/lens-window/application-window/lens-window-injection-token.ts
jansav: Done

src/main/start-main-application/lens-window/hide-all-windows/close-all-windows.injectable.ts
jansav: `for...of` is something that you cannot compose. 

Imagine:


```typescript
for (const item of items) {
  doSomething(item);
}
```

vs.

```typescript
someItems.forEach(doSomething);
```


Now let's say we want to filter and sort items before we do it.

```typescript
for (const item of items.filter(x => x.length > 5).sort()) {
  doSomething(item);
}
```

vs.

```typescript
someItems.filter(x => x.length > 5).sort().forEach(doSomething);
```

or even

```typescript
pipeline(
  someItems,
  filter(x => x.length > 5),
  sortBy(identity),
  forEach(doSomething)
)
```

Which one leads to cleaner code? :)

src/main/routes/metrics/get-metric-providers-route.injectable.ts
jansav: Reverted

src/main/menu/show-about.injectable.ts
jansav: Same as above about code-style. TLDR: we don't do any formatting ourself. It's job for a computer.

src/main/app-updater.ts
jansav: `autoUpdater.isUpdaterActive()` is the underlying reason. At the moment we don't have idea if "active" status of updater can change, so easier to keep it as function for now. Of course real deal would be to make it reactive but that's out of scope here.

src/common/cluster/cluster.ts
jansav: Done

src/common/vars/is-linux.injectable.ts
jansav: There is now `platformInjectable` which is overridden for same effect. This allowed us to get rid of `causesSideEffects` in `isLinux`, `isWindows` and `isMac`.

Needs to said:

All `isLinux`, `isWindows`, `isMac` and such are breach of Open Closed Principle and thus live on borrowed time. This is topic for the future.

src/main/start-main-application/start-main-application.injectable.ts
jansav: We changed name of `whenApplicationIsLoading` to `onLoadOfApplication` to better indicate that adding stuff here adds to stuff that happens on load of application.

Also, we renamed `applicationIsLoadingWindow` to `splashWindow` to better distinguish it from `onLoadOfApplication` which has different responsibilities.

src/main/protocol-handler/lens-protocol-router-main/lens-protocol-router-main.ts
jansav: Changed to TODO.

src/main/prometheus/prometheus-provider-registry.injectable.ts
jansav: 1. Same naming as before, no changes here.
2. Preference over unique file names leads to better dev-experience because of discoverability.

src/main/kubeconfig-manager/create-kubeconfig-manager.injectable.ts
jansav: True, renamed.

src/main/lens-proxy-port-number-state.injectable.ts
jansav: Agreed, done.

src/main/tray/stop-tray.injectable.ts
jansav: Decided to use `installTray` and `uninstallTray`

src/renderer/getDi.tsx
jansav: Still, we removed this for being YAGNI.

src/renderer/components/+preferences/extension-preference-item-registrator.injectable.ts
jansav: Let's discuss this separately when the time comes to test waters with `computedInjectMany` to get rid of those nasty installation counters.

src/main/index.ts
jansav: > This await is useless

It is as so to make bigger deal about unawaited promise.


> Why not just have injecting startMainApplicationInjectable do all the work immediately?

It would be unexpected to have the act of inject do something operational while injected instance is irrelevant.

src/common/test-utils/flush-promises.ts
jansav: Check e.g. https://www.npmjs.com/package/flush-promises

src/common/runnable/run-many-for.ts
jansav: What would this improved code look like?

src/common/utils/get-startable-stoppable.ts
jansav: Fixed.

src/common/app-paths/app-paths.test.ts
jansav: Changed to `beforeApplicationStart`

jansav:
> It seems most good.

All comments addressed now.

src/extensions/lens-main-extension.ts
jansav: I can see that you have done the work required to get dependencies for these abstract classes. So doing this after branches have been merged.

src/common/runnable/run-many-for.test.ts
```diff
@@ -0,0 +1,225 @@
...
+
+      rootDi.register(someInjectable, someOtherInjectable);
+
+      const runMany = runManyFor(rootDi)(someInjectionTokenForRunnables);
+
+      actualPromise = runMany();
+    });
+
+    it("runs all runnables at the same time", () => {
+      expect(runMock).toHaveBeenCalledTimes(2);
```
jim-docker: could be the same runnable was called twice? This test would incorrectly pass?

src/common/runnable/run-many-for.test.ts
```diff
@@ -0,0 +1,225 @@
...
+      const someInjectionTokenForRunnables = getInjectionToken<Runnable>({
+        id: "some-injection-token",
+      });
+
+      const someInjectable1 = getInjectable({
+        id: "some-injectable-1",
+
+        instantiate: (di) => ({
+          run: () => runMock("third-level-run"),
+          runAfter: di.inject(someInjectable2),
```
jim-docker, line 83: what if `runAfter` is set to a runnable that doesn't match the injection token? Should there be a test for that (not sure this is handled)?

src/common/runnable/run-many-for.ts
Nokel81:
```ts
export type Run<Param> = (parameter: Param) => Promise<void>;
export type RunMany = <Param>(injectionToken: InjectionToken<Runnable<Param>, void>) => Run<Param>;

  return (injectionToken) => (
    async (param) => {
      ...
    }
  );
}
```

src/common/__tests__/cluster-store.test.ts
```diff
@@ -84,15 +81,14 @@ describe("cluster-store", () => {
...
+    mainDi.override(directoryForTempInjectable, () => "some-temp-directory");
 
     mainDi.permitSideEffects(getConfigurationFileModelInjectable);
     mainDi.permitSideEffects(appVersionInjectable);
+    mainDi.permitSideEffects(clusterStoreInjectable);
 
-    await mainDi.runSetups();
-
-    createCluster = mainDi.inject(createClusterInjectionToken);
+    mainDi.unoverride(clusterStoreInjectable);
```
jim-docker, line 92: `unoverride`?

jansav: `unoverride` is mechanism that is used when something is globally overridden, but then we want to write technical unit test for it. Most commonly used when writing technical unit test for implementation of an abstraction. 

This case it's used to allow this test to inject `clusterStore` even thought it is globally overridden from all other unit tests. `clusterStore` is overridden globally for now having side-effects.

src/common/runnable/run-many-for.test.ts
jansav: Good. Changed test to

```typescript
it("runs all runnables at the same time", () => {
  expect(runMock.mock.calls).toEqual([
    ["some-call"],
    ["some-other-call"],
  ]);
});
```

src/common/runnable/run-many-for.test.ts
jansav: Excellent catch. Added tests for this case. It will throw with error message informing you that it needs to be fixed.

src/common/runnable/run-many-for.ts
jansav: Most of the runnables are called without parameters.

src/extensions/lens-main-extension.ts
jansav: Changed to use dependencies.

Iku-turso (approved):

Iku-turso (approved):

Nokel81 (approved):

src/common/__tests__/cluster-store.test.ts
jim-docker: seemed like a code smell. So for unit tests we globally override everything then unoverride what we're trying to test?

Issue #5333, opened 2022-05-05
kubectl check fails on 1.24 clusters


kaspars:
**Describe the bug**
When connecting to a cluster running v1.24, Lens tries to use the corresponding version of kubectl, but the version check `kubectl version --client true --output json` fails with `error: extra arguments: [true]`. Thus you can't open shell on pod, and do other things that depend on kubectl.

**To Reproduce**
1. Start Lens from shell, to see the errors
2. Select a kubernetes cluster, running v1.24
3. Open Workloads->Pods, try to open Shell on any pod
4. Nothing happens
5. Inspect the error from Lens output

**Expected behavior**
If you remove `true` from `kubectl version --client true --output json`, the check will work.

**Screenshots**

**Environment (please complete the following information):**
- Lens Version: 5.4.6-latest.20220428.1
- OS: OSX

**Logs:**
When you run the application executable from command line you will see some logging output. Please paste them here:
```
info:    ▪ Downloading kubectl 1.24.0 from https://storage.googleapis.com/kubernetes-release/release/v1.24.0/bin/darwin/arm64/kubectl to /Users/kaspars/Library/Application Support/Lens/binaries/kubectl/1.24.0/kubectl +1s
error:   ▪ Local kubectl failed to run properly (Command failed: /Users/kaspars/Library/Application Support/Lens/binaries/kubectl/1.24.0/kubectl version --client true --output json
error: extra arguments: [true]
), unlinking +16s
```

kaspars:
Another thing which doesn't work with v1.24 clusters is "Master" resources (CPU and Memory) tab in "Cluster" section.

Upgrade to kubernetes v1.24.0 removes `node-role.kubernetes.io/master=` label from master nodes, and leaves only `node-role.kubernetes.io/control-plane=`. Lens should look for both of them to detect "Master" nodes.

PR #5334, opened 2022-05-05
"kubectl version --client" doesn't have extra values

kaspars:
`--client true` has been working with previous versions, but not with 1.24

### DIFF
--- a/src/main/kubectl/kubectl.ts
+++ b/src/main/kubectl/kubectl.ts
@@ -155,7 +155,7 @@ export class Kubectl {
       try {
         const args = [
           "version",
-          "--client", "true",
+          "--client",
           "--output", "json",
         ];
         const { stdout } = await promiseExecFile(path, args);


kaspars:
Related to this: https://github.com/lensapp/lens/issues/5333

Nokel81:
Please sign your commit.

src/main/kubectl/kubectl.ts
```diff
@@ -155,7 +155,7 @@ export class Kubectl {
       try {
         const args = [
           "version",
-          "--client", "true",
```
Nokel81, line 158: This won't work because we support versions of kubectl going back to 1.10

jim-docker: @Nokel81 did you look at the kubectl code for how options are handled? This change may be ok? it works on kube v1.16.15, though I don't know about 1.10. We may have been misinterpreting the kubectl docs on various options(?) Many have the default of `false`, which you get by not specifying the option. If you specify `--option` this means `true`.  Like for the set command example (https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#set):

```
kubectl set env pods --all --list
```
`all` and `list` are described the same way as `client` is for version. No need to specify `true` explicitly

Nokel81: Oh okay. I guess I misunderstood the docs since I didn't look at the code.

Nokel81 (approved):
Please sign your commit

src/main/kubectl/kubectl.ts
jim-docker: I tested with kubernetes 1.13.0 (oldest you can install with minikube) and it is fine. Looks like 1.24 is finickier so I think this PR is good and necessary moving forward.

jim-docker (approved):

jim-docker:
@kaspars can you sign this commit?

Nokel81:
Closing as this has been resolved with #5415

Nokel81:
Thanks for your contribution anyway

PR #5352, opened 2022-05-09
Revert upgrade to react 18

Nokel81:

This PR temporarily reverts the upgrade work to react 18 as it has caused a few too many issues with the extension API. Hopefully we can revert this reversion in the near future.


package.json
```diff
@@ -252,8 +252,8 @@
     "p-limit": "^3.1.0",
     "path-to-regexp": "^6.2.0",
     "proper-lockfile": "^4.1.2",
-    "react": "^18.0.0",
-    "react-dom": "^18.0.0",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
```
ixrock, line 256: I think having latest `react@18` with outdated types from `"@types/react-dom": "^17.0.14` quite acceptable.. unless we figure out/migrate to latest `@types/react-*`? 
I would keep latest `react` anyway since it should be backward compatible yet for some outdated react apis, e.g. `ReactDOM.render` should be still available.

ixrock (approved):

package.json
Nokel81: We found that both the lenscloud extension and the Kube resource map extension throw errors after this upgrade without trying to fix them.

jim-docker (approved):

PR #5354, opened 2022-05-09
Fix crash when changing os theme kind

### LINKED ISSUE #5171: Crash when switching OS theme

Nokel81:

fixes #5171

### DIFF
--- a/src/renderer/theme.store.ts
+++ b/src/renderer/theme.store.ts
@@ -36,7 +36,7 @@ export class ThemeStore extends Singleton {
     "lens-light": lensLightThemeJson as Theme,
   });
 
-  @observable osNativeThemeType: "dark" | "light" | undefined;
+  @observable private osNativeThemeType: "dark" | "light" | undefined;
 
   @computed private get colorThemePreference(): ThemeId | "system" {
     return UserStore.getInstance().colorTheme;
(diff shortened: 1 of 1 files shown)


src/renderer/theme.store.ts
```diff
@@ -113,25 +119,12 @@ export class ThemeStore extends Singleton {
     });
   }
 
-  bindNativeThemeUpdateEvent() {
-    ipcRenderer.on(setNativeThemeChannel, (event, theme: "dark" | "light") => {
-      this.osNativeTheme = theme;
-      this.applyTheme(theme);
```
Nokel81, line 119: This call was the main source of the crash because it treated `"dark"` as a `ThemeId`

Nokel81:
To reproduce the crash, delete your OpenLens's `lens-user-store.json` file and then start `make dev`, once it starts and the window is open go to `System Settings` and switch your theme preference. On `master` this will crash the application.

ixrock (approved):

aleksfront (approved):
LGTM

PR #5360, opened 2022-05-11
Fix <Switch/> checked callback attribute value

aleksfront:
Fixing `checked` value from `onChange` callback to be consistent what `event.target.checked` natively returns.

To avoid things like https://github.com/lensapp/lenscloud-lens-extension/pull/1654#discussion_r866658781

### DIFF
--- a/src/renderer/components/switch/switch.tsx
+++ b/src/renderer/components/switch/switch.tsx
@@ -17,7 +17,13 @@ export function Switch({ children, disabled, onChange, ...props }: SwitchProps) 
   return (
     <label className={cssNames(styles.Switch, { [styles.disabled]: disabled })} data-testid="switch">
       {children}
-      <input type="checkbox" role="switch" disabled={disabled} onChange={(event) => onChange?.(props.checked, event)} {...props}/>
+      <input
+        type="checkbox"
+        role="switch"
+        disabled={disabled}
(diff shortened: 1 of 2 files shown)


src/renderer/components/switch/switch.tsx
```diff
@@ -17,7 +17,13 @@ export function Switch({ children, disabled, onChange, ...props }: SwitchProps)
   return (
     <label className={cssNames(styles.Switch, { [styles.disabled]: disabled })} data-testid="switch">
       {children}
-      <input type="checkbox" role="switch" disabled={disabled} onChange={(event) => onChange?.(props.checked, event)} {...props}/>
+      <input
+        type="checkbox"
+        role="switch"
+        disabled={disabled}
+        onChange={(event) => onChange?.(!props.checked, event)}
```
ixrock: Why not?
```suggestion
        onChange={(event) => onChange?.(event.target.checked, event)}
```

Otherwise if you passing `props.checked` as first argument it's assuming `props.checked` must be required prop for `<Switch/>`? Also it means it's always gonna be only in controlled react component form? [read-more](https://reactjs.org/docs/uncontrolled-components.html)

Nokel81 (approved):

ixrock (approved):

PR #5361, opened 2022-05-11
auto-update notifications on the status bar

jim-docker:
moved some update notifications to status bar

https://user-images.githubusercontent.com/40840436/172835092-8f18cb43-d571-4914-b91c-cec7f5631cbd.mov


src/common/application-update/progress-of-update-download/progress-of-update-download.injectable.ts
```diff
@@ -3,10 +3,11 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
 import { getInjectable } from "@ogre-tools/injectable";
+import type { JsonObject } from "type-fest";
 import createSyncBoxInjectable from "../../utils/sync-box/create-sync-box.injectable";
 import { syncBoxInjectionToken } from "../../utils/sync-box/sync-box-injection-token";
 
-export interface ProgressOfDownload {
```
Nokel81, line 9: If you rebase with master this change isn't necessary anymore.

src/renderer/components/status-bar/status-bar.tsx
```diff
@@ -19,7 +20,9 @@ interface Dependencies {
 }
 
 const NonInjectedStatusBar = observer(({ items }: Dependencies & StatusBarProps) => {
-  const { left, right } = items.get();
+  const { left: leftItems, right } = items.get();
+
+  const left = [AutoUpdateComponent, ...leftItems];
```
Nokel81: This should be already part of the `items` dependency.

Nokel81 (commented):
Just some minor comments

src/renderer/components/status-bar/status-bar.tsx
jim-docker: You mean it should be inserted by `statusBarItemsInjectable`?

Nokel81: yes

jim-docker: ok. I guess if there were for a need for just the status bar items from extensions we could create another (base) injectable

Nokel81: We could yes

src/renderer/application-update/application-update-status-listener.injectable.ts
```diff
@@ -5,39 +5,46 @@
 import { getInjectable } from "@ogre-tools/injectable";
 import type { ApplicationUpdateStatusChannel, ApplicationUpdateStatusEventId } from "../../common/application-update/application-update-status-channel.injectable";
 import applicationUpdateStatusChannelInjectable from "../../common/application-update/application-update-status-channel.injectable";
-import showInfoNotificationInjectable from "../components/notifications/show-info-notification.injectable";
+//import showInfoNotificationInjectable from "../components/notifications/show-info-notification.injectable";
 import type { MessageChannelListener } from "../../common/utils/channel/message-channel-listener-injection-token";
 import { messageChannelListenerInjectionToken } from "../../common/utils/channel/message-channel-listener-injection-token";
+import AutoUpdateStateInjectable from "../../common/auto-update/auto-update-state.injectable";
 
 const applicationUpdateStatusListenerInjectable = getInjectable({
```
jansav: Hmm. I'm just wondering if we could get rid of this injectable since we are not doing any explicit actions anymore. 

The component (status-bar-item) could just inject e.g. `discoveredUpdateVersion` and do the things based on that directly. In that way our code would be even simpler.

src/renderer/components/status-bar/status-bar-items.injectable.tsx
```diff
@@ -21,13 +22,17 @@ interface Dependencies {
...
+function getStatusBarItems({ registrations }: Dependencies): IComputedValue<StatusBarItems> {
   return computed(() => {
-    const res: RegisteredStatusBarItems = {
+    const res: StatusBarItems = {
       left: [],
       right: [],
     };
 
+    // add Lens specific components
+    res.left.push(AutoUpdateComponent);
```
jansav: For the first item, sure. At least when next time emerges, we need to make it comply with Open Closed Principle so that we treat internal status bar items same as items belonging to extensions. I would definitely do it immediately though. 

Check these for example:
- `tray-menu-items.injectable.ts`
- `tray-menu-item-registrator.injectable.ts`
- `tray-menu-item-injection-token.ts`

jansav: @jim-docker For the examples I provided in earlier comment, notice that #5662 simplifies it even more.

src/renderer/components/status-bar/status-bar-item-registrator.injectable.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+      if (pos !== "left" && pos !== "right") {
+        throw new TypeError("StatusBarRegistration.components.position must be either 'right' or 'left'");
+      }
+
+      position = pos;
+
+      component = Item;
+    } else {
+      // throw?
```
Nokel81: I would return `[]`; here

src/renderer/components/status-bar/status-bar-item-registrator.injectable.tsx
```diff
@@ -0,0 +1,84 @@
...
+const toItemInjectableFor = (extension: LensRendererExtension, getRandomId: () => string) => {
+  const _toItemInjectable = () => (registration: StatusBarRegistration): Injectable<StatusBarItem, StatusBarItem, void>[] => {
+    const id = `${getRandomId()}-status-bar-item-for-extension-${extension.sanitizedExtensionId}`;
+    let component: React.ComponentType;
+    let position: "left" | "right";
+
+    if (registration.item) {
+      const { item } = registration;
+
+      // default for old API is "right"
```
Nokel81: I don't think that the implementation should need to know about this so I think specifying `position = "right"` here would be better than a comment.

src/renderer/components/status-bar/status-bar-item-registrator.injectable.tsx
```diff
@@ -0,0 +1,84 @@
...
+        position,
+        visible: computed(() => true),
+      }),
+
+      injectionToken: statusBarItemInjectionToken,
+    })];
+
+  };
+
+  return _toItemInjectable();
```
Nokel81: nit: it seems strange to have `_toItemInjectable` be a function here when all it does it return a function itself. Could line 34 just be `return (registration: StatusBarRegistration): Injectable<StatusBarItem, StatusBarItem, void>[] => {` instead?

Nokel81 (commented):
Looks very good.

The only comment about the UI is that the spinner before downloading is too big for the status bar.

src/renderer/components/status-bar/status-bar-item-registrator.injectable.tsx
jim-docker: Already done (as part of getting the tests to pass)

src/renderer/components/status-bar/status-bar-item-registrator.injectable.tsx
jim-docker: This was copy/paste but I have already added the explicit init of `position`. Will remove the comment.

src/renderer/components/status-bar/status-bar-item-registrator.injectable.tsx
```diff
@@ -0,0 +1,83 @@
...
+      const { position: pos = "right", Item } = registration.components;
+
+      if (pos !== "left" && pos !== "right") {
+        throw new TypeError("StatusBarRegistration.components.position must be either 'right' or 'left'");
+      }
+
+      position = pos;
+      component = Item;
+    } else {
+      logger.warn("StatusBarRegistration must have valid item or components field");
```
jansav, line 66: `logger` should be injected. It's overridden in unit tests so that logging are not present there at all being e.g. slow and noisy

src/renderer/components/status-bar/status-bar-items.injectable.tsx
```diff
@@ -0,0 +1,69 @@
...
+    for (const registration of registrations.get()) {
+      if (!registration || typeof registration !== "object") {
+        continue;
+      }
+
+      const { position = "right", component, visible } = registration;
+
+      if (position !== "left" && position !== "right") {
+        throw new TypeError("StatusBarRegistration.components.position must be either 'right' or 'left'");
+      }
```
jansav: This part seems to be there to serve Extension API? Our code is protected by typescript, so we don't need it. 

And since it's already present in `registrator`, you can just delete it here.

src/renderer/components/status-bar/status-bar-items.injectable.tsx
```diff
@@ -0,0 +1,69 @@
...
+  return computed(() => {
+    const res: StatusBarItems = {
+      left: [],
+      right: [],
+    };
+
+    for (const registration of registrations.get()) {
+      if (!registration || typeof registration !== "object") {
+        continue;
+      }
```
jansav: What would be case that registration does not exist or wouldn't be object? Stuff coming from extension api? If so, this belongs to `registrator`. In our code base we shouldn't need it, since the interface is enforced.

jansav: Also wouldn't it make sense to throw in this case? So that extension developer gets the feedback immediately.

src/renderer/components/status-bar/status-bar-items.injectable.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+      const { position = "right", component, visible } = registration;
+
+      if (position !== "left" && position !== "right") {
+        throw new TypeError("StatusBarRegistration.components.position must be either 'right' or 'left'");
+      }
+
+      if (!visible) {
```
jansav: Ain't visible `IComputedValue<boolean>`? So should it be `visible.get()`?

src/renderer/components/status-bar/status-bar.test.tsx
```diff
@@ -5,18 +5,16 @@
```
jansav, line 5: This test should be written as behaviour. Let's do this together today when you start your workday? :) Shouldn't be that hard to modify and I'll bet we'll catch some bugs while we're doing it.

src/renderer/components/status-bar/status-bar-items.injectable.tsx
```diff
@@ -0,0 +1,69 @@
...
+  left: React.ComponentType<StatusBarItemProps>[];
+}
+
+interface Dependencies {
+  registrations: IComputedValue<StatusBarItem[]>;
+}
+
+function getStatusBarItems({ registrations }: Dependencies): IComputedValue<StatusBarItems> {
+  return computed(() => {
+    const res: StatusBarItems = {
```
jansav, line 24: nit: would be easier to read if we utilized some functional programming here. :)

e.g.

```typescript
// This function is pure function, so it can be located outside of the getStatusBarItems, e.g. at bottom of file
const toVisibleComponents = (items: StatusBarItem[]) => pipeline(
  items,
  filter((item) => item.visible.get()),
  map((item) => item.component),
);

const result = pipeline(
  items.get(),
  partition((item) => item.position === "left"),
  map(toVisibleComponents),
  ([left, right]) => ({ left, right: reverse(right) }),
);
```

src/renderer/components/status-bar/status-bar-item-registrator.injectable.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  injectionToken: extensionRegistratorInjectionToken,
+});
+
+export default statusBarItemRegistratorInjectable;
+
+const toItemInjectableFor = (extension: LensRendererExtension, getRandomId: () => string) => {
+  return (registration: StatusBarRegistration): Injectable<StatusBarItem, StatusBarItem, void>[] => {
+    const id = `${getRandomId()}-status-bar-item-for-extension-${extension.sanitizedExtensionId}`;
+    let component: React.ComponentType;
```
jansav, line 39: nit: kinda hard to follow since we are using variables which are set in different timeslots. Could be simplified.

src/renderer/components/status-bar/auto-update-component.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  const { percentage, failed } = progressOfUpdateDownload.value.get();
+
+  if (downloadingUpdateState.value.get()) {
+
+    assert(discoveredVersion);
+
+    const roundedPercentage = Math.round(percentage);
+
+    return downloading(discoveredVersion.version, roundedPercentage);
```
jansav: nit: At least for me it's bit surprising that we are using functions here. Would be easier to follow if these were components. That would also give us possibility to add observers (and therefore not e.g. reload whole "auto-update-component" when only percentage changes.) But again, as long it works, I don't see any huge benefits here. I'm fine with it.

src/renderer/components/status-bar/auto-update-component.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  return DivWithTestId(`Download for version ${version} ${percentDone}%...`);
+};
+
+const downloadFailed = (errMsg: string) => DivWithTestId(errMsg);
+
+const idle = () => DivWithTestId("");
+
+
+export const NonInjectedAutoUpdateComponent = observer(({
```
jansav, line 67: Could name of component be something like `StatusOfApplicationUpdate`

src/renderer/components/status-bar/auto-update-component.tsx
```diff
@@ -0,0 +1,120 @@
...
+  });
+
+  if (idling) {
+    return idle();
+  }
+
+  return note(version ?? "");
+};
+
+const DivWithTestId = (text: string) => <div data-testid="auto-update-component">{text}</div>;
```
jansav: We should use different `test-id` for each case. In that way our unit test will be more strict and prevent me from breaking it accidentally.

src/renderer/components/status-bar/auto-update-component.tsx
```diff
@@ -0,0 +1,120 @@
...
+    {DivWithTestId("Checking for updates..." )}
+  </>
+);
+
+const available = (version: string) => DivWithTestId(`${version ?? "Update"} is available`);
+
+const notAvailable = () => DivWithTestId("No new updates available");
+
+const downloading = (version: string, percentDone: number) => {
+  if ( percentDone === 0 ) {
```
jansav: Is there a reason why we are not showing the spinner when downloading has just started? (percentage at 0)

src/behaviours/application-update/installing-update.test.ts
```diff
@@ -108,7 +103,7 @@ describe("installing update", () => {
         });
 
         it("notifies the user", () => {
-          expect(showInfoNotificationMock).toHaveBeenCalledWith("No new updates available");
+          expect(rendered.getByTestId("auto-update-component")).toHaveTextContent("No new updates available");
```
jansav: If we used different `test-id` for each of these, we wouldn't have to `toHaveTextContent`. We could just `toBeInTheDocument()` which would make this test less prone to fail for wrong reason (e.g. we are changing notification text)

src/behaviours/application-update/installing-update.test.ts
```diff
@@ -24,19 +23,17 @@ describe("installing update", () => {
   let checkForPlatformUpdatesMock: AsyncFnMock<CheckForPlatformUpdates>;
   let downloadPlatformUpdateMock: AsyncFnMock<DownloadPlatformUpdate>;
   let setUpdateOnQuitMock: jest.Mock;
-  let showInfoNotificationMock: jest.Mock;
 
   beforeEach(() => {
+    jest.useFakeTimers();
```
jansav, line 28: This doesn't seem to be used at all? Which tells me that the code related to idling is not tested at all. Let's take a look this together so we can add test for that case too.

jansav:
Let's have a huddle to discuss about the things I mentioned. Should be relative small stuff, so after that we can get this merged.

src/behaviours/application-update/installing-update.test.ts
```diff
@@ -113,6 +113,12 @@ describe("installing update", () => {
         it("renders", () => {
           expect(rendered.baseElement).toMatchSnapshot();
         });
+
+        it.skip("when 5 seconds elapses, clears the notification to the user", () => {
```
jim-docker, line 117: could not get this to work, maybe you can figure it out @jansav ? I tested manually and it does work (clears the message after 5 seconds)

src/renderer/components/status-bar/auto-update-component.tsx
```diff
@@ -28,51 +29,39 @@ interface EndNoteProps {
...
+  if (start + 5000 <= reactiveDateNow()) {
     return idle();
   }
-
+  
   return note(version ?? "");
-};
-
-const DivWithTestId = (text: string) => <div data-testid="auto-update-component">{text}</div>;
+});
```
jim-docker: modified this at @Nokel81's suggestion, in an attempt to get the behavioural test to work.

src/renderer/components/status-bar/auto-update-component.tsx
```diff
@@ -83,15 +72,13 @@ export const NonInjectedAutoUpdateComponent = observer(({
...
+  const { failed } = progressOfUpdateDownload.value.get();
 
   if (downloadingUpdateState.value.get()) {
 
     assert(discoveredVersion);
 
-    const roundedPercentage = Math.round(percentage);
-
-    return downloading(discoveredVersion.version, roundedPercentage);
+    return downloading(discoveredVersion.version);
```
jim-docker, line 81: no longer prints percentage.

jim-docker (commented):
Spinner is still too large, css magic anyone?

src/behaviours/application-update/installing-update.test.ts
jansav: It is actually working in isolation. I've have same trouble in my branch and this just gave me idea what is causing the trouble. It seems that `now()` from `mobx-utils` is using shared global state. I'm gonna double check and figure it out.

jansav: So I managed to create minimal unit test to prove the point:

```typescript
import { computed, observe } from "mobx";
import { now as reactiveNow } from "mobx-utils";

describe("reactiveNow", () => {
  let actual: boolean;

  beforeEach(() => {
    jest.useFakeTimers();

    jest.setSystemTime(new Date("2015-10-21T07:28:00Z"));

    const someComputed = computed(() => {
      const currentTimestamp = reactiveNow(1000);

      return currentTimestamp > new Date("2015-10-21T07:28:00Z").getTime();
    });

    observe(someComputed, (changed) => {
      actual = changed.newValue as boolean;
    }, true);
  });

  it("when time passes, returns true", () => {
    jest.advanceTimersByTime(1000);

    expect(actual).toBe(true);
  });

  it("does not share the state from previous test", () => {
    expect(actual).toBe(false);
  });
});
```

Key here is that if you focus on either of tests individually, they work. But since both are running, the latter fails. 

Opened issue here: https://github.com/mobxjs/mobx-utils/issues/306

But while we're waiting them to solve the issue, I'll figure out solution for us.

jansav: Fixed the issue here: https://github.com/lensapp/lens/pull/5764

It was more than earlier described.

src/renderer/components/status-bar/auto-update-component.tsx
jansav: Cool. :)

jansav (approved):

aleksfront (approved):
Things should be improved next: 
1. Do not show "No updates available" notification on app start.

https://user-images.githubusercontent.com/9607060/176879404-001c7d7e-8a0c-4662-9b9b-7ca43bd3ee47.mov

2. Hide empty status bar elements to prevent rendering them in the dom. Currently 
```
const idle = () => <div data-testid="app-update-idle"></div>;
```
prevents us to do it.


https://user-images.githubusercontent.com/9607060/176879800-4645bb54-de06-49cd-8ac5-e580c84c60f9.mov

PR #5369, opened 2022-05-11
Upgrade to electron@^19

Nokel81:

blocked on https://github.com/lensapp/lenscloud-lens-extension/issues/1660


Nokel81:
Electron 15 and 16 have reached end-of-life and are no longer supported. This cannot be merged as the `<webview>`'s used by the extension API currently cannot open windows.

They should be removed entirely.

ixrock:
Any updates?

Nokel81:
@ixrock Yeah I will get back to fixing this PR up. It is currently blocked on removing webview usages in some of the bundled extensions first.

aleksfront:
Upvoting for upgrade. Bumped into some other bugs which will be fixed from electron 16.

aleksfront (approved):
Looks good. If you face `pty.node` error while compiling
```
[0] App threw an error during load
[0] Error: The module '/Users/user/lens/node_modules/node-pty/build/Release/pty.node'
[01 was compiled against a different Node.is version using
[01 NODE MODULE VERSION 98. This version of Node.is requires
[0] NODE MODULE_VERSION 106.
```
just manually remove `node_modules` folder.

src/main/start-main-application/lens-window/application-window/create-electron-window.injectable.ts
```diff
@@ -76,9 +76,7 @@ const createElectronWindowInjectable = getInjectable({
         webPreferences: {
           nodeIntegration: true,
           nodeIntegrationInSubFrames: true,
-          webviewTag: true,
```
jakolehm, line 79: Instead of removing this line should we keep it with `false` value (to make it clear)?

jakolehm (approved):

src/main/start-main-application/lens-window/application-window/create-electron-window.injectable.ts
Nokel81: Well th default is false and I don't see electron changing that default now.

chenhunghan:
`"@types/node": "^16.14.2"` seems wrong, the latest version on npm is `16.11.49`, and what actually installed in this PR is `16.11.34` (according to lockfile)

Issue #5214, opened 2022-04-07
Orange indicator for too long since last update


Nokel81:
If an update has not been found for a long time or if the update server is not accessible then the icon from https://github.com/lensapp/lens/issues/5207 should be set as orange with some information about checking the website for an update.

PR #5376, opened 2022-05-12
TopBar Update button

aleksfront:
UI Part of #5214 

Update button with 3 warning levels, each of them changes button color. Clicking 'Update' button will show menu with `Relaunch Lens to Update` item.

Dark theme:

https://user-images.githubusercontent.com/9607060/168063398-dbb1a94c-d110-455a-8006-8ca3f5ec69e8.mov

update medium dark
update high dark

Light theme:
update high light
update medium light
update light light

Update button is isolated and currently completely hidden. It will show up when some `warningLevel` prop will be passed.
no update button

### DIFF
--- a/src/renderer/components/update-button/__tests__/update-button.test.tsx
+++ b/src/renderer/components/update-button/__tests__/update-button.test.tsx
@@ -5,7 +5,7 @@
 
 import { render } from "@testing-library/react";
 import React from "react";
-import { UpdateButton } from "..";
+import { UpdateButton } from "../update-button";
 import "@testing-library/jest-dom/extend-expect";
 
 const update = jest.fn();
--- a/src/renderer/components/update-button/__tests__/update-button.test.tsx
+++ b/src/renderer/components/update-button/__tests__/update-button.test.tsx
@@ -1,0 +1,65 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import { render } from "@testing-library/react";
+import React from "react";
+import { UpdateButton } from "..";
+import "@testing-library/jest-dom/extend-expect";
+
+const update = jest.fn();
+
+describe("<UpdateButton/>", () => {
+  it("should not render if no warning level prop passed", () => {
+    const { queryByTestId } = render(<UpdateButton update={update} />);
+
+    expect(queryByTestId("update-button")).not.toBeInTheDocument();
+  });
+
+  it("should render if warning level prop passed", () => {
+    const { getByTestId } = render(<UpdateButton update={update} warningLevel="light" />);
+
+    expect(getByTestId("update-button")).toMatchSnapshot();
+  });
+
+  it("should open menu when clicked", () => {
+    const { getByTestId } = render(<UpdateButton update={update} warningLevel="light" />);
+
+    const button = getByTestId("update-button");
+
+    button.click();
+
+    expect(getByTestId("update-lens-menu-item")).toBeInTheDocument();
+  });
+
+  it("should call update function when menu item clicked", () => {
+    const { getByTestId } = render(<UpdateButton update={update} warningLevel="light" />);
+
+    const button = getByTestId("update-button");
+
+    button.click();
+
+    const menuItem = getByTestId("update-lens-menu-item");
+
+    menuItem.click();
+
+    expect(update).toHaveBeenCalled();
+  });
+
+  it("should have class name with medium warning level", () => {
+    const { getByTestId } = render(<UpdateButton update={update} warningLevel="medium" />);
+
+    const button = getByTestId("update-button");
+
+    expect(button.className.toLowerCase().includes("medium")).toBeTruthy();
+  });
+
+  it("should have class name with high warning level", () => {
+    const { getByTestId } = render(<UpdateButton update={update} warningLevel="high" />);
+
+    const button = getByTestId("update-button");
+
+    expect(button.className.toLowerCase().includes("high")).toBeTruthy();
+  });
+});
--- a/src/renderer/components/update-button/__tests__/update-button.test.tsx
+++ b/src/renderer/components/update-button/__tests__/update-button.test.tsx
@@ -3,7 +3,7 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
 
-import { render } from "@testing-library/react";
+import { render, act } from "@testing-library/react";
 import React from "react";
 import { UpdateButton } from "../update-button";
 import "@testing-library/jest-dom/extend-expect";
@@ -27,12 +27,12 @@ describe("<UpdateButton/>", () => {
     expect(getByTestId("update-button")).toMatchSnapshot();
   });
 
-  it("should open menu when clicked", () => {
+  it("should open menu when clicked", async () => {
     const { getByTestId } = render(<UpdateButton update={update} warningLevel="light" />);
 
     const button = getByTestId("update-button");
 
-    button.click();
+    act(() => button.click());
 
     expect(getByTestId("update-lens-menu-item")).toBeInTheDocument();
   });
@@ -42,7 +42,7 @@ describe("<UpdateButton/>", () => {
 
     const button = getByTestId("update-button");
 
-    button.click();
+    act(() => button.click());
 
     const menuItem = getByTestId("update-lens-menu-item");
 
--- a/src/renderer/components/update-button/__tests__/update-button.test.tsx
+++ b/src/renderer/components/update-button/__tests__/update-button.test.tsx
@@ -51,19 +51,27 @@ describe("<UpdateButton/>", () => {
     expect(update).toHaveBeenCalled();
   });
 
-  it("should have class name with medium warning level", () => {
+  it("should have light warning level", () => {
+    const { getByTestId } = render(<UpdateButton update={update} warningLevel="light" />);
+
+    const button = getByTestId("update-button");
+
+    expect(button.dataset.warningLevel).toBe("light");
+  });
+
+  it("should have medium warning level", () => {
     const { getByTestId } = render(<UpdateButton update={update} warningLevel="medium" />);
 
     const button = getByTestId("update-button");
 
-    expect(button.className.toLowerCase().includes("medium")).toBeTruthy();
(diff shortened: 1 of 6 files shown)


src/renderer/components/update-button/update-button.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { noop } from "lodash";
+import { cssNames } from "../../utils";
+
+interface UpdateButtonProps {
+  warningLevel?: "light" | "medium" | "high";
+  update: () => void;
+}
+
+export function UpdateButton({ warningLevel, update }: UpdateButtonProps) {
+  const id = uniqueId("update_button_");
```
Nokel81: I don't particularly like the fact that this is non-deterministic. Can we just have `id` passed in as a required prop?

src/renderer/components/update-button/styles.module.scss
```diff
@@ -0,0 +1,42 @@
...
+  --active-color-hsl: 0deg 76% 62%;
+}
+
+.warningMedium {
+  --active-color-hsl: 36deg 100% 50%;
+}
+
+:global(.theme-light) .warningMedium {
+  // Making medium color a bit darker for light theme
+  --active-color-hsl: 36deg 100% 46%;
```
Nokel81: Why is this not just part of the theme then?

src/renderer/components/update-button/__tests__/update-button.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import { render } from "@testing-library/react";
+import React from "react";
+import { UpdateButton } from "..";
+import "@testing-library/jest-dom/extend-expect";
+
+const update = jest.fn();
```
Nokel81, line 11: Shouldn't this be a local to the tests and refreshed in a `beforeEach` block?

Nokel81:
Also, do we need the hamburger menu icon as well?

aleksfront:
> Also, do we need the hamburger menu icon as well?

Yes, it shows that you shouldn't afraid to click `Update` button because it contains dropdown menu. So there won't be any immediate action if you click.

src/renderer/components/update-button/update-button.tsx
aleksfront: It is just used by the following `<Menu>` to find element which will trigger menu to open. So no need to make it a required prop, I want to make this component to be as easy as possible.

src/renderer/components/update-button/styles.module.scss
aleksfront: Using HSL format allowing to make color opacity transformations using native css like so:
```
background: hsl(var(--active-color-hsl) / 15%);
```

where `15%` describes opacity layer. I can ofc put this colors into theme file if we find it important, but in my opinion theme file should be cleaned and trying not to add non-important colors there. Any other opinions on that? @ixrock

src/renderer/components/update-button/update-button.tsx
Nokel81: Yeah but it makes snapshots harder because the ID's will change if we reorder the tests.

src/renderer/components/update-button/styles.module.scss
Nokel81: I meant for the yellow/red/green colours to be part of theme (not the transparency bit) because I could easily envision them being reused.

jakolehm:
> > Also, do we need the hamburger menu icon as well?
> 
> Yes, it shows that you shouldn't afraid to click `Update` button because it contains dropdown menu. So there won't be any immediate action if you click.

@aleksfront could we test how three dots looks as a replacement for hamburger? (I think Chrome uses three dots, not hamburger)

Nokel81:
Firefox uses hamburger I think, haven't used chrome in ages.

jakolehm:
> Firefox uses hamburger I think, haven't used chrome in ages.

That's fine but Lens itself does not use hamburger (anywhere?), within Lens three-dots is more common and we really should try to keep UI unified (unless we have really good reason).

jakolehm:
> > Firefox uses hamburger I think, haven't used chrome in ages.
> 
> That's fine but Lens itself does not use hamburger (anywhere?), within Lens three-dots is more common and we really should try to keep UI unified (unless we have really good reason).

Arrow down is also quite commonly used in Lens.

src/renderer/components/update-button/styles.module.scss
ixrock: @aleksfront I'd prefer to keep colors in HEX if possible or why we cannot keep colors per theme in rgba() or hexa formats? Kinda like this:

```scss
:root {
 --lens-red: #EE0000; // no-alpha
 --lens-green: #00FF0AA; // with alpha via hex-format
}

:root[data-theme=dark] {
  --lens-red: red;
 --lens-green: rgba(0, 100, 0, 50%); 
}
```

or here is other fancy way to control global themed `--red` color or it's chunks separately, see example:

```scss
:root {
    // might be themed variables
    --op: 30%;
    --bg: var(--red-final, rgba(var(--red, 0), var(--glue, 0), var(--blue, 0), var(--op)));
}

.SomeComponent {
     background: var(--bg, red);

    &.dark {
      --red-final: rgba(255,0,0,.5); // has more priority if defined
      --red: 255; // doesn't affected since ^
      --blue: 100; // doesn't affected since ^
    }
}

 ```

Comment (ixrock) [reply src/renderer/components/update-button/styles.module.scss]: **Demo:** _(could it simplify your  issue somehow?)_

https://user-images.githubusercontent.com/6377066/168426222-9f006428-c276-4047-a890-b097b7536b37.mov

File: src/renderer/components/update-button/update-button.tsx [hunk 4]
```diff
@@ -0,0 +1,70 @@
...
+
+import React, { useState } from "react";
+import { Menu, MenuItem } from "../menu";
+import uniqueId from "lodash/uniqueId";
+import { noop } from "lodash";
+import { cssNames } from "../../utils";
+
+interface UpdateButtonProps {
+  warningLevel?: "light" | "medium" | "high";
+  update: () => void;
```
ixrock, line 17: Maybe it should be optional and you won't need things like this?

```tsx
<UpdateButton update={noop} /> // top-bar.tsx
```

src/renderer/components/update-button/update-button.tsx
ixrock: Maybe this? Saves only initially rendered `props.id` if provided or generating random one:
Or if we need to support dynamic `props.id` for component, then go with a `getter` or `computed (mobx)` expression.

```suggestion
  const id = this.props.id ?? uniqueId("update_button_");
```

src/renderer/components/update-button/update-button.tsx
aleksfront: This is intentional. Without `update` function this component doesn't make any sense. Current form with `update={noop}` is used only temporary before business logic for this button will be ready.

src/renderer/components/update-button/styles.module.scss
aleksfront: Thanks for suggestion. I've used pseudo-element with variable opacity to perform background changes. So no need to add extra colors + accent color goes from theme.

aleksfront:
update dropdown high
update dropdown medium
update dropdown theme-light

@Nokel81 @ixrock PR is ready for another review. Please take a look.

src/renderer/components/update-button/__tests__/update-button.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+    expect(button.className.toLowerCase().includes("medium")).toBeTruthy();
+  });
+
+  it("should have class name with high warning level", () => {
+    const { getByTestId } = render(<UpdateButton update={update} warningLevel="high" />);
+
+    const button = getByTestId("update-button");
+
+    expect(button.className.toLowerCase().includes("high")).toBeTruthy();
+  });
```
ixrock: Maybe we should not rely on `classNames`? _(it's gonna be broken test in future imho)_
Better to assign `<Update data-warning-level={warningLevel}/>` and check this value in tests.

Also you're testing here only 2 of 3  `warningLevel` possible values (`light` used previously in test to check just button existence, but it didn't check anything about applied update to the DOM/css like you do here).

src/renderer/components/update-button/update-button.tsx
```diff
@@ -0,0 +1,63 @@
...
+  };
+
+  if (!warningLevel) {
+    return null;
+  }
+
+  return (
+    <>
+      <button
+        data-testid="update-button"
```
ixrock, line 36: _To improve tests:_

```suggestion
        data-testid="update-button"
        data-warning-level={warningLevel}
```

ixrock (dismissed):

src/renderer/components/update-button/__tests__/update-button.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    expect(update).toHaveBeenCalled();
+  });
+
+  it("should have class name with medium warning level", () => {
+    const { getByTestId } = render(<UpdateButton update={update} warningLevel="medium" />);
+
+    const button = getByTestId("update-button");
+
+    expect(button.className.toLowerCase().includes("medium")).toBeTruthy();
```
ixrock:
```suggestion
    expect(button.dataset["warning-level"]).toBe("medium");
```

Nokel81 (dismissed):
Looks good, but Roman's comments seem applicable.

aleksfront:
Fixed tests a bit.

src/renderer/components/update-button/update-button.tsx
```diff
@@ -50,6 +50,7 @@ export function UpdateButton({ warningLevel, update, id }: UpdateButtonProps) {
         isOpen={opened}
         close={toggle}
         open={toggle}
+        aria-expanded={opened}
```
ixrock: What it does?

ixrock (dismissed):

src/renderer/components/update-button/update-button.tsx
aleksfront: A leftover from fixing tests.

ixrock (approved):

Nokel81 (approved):

PR #5389, opened 2022-05-13
Refactoring extension-app compatibility checks

ixrock:
Follow-up to https://github.com/lensapp/lens/pull/5290#discussion_r872186241

### DIFF
--- a/src/extensions/__tests__/is-compatible-extension.test.ts
+++ b/src/extensions/__tests__/is-compatible-extension.test.ts
@@ -11,40 +11,54 @@ import type { LensExtensionManifest } from "../lens-extension";
 
 describe("Extension/App versions compatibility checks", () => {
   it("is compatible with exact version matching", () => {
-    expect(isCompatible("5.5.0", "5.5.0")).toBeTruthy();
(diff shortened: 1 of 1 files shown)


src/extensions/__tests__/is-compatible-extension.test.ts
```diff
@@ -9,108 +9,47 @@ import {
...
+    expect(() => isCompatible("~2.0", "2.0")).toThrow(/Invalid format/i);
+    expect(() => isCompatible("*", "1.0")).toThrow(/Invalid format/i);
   });
 });
 
-function getExtensionManifestMock(
-  {
-    lensEngine = "1.0",
-  } = {}): LensExtensionManifest {
+function isCompatible(extensionLensEngine = "^1.0", appVersion = "1.0"): boolean {
```
Nokel81: nit: I would prefer object deconstructing here so that I know when reading the tests which value means which.

ixrock: Fixed

Nokel81 (approved):

jansav (approved):

PR #5399, opened 2022-05-16
Fix downloading cluster specific kubectl

Nokel81:
- The normalization of the architecture was incorrect as the kubectl
  download server uses a different normalization than what OpenLens does
  for its bundled binaries.

- Made that split explicit by introducing some new injectables

- Made the Kubectl class not depend on any global shared state by
  introducing new injectables


jim-docker (commented):
The downloading is working but Lens is still using the bundled version in the terminal:

Screen Shot 2022-05-16 at 3 39 55 PM

See the [SHELL-SESSION] error

jim-docker:
> 
> See the [SHELL-SESSION] error

Actually, this error is also in the latest master (it shows every 30s)

Nokel81:
> Actually, this error is also in the latest master (it shows every 30s)

Okay that should be fixed then. But it does look like we are not picking the correct `kubectl` version anyway. I will look into this.

Nokel81:
@jim-docker Fixed.

Nokel81:
Also, how did you encounter that other issue, I have not seen it.

jim-docker:
> Also, how did you encounter that other issue, I have not seen it.

I see it on all 3 platforms, I have a terminal in the dock, which opens when connecting to a cluster. Error appears every ~30s

src/main/kubectl/kubectl.ts
```diff
@@ -299,7 +299,7 @@ export class Kubectl {
       "elif test -f \"$HOME/.profile\"; then",
       "  . \"$HOME/.profile\"",
       "fi",
-      `export PATH="${binariesDir}:${kubectlPath}:$PATH"`,
+      `export PATH="${kubectlPath}:${binariesDir}:$PATH"`,
```
jim-docker, line 302: these changes don't help on Windows (it still uses the bundled kubectl)

Nokel81: Fixed

Nokel81:
Closing in favour of https://github.com/lensapp/lens/pull/5413, because master is diverging too far

Will make a PR to merge the release branch back into master once 5.5 is out

Nokel81:
Marking this PR as `skip-changelog` so that it isn't displayed as part of the change-log. We need this PR here because the other PR is targeting the `release/v5.5` branch.

jim-docker (approved):

PR #5403, opened 2022-05-17
Fix changes to InputValidator to resolve accidental breaking changes

Nokel81:

### DIFF
--- a/src/renderer/components/+extensions/install.tsx
+++ b/src/renderer/components/+extensions/install.tsx
@@ -9,14 +9,12 @@ import { prevDefault } from "../../utils";
 import { Button } from "../button";
 import { Icon } from "../icon";
 import { observer } from "mobx-react";
-import { Input, InputValidators } from "../input";
+import { asyncInputValidator, Input, InputValidators } from "../input";
 import { SubTitle } from "../layout/sub-title";
 import { TooltipPosition } from "../tooltip";
 import type { ExtensionInstallationStateStore } from "../../../extensions/extension-installation-state-store/extension-installation-state-store";
-import extensionInstallationStateStoreInjectable
-  from "../../../extensions/extension-installation-state-store/extension-installation-state-store.injectable";
+import extensionInstallationStateStoreInjectable from "../../../extensions/extension-installation-state-store/extension-installation-state-store.injectable";
 import { withInjectables } from "@ogre-tools/injectable-react";
-import { inputValidator } from "../input/input_validators";
 
 export interface InstallProps {
   installPath: string;
@@ -30,17 +28,22 @@ interface Dependencies {
   extensionInstallationStateStore: ExtensionInstallationStateStore;
 }
 
-const installInputValidators = [
-  InputValidators.isUrl,
-  InputValidators.isPath,
-  InputValidators.isExtensionNameInstall,
-];
+const installInputValidator = asyncInputValidator({
+  validate: async (value) => {
+    if (
+      InputValidators.isUrl.validate(value)
+      || InputValidators.isExtensionNameInstall.validate(value)
+    ) {
+      return;
+    }
 
-const installInputValidator = inputValidator({
-  message: "Invalid URL, absolute path, or extension name",
-  validate: (value: string, props) => (
-    installInputValidators.some(({ validate }) => validate(value, props))
-  ),
+    try {
+      return await InputValidators.isPath.validate(value);
+    } catch {
+      throw new Error("Invalid URL, absolute path, or extension name");
+    }
+  },
+  debounce: InputValidators.isPath.debounce,
 });
 
 const NonInjectedInstall: React.FC<Dependencies & InstallProps> = ({
--- a/src/renderer/components/+extensions/install.tsx
+++ b/src/renderer/components/+extensions/install.tsx
@@ -9,12 +9,13 @@ import { prevDefault } from "../../utils";
 import { Button } from "../button";
 import { Icon } from "../icon";
 import { observer } from "mobx-react";
-import { asyncInputValidator, Input, InputValidators } from "../input";
+import { Input, InputValidators } from "../input";
 import { SubTitle } from "../layout/sub-title";
 import { TooltipPosition } from "../tooltip";
 import type { ExtensionInstallationStateStore } from "../../../extensions/extension-installation-state-store/extension-installation-state-store";
 import extensionInstallationStateStoreInjectable from "../../../extensions/extension-installation-state-store/extension-installation-state-store.injectable";
 import { withInjectables } from "@ogre-tools/injectable-react";
+import { unionInputValidatorsAsync } from "../input/input_validators";
 
 export interface InstallProps {
   installPath: string;
@@ -28,23 +29,14 @@ interface Dependencies {
   extensionInstallationStateStore: ExtensionInstallationStateStore;
 }
 
-const installInputValidator = asyncInputValidator({
-  validate: async (value) => {
-    if (
-      InputValidators.isUrl.validate(value)
-      || InputValidators.isExtensionNameInstall.validate(value)
-    ) {
-      return;
-    }
-
-    try {
-      return await InputValidators.isPath.validate(value);
-    } catch {
-      throw new Error("Invalid URL, absolute path, or extension name");
-    }
+const installInputValidator = unionInputValidatorsAsync(
+  {
+    message: "Invalid URL, absolute path, or extension name",
   },
-  debounce: InputValidators.isPath.debounce,
-});
+  InputValidators.isUrl,
+  InputValidators.isExtensionNameInstall,
+  InputValidators.isPath,
+);
 
 const NonInjectedInstall: React.FC<Dependencies & InstallProps> = ({
   installPath,
--- a/src/renderer/components/input/input.tsx
+++ b/src/renderer/components/input/input.tsx
@@ -191,6 +191,8 @@ export class Input extends React.Component<InputProps, State> {
             return this.getValidatorError(value, validator) || (error instanceof Error ? error.message : String(error));
           }
         })());
+      } else if (!validator.validate(value, this.props)) {
+        errors.push(this.getValidatorError(value, validator));
       }
     }
 
--- a/src/renderer/components/input/input.tsx
+++ b/src/renderer/components/input/input.tsx
@@ -12,15 +12,25 @@ import { Icon } from "../icon";
 import type { TooltipProps } from "../tooltip";
 import { Tooltip } from "../tooltip";
 import * as Validators from "./input_validators";
-import type { InputValidator } from "./input_validators";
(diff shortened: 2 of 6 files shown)


jweak (dismissed):
LGTM

jweak (dismissed):

src/renderer/components/+extensions/install.tsx
```diff
@@ -30,17 +28,22 @@ interface Dependencies {
   extensionInstallationStateStore: ExtensionInstallationStateStore;
 }
 
-const installInputValidators = [
-  InputValidators.isUrl,
-  InputValidators.isPath,
-  InputValidators.isExtensionNameInstall,
-];
```
aleksfront: It looks more readable to keep all validators in array comparing to new version where different logic mixed in one function.

src/renderer/components/input/input.tsx
```diff
@@ -168,32 +175,16 @@ export class Input extends React.Component<InputProps, State> {
...
-        if (!validationId) {
-          this.validationId = validationId = uniqueId("validation_id_");
-        }
-        asyncValidators.push(
-          result.then(
-            () => null, // don't consider any valid result from promise since we interested in errors only
-            error => this.getValidatorError(value, validator) || error,
-          ),
-        );
-      }
```
aleksfront: Quite heavy code block removed, which is super great. But to make 100% sure we don't need this anywhere?

src/renderer/components/input/input_validators.ts
```diff
@@ -92,26 +116,24 @@ export const isExtensionNameInstall = inputValidator({
...
-    } catch {
-      throw new AsyncInputValidationError(`${value} is not a valid file path`);
+    if (!await fse.pathExists(value)) {
+      throw new Error(`"${value}" is not a valid file path`);
     }
   },
 });
 
-export const minLength = inputValidator({
+export const minLength = inputValidator<true>({
```
aleksfront: First time see this notation, does it mean same as `inputValidator<boolean>` or it can't be `false`?

aleksfront (commented):
Can we specify what exact issues it solves and where is the actual fixes / additional improvements? I'm just a bit worried that we're adding extra code here.

src/renderer/components/+extensions/install.tsx
Nokel81: The problem is that the old code was had a bug in it. It would cause a "UnhandledPromiseRejectionError" because `isPath` is an async validator

src/renderer/components/input/input.tsx
Nokel81: Is caused the duplicate error messages bug. Let me find the issue about that.

src/renderer/components/input/input_validators.ts
Nokel81: It means that the second parameter for the validate function is required. I can make that easier to read.

src/renderer/components/input/input.tsx
Nokel81: I cannot find it, but here is a screenshot of the bug (from master)
Screen Shot 2022-05-17 at 3 47 33 PM

src/renderer/components/input/input_validators.ts
Nokel81: I have backed out the ability to specify that it is required, thus making this easier to read.

src/renderer/components/+extensions/install.tsx
aleksfront: Well, I still think this new function `installInputValidator` should be simplified or splitted to smaller ones. If we want to add some new validator to the list, we will touch code of this `installInputValidator` and probably add some another ifs or replace conditions which is not good. Not good because it can break other things.

Maybe we can use `for ... of` to iterate through the validators array https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of

My quick checks shows that it "handles" async and non-async items:

```
const x = () => true
const y = () => true
const z = () => {
    return new Promise(resolve => {
        resolve(true);
    });
}

for await (const variable of [x, y, z]) {
    console.log(variable())
}

// prints
// VM593:2 true
// VM593:2 Promise {<fulfilled>: true}
```

src/renderer/components/+extensions/install.tsx
```diff
@@ -30,17 +28,22 @@ interface Dependencies {
...
+const installInputValidator = asyncInputValidator({
+  validate: async (value) => {
+    if (
+      InputValidators.isUrl.validate(value)
+      || InputValidators.isExtensionNameInstall.validate(value)
+    ) {
+      return;
```
chenhunghan: Do we need to return `true`? A bit confusing for me if a validator doesn't return boolean..

src/renderer/components/input/input.tsx
```diff
@@ -153,7 +161,7 @@ export class Input extends React.Component<InputProps, State> {
   async validate() {
     const value = this.getValue();
     let validationId = (this.validationId = ""); // reset every time for async validators
-    const asyncValidators: Promise<any>[] = [];
+    const asyncValidators: Promise<React.ReactNode>[] = [];
```
chenhunghan, line 171: Could you please explain why is validators an array of `Promise<React.ReactNode>`? I went through the code but don't get why they are `Promise<React.ReactNode>`

src/renderer/components/input/input_validators.ts
```diff
@@ -4,37 +4,59 @@
...
 );
 
-export function inputValidator<IsAsync extends boolean = false>(validator: InputValidator<IsAsync>): InputValidator<IsAsync> {
+export function asyncInputValidator(validator: InputValidator<true, false>): InputValidator<true, false> {
+  return validator;
+}
+
+export function inputValidator(validator: InputValidator<false, false>): InputValidator<false, false> {
+  return validator;
+}
```
chenhunghan: I don't get how these two different.. just the name? They seems to get the same input, and return the same output.

chenhunghan: ah, now I found the diff, `InputValidator<false, false>` vs `InputValidator<true, false>`

src/renderer/components/input/input_validators.ts
```diff
@@ -92,30 +114,26 @@ export const isExtensionNameInstall = inputValidator({
...
-export const isPath = inputValidator<true>({
+export const isPath = asyncInputValidator({
   debounce: 100,
   condition: ({ type }) => type === "text",
   validate: async value => {
-    try {
-      await fse.pathExists(value);
-    } catch {
-      throw new AsyncInputValidationError(`${value} is not a valid file path`);
+    if (!await fse.pathExists(value)) {
```
chenhunghan, line 179: Looks like old code catches the error thrown from `fse.pathExists`, do we still need to catch this exception, or it's handled somewhere else?

src/renderer/components/+extensions/install.tsx
Nokel81: Fair enough, the rule is that synchronous validators must return booleans but async validators reject to show errors.

src/renderer/components/+extensions/install.tsx
Nokel81: The problem is that in the async case, it isn't that the promise should resolve to `true` but that it shouldn't reject. This is a hold over from how they were previously implemented.

src/renderer/components/input/input.tsx
Nokel81: The reason is on line 182 below. Basically this is here to signify that the promises added to this list should resolve to something that can be rendered (the error message)

src/renderer/components/input/input_validators.ts
Nokel81: The old code was wrong `fse.pathExists` never throws, it always resolves.

src/renderer/components/+extensions/install.tsx
ixrock: Haven't read everything in this PR.. but just quick question:
if we don't want to that some async validator _might throw exception_ why we cannot just wrap it with extra `.catch(() => true)` or something?

Nokel81: The thing is that async validators have always used `throw` to mean "failed to validate". This PR does not change that and doesn't try to. Also, the error message from the rejection was always used as the message displayed to the user.

I agree that a better and more unified design could have been fixed (either always using throw/reject or always using `return boolean`).

ixrock: If I'm not mistaken if we wrap every promise with extra `.catch(() => {})` we can filter all relevant errors / skip   `UnhandledPromiseRejectionError` or just passthrough inside catch with `throw` again.

e.g. something like this wouldn't work?

```ts
// validation cycle
await Promise.resolve(validator()).catch(err => {
   if(err instanceof UnhandledPromiseRejectionError) return true; // all good
   throw String(err); // pass real validation error
})
```

Nokel81: No unfortunately because we won't get back `UnhandledPromiseRejectionError`.

What I think I can do though, is make a "union" function for joining several input validators together.

ixrock: What about catching this error with `process.on('unhandledRejection')`? Would it make sense?

```ts
process.on('unhandledRejection', error => {
  // Will print "unhandledRejection err is not defined"
  console.log('unhandledRejection', error.message);
});

new Promise((_, reject) => reject(new Error('woops'))).
  catch(error => {
    // Will not execute
    console.log('caught', err.message);
  });
```

Nokel81: That seems very hacky and wrong, especially since that is global shared state. It isn't that hard to make a union function that works correctly. Just shows the horribleness of the current design.

Nokel81: I have pushed the unionizing function, PTAL.

src/renderer/components/input/input_validators.ts
```diff
@@ -4,41 +4,125 @@
...
+            return;
+          }
+        }
+      }
+
+      /**
+       * If no validator returns `true` then mark as invalid by throwing. The message will be
+       * obtained from the `message` field.
+       */
+      throw {};
```
panuhorsmalahti: Why not `throw new Error();`?

aleksfront (commented):
I still got the errors trying to compile extension which uses validators. Compilation goes fine with previous `k8slens/extensions` version `5.5.0-git.c376b848c9.0`. Error example:
validator errors

I guess we have to decide between 2 options:
1. Fine-tune this PR to remove extension compilation errors (sine the name of the PR contains "resolve accidental breaking changes").
2. Just to change typings within our extension to match them provided in current PR.

src/renderer/components/input/input_validators.ts
Nokel81: Sure, it doesn't matter.

jweak (approved):

chenhunghan (approved):
LGTM(?)

aleksfront (approved):

PR #5408, opened 2022-05-17
Switch to always using latest minikube in CI

Nokel81:

This PR fixes the issue seen at https://github.com/lensapp/lens/runs/6472618905?check_suite_focus=true where if there is a minikube upgrade then CI on ubuntu will fail because there is an interactive prompt which never gets resolved.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -73,7 +73,9 @@ jobs:
 
       - name: Install integration test dependencies
         id: minikube
-        uses: medyagh/setup-minikube@5a9a7104d7322fa40424de8855c84685e89cefd7
+        uses: medyagh/setup-minikube@master
+        with:
+          minikube-version: latest
         if: runner.os == 'Linux'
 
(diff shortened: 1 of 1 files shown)


ixrock (approved):

.github/workflows/test.yml
```diff
@@ -73,7 +73,9 @@ jobs:
 
       - name: Install integration test dependencies
         id: minikube
-        uses: medyagh/setup-minikube@5a9a7104d7322fa40424de8855c84685e89cefd7
+        uses: medyagh/setup-minikube@master
```
jim-docker, line 76: This means that there can be other kinds of sudden failures due to constantly changing minikube versions?

jim-docker (approved):

.github/workflows/test.yml
Nokel81: Kubernetes only does updates about 3 times a year, so I doubt that that will happen often, but yes that is a concern.

PR #5414, opened 2022-05-18
Custom terminal fonts support

### LINKED ISSUE #5132: Changing terminal font family clears open terminal
Changing terminal font family on preferences page causes any existing text in a previously open terminal to be cleared (in currently open dashboard):

https://user-images.githubusercontent.com/40840436/160643000-74edc3f9-e3b6-47e4-ba84-93a9be6282b1.mov

happens in 5.4.4

### LINKED ISSUE #5133: terminal font family other than RobotoMono have large char spacing
Even monospaced fonts like CourierNew seem to have too large spacing between characters:

Screen Shot 2022-03-29 at 10 43 15 AM

Only RobotoMono has correct spacing

happens on 5.4.4

ixrock:
New fonts list for terminal with current `fontSize` live-preview: _(fonts downloaded from https://fonts.google.com/)_

Screenshot 2022-05-25 at 16 48 44

fix https://github.com/lensapp/lens/issues/5132
fix https://github.com/lensapp/lens/issues/5133


ixrock:
@lensapp/lens-maintainers do we still need to support custom font as `string`? _(might not work as good as user expecting)_ + font must be installed globally + it should be `monospace` for good reading view. Wdyt?


**Demo:**

https://user-images.githubusercontent.com/6377066/170259059-32c39ce9-e764-409d-ac51-f50b3257a92e.mov

src/renderer/components/+preferences/terminal.tsx
```diff
@@ -88,11 +95,11 @@ const NonInjectedTerminal = observer(({
...
         <section>
           <SubTitle title="Font family" />
-          <Input
-            theme="round-black"
-            type="text"
+          <Select
+            themeName="lens"
             value={userStore.terminalConfig.fontFamily}
-            onChange={(value) => userStore.terminalConfig.fontFamily = value}
+            options={supportedCustomFonts}
```
Nokel81, line 118: Minor: It would be cool if the labels for these options were in the font that they represent.

Nokel81 (dismissed):
LGTM!

src/renderer/components/fonts.scss
```diff
@@ -3,34 +3,66 @@
...
+  src: local("Ubuntu Mono"), url("../fonts/UbuntuMono-Regular.ttf") format("truetype");
+  font-display: block;
+}
+
+// Patched RobotoMono font with icons
+// RobotoMono Windows Compatible for using in terminal
+// https://github.com/ryanoasis/nerd-fonts/tree/master/patched-fonts/RobotoMono
+@font-face {
+  font-family: "RobotoMono";
+  src: local("RobotoMono"), url("../fonts/roboto-mono-nerd.ttf") format("truetype");
```
ixrock: Not sure if/when it's needed _(patched font)_ Anybody any clue? 
Could we replace it with normal roboto-mono dowloaded from [google.font](https://fonts.google.com/) ?

Nokel81: I think that is a @aleksfront question.

src/renderer/components/mixins.scss
```diff
@@ -59,20 +59,3 @@
     @content; // css-modules (*.module.scss)
   }
 }
-
-// Makes custom @font-family available at earlier stages.
-// Element must exist in DOM as soon as possible to initiate preloading.
-@mixin font-preload($fontFamily) {
```
ixrock, line 65: Seems like overkill since can serve only single element and require duplicating exported font name (declared in `fonts.scss`). More simple solution would be to define font directly via `style="font-family: %ExportedNameInFontsCss%"` _(see: `template.html`)_ and enable early-preloading therefore.

If early-preloading not the point it's just enough to import and bundle custom font in `fonts.scss`

src/renderer/components/cluster-manager/cluster-frame-handler.ts
```diff
@@ -113,9 +112,9 @@ export class ClusterFrameHandler {
 
           return undefined;
         },
-        (view) => {
+        (view: LensView) => {
           logger.info(`[LENS-VIEW]: cluster id=${clusterId} should now be visible`);
-          view.frame.style.display = "flex";
+          view.frame.classList.remove("hidden");
```
ixrock, line 117: _FYI:_ It's always better to move all styles management if possible to CSS files and just toggle classnames in JS.

src/renderer/components/dock/terminal/terminal.ts
```diff
@@ -197,12 +204,21 @@ export class Terminal {
...
-  setFontFamily = (family: string) => {
-    this.xterm.options.fontFamily = family;
+  setFontFamily = (fontFamily: string) => {
+    logger.info(`[TERMINAL]: set fontFamily to ${fontFamily}`);
+
+    this.xterm.options.fontFamily = fontFamily;
+    this.fit();
+
+    // provide css-variable within `:root {}`
+    document.documentElement.style.setProperty("--font-terminal", fontFamily);
```
ixrock, line 221: Exported global "live" css-variable.

src/renderer/components/+preferences/terminal.tsx
ixrock: Good idea. Gonna take a look soonish. 
_(since all fonts are preloaded and ready to use it's easy-picky addition)_

ixrock: Screenshot 2022-05-25 at 16 48 44

src/renderer/components/fonts.scss
aleksfront: This patched font contains some fancy glyphs modern shells can use. So this kind of things are possible with `roboto-mono-nerd`:
shell glyphs

Other fonts will show blank squares instead of icons/glyphs.

aleksfront (dismissed):
Need to check iframe styles a bit more:

https://user-images.githubusercontent.com/9607060/170486618-ad3035fd-d658-43f9-833a-ce697a8a6ca2.mov

Also:
1. What if user wants to use font not listed in the dropdown?
2. What was the criterion for choosing these fonts? In short: why no more and no less fonts?
3. Can we somehow list font names from the google fonts and download them on demand? It looks a bit excessively that we contain all mono fonts in the bundle (2.1 MB) but I suggest their usage is very rare.

src/renderer/components/fonts.scss
Nokel81: Could we have those glifs patched in, or maybe as a fallback?

ixrock:
> 1. What if user wants to use font not listed in the dropdown?
> 2. What was the criterion for choosing these fonts? In short: why no more and no less fonts?
> 3. Can we somehow list font names from the google fonts and download them on demand? It looks a bit excessively that we contain all mono fonts in the bundle (2.1 MB) but I suggest their usage is very rare.

@aleksfront 

1. we have decided with @Nokel81  to drop support custom font-family as any `string` since it might not look good always in all cases in terminal's canvas _(e.g. font should be monospace and globally installed/available)_
2. picked mostly random different monospaced fonts _(total it's 8 fonts now and should be enough for terminal usage)_
3. it's better to have such things bundled with the app for security reasons (but it's under discussion and we can try to serve google fonts (e.g. all monospaced) via `<link href='..' anonymous>` if all good), but I doubt it would work fine in all cases as you can see we have to "dance" around early-loading/early-usages of  fonts even with bundled ones _(e.g. opened terminal on initial page reload)_.

jim-docker:
Something weird going on, not sure if this is a related bug:

https://user-images.githubusercontent.com/40840436/171225109-aee59b38-0ba3-4e04-bd7c-fda18c35750c.mov


maybe just need to rebase to latest master?

ixrock:
@aleksfront incorrect `iframe`s positioning should be fixed now, PTALA
@jim-docker cannot reproduce your issue:

https://user-images.githubusercontent.com/6377066/171359712-819ac917-d4d7-4ed1-930f-c21d90e128c7.mov

src/renderer/components/fonts.scss
aleksfront: Not sure what you mean, contain glifs separately somehow? Hovewer, as I said before, these fancy shell icons are only possible with patched Roboto.

aleksfront:
Noticed another font-related issue. Not sure what causes it, but in your branch our primary fonts looks a bit different, a more like old-school `Arial` style.

`terminal-fonts-customization`:
bit different font

`master`:

master fonts

ixrock:
> Noticed another font-related issue. Not sure what causes it, but in your branch our primary fonts looks a bit different, a more like old-school `Arial` style.

@aleksfront Good catch! That's because I have removed deprecated package `typeface-roboto` and forgot to download latest `Roboto`-font  from google fonts. Now it's fixed. PTAL.

Nokel81 (dismissed):

jim-docker:
> @jim-docker cannot reproduce your issue:

It doesn't feel font related, but I can easily reproduce it by changing the font size. Connect to a cluster with a terminal open. Go to the Preferences and change the font size by deleting what's there and typing in a new one. Observe that the Font Family field goes blank temporarily. Click X to return to the cluster dashboard and it is locked up:

https://user-images.githubusercontent.com/40840436/171503453-47d6cb6e-1bcd-4bea-a30c-ef26dd9c615a.mov

I checked out latest of this branch, and ran `make clean && make dev`

Maybe it's related to local data, what font I had previously set (migration issue?)

ixrock:
> Observe that the Font Family field goes blank temporarily.

This is expected since current `fontSize` is "live" and used to show actual terminal font in the list.

@jim-docker so this issue you can easily reproduce in current branch, but it's wont appear in `master`? 
If so and the issue arise only in PR's branch then further investigation might be required..

Regarding to migration: I think if user previously has changed font to something else (e.g. `Courier New`) it should continue work for terminal as before and could be changed from now only via current predefined list _(8 fonts available)_.

aleksfront:
There are some problems with `Font Size` field which in my opinion should be fixed in current (or new targeted this one) PR. Font Size field works in conjunction with Font Family dropdown and produces couple of bugs.

1. An issue @jim-docker found earlier (can't reproduce it on my side though)
2. Weird way to change value when you erase previous one with backspace
3. Ability to add 0 as well as very huge size number

https://user-images.githubusercontent.com/9607060/171632443-08fb982f-8799-476b-8fee-de217ddca990.mov



I suggest to simply convert font size field to dropdown with list of predefined values. It should fix all the issues.

jim-docker:
> This is expected since current `fontSize` is "live" and used to show actual terminal font in the list.
> 

I thought so, just wondered if it could have a side effect to cause the bug.

> @jim-docker so this issue you can easily reproduce in current branch, but it's wont appear in `master`? If so and the issue arise only in PR's branch then further investigation might be required..
> 

I'll try on master, but I think this PR is not based on the current `HEAD` of master? I can try to rebase it locally and see if the bug persists.

ixrock:
> I'll try on master, but I think this PR is not based on the current `HEAD` of master? I can try to rebase it locally and see if the bug persists.

it's a bit away from master, no need to testing right now, let's replace `fontSize` input to select first _or restrict it's value somehow (`min`/`max`/etc)_

jim-docker:
Doesn't happen on latest master. Still happens on this branch, rebased on latest master. Eventually I see the same gray screen we've noticed when returning from sleep (with same error in `main`):

Screen Shot 2022-06-02 at 9 41 32 AM

Probably this is not a bug in this PR, but this PR manifests the bug.

ixrock:
Now invalid changes not propagated to input:

https://user-images.githubusercontent.com/6377066/171928612-6af474af-d3cb-4121-8a1d-d5cb20f083eb.mov

src/renderer/components/input/input.tsx
```diff
@@ -173,22 +175,18 @@ export class Input extends React.Component<InputProps, State> {
             error => this.getValidatorError(value, validator) || error,
           ),
         );
-      } else {
-        if (!validator.validate(value, this.props)) {
-          errors.push(this.getValidatorError(value, validator));
-        }
```
ixrock, line 179: This was causing to doubling same error message for sync validators:
image

Nokel81: I thought I also fixed this... Ah it is part of https://github.com/lensapp/lens/pull/5403 which hasn't been merged yet.

jim-docker (dismissed):
bug I found is gone

aleksfront (approved):
LGTM

Nokel81 (approved):

PR #5419, opened 2022-05-18
Generate @k8slens/extensions via tsc only

Nokel81:
- This method only produces the relavent .d.ts files


Nokel81:
Closing as it seems that we actually do want `.js` files in the extension.

PR #5426, opened 2022-05-19
Fix error in legacyRegisterApi on main

Nokel81:

### DIFF
--- a/src/common/k8s-api/kube-api.ts
+++ b/src/common/k8s-api/kube-api.ts
@@ -312,12 +312,13 @@ export interface DeleteResourceDescriptor extends ResourceDescriptor {
  * @deprecated In the new extension API, don't expose `KubeApi`'s constructor
  */
 function legacyRegisterApi(api: KubeApi<any, any>): void {
-  const di = getEnvironmentSpecificLegacyGlobalDiForExtensionApi(Environments.renderer);
-
-  if (di) {
+  try {
+    const di = getEnvironmentSpecificLegacyGlobalDiForExtensionApi(Environments.renderer);
     const autoRegistrationEmitter = di.inject(autoRegistrationEmitterInjectable);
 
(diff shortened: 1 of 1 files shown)


src/common/k8s-api/kube-api.ts
```diff
@@ -312,12 +312,13 @@ export interface DeleteResourceDescriptor extends ResourceDescriptor {
...
-  const di = getEnvironmentSpecificLegacyGlobalDiForExtensionApi(Environments.renderer);
-
-  if (di) {
+  try {
+    const di = getEnvironmentSpecificLegacyGlobalDiForExtensionApi(Environments.renderer);
     const autoRegistrationEmitter = di.inject(autoRegistrationEmitterInjectable);
 
     autoRegistrationEmitter.emit("kubeApi", api);
+  } catch {
+    // ignore error
```
aleksfront, line 331: Seems a bit of strange to me. Why we ever need code above if the application works fine without it? (I suggest that we continue to work okay after the `catch` block).

Nokel81: We need the above code to work in `renderer`, but if the above code is called in `main` it throws (which we don't want).

We don't want the code to run in `main` because the `KubeApi` class (and `ApiManger` for that matter) assume that they get the cluster ID from the hostname of the browser window (which doesn't really exist on `main`).

aleksfront: Okay, maybe we need to add a comment about it.

aleksfront (dismissed):

src/common/k8s-api/kube-api.ts
Nokel81: Sure can do

aleksfront (approved):

jansav (approved):

jansav:
This is good thing. It highlights even more the damage of current design while allowing us to refactor it over time.

PR #5430, opened 2022-05-19
Fix not being able to switch metrics if none are available

Nokel81:
- Previously if there were no metrics for the current selected options then you couldn't switch options.

https://user-images.githubusercontent.com/8225332/169358252-95807813-68fd-4f4b-8622-0e57ab2f1f75.mov

### DIFF
--- a/src/renderer/components/+cluster/cluster-metric-switchers.tsx
+++ b/src/renderer/components/+cluster/cluster-metric-switchers.tsx
@@ -13,6 +13,7 @@ import { MetricNodeRole, MetricType } from "./cluster-overview-store/cluster-ove
 import clusterOverviewStoreInjectable from "./cluster-overview-store/cluster-overview-store.injectable";
 import { withInjectables } from "@ogre-tools/injectable-react";
 import nodeStoreInjectable from "../+nodes/store.injectable";
+import { normalizeMetrics } from "../../../common/k8s-api/endpoints/metrics.api";
 
 interface Dependencies {
   clusterOverviewStore: ClusterOverviewStore;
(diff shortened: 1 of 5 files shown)


Nokel81:
To further elaborate, the bug manifested itself that if there were CPU metrics but not MEMORY metrics and you switched to the memory metrics it was not possible to switch back

src/renderer/components/+cluster/cluster-metrics.module.scss
```diff
@@ -17,7 +17,6 @@
   }
 
   .empty {
-    margin-top: -45px;
```
Nokel81, line 20: This change (and the one below) is a required part of the fix. With this here it was covering the radio buttons and preventing them from being clicked.

src/renderer/components/+cluster/cluster-metric-switchers.tsx
```diff
@@ -24,32 +25,50 @@ const NonInjectedClusterMetricSwitchers = observer(({
...
+  const hasCpuMetrics = normalizeMetrics(cpuUsage).data.result[0].values.length > 0;
+  const hasMemoryMetrics = normalizeMetrics(memoryUsage).data.result[0].values.length > 0;
 
   return (
     <div className="flex gaps" style={{ marginBottom: "calc(var(--margin) * 2)" }}>
       <div className="box grow">
         <RadioGroup
           asButtons
-          className={cssNames("RadioGroup flex gaps", { disabled: disableRoles })}
+          className={cssNames("RadioGroup flex gaps")}
```
aleksfront: Nit: no need for `cssNames` if we're not composing classnames anymore.

src/renderer/components/+cluster/cluster-metric-switchers.tsx
```diff
@@ -24,32 +25,50 @@ const NonInjectedClusterMetricSwitchers = observer(({
...
+            value={MetricNodeRole.WORKER}
+            disabled={!hasWorkerNodes}
+          />
         </RadioGroup>
       </div>
       <div className="box grow" style={{ textAlign: "right" }}>
         <RadioGroup
           asButtons
-          className={cssNames("RadioGroup flex gaps", { disabled: disableMetrics })}
+          className={cssNames("RadioGroup flex gaps")}
```
aleksfront: No need for `cssNames` here.

aleksfront (dismissed):

src/renderer/components/radio/radio.tsx
```diff
@@ -85,6 +85,7 @@ export function Radio<T>({
         type="radio"
         checked={checked}
         onChange={() => ctx.onSelect(value)}
+        disabled={disabled || ctx.disabled}
```
ixrock, line 88: Shouldn't it be like this (since `props.disabled` is optional?)

```suggestion
        disabled={disabled ?? ctx.disabled}
```

src/renderer/components/+cluster/cluster-metric-switchers.tsx
```diff
@@ -24,32 +24,50 @@ const NonInjectedClusterMetricSwitchers = observer(({
...
+            value={MetricNodeRole.WORKER}
+            disabled={!hasWorkerNodes}
+          />
         </RadioGroup>
       </div>
       <div className="box grow" style={{ textAlign: "right" }}>
         <RadioGroup
           asButtons
-          className={cssNames("RadioGroup flex gaps", { disabled: disableMetrics })}
+          className="RadioGroup flex gaps"
```
ixrock, line 57: We should start probably getting rid from such global classnames like `flex gaps` and moving toward css-modules, where we:

- replace those classes with pure css-rules, e.g. classname old global `<div className="Item flex box grow"` becomes css-rule `.Item { display: flex; flex: 1; }` 
- or
- re-use some corresponding classes from `tailwindcss`, e.g. `.Item { @apply flex flex-grow; }`

But this is out of scope for current PR maybe and should be considered as suggestion for all future PRs.

ixrock (approved):

aleksfront (approved):

src/renderer/components/radio/radio.tsx
Nokel81: No because `disabled` is defaulted to `false` above

PR #5431, opened 2022-05-19
Reenable generating of .js files for extension api package

Nokel81:


Nokel81:
I don't actually know if this is required.... @chenhunghan

package.json
```diff
@@ -21,7 +21,7 @@
     "compile": "env NODE_ENV=production concurrently yarn:compile:*",
     "compile:main": "yarn run webpack --config webpack/main.ts",
     "compile:renderer": "yarn run webpack --config webpack/renderer.ts",
-    "compile:extension-types": "yarn run tsc --project tsconfig.extension-api.json",
```
panuhorsmalahti, line 24: Where is `tsconfig.extension-api.json` used after this change?

Nokel81: It is not, I guess I should remove it. I think in the future we might return to using tsc if we no longer output anything other than types from the extension API.

panuhorsmalahti (approved):

jakolehm (approved):

PR #5432, opened 2022-05-19
Fix cluster connection failing after execHelm

### LINKED ISSUE #5423: Connection to cluster failing and using proxy after browsing Helm Charts in connected cluster
**Describe the Bug**
Unable to connect to clusters (which don't require proxy connection) after having been connected and browsed Helm Charts (which does require proxy connection through corporate proxy).

**To Reproduce**
1. Connect to cluster(s)
2. Click on Helm > Charts
3. Disconnect from a cluster
4. On Reconnect to cluster, proxy error

**Expected behavior**
No error connecting to cluster which normally connects without issue.
Believe this is related to https://github.com/lensapp/lens/pull/5273

**Screenshots**
https://user-images.githubusercontent.com/34988938/169263173-5225337d-9044-4d7a-b73c-c6e5de9339d8.mp4

- Lens Version: 5.5.0-beta1
- OS: Windows 11 21H2
- Installation method (e.g. snap or AppImage in Linux): Lens Installer/Updater for Windows

**Logs:**
No logs of significance re: proxy setting from what I can see
```
error:   Γû¬ [CLUSTER]: Failed to connect to "fastbpm-eastus2-prod-aks-1-aks": StatusCodeError: 500 - "x509: certificate signed by unknown authority\n" +369ms
```

**Kubeconfig:**
No issue with Kubeconfig

**Additional context**
Connections to cluster do not require proxy. Proxy is required to reach Helm repos/charts. Seems to set proxy for helm and remains stuck/in-place for all other new connections to clusters.
Thank you for fixing proxy connection for helm though - this is a smaller issue which has a workaround for now. Simply need to force close Lens (from system tray/processes) and re-open.

Nokel81:
- The bug was that we were only assigning a reference to the global process.env instead of cloning it, so we were accidentally modifying the global

fixes #5423

### DIFF
--- a/src/main/helm/exec.ts
+++ b/src/main/helm/exec.ts
@@ -17,7 +17,7 @@ export async function execHelm(args: string[], options?: BaseEncodingOptions & E
   try {
     const opts = { ...options };
 
-    opts.env ??= process.env;
+    opts.env ??= { ...process.env };
 
     if (!opts.env.HTTPS_PROXY && UserStore.getInstance().httpsProxy) {
       opts.env.HTTPS_PROXY = UserStore.getInstance().httpsProxy;


src/main/helm/exec.ts
```diff
@@ -17,7 +17,7 @@ export async function execHelm(args: string[], options?: BaseEncodingOptions & E
   try {
     const opts = { ...options };
 
-    opts.env ??= process.env;
+    opts.env ??= { ...process.env };
```
Nokel81, line 20: I think that our upcoming work should help make this less likely to slip though.

aleksfront (approved):

PR #5438, opened 2022-05-24
Allow bundled extensions to use codesign

jakolehm:
Allows bundled extensions to bundle binaries and sign them with codesign for example during prepack. This is a macOS requirement, even binaries within npm tgz needs to be signed properly.


ixrock (approved):

msa0311 (approved):

Nokel81:
Will merge into the `release/v5.5` myself.

PR #5433, opened 2022-05-20
Introduce way to install update directly from tray. Make application updater unit testable...

jansav:
This PR will:

1. ### Introduce behaviours for updating application
    - Checking for updates automatically
    - Downgrading version update
    - Installing update
    - Installing update from different update channels
    - Installing update using tray

2. ### Introduce way to install update using tray without having to check for updates again
   
2. ### Make notifications unit testable as part of behaviours
 
3. ### Make stuff happening when root frame is rendered unit testable as part of behaviours
 
4. ### Introduce concept called SyncBox to handle situation where main/renderer wants to react to same observable
    - No need to know anything about IPC anymore
    - Usages: `discoveredUpdateVersionInjectable`, `progressOfUpdateDownloadInjectable`, `updateIsBeingDownloadedInjectable` and `updatesAreBeingDiscoveredInjectable`

5. ### Channel Abstraction to handle two-way-communication
    - Consolidates most of the usages of old to new one. Note: some still left untouched for reasons.
    - To see usages, check usages of `channelInjectionToken` and `channelListenerInjectionToken`

6. ### Introduce a way to ask Yes/No from user in main process.
    - Shows notification in window to ask yes/no question from user
    - Check `askBooleanInjectable`

7. ### Introduce higher order functions for error logging and error suppression

8. ### Remove lot of boilerplate from existing unit tests

PR #5445, opened 2022-05-25
Change update warning level over time

### LINKED ISSUE #5207: Update ready for installation display
A colour (and maybe shape) changing display icon that goes from green -> orange -> red depending on how old the oldest non-applied downloaded update for Lens is. Namely, once an update has been downloaded (that still applies vis-à-vis update channel) a green update icon should appear. After 20 days it should change to orange, and after 25 days it should change to red.

Any UI from #1909 or #4948 should be able to change depending on how much time passed since last version is released. Similar to what Chrome and Firefox do if users are more that one major version behind.

**Why is this needed**:
So that a user can know that there are several versions that they are missing features and bug fixes from. And nudge them into updating.

aleksfront:
Changing update warning level over time and show `<UpdateButton>` with appropriate colors.

After receiving `update-downloaded` ipc event from `electron-updater` newly introduced class would set `light` warning level and start check interval.

Warning level following requirements described in https://github.com/lensapp/lens-ide/blob/main/requirements/AUTO_UPDATE.md#technical-requirements (R7-R9). For example, it will go `medium` when 20 days passed after update being downloaded.

Here is the simulation of how button would get changed over time:

https://user-images.githubusercontent.com/9607060/171170380-828df158-b425-48d0-8d8f-6c281d26032f.mov

**Testing**
To simulate `update-downloaded` event `broadcastMessage(UpdateAvailableChannel)` should be fired from somewhere. This will show update button.

Clicking `Relaunch to Update Lens` doesn't work locally because of this https://www.electronjs.org/docs/latest/api/auto-updater#autoupdaterquitandinstall.

Resolves #5207


src/common/__tests__/timezones.test.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+describe("Timezones", () => {
+  it("should always be UTC", () => {
+    expect(new Date().getTimezoneOffset()).toBe(0);
```
ixrock, line 8: Why you're testing standard JS lib and moreover the test should be running only at server with UTC==0?

src/jest.timezone.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+// Setting the timezone to UTC to ensure same timezone for CI and local environments
+module.exports = async () => {
+  process.env.TZ = "UTC";
```
ixrock, line 8: Why should we consider timezone in tests at all? Would be easier just convert all dates to the same TZ (no matter which) before comparing?

Nokel81:
Without looking at the code, the last bit of the video confuses me. Why does it go grey? Is that you reloading the application?

aleksfront:
> Without looking at the code, the last bit of the video confuses me. Why does it go grey? Is that you reloading the application?

Yes, it was page reload, no bugs. Showing that update status still the same after page reload.

Nokel81:
I think this PR could easily be moved over to the work that was merged into master from https://github.com/lensapp/lens/pull/5433

src/jest.timezone.ts
jansav: Unit tests should have mocked date and timezone. Just to make them more deterministic. 
Would be great to have global test setup (in unit tests) which sets current time to `2015-10-21T07:28:00Z` and timezone to `UTC` of course. This would remove complete set of tests failing for a wrong reason.

Selection of "current time" fake is very important. It needs to be:
- easy to remember
- easy to distinct

In general, we should keep all dates UTC and convert only when showing to user. Common mistake is to convert dates too early. That makes a tricky topic a lot simpler.

jansav: Unit tests should have mocked date and timezone. Just to make them more deterministic. 
Would be great to have global test setup (in unit tests) which sets current time to `2015-10-21T07:28:00Z` and timezone to `UTC` of course. This would remove complete set of tests failing for a wrong reason.

Selection of "current time" fake is very important. It needs to be:
- easy to remember
- easy to distinct

Common mistake is to convert dates too early. We should keep all dates UTC and convert only when showing to user. That makes a tricky topic a lot simpler.

src/common/__tests__/timezones.test.ts
jansav: Only requirement for this test suite would be to act as documentation. It should carefully explain why we are faking timezone (and time, check other comment about setting determistic current time in test setup). 

Currently name of test doesn't give any explanation why we are doing this.

jansav:
Not proceeding with review till rebased with master.

Iku-turso:
There's a lot of cool and significant improvements to be made here, but let's wait for that rebase first. Shout when that happens? :)

src/renderer/app-update-warning/app-update-warning.injectable.ts
```diff
@@ -0,0 +1,23 @@
...
+import { AppUpdateWarning } from "./app-update-warning";
+
+const appUpdateWarningInjectable = getInjectable({
+  id: "app-update-warning",
+
+  instantiate: (di) => {
+    AppUpdateWarning.resetInstance();
+    
+    return AppUpdateWarning.createInstance({
+      ipcRenderer: di.inject(ipcRendererInjectable),
```
ixrock: Why `ipcRenderer` needs to be injectable?

Nokel81: So that in every unit test it can be a fresh mock. Though it is probably better if this was instead a `SyncBox` or some `IComputedValue` derived from the `SyncBox`'s that Janne and Mikko introduced in #5433

ixrock: Unit tests takes too much.. We end-up that way having injectable for everything what is "better for tests"..

jansav: We're moving to direction where `ipc` is going to be technical detail from the perspective of operational code and unit tests. We'll be creating easy-to-use tools on top of that (e.g. `SyncBox` that @Nokel81 mentioned) which make sure that you don't need to know anything about the technical implementation. 

Implementation here shouldn't use `ipcRenderer` at all. `IpcRenderer` or `IpcMain` is not welcome to operational code anymore. It's too low level implementation, which directly couples us to electron.

There is better alternatives for the job. Currently we have lower level implementations: `messageToChannel`, `requestFromChannel` and higher level implementation: `SyncBox` which can be utilized.

ixrock: I don't suggest to use `ipcRenderer` directly from electron's import.. But single abstraction in module/object/function would be enough imho.

Iku-turso: > Unit tests takes too much.. We end-up that way having injectable for everything what is "better for tests"..

Remember, unit testing alone is a lot, but it is by far not the only motivation to organise code like this (i.e. Dependency Inversion Principle). We also get stuff like simplicity, modularity, high cohesion, low coupling, composability, maintainability, reusability, clean abstractions, controllable side-effects, deterministic implementations, clean water, sandy beaches and margaritas.

jansav: > I don't suggest to use ipcRenderer directly from electron's import.. But single abstraction in module/object/function would be enough imho.

Single abstraction around IPC leads to re-inventing the wheel many times. That's why we are introducing tools like `SyncBox`. What's the downside of having e.g. `SyncBox` which makes sure that the injected value is exactly same in all processes, so that you don't have to worry about it anymore? 

Another example that comes to my mind is that you want to trigger function in other process. (we don't have easy-to-use tool for this yet, but could be implemented if/when requirement emerges). With correct tools, you can just inject the function and call it. Without the tools, you need to re-invent the wheel every time and figure out how you inform the other process that you wish to trigger something.

src/renderer/app-update-warning/__tests__/app-update-warning.test.ts
```diff
@@ -0,0 +1,169 @@
+/**
```
Iku-turso: Monseigneur, all the unit tests in this PR are technical, and really need to be written as behavioural instead :)

We could probably start considering the technical unit tests as rare exceptions, and behavioural ones as default.

src/renderer/app-update-warning/__tests__/app-update-warning.test.ts
```diff
@@ -0,0 +1,169 @@
...
+      ["2022-06-24T04:31:00.000Z"],
+    ];
+
+    const highDates = [
+      ["2022-06-24T05:31:00.000Z"],
+      ["2022-06-25T05:31:00.000Z"],
+      ["2027-05-28T05:31:00.000Z"],
+    ];
+
+    describe.each(lightDates)("given AppUpdateWarning with update date set to %s", (date) => {
```
Iku-turso: Name of test does not sufficiently explain relevance.

Iku-turso: `AppUpdateWarning` is a technical phenomena, and thus not welcome in name of unit test.

src/renderer/app-update-warning/__tests__/app-update-warning.test.ts
```diff
@@ -0,0 +1,169 @@
...
+      });
+    });
+
+    describe.each(mediumDates)("given AppUpdateWarning with update date set to %s", (date) => {
+      beforeAll(() => {
+        jest.useFakeTimers("modern");
+        jest.setSystemTime(new Date(date).getTime());
+      });
+
+      afterAll(() => {
```
Iku-turso: Nit: clean-up is not required?

src/renderer/app-update-warning/__tests__/app-update-warning.test.ts
aleksfront: Can you give an example how behavioural version of such tests would be look like? Will it be some difference if we try to test how warning level would change over time?

src/renderer/app-update-warning/__tests__/app-update-warning.test.ts
```diff
@@ -0,0 +1,169 @@
...
+  describe("given AppUpdateWarning with warning level changing over time", () => {
+    let appUpdateWarning: AppUpdateWarning;
+
+    beforeEach(() => {
+      jest.useFakeTimers("modern");
+      jest.setSystemTime(new Date("2022-05-30T05:30:00.000Z").getTime());
+      jest.spyOn(global, "setInterval");
+
+      appUpdateWarning = di.inject(appUpdateWarningInjectable);
+      appUpdateWarning.init();
```
Iku-turso: Description of `beforeEach` does not document this initialization.

src/renderer/app-update-warning/__tests__/app-update-warning.test.ts
```diff
@@ -0,0 +1,169 @@
...
+    
+    afterEach(() => {
+      jest.useRealTimers();
+    });
+
+    it("calls setInterval with correct arguments", () => {
+      const onceADay = 1000 * 60 * 60 * 24;
+
+      expect(setInterval).toHaveBeenCalledTimes(1);
+      expect(setInterval).toHaveBeenCalledWith(expect.any(Function), onceADay);
```
Iku-turso: Multiple expects per test prevent accurate name of test.

src/renderer/app-update-warning/__tests__/app-update-warning.test.ts
```diff
@@ -0,0 +1,169 @@
...
+    });
+    
+    afterEach(() => {
+      jest.useRealTimers();
+    });
+
+    it("calls setInterval with correct arguments", () => {
+      const onceADay = 1000 * 60 * 60 * 24;
+
+      expect(setInterval).toHaveBeenCalledTimes(1);
```
Iku-turso: If there's already fake timers, why not test this realistically by advancing it, instead of spying it? Something like `jest.advanceFakeTimers(onceADay)`.

Iku-turso: This tests nothing valuable :) Global `setInterval` could have been called by just about anything, even a third party library.

src/renderer/app-update-warning/__tests__/app-update-warning.test.ts
```diff
@@ -0,0 +1,169 @@
...
+
+      appUpdateWarning = di.inject(appUpdateWarningInjectable);
+      appUpdateWarning.init();
+    });
+    
+    afterEach(() => {
+      jest.useRealTimers();
+    });
+
+    it("calls setInterval with correct arguments", () => {
```
Iku-turso: Test does not test what name of test implies.

src/renderer/app-update-warning/__tests__/app-update-warning.test.ts
```diff
@@ -0,0 +1,169 @@
...
+      const onceADay = 1000 * 60 * 60 * 24;
+
+      expect(setInterval).toHaveBeenCalledTimes(1);
+      expect(setInterval).toHaveBeenCalledWith(expect.any(Function), onceADay);
+    });
+
+    it("shows light warning level if less than 20 days passed", () => {
```
Iku-turso: Improved naming convention ([given-when-then](https://en.wikipedia.org/wiki/Given-When-Then)) would be something like: "when less than 20 days passes, (then) still shows light warning level"

The ramifications of this are pretty considerable, but let's talk about that elsewhere cool? :)

src/renderer/app-update-warning/app-update-warning.injectable.ts
```diff
@@ -0,0 +1,23 @@
...
+import { getInjectable } from "@ogre-tools/injectable";
+import ipcRendererInjectable from "../app-paths/get-value-from-registered-channel/ipc-renderer/ipc-renderer.injectable";
+import sessionStorageInjectable from "../utils/session-storage.injectable";
+import { AppUpdateWarning } from "./app-update-warning";
+
+const appUpdateWarningInjectable = getInjectable({
+  id: "app-update-warning",
+
+  instantiate: (di) => {
+    AppUpdateWarning.resetInstance();
```
Iku-turso: `AppUpdateWarning` is a new global shared state, which we are trying to get rid of.

src/renderer/app-update-warning/app-update-warning.injectable.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import ipcRendererInjectable from "../app-paths/get-value-from-registered-channel/ipc-renderer/ipc-renderer.injectable";
```
Iku-turso: `ipcRendererInjectable` is a pretty low-level abstraction to be used in code like this. Use a new `MessageChannel` instead. There's also something called `SyncBox`.

src/renderer/app-update-warning/app-update-warning.ts
```diff
@@ -0,0 +1,120 @@
...
+import type { IpcRenderer } from "electron";
+
+interface Dependencies {
+  readonly ipcRenderer: IpcRenderer;
+  readonly sessionStorage: Storage;
+}
+
+const onceADay = 1000 * 60 * 60 * 24;
+
+export class AppUpdateWarning extends Singleton {
```
Iku-turso: Hmm. Baseclass `Singleton` should not be used for any new code, or at all. I thought it was deprecated, will check again.

src/renderer/app-update-warning/app-update-warning.ts
```diff
@@ -0,0 +1,120 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { makeObservable, observable } from "mobx";
+import { UpdateAvailableChannel } from "../../common/ipc";
```
Iku-turso: Such channel should reside in a feature-specific directory, instead of this common one.

src/renderer/app-update-warning/app-update-warning.ts
```diff
@@ -0,0 +1,120 @@
...
+import { Singleton } from "../utils";
+import type { IpcRenderer } from "electron";
+
+interface Dependencies {
+  readonly ipcRenderer: IpcRenderer;
+  readonly sessionStorage: Storage;
```
Iku-turso: Hmm. Shouldn't this be `localStorage` instead of `sessionStorage`? I'm not sure.

src/renderer/app-update-warning/app-update-warning.ts
```diff
@@ -0,0 +1,120 @@
...
+      clearInterval(this.interval);
+    }
+  }
+
+  private get daysAfterUpdateAvailable() {
+    if (!this.updateAvailableDate) {
+      return 0;
+    }
+
+    const today = Date.now();
```
Iku-turso: Explicit side-effect.

src/renderer/app-update-warning/app-update-warning.ts
```diff
@@ -0,0 +1,120 @@
...
+export class AppUpdateWarning extends Singleton {
+  @observable warningLevel: "high" | "medium" | "light" | "" = "";
+  @observable private updateAvailableDate: Date | null = this.getDateFromSessionStorage();
+  private interval: NodeJS.Timeout | null = null;
+
+  constructor(private dependencies: Dependencies) {
+    super();
+    makeObservable(this);
+
+    dependencies.ipcRenderer.on(UpdateAvailableChannel, () => {
```
Iku-turso: Construction time side-effect.

src/renderer/app-update-warning/app-update-warning.ts
```diff
@@ -0,0 +1,120 @@
+/**
```
Iku-turso: This entire implementation is stateful, when it could be reactive. This is a major root cause for considerable complexity and buggyness in the codebase.

src/renderer/app-update-warning/app-update-warning.ts
```diff
@@ -0,0 +1,120 @@
...
+    this.startRefreshLevelInterval();
+  }
+
+  private saveDateToSessionStorage() {
+    if (this.updateAvailableDate) {
+      this.dependencies.sessionStorage.setItem("when-update-available", this.updateAvailableDate.toISOString());
+    }
+  }
+
+  private getDateFromSessionStorage() {
```
Iku-turso: Too many responsibilities indicates a god class.

src/renderer/components/update-button/update-button.tsx
```diff
@@ -11,13 +11,21 @@ import { Menu, MenuItem } from "../menu";
...
+import appUpdateWarningLevelInjectable from "../../app-update-warning/app-update-warning-level.injectable";
+import type { IComputedValue } from "mobx";
+import quitAndInstallUpdateInjectable from "../../../main/electron-app/features/quit-and-install-update.injectable";
 
 interface UpdateButtonProps extends HTMLAttributes<HTMLButtonElement> {
-  warningLevel?: "light" | "medium" | "high";
+}
+
+interface Dependencies {
+  warningLevel?: IComputedValue<"light" | "medium" | "high" | "">;
```
Iku-turso: This is already nullable, why the empty string? :) Looks like it shouldn't even be nullable?

src/renderer/utils/session-storage.injectable.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+ 
+const sessionStorageInjectable = getInjectable({
```
Iku-turso: This should be `causesSideEffects: true`.

Iku-turso (dismissed):
Hey, I just noticed my comments coming out as a bit grumpy 😱 Don't worry about that, and consider them more as notes to myself. There's IMHO too many and too big things to properly address here, so how about we take a look at that stuff together, hmm? :)

Here's the main gripes:
1. Tests are not behavioural
2. Tests are not generally up to standard
3. Statefulness and polling is being used over reactivity
4. Usages of Singleton, shared global state and a store-like god class

src/renderer/app-update-warning/__tests__/app-update-warning.test.ts
Iku-turso: > Can you give an example how behavioural version of such tests would be look like?

I could if you really want, but are the current behaviours not a sufficient example already? I propose in a later comment that we look at this together :) So how about it, bud? :)

> Will it be some difference if we try to test how warning level would change over time?

Deeply sorry, I don't understand :(

src/renderer/app-update-warning/app-update-warning.injectable.ts
ixrock: I don't think simple abstraction-layer/wrapper is re-inventing the wheel, there is nothing to invent, it does the job and incapsulates "native api" where we hide it and keep simple and in single place / source-of-truth.
That way  at some point you can replace for example `IPC` to `REST` in single module/place without a hassle. 
If framework/tools becomes too opinionated they becomes burden.

> With correct tools, you can just inject the function and call it

With correct implementation you just import module and it just works™

src/common/__tests__/timezones.test.ts
jansav: This test isn't deterministic. It fails e.g. on my computer.

jansav (changes requested):
Once test for timezones is fixed, I'm happy with this.

src/renderer/app-update-warning/app-update-warning.injectable.ts
jansav: > I don't think simple abstraction-layer/wrapper is re-inventing the wheel, there is nothing to invent, it does the job and incapsulates "native api" where we hide it and keep simple and in single place / source-of-truth.
That way at some point you can replace for example IPC to REST in single module/place without a hassle.
If framework/tools becomes too opinionated they becomes burden.

That's what we are doing here. Instead of introducing single abstraction layer for a tool, we are introducing one abstraction layer for purpose and using a tool to implement that abstraction. That means that we are able to change our mind which tool we are using for what purpose. One purpose at the time.

> With correct implementation you just import module and it just works™

Well, that would lead to place where you need to specify the dependencies when called. If that's not the case, we couldn't depend from the interfaces and would have to depend from the implementations. Therefore we wouldn't be able to replace one implementation with something else which just fulfills the required interface.

One common used example is how you access database. You shouldn't depend from e.g. DynamoDB, you should depend from interface that gets implemented with DynamoDB stuff. (or file system, in-memory, whatever.)

jansav (dismissed):

src/renderer/components/update-button/update-button.tsx
```diff
@@ -26,19 +33,19 @@ export function UpdateButton({ warningLevel, update, id }: UpdateButtonProps) {
     setOpened(!opened);
   };
 
-  if (!warningLevel) {
+  if (!warningLevel.get()) {
```
Nokel81: nit: this value should be put into a variable

src/renderer/components/update-button/update-warning-level.injectable.ts
```diff
@@ -0,0 +1,45 @@
...
+  id: "update-warning-level",
+
+  instantiate: (di) => {
+    const updateDownloadedDateTime = di.inject(updateDownloadedDateTimeInjectable);
+
+    return computed(() => {
+      const downloadedAt = updateDownloadedDateTime.value.get();
+
+      if (!downloadedAt) {
+        return "";
```
Nokel81, line 20: nit: We probably shouldn't be using `""` as the sentinel value.

Nokel81 (dismissed):

src/renderer/components/update-button/update-warning-level.injectable.ts
aleksfront: It converts safely to falsey value in `UpdateButton` condition.

```
if (!warningLevel.get()) {
```

Nokel81 (approved):

jansav (approved):

PR #5449, opened 2022-05-25
Make terminal tab icon the same as the menu item icon

Nokel81:
Just something off than I noticed.

before:
Screen Shot 2022-05-25 at 3 33 43 PM

after:
Screen Shot 2022-05-25 at 3 32 07 PM


aleksfront (changes requested):
I would invert your change and use material icon version inside the `Terminal` menu item. Svg version is non-standart and was introduced by the times when material icons haven't got own `terminal` icon yet.

Svg terminal icon would be eliminated for the next major version.

Nokel81:
@aleksfront Here is how it looks with your recommendations
Screen Shot 2022-05-26 at 9 14 17 AM

aleksfront (approved):

ixrock (approved):

PR #5451, opened 2022-05-25
Fix completely overriding PATH in shellSync

### LINKED ISSUE #4931: Shell sync overrides modified PATH
During shell sync we override the PATH variable on non-snap environments. However, if the user has modified the PATH, for instance by setting up a virtual environment for certain dependencies, this gets overridden.

We should instead do some sort of concat or maybe merge.

Nokel81:
- This allows local changes to PATH in the shell that is used to open
  lens to persist after shellSync.

- This helps with using python virtual environments

fixes #4931

### DIFF
--- a/src/common/utils/__tests__/union-env-path.test.ts
+++ b/src/common/utils/__tests__/union-env-path.test.ts
@@ -1,0 +1,32 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import path from "path";
+import { unionPATHs } from "../union-env-path";
+
+describe("unionPATHs", () => {
+  it("return the same path if given only one with no double delimiters", () => {
+    expect(unionPATHs(`/bin/bar${path.delimiter}/usr/bin`)).toBe(`/bin/bar${path.delimiter}/usr/bin`);
+  });
+
+  it("return equivalent path if given only one with no double delimiters", () => {
+    expect(unionPATHs(`/bin/bar${path.delimiter}${path.delimiter}/usr/bin`)).toBe(`/bin/bar${path.delimiter}/usr/bin`);
+  });
+
+  it("should remove duplicate entries, appending non duplicates in order received", () => {
+    expect(unionPATHs(
+      `/bin/bar${path.delimiter}/usr/bin`,
+      `/bin/bar${path.delimiter}/usr/lens/bat`,
+    )).toBe(`/bin/bar${path.delimiter}/usr/bin${path.delimiter}/usr/lens/bat`);
+  });
+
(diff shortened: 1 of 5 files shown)


src/common/utils/iter.ts
```diff
@@ -233,3 +233,18 @@ export function every<T>(src: Iterable<T>, fn: (val: T) => any): boolean {
 
   return true;
 }
+
+/**
+ * Takes two iterators and creates a new iterator over both in sequence.
+ * @param first The first iterable to iterate over
+ * @param second The second iterable to iterate over
+ */
+export function* chain<T>(first: Iterable<T>, second: Iterable<T>): Iterable<T> {
```
ixrock: It's unused in PR?

Nokel81: Good catch, I added this in a previous iteration of `unionPATHs`, will remove.

src/main/shell-sync.ts
```diff
@@ -25,7 +26,8 @@ export async function shellSync() {
   }
 
   if (!isSnap) {
-    process.env.PATH = env.PATH;
+    // Prefer the synced PATH over the initial one
+    process.env.PATH = unionPATHs(env.PATH ?? "",  process.env.PATH ?? "");
```
ixrock, line 30: What happens if both `env.PATH` and `process.env.PATH` are defined?  Or it's expected to be merged if so?

ixrock (approved):

src/main/shell-sync.ts
Nokel81: That is expected, and yes it is expected that they will be merged.

Furthermore, the entries within each will be de-duplicated as well.

src/common/utils/__tests__/union-env-path.test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import path from "path";
+import { unionPATHs } from "../union-env-path";
+
+describe("unionPATHs", () => {
+  it("return the same path if given only one with no double delimiters", () => {
+    expect(unionPATHs(`/bin/bar${path.delimiter}/usr/bin`)).toBe(`/bin/bar${path.delimiter}/usr/bin`);
+  });
+
+  it("return equivalent path if given only one with no double delimiters", () => {
```
jim-docker, line 14: should this read:
```suggestion
  it("return equivalent path if given only one with double delimiters", () => {
```

Nokel81:
Closing for the time being as neither @jim-docker nor I can reproduce the linked issue.

jim-docker (approved):
Couldn't actually reproduce the bug to confirm this fixes it. Regardless this shouldn't hurt anything(?)

src/common/utils/__tests__/union-env-path.test.ts
Nokel81: Will fix in follow up PR

PR #5479, opened 2022-05-27
Fix: app-crash with multiple usages of active MonacoEditor component

ixrock:
How to reproduce: 
- open one pod on monaco editor (ie edit), and then try to open another pod details (or try to edit it as well)

Partially reverted from #5264 

image


aleksfront (approved):
Quite surprising!

nevalla (approved):
Tested and works fine now.

ksa-real:
Can you do an emergency release?

sinchanroy:
What needs to be done from userside to solve this ?  Do we download the latest release ?

ksa-real:
I guess users will auto-update as usual. But generally, it would be nice to somehow be able to roll back in situations like that, e.g. by downloading an older version of the app. Owners could force roll-back from their end as well.

Nokel81:
Yes, we will do another emergency release today in a few hours

PR #5483, opened 2022-05-27
Remove legacy renderBooleans prop

Nokel81:
- Replace with direct serialization at source

- Add tests to verify behaviour


src/common/k8s-api/endpoints/types/json-schema-props.ts
```diff
@@ -0,0 +1,93 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import type { JsonValue } from "type-fest";
+import type { ExternalDocumentation } from "./external-documentation";
+
+export interface JSONSchemaProps {
```
Nokel81, line 9: I added these types to make writing the unit tests easier.

src/renderer/components/drawer/drawer-item.tsx
```diff
@@ -5,14 +5,13 @@
...
 import React from "react";
-import { cssNames, displayBooleans } from "../../utils";
+import { cssNames } from "../../utils";
 
 export interface DrawerItemProps extends React.HTMLAttributes<HTMLDivElement> {
   name: React.ReactNode;
   title?: string;
   labelsOnly?: boolean;
   hidden?: boolean;
-  renderBoolean?: boolean; // show "true" or "false" for all of the children elements are "typeof boolean"
```
Nokel81, line 15: Technically this is a breaking change (when building an extension) but I think that since this really was an internal hack we can remove it.

src/renderer/components/+custom-resources/__tests__/custom-resource-details.test.tsx
```diff
@@ -0,0 +1,198 @@
...
+  let render: DiRender;
+
+  beforeEach(() => {
+    const di = getDiForUnitTesting({ doGeneralOverrides: true });
+
+    render = renderFor(di);
+  });
+
+  describe("with a CRD with a boolean field", () => {
+    const crd = new CustomResourceDefinition({
```
jansav: This `crd` should be in `beforeEach`. It's now shared between tests. Just to limit possibility of one test affecting other test.

Nokel81: Sounds good

Nokel81:
@jansav PTALA

ixrock (dismissed):

src/renderer/components/drawer/drawer-item.tsx
jim-docker: It is described in the API reference, can you just mark it here as deprecated (and now no longer used)?

jim-docker (dismissed):

jim-docker (approved):

ixrock (approved):

PR #5484, opened 2022-05-27
Fix error in EditResource 

Nokel81:

This bug doesn't exist on master because of #5195

### DIFF
--- a/src/renderer/components/dock/edit-resource/view.tsx
+++ b/src/renderer/components/dock/edit-resource/view.tsx
@@ -45,8 +45,8 @@ class NonInjectedEditResource extends React.Component<EditResourceProps & Depend
         const tabData = this.props.editResourceStore.getData(this.props.tab.id);
         const obj = this.resource;
 
-        if (!obj) {
-          if (store?.isLoaded) {
+        if (!obj && store) {
+          if (store.isLoaded) {
             // auto-close tab when resource removed from store
             this.props.closeTab(this.props.tab.id);
           } else if (!store.isLoading) {


src/renderer/components/dock/edit-resource/view.tsx
```diff
@@ -45,8 +45,8 @@ class NonInjectedEditResource extends React.Component<EditResourceProps & Depend
         const tabData = this.props.editResourceStore.getData(this.props.tab.id);
         const obj = this.resource;
 
-        if (!obj) {
-          if (store?.isLoaded) {
+        if (!obj && store) {
+          if (store.isLoaded) {
             // auto-close tab when resource removed from store
             this.props.closeTab(this.props.tab.id);
           } else if (!store.isLoading) {
```
Nokel81, line 52: The actual error reported was on this line.

ixrock (approved):

src/renderer/components/dock/edit-resource/view.tsx
```diff
@@ -45,8 +45,8 @@ class NonInjectedEditResource extends React.Component<EditResourceProps & Depend
         const tabData = this.props.editResourceStore.getData(this.props.tab.id);
         const obj = this.resource;
 
-        if (!obj) {
-          if (store?.isLoaded) {
+        if (!obj && store) {
+          if (store.isLoaded) {
```
jim-docker, line 49: So the bug was that if store was null/undefined then execution shouldn't have gone to the `else` block? These are tricky to see, and made more common by the convenience of the `?.` operator

jim-docker (approved):

src/renderer/components/dock/edit-resource/view.tsx
Nokel81: No the bug was that if the `store` was `null` or `undefined` nothing should have happened, but instead it did go to the `else` block.

jim-docker: thanks (I meant shouldn't, corrected)

PR #5488, opened 2022-05-27
Ensure the cluster dashboard start url is valid

jim-docker:
ensure the start url is the workloads overview when the cluster overview is not available due to rbac restrictions

fixes #5487
fixes #5459
fixes #5462

~~This may also fix #5459 / #5462 though there is a separate PR coming for that. I think that PR should duplicate these changes to `start-url.injectable.ts`.~~

PR #5518, opened 2022-05-31
Fix cluster frame display issue

Nokel81:
- Add some defensive code to prevent this sort of infinite loop

- Add some unit tests

This is the version of https://github.com/lensapp/lens/pull/5488 for `master`. It includes some unit tests.


src/common/__tests__/cluster-store.test.ts
```diff
@@ -94,6 +95,7 @@ describe("cluster-store", () => {
     mainDi.permitSideEffects(getConfigurationFileModelInjectable);
     mainDi.permitSideEffects(appVersionInjectable);
     mainDi.permitSideEffects(clusterStoreInjectable);
+    mainDi.permitSideEffects(fsInjectable);
```
jansav, line 98: Doesn't seem right, but neither does `mainDi.permitSideEffects(getConfigurationFileModelInjectable);`. Guess there is good reason for it for now :)

src/common/cluster-store/cluster-store.ts
```diff
@@ -111,12 +115,24 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
     return undefined;
   }
 
+  private createNewCluster(model: ClusterModel): Cluster {
```
jansav: Not welcome here at all. This logic belongs to `createCluster` dependency function.

src/main/catalog-sources/__test__/kubeconfig-sync.test.ts
```diff
@@ -54,10 +55,9 @@ describe("kubeconfig-sync.source tests", () => {
     di.override(kubectlDownloadingNormalizedArchInjectable, () => "amd64");
     di.override(normalizedPlatformInjectable, () => "darwin");
 
-    di.override(clusterStoreInjectable, () =>
-      ClusterStore.createInstance({ createCluster: () => null as never }),
-    );
-
+    di.permitSideEffects(fsInjectable);
```
jansav, line 58: Doesn't seem right to introduce real `fs` to unit tests.

src/renderer/frames/cluster-frame/cluster-frame.tsx
```diff
@@ -62,7 +68,16 @@ export const NonInjectedClusterFrame = observer(({
         {
           Component
             ? <Component />
-            : <Redirect to={startUrl.get()} />
+            // NOTE: this check is to prevent an infinite loop
```
jansav, line 71: I don't follow. How could anyone encounter this?

src/renderer/frames/cluster-frame/start-url.injectable.ts
```diff
@@ -4,29 +4,30 @@
...
+      }
+
+      if (workloadOverviewRoute.isEnabled.get()) {
+        return workloadOverviewRoute.path;
+      }
 
-      return resources.every((resourceName) => isAllowedResource(resourceName))
-        ? clusterOverviewUrl
-        : workloadOverviewUrl;
+      /**
```
jansav, line 26: I don't follow. Why is this needed?

src/common/__tests__/cluster-store.test.ts
Nokel81: Well until we factor out all of this (and probably the library that we are using) we have to continue to use `mockFs`.

src/main/catalog-sources/__test__/kubeconfig-sync.test.ts
Nokel81: I am not, this test previously (and still does) use `mockFs`. I just changed how it is used.

src/renderer/frames/cluster-frame/start-url.injectable.ts
Nokel81: 1. Because `workloadOverviewRoute.isEnabled` is typed as `IComputedValue<boolean>` it *could* be `false` so this is a branch of code that _could_ be run.
2. The other option is to just not check `workloadOverviewRoute.isEnabled` at all and then always return `workloadOverviewRoute.path` but then if in the future we change the definition of `workloadOverviewRoute.isEnabled` to not always be defined we might get into the same case again.

src/renderer/frames/cluster-frame/cluster-frame.tsx
Nokel81: People did: https://github.com/lensapp/lens/issues/5462

The bug was that there was a disconnect between the `Redirect` and whether `startUrl` was valid.

jansav: Hmm. I'd say this kludge to get rid of the problem. However, this doesn't solve the real issue, which is to figure out why this is possible and make it impossible. 

Maybe good enough for now.

jansav (approved):

src/renderer/frames/cluster-frame/cluster-frame.tsx
Nokel81: I did fix the problem as well. This is here to prevent the same problem from occurring again in the future.

src/common/cluster-types.ts
```diff
@@ -67,6 +67,15 @@ export interface ClusterModel {
   labels?: Record<string, string>;
 }
 
+/**
+ * This data is retreived from the kubeconfig file before calling the cluster constructor.
+ *
+ * That is done to remove the external dependency on the construction of Cluster instances.
```
jim-docker, line 73: Yay!

src/renderer/frames/cluster-frame/cluster-frame.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+    di.override(directoryForUserDataInjectable, () => "/some/irrelavent/path");
+    di.override(storesAndApisCanBeCreatedInjectable, () => true);
+
+    const createCluster = di.inject(createClusterInjectable);
+
+    cluster = createCluster(
+      {
+        contextName: "my-cluster",
+        id: "123456",
+        kubeConfigPath: "/irrelavent",
```
jim-docker, line 54: I know it's irrelevant but it's irrelevant :)

jim-docker (approved):

src/renderer/frames/cluster-frame/cluster-frame.test.tsx
Nokel81: LoL

PR #5519, opened 2022-05-31
Fix Tooltip not showing when switching hotbars

### LINKED ISSUE #5517: Hotbar switching no longer displays name as tooltip
Regression after https://github.com/lensapp/lens/pull/4679

The tooltip which should show up when switching hotbars via the arrows no longer shows up in 5.5.2

Nokel81:

fixes #5517

### DIFF
--- a/src/renderer/components/tooltip/tooltip.tsx
+++ b/src/renderer/components/tooltip/tooltip.tsx
@@ -10,7 +10,7 @@ import { createPortal } from "react-dom";
 import { observer } from "mobx-react";
 import type { IClassName } from "../../utils";
 import { cssNames, autoBind } from "../../utils";
-import { observable, makeObservable, action, computed } from "mobx";
+import { observable, makeObservable, action } from "mobx";
 
 export enum TooltipPosition {
   TOP = "top",
@@ -62,10 +62,6 @@ export class Tooltip extends React.Component<TooltipProps> {
     super(props);
     makeObservable(this);
     autoBind(this);
-  }
-
-  @computed get renderToDOM() {
-    return this.props.visible ?? this.isVisible;
   }
 
   get targetElem(): HTMLElement | null {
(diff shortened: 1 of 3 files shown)


jakolehm (dismissed):

src/renderer/components/tooltip/tooltip.tsx
```diff
@@ -64,6 +64,10 @@ export class Tooltip extends React.Component<TooltipProps> {
     autoBind(this);
   }
 
+  @computed get renderToDOM() {
```
ixrock: Looks like a method name. 
Moreover I'd like to keep this more simple. We have already `this.isContentVisible` and `this.isVisible`, isn't it too much for about almost same thing?

So the only thing we should check in render before showing actual DOM should be like: 

```tsx
render() {
 if(!this.isVisible || (this.isVisible && !this.isContentVisible)) return null;
 // otherwise render content
}
```

But we shouldn't reset `isVisible` immedialtely on mouse leave (because of css-animation, see `transition` in `tooltip.scss`).
We should reset it only after animation ends, kinda:

```tsx
  // another more detailed example

  @observable isVisible = Boolean(this.props.visible);
  @observable className = { enter: false, leave: false };

  @action
  enter() {
    this.isVisible = true; // triggers render() to apply css-animation in existing dom
    this.className.leave = false; // reset if in the middle of previous mouseleave animation

    requestAnimationFrame(() => {
      this.className.enter = true;
    });
  }

  @action
  leave() {
    if (!this.isVisible) return;
    this.className.leave = true;
  }

  @action
  reset() {
    this.isVisible = false;
    this.className.enter = false;
    this.className.leave = false;
  }

  onTransitionEnd(evt: React.TransitionEvent) {
    var { enter, leave } = this.className;
    if (enter && leave) {
      this.reset();
    }
  }

  render() {
    var { isVisible, className } = this;

    if (!isVisible && !className.leave) {
      return null; // exit
    }
    // otherwise render
```

Nokel81: I can change the name, fair enough it does sound like a method. We can do some more cleanup later, I didn't want to touch this too much.

Nokel81: I have removed this getter and just inlined it.

aleksfront (approved):
LGTM

ixrock (approved):

PR #5520, opened 2022-05-31
Fix memoryUsage display for Nodes

### LINKED ISSUE #5457: Pods Memory in deployment view - per pod view is shown as Mib instead of GiB
open deployment view - per pod view the memory is shown as MiB instead of Gib but the number is correct.

lens-memory issue

lens version - Lens: 5.5.0-latest.20220525.1

Nokel81:

fixes #5457

### DIFF
--- a/src/renderer/components/+nodes/route.tsx
+++ b/src/renderer/components/+nodes/route.tsx
@@ -21,7 +21,6 @@ import { KubeObjectStatusIcon } from "../kube-object-status-icon";
 import { Badge } from "../badge/badge";
 import { eventStore } from "../+events/legacy-store";
 import { makeObservable, observable } from "mobx";
-import isEmpty from "lodash/isEmpty";
 import { KubeObjectAge } from "../kube-object/age";
 
 enum columnId {
@@ -48,7 +47,7 @@ interface UsageArgs {
 
 @observer
 export class NodesRoute extends React.Component {
-  @observable.ref metrics: Partial<NodeMetricData> = {};
+  @observable.ref metrics?: NodeMetricData;
   private metricsWatcher = interval(30, async () => this.metrics = await getMetricsForAllNodes());
 
   constructor(props: any) {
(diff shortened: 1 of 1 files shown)


src/renderer/components/+nodes/route.tsx
```diff
@@ -124,7 +125,7 @@ export class NodesRoute extends React.Component {
     return this.renderUsage({
       node,
       title: "Memory",
-      metricNames: ["workloadMemoryUsage", "memoryAllocatableCapacity"],
+      metricNames: ["memoryUsage", "memoryAllocatableCapacity"],
```
aleksfront, line 128: Is it really fixes the issue? Does it mean prometheus brings `workloadMemoryUsage` metric with **exact** same numbers but with different multipliers as `memoryUsage`? The meaning of `workloadMemoryUsage` is unclear to me then.

What I mean is:
`workloadMemoryUsage` gives us `11.2 MiB`
and
`memoryUsage` gives us `11.2 GiB`?

Nokel81: No it is not the exact same. But it is the exact same metric we use for the other graphs.

ixrock (approved):

src/renderer/components/+nodes/route.tsx
aleksfront: > No it is not the exact same

Need to clarify this. If it not the same then numbers would be different. But example here https://github.com/lensapp/lens/issues/5457 shows that only suffix should be changed. How we can be 100% sure that changing metric actually fixing the issue?

Nokel81: It might be in his case that it is exactly 8x smaller (modulo rounding) but in my testing it wasn't exactly that ratio. However what I did find is that this is our only use of `workloadMemoryUsage` whereas on the cluster overview we use `memoryUsage`

Nokel81: Maybe @nevalla remembers the difference (I am not an expert on Prometheus).

Nokel81:
Closing as this is not actually the bug reported...

PR #5531, opened 2022-06-01
Prevent duplicate app windows

Iku-turso:
Note: base-branch of this PR is #5433 instead of master, because the requirements of unit testing are introduced in the branch.

Note: this fixes bug about multiple application windows opening if user reopens Lens using tray while the splash screen was still visible. The bug was reproduced in a unit test and some behaviours around starting and stopping Lens where created.

Relevant test:
`opening application window using tray › given application has started › when the application window is closed › when an application window is reopened using tray › given opening of splash window has not finished yet, but another attempt to open the application is made › does not open any new windows`

### DIFF
--- a/src/behaviours/quitting-and-restarting-the-app/opening-application-window-using-tray.test.ts
+++ b/src/behaviours/quitting-and-restarting-the-app/opening-application-window-using-tray.test.ts
@@ -86,6 +86,18 @@ describe("opening application window using tray", () => {
           expectWindowsToBeOpen([]);
         });
 
+        describe("given opening of splash window has not finished yet, but another attempt to open the application is made", () => {
+          beforeEach(() => {
+            createElectronWindowMock.mockClear();
+
+            applicationBuilder.tray.click("open-app");
+          });
+
+          it("does not open any new windows", () => {
+            expect(createElectronWindowMock).not.toHaveBeenCalled();
+          });
+        });
+
         describe("when opening of splash window resolves", () => {
           beforeEach(async () => {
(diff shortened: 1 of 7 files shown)


src/behaviours/quitting-and-restarting-the-app/opening-application-window-using-tray.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+        describe("when opening of splash window resolves", () => {
+          beforeEach(async () => {
+            await resolveOpeningOfWindow("splash");
+          });
+
+          it("still only splash window is open", () => {
+            expectWindowsToBeOpen(["splash"]);
+          });
+
```
Nokel81, line 109: I would like to see another test along the lines of "given another attempt to open the application is made > shows window immediately without waiting for it to load".

I believe that is what the previous behaviour was and is useful when you get a crash before the window has reported that it is loaded, that way we can open the dev tools and look at the logs.

src/main/start-main-application/lens-window/application-window/create-lens-window.injectable.ts
```diff
@@ -35,31 +35,45 @@ const createLensWindowInjectable = getInjectable({
...
+      let windowIsOpening = false;
+
       return {
+        id: configuration.id,
+
         get visible() {
           return !!browserWindow;
         },
+
+        get opening() {
```
Nokel81, line 58: nit: `isOpening` might read a bit better.

PR #5532, opened 2022-06-01
Fix *.modules.scss that don't get correctly typed

Nokel81:

This was only a problem within vscode (and maybe other editors) which is why I marked it as a `chore`.

### DIFF
--- a/types/mocks.d.ts
+++ b/types/mocks.d.ts
@@ -10,6 +10,14 @@ declare module "@hapi/subtext"
 
 // Support import for custom module extensions
 // https://www.typescriptlang.org/docs/handbook/modules.html#wildcard-module-declarations
+declare module "*.module.scss" {
+  const classes: { [key: string]: string };
+  export default classes;
+}
+declare module "*.module.css" {
+  const classes: { [key: string]: string };
+  export default classes;
+}
 declare module "*.scss" {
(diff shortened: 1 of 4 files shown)


types/mocks.d.ts
```diff
@@ -10,14 +10,6 @@ declare module "@hapi/subtext"
 
 // Support import for custom module extensions
 // https://www.typescriptlang.org/docs/handbook/modules.html#wildcard-module-declarations
-declare module "*.module.scss" {
```
Nokel81: Since we have that package which types them we no longer need these.

aleksfront (approved):
LGTM

ixrock (approved):
Why only these files? Looks like a hack.

Nokel81:
@ixrock I looked at the typescript server logs in vscode to figure out why a specific file was being typed as `{}` instead of being typed as `{ AddCluster: string; }`. It reported an error from the CSS modules plugin for typescript saying that the `theme-light` mixin was undefined and that it considered the given file the "root file" for that use.

So I figured that importing the file that defines the `theme-light` mixin would help solve the issue. 

I also tried getting the plugin to accept the mixin file as "always imported" but I couldn't figure out how. I also don't know how the actual transpiler does it.

PR #5533, opened 2022-06-01
Prevent duplicate app windows

jansav:
Note: base-branch of this PR is https://github.com/lensapp/lens/pull/5433 instead of master, because the requirements of unit testing are introduced in the branch.

Note: this fixes bug about multiple application windows opening if user reopens Lens using tray while the splash screen was still visible. The bug was reproduced in a unit test and some behaviours around starting and stopping Lens where created.

Relevant test:
opening application window using tray › given application has started › when the application window is closed › when an application window is reopened using tray › given opening of splash window has not finished yet, but another attempt to open the application is made › does not open any new windows

### DIFF
--- a/src/behaviours/quitting-and-restarting-the-app/opening-application-window-using-tray.test.ts
+++ b/src/behaviours/quitting-and-restarting-the-app/opening-application-window-using-tray.test.ts
@@ -86,6 +86,18 @@ describe("opening application window using tray", () => {
           expectWindowsToBeOpen([]);
         });
 
+        describe("given opening of splash window has not finished yet, but another attempt to open the application is made", () => {
+          beforeEach(() => {
+            createElectronWindowMock.mockClear();
+
+            applicationBuilder.tray.click("open-app");
+          });
+
+          it("does not open any new windows", () => {
+            expect(createElectronWindowMock).not.toHaveBeenCalled();
+          });
+        });
+
         describe("when opening of splash window resolves", () => {
           beforeEach(async () => {
             await resolveOpeningOfWindow("splash");
@@ -101,7 +113,7 @@ describe("opening application window using tray", () => {
             expectWindowsToBeOpen(["only-application-window"]);
           });
 
-          describe("given opening has not finished yet, but another attempt to open the application is made", () => {
+          describe("given opening of application window has not finished yet, but another attempt to open the application is made", () => {
             beforeEach(() => {
               createElectronWindowMock.mockClear();
 
--- a/src/behaviours/quitting-and-restarting-the-app/opening-application-window-using-tray.test.ts
+++ b/src/behaviours/quitting-and-restarting-the-app/opening-application-window-using-tray.test.ts
@@ -1,0 +1,151 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import type { ApplicationBuilder } from "../../renderer/components/test-utils/get-application-builder";
+import { getApplicationBuilder } from "../../renderer/components/test-utils/get-application-builder";
+import { lensWindowInjectionToken } from "../../main/start-main-application/lens-window/application-window/lens-window-injection-token";
+import applicationWindowInjectable from "../../main/start-main-application/lens-window/application-window/application-window.injectable";
+import createElectronWindowForInjectable from "../../main/start-main-application/lens-window/application-window/create-electron-window-for.injectable";
+import type { AsyncFnMock } from "@async-fn/jest";
+import asyncFn from "@async-fn/jest";
+
+import type {
+  ElectronWindow,
+  LensWindowConfiguration,
+} from "../../main/start-main-application/lens-window/application-window/create-lens-window.injectable";
+
+import { flushPromises } from "../../common/test-utils/flush-promises";
+import type { DiContainer } from "@ogre-tools/injectable";
+
+describe("opening application window using tray", () => {
+  describe("given application has started", () => {
+    let applicationBuilder: ApplicationBuilder;
+
+    let createElectronWindowMock: AsyncFnMock<
+      (configuration: LensWindowConfiguration) => ElectronWindow
+    >;
+
+    let expectWindowsToBeOpen: (windowIds: string[]) => void;
+    let resolveOpeningOfWindow: (windowId: string) => Promise<void>;
+
+    beforeEach(async () => {
+      applicationBuilder = getApplicationBuilder().beforeApplicationStart(
+        ({ mainDi }) => {
+          createElectronWindowMock = asyncFn();
+
+          mainDi.override(
+            createElectronWindowForInjectable,
+
+            () => (configuration) => () =>
+              createElectronWindowMock(configuration),
+          );
+
+          expectWindowsToBeOpen = expectWindowsToBeOpenFor(mainDi);
+
+          resolveOpeningOfWindow = resolveOpeningOfWindowFor(
+            createElectronWindowMock,
+          );
+        },
+      );
+
+      const renderPromise = applicationBuilder.render();
+
+      await flushPromises();
+
+      await resolveOpeningOfWindow("splash");
+      await resolveOpeningOfWindow("only-application-window");
+
+      await renderPromise;
+    });
+
+    it("only an application window is open", () => {
+      expectWindowsToBeOpen(["only-application-window"]);
+    });
+
+    describe("when the application window is closed", () => {
+      beforeEach(() => {
+        const applicationWindow = applicationBuilder.dis.mainDi.inject(
+          applicationWindowInjectable,
+        );
+
+        applicationWindow.close();
+      });
+
+      it("no windows are open", () => {
+        expectWindowsToBeOpen([]);
+      });
+
+      describe("when an application window is reopened using tray", () => {
+        beforeEach(() => {
+          applicationBuilder.tray.click("open-app");
+        });
+
+        it("still no windows are open", () => {
+          expectWindowsToBeOpen([]);
+        });
+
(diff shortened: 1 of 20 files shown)


Iku-turso (dismissed):

src/behaviours/quitting-and-restarting-the-app/opening-application-window-using-tray.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+        describe("when opening of splash window resolves", () => {
+          beforeEach(async () => {
+            await resolveOpeningOfWindow("splash");
+          });
+
+          it("still only splash window is open", () => {
+            expectWindowsToBeOpen(["splash"]);
+          });
+
```
Nokel81, line 202: I would like to see another test along the lines of "given another attempt to open the application is made > shows window immediately without waiting for it to load".

I believe that is what the previous behaviour was and is useful when you get a crash before the window has reported that it is loaded, that way we can open the dev tools and look at the logs.

src/main/start-main-application/lens-window/application-window/create-lens-window.injectable.ts
```diff
@@ -35,31 +35,45 @@ const createLensWindowInjectable = getInjectable({
...
+      let windowIsOpening = false;
+
       return {
+        id: configuration.id,
+
         get visible() {
           return !!browserWindow;
         },
+
+        get opening() {
```
Nokel81: nit: isOpening might read a bit better.

src/behaviours/quitting-and-restarting-the-app/opening-application-window-using-tray.test.ts
jansav: Hm, sounds definitely like something that we should still have.

Nokel81: Doesn't look like it has been added yet...

jansav: Yes, it hasn't. We'll be working on this tomorrow.

src/behaviours/quitting-and-restarting-the-app/opening-application-window-using-tray.test.ts
```diff
@@ -0,0 +1,163 @@
...
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import type { ApplicationBuilder } from "../../renderer/components/test-utils/get-application-builder";
+import { getApplicationBuilder } from "../../renderer/components/test-utils/get-application-builder";
+import { lensWindowInjectionToken } from "../../main/start-main-application/lens-window/application-window/lens-window-injection-token";
+import applicationWindowInjectable from "../../main/start-main-application/lens-window/application-window/application-window.injectable";
+import createElectronWindowForInjectable from "../../main/start-main-application/lens-window/application-window/create-electron-window-for.injectable";
+import type { AsyncFnMock } from "@async-fn/jest";
+import asyncFn from "@async-fn/jest";
```
ixrock, line 12: _nit:_ 

```suggestion
import asyncFn, { type AsyncFnMock } from "@async-fn/jest";
```

ixrock (dismissed):

src/behaviours/quitting-and-restarting-the-app/opening-application-window-using-tray.test.ts
jansav: We're using auto import + eslint which makes two separate lines. So if we want to switch to using suggested, let's update eslint to do the work for us. :)

src/behaviours/quitting-and-restarting-the-app/opening-application-window-using-tray.test.ts
Iku-turso: Added, and then some.

src/main/start-main-application/lens-window/application-window/create-lens-window.injectable.ts
Iku-turso: Right you are, renamed. Also instead of "opening" there's now a distinction between "starting" and "showing" a window. Therefore this naming is `isStarting`.

src/common/utils/channel/channel.test.ts
```diff
@@ -94,12 +94,12 @@ describe("channel", () => {
       });
     });
 
-    it("given multiple shown windows, when sending message, triggers listeners in all windows", async () => {
+    it("given multiple started started windows, when sending message, triggers listeners in all windows", async () => {
```
Nokel81: nit: duplicate wording

src/main/start-main-application/lens-window/application-window/wait-until-bundled-extensions-are-loaded.injectable.ts
```diff
@@ -0,0 +1,29 @@
...
+  instantiate: (di) => {
+    const ipcMain = di.inject(ipcMainInjectable);
+
+    return async () => {
+      const viewHasLoaded = new Promise<void>((resolve) => {
+        ipcMain.once(bundledExtensionsLoaded, () => resolve());
+      });
+
+      await viewHasLoaded;
+      await delay(50); // wait just a bit longer to let the first round of rendering happen
```
Nokel81, line 22: Sometime in the future we should add a `requestAnimationFrame` wrapper into the renderer so that this delay is not needed.

src/main/electron-app/runnables/dock-visibility/hide-dock-for-last-closed-window.injectable.ts
```diff
@@ -20,7 +20,7 @@ const hideDockForLastClosedWindowInjectable = getInjectable({
       run: () => {
         const visibleWindows = pipeline(
           getLensWindows(),
-          filter(window => !!window.visible),
+          filter(window => !!window.isVisible),
         );
```
Nokel81: This code has been duplicated in several places, it probably should be its own injectable.

Nokel81 (commented):
Looks great, just some minor things.

Nokel81 (dismissed):

src/main/start-main-application/lens-window/application-window/wait-until-bundled-extensions-are-loaded.injectable.ts
jansav: Absolutely. Here we just extracted this to separate injectable so that we can override it from unit tests.

src/main/electron-app/runnables/dock-visibility/hide-dock-for-last-closed-window.injectable.ts
jansav: Sure. Normally we've done it when same thing is duplicated 3-4 times. But this is clearly case in which we don't have to wait since we know already it's not going to be wrong abstraction. :) Done.

src/common/utils/channel/channel.test.ts
jansav: Good catch, fixed.

Nokel81 (dismissed):

src/main/utils/channel/message-to-channel.injectable.ts
```diff
@@ -2,32 +2,24 @@
...
+import getVisibleWindowsInjectable from "../../start-main-application/lens-window/get-visible-windows.injectable";
 
 const messageToChannelInjectable = getInjectable({
   id: "message-to-channel",
 
   instantiate: (di) => {
-    const getAllLensWindows = () => di.injectMany(lensWindowInjectionToken);
+    const getVisibleWindows = di.inject(getVisibleWindowsInjectable);
 
     // TODO: Figure out way to improve typing in internals
```
Nokel81, line 17: This can be sort of fixed by 

1. Wrap the return value in `(...)` and then add an `as MessageToChannel` cast.

Also, I think it would be possible to simply the type `MessageToChannel` to the following:

```ts
export interface MessageToChannel {
  (channel: MessageChannel<void>): void;
  <Message>(channel: MessageChannel<Message>, message: Message): void;
}
```

Nokel81 (approved):

jim-docker (approved):
It's odd that, when clicking on "Open Lens" from the tray shows the splash window and doesn't show the app window until it's "ready", but clicking on "Open Lens" from the tray a second time, while the splash window is still shown, the app window is shown immediately, and it is empty until it becomes "ready". Would be cleaner if the app window does not show until it is "ready", regardless of how many times "Open Lens" is clicked.

Nokel81:
> It's odd that, when clicking on "Open Lens" from the tray shows the splash window and doesn't show the app window until it's "ready", but clicking on "Open Lens" from the tray a second time, while the splash window is still shown, the app window is shown immediately, and it is empty until it becomes "ready". Would be cleaner if the app window does not show until it is "ready", regardless of how many times "Open Lens" is clicked.

This was a request of mine since that is how it used to work. It is a useful debugging tool for when the app window crashes so that we can then open the dev tools.

jansav:
> > It's odd that, when clicking on "Open Lens" from the tray shows the splash window and doesn't show the app window until it's "ready", but clicking on "Open Lens" from the tray a second time, while the splash window is still shown, the app window is shown immediately, and it is empty until it becomes "ready". Would be cleaner if the app window does not show until it is "ready", regardless of how many times "Open Lens" is clicked.
> 
> This was a request of mine since that is how it used to work. It is a useful debugging tool for when the app window crashes so that we can then open the dev tools.

I think in the future we should catch the errors which prevent opening of window somehow, and show some kind of error dialog for them. Then we could go back to implementation where we show application window only once it's ready. 
But since this isn't the case yet, we need to have way to access developer tools to see the real error preventing window to be started. 

Motivation is documented in the behaviour, so we should be fine and we can change this in the future if we so decide.

PR #5548, opened 2022-06-02
Remove all manual uses of v8 serialization

Nokel81:
- Seems that sometimes the is a versioning mismatch within the binary format causes cluster screen the go grey


jansav (approved):

aleksfront (approved):
LGTM

PR #5562, opened 2022-06-06
Electron v15.5.7

jakolehm:
- bumps Electron to 15.x series which is currently supported version (14.x is not)
- bumps node-pty because current stable version does not work with newer Electron versions
- changes nodejs version to 16 (because Electron 15.x has nodejs 16)

### DIFF
--- a/package.json
+++ b/package.json
@@ -313,7 +313,7 @@
     "@types/md5-file": "^4.0.2",
     "@types/mini-css-extract-plugin": "^2.4.0",
     "@types/mock-fs": "^4.13.1",
-    "@types/node": "14.18.18",
+    "@types/node": "^16.11.38",
     "@types/node-fetch": "^2.6.1",
     "@types/npm": "^2.0.32",
     "@types/proper-lockfile": "^4.1.2",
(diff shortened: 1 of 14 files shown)


src/renderer/components/test-utils/get-application-builder.tsx
```diff
@@ -209,6 +209,7 @@ export const getApplicationBuilder = () => {
           {
             menu: null as never,
             commandId: 0,
+            userAccelerator: null,
```
jakolehm, line 212: Does not pass typings without this.

package.json
```diff
@@ -355,7 +355,7 @@
     "css-loader": "^6.7.1",
     "deepdash": "^5.3.9",
     "dompurify": "^2.3.8",
```
Nokel81, line 357: Can you also bump the `@types/node` package to the one for node 16?

aleksfront (dismissed):

package.json
jakolehm: Good catch, fixed.

Nokel81 (dismissed):

Nokel81 (approved):

nevalla (approved):

PR #5567, opened 2022-06-06
Add reactive tray icon for when update is available

### LINKED ISSUE #1909: Change tray icon when update is available.

Nokel81:

fixes https://github.com/lensapp/lens/issues/1909

~Notes:~

- ~I had to manually make the update available icon (instead of generating it) so this will also have to be done in lens-ide~


src/renderer/components/test-utils/get-application-builder.tsx
```diff
@@ -166,15 +170,84 @@ export const getApplicationBuilder = () => {
     computed(() => []),
   );
 
-  let trayMenuItemsStateFake: TrayMenuItem[];
+  let commandId = 0;
+  const makeFakeMenuItem = (opts: MenuItemConstructorOptions, menu: Menu): MenuItem => {
```
jansav: Important: Let's keep our own interface here. This is now Electron's opinion how menu item looks and way too complicated. Our interface can be simpler. That also makes us less tightly coupled with Electron.

src/main/tray/reactive-tray-menu-items/tray-menu.injectable.ts
```diff
@@ -0,0 +1,28 @@
...
+import convertToElectronMenuTemplateInjectable from "./convert-to-electron-menu-template.injectable";
+
+const trayMenuInjectable = getInjectable({
+  id: "tray-menu",
+  instantiate: (di) => {
+    const trayMenuItems = di.inject(trayMenuItemsInjectable);
+    const convertToElectronMenuTemplate = di.inject(convertToElectronMenuTemplateInjectable);
+    const buildMenuFromTemplate = di.inject(buildMenuFromTemplateInjectable);
+
+    return computed(() => (
```
jansav: nit: could be easier to read. Not interesting as long as there is unit test.

```typescript
pipeline(
  trayMenuItems.get(),
  convertToElectronMenuTemplate,
  buildMenuFromTemplate
)
```

src/main/tray/reactive-tray-menu-items/convert-to-electron-menu-template.injectable.ts
```diff
@@ -0,0 +1,63 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import loggerInjectable from "../../../common/logger.injectable";
+import type { TrayMenuItem } from "../tray-menu-item/tray-menu-item-injection-token";
+
+const convertToElectronMenuTemplateInjectable = getInjectable({
```
jansav: Nit: Hmm. This could be just a function (not injectable at all). Basically it's just a mapper function. Logger is dependency which can be removed since the implementations should handle it separately. (and they already do, so it's just left-over from refactorings. Same goes for try-catch)

src/behaviours/application-update/installing-update-using-tray.test.ts
```diff
@@ -59,15 +59,14 @@ describe("installing update using tray", () => {
     });
```
jansav, line 63: Important: There should be unit tests for the changing icon here.

src/main/electron/build-from-template.injectable.ts
```diff
@@ -0,0 +1,17 @@
+/**
```
jansav: Nit: This could be implementation detail (not injectable at all). Done right, we should override the use place of this anyway.

src/renderer/components/test-utils/get-application-builder.tsx
Nokel81: I think I understand.

src/main/tray/reactive-tray-menu-items/convert-to-electron-menu-template.injectable.ts
ixrock: TBH most of the code with `DI` hard to read and not always clear _why something needed somewhere_. 
That creates huge issue in understanding how code works in general and as well as some part of it (especially now with hundreds on injectables).
I guess the goal is: _"there is injectable for everything"_ but that sounds scary and obscure :)

aleksfront: I would partially agree that code is harder to read now, I guess this is one of drawbacks from injectables. Another problem I can mention is no documentation about DI tools we use which slows the pace a lot and leads to write code without full understanding. Plus code navigation is hard.

On the other hand, we got better test deploying + separation of concerns which is good.

jansav: > TBH most of the code with DI hard to read and not always clear why something needed somewhere.
That creates huge issue in understanding how code works in general and as well as some part of it (especially now with hundreds on injectables).

I can see the pain you have. However, it's not because of `DI`. It's because of bad naming, functions doing too many things etc. However, the aid is closer than you think. Goal is to make it so that if you don't understand the code, you can just comment it and run unit tests which will fail and explain why the code is needed. But, needs to be said: if code isn't readable or understandable there is something else wrong with it. :) 

> I guess the goal is: "there is injectable for everything" but that sounds scary and obscure :)

No, `injectable` is not the goal. Goal is to keep code as simple as possible. That means code for implementation and code for tests. 

By keep the interfaces minimal and honest, you can just wish from the universe, and it will answer. However, this should be the case even if there is no `DI`.

> Another problem I can mention is no documentation about DI tools we use which slows the pace a lot and leads to write code without full understanding.

Library used is 100 % unit tested, so there is a lot documentation. Take a look [here](https://github.com/ogre-works/ogre-tools/tree/master/packages/injectable/src/dependency-injection-container). Library isn't the important part though. Important part is the thing that we are solving by using the library. 

> Plus code navigation is hard

Sometimes, yes. There is lot to improve here. E.g. in the future we should start combining all the code for a feature in one place. Everything related to `application-update` is located under `src/behaviours/application-update` and if I want to remove the feature, I can just remove the directory. Also we need to remember that there is many ways that you can navigate in codebase, I'd say we are utilizing file (and text)-search  a lot, which makes naming even more important.


If it's OK for you guys, let's have these discussions elsewhere? :) 

--

But to stay on topic here, @Nokel81 my point is that you can simplify (remove injectable) since it's not needed. It's just a technical detail which can be unit tested separately.

src/common/utils/sync-box/sync-box-injection-token.ts
```diff
@@ -4,12 +4,21 @@
...
-export interface SyncBox<TValue extends JsonValue> {
+type AsJson<T> = T extends string | number | boolean | null
+  ? T
+  : T extends Function
+    ? never
+    : T extends Array<infer V>
+      ? AsJson<V>[]
+      : T extends object
+        ? { [K in keyof T]: AsJson<T[K]> }
+        : never;
```
ixrock, line 16: What it does and why we need such complex type definition?

build/generate-tray-icons.ts
```diff
@@ -3,48 +3,128 @@
...
+              height: 60,
+            })
+            .toBuffer()
+        ),
+        top: 66,
+        left: 66,
+      },
+    ])
+    .toBuffer();
+}
```
ixrock, line 93: Looks like over engineered.. why not just combine `logo.svg` with `notice.svg` in graphics editor and inverse colors while generating PNGs?

Nokel81: Two main reasons

1. So that we can have the code to also generate the same on the lens-ide side (the logo was changed for OpenLens to be round whereas the one for Lens Desktop is squared off)
2. This is much easier to reproduce then manually doing it in a graphics editor especially because there are so many different sizes for the icons.

src/common/utils/sync-box/sync-box-injection-token.ts
Nokel81: The previous definition used `TValue extends JsonValue` which doesn't work because that requires adding an index field to all the objects that we want to serialize. 

I can add some comments as to why this is necessary of you would like.

build/generate-tray-icons.ts
ixrock: 1. How this code can do the different things without modifications? If not -> you still need to edit something and most easy way to update image it's just replace it
2. I'm not agains generating different sizes programatically :) only against composition from different files _(you're using positioning, etc)._

but maybe this is fine.. @aleksfront wdyt?

src/common/utils/sync-box/sync-box-injection-token.ts
jansav: Hm. I'm thinking if we really need anything else than JSON. It would make things a lot simpler. 
However, if you think it's needed, sure. We can simplify later if we so decide. :)

build/generate-tray-icons.ts
jansav: I do agree that the simpler solution would be that we have set of icons already generated in the repositories. We could make it so that icons could be configured in `package.json` so that we can have different icons in lens-ide. In that way we are not tied to having same file names or even directory structure in lens-ide and it would make it easier to understand what happens. (package.json -files are merged and then used as a configuration)

```json

{
  "icons": {
    "tray": {
      "normal": "./icons/tray/normal-icon.svg",
      "normal-2x": "./icons/tray/normal-icon@2x.svg",
      "update-available": "./icons/tray/update-available-icon.svg"
   }
  }
}

```

However, having automation that creates the icons is good. Manually editing icons leads to place where changing the icon is hard and can be done only by someone who has correct tools for the job.

We could have best of both worlds; script that generates icons (NPM-script, which is only triggered manually) but still have all generated icons in the repositories. 

---
This is bit against Open Closed Principle and we could make it comply with it, the simplicity is the winner.

Keep in mind that what @Nokel81 suggests here is already an improvement, so I don't see this as critical to fix right now.

jansav: But, needs to be said that this is already an improvement. So I don't see this as critical.

ixrock (approved):

src/main/tray/reactive-tray-menu-items/convert-to-electron-menu-template.injectable.ts
Iku-turso: This is good conversation. Remember that not all the benefits of complying to **Dependency Inversion Principle** or the **Dependency Inversion Container** pattern are apparent right away. For one, after a coder gains familiarity, e.g. navigation actually gets easier since dependencies are more likely to do one thing well, and describe that in its name with credibility. This should mean less need to visit the code. Also from that, we will have stuff like composability, clear abstractions, decoupling and cohesion. Cool stuff right guys? :)

jansav (approved):

PR #5573, opened 2022-06-07
Revert breaking the types of DrawerTitle

Nokel81:

This was a breakage from #4131

### DIFF
--- a/src/renderer/components/drawer/drawer-title.tsx
+++ b/src/renderer/components/drawer/drawer-title.tsx
@@ -9,7 +9,12 @@ import { cssNames } from "../../utils";
 
 export interface DrawerTitleProps {
   className?: string;
-  children: React.ReactNode;
+  children?: React.ReactNode;
+
+  /**
+   * @deprecated Prefer passing the value as `children`
+   */
+  title?: React.ReactNode;
 
   /**
    * Specifies how large this title is


src/renderer/components/drawer/drawer-title.tsx
```diff
@@ -9,7 +9,12 @@ import { cssNames } from "../../utils";
 
 export interface DrawerTitleProps {
   className?: string;
-  children: React.ReactNode;
+  children?: React.ReactNode;
+
+  /**
+   * @deprecated Prefer passing the value as `children`
+   */
+  title?: React.ReactNode;
```
ixrock, line 17: Why it was deprecated and `props.children` preferred?

ixrock (approved):

src/renderer/components/drawer/drawer-title.tsx
Nokel81: Mostly because the react docs imply that a container like `DrawerTitle` should prefer the children props:

> Some components don’t know their children ahead of time. This is especially common for components like Sidebar or Dialog that represent generic “boxes”.
> 
> We recommend that such components use the special children prop to pass children elements directly into their output:

stevejr (approved):

PR #5581, opened 2022-06-08
Enable csp on lens proxy

### LINKED ISSUE #5576: Use Content-Security-Policy header
LensProxy should set CSP header (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) that allows only `script-src 'unsafe-eval' 'self'; img-src *` by default. Distro builders should be able to provide custom CSP via package.json.

jakolehm:
Fixes #5576

### DIFF
--- a/src/common/vars.ts
+++ b/src/common/vars.ts
@@ -143,3 +143,4 @@ export const appSemVer = new SemVer(packageInfo.version);
 export const docsUrl = "https://docs.k8slens.dev/main/" as string;
 
 export const sentryDsn = packageInfo.config?.sentryDsn ?? "";
+export const contentSecurityPolicy = packageInfo.config?.contentSecurityPolicy || "script-src 'unsafe-eval' 'self'; frame-src http://*.localhost:*/; img-src *";
--- a/src/common/vars.ts
+++ b/src/common/vars.ts
@@ -143,4 +143,4 @@ export const appSemVer = new SemVer(packageInfo.version);
 export const docsUrl = "https://docs.k8slens.dev/main/" as string;
 
 export const sentryDsn = packageInfo.config?.sentryDsn ?? "";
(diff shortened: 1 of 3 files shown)


src/main/lens-proxy/lens-proxy.ts
```diff
@@ -239,6 +239,10 @@ export class LensProxy {
       }
     }
 
+    if (contentSecurityPolicy) {
+      res.setHeader("Content-Security-Policy", contentSecurityPolicy);
+    }
```
Nokel81, line 244: Won't that always be defined?

jakolehm: In theory it can be an empty string.

jakolehm: Nevermind, it can't be.

src/common/vars.ts
```diff
@@ -143,3 +143,4 @@ export const appSemVer = new SemVer(packageInfo.version);
 export const docsUrl = "https://docs.k8slens.dev/main/" as string;
 
 export const sentryDsn = packageInfo.config?.sentryDsn ?? "";
+export const contentSecurityPolicy = packageInfo.config?.contentSecurityPolicy || "script-src 'unsafe-eval' 'self'; frame-src http://*.localhost:*/; img-src *";
```
ixrock:
```suggestion
export const contentSecurityPolicy = packageInfo.config.contentSecurityPolicy;
```

Why not to keep defaults directly in `package.json`?

ixrock (dismissed):

src/common/vars.ts
jakolehm: @ixrock good point, will move defaults to package.json.

jakolehm:
@Nokel81 @ixrock ptal again.

ixrock (approved):
LGTM

Nokel81 (approved):

PR #5597, opened 2022-06-09
Fix memory leak in unit tests

Nokel81:
- Upgrade to jest 28
- Switch to using swc
- Use @side/jest-runtime as the jest-runtime


jim-docker:
Are all the changes to tests related to the memory issue/swc addition?

Nokel81:
Yes

Nokel81:
And the jest upgrade

jakolehm (commented):
I would split swc out from this PR because it will affect runtime and might introduce unexpected bugs. Or is there a reason to do both of these changes in the same PR?

jansav (commented):
I'm OK with this. All meaningful changes were made to test code. SWC and runtime if used only in unit tests so we're not breaking anything in the application itself. 

Of course in the future we need to make sure that code is compiled with exactly same tool and configuration for unit tests and production.

There is some typescript errors which make unit tests to fail. I'll approve when those have been fixed.

jakolehm:
> All meaningful changes were made to test code. SWC and runtime if used only in unit tests so we're not breaking anything in the application itself.

Oh, didn't realize that we use swc only for tests. Then it's probably fine (although feels a bit bad to use different "compiler" for unit tests).

src/extensions/__tests__/extension-loader.test.ts
```diff
@@ -125,42 +126,39 @@ describe("ExtensionLoader", () => {
...
-            "absolutePath": "/test/1",
-            "id": "manifest/path",
-            "isBundled": false,
-            "isEnabled": true,
-            "manifest": Object {
-              "name": "TestExtension",
-              "version": "1.0.0",
-            },
-            "manifestPath": "manifest/path",
+    await delay(10);
```
jansav, line 133: Note: This `delay` (and the old `setTimeout`) is very bad thing. But since you are not adding it, and just changing the way of doing it, it's fine for now.

src/extensions/extension-discovery/extension-discovery.test.ts
```diff
@@ -114,13 +117,14 @@ describe("ExtensionDiscovery", () => {
         },
         manifestPath: path.normalize("some-directory-for-user-data/node_modules/my-extension/package.json"),
       });
-      done();
+      letTestFinish.set(true);
```
jansav, line 120: Do we really need this? If the motivation is that you are not allowed to have `async` and `done` in same test, could we just remove `async` since it wasn't in use at all? :)

src/main/protocol-handler/__test__/router.test.ts
```diff
@@ -33,49 +32,36 @@ function throwIfDefined(val: any): void {
...
-    mockFs.restore();
-  });
-
   it("should throw on non-lens URLS", async () => {
-    try {
-      expect(await lpr.route("https://google.ca")).toBeUndefined();
-    } catch (error) {
-      expect(error).toBeInstanceOf(Error);
-    }
+    await lpr.route("https://google.ca");
```
jansav, line 58: Are we changing the behaviour for a reason? If so, name of test lies. If not, the test is wrong.

src/renderer/components/layout/top-bar/start-state-sync.injectable.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { action } from "mobx";
+import { beforeFrameStartsInjectionToken } from "../../../before-frame-starts/before-frame-starts-injection-token";
+import ipcRendererInjectable from "../../../utils/channel/ipc-renderer.injectable";
+import topBarStateInjectable from "./state.injectable";
+
+// TODO: replace with a SyncBox
+const startTopbarStateSyncInjectable = getInjectable({
+  id: "start-topbar-state-sync",
```
jansav, line 13: `causesSideEffects: true`

jansav:
@Nokel81 ping me when you are available and let's get this merged. My comments are small stuff and some of them just need little bit explanation. :) We can have huddle.

jansav (dismissed):

jakolehm (dismissed):

jansav (dismissed):

jansav (approved):

PR #5603, opened 2022-06-13
Fix issue where releases are not reloaded when new release is added e…

### LINKED ISSUE #5269: Helm Releases sometimes do not refresh
~Helm repos are not auto updated and have to close lens from icon tray and then repoen.~

Helm releases are not auto updated and have to disconnect from cluster and come back to see them.

The latter is more important IMO.

Please either auto update or add a refresh button.

**Why is this needed**:
Slows down workflow.

- Kubernetes distribution: [ AKS,.]
- Desktop OS: <win10 with kubeconfig in ubuntu 20.04 WSL2 on >

jansav:
…xternally (e.g. from terminal)

Notice:

There should be behaviour for this, but adding it would lead to cascade of refactorings so I chose not do it for now. 

How to test:

1. Open Releases
2. Open Terminal
3. `$ helm install --generate-name bitnami/consul`
4. Make sure that releases are reloaded (this is the part that wasn't working before)
5. `$ helm list` and `$ helm uninstall X`
6. Make sure that releases are reloaded (Notice how the release which is uninstalled will stay in the list with status "Uninstalling". If this is not intended behaviour, then it's separate bug somewhere in `SecretStore`)

### DIFF
--- a/src/renderer/components/+helm-releases/release-secrets.injectable.ts
+++ b/src/renderer/components/+helm-releases/release-secrets.injectable.ts
@@ -1,0 +1,37 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { computed, onBecomeObserved, onBecomeUnobserved } from "mobx";
+import subscribeStoresInjectable from "../../kube-watch-api/subscribe-stores.injectable";
+import secretStoreInjectable from "../+config-secrets/store.injectable";
+
+const releaseSecretsInjectable = getInjectable({
+  id: "release-secrets",
+
+  instantiate: (di) => {
+    const subscribeStores = di.inject(subscribeStoresInjectable);
+    const secretStore = di.inject(secretStoreInjectable);
+
+    const releaseSecrets = computed(() =>
+      secretStore.contextItems.filter((secret) =>
+        secret.type.startsWith("helm.sh/release"),
+      ),
+    );
+
+    let unsubscribe: () => void;
+
+    onBecomeObserved(releaseSecrets, () => {
+      unsubscribe = subscribeStores([secretStore]);
+    });
+
+    onBecomeUnobserved(releaseSecrets, () => {
+      unsubscribe?.();
+    });
+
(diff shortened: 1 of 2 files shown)


Nokel81:
fixes https://github.com/lensapp/lens/issues/5269

src/renderer/components/+helm-releases/release-secrets.injectable.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    const releaseSecrets = computed(() =>
+      secretStore.contextItems.filter((secret) =>
+        secret.type.startsWith("helm.sh/release"),
+      ),
+    );
+
+    let unsubscribe: () => void;
+
+    onBecomeObserved(releaseSecrets, () => {
```
Nokel81, line 25: This is neat

src/renderer/components/+helm-releases/releases.injectable.ts
```diff
@@ -7,17 +7,21 @@ import { asyncComputed } from "@ogre-tools/injectable-react";
...
 
   instantiate: (di) => {
     const clusterContext = di.inject(clusterFrameContextInjectable);
     const namespaceStore = di.inject(namespaceStoreInjectable);
+    const releaseSecrets = di.inject(releaseSecretsInjectable);
 
     return asyncComputed(async () => {
       const contextNamespaces = namespaceStore.contextNamespaces || [];
 
+      void releaseSecrets.get();
```
Nokel81, line 23: Why is this enough?

Nokel81 (approved):

src/renderer/components/+helm-releases/release-secrets.injectable.ts
jansav: Need to remember that this is very dangerous tool. It's easy to create stuff which goes to endless observe-unobserve loop. :)

src/renderer/components/+helm-releases/releases.injectable.ts
jansav: `asyncComputed` works as `computed` does. So if you observe some observable, and the value changes, it will re-compute. Here `releaseSecrets` changes when something is installed, so that causes `asyncComputed` to recompute.

Iku-turso (approved):

Issue #3955, opened 2021-10-05
Add support for specifying prometheus labels


Nokel81:
**What would you like to be added**:
The ability to specify some labels as part of the prometheus settings in Lens.

**Why is this needed**:
As a follow up to https://github.com/lensapp/lens/pull/3653, it is now more obvious that there is not standard prometheus labelling protocol for kube objects. Most commonly the use of `node` vs `kubernetes_node`. It would be better if the user could specify this within Lens.

ruckc:
I'd reallt like this.  I deploy prometheus through kube-prometheus-stack helm chart, and one service has the `operated-prometheus=true` label and difference services get the `self-monitor=true` label.

ineentho:
Version 15 of the [prometheus-community/prometheus](https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus) chart updated the default metric labels, removing the kubernetes_ prefix. This aligns it with prometheus-operator, but breaks Lens cluster-wide stats by default.

See this merge request: https://github.com/prometheus-community/helm-charts/pull/1520

Example of metric before 15:

```
node_cpu_seconds_total{app="prometheus", app_kubernetes_io_instance="prometheus", chart="prometheus-14.11.1", component="node-exporter", cpu="0", heritage="Helm", instance="10.0.0.102:9100", job="kubernetes-service-endpoints", kubernetes_name="prometheus-node-exporter", kubernetes_namespace="monitoring", kubernetes_node="main", mode="system", release="prometheus"}
```

after:
```
node_cpu_seconds_total{app="prometheus", app_kubernetes_io_instance="prometheus", chart="prometheus-15.0.2", component="node-exporter", cpu="0", heritage="Helm", instance="10.0.0.102:9100", job="kubernetes-service-endpoints", mode="system", namespace="monitoring", node="main", release="prometheus", service="prometheus-node-exporter"}
```

I'm not sure what the correct course of action is, but any way to solve this situation would be useful.

Nokel81:
@ineentho For the time being does switching to `prometheus=operator` work for you?

tyriis:
@ineentho I can confirm, switching to `prometheus=operator` is solving the issue for me with the community charts.

sidh:
`prometheus=operator` doesn't helm in my case :(

nabokihms:
I think it can also be helpful to provide the ability to specify custom labels and provide Prometheus queries for all metrics. Let's call it the `Custom Prometheus' option.

[...]

PR #5604, opened 2022-06-13
Fix broken lens metrics.

iciclespider:
1. There are no `kubernetes_name` labels, this really should be `kubernetes_node`
2. The `nodes` `cpuCapaicity` metric was incorrectly coded to use`kube_node_status_allocatable` , this should use `kube_node_status_capacity` like all the other cpu and memory capacity metrics do.


Nokel81:
This is sort of a duplicate of #3955

Nokel81:
I am going to close this as a duplicate of #6157 and https://github.com/lensapp/lens/pull/6139 which allow both v14 and v15 to be used. Thanks for the PR none the less.

PR #5607, opened 2022-06-13
Improved metrics for updating application

### LINKED ISSUE #5064: Improve metrics gathering for update checks
More granular metrics tracking for updates. For instance:

- timestamps of "checking for updates"
- how the update was applied
- timestamps of successful updates
- currently running version
- ...

jansav:


src/behaviours/application-update/analytics-for-installing-update.test.ts
```diff
@@ -0,0 +1,158 @@
...
+
+      processCheckingForUpdates();
+    });
+
+    it("sends event to analytics about checking for updates", () => {
+      expect(analyticsListenerMock.mock.calls).toEqual([
+        [
+          {
+            name: "app",
+            action: "checking-for-updates",
```
Nokel81, line 150: Given that this can be done both automatically (I think every 2 hours) and manually, I think that how it was triggered should be part of the metric.

src/common/utils/date/get-current-date-time.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import moment from "moment";
+
+export const getCurrentDateTime = () => moment().utc().format();
```
Nokel81, line 7: Shouldn't this be an `injectable`, so that you don't to do `global.Data.now = ...` in the tests.

src/main/application-update/quit-and-install-update.injectable.ts
```diff
@@ -0,0 +1,45 @@
...
+      emitEvent({
+        name: "app",
+        action: "start-installing-update",
+
+        params: {
+          version: discoveredVersion.version,
+          updateChannel: discoveredVersion.updateChannel.id,
+          currentDateTime: getCurrentDateTime(),
+        },
+      });
```
Nokel81, line 38: What criteria should we use for when these should be split out into their own injectable functions instead of being inline? I see that there is `emitCurrentVersionToAnalyticsInjectable` but not for this as well.

src/common/utils/date/get-current-date-time.ts
jansav: I'd like to say yes, but previous experience says that it's best to override `Date` by messing with globals, because of the all of the libraries that use it. If we used `injectable` for it, then we would have absolute control in our code base, but if (and when) we use some library that uses `Date.now`, we will get separate opinion which is easy way encounter hard-to-debug failing unit tests.

src/main/application-update/quit-and-install-update.injectable.ts
jansav: Common sense. Things that I consider (not in any particular order):

1. Am I able to keep usage cleaner by giving name for it? (-> function in same file)
2. Does it have it's own dependencies? (-> injectable)
3. Is it used in multiple places? (-> injectable if there is dependencies, otherwise function)
4. Is it something that need to be overridden? (-> injectable)
5. Does it cause side effects? (-> injectable)

So, I would definitely not create injectable just for sake of injectable. There needs to be reason for it. 

`emitCurrentVersionToAnalyticsInjectable` is because it's `runnable`, which is ran when application is loaded.

src/behaviours/application-update/analytics-for-installing-update.test.ts
jansav: Great. Added.

src/main/application-update/quit-and-install-update.injectable.ts
Nokel81: Okay those are roughly the criteria I would have followed too, though given that this might make usage clearer with its own name and it has some dependencies of its own I would have moved it to an injectable.

Minor though.

src/common/utils/date/get-current-date-time.ts
Nokel81: Fair enough, thought after talking with @nevalla it seems that the timestamp is received automatically, so we don't need to include it.

jim-docker (approved):

src/common/utils/date/get-current-date-time.ts
jansav: I'd say let's keep it there. Sometimes it's easier to access data from event itself, and not from the metadata. It all depends from the choice of service that we are using.

Nokel81 (approved):

PR #5609, opened 2022-06-13
Add behaviour for tray menu items originating from extensions

jansav:


src/behaviours/preferences/navigation-to-extension-specific-preferences.test.tsx
```diff
@@ -48,7 +48,7 @@ describe("preferences - navigation to extension specific preferences", () => {
         const getRendererExtensionFake = getRendererExtensionFakeFor(applicationBuilder);
         const testExtension = getRendererExtensionFake(extensionStubWithExtensionSpecificPreferenceItems);
 
-        applicationBuilder.addExtensions(testExtension);
+        applicationBuilder.extensions.renderer.enable(testExtension);
```
Nokel81: why is this one not awaited and the rest are?

src/behaviours/tray/clicking-tray-menu-item-originating-from-extension.test.ts
```diff
@@ -0,0 +1,134 @@
...
+
+      it("logs the error", () => {
+        expect(logErrorMock).toHaveBeenCalledWith(
+          '[TRAY]: Clicking of tray item "some-label" from extension "some-extension-id" failed.',
+          "some-rejection",
+        );
+      });
+    });
+
+    describe("when disabling extension", () => {
```
Nokel81: I would call this "when extension has subsequently been disabled".

src/behaviours/tray/clicking-tray-menu-item-originating-from-extension.test.ts
```diff
@@ -0,0 +1,134 @@
...
+
+        expect(
+          applicationBuilder.tray.get(
+            "some-label-tray-menu-item-for-extension-some-extension-id-instance-1",
+          ),
+        ).toBeNull();
+      });
+
+      // Note: Motivation here is to make sure that enabling same extension does not throw
+      it("when extension is re-enabled, has the tray menu item from extension", async () => {
```
Nokel81, line 103: And then shouldn't this also be in another `describe`

src/renderer/components/test-utils/get-application-builder.tsx
```diff
@@ -169,26 +188,23 @@ export const getApplicationBuilder = () => {
...
   let trayMenuIconPath: string;
 
   mainDi.override(electronTrayInjectable, () => ({
     start: () => {
       trayMenuIconPath = iconPaths.normal;
     },
     stop: () => {},
-    setMenuItems: (items) => {
-      trayMenuItemsStateFake = items;
-    },
```
Nokel81, line 196: why remove this? Don't we want to test non-extension menu items?

src/behaviours/preferences/navigation-to-extension-specific-preferences.test.tsx
jansav: Hm, good catch. It's because in upcoming PR there is change to remove async from enabling extension for good. However, in here `await` is correct.

src/behaviours/tray/clicking-tray-menu-item-originating-from-extension.test.ts
jansav: Went with "when extension is disabled"

src/behaviours/preferences/navigation-to-extension-specific-preferences.test.tsx
jansav: Fixed

src/behaviours/tray/clicking-tray-menu-item-originating-from-extension.test.ts
jansav: No need if there is only one test. We could add second test just for documentative purpose. But it would be empty test, so comment does same trick.

src/renderer/components/test-utils/get-application-builder.tsx
jansav: Of course we do. They can be tested just like any item originating from extension. Having this fake here, skipped some business logic we had for tray menu items. (e.g. filtering based on visibility) which is very important to be part of unit test. Changes to `applicationBuilder.tray.click` and `applicationBuilder.tray.get` made this redundant.

Nokel81: I still don't understand, wouldn't it be a better to test the runables are set up correctly as well? And thus this would still be needed?

Nokel81 (dismissed):

src/renderer/components/test-utils/get-application-builder.tsx
Nokel81: As discussed in the huddle, this is not currently possible because of some other things. We will get to that point.

src/behaviours/tray/clicking-tray-menu-item-originating-from-extension.test.ts
```diff
@@ -0,0 +1,134 @@
...
+        id: "some-extension-id",
+        trayMenus: [{ label: "some-label", click: clickMock }],
+      });
+
+      await applicationBuilder.extensions.main.enable(someExtension);
+    });
+
+    it("when item is clicked, triggers the click handler", () => {
+      applicationBuilder.tray.click(
+        "some-label-tray-menu-item-for-extension-some-extension-id-instance-1",
```
jim-docker, line 45: where does this identifier come from?

src/renderer/components/test-utils/get-application-builder.tsx
```diff
@@ -244,18 +297,20 @@ export const getApplicationBuilder = () => {
...
       getIconPath: () => trayMenuIconPath,
+
       click: async (id: string) => {
-        const menuItem = pipeline(
-          trayMenuItemsStateFake,
-          find((menuItem) => menuItem.id === id),
-        );
+        const trayMenuItems = mainDi.inject(trayMenuItemsInjectable).get();
+
+        const menuItem = trayMenuItems.find(matches({ id })) ?? null;
```
jim-docker, line 307: would that you could just call `get()`?

jim-docker (dismissed):

src/behaviours/tray/clicking-tray-menu-item-originating-from-extension.test.ts
jansav: From above:

```
someExtension = new SomeTestExtension({
  id: "some-extension-id",
  trayMenus: [{ label: "some-label", click: clickMock }],
});
```

And rest from implementation.

jansav:
Rebased. Mind re-approving @Nokel81 @jim-docker. Feel free to merge when there is 2 approves and everything passes

Iku-turso (approved):

Iku-turso (approved):

Nokel81 (approved):

PR #5611, opened 2022-06-13
Custom columns for kubernetes cluster catalog category

jakolehm:
image


src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -132,6 +132,14 @@ export class KubernetesCluster<
         break;
     }
   }
+
+  get k8sVersion() {
+    return this.metadata.kubeVersion?.replace("unknown", "") || "";
```
Nokel81: Do you actually want to replace the first occurrence of `unknown` or do you want to just want to return `""` instead of `"unknown"`?

src/common/catalog/categories/kubernetes-cluster.injectable.ts
```diff
@@ -4,12 +4,10 @@
  */
 import { getInjectable } from "@ogre-tools/injectable";
 import { KubernetesClusterCategory } from "../../catalog-entities/kubernetes-cluster";
-import { builtInCategoryInjectionToken } from "../category-registry.injectable";
 
 const kubernetesClusterCategoryInjectable = getInjectable({
   id: "kubernetes-cluster-category",
   instantiate: () => new KubernetesClusterCategory(),
-  injectionToken: builtInCategoryInjectionToken,
 });
```
Nokel81, line 11: I don't think that this should be left in if this injection token is being removed. Because we expose this to the extension API an extension might try and listen for the events on this instance, but if it is not registered in the category registry then no events will be emitted on this instance.

src/renderer/catalog-entities/kubernetes-cluster.tsx
```diff
@@ -0,0 +1,63 @@
...
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import React from "react";
+import { categoryVersion } from "../../common/catalog";
+import type { CatalogCategorySpec, CatalogEntityConstructor } from "../../common/catalog";
+import { KubernetesCluster, KubernetesClusterCategory as KubernetesClusterCommonCategory } from "../../common/catalog-entities";
+import styles from "./kubernetes-cluster.module.scss";
+
+export class KubernetesClusterCategory extends KubernetesClusterCommonCategory {
```
Nokel81, line 12: The other option could be to have `displayColumns` be an optional parameter to the category's constructor. That way we don't need a new base class.

Another option would be to expand the `categoryColumnsInjectable` with an injection token for internal registrations.

Nokel81 (changes requested):

src/common/catalog/categories/kubernetes-cluster.injectable.ts
jakolehm: @Nokel81 what's your suggestion? We cannot expose main or renderer because the extension api is common.

Nokel81: Well this will cause bugs so we cannot leave it like this.

Either we: 

1. Have only `kubernetesClusterCategoryInjectable` which then does `di.inject(kubernetesClusterCategoryColumnsInjectable)`
2. Have the extension API expose a `legacyGlobalForExtensionApi` based on a injection token and use a jump pad for the auto registration.

jakolehm: @Nokel81 Yes I understand that we need to fix this. Still I started to wonder why are we exposing instance of `KubernetesClusterCategory`, shouldn't all instances be available through `CatalogCategoryRegistry`.... feels like a bug to me.

Nokel81:
Closing this as I will open a new PR with this feature shortly.

PR #5626, opened 2022-06-14
Fix failing to start shell being hidden from user

### LINKED ISSUE #5513: Unable to connect to Pods' shell using lens
click shell Always present  Connecting ...  But it doesn't connect

Nokel81:

fixes https://github.com/lensapp/lens/issues/5513


Nokel81:
This is now blocked on https://github.com/xtermjs/xterm.js/pull/4214 being merged and released because without it the tests fail

Nokel81:
This is now blocked on https://github.com/xtermjs/xterm.js/pull/4217, since the first PR that I created was rejected for not being [Trusted Types](https://www.w3.org/TR/2022/WD-trusted-types-20220927/) compliant

src/features/terminal/opening-terminal-tab.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+      expect(convertCanvasToPngBuffer(textCanvas)).toMatchImageSnapshot();
+    });
+
+    it.skip("connects websocket to main", () => {
+
+    });
+
+    it.skip("displays the values on screen", () => {
+
+    });
```
jansav, line 140: Skipped empty tests?

src/main/start-main-application/runnables/setup-lens-proxy.injectable.ts
```diff
@@ -82,7 +82,7 @@ const setupLensProxyInjectable = getInjectable({
     };
   },
 
-  causesSideEffects: true,
+  // causesSideEffects: true,
```
jansav: Remove line

src/renderer/components/dock/terminal/store.ts
```diff
@@ -20,69 +20,71 @@ interface Dependencies {
...
     });
     const terminal = this.dependencies.createTerminal(tab.id, api);
 
-    this.connections.set(tab.id, api);
-    this.terminals.set(tab.id, terminal);
+    runInAction(() => {
+      this.connections.set(tab.id, { api, terminal });
+    });
+
+    const connectionPromise = api.connect();
```
jansav, line 56: Why are we not `await`ing for promise?

src/common/utils/channel/channel.test.ts
```diff
@@ -11,15 +11,15 @@ import { getApplicationBuilder } from "../../../renderer/components/test-utils/g
...
-import type { RequestChannel } from "./request-channel-listener-injection-token";
+import type { RequestChannel } from "./request-channel";
 import type { AsyncFnMock } from "@async-fn/jest";
 import asyncFn from "@async-fn/jest";
 import { getPromiseStatus } from "../../test-utils/get-promise-status";
 import { runInAction } from "mobx";
 import type { RequestChannelHandler } from "../../../main/utils/channel/channel-listeners/listener-tokens";
 import { getRequestChannelListenerInjectable } from "../../../main/utils/channel/channel-listeners/listener-tokens";
+import type { RequestFromChannel } from "../../../renderer/utils/channel/request-from-channel.injectable";
+import requestFromChannelInjectable from "../../../renderer/utils/channel/request-from-channel.injectable";
```
jansav, line 22: We shouldn't be removing the injection token. `requestFromChannel` is contract which is implemented separately in renderer and main but the **contract** is same.

jansav: Important thing is that channel is something that can be called from anywhere and it doesn't know who answers. That being said, the "request" or "message" doesn't even have to go over IPC, the listener and caller can be located in same environment. (It doesn't work like this yet due YAGNI but conceptually that's what **Channel** is.)

src/renderer/utils/sync-box/provide-initial-values-for-sync-boxes.injectable.ts
```diff
@@ -6,26 +6,26 @@ import { getInjectable } from "@ogre-tools/injectable";
...
+import requestFromChannelInjectable from "../channel/request-from-channel.injectable";
 
 const provideInitialValuesForSyncBoxesInjectable = getInjectable({
   id: "provide-initial-values-for-sync-boxes",
 
   instantiate: (di) => ({
     id: "provide-initial-values-for-sync-boxes",
     run: async () => {
-      const requestFromChannel = di.inject(requestFromChannelInjectionToken);
+      const requestFromChannel = di.inject(requestFromChannelInjectable);
```
jansav, line 20: Revert, see above discussion about the requirement for `injectionToken`.

src/common/app-paths/app-paths.test.ts
```diff
@@ -56,6 +56,10 @@ describe("app-paths", () => {
     });
   });
 
+  afterEach(() => {
+    builder.quit();
+  });
+
```
jansav: No, this implies that there is some shared global state in the builder. We need to figure out what that is and remove it. There shouldn't be any requirement for cleanups after test.

src/renderer/components/test-utils/get-application-builder.tsx
```diff
@@ -703,6 +706,13 @@ export const getApplicationBuilder = () => {
       await startApplication({ shouldStartHidden: true });
     },
 
+    quit() {
```
jansav: 1. This isn't correct way to quit the application. This should use all same code than we use when application is quit in real life. Extract the function used in production and use it here.

2. `quit` shouldn't be part of all behavioural tests. There should be explicit behavioural tests for quitting the application and rest of them shouldn't know anything about it. That being said, `afterEach` requirement is not welcome.

package.json
```diff
@@ -449,7 +449,7 @@
     "webpack-cli": "^4.9.2",
     "webpack-dev-server": "^4.11.1",
     "webpack-node-externals": "^3.0.0",
-    "xterm": "^4.19.0",
+    "xterm": "^5.0.0",
```
jansav: This commit removes the last usage for earlier introduce image-snapshot dependency. Could you remove it?

package.json
```diff
@@ -450,27 +449,15 @@
     "webpack-dev-server": "^4.11.1",
     "webpack-node-externals": "^3.0.0",
     "xterm": "^5.0.0",
-    "xterm-addon-fit": "^0.5.0"
+    "xterm-addon-fit": "^0.5.0",
```
jansav: Why is this commit part of this monster PR at all? Doesn't have anything to do with fixing something, it's adding stuff.

src/main/lens-proxy/proxy-functions/shell-api-request.injectable.ts
```diff
@@ -27,14 +27,21 @@ const shellApiRequestInjectable = getInjectable({
       const nodeName = searchParams.get("node");
       const shellToken = searchParams.get("shellToken");
 
+      console.log("got shell api request", { tabId, clusterId: cluster?.id, nodeName });
```
jansav: Switch to using logger.

package.json
```diff
@@ -86,7 +87,9 @@
     "version": "yarn run version-checkout && git add package.json && yarn run version-commit",
     "postversion": "git push --set-upstream ${GIT_REMOTE:-origin} release/v$npm_package_version",
     "precreate-release-pr": "npx swc ./scripts/create-release-pr.ts -o ./scripts/create-release-pr.mjs",
-    "create-release-pr": "node ./scripts/create-release-pr.mjs"
+    "create-release-pr": "node ./scripts/create-release-pr.mjs",
+    "prefix-canvas-deps": "npx swc ./scripts/fix-canvas-deps.ts -o ./scripts/fix-canvas-deps.mjs",
+    "fix-canvas-deps": "node ./scripts/fix-canvas-deps.mjs"
```
jansav, line 94: What is the requirement for `canvas` in the first place? This is complexity which shouldn't be included without very good reason.

src/features/cluster/kube-object-status-icon/extension-api/reactively-hide-kube-object-status.test.tsx
```diff
@@ -71,24 +72,16 @@ describe("reactively hide kube object status", () => {
...
   });
 
-  it("does not show the kube object status", () => {
-    const actual = rendered.baseElement.querySelectorAll(
-      ".KubeObjectStatusIcon",
-    );
-
-    expect(actual).toHaveLength(0);
+  it("does not show the kube object status", async () => {
+    await waitFor(() => expect(rendered.baseElement.querySelectorAll(".KubeObjectStatusIcon")).toHaveLength(0));
```
jansav, line 72: This is slow unit test. In unit test we should always know if the item is there or not.

src/features/cluster/kube-object-status-icon/extension-api/reactively-hide-kube-object-status.test.tsx
```diff
@@ -71,24 +72,16 @@ describe("reactively hide kube object status", () => {
...
     runInAction(() => {
       someObservable.set(true);
     });
 
-    const actual = rendered.baseElement.querySelectorAll(
-      ".KubeObjectStatusIcon",
-    );
-
-    expect(actual).toHaveLength(1);
+    await waitFor(() => expect(rendered.baseElement.querySelectorAll(".KubeObjectStatusIcon")).toHaveLength(1));
```
jansav, line 80: This is slow unit test. In unit test we should always know if the item is there or not.

src/features/quitting-and-restarting-the-app/opening-application-window-using-tray.test.ts
```diff
@@ -136,7 +137,7 @@ describe("opening application window using tray", () => {
           });
 
           it("starts loading of content for the application window", () => {
-            expect(callForApplicationWindowHtmlMock).toHaveBeenCalledWith("https://lens.app:42");
+            expect(callForApplicationWindowHtmlMock).toHaveBeenCalledWith(matches((val) => val.startsWith("http://lens.app:")));
```
jansav: Side-effect with the port number :(

src/renderer/api/create-terminal-api.injectable.ts
```diff
@@ -17,19 +17,25 @@ export type CreateTerminalApi = (query: TerminalApiQuery) => TerminalApi;
...
       defaultParams: di.inject(defaultWebsocketApiParamsInjectable),
       logger: di.inject(loggerInjectable),
       currentLocation: di.inject(currentLocationInjectable),
-      hostedClusterId,
     };
 
-    return (query) => new TerminalApi(deps, query);
+    return (query) => {
+      const hostedClusterId = di.inject(hostedClusterIdInjectable);
+
```
jansav, line 31: This injection should be done in the instantiate time and not in the function.

src/common/cluster/cluster.ts
```diff
@@ -20,12 +20,12 @@ import { disposer, isDefined, isRequestError, toJS } from "../utils";
 import type { Response } from "request";
```
jansav: "Update (mostly just) snapshots and a few other tests"

Please be more careful with commit names, this one does a lot more than it says. Almost skipped reviewing because it said "update snapshots"

src/features/helm-charts/listing-active-helm-repositories-in-preferences.test.ts
```diff
@@ -43,7 +43,7 @@ describe("listing active helm repositories in preferences", () => {
 
     builder.beforeApplicationStart((mainDi) => {
       mainDi.override(readYamlFileInjectable, () => readYamlFileMock);
-      mainDi.override(execFileInjectable, () => execFileMock);
+      mainDi.override(exrequestPublicHelmRepositoriesInjectable
```
jansav: Typo?

src/features/preferences/navigation-to-kubernetes-preferences.test.ts
```diff
@@ -30,7 +30,7 @@ describe("preferences - navigation to kubernetes preferences", () => {
         mainDi.override(
           getActiveHelmRepositoriesInjectable,
           () => async () => ({ callWasSuccessful: true, response: [] }),
-        );
+        );requestPublicHelmRepositoriesInjectable;
```
jansav: Typo?

src/features/window-location/main/for-top-bar-state-handler.injectable.ts
```diff
@@ -0,0 +1,28 @@
...
+    const getVisibleWindows = di.inject(getVisibleWindowsInjectable);
+    const topBarState = di.inject(topBarStateInjectable);
+
+    return () => {
+      const windows = getVisibleWindows();
+
+      topBarState.set({
+        prevEnabled: windows.some(window => window.canGoBack()),
+        nextEnabled: windows.some(window => window.canGoForward()),
+      });
```
jansav, line 23: This seems fundamentally wrong. At the moment we have only one window, but if we had more, this would result in very weird behaviour.

src/common/fetch/lens-fetch.injectable.ts
```diff
@@ -0,0 +1,33 @@
...
+import lensProxyPortInjectable from "../../main/lens-proxy/lens-proxy-port.injectable";
+import lensProxyCertificateInjectable from "../certificate/lens-proxy-certificate.injectable";
+import nodeFetchModuleInjectable from "./fetch-module.injectable";
+
+export type LensFetch = (pathnameAndOrQuery: string, init?: Omit<RequestInit, "agent">) => Promise<Response>;
+
+/**
+ * This fetch is for requesting items from the Lens Proxy and thus does not cause side effects
+ */
+const lensFetchInjectable = getInjectable({
```
jansav, line 17: How about we override this in behavioural test with a fake that does whatever the "server" does. So basically maps the request to correct route in backend.

src/features/shell-sync/main/compute-unix-shell-environment.global-override-for-injectable.ts
```diff
@@ -6,6 +6,7 @@
 import { getGlobalOverride } from "../../../common/test-utils/get-global-override";
 import computeUnixShellEnvironmentInjectable from "./compute-unix-shell-environment.injectable";
 
-export default getGlobalOverride(computeUnixShellEnvironmentInjectable, () => async () => {
-  throw new Error("Tried to get unix shell env without override");
-});
+export default getGlobalOverride(computeUnixShellEnvironmentInjectable, () => async () => ({
+  callWasSuccessful: true,
+  response: process.env,
```
jansav, line 11: This is introducing side effect to override. In global override the response should be always same.

package.json
```diff
@@ -78,7 +79,7 @@
     "build:theme-vars": "yarn run ts-node build/build_theme_vars.ts",
     "lint": "PROD=true yarn run eslint --ext js,ts,tsx --max-warnings=0 .",
     "lint:fix": "yarn run lint --fix",
-    "postinstall": "yarn run fix-canvas-deps",
+    "postinstall": "yarn run fix-canvas-deps && yarn run compile:node-fetch",
```
jansav: Both of these introduce level of complexity that makes me question why they are **really** needed.

src/common/app-paths/app-paths.test.ts
```diff
@@ -7,17 +7,17 @@ import getElectronAppPathInjectable from "../../main/app-paths/get-electron-app-
...
+import { setupInitializingApplicationBuilder } from "../../features/test-utils/application-builder";
 import type { DiContainer } from "@ogre-tools/injectable";
 import appPathsInjectable from "./app-paths.injectable";
 
 describe("app-paths", () => {
   let builder: ApplicationBuilder;
 
-  beforeEach(() => {
-    builder = getApplicationBuilder();
+  setupInitializingApplicationBuilder(b => builder = b);
```
jansav, line 18: Ugh. This would be hiding a lot of important details. It also introduces dependency between order of these before each loops. If the motivation is to remove the requirement for `afterEach` from usages, then we should figure out why we need the `afterEach` at all. It shouldn't be there.

src/common/cluster/cluster.ts
```diff
@@ -668,19 +731,26 @@ export class Cluster implements ClusterModel {
...
       return [];
     }
   }
 
   shouldShowResource(resource: KubeApiResourceDescriptor): boolean {
+    // console.log({
+    //   resource,
+    //   allowed: toJS([...this.allowedResources]),
+    // });
+
```
jansav: Debug code

src/common/utils/collection-functions.ts
```diff
@@ -23,6 +24,21 @@ export function getOrInsert<K, V>(map: Map<K, V>, key: K, value: V): V {
...
+export async function getOrInsertWithObservable<K>(box: IObservableValue<K | undefined>, getter: () => Promise<K>): Promise<K> {
+  const val = box.get();
+
+  if (val) {
+    return val;
+  }
+
+  const newVal = await getter();
+
+  runInAction(() => box.set(newVal));
```
jansav, line 36: Dangerous. Transaction should be controlled from the useplace.

src/common/fs/change-path-mode.global-override-for-injectable.ts
```diff
@@ -1,11 +0,0 @@
-/**
```
jansav: This commit doesn't explain at all why these global overrides were removed?

src/jest.setup.ts
```diff
@@ -74,6 +74,13 @@ Object.defineProperty(window, "matchMedia", {
   })),
 });
 
+/**
+ * This is needed for getting XTermJs to work in tests
+ */
```
jansav, line 79: How about we override x-term-js with something that mimics the behaviour but is in our full control?

scripts/fix-canvas-deps.ts
```diff
@@ -0,0 +1,46 @@
+/**
```
Iku-turso, line 1: This whole file feels like an amazing kludge, and brittle too. Is there really no other way?

src/features/window-location/main/for-top-bar-state-handler.injectable.ts
Nokel81: Fair enough, however it is what the current state was so I just changed it to be injectable.

src/renderer/api/create-terminal-api.injectable.ts
Nokel81: Yes good point.

src/common/utils/channel/channel.test.ts
Nokel81: The contract isn't the same. A request channel being requested on has no meaning on `main` and thus `main` shouldn't even know about it.

src/renderer/components/dock/terminal/store.ts
Nokel81: So that the users of `connect` can be sync for when they want to do things that are not related to the connection being fully active.

src/common/fetch/lens-fetch.injectable.ts
Nokel81: Yeah sure.

src/common/utils/collection-functions.ts
Nokel81: Not when you expect the helper function to be an action

scripts/fix-canvas-deps.ts
Nokel81: Unfortunately I could not find one after banging my head for a few days. The fact that we use Electron seem to break some of the dependencies auto detecting certain configuration things.

src/common/utils/channel/channel.test.ts
jansav: `main` can request stuff from channel. It just doesn't know who answers it, might be main or renderer.

Nokel81: No it cannot, where is that possible?

package.json
Nokel81: Sure also

package.json
Nokel81: Canvas is required for some of the tests IIRC, might not be necessary if we upgrade to xtermJs 5 which adds the DOM based renderer for testing

src/jest.setup.ts
Nokel81: I don't think it is worth it. But maybe

package.json
jansav: With this complexity, I would say that we stop testing xtermjs directly and only test the contract (our wrapper).

PR #5630, opened 2022-06-14
modify app test rendering to include status bar

jim-docker:
With some auto-upate info moving to the status bar from notifications, the snapshots will need to include the status bar in order to reflect this change.


Nokel81 (approved):

jansav (approved):

jansav:
Notice that when we have done our job well, we are switching to use `bootstrap.tsx` directly in unit tests, so that this won't be needed to update anymore. :) Can't do that yet, since there is too many usages of components with shared global state in `cluster-frame.tsx` and `root-frame.tsx`.

PR #5635, opened 2022-06-15
Move renderer phenomenon out from common by introducing different …

jansav:
…implementations for environments

### DIFF
--- a/src/common/utils/is-allowed-resource.injectable.ts
+++ b/src/common/utils/is-allowed-resource.injectable.ts
@@ -4,8 +4,8 @@
  */
 import { getInjectable, lifecycleEnum } from "@ogre-tools/injectable";
 import { computed } from "mobx";
-import allowedResourcesInjectable from "../cluster-store/allowed-resources.injectable";
 import type { KubeResource } from "../rbac";
+import { allowedResourcesInjectionToken } from "../cluster-store/allowed-resources-injection-token";
 
 export type IsAllowedResource = (resource: KubeResource) => boolean;
 
@@ -13,7 +13,7 @@ const isAllowedResourceInjectable = getInjectable({
   id: "is-allowed-resource",
 
   instantiate: (di, resourceName: string) => {
-    const allowedResources = di.inject(allowedResourcesInjectable);
+    const allowedResources = di.inject(allowedResourcesInjectionToken);
 
     return computed(() => allowedResources.get().has(resourceName));
   },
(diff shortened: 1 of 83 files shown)


src/common/utils/is-allowed-resource.injectable.ts
```diff
@@ -4,16 +4,16 @@
...
+import { allowedResourcesInjectionToken } from "../cluster-store/allowed-resources-injection-token";
 
 export type IsAllowedResource = (resource: KubeResource) => boolean;
 
 const isAllowedResourceInjectable = getInjectable({
   id: "is-allowed-resource",
 
   instantiate: (di, resourceName: string) => {
-    const allowedResources = di.inject(allowedResourcesInjectable);
+    const allowedResources = di.inject(allowedResourcesInjectionToken);
```
jansav, line 16: Important bit here. Switching to use `injectionToken`

src/main/create-cluster/allowed-resources.injectable.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { computed } from "mobx";
+import { allowedResourcesInjectionToken } from "../../common/cluster-store/allowed-resources-injection-token";
+
+// TODO: Figure out implementation for this later.
```
jansav, line 9: Important bit here: Dummy implementation for main till we have real one.

jansav:
Everything else is just relocation of files from `common` to `renderer`.

src/main/create-cluster/allowed-resources.injectable.ts
Nokel81: I don't think we can ever have a real one. What I think is that we should try and move `isAllowedResource` to be only within `renderer`.

Nokel81 (dismissed):

src/main/create-cluster/allowed-resources.injectable.ts
jansav: As we discussed in huddle, we surely can. When we have removed the iframe, we can call e.g. `navigateToPods(clusterId)` which then will have the requirement of checking if cluster has pods. Or even application menu containing all clusters and ways to navigate to pages in them. But that's enough speculation :) We'll get back on this when the time comes.

jansav:
Accidently merged: https://github.com/lensapp/lens/pull/5636 to this. Not gonna unmerge them again, since both of these were approved. (just to save time, since these are just chores) @Nokel81 @jim-docker could you re-approve this?

This PR's original changes are in the first commit, so if you need to check it still, review commit by commit.

PR #5641, opened 2022-06-15
Fix html font size and adjust usage of rem units

aleksfront:
Changes made in https://github.com/lensapp/lens/pull/5414/files#diff-e16a68bd7aaebebc826825c0989e76ce22c5ea6927dce3fdd1be53d705238fc3 removed legacy value for setting HTLM font-size:
```
html {
  font-size: 62.5%; // 1 rem == 10px
}
```

This caused all elements where `rem` units are used to increase their actual pixel size. Following by that many components became visually bigger than they should be. For example:
big ns selector
big remove cluster dialog

big catalog title

Current PR provides following changes:
1. It drops custom font-size for the `HTML` element which now takes 16px by default in the most browsers. It basically means that `1rem === 16px` and considered a good standard. Now libraries like tailwindcss will work properly (set proper heights, paddings and margins), because it heavily uses `rems` and expects browser-default font-size as a base (https://v2.tailwindcss.com/docs/customizing-spacing#default-spacing-scale).
2. Current usage of `rem` units adjusted or replaced by our internal sizing variables `var(--padding)` & `var(--margin)`.

catalog title

remove cluster dialog

ns selector

win buttons
cluster sidebar

### DIFF
--- a/src/behaviours/helm-charts/__snapshots__/add-custom-helm-repository-in-preferences.test.ts.snap
+++ b/src/behaviours/helm-charts/__snapshots__/add-custom-helm-repository-in-preferences.test.ts.snap
@@ -886,18 +886,20 @@ exports[`add custom helm repository in preferences when navigating to preference
                     class="repos"
                   >
                     <div
-                      class="item flex gaps align-center justify-space-between mt-3"
+                      class="item flex gaps align-center justify-space-between repo"
                     >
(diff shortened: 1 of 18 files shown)


Nokel81 (approved):
LGTM!

ixrock (approved):

PR #5654, opened 2022-06-16
Try solving flaky unit test by switch to use different implementation…

jansav:
… of setImmediate


jansav:
Nope, didn't solve it.

PR #5588, opened 2022-06-09
Fix triple click breaking inputs

Nokel81:
Signed-off-by: Sebastian Malton <sebastian@malton.name>

fixes https://github.com/lensapp/lens/issues/5580

PR #5434, opened 2022-05-20
Fix envFrom.prefix not being applied in container's env view

Nokel81:
Signed-off-by: Sebastian Malton <sebastian@malton.name>

fixes #5425

PR #5657, opened 2022-06-16
Readd space between key and value within ContainerEnv

### LINKED ISSUE #5656: Space missing between key and value in Environment
There used to be a space between the key and the value in the Environment section of a pod. Now the space is missing, which makes it more difficult to e.g. copy the value.

Nokel81:
- Add some snapshot tests

NOTES:
- Noticed that https://github.com/lensapp/lens/pull/4201 is messing up with the space that was reintroduced in this PR so it is not possible to triple click just the value. I am inclined to close #5588 and instead remove the code and reintroduce #4121 in some form.


src/renderer/components/+workloads-pods/__tests__/pod-container-env.test.tsx
```diff
@@ -0,0 +1,269 @@
...
+        selfLink: "/api/v1/pods/default/my-pod",
+        uid: "1234",
+      },
+      spec: {
+        containers: [container],
+      },
+    });
+    const result = render(<ContainerEnvironment container={container} namespace={pod.getNs()} />);
+
+    expect(result.baseElement).toMatchSnapshot();
```
jansav: Since these are technical unit tests, we could just find the environment value and check that the format is correct. I'd save the snapshots for behaviours so that we can get the feeling from whole application. And there those are used just as additions, never alone to test some business logic in the implementation.

Nokel81: Okay that makes sense will change them.

src/renderer/components/+workloads-pods/pod-container-env.tsx
```diff
@@ -124,75 +149,35 @@ export const ContainerEnvironment = observer((props: ContainerEnvironmentProps)
...
+          className="variable"
+          key={name}
+          data-testid={`envFrom-secret-${secretName}-${name}`}
+        >
+          <span className="var-name">{name}</span>
+          {`= `}
+          <SecretKey
+            secret={secret}
+            field={name}
+          />
```
jansav: This part seems to be duplicated quite many times. Maybe you could introduce non-exported component to bottom of file which does the trick? :)

src/renderer/components/+workloads-pods/pod-container-env.tsx
```diff
@@ -124,75 +149,35 @@ export const ContainerEnvironment = observer((props: ContainerEnvironmentProps)
...
-      );
-    });
+      ));
   };
 
   return (
     <DrawerItem name="Environment" className="ContainerEnvironment">
-      {env && renderEnv()}
-      {envFrom && renderEnvFrom()}
+      {renderEnv()}
```
jansav: This would read better if it was 
```
{env && renderEnv(env)}
{envFrom && renderEnvFrom(envFrom)}
``` 

That would also make `renderEnv` and `renderEnvFrom` more pure-function (and then those could be moved outside of the component.

src/renderer/components/+workloads-pods/__tests__/pod-container-env.test.tsx
```diff
@@ -0,0 +1,273 @@
...
+        selfLink: "/api/v1/pods/default/my-pod",
+        uid: "1234",
+      },
+      spec: {
+        containers: [container],
+      },
+    });
+    const result = render(<ContainerEnvironment container={container} namespace={pod.getNs()} />);
+
+    expect(result.getByTestId("env-foobar").innerHTML).toBe(`<span class="var-name">foobar</span>= https://localhost:12345`);
```
jansav: Hmm. Could it be bit less technical? 

```
expect(result.getByTestId("env-foobar")).toHaveTextContent("foobar= https://localhost:12345");
```

Nokel81: Yes, I completely forgot about those extended `expect` functions.

src/renderer/components/+workloads-pods/pod-container-env/env.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+      if (valueFrom?.secretKeyRef) {
+        const { secretKeyRef } = valueFrom;
+
+        secretStore
+          .load({ name: secretKeyRef.name, namespace })
+          .catch(error => logger.warn(`[CONTAINER-ENV]: failed to load Secret ${secretKeyRef.name} in ns=${namespace}`, error));
+      }
+    }
+  }), []);
```
jansav, line 55: Render-time side-effect. I know you just moved it here, so I don't see this as something you should fix now. 

Real solution would be that this component gets the values that it needs as dependency and that dependency would do the loading etc. Case where `asyncComputed` can be utilized to completely remove the `useEffect` and more importantly `autorun`.

But as said, not now, since you are just extracting it as own component.

src/renderer/components/+workloads-pods/pod-container-env/env-from.tsx
```diff
@@ -0,0 +1,120 @@
...
+          .catch(error => logger.warn(`[CONTAINER-ENV]: failed to load Secret ${secretRef.name} in ns=${namespace}`, error));
+      }
+
+      if (configMapRef?.name) {
+        configMapStore
+          .load({ name: configMapRef.name, namespace })
+          .catch(error => logger.warn(`[CONTAINER-ENV]: failed to load ConfigMap ${configMapRef.name} in ns=${namespace}`, error));
+      }
+    }
+  }), []);
```
jansav, line 51: Render time side-effect. Also looks some-what duplicated. Could be wrong. But as said in other comment, I know you are just extracting this as of component, so no changes needed for now.

src/renderer/components/+workloads-pods/pod-container-env/view.test.tsx
```diff
@@ -0,0 +1,277 @@
...
+describe("<ContainerEnv />", () => {
+  let render: DiRender;
+  let secretStore: jest.Mocked<Pick<SecretStore, "load" | "getByName">>;
+  let configMapStore: jest.Mocked<Pick<ConfigMapStore, "load" | "getByName">>;
+
+  beforeEach(() => {
+    const di = getDiForUnitTesting({ doGeneralOverrides: true });
+
+    secretStore = ({
+      load: jest.fn().mockImplementation(async () => {
```
jansav, line 27: nit: `jest.fn().mockImplementation(...)` is same as `jest.fn(...)`

Improvement for the future: the promise should be controlled, we should test how component looks when promise has no resolved yet. That's where `asyncFn()` comes in.

src/renderer/components/+workloads-pods/pod-container-env/view.test.tsx
```diff
@@ -0,0 +1,277 @@
+/**
```
jansav, line 1: All of these should be in behaviours. But I'm not sure if we have all bits and pieces ready to test `Details`-panels in the behaviours. So OK for now.

src/renderer/components/+workloads-pods/pod-container-env/view.test.tsx
Nokel81: That is incorrect (which is why I did it this way), the types flow differently if you first bind `jest.fn()` and then bind the implementation.

src/renderer/components/+workloads-pods/pod-container-env/env-from.tsx
Nokel81: Probably could just replace this with a top level subscribe instead.

Nokel81:
@jansav So is this good enough to merge?

jansav (dismissed):
Ah, sorry for keeping you waiting. Didn't get any notification about this.

aleksfront (commented):
Quite a big change for what PR title states. It's a way more refactoring kind.

Is there a reason why semicolon `:` changed to `=` for the env variables? And there's no space after var name. See example:

container environment

Nokel81:
@aleksfront It was changed to `=` because that is what unix uses to denote env vars everywhere

As for the additional space between the name and the `=` that wasn't asked for since I didn't bother adding it.

aleksfront:
> @aleksfront It was changed to `=` because that is what unix uses to denote env vars everywhere
> 
> As for the additional space between the name and the `=` that wasn't asked for since I didn't bother adding it.

That's how `kubectl` shows environment info (using semicolon):
kubectl env

aleksfront (approved):
LGTM, however, current PR will probably have conflicts with https://github.com/lensapp/lens/pull/5434

Nokel81:
Yeah I will have to solve the the conflicts after this is merged

PR #5662, opened 2022-06-17
Simplify logic for injecting many instances which can reactively change e.g. based on extension being enabled or disabled

jansav:
1. Introducing `computedInjectMany` which works as `injectMany` does, but always returns type `IComputedValue<T>`. When this value is observed, it is recomputed every time when new implementations for the `injectionToken` are registered or deregistered. 

2. Enabling extension will cause registration of new injectables. Disabling extension will cause deregistration of those injectables. This means that we don't have to care about extension being enabled or disabled anymore when using e.g. `front-end-routes`.

---- 
**Also:** Removed `di.inject(currentPathParametersInjectable)` for being not typed correctly and being easy way to achieve circular dependency. Migrated to using `di.inject(routePathParametersInjectable, route)`.

This was done because switching to use `computedInjectMany` revealed circular dependencies.


package.json
```diff
@@ -208,10 +208,11 @@
...
     "@material-ui/styles": "^4.11.5",
-    "@ogre-tools/fp": "7.1.0",
-    "@ogre-tools/injectable": "7.1.0",
-    "@ogre-tools/injectable-extension-for-auto-registration": "7.1.0",
-    "@ogre-tools/injectable-react": "7.1.0",
+    "@ogre-tools/fp": "8.0.0",
+    "@ogre-tools/injectable": "8.0.0",
+    "@ogre-tools/injectable-extension-for-auto-registration": "8.0.0",
+    "@ogre-tools/injectable-react": "8.0.0",
+    "@ogre-tools/injectable-extension-for-mobx": "8.0.0",
```
Nokel81: Can you add these via `yarn add` and not manually because these are not in sorted order which messes up other additions.

src/behaviours/extension-special-characters-in-page-registrations.test.tsx
```diff
@@ -23,7 +23,7 @@ describe("extension special characters in page registrations", () => {
       extensionWithPagesHavingSpecialCharacters,
     );
 
-    await applicationBuilder.addExtensions(testExtension);
+    applicationBuilder.extensions.renderer.enable(testExtension);
```
Nokel81, line 26: Have we already merged the other PR that fixes these calls to be `await`-ed?

src/renderer/components/+preferences/telemetry-preference-items.injectable.ts
```diff
@@ -3,48 +3,22 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
 import { getInjectable, getInjectionToken } from "@ogre-tools/injectable";
-import rendererExtensionsInjectable from "../../../extensions/renderer-extensions.injectable";
-import { pipeline } from "@ogre-tools/fp";
-import { filter, overSome } from "lodash/fp";
-import { computed } from "mobx";
-import type { LensRendererExtension } from "../../../extensions/lens-renderer-extension";
+import { computedInjectManyInjectable } from "@ogre-tools/injectable-extension-for-mobx";
```
Nokel81, line 6: Why is this an injectable?

src/behaviours/preferences/navigation-to-extension-specific-preferences.test.tsx
```diff
@@ -44,11 +44,11 @@ describe("preferences - navigation to extension specific preferences", () => {
     });
 
     describe("when extension with specific preferences is enabled", () => {
-      beforeEach(() => {
+      beforeEach(async () => {
         const getRendererExtensionFake = getRendererExtensionFakeFor(applicationBuilder);
         const testExtension = getRendererExtensionFake(extensionStubWithExtensionSpecificPreferenceItems);
 
-        applicationBuilder.addExtensions(testExtension);
+        await applicationBuilder.extensions.renderer.enable(testExtension);
```
Nokel81: unnecessary await?

Nokel81 (dismissed):

package.json
jansav: Done

src/behaviours/extension-special-characters-in-page-registrations.test.tsx
jansav: This PR makes `enable` synchronous.

Nokel81 (approved):

src/renderer/components/+preferences/telemetry-preference-items.injectable.ts
jansav: It has its own dependencies.

src/behaviours/preferences/navigation-to-extension-specific-preferences.test.tsx
jansav: Good catch, removed.

Iku-turso (approved):

PR #5450, opened 2022-05-25
Fix crash when using an inline svg Icon

Nokel81:
Signed-off-by: Sebastian Malton <sebastian@malton.name>

fixes #5448

PR #5667, opened 2022-06-17
Fix crash in HelmChartDetails

### LINKED ISSUE #5651: Helm Chart open app crash
```
Uncaught TypeError: abort_controller__WEBPACK_IMPORTED_MODULE_14__.AbortController is not a constructor
    at new NonInjectedHelmChartDetails (helm-chart-details.tsx:55)
    at constructClassInstance (react-dom.development.js:12716)
    at updateClassComponent (react-dom.development.js:17425)
    at beginWork (react-dom.development.js:19073)
    at HTMLUnknownElement.callCallback (react-dom.development.js:3945)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:3994)
    at invokeGuardedCallback (react-dom.development.js:4056)
    at beginWork$1 (react-dom.development.js:23964)
    at performUnitOfWork (react-dom.development.js:22776)
    at workLoopSync (react-dom.development.js:22707)
```

Need to check if it's only related to dev mode (since error mentioned webpack).

https://user-images.githubusercontent.com/9607060/174011324-02e1d434-23b9-4548-858a-50080d6bba67.mov

Nokel81:

fixes https://github.com/lensapp/lens/issues/5651


aleksfront (commented):
While testing I found an issue where there's a loop trying to fetch helm chart image for those charts in which there is no image initially.

https://user-images.githubusercontent.com/9607060/174541089-3f93b2ee-8fce-4852-9879-594b2629e239.mov

I do not know if the issue presented only in this PR or also in `master` because charts in master are broken. But it's worth to fix it there if possible.

Nokel81:
@aleksfront I think that is a longer standing bug, a regression probably from a combination of the webpack 5 upgrade and #5450

Nokel81:
@aleksfront Fixed and pushed PTAL.

src/renderer/components/+helm-charts/helm-chart-details.tsx
```diff
@@ -46,6 +46,7 @@ interface Dependencies {
 
 @observer
 class NonInjectedHelmChartDetails extends Component<HelmChartDetailsProps & Dependencies> {
+  private readonly imgLoadingFailed = observable.set<string>(); // The IDs of the HelmChart instances
```
aleksfront: A code readability comment: the name of the variables should be self-descriptive. There is something wrong with it if we have such a comment afterwards.

How about that?

```
private readonly helmChartsWithFailedImagesLoad = ...
```

src/renderer/components/+helm-charts/helm-chart-details.tsx
```diff
@@ -118,14 +119,32 @@ class NonInjectedHelmChartDetails extends Component<HelmChartDetailsProps & Depe
...
 
+  private renderIcon(chart: HelmChart) {
+    const icon = chart.getIcon();
+
+    if (!icon || this.imgLoadingFailed.has(chart.getId())) {
+      return (
+        <div
+          className="intro-logo"
+          dangerouslySetInnerHTML={{ __html: HelmLogoPlaceholder }}
+        />
```
aleksfront: Can't we use `img` with `HelmLogoPlaceholder` as its `src` to avoid using something like `dangerouslySetInnerHTML={{ __html: HelmLogoPlaceholder }}`

src/renderer/components/+helm-charts/helm-charts.tsx
```diff
@@ -37,6 +39,8 @@ interface Dependencies {
 
 @observer
 class NonInjectedHelmCharts extends Component<Dependencies> {
+  private readonly imgLoadingFailed = observable.set<string>(); // The IDs of the HelmChart instances
```
aleksfront: We should use self-explanatory variable name, this will allow us to avoid such comments.

src/renderer/components/+helm-charts/helm-charts.tsx
```diff
@@ -76,6 +80,22 @@ class NonInjectedHelmCharts extends Component<Dependencies> {
...
+
+    return (
+      <img
+        src={icon}
+        onLoad={evt => evt.currentTarget.classList.add("visible")}
+        onError={() => this.imgLoadingFailed.add(chart.getId())}
+      />
+    );
+  }
+
```
aleksfront: Seems like repeatable code from `helm-chart-details.tsx`. Good chance to move it to separate component?

src/renderer/components/+helm-charts/get-char-details.injectable.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export type GetChartDetails = (repo: string, name: string, opts?: GetChartDetailsOptions) => Promise<IHelmChartDetails>;
+
+const getChartDetailsInjectable = getInjectable({
+  id: "get-chart-details",
+  instantiate: (): GetChartDetails => getChartDetails,
+  causesSideEffects: true,
+});
+
+export default getChartDetailsInjectable;
```
aleksfront, line 17: @jansav Do we need to create an injectable for such cases where there're no injectables inside of it? Why not just use `getChartDetails` function directly?

aleksfront:
> @aleksfront Fixed and pushed PTAL.

Thanks. Checked that bug is fixed now.

src/renderer/components/+helm-charts/helm-chart-details.tsx
Nokel81: No because we currently bundle the `.svg` as the raw avg data, not as a data URL. And we have to continue bundling like that because we have to keep `<Icon>` from breaking.

I could factor out that and make it a seperate component of we want.

src/renderer/components/+helm-charts/get-char-details.injectable.ts
Nokel81: This is so that I can override it in tests. In the future it would be better to make this sort of route fully injectable but that would take too long.

Nokel81:
@aleksfront I have factored out `HelmChartIcon` which also nicely solves the variable naming problem.

src/renderer/components/+helm-charts/helm-chart-details.tsx
aleksfront: Removed usage of `dangerouslySetInnerHTML={{ __html: HelmLogoPlaceholder }}` and used inline version of placeholder icon. Plus, adjusted list styles a bit.

helm chart placeholder light

helm chart placeholder

aleksfront (dismissed):

aleksfront (approved):

src/renderer/components/+helm-charts/get-char-details.injectable.ts
jansav: @Nokel81 approach seems right. In the future we might want to override something else which `getChartDetails` gets as dependency. But for now this seems fine.

src/renderer/components/+helm-charts/icon.tsx
```diff
@@ -0,0 +1,49 @@
...
+  chart,
+  className,
+}: HelmChartIconProps) => {
+  const [failedToLoad, setFailedToLoad] = useState(false);
+  const icon = chart.getIcon();
+
+  if (!icon || failedToLoad) {
+    return (
+      <div className={className}>
+        <svg viewBox="0 0 722.8 702" xmlns="http://www.w3.org/2000/svg">
```
jansav, line 24: nit: This `svg` should be in separate file. Now it's impossible to preview.

src/renderer/components/notifications/show-checked-error.injectable.ts
```diff
@@ -0,0 +1,32 @@
...
+import { getInjectable } from "@ogre-tools/injectable";
+import { JsonApiErrorParsed } from "../../../common/k8s-api/json-api";
+import loggerInjectable from "../../../common/logger.injectable";
+import type { Disposer } from "../../utils";
+import type { CreateNotificationOptions } from "./notifications.store";
+import showErrorNotificationInjectable from "./show-error-notification.injectable";
+
+export type ShowCheckedErrorNotification = (message: unknown, fallback: string, opts?: CreateNotificationOptions) => Disposer;
+
+const showCheckedErrorNotificationInjectable = getInjectable({
```
jansav, line 14: I don't understand the concept of `checkedErrorNotification`. Seems like something that shouldn't exist.

src/renderer/components/+helm-charts/helm-chart-details.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+import type { CreateInstallChartTab } from "../dock/install-chart/create-install-chart-tab.injectable";
+import createInstallChartTabInjectable from "../dock/install-chart/create-install-chart-tab.injectable";
+import { Notifications } from "../notifications";
+import type { DiRender } from "../test-utils/renderFor";
+import { renderFor } from "../test-utils/renderFor";
+import type { GetChartDetails } from "./get-char-details.injectable";
+import getChartDetailsInjectable from "./get-char-details.injectable";
+import { HelmChartDetails } from "./helm-chart-details";
+
+describe("<HelmChartDetails />", () => {
```
jansav, line 24: As you might have guessed, this should tested as part of behaviour. :) But yeah, I know that this part of application is not that easy to access in behavioural unit tests yet.

src/renderer/components/+helm-charts/icon.tsx
Nokel81: Yeah... @aleksfront converted move this instead of the previous version which `dangerouslySetInnerHTML`. I think having a component plus svg file would be best.

src/renderer/components/notifications/show-checked-error.injectable.ts
Nokel81: It is a helper around `showErrorNotification`, maybe it should be called `checkedShowErrorNotification` instead. It does some additional logic to make sure that the value that gets through to "reasonable" with a declarative fallback message for better UX.

jansav (approved):

PR #5687, opened 2022-06-21
fix jsonnet error with kube-prometheus >= v0.8

elvischan-crypto:
kubelet ServiceMonitor moved to kubernetesControlPlane starting from v0.8

### DIFF
--- a/jsonnet/lens/custom-prometheus.jsonnet
+++ b/jsonnet/lens/custom-prometheus.jsonnet
@@ -24,7 +24,7 @@
       },
     },
   },
-  prometheus+:: {
+  kubernetesControlPlane+:: {
     serviceMonitorKubelet+: {
       spec+: {
         endpoints: std.map(function(endpoint)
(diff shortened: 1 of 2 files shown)


Nokel81:
Thanks for the PR, given that some people might still be running 0.8, I think that this should probably be a switch or maybe dependent on the version.

elvischan-crypto:
sure, and my bad that it should be >= 0.8.

Also, with the cloud provide phasing out the k8s 1.19 such as eks 1.19 EOS end of this month, I think more and more people would need to update >= 0.8

Nokel81:
Oh I have no doubt that more people will be using the newer versions. We tend to try and support more than that though.

Nokel81:
@elvischan-crypto To get this PR merged, I think all that would be needed is adding a note in `troubleshooting/custom-prometheus.md` that says that the file is for >= v0.8

Nokel81 (approved):

elvischan-crypto:
sorry for late reply, thanks for updating the docs

PR #5690, opened 2022-06-21
Helper to resolve proxy from URL in any environment, or using extension-API

### LINKED ISSUE #5689: Extension api for resolving a proxy for an url
Api for something like:

```ts
electron.webContents.getAllWebContents()[0]?.session.resolveProxy(url)
```

Iku-turso:
Note: given there are no open windows (ie. webContents), the helper will throw, and log the error.

Closes #5689


src/behaviours/proxy/resolve-proxy/main/electron.injectable.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import electron from "electron";
+
+const electronInjectable = getInjectable({
```
jakolehm: `electronInjectable` sounds like something that is common for other stuff. Are we going to move this whenever there is a need?

src/extensions/common-api/proxy.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import { asLegacyGlobalFunctionForExtensionApi } from "../as-legacy-globals-for-extension-api/as-legacy-global-function-for-extension-api";
+import { resolveProxyInjectionToken } from "../../behaviours/proxy/resolve-proxy/common/resolve-proxy-injection-token";
+
+export const resolveProxy = asLegacyGlobalFunctionForExtensionApi(resolveProxyInjectionToken);
```
jakolehm: Is there a "non-legacy" way to export functions for extensions?

jakolehm: `resolveProxy` as a name is very generic (would `resolveSystemProxy` make more sense?). What kind of behavior extension api user expects from this? 🤔 At least we should have api docs (tsdoc) that shortly summarizes what it is.

jakolehm (commented):
Functionality looks good but I'm not sure about the api naming and the fact that we decided to use `src/behaviours` for this kind of stuff. Folder structure is something we can shuffle later but we probably need to rethink the api name _or_ change the functionality a bit.

src/behaviours/proxy/resolve-proxy/main/electron.injectable.ts
Iku-turso: Right you are, it is general, and will move to more shared scope when something else requires it :)

src/extensions/common-api/proxy.ts
Iku-turso: > Is there a "non-legacy" way to export functions for extensions?

All stuff exposed in current extension-API is involuntary **global state**. A completely clean function can be exported as it is, "non-legacy"... But that other stuff with side-effects and lifecycles and dependencies and all, cannot be but packaged as legacy :/

This is a considerable damage that could be fluently solved in next iteration of extension-API.

Iku-turso: > would resolveSystemProxy make more sense?

A bit I guess. Renamed :)

Iku-turso: > What kind of behavior extension api user expects from this? 🤔 At least we should have api docs (tsdoc) that shortly summarizes what it is.

Hmm. Electron's own documentation for `session.resolveProxy()` is pretty [terse](https://www.electronjs.org/docs/latest/api/session#sesresolveproxyurl), as it just says "Resolves with the proxy information for `url`." Duh. I suppose you already have better understanding for what this does? Share? :)

Iku-turso: TSDoc added from conversation elsewhere.

jakolehm (approved):
LGTM

jansav (approved):

PR #5691, opened 2022-06-21
Fix check on NamespaceSelectFilter not updating

### LINKED ISSUE #5639: Namespace selector doesn't update checks
There is 2 problems with namespace selector:

1. It changes it's width unexpectedly.
2. It doesn't update ticks showing what namespace selected (all items are always checked).

https://user-images.githubusercontent.com/9607060/173817816-cd46100c-5ad8-4660-9bb0-2fafbc62f0c1.mov

Nokel81:

fixes https://github.com/lensapp/lens/issues/5639

### DIFF
--- a/src/renderer/components/+namespaces/namespace-select-filter-model/namespace-select-filter-model.tsx
+++ b/src/renderer/components/+namespaces/namespace-select-filter-model/namespace-select-filter-model.tsx
@@ -3,7 +3,7 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
 import React from "react";
-import { observable, action, computed, makeObservable, comparer } from "mobx";
+import { observable, action, computed, makeObservable, comparer, reaction } from "mobx";
 import type { NamespaceStore } from "../store";
 import type { ActionMeta } from "react-select";
 import { Icon } from "../../icon";
@@ -32,11 +32,28 @@ export class NamespaceSelectFilterModel {
   constructor(private readonly dependencies: Dependencies) {
     makeObservable(this);
     autoBind(this);
+
+    reaction(
+      () => this.menuIsOpen.get(),
+      (isOpen) => {
+        if (!isOpen) { // falling edge of menu being open
+          this.optionsSortingSelected.replace(this.selectedNames.get());
+        }
+      },
+    );
   }
 
-  readonly selectedNames = computed(() => new Set(this.dependencies.namespaceStore.contextNamespaces), {
+  readonly menuIsOpen = observable.box(false);
+
+  private readonly selectedNames = computed(() => new Set(this.dependencies.namespaceStore.contextNamespaces), {
     equals: comparer.structural,
   });
+
+  /**
+   * This set is only updated on the falling edge of the menu being open. That way while the menu is
+   * open the order of the items doesn't change
+   */
+  private readonly optionsSortingSelected = observable.set<string>(this.selectedNames.get());
 
   readonly options = computed((): readonly NamespaceSelectFilterOption[] => {
     const baseOptions = this.dependencies.namespaceStore.items.map(ns => ns.getName());
@@ -44,8 +61,8 @@ export class NamespaceSelectFilterModel {
 
     baseOptions.sort((
       (left, right) =>
-        +selectedNames.has(right)
-        - +selectedNames.has(left)
+        +this.optionsSortingSelected.has(right)
+        - +this.optionsSortingSelected.has(left)
     ));
 
     return [
@@ -81,8 +98,6 @@ export class NamespaceSelectFilterModel {
       </div>
     );
   }
-
-  readonly menuIsOpen = observable.box(false);
 
   @action
   closeMenu() {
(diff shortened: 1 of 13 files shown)


jim-docker:
I think this is a regression:

https://user-images.githubusercontent.com/40840436/175970196-e09e0080-56b4-4e3b-ae2a-f27a8f8be4e9.mov

Scroll the namespace list down, then Cmd+Click on an item. List scrolls to top.

Nokel81:
@jim-docker Yes that is a bug, and I also noticed that the placeholder is wrong sometimes too in your video.

Nokel81:
@jim-docker Resolved that issue and added some more tests. PTALA

jim-docker:
Looks good now. Something still odd about the dynamic width changing of the placeholder and actual dropdown. They're a bit out of sync (dropdown list doesn't adjust till you move to a different item):

https://user-images.githubusercontent.com/40840436/176048114-6aafcce3-ee8b-4dcf-949d-f0531bd5fd0f.mov

Do we really need the namespace selector to adjust its width dynamically? Looks weird to me. Just set it to its max width (as per how much width is available to it)?

Nokel81:
Yeah I can set it to a fixed width.

src/renderer/components/+namespaces/namespace-select-filter.test.tsx
```diff
@@ -0,0 +1,252 @@
...
+              });
+
+              describe("when 'test-13' is clicked", () => {
+                beforeEach(() => {
+                  result.getByText("test-13").click();
+                });
+
+                it("has all of 'test-1', 'test-3', and 'test-13' selected in the store", () => {
+                  expect(new Set(namespaceStore.contextNamespaces)).toEqual(new Set(["test-1", "test-3", "test-13"]));
+                });
```
jim-docker, line 170: Could also test that when one of them is clicked again it is no longer in  `namespaceStore.contextNamespaces`?

jim-docker:
> Yeah I can set it to a fixed width.

just my opinion, others might prefer the dynamic width?

src/renderer/components/+namespaces/namespace-select-filter-model/namespace-select-filter-model.tsx
```diff
@@ -3,157 +3,179 @@
...
+  let didToggle = false;
+  let isMultiSelection = false;
+  const menuState = observableCrate(SelectMenuState.Close, [{
+    from: SelectMenuState.Close,
+    to: SelectMenuState.Open,
+    onTransition: () => {
+      optionsSortingSelected.replace(selectedNames.get());
+      didToggle = false;
+    },
+  }]);
```
jansav, line 58: As discussed earlier, this is basically `reaction` so we shouldn't need it at all if we would reorganize code. So it's good for now, but I wouldn't create more usages on `observableCrate` since usage of `reaction` and `autorun` should be very limited in `renderer`. (`mobx-react` does the trick already for us.)

src/renderer/components/+namespaces/namespace-select-filter-model/namespace-select-filter-model.tsx
```diff
@@ -3,157 +3,179 @@
...
+              data-testid={`namespace-select-filter-option-${option.value}-selected`}
+            />
+          )}
+        </div>
+      );
+    },
+  };
+
+  return model;
 }
```
jansav, line 181: In the future we should separate the logic for `Select` and `NamespaceSelect` (e.g. whether select should be closed after option is selected is responsibility of `Select`. Same goes for multi selection).

`NamespaceSelect` shouldn't be any different from any other selects. I'd like to argue that we don't when need it when we separate the business logic for getting options, what happens when option is selected and the UI code.

But as said, that's for the future.

src/renderer/components/+namespaces/namespace-select-filter.test.tsx
```diff
@@ -0,0 +1,252 @@
...
+      createNamespace("test-5"),
+      createNamespace("test-6"),
+      createNamespace("test-7"),
+      createNamespace("test-8"),
+      createNamespace("test-9"),
+      createNamespace("test-10"),
+      createNamespace("test-11"),
+      createNamespace("test-12"),
+      createNamespace("test-13"),
+    ]);
```
jansav, line 59: 1. Setting `observable` should always happen in `action`.
2. In this case it would be better if you would change order of lines 45 and 61 so that `render` happens first. That would make sure that the component is prepared for reactive data.

src/renderer/components/+namespaces/namespace-select-filter.test.tsx
```diff
@@ -0,0 +1,252 @@
...
+              beforeEach(() => {
+                result.getByText("test-3").click();
+              });
+
+              it("renders", () => {
+                expect(result.baseElement).toMatchSnapshot();
+              });
+
+              it("has both 'test-1' and 'test-3' as selected in the store", () => {
+                expect(new Set(namespaceStore.contextNamespaces)).toEqual(new Set(["test-1", "test-3"]));
```
jansav, line 152: Why are you using `Set` here? 

```typescript
expect(namespacesStore.contextNamespaces).toEqual(["test-1", "test-3"]);
```

jansav (approved):

src/renderer/components/+namespaces/namespace-select-filter.test.tsx
Nokel81: So that order doesn't matter.

jim-docker (approved):

PR #5712, opened 2022-06-23
Start doing Architectural Decision Records (ADRs)

Iku-turso:
As reference for Architecture Decision Records, there's [this article](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions).

This PR also introduces related CLI-tool to maintain the ADRs, [adr-tools](https://github.com/npryce/adr-tools).

### DIFF
--- a/.adr.json
+++ b/.adr.json
@@ -1,0 +1,6 @@
+{
+  "language": "en",
+  "path": "docs/architecture/decisions/",
+  "prefix": "",
+  "digits": 4
+}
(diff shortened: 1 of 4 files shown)


Nokel81:
How does this introduce `adr-tools`?

jansav:
> How does this introduce `adr-tools`?

Hm. It doesn't. I'll update the PR to contain them in near future since this is not that important yet.

jansav (approved):

jakolehm (approved):

PR #5728, opened 2022-06-24
Fix selecting items always scrolling to top

### LINKED ISSUE #5699: Table scrolls to selected element on click
If selecting any element within Table, the component scrolls it in way to make clicked element to be visible first.

Table should not scroll when selecting its elements with mouse.

https://user-images.githubusercontent.com/9607060/175027200-4d205ea6-fb90-4694-9d32-77c161cb7b55.mov

Nokel81:

fixes #5699


src/renderer/components/virtual-list/virtual-list.test.tsx
```diff
@@ -0,0 +1,81 @@
+/**
```
aleksfront, line 1: It's nice we have such tests now.

aleksfront (approved):
LGTM

ixrock (approved):

Issue #5208, opened 2022-04-06
Must upgrade immediately modal


Nokel81:
**What would you like to be added**:
If Lens has been running for a long time with an update downloaded and pending then a modal should be displayed requiring the user to quit so that the update can be applied.

PhilippKuntschik:
how would this perform if I build the binary from scratch? see discussion [#5444 ](https://github.com/lensapp/lens/issues/5444#issuecomment-1173742061)

Nokel81:
@PhilippKuntschik Nothing will happen. This only affects instances that have a configured update source which isn't configured in this repo at all.

PhilippKuntschik:
it would probably be good to have https://github.com/lensapp/lens/releases/latest as an update source once https://github.com/lensapp/lens/pull/5772 is merged

disconn3ct:
> @PhilippKuntschik Nothing will happen. This only affects instances that have a configured update source which isn't configured in this repo at all.

Why is this being added to a repo that it is explicitly not used in? Can we add other unrelated stuff?

PR #5737, opened 2022-06-27
Split root and cluster frames in smaller pieces 

jansav:
Motivation: 

Make it possible to add new component (e.g. the force-updating modal which will come in next PR) without modifications. Include these components in behavioural unit tests (some of them are still overridden for causing side-effects)

Other small stuff:
~~1. Make notifications in application update deterministic. (caused some flaky snapshots)~~
2. Simplify one existing unit test now that `deleteClusterDialog` is included in behaviours.
3. Remove import time side-effect causing memory leak

**Most of the changes are snapshots, so please, don't waste your time on them. :) There is filter in review tab that you can use to remove all `.snap` files.**

Note: this is prerequisite for upcoming PR about #5208.


src/behaviours/application-update/installing-update.test.ts
```diff
@@ -89,7 +89,10 @@ describe("installing update", () => {
       });
 
       it("notifies the user that checking for updates is happening", () => {
-        expect(showInfoNotificationMock).toHaveBeenCalledWith("Checking for updates...");
+        expect(showInfoNotificationMock).toHaveBeenCalledWith(
```
jim-docker: https://github.com/lensapp/lens/pull/5361 removes `showInfoNotificationMock` as update status info is now put on the status bar instead of using notifications.

src/main/getDiForUnitTesting.ts
```diff
@@ -112,23 +113,27 @@ export function getDiForUnitTesting(opts: { doGeneralOverrides?: boolean } = {})
...
+  const filePaths = getInjectableFilePaths();
 
-    di.register({
-      ...injectableInstance,
-      aliases: [injectableInstance, ...(injectableInstance.aliases || [])],
-    });
-  }
+  const injectables = filePaths.map(filePath => require(filePath).default);
+
+  chunk(100)(injectables).forEach(chunkInjectables => {
```
jim-docker, line 119: performance enhancement?
Don't need `aliases` any more?

src/renderer/application-update/application-update-status-listener.injectable.ts
```diff
@@ -19,25 +19,25 @@ const applicationUpdateStatusListenerInjectable = getInjectable({
     const eventHandlers: Record<ApplicationUpdateStatusEventId, { handle: (version?: string) => void }> = {
       "checking-for-updates": {
         handle: () => {
-          showInfoNotification("Checking for updates...");
+          showInfoNotification("Checking for updates...", { id: "checking-for-updates" });
```
jim-docker: This file (and associated broadcast channel files) are removed in https://github.com/lensapp/lens/pull/5361

src/renderer/getDiForUnitTesting.tsx
```diff
@@ -103,8 +111,35 @@ export const getDiForUnitTesting = (opts: { doGeneralOverrides?: boolean } = {})
 
     di.override(lensResourcesDirInjectable, () => "/irrelevant");
 
-    di.override(watchHistoryStateInjectable, () => () => () => {});
+    // TODO: Remove side-effects and shared global state
+    di.override(commandContainerRootFrameChildComponentInjectable, () => ({
```
jim-docker, line 116: what about the other root frame child components? They don't need overriding?

jim-docker (dismissed):
looks good

src/behaviours/application-update/installing-update.test.ts
jansav: Good. :)

src/main/getDiForUnitTesting.ts
jansav: `aliases` was leftover from the days we used them. Library haven't supported aliases for a while anymore.

> performance enhancement?

Yep, this is due the fact that too many arguments for a function seems to slow it down. Weird, but that's the case.

src/renderer/application-update/application-update-status-listener.injectable.ts
jansav: Good.

src/renderer/getDiForUnitTesting.tsx
jansav: No, `commandContainer` seems to be only one.

jansav:
This is now ready to be reviewed. @jim-docker @Nokel81. Fixed the memory leak caused by import time side-effect to make the unit tests pass.

src/main/getDiForUnitTesting.ts
```diff
@@ -111,23 +112,27 @@ export function getDiForUnitTesting(opts: { doGeneralOverrides?: boolean } = {})
...
     di.override(electronInjectable, () => ({}));
     di.override(waitUntilBundledExtensionsAreLoadedInjectable, () => async () => {});
     di.override(getRandomIdInjectable, () => () => "some-irrelevant-random-id");
-    di.override(hotbarStoreInjectable, () => ({ load: () => {} }));
+
+    di.override(hotbarStoreInjectable, () => ({
+      load: () => {},
+      getActive: () => ({ name: "some-hotbar", items: [] }),
+      getDisplayIndex: () => "0",
+    }) as unknown as HotbarStore);
```
Nokel81, line 134: Thinking out loud.

I wonder if it would be better if we could somehow mock out `getConfigurationFileModelInjectable` instead of the individual stores.

Nokel81 (approved):

jim-docker (approved):

src/main/getDiForUnitTesting.ts
jansav: Agreed. It will come naturally in the future. Key thing is that stores contain so much business logic which we would like to keep in the scope of behavioural tests.

Issue #5786, opened 2022-07-05
Extension API fails to respect `isEnabledForCluster`


Nokel81:
Because of the changes being made to `ExtensionLoader` the `isEnabledForCluster()` method on extensions is not being respected for most registrators. This means that even if that method is `return false` an extension's `clusterPages`, `clusterPageMenus`, etc... will still be shown.

Affects at least 5.5 and master, probably also some of 5.4.

Should not be referenced for the following (since these are only used in the root frame):
- `entitySettings`
- `globalPages`
- `catalogEntityDetailItems`
- `appPreferences`
- `appPreferenceTabs`
- `statusBarItems`
- `commands` (but should be called for commands in the cluster frame)
- `welcomeMenus`
- `welcomeBanners`
- `topBarItems`
- `additionalCategoryColumns`
- `customCategoryViews`

Should be references for the following:

- [x] `clusterPages` #5800 
- [x] `clusterPageMenus` #5801 
- [x] `kubeObjectStatusTexts` #5802 
- [x] `kubeObjectDetailItems` #5805 
- [x] `kubeObjectMenuItems` #5803 
- [x] `kubeWorkloadsOverviewItems` #5804 
- [ ] #5830

PR #5800, opened 2022-07-07
Introduce kludge to make "isEnabledForCluster" work again for cluster pages PART 1

jansav:
This is first part of PRs to tackle #5786. 

# What

`isEnabledForCluster` hasn't worked properly for a while. Here implement kludge to get it working again. Real solution is to expose `enabled: IComputedValue<boolean>` to Extension API which will be done in latter PR. Plan is to deprecate `isEnabledForCluster` and expose `enabled` so that extensions can start the migration. However, we don't want to break the Extension API so that's why we need to keep supporting `isEnabledForCluster` for a while.


src/extensions/extension-loader/extension-is-enabled-for-cluster.injectable.ts
```diff
@@ -0,0 +1,30 @@
...
+  cluster: KubernetesCluster;
+}
+
+const extensionIsEnabledForClusterInjectable = getInjectable({
+  id: "extension-is-enabled-for-cluster",
+
+  instantiate: async (
+    di,
+    { extension, cluster }: ExtensionIsEnabledForCluster,
+  ) => (await extension.isEnabledForCluster(cluster)) as boolean,
```
Nokel81, line 20: Why is this cast needed?

Nokel81 (dismissed):

src/extensions/extension-loader/extension-is-enabled-for-cluster.injectable.ts
jansav: `isEnabledForCluster` is `(cluster: KubernetesCluster) => Promise<Boolean>` which is different than `Promise<boolean>` And in the codebase there is more usages on `boolean` than `Boolean`

Nokel81: Oh I see, that certainly wasn't intended.

jansav: Yep. Can't fix the type in that function because it would be breaking change :(

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -282,7 +283,7 @@ export class ExtensionLoader {
       const extension = ext as LensRendererExtension;
 
       // getCluster must be a callback, as the entity might be available only after an extension has been loaded
-      if ((await extension.isEnabledForCluster(entity)) === false) {
+      if (await this.dependencies.getExtensionIsEnabledForCluster(extension, entity)) {
```
jim-docker: Was the `=== false` a bug?

jim-docker (dismissed):

jim-docker:
This is probably a different issue but I noticed that the cluster overview page clobbers an extension's cluster page when there is a cluster manager (or watch) update:
 
https://user-images.githubusercontent.com/40840436/178317853-05d8e407-101b-4c7d-8de8-31399bff062e.mov

Nokel81 (dismissed):

src/extensions/extension-loader/extension-loader.ts
jansav: Good catch. Fixed, but this is removed in latter PR anyway.

jansav:
> This is probably a different issue but I noticed that the cluster overview page clobbers an extension's cluster page when there is a cluster manager (or watch) update:

Actually it was bug here. It was caused because cluster metadata is updated with every watch causing `computed` to re-compute. 

Fixed now

jim-docker (dismissed):

jim-docker (dismissed):

jim-docker (approved):

Nokel81 (approved):

PR #5803, opened 2022-07-07
Introduce kludge to make "isEnabledForCluster" work again for kube object menu items PART 4

jansav:
Part 4 of #5800 


src/renderer/components/kube-object-menu/kube-object-menu-item-injection-token.ts
```diff
@@ -0,0 +1,21 @@
...
+import type { IComputedValue } from "mobx";
+import type { KubeObjectMenuProps } from "./kube-object-menu";
+
+export interface KubeObjectMenuItem {
+  kind: string;
+  apiVersions: string[];
+  enabled: IComputedValue<boolean>;
+
+  // TODO: Figure out why "KubeObject" does not work here.
+  Component: React.ElementType<KubeObjectMenuProps<any>>;
```
Nokel81: The reason that using `KubeObject` here doesn't work is that the components being registered say that they require `class T extends KubeObject` which is not the same and typescript rightly complains that any arbitrary instance unrelated to `T` could be passed in.

Personally, I would use `KubeObject` here (and factor out this type as:

```ts
export type KubeObjectMenuItemComponent = React.ElementType<KubeObjectMenuProps<KubeObject>>;
```

and then at the uses of it I would add an `as KubeObjectMenuItemComponent` cast.

src/renderer/components/kube-object-menu/kube-object-menu-items.injectable.ts
```diff
@@ -0,0 +1,44 @@
...
+        ),
+
+        sortBy((item) => item.orderNumber),
+        map((item) => item.Component),
+      ),
+    );
+  },
+
+  lifecycle: lifecycleEnum.keyedSingleton({
+    getInstanceKey: (di, { kubeObject }: { kubeObject: KubeObject }) =>
```
Nokel81: shouldn't it be `KubeObject | undefined`?

Nokel81 (approved):

src/renderer/components/kube-object-menu/kube-object-menu-item-injection-token.ts
jansav: Changed.

src/renderer/components/kube-object-menu/kube-object-menu-items.injectable.ts
jansav: Hmm. In here it looks like it yeah, but from the perspective of usage (kube-object-menu) it doesn't. I'd would like to be sure it's not undefined.

PR #5804, opened 2022-07-07
Introduce kludge to make "isEnabledForCluster" work again for workload overview details PART 5

jansav:
Part 5 of #5800


src/renderer/components/+workloads-overview/workload-overview-details/workload-overview-detail-registrator.injectable.ts
```diff
@@ -0,0 +1,57 @@
...
+          id,
+
+          instantiate: () => ({
+            Component: registration.components.Details,
+
+            enabled: computed(() =>
+              extensionShouldBeEnabledForClusterFrame.value.get(),
+            ),
+
+            orderNumber: 200 - (registration.priority || 50),
```
Nokel81: where does this come from? I don't remember ever seeing a `200`.

jansav: Just picked some larger number than we use in our items to make items coming from extension appear last.

Nokel81: I think that it was specifically designed so that the extension API could have items appear before the in-tree ones. Plus this doesn't really ensure that.

jansav: > Plus this doesn't really ensure that.

Sure it doesn't. 

> I think that it was specifically designed so that the extension API could have items appear before the in-tree ones.

OK, I'll change it.

PR #5805, opened 2022-07-07
Introduce kludge to make "isEnabledForCluster" work again for kube object details PART 6

jansav:
Part 6 of #5800 

In this PR, we replace `KubeObjectDetailsRegistry` with new alternative which works properly with `isEnabledForCluster`.

Signed-off-by: Janne Savolainen <janne.savolainen@live.fi>

Closes #4632

PR #5816, opened 2022-07-08
feat: New view for PriorityClasses under Config menu

dex4er:
This is new view for PriorityClasses v1 which are since Kubernetes v1.14 and there is no other way to present this resource in Lens.

Screenshots:

image

image

image

### DIFF
--- a/src/common/k8s-api/endpoints/priority-class.api.injectable.ts
+++ b/src/common/k8s-api/endpoints/priority-class.api.injectable.ts
@@ -1,0 +1,19 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import assert from "assert";
+import { storesAndApisCanBeCreatedInjectionToken } from "../stores-apis-can-be-created.token";
+import { PriorityClassApi } from "./priority-class.api";
+
+const priorityClassApiInjectable = getInjectable({
+  id: "priority-class-api",
+  instantiate: (di) => {
+    assert(di.inject(storesAndApisCanBeCreatedInjectionToken), "PriorityClassApi is only available in certain environments");
+
+    return new PriorityClassApi();
+  },
+});
+
+export default priorityClassApiInjectable;
--- a/src/renderer/components/+config-priority-classes/priority-classes-details.tsx
+++ b/src/renderer/components/+config-priority-classes/priority-classes-details.tsx
@@ -9,9 +9,8 @@ import React from "react";
 import { observer } from "mobx-react";
 import { DrawerItem } from "../drawer";
 import type { KubeObjectDetailsProps } from "../kube-object-details";
-import { PriorityClass } from "../../../common/k8s-api/endpoints";
+import type { PriorityClass } from "../../../common/k8s-api/endpoints";
 import { KubeObjectMeta } from "../kube-object-meta";
-import logger from "../../../common/logger";
 
 export interface PriorityClassesDetailsProps extends KubeObjectDetailsProps<PriorityClass> {
 }
@@ -21,16 +20,6 @@ export class PriorityClassesDetails extends React.Component<PriorityClassesDetai
 
   render() {
     const { object: pc } = this.props;
-
-    if (!pc) {
-      return null;
-    }
-
-    if (!(pc instanceof PriorityClass)) {
-      logger.error("[PriorityClassesDetails]: passed object that is not an instanceof PriorityClass", pc);
-
-      return null;
-    }
 
     return (
       <div className="PriorityClassesDetails">
--- a/src/renderer/components/+config-priority-classes/priority-classes-details.tsx
+++ b/src/renderer/components/+config-priority-classes/priority-classes-details.tsx
@@ -1,0 +1,54 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import "./priority-classes.scss";
+
+import React from "react";
+import { observer } from "mobx-react";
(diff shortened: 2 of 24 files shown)


src/common/k8s-api/endpoints/priority-class.api.injectable.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import assert from "assert";
+import { storesAndApisCanBeCreatedInjectionToken } from "../stores-apis-can-be-created.token";
+import { PriorityClassApi } from "./priority-class.api";
+
+const PriorityClassApiInjectable = getInjectable({
```
Nokel81: nit: this should be camalCase

src/renderer/components/+config-priority-classes/legacy-store.ts
```diff
@@ -0,0 +1,12 @@
...
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import { asLegacyGlobalForExtensionApi } from "../../../extensions/as-legacy-globals-for-extension-api/as-legacy-global-object-for-extension-api";
+import priorityClassStoreInjectable from "./store.injectable";
+
+/**
+ * @deprecated use `di.inject(priorityClassStoreInjectable)` instead
+ */
+export const priorityClassStore = asLegacyGlobalForExtensionApi(priorityClassStoreInjectable);
```
Nokel81: nit: unless you have some uses of this I wouldn't bother adding it.

src/renderer/components/+config-priority-classes/priority-classes.tsx
```diff
@@ -0,0 +1,66 @@
...
+@observer
+export class PriorityClasses extends React.Component<PriorityClassesProps> {
+  render() {
+    return (
+      <SiblingsInTabLayout>
+        <KubeObjectListLayout
+          isConfigurable
+          tableId="configuration_priority_classes"
+          className="PriorityClasses"
+          store={priorityClassStore}
```
Nokel81, line 51: This entire component could probably be done using `withInjectables` so that this store doesn't have to use the legacy globals.

Nokel81 (commented):
Thanks very much for this PR, looks very good. Just some minor things before merging.

src/renderer/components/+config-priority-classes/legacy-store.ts
dex4er: I didn't found single component without legacy-store.ts so if you have another idea how to not use `store` attribute in `React.Component` then maybe you can change it.

src/renderer/components/+config-priority-classes/priority-classes.tsx
dex4er: Again: all other components don't use `withInjectables` then this change would be out of the scope for this small feature.

src/renderer/components/+config-priority-classes/legacy-store.ts
Nokel81: https://github.com/lensapp/lens/blob/master/src/renderer/components/%2Bworkloads-pods/pods.tsx is a great example.

dex4er: Ok I refactored

src/renderer/components/+config-priority-classes/priority-classes.tsx
dex4er: Ok I refactored

Nokel81 (dismissed):

src/renderer/components/+config-priority-classes/priority-classes.tsx
Nokel81: Thanks!

jansav:
Hey @dex4er, noticed that you are adding stuff to `kube-object-detail-registry` which was just removed here #5805.

It's easy conflict that you are facing now, just introduce new implementation of `kubeObjectDetailItemInjectionToken`. You can use this as an example: https://github.com/lensapp/lens/blob/78a6e8d1b3afebc8331dfffc981251f7d30d434d/src/renderer/components/kube-object-details/kube-object-detail-items/implementations/cluster-role-binding-detail-item.injectable.ts

src/renderer/components/+config-priority-classes/priority-classes-details.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+    if (!pc) {
+      return null;
+    }
+
+    if (!(pc instanceof PriorityClass)) {
+      logger.error("[PriorityClassesDetails]: passed object that is not an instanceof PriorityClass", pc);
+
+      return null;
+    }
```
jansav: These defensive code shouldn't be required anymore. `enabled` in the implementation of `kubeObjectDetailItemInjectionToken` should take care of these.

jansav (commented):
Small things here and there.

It's good to see how much work it still requires to add new type. This just tells me that we have still stuff we can improve to make this type of addition even more easy. :)

jim-docker:
Seeing this right after connecting to my local minikube cluster:

Screen Shot 2022-07-14 at 3 26 17 PM

(`make clean && make dev`)

dex4er:
Works for me with Minikube with k8s v1.21:

image

src/renderer/components/+config-priority-classes/priority-classes-details.tsx
dex4er: I removed this code.

jim-docker:
> Works for me with Minikube with k8s v1.21:
> 

issue is gone after pulling today's commits

Nokel81 (approved):

jansav (approved):

jansav:
@dex4er Thank you. 🎉

PR #5818, opened 2022-07-08
Fix open CommandDialog

### LINKED ISSUE #5411: Command palette hotkey does not work on Linux
**Describe the bug**
The command palette hotkey `Ctrl+Shift+P` does not appear to work on Linux. However, pressing `Ctrl+Shift+E` does work for extensions.

**To Reproduce**
1. Press `Ctrl+Shift+P` on Linux
2. No command palette is displayed

**Expected behavior**
Hotkeys should function on Linux

- Lens Version: 5.4.6
- OS: Linux (Arch)
- Installation method (e.g. snap or AppImage in Linux): AppImage (AUR: lens-bin)

**Logs:**
No logs produced on pressing key combo

**Kubeconfig:**
N/A

**Additional context**
Not sure if this is related or Linux-specific, but `Esc` does not work if a text field is selected (e.g. on entering the `Extensions` screen, the search field is auto-selected, and `Esc` refuses to close the page unless the field is unfocused).

Also unsure if there is meant to be a quit hotkey (tried `Ctrl+q`, `Ctrl+w`, `Ctrl+Shift+w` with no effect), or if it also just isn't working for Linux.

Nokel81:

fixes #5411

### DIFF
--- a/src/behaviours/command-pallet/keyboard-shortcuts.test.tsx
+++ b/src/behaviours/command-pallet/keyboard-shortcuts.test.tsx
@@ -1,0 +1,148 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import type { RenderResult } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import platformInjectable from "../../common/vars/platform.injectable";
+import { type ApplicationBuilder, getApplicationBuilder } from "../../renderer/components/test-utils/get-application-builder";
+
+describe("Command Pallet: keyboard shortcut tests", () => {
+  let applicationBuilder: ApplicationBuilder;
+  let rendered: RenderResult;
+
+  beforeEach(async () => {
+    applicationBuilder = getApplicationBuilder();
+  });
+
+  describe("when on macOS", () => {
+    beforeEach(async () => {
+      applicationBuilder.dis.rendererDi.override(platformInjectable, () => "darwin");
+      rendered = await applicationBuilder.render();
+    });
+
+    it("renders", () => {
+      expect(rendered.baseElement).toMatchSnapshot();
+    });
+
+    it("does not show the command pallet yet", () => {
+      const actual = rendered.queryByTestId("command-container");
+
+      expect(actual).toBeNull();
+    });
+
+    describe("when pressing ESC", () => {
+      beforeEach(() => {
+        userEvent.keyboard("{Escape}");
+      });
+
+      it("renders", () => {
+        expect(rendered.baseElement).toMatchSnapshot();
+      });
+
+      it("does not show the command pallet yet", () => {
+        const actual = rendered.queryByTestId("command-container");
+
+        expect(actual).toBeNull();
+      });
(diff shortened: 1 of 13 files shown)


src/behaviours/command-pallet/keyboard-shortcuts.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+        });
+
+        it("no longer shows the command pallet", () => {
+          const actual = rendered.queryByTestId("command-container");
+
+          expect(actual).toBeNull();
+        });
+      });
+    });
+  });
```
jim-docker, line 147: no Windows test?

src/renderer/components/command-palette/command-container.tsx
```diff
@@ -34,47 +35,38 @@ interface Dependencies {
...
     disposeOnUnmount(this, [
       this.props.legacyOnChannelListen(ipcChannel, action),
-      addWindowEventListener("keydown", this.onKeyboardShortcut(action)),
-      addWindowEventListener("keyup", this.escHandler, true),
+      addWindowEventListener("keydown", onKeyboardShortcut(
+        isMac
+          ? "Shift+Cmd+P"
+          : "Shift+Ctrl+P",
+        action,
+      )),
```
jansav, line 63: Future: I'd like to see keyboard-shortcuts as a feature. So that we don't re-invent the wheel in every place. There could be `keyboardShortcutInjectionToken` which you just implement.

It would look something like:

```typescript
interface KeyboardShortcut {
  id: string;
  keyStrokes: string[];
  action: () => void | Promise<void>;
  enabled: IComputedValue<boolean>;
}
```



By doing it like that, we get the possibility of having customizable keyboard shortcuts in the future if we so decide.

src/behaviours/command-pallet/keyboard-shortcuts.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import type { RenderResult } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import platformInjectable from "../../common/vars/platform.injectable";
+import { type ApplicationBuilder, getApplicationBuilder } from "../../renderer/components/test-utils/get-application-builder";
+
+describe("Command Pallet: keyboard shortcut tests", () => {
```
jansav, line 11: Good stuff here. This will be great tool if we ever decide to extract keyboard-shortcuts as a feature, because this test shouldn't break. Therefore it allows large-scope refactoring, which one of the goals of behavioural tests. :)

src/renderer/components/command-palette/command-container.tsx
jansav, line 63: All this mount-time stuff would be removed from here if we had the keyboard-shortcuts -feature. 

Just saying because normally I would say that this is logic that is not welcome here at all, but for now I think it's OK because we shouldn't introduce the keyboard-shortcuts feature here :)

jansav (approved):

jim-docker (approved):

PR #5824, opened 2022-07-11
Expose reactive ways to hide items in cluster frame using Extension API - PART 7

### LINKED ISSUE #5786: Extension API fails to respect `isEnabledForCluster`

jansav:
Deprecation of `isEnabledForCluster` which is not good enough. 

1. Extension is should always be enabled or disabled. It's not possible to disable extension for X-part of the application. Keep it simple.
2. Extension has full control whether some registration should be used or not. Previously decision could only be done once when extension was activating, but now there can be business logic tied to that decision.
3. We're already using similar stuff in OpenLens, so it makes sense to expose them to Extension API.

This is part of the #5800 

Fixes #5786

### DIFF
--- a/src/behaviours/cluster/kube-object-status-icon/extension-api/reactively-hide-kube-object-status.test.tsx
+++ b/src/behaviours/cluster/kube-object-status-icon/extension-api/reactively-hide-kube-object-status.test.tsx
@@ -1,0 +1,132 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import type { RenderResult } from "@testing-library/react";
+import type { ApplicationBuilder } from "../../../../renderer/components/test-utils/get-application-builder";
+import { getApplicationBuilder } from "../../../../renderer/components/test-utils/get-application-builder";
+import { getExtensionFakeFor } from "../../../../renderer/components/test-utils/get-extension-fake";
+import { getInjectable } from "@ogre-tools/injectable";
+import { frontEndRouteInjectionToken } from "../../../../common/front-end-routing/front-end-route-injection-token";
+import type { IObservableValue } from "mobx";
+import { observable, runInAction, computed } from "mobx";
+import React from "react";
+import { navigateToRouteInjectionToken } from "../../../../common/front-end-routing/navigate-to-route-injection-token";
+import { routeSpecificComponentInjectionToken } from "../../../../renderer/routes/route-specific-component-injection-token";
+import extensionShouldBeEnabledForClusterFrameInjectable from "../../../../renderer/extension-loader/extension-should-be-enabled-for-cluster-frame.injectable";
+import { KubeObject } from "../../../../common/k8s-api/kube-object";
+import { KubeObjectStatusLevel } from "../../../../common/k8s-api/kube-object-status";
+import { KubeObjectStatusIcon } from "../../../../renderer/components/kube-object-status-icon";
+
+describe("reactively hide kube object status", () => {
+  let builder: ApplicationBuilder;
+  let rendered: RenderResult;
+  let someObservable: IObservableValue<boolean>;
+
+  beforeEach(async () => {
+    builder = getApplicationBuilder();
+
+    const rendererDi = builder.dis.rendererDi;
+
+    rendererDi.unoverride(extensionShouldBeEnabledForClusterFrameInjectable);
+
+    rendererDi.register(testRouteInjectable, testRouteComponentInjectable);
+
+    builder.setEnvironmentToClusterFrame();
+
+    const getExtensionFake = getExtensionFakeFor(builder);
+
+    someObservable = observable.box(false);
+
+    const testExtension = getExtensionFake({
+      id: "test-extension-id",
+      name: "test-extension",
+
+      rendererOptions: {
+        kubeObjectStatusTexts: [
+          {
+            kind: "some-kind",
+            apiVersions: ["some-api-version"],
+
+            resolve: () => ({
+              level: KubeObjectStatusLevel.CRITICAL,
+              text: "some-kube-object-status-text",
+            }),
+
+            visible: computed(() => someObservable.get()),
+          },
+        ],
+      },
+    });
+
+    rendered = await builder.render();
+
+    const navigateToRoute = rendererDi.inject(navigateToRouteInjectionToken);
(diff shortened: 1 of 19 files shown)


src/behaviours/cluster/kube-object-status-icon/extension-api/reactively-hide-kube-object-status.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+  let builder: ApplicationBuilder;
+  let rendered: RenderResult;
+  let someObservable: IObservableValue<boolean>;
+
+  beforeEach(async () => {
+    builder = getApplicationBuilder();
+
+    const rendererDi = builder.dis.rendererDi;
+
+    rendererDi.unoverride(extensionShouldBeEnabledForClusterFrameInjectable);
```
jim-docker, line 31: why do you need to do this? (curious)

src/renderer/components/kube-object-details/kube-object-detail-items/kube-object-detail-item-registrator.injectable.ts
```diff
@@ -45,20 +43,28 @@ const kubeObjectDetailItemRegistratorInjectable = getInjectable({
...
+
+                if (!isRelevantKubeObject(kubeObject.get())) {
+                  return false;
+                }
+
+                if (!registration.visible) {
+                  return true;
+                }
+
+                return registration.visible.get();
```
jim-docker: nit, use
```
    return registration.visible ? registration.visible.get() : true;
```
like for the others?

src/behaviours/routes/extension-api/reactively-disable-pages.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+import type { RenderResult } from "@testing-library/react";
+import type { IObservableValue } from "mobx";
+import { observable, runInAction, computed } from "mobx";
+import React from "react";
+import type { TestExtensionRenderer } from "../../../renderer/components/test-utils/get-extension-fake";
+import type { ApplicationBuilder } from "../../../renderer/components/test-utils/get-application-builder";
+import { getApplicationBuilder } from "../../../renderer/components/test-utils/get-application-builder";
+import { getExtensionFakeFor } from "../../../renderer/components/test-utils/get-extension-fake";
+
+describe("reactively disable pages", () => {
```
jim-docker: "reactively disable global pages" is less ambiguous

src/behaviours/routes/extension-api/reactively-disable-pages.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+    expect(actual).not.toBeInTheDocument();
+  });
+
+  it("given page becomes enabled, when navigating to the page, shows the page", () => {
+    runInAction(() => {
+      someObservable.set(true);
+    });
+
+    rendererTestExtension.navigate();
```
jim-docker: shouldn't it work even if you navigate before the `runInAction()` above?

src/extensions/registries/page-registry.ts
```diff
@@ -14,6 +15,7 @@ export interface PageRegistration {
   id?: string;
   params?: PageParams<string | Omit<PageParamInit<any>, "name" | "prefix">>;
   components: PageComponents;
+  enabled?: IComputedValue<boolean>;
```
jim-docker, line 18: should we use `visible` to be consistent with the others? Especially if this is in the extension api.

jim-docker (dismissed):
Nice!

jim-docker:
not sure why this would have conflicts, sorry if I messed it up.

src/renderer/components/kube-object-details/kube-object-detail-items/kube-object-detail-item-registrator.injectable.ts
jansav: Changed.

src/behaviours/routes/extension-api/reactively-disable-pages.test.tsx
jansav: Changed.

src/behaviours/routes/extension-api/reactively-disable-pages.test.tsx
jansav: Actually no. Before `someObservable.set(true);` `navigate()` would try navigating to page but then get redirected to front page. It would work if we didn't have that redirection

src/behaviours/cluster/kube-object-status-icon/extension-api/reactively-hide-kube-object-status.test.tsx
jansav: `extensionShouldBeEnabledForClusterFrameInjectable` is globally overridden to always return `true`. Just so that you don't need to worry about it in 99% of behaviours since it's not interesting there.

src/extensions/registries/page-registry.ts
jansav: I almost did so, but then figured out that `visible` doesn't really work with pages.

jansav:
> not sure why this would have conflicts, sorry if I messed it up.

No problem. :)

jim-docker (approved):

src/extensions/registries/page-registry.ts
jim-docker: you mean it doesn't really work technically or conceptually?

msa0311 (approved):

PR #5841, opened 2022-07-18
Fix installation of helm charts

### LINKED ISSUE #5827: Installing Helm Chart doesn't work properly
**Describe the bug**

Issues:

1. Clicking "Install" button results to "Bad Request" notification.
2. Dock tab for doesn't load properly when you have already tried to install some chart (leave dock tab open for it) and try installing second chart.
3. Clicking "View Helm Release" after installing opens the Details for Release but then quickly closes it again. After that any details for releases cannot be opened.
4. In confirmation dialog for deleting release there is missing space between "Remove" and name of release. 
5. Changing version in dock tab for installing chart doesn't work.

- Lens Version: 5.6.0-alpha.6.20220706.1
- OS: OSX
- Installation method: Website + auto update

https://user-images.githubusercontent.com/3083083/178434974-6dead347-9277-4511-8fa0-f1a99cd7c9ad.mov

image

jansav:
Fixes #5827 

Notice the added behaviours, they describe completely how it should work. There's still lot to improve in the implementation, but that's enough for now.


src/renderer/components/+helm-charts/details/versions/helm-chart-details-version-selection.injectable.ts
```diff
@@ -0,0 +1,59 @@
...
+    const versionsOfSelectedHelmChart = di.inject(
+      versionsOfSelectedHelmChartInjectable,
+      chart,
+    );
+
+    const state = observable.box<HelmChart>();
+
+    return {
+      value: computed(
+        () => state.get() || versionsOfSelectedHelmChart.value.get()[0],
```
Nokel81, line 36: nit, using `??` is more semantically correct.

src/renderer/components/dock/editor-panel.tsx
```diff
@@ -59,7 +61,7 @@ const NonInjectedEditorPanel = observer(({
       autoFocus={autoFocus}
       id={tabId}
       value={value}
-      className={cssNames(styles.EditorPanel, className)}
+      className={cssNames(styles.EditorPanel, className, { hidden })}
```
Nokel81, line 64: This seems strange, why is this needed?

src/renderer/components/dock/install-chart/chart-data/call-for-helm-chart-values.injectable.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export type CallForHelmChartValues = (
+  repo: string,
+  name: string,
+  version: string
+) => Promise<string>;
+
+const callForHelmChartValuesInjectable = getInjectable({
+  id: "call-for-helm-chart-values",
+  instantiate: (): CallForHelmChartValues => getChartValues,
```
Nokel81, line 16: Shouldn't this also be `causesSideEffects`?

src/renderer/components/dock/install-chart/get-random-install-chart-tab-id.injectable.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import getRandomIdInjectable from "../../../../common/utils/get-random-id.injectable";
+
+const getRandomInstallChartTabIdInjectable = getInjectable({
+  id: "get-random-install-chart-tab-id",
+  instantiate: (di) => di.inject(getRandomIdInjectable),
+});
+
```
Nokel81, line 12: This is weird, why have this additional level of abstraction?

src/renderer/components/dock/install-chart/install-chart-model.injectable.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+  instantiate: async (di, tabId: string) => {
+    const store = di.inject(installChartTabStoreInjectable);
+    const callForHelmChartValues = di.inject(callForHelmChartValuesInjectable);
+    const callForHelmChartVersions = di.inject(callForHelmChartVersionsInjectable);
+    const callForCreateHelmRelease = di.inject(callForCreateHelmReleaseInjectable);
+    const dockStore = di.inject(dockStoreInjectable);
+    const navigateToHelmReleases = di.inject(navigateToHelmReleasesInjectable);
+    const closeTab = () => dockStore.closeTab(tabId);
+    const callChartByTab = async () => await waitUntilDefined(() => store.getData(tabId));
```
Nokel81: nit: this is a strange variable name for something that doesn't do an API call.

src/renderer/components/dock/install-chart/install-chart-model.injectable.tsx
```diff
@@ -0,0 +1,281 @@
...
+import type { SingleValue } from "react-select";
+import type { CallForHelmChartVersions } from "../../+helm-charts/details/versions/call-for-helm-chart-versions.injectable";
+import callForHelmChartVersionsInjectable from "../../+helm-charts/details/versions/call-for-helm-chart-versions.injectable";
+
+const installChartModelInjectable = getInjectable({
+  id: "install-chart-model",
+
+  instantiate: async (di, tabId: string) => {
+    const store = di.inject(installChartTabStoreInjectable);
+    const callForHelmChartValues = di.inject(callForHelmChartValuesInjectable);
```
Nokel81, line 35: comment: what is the benefit of naming these sort of injectables with the `callFor` prefix? Wouldn't `getHelmChartValues` be descriptive enough?

src/renderer/getDiForUnitTesting.tsx
```diff
@@ -111,6 +112,9 @@ export const getDiForUnitTesting = (opts: { doGeneralOverrides?: boolean } = {})
 
     di.override(historyInjectable, () => createMemoryHistory());
     di.override(legacyOnChannelListenInjectable, () => () => noop);
+
+    di.override(storageSaveDelayInjectable, () => 0);
```
Nokel81, line 116: Why is this override different than that of `main`'s `getDiForUnitTesting`?

jansav: This is used only by `createStorage` which is `renderer` thing.

src/renderer/components/dock/install-chart/get-random-install-chart-tab-id.injectable.ts
jansav: Overriding in test. If I would have overridden `getRandomIdInjectable` then all of use places (even the ones not relevant here) would be overridden.

src/renderer/components/dock/editor-panel.tsx
jansav: Because `MonacoEditor` works incorrectly

```typescript
const SomeComponent = observer(({ someProp, value }: Props) => (
  <div>
    {someProp.get() ? <MonacoEditor value={value.get()} /> : <div>Something else</div>}
  </div>
));

```

1. `someProp.get()` returns truthy value, `MonacoEditor` is rendered
2. `someProp.get()` changes to falsy value, `MonacoEditor` is not rendered
3. `value.get()` changes
4. `someProp.get()`  changes back to truthy value, `MonacoEditor` renders with old value

src/renderer/components/dock/install-chart/chart-data/call-for-helm-chart-values.injectable.ts
jansav: Sure, good catch.

jansav: Fixed

src/renderer/components/dock/install-chart/install-chart-model.injectable.tsx
jansav: Actually it was leftover. It's used only for waiting till chart is present in store so I named it `waitForChart` and made it `() => Promise<void>`

src/renderer/components/dock/install-chart/install-chart-model.injectable.tsx
jansav: It's just matter of preference and what we're used to. 

We've been using `callForX` for stuff that needs to be get from somewhere outside of the system. E.g. API call, file system, local storage. `getX` doesn't really communicate enough because it could also mean that we are getting some static value from somewhere or e.g. some deeply nested value from the input.

src/renderer/getDiForUnitTesting.tsx
Nokel81: Then why should `main` even know about it?

src/renderer/components/dock/editor-panel.tsx
Nokel81: Then shouldn't we fix `MonacoEditor`? That sounds like it isn't checking for an updated `value` prop (in `componentDidUpdate` or something).

jansav: Yes, we should. Draw the line here, because there is lot to improve in `MonacoEditor`. Issue is that it's calling `this.restoreViewState(this.model);` when it creates the editor. Didn't go any deeper than that.

src/renderer/getDiForUnitTesting.tsx
jansav: It doesn't?

Nokel81 (approved):

jim-docker:
https://user-images.githubusercontent.com/40840436/179756154-eb74dcb4-2325-4a7c-a7d3-3428355de3e7.mov

Or maybe this is related to the Upgrade bug

jansav:
> 
> 
> https://user-images.githubusercontent.com/40840436/179756154-eb74dcb4-2325-4a7c-a7d3-3428355de3e7.mov
> 
> 
> 
> Or maybe this is related to the Upgrade bug

Yes, this is related to the other issue I created today. #5852

jim-docker:
The release details page slides in a few extra times (sometimes it doesn't happen till ~30 seconds later)
Also, after removing a release, the release list gets redrawn a few extra times.

https://user-images.githubusercontent.com/40840436/179889331-fb8a374a-84f4-4494-9bb0-32fdb433de70.mov

jim-docker:
crash after removing a release with an upgrade tab open

https://user-images.githubusercontent.com/40840436/179889451-ec409597-387a-42b4-a254-2ba254b5955a.mov

jim-docker (approved):
approving in case found issues should be fixed in another PR

PR #5845, opened 2022-07-18
Add guards to prevent activation errors slipping through

Nokel81:

### DIFF
--- a/src/common/cluster/cluster.ts
+++ b/src/common/cluster/cluster.ts
@@ -340,15 +340,35 @@ export class Cluster implements ClusterModel, ClusterState {
     }
 
     if (this.disconnected || !this.accessible) {
-      await this.reconnect();
+      try {
+        this.broadcastConnectUpdate("Starting connection ...");
+        await this.reconnect();
+      } catch (error) {
+        this.broadcastConnectUpdate(`Failed to start connection: ${error}`, true);
+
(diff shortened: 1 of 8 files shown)


msa0311 (approved):

src/common/cluster/cluster.ts
```diff
@@ -340,15 +340,35 @@ export class Cluster implements ClusterModel, ClusterState {
...
     if (this.disconnected || !this.accessible) {
-      await this.reconnect();
+      try {
+        this.broadcastConnectUpdate("Starting connection ...");
+        await this.reconnect();
+      } catch (error) {
+        this.broadcastConnectUpdate(`Failed to start connection: ${error}`, true);
+
+        return;
+      }
```
jansav, line 350: We should do it in a way that doesn't require try-catches.

E.g. here it should be

```typescript

const { wasSuccessful } = await this.reconnect();

if (wasSuccessful) {
  // ...
} else {
  // ...
}

```

Adding try-catches is just invitation for this type of bugs. Try-catch should be in the first place that might throw. Here it might be some shell command or `spawn` w.e.  

We have ~230 try-catch blocks in our code base. By doing it in this way, we would have ~90% less. And by doing that, this type of bugs would be impossible.

jansav (approved):

PR #5850, opened 2022-07-19
Fix opening of release details

### LINKED ISSUE #5767: Namespace error when upgrading a helm chart
**Describe the bug**
I am trying to upgrade a helm chart "cert-manager" that I had added to the namespace "cert-manager" from the UI

I get the following error message

> the namespace from the provided object "kube-system" does not match the namespace "cert-manager". You must pass '--namespace=kube-system' to perform this operation.

**To Reproduce**
1. Go to Lens > Helm > Releases
2. Click on a release, the upgrade
3. Select a higher version

**Expected behavior**
The helm chart should be upgraded in its namespace

**Screenshots**
Screenshot 2022-07-01 at 12 01 15

- Lens Version: Lens: 5.5.4-latest.20220609.2
- OS: macOs 12.3.1 (21E258)
- Installation method (e.g. snap or AppImage in Linux): manual download

**Logs:**
```
N/A
```

**Kubeconfig:**
```
cyrilduchon-doris@Cyril-DD-MacbookPro staging % cat ~/.kube/config
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: XXX==
    server: https://XXX.yl4.eu-central-1.eks.amazonaws.com
  name: arn:aws:eks:eu-central-1:XXX:cluster/mjg-dev
- cluster:
    certificate-authority-data: XXX==
    server: https://XXX.gr7.eu-central-1.eks.amazonaws.com
  name: arn:aws:eks:eu-central-1:XXX:cluster/mjg-production
contexts:
- context:
    cluster: arn:aws:eks:eu-central-1:XXX:cluster/mjg-dev
    user: arn:aws:eks:eu-central-1:XXX:cluster/mjg-dev
  name: arn:aws:eks:eu-central-1:XXX:cluster/mjg-dev
- context:
    cluster: arn:aws:eks:eu-central-1:113227347315:cluster/mjg-production
    user: arn:aws:eks:eu-central-1:113227347315:cluster/mjg-production
  name: arn:aws:eks:eu-central-1:113227347315:cluster/mjg-production
current-context: arn:aws:eks:eu-central-1:113227347315:cluster/mjg-dev
kind: Config
preferences: {}
users:
- name: arn:aws:eks:eu-central-1:xxx:cluster/mjg-dev
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1alpha1
      args:
      - --region
      - eu-central-1
      - eks
      - get-token
      - --cluster-name
      - mjg-dev
      command: aws
      env: null
      interactiveMode: IfAvailable
      provideClusterInfo: false
- name: arn:aws:eks:eu-central-1:xxx:cluster/mjg-production
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1alpha1
      args:
      - --region
      - eu-central-1
      - eks
      - get-token
      - --cluster-name
      - mjg-production
      command: aws
      env: null
      interactiveMode: IfAvailable
      provideClusterInfo: false
```

**Additional context**

### LINKED ISSUE #5849: Unrelated resources shown in release details
**Describe the bug**
When I open details for a Helm Release, there is list of resources which contain clearly unrelated resources.

**To Reproduce**
1. Install two helm charts
2. Open details for helm release
3. Scroll down to see list of resources
5. See that it lists unrelated resources 

**Expected behavior**
Only resources created by release should be listed

**Screenshots**

https://user-images.githubusercontent.com/3083083/179717318-1e4505b7-7214-4577-84aa-c9b0a5fc7dcb.mov

- Lens Version: 5.6.0-alpha.6.2
- OS: OSX

jansav:
Fixes #5767
Fixes #5849


jansav:
Working on adding behaviour to cover this

src/main/helm/helm-release-manager.ts
```diff
@@ -191,7 +191,6 @@ async function getResources(name: string, namespace: string, kubeconfigPath: str
   ];
   const kubectlArgs = [
     "get",
-    "--namespace", namespace,
```
Nokel81, line 194: Why would removing the namespace mean that the "correct" resources are retrieved?

jansav: This part fixes the crash.  https://github.com/lensapp/lens/issues/5767

jansav: Other changes fixes https://github.com/lensapp/lens/issues/5849.

But more changes are coming when I get the behaviours in because I can already see two blindspots which lead to crash in correct circumstance.

src/renderer/components/+helm-releases/release-details/release-details.tsx
```diff
@@ -70,7 +71,11 @@ class NonInjectedReleaseDetails extends Component<ReleaseDetailsProps & Dependen
   }
 
   @computed get details() {
-    return this.props.releaseDetails.value.get();
+    const details = this.props.releaseDetails.value.get();
+
+    assert(details);
+
+    return details;
   }
```
Nokel81: Given that this is only used in the `renderNotes` function below and that is only used once in `renderContent`, I think we probably could just drop this computed and just pipe the value through.

jansav: Resolving this because I implemented details to remove many bugs and blindspots.

jansav:
@Nokel81 @jim-docker This is now ready for review. It's fully behavioural tested, so please review that behaviour is correct.

src/main/getDiForUnitTesting.ts
```diff
@@ -124,6 +124,16 @@ export function getDiForUnitTesting(opts: { doGeneralOverrides?: boolean } = {})
...
+    const globalOverrideFilePaths = getGlobalOverridePaths();
+
+    const globalOverrides = globalOverrideFilePaths.map(
+      (filePath) => require(filePath).default,
+    );
+
+    globalOverrides.forEach(globalOverride => {
+      di.override(globalOverride.injectable, globalOverride.overridingInstantiate);
+    });
+
```
jansav, line 136: Note: This is here so that we can stop bloating this file. We can now just add file with postfix `.global-override-for-injectable.ts` which is used for override automatically.

src/renderer/components/+helm-releases/call-for-helm-releases/call-for-helm-releases.global-override-for-injectable.ts
```diff
@@ -0,0 +1,15 @@
...
+import { getGlobalOverride } from "../../../../common/test-utils/get-global-override";
+
+export default getGlobalOverride(
+  callForHelmReleasesInjectable,
+  () => () => {
+    throw new Error(
+      "Tried to call for helm releases without explicit override.",
+    );
+  },
+);
```
jansav, line 15: Note: First example of global override.

src/main/getDiForUnitTesting.ts
jansav: In future PR, we can consolidate all overrides here to new way.

src/renderer/components/+helm-releases/release-details/release-details-content.tsx
```diff
@@ -0,0 +1,210 @@
...
+      testIdForClose="close-helm-release-detail"
+      toolbar={
+        !isLoading && (
+          <HelmReleaseMenu
+            release={model.release}
+            toolbar
+            hideDetails={model.close}
+          />
+        )
+      }
```
jansav, line 56: Notice: Here the pain is that we have to set `title` and `toolbar` even though we don't have the data yet. Drawer needs to be refactored to place where we can set `title` and `toolbar` later.

It would look something like this:

```jsx
return (
  <Drawer open={true}>
    <DrawerHeader>
      <span>{model.release.getName()}</span>
     
      <HelmReleaseMenu toolbar release={model.release} />
    </DrawerHeader>

    <DrawerContent>
      Some content
    </DrawerContent>
  </Drawer>
)
```

In that case we could:

```jsx
return (
  <Drawer open={true}>
    <ReleaseDetailsContent />
  </Drawer>
);
```

And now `ReleaseDetailsContent` could have `async` dependency and `getPlaceholder` which returns spinner that looks good for drawer.

---

**Key thing is to start combining different components and stop expanding the scope of a single component by adding more props. That gives us flexibility, when latter only limits our ability to change our mind later.**

src/renderer/components/+helm-releases/release-details/release-details-model/release-details-model.injectable.tsx
```diff
@@ -0,0 +1,312 @@
...
+  callForHelmReleaseConfiguration: CallForHelmReleaseConfiguration;
+  getResourceDetailsUrl: GetResourceDetailsUrl;
+  updateRelease: CallForHelmReleaseUpdate;
+  showCheckedErrorNotification: ShowCheckedErrorNotification;
+  showSuccessNotification: ShowNotification;
+  createUpgradeChartTab: (release: HelmRelease) => string;
+  navigateToHelmReleases: NavigateToHelmReleases;
+}
+
+export class ReleaseDetailsModel {
```
jansav, line 105: Note: I'd call this view model since it is the only dependency that `ReleaseDetails` has and it combines all of the functionality and business logic details needs. It doesn't expose anything else than is required and it's allowed to be injected only by `ReleaseDetails`. Therefore it serves only one purpose.

src/renderer/components/dock/upgrade-chart/view.tsx
```diff
@@ -96,7 +98,7 @@ export class NonInjectedUpgradeChart extends React.Component<UpgradeChartProps &
     const versions = await helmChartStore.getVersions(release.getChart());
 
     this.versions.replace(versions);
-    this.version = this.versions[0];
+    this.version = first(this.versions);
```
jansav, line 101: Note: had to do this since opening tab for the purpose of upgrading chart shows warning without the data. There is still open issue for the crashes in upgrade #5852.

src/main/getDiForUnitTesting.ts
Nokel81: Is this worth it though since the end game would be to have as few of these global overrides as possible?

src/renderer/components/+helm-releases/release-details/release-details-model/release-details-model.injectable.tsx
```diff
@@ -0,0 +1,312 @@
...
+    // TODO: Reorganize Drawer to allow setting of header-bar in children to make "getPlaceholder" from injectable usable.
+    load();
+
+    return model;
+  },
+
+  lifecycle: lifecycleEnum.keyedSingleton({
+    getInstanceKey: (di, release: TargetHelmRelease) =>
+      `${release?.namespace}/${release?.name}`,
+  }),
```
Nokel81: This seems weird, why is it `release?.namespace` and not `release.namespace`? If the reason is that `release` could be `undefined | null` then why is the function not declaring that?

src/renderer/components/+helm-releases/release-details/release-details-model/release-details-model.injectable.tsx
```diff
@@ -0,0 +1,312 @@
...
+    this.dependencies.createUpgradeChartTab(this.release);
+
+    this.close();
+  };
+
+  @computed get id() {
+    const { name, namespace } = this.dependencies.targetRelease;
+
+    return `${namespace}/${name}`;
+  }
```
Nokel81: nit: this seems strange, why is not just assigned once at construction time?

src/renderer/components/dock/upgrade-chart/view.tsx
Nokel81: What is the difference then? Or do you mean a mobx warning in the console?

src/behaviours/helm-releases/showing-details-for-helm-release.test.ts
```diff
@@ -0,0 +1,604 @@
...
+
+      rendererDi.override(
+        callForHelmReleasesInjectable,
+        () => callForHelmReleasesMock,
+      );
+
+      rendererDi.override(
+        callForHelmReleaseInjectable,
+        () => callForHelmReleaseMock,
+      );
```
Nokel81, line 80: why the duplicate calls to override?

Nokel81 (commented):
The behaviours look good

src/renderer/components/+helm-releases/release-details/release-details-model/release-details-model.injectable.tsx
jansav: Good catch. Removed the optional chaining.

src/renderer/components/+helm-releases/release-details/release-details-model/release-details-model.injectable.tsx
jansav: Sure, changed.

src/behaviours/helm-releases/showing-details-for-helm-release.test.ts
jansav: `releases` vs `release` :)

jansav:
> What is the difference then? Or do you mean a mobx warning in the console?

No difference in behaviour, but gets rid of MobX warning.

jansav:
> Is this worth it though since the end game would be to have as few of these global overrides as possible?

It is, because I don't see how we could get rid of global overrides before this file gets out of control. In the end, it's very easy to remove if we don't need it. Even though we wouldn't refactor existing ones, adding new is a lot easier because now global override can live right next to the actual implementation.

Nokel81 (approved):

src/renderer/components/+helm-releases/release-details/release-details-content.tsx
aleksfront: Totally agree, let's do this.

aleksfront (approved):

PR #5855, opened 2022-07-19
Fix kubeconfig-sync sometimes producing multiple identical entities

### LINKED ISSUE #3872: No longer seeing a blue 'LED' on connected clusters in the Hotbar

Nokel81:
- This happens when there is an overlapping sync of a folder and a file
  within that folder both producing the same entities

- This also fixes the LED not showing up on some KubernetesCluster
  instances because of this duplication the connected status was being
  set on the first instance, but renderer would use the second (because
  JS's Map constructor deduplicates its constructor initializer by
  taking the last)

- Added some tests to cover this case


jim-docker:
This is a big refactor for the 5.6.0 milestone...

Nokel81:
We can push this to the next milestone

aleksfront (commented):
I can't reproduce the issue mentioned in the description:

> This happens when there is an overlapping sync of a folder and a file
within that folder both producing the same entities

Tested with 2 different file/folder structures:
```
configfolder -> config
             -> deepfolder -> config
``` 
and
```
configfolder -> config
             -> config 2
```

all of the config files contained same minikube cluster (copy pasted kubeconfig from one file to another)

In my opinion Lens seems work as expected treating these clusters as separate ones. Both of clusters switched okay with proper LED lights without any sync problem. Example:


https://user-images.githubusercontent.com/9607060/187372472-2bb205e1-306b-4060-be97-4bfcf21d7e92.mov

src/common/logger/child-logger.injectable.ts
```diff
@@ -0,0 +1,37 @@
...
+import { getInjectable, lifecycleEnum } from "@ogre-tools/injectable";
+import type { Logger } from "../logger";
+import loggerInjectable from "../logger.injectable";
+
+const childLoggerInjectable = getInjectable({
+  id: "child-logger",
+  instantiate: (di, prefix): Logger => {
+    const logger = di.inject(loggerInjectable);
+
+    return {
```
Iku-turso: Note: it's a questionable practise here and elsewhere to bundle all these responsibilities together.

src/common/logger/child-logger.injectable.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable, lifecycleEnum } from "@ogre-tools/injectable";
+import type { Logger } from "../logger";
+import loggerInjectable from "../logger.injectable";
+
+const childLoggerInjectable = getInjectable({
+  id: "child-logger",
```
Iku-turso: Proposed name: "prefixed-logger" :)

src/main/catalog-sources/__test__/kubeconfig-sync.test.ts
```diff
@@ -151,7 +164,7 @@ describe("kubeconfig-sync.source tests", () => {
 
       fs.writeFileSync(filePath, contents);
```
Iku-turso: Opportunity to get rid of another `mock-fs` ;) Fix if feeling boy-scoutish :)

While we are at it, maybe these technical unit tests could be more awesome as part of behavioural tests instead. WDYT? :)

src/common/fs/watch.injectable.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { watch } from "chokidar";
+
+const watchInjectable = getInjectable({
```
Iku-turso: Nit: this foreign 3rd party abstraction is accessed directly from our domestic code, hard-coupling us to chokidar exactly. We could just as well invent our own abstraction that would be simple, stable and controlled in single place.

This may not seem super meaningful here, but as a policy, not letting foreign abstractions bleed in domestic code can sure be a life-saver, and a good investment on top of that.

src/main/start-main-application/runnables/kube-config-sync/add-source.injectable.ts
```diff
@@ -0,0 +1,25 @@
+/**
```
Iku-turso, line 1: As modern code like this emerges, it would be great if it did so accompanied by those darn behavioural tests ;) WDYT?

Iku-turso (changes requested):
Discuss: Add or migrate to behavioural tests? Are you too starting to have the feeling that PR's that don't pull their weight in contributions to unit testing don't kinda feel up-to-par anymore? No? :)

Iku-turso (dismissed):

src/common/logger/child-logger.injectable.ts
Iku-turso: Discussion handled IRL.

src/main/catalog-sources/__test__/kubeconfig-sync.test.ts
Iku-turso: Discussion handled IRL.

src/common/fs/watch.injectable.ts
Iku-turso: Discussion handled IRL.

src/main/start-main-application/runnables/kube-config-sync/add-source.injectable.ts
Iku-turso: Discussion handled IRL.

Iku-turso (dismissed):

PR #5870, opened 2022-07-21
Fix terminal freezing self and whole app when internet is disconnected

### LINKED ISSUE #5868: Opening terminal without network connection makes application freeze
**Describe the bug**

When you try to open terminal with network being disabled, the whole application just freezes. Only way to shut it down is to force quit.

**To Reproduce**
1. Open any cluster
2. Turn off networking (e.g. WIFI)
3. Open new terminal tab
4. See the freeze

**Expected behavior**
Error that explains you need network to access terminal (if that's really the case? e.g. local clusters?).

- Lens Version: 5.6.0-alpha.7.1
- OS: OSX

Nokel81:
- Remove uses to navigator.onLine as not needed and because that value
  cannot be trusted (see several bug reports on chromium)

- Fix infinite loop when socket is not connected in flush

- Add some simple tests

### DIFF
--- a/src/renderer/api/__tests__/websocket-api.test.ts
+++ b/src/renderer/api/__tests__/websocket-api.test.ts
@@ -1,0 +1,37 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import type { WebSocketEvents } from "../websocket-api";
+import { WebSocketApi } from "../websocket-api";
+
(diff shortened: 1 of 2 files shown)


jansav (approved):
Seems to fix the issue. 

It seems that only usage of `WebSocketApi` is `TerminalApi` so this works. However, if we had another usage of `WebSocketApi` which require networking, I don't see how this would work. 

Let's figure that out in the future.

Nokel81:
@jansav I would have to play around with websockets some more but from my research using `navigator.onLine` is never the correct solution.

aleksfront (approved):
LGTM

PR #5875, opened 2022-07-22
Resolve kube object status texts again on re-render to make it look like it's reactive

jansav:
~~This verifies that kube object status texts (the warning icons e.g. next to pod name in Pods -list) in general are working as expected.~~

~~However, this doesn't verify that e.g. warnings for pods are shown when they are supposed to.~~

~~@msa0311 previously encountered a situation where status warning for Pod would appear, but not disappear after issue was cleared.~~


jansav:
Converting back to draft since it came clear that current understanding how it should work is wrong. Extensions are relying `resolve` to be called on each render and component gets re-rendered because some other change causes parent component to change. 

This means that `kube-object-status-icon` is not usable as a standalone component because there is no way that it could react to changes.

jansav:
Ready to be reviewed

src/renderer/components/kube-object-status-icon/kube-object-status-icon.tsx
```diff
@@ -133,6 +139,10 @@ class NonInjectedKubeObjectStatusIcon extends React.Component<KubeObjectStatusIc
   }
 }
 
+function isNotEmpty<T>(item: T | null | undefined): item is T {
+  return !!item;
+}
```
Nokel81: nit: this already exists in the form of `isDefined` in `src/common/utils/type-narrowing.ts`. Also this function is not quite correct since `isNotEmpty(0)` would return `false`

Nokel81 (dismissed):

Nokel81:
> This means that `kube-object-status-icon` is not usable as a standalone component because there is no way that it could react to changes.

This is not quite true because an `observer` component tracks all changes to `observable` or `computed` state through all function calls (components excluded).

src/renderer/components/kube-object-status-icon/kube-object-status-icon.tsx
jansav: Removed new implementation to prefer existing.

jansav:
> > This means that `kube-object-status-icon` is not usable as a standalone component because there is no way that it could react to changes.
> 
> This is not quite true because an `observer` component tracks all changes to `observable` or `computed` state through all function calls (components excluded).

Sure. You can always find a way to hack around it, but it shouldn't be done. If we want to have something which works as standalone component, then it would make more sense to present reactive way to control visibility.

Nokel81 (approved):

src/behaviours/cluster/kube-object-status-icon/show-status-for-a-kube-object.test.tsx
```diff
@@ -0,0 +1,376 @@
...
+    let rendererDi: DiContainer;
+    let rendered: RenderResult;
+    let rerenderParent: () => void;
+
+    beforeEach(async () => {
+      rendered = await builder.render();
+
+      rendererDi = builder.dis.rendererDi;
+
+      const someAtom = rendererDi.inject(someAtomInjectable);
```
aleksfront, line 145: Would be good to know better what atom is and why it it needed in tests.

aleksfront (approved):

src/behaviours/cluster/kube-object-status-icon/show-status-for-a-kube-object.test.tsx
jansav: `Atom` is concept of `mobx`. Documentation here: https://mobx.js.org/custom-observables.html
Basically it's something like `observable` or `computed` but without any value. There's just a way to say:
- `atom.reportObserved()` which means that you are interested of it, e.g. component re-rendering is controlled by it (like when you call `someObservable.get()` in component.)
- `atom.reportChanged()` which means that every place with `reportObserved()` should be recomputed (like when you call `someObservable.set("some-value")`)

In this test I'm using it to control when parent component should be re-rendered. 
I could do the same with `observable.box` which I would dereference (`.get()`) in the component and `.set("some-value")` when I want to re-render. Why Atom? - because it doesn't require the value which changes.

PR #5877, opened 2022-07-22
Kludge editing of kube resources to work again

### LINKED ISSUE #5876: Editing kube resource doesn't work.
**Describe the bug**

When trying to edit any kube resource, the resource is not updated at all. 

**To Reproduce**
1. Open any cluster
2. Go to e.g. Pods (applies to any kube resource)
3. Open Pod to edit
4. Change some value
5. Click Save & Close (or Save -> refresh)
6. Open Edit again for same Pod
7. See that changed values are not there

**Screenshots**

https://user-images.githubusercontent.com/3083083/180416316-f8b54dff-76b6-44b0-b79c-4e73efe8859d.mov

- Lens Version: 5.6.0-alpha.7.1
- OS: OSX

This is regression of #5195

jansav:
Kludge because this is way too complicated and cannot be unit tested at the moment. 

Fixes #5876

### DIFF
--- a/src/renderer/components/dock/edit-resource/view.tsx
+++ b/src/renderer/components/dock/edit-resource/view.tsx
@@ -4,11 +4,11 @@
  */
 
 import React from "react";
-import { autorun, makeObservable, observable } from "mobx";
+import { autorun, computed, makeObservable, observable } from "mobx";
 import { disposeOnUnmount, observer } from "mobx-react";
 import yaml from "js-yaml";
 import type { DockTab, TabId } from "../dock/store";
-import type { EditingResource, EditResourceTabStore } from "./store";
+import type { EditResourceTabStore } from "./store";
 import { InfoPanel } from "../info-panel";
 import { Badge } from "../../badge";
 import { EditorPanel } from "../editor-panel";
@@ -31,7 +31,6 @@ interface Dependencies {
 }
 
 interface SaveDraftArgs {
-  tabData: EditingResource;
   resource: KubeObject;
   store: KubeObjectStore;
 }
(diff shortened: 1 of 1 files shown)


src/renderer/components/dock/edit-resource/view.tsx
```diff
@@ -101,14 +100,15 @@ class NonInjectedEditResource extends React.Component<EditResourceProps & Depend
...
     if (this.error) {
       return null;
     }
 
-    const currentVersion = yaml.load(this.draft);
-    const firstVersion = yaml.load(tabData.firstDraft ?? this.draft);
-    const patches = createPatch(firstVersion, currentVersion);
+    const currentVersion = yaml.load(this.configurationWhenOpened);
+    const firstVersion = yaml.load(this.configurationInView ?? this.configurationWhenOpened);
+    const patches = createPatch(currentVersion, firstVersion);
```
jansav, line 110: Important change here. It seems that #5195 changed the meaning of these variables, which mean that the order was changed.

src/renderer/components/dock/edit-resource/view.tsx
```diff
@@ -123,8 +123,12 @@ class NonInjectedEditResource extends React.Component<EditResourceProps & Depend
     );
   }
 
+  @computed get configurationInView() {
+    return this.props.editResourceStore.getData(this.tabId)?.draft;
+  }
+
```
jansav, line 129: This computed is crucial. `this.tabData` or `tabData` from `save()` are not correct. Again, this is kludge because that's simply not OK..

Nokel81: With testing (and reading code) your claim about `computed` is incorrect. Since this value is only read in one place (essentially an action) it does not need to be computed.

jansav: `@computed` maybe, since `getData` already reads from reactive sure. So maybe I should have said "This getter is crucial", since `get tabData()` cannot be used.

aleksfront (approved):

PR #5881, opened 2022-07-22
Fix trying to connect to cluster whose entity doesn't exist

Nokel81:
- If a Cluster is trying to be connected to but there is no associated
  entity then don't try and connect

### DIFF
--- a/src/renderer/components/cluster-manager/cluster-view.tsx
+++ b/src/renderer/components/cluster-manager/cluster-view.tsx
@@ -68,6 +68,11 @@ class NonInjectedClusterView extends React.Component<Dependencies> {
   bindEvents() {
     disposeOnUnmount(this, [
       reaction(() => this.clusterId, async (clusterId) => {
+        // TODO: replace with better handling
+        if (clusterId && !this.props.entityRegistry.getById(clusterId)) {
+          return this.props.navigateToCatalog(); // redirect to catalog when the clusterId does not correspond to an entity
+        }
+
         this.props.clusterFrames.setVisibleCluster(clusterId);
         this.props.clusterFrames.initView(clusterId);
         requestClusterActivation(clusterId, false); // activate and fetch cluster's state from main
--- a/src/renderer/api/catalog/entity/registry.ts
+++ b/src/renderer/api/catalog/entity/registry.ts
@@ -60,7 +60,7 @@ export class CatalogEntityRegistry {
     // If the entity was not found but there are rawEntities to be processed,
     // try to process them and return the entity.
     // This might happen if an extension registered a new Catalog category.
-    if (this.activeEntityId.get() && !entity && this.rawEntities.length > 0) {
+    if (this.activeEntityId && !entity && this.rawEntities.length > 0) {
       this.processRawEntities();
 
       return this.getActiveEntityById();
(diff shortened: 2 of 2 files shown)


src/renderer/components/cluster-manager/cluster-view.tsx
```diff
@@ -68,6 +68,11 @@ class NonInjectedClusterView extends React.Component<Dependencies> {
   bindEvents() {
     disposeOnUnmount(this, [
       reaction(() => this.clusterId, async (clusterId) => {
+        // TODO: replace with better handling
+        if (clusterId && !this.props.entityRegistry.getById(clusterId)) {
+          return this.props.navigateToCatalog(); // redirect to catalog when the clusterId does not correspond to an entity
+        }
+
```
Nokel81, line 75: This code is safe because it is run in the main frame so there should be no timing issues with the catalog not being synced yet.

src/renderer/components/cluster-manager/cluster-view.tsx
```diff
@@ -68,6 +68,11 @@ class NonInjectedClusterView extends React.Component<Dependencies> {
   bindEvents() {
     disposeOnUnmount(this, [
       reaction(() => this.clusterId, async (clusterId) => {
+        // TODO: replace with better handling
+        if (clusterId && !this.props.entityRegistry.getById(clusterId)) {
```
jim-docker, line 72:
```suggestion
        if (!clusterId || !this.props.entityRegistry.getById(clusterId)) {
```
You don't want it activating if `clusterId` is undefined, right? Although I guess that is the pre-existing behaviour...

Nokel81: I think we do because that clears the current cluster value on main

aleksfront (commented):
It would be good to have some bit more description: when it occasionally happens (cluster without an entity) and for what reason? Don't we have app crash in this case but only reconnection attempts?

src/renderer/components/cluster-manager/cluster-view.tsx
```diff
@@ -68,6 +68,11 @@ class NonInjectedClusterView extends React.Component<Dependencies> {
...
+          return this.props.navigateToCatalog(); // redirect to catalog when the clusterId does not correspond to an entity
```
panuhorsmalahti, line 73: So when this bug happens user is redirected to the Catalog page, and then has to try to connect again, hoping it will work?

Nokel81: So the bug is that the file for the cluster did exist so `main` has an instance of `Cluster` for it, but the file has been removed so when the `iframe` is created (which happens before the activation of the cluster succeeds) that file isn't there and thus throws an error in the constructor.

The other options for this is to change the constructor of the cluster so that the data from the file is not a requirement but I would want more tests before doing that.

Nokel81:
> It would be good to have some bit more description: when it occasionally happens (cluster without an entity) and for what reason? Don't we have app crash in this case but only reconnection attempts?

Sure

1. when it occasionally happens (cluster without an entity) and for what reason? 
    - If an entity no longer exists then I figured it would be best to disallow trying to connect to a cluster. Since the only way to connect to the cluster at that point is navigation via the history stack (the back arrows etc.)
2. Don't we have app crash in this case but only reconnection attempts?
    - In neither case do we have an app crash. On dev mode there I `iframe` goes into an unhanded promise rejection handler and displays the error where as in a prod build the error is only logged to the console.

jim-docker (approved):
can't seem to get the extension to show up in order to test

msa0311 (approved):

PR #5906, opened 2022-07-26
Fix editing of kube resource

### LINKED ISSUE #5879: Kube editor doesn't refresh contents when jumping between tabs
When jumping between `Edit resource` tabs some of then have exact same content which means Monaco Editor doesn't refresh tab data properly.

https://user-images.githubusercontent.com/9607060/180454118-a66a9f9f-4a68-46bc-8202-6bdb0525187f.mov

jansav:
Fixes #5879

### DIFF
--- a/src/behaviours/cluster/namespaces/edit-namespace-from-new-tab.test.tsx
+++ b/src/behaviours/cluster/namespaces/edit-namespace-from-new-tab.test.tsx
@@ -1,0 +1,992 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import type { DiContainer } from "@ogre-tools/injectable";
+import type { RenderResult } from "@testing-library/react";
+import { fireEvent } from "@testing-library/react";
+import type { ApplicationBuilder } from "../../../renderer/components/test-utils/get-application-builder";
+import { getApplicationBuilder } from "../../../renderer/components/test-utils/get-application-builder";
+import navigateToNamespacesInjectable from "../../../common/front-end-routing/routes/cluster/namespaces/navigate-to-namespaces.injectable";
+import React from "react";
+import createEditResourceTabInjectable from "../../../renderer/components/dock/edit-resource/edit-resource-tab.injectable";
+import { KubeObject } from "../../../common/k8s-api/kube-object";
+import getRandomIdForEditResourceTabInjectable from "../../../renderer/components/dock/edit-resource/get-random-id-for-edit-resource-tab.injectable";
+import type { AsyncFnMock } from "@async-fn/jest";
+import asyncFn from "@async-fn/jest";
+import type { CallForResource } from "../../../renderer/components/dock/edit-resource/edit-resource-model/call-for-resource/call-for-resource.injectable";
+import callForResourceInjectable from "../../../renderer/components/dock/edit-resource/edit-resource-model/call-for-resource/call-for-resource.injectable";
+import type { CallForPatchResource } from "../../../renderer/components/dock/edit-resource/edit-resource-model/call-for-patch-resource/call-for-patch-resource.injectable";
+import callForPatchResourceInjectable from "../../../renderer/components/dock/edit-resource/edit-resource-model/call-for-patch-resource/call-for-patch-resource.injectable";
+import dockStoreInjectable from "../../../renderer/components/dock/dock/store.injectable";
+import { Namespace } from "../../../common/k8s-api/endpoints";
+import showSuccessNotificationInjectable from "../../../renderer/components/notifications/show-success-notification.injectable";
+import showErrorNotificationInjectable from "../../../renderer/components/notifications/show-error-notification.injectable";
+import readJsonFileInjectable from "../../../common/fs/read-json-file.injectable";
+import directoryForLensLocalStorageInjectable from "../../../common/directory-for-lens-local-storage/directory-for-lens-local-storage.injectable";
+import hostedClusterIdInjectable from "../../../renderer/cluster-frame-context/hosted-cluster-id.injectable";
+import { controlWhenStoragesAreReady } from "../../../renderer/utils/create-storage/storages-are-ready";
+
+jest.mock("../../../renderer/components/tooltip/withTooltip", () => ({
+  withTooltip:
+    (Target: any) =>
+      ({ tooltip, ...props }: any) => {
+        if (tooltip) {
+          const testId = props["data-testid"];
+
+          return (
+            <>
+              <Target
+                tooltip={tooltip.children ? undefined : tooltip}
+                {...props}
+              />
+              <div data-testid={testId && `tooltip-content-for-${testId}`}>
+                {tooltip.children || tooltip}
+              </div>
+            </>
+          );
+        }
+
+        return <Target {...props} />;
+      },
+}));
+
+describe("cluster/namespaces - edit namespace from new tab", () => {
+  let builder: ApplicationBuilder;
+  let callForNamespaceMock: AsyncFnMock<CallForResource>;
+  let callForPatchNamespaceMock: AsyncFnMock<CallForPatchResource>;
+  let showSuccessNotificationMock: jest.Mock;
+  let showErrorNotificationMock: jest.Mock;
+  let storagesAreReady: () => Promise<void>;
+
+  beforeEach(() => {
+    builder = getApplicationBuilder();
+
+    builder.setEnvironmentToClusterFrame();
+
+    callForNamespaceMock = asyncFn();
+    callForPatchNamespaceMock = asyncFn();
+
+    showSuccessNotificationMock = jest.fn();
+    showErrorNotificationMock = jest.fn();
+
+    builder.beforeApplicationStart(({ rendererDi }) => {
+      rendererDi.override(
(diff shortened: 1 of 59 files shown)


src/behaviours/cluster/namespaces/edit-namespace-from-new-tab.test.tsx
```diff
@@ -0,0 +1,992 @@
...
+        });
+
+        xit("does not show edit resource tab yet", () => {
+
+        });
+
+        describe("when clicking to edit namespace", () => {
+          beforeEach(() => {
+            // TODO: Make implementation match the description (tests above)
+            const namespaceStub = KubeObject.create(someNamespaceDataStub);
```
Nokel81: Why are we not using `new Namespace`?

src/behaviours/cluster/namespaces/edit-namespace-from-new-tab.test.tsx
```diff
@@ -0,0 +1,992 @@
...
+              expect(input.value).toBe(`apiVersion: some-api-version
+kind: Namespace
+metadata:
+  uid: some-uid
+  name: some-name
+  resourceVersion: some-resource-version
+  selfLink: /apis/some-api-version/namespaces/some-uid
+  somePropertyToBeRemoved: some-value
+  somePropertyToBeChanged: some-old-value
+`);
```
Nokel81, line 271: Wouldn't this be better as a snapshot?

src/common/k8s-api/kube-api/get-kube-api-from-path.injectable.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  instantiate: (di) => {
+    const kubeApis = di.injectMany(kubeApiInjectionToken);
+
+    return (apiPath: string) => {
+      const parsed = parseKubeApi(apiPath);
+
+      const kubeApi = kubeApis.find((api) => api.apiBase === parsed.apiBase);
+
+      return (kubeApi as KubeApi) || undefined;
```
Nokel81, line 21: isn't `kubeApi` already typed as `KubeApi | undefined`?

src/renderer/components/dock/edit-resource/edit-resource-tab.injectable.ts
```diff
@@ -10,13 +10,15 @@ import type { DockStore, DockTabCreateSpecific, TabId } from "../dock/store";
...
+import getRandomIdForEditResourceTabInjectable from "./get-random-id-for-edit-resource-tab.injectable";
 
 interface Dependencies {
   dockStore: DockStore;
   editResourceStore: EditResourceTabStore;
+  getRandomId: () => string;
 }
 
-const createEditResourceTab = ({ dockStore, editResourceStore }: Dependencies) => (object: KubeObject, tabParams: DockTabCreateSpecific = {}): TabId => {
+const createEditResourceTab = ({ dockStore, editResourceStore, getRandomId }: Dependencies) => (object: KubeObject, tabParams: DockTabCreateSpecific = {}): TabId => {
```
Nokel81: nit: can you move these dependencies onto separate lines to make future diffs easier.

src/renderer/components/dock/edit-resource/view.tsx
```diff
@@ -4,169 +4,82 @@
...
-          }}
-          onError={error => this.error = String(error)}
-        />
+        {model.shouldShowErrorAboutNoResource && (
+          <Notice>
+            Resource not found
+          </Notice>
+        )}
+
+        {!model.shouldShowErrorAboutNoResource && (
```
Nokel81, line 36: nit: could combine these

src/renderer/components/dock/edit-resource/edit-resource-model/call-for-resource/call-for-resource.injectable.ts
```diff
@@ -0,0 +1,48 @@
...
+import { parseKubeApi } from "../../../../../../common/k8s-api/kube-api-parse";
+import getKubeApiFromPathInjectable from "../../../../../../common/k8s-api/kube-api/get-kube-api-from-path.injectable";
+import type { AsyncResult } from "../../../../../../common/utils/async-result";
+import { getErrorMessage } from "../../../../../../common/utils/get-error-message";
+
+export type CallForResource = (
+  selfLink: string
+) => Promise<AsyncResult<KubeObject | undefined>>;
+
+const callForResourceInjectable = getInjectable({
```
Nokel81, line 17: Is the whole of this really necessary for unit testing or for the correct behaviour?

jansav: Both. This is something that I would expect to see in the unit test. `when I do something, calls for resource` 
In the beginning I had it implemented using stores, but changed since that would introduce race-condition when you are refreshing page with tab open. (store might not be loaded yet or not going to be loaded at all, therefore we would have to make sure we load the store here) 

Currently I don't see a place for `ApiManager`, `XStore extends KubeObjectStore` and `XApi extends KubeApi` because all of these can be done a lot simpler. More on this topic in the future if we start doing anything about it.

src/renderer/components/dock/edit-resource/view.tsx
jansav: I'd say this is easier to read because there is less depth.

src/common/k8s-api/kube-api/get-kube-api-from-path.injectable.ts
jansav: Line `19` has type `KubeApi<any, any> | undefined` which is causing the trouble. Line `21` is there to make it `KubeApi<KubeObject, KubeJsonApiDataFor<KubeObject>> | undefined`.

This is due the fact that at the moment injection token has to be typed `KubeApi<any, any>` to make it work properly. That's something that we need to figure out later.

src/behaviours/cluster/namespaces/edit-namespace-from-new-tab.test.tsx
jansav: I decided not to use that because we utilize snapshots for UI. It's easy to feel "ah, there's 4 snapshots that are failing" -> check first one and see that it is valid change -> update all because you are thinking that it's the same issue on every snapshot.

src/behaviours/cluster/namespaces/edit-namespace-from-new-tab.test.tsx
jansav: Changed.

src/renderer/components/dock/edit-resource/edit-resource-tab.injectable.ts
jansav: I've inlined the implementation being so simple. Managed to get rid of requirement for `Dependencies`

aleksfront (dismissed):
LGTM, no bugs found.

src/renderer/components/dock/edit-resource/edit-resource-model/call-for-resource/call-for-resource.injectable.ts
Nokel81: Well what about CRD APIs? Wouldn't the place for ApiManager be to handle all of those things?

As for the behaviour, maybe that behaviour is not correct? Should we really always call for resource?

jansav: > Well what about CRD APIs? Wouldn't the place for ApiManager be to handle all of those things? 

Not relevant here. We'll figure out when the time comes. I was just communicating how I currently feel, I might be wrong. :)

jansav: > As for the behaviour, maybe that behaviour is not correct? Should we really always call for resource?

Definitely yes. Editing resource shouldn't have anything to do with how resources are listed. Those are two separate features which **may** share code (or even data) but they should work in isolation. There's no need to couple features together.

jansav: Benefit of this is that now change in how we list resources cannot broke editing resources. I'd say that's quite good thing?

Nokel81: How is it not relevant? If I want to edit a Custom Resource instance I don't see behaviour to cover that and I also don't see implementation to cover that, plus the user should still be able to do it.

jansav: Well that's different thing. I've just added behaviour for testing namespaces, but surely there should be a behaviour for editing custom resources.

Nokel81 (approved):

src/renderer/components/dock/edit-resource/edit-resource-model/call-for-resource/call-for-resource.injectable.ts
jansav: Good catch, editing custom resources didn't work. Thanks @Nokel81. Fixed now

msa0311 (approved):

jim-docker (approved):

PR #5909, opened 2022-07-26
Fix ResourceQuotaDetails quotas display bugs

### LINKED ISSUE #5892: Resource quota details odities
The quotas are displayed as progress bars of sorts (but display full if 0 is used)
Screen Shot 2022-07-25 at 3 11 46 PM

Nokel81:

fixes https://github.com/lensapp/lens/issues/5892

### DIFF
--- a/src/common/utils/convertCpu.ts
+++ b/src/common/utils/convertCpu.ts
@@ -3,18 +3,36 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
 
+import { TypedRegEx } from "typed-regex";
+
 // Helper to convert CPU K8S units to numbers
 
-const thousand = 1000;
-const million = thousand * thousand;
-const shortBillion = thousand * million;
+const unitConverters = new Map([
+  ["m", 1000 ** -1], // milli
+  ["", 1000 ** 0], // no units
+  ["k", 1000 ** 1],
+  ["M", 1000 ** 2],
+  ["G", 1000 ** 3],
+  ["P", 1000 ** 4],
+  ["T", 1000 ** 5],
+  ["E", 1000 ** 6],
+]);
 
-export function cpuUnitsToNumber(cpu: string) {
-  const cpuNum = parseInt(cpu);
+const cpuUnitsRegex = TypedRegEx("^(?<digits>\\d+)(?<unit>.*)$");
 
-  if (cpu.includes("m")) return cpuNum / thousand;
-  if (cpu.includes("u")) return cpuNum / million;
-  if (cpu.includes("n")) return cpuNum / shortBillion;
+export function cpuUnitsToNumber(value: string) {
+  const match = cpuUnitsRegex.captures(value);
 
-  return parseFloat(cpu);
+  if (!match) {
+    return undefined;
+  }
+
+  const { digits, unit } = match;
+  const convertion = unitConverters.get(unit);
+
+  if (convertion === undefined) {
+    return undefined;
+  }
+
+  return parseInt(digits) * convertion;
 }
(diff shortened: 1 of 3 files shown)


jim-docker:
Nice, it actually works now. But, perhaps unrelated, is this right?

Screen Shot 2022-07-28 at 5 29 51 PM

cpu is `500m / 1k`? Isn't that 0.05% not 50%? Used the Resource Quotas example from https://kubernetes.io/docs/concepts/policy/resource-quotas/#:~:text=apiVersion%3A%20v1,%22low%22%5D (the above png shows the pods-high resource quota, which has `hard.cpu = "1000"`

Nokel81:
I have now also fixed what @jim-docker found, 

here is a screenshot:

Screen Shot 2022-07-29 at 3 38 08 PM

src/common/utils/convertCpu.ts
```diff
@@ -3,18 +3,36 @@
...
+  }
 
-export function cpuUnitsToNumber(cpu: string) {
-  const cpuNum = parseInt(cpu);
+  const { digits, unit } = match;
+  const convertion = unitConverters.get(unit);
 
-  if (cpu.includes("m")) return cpuNum / thousand;
-  if (cpu.includes("u")) return cpuNum / million;
-  if (cpu.includes("n")) return cpuNum / shortBillion;
```
jim-docker, line 17: Could `u` and `n` be from an older kube version? Could keep them around just to be safe?

src/common/utils/convertCpu.ts
```diff
@@ -3,18 +3,36 @@
...
+  const match = cpuUnitsRegex.captures(value);
+
+  if (!match) {
+    return undefined;
+  }
 
-export function cpuUnitsToNumber(cpu: string) {
-  const cpuNum = parseInt(cpu);
+  const { digits, unit } = match;
+  const convertion = unitConverters.get(unit);
```
jim-docker: nit
```suggestion
  const conversion = unitConverters.get(unit);
```

src/renderer/components/+config-resource-quotas/resource-quota-details.tsx
```diff
@@ -53,7 +67,7 @@ function renderQuotas(quota: ResourceQuota): JSX.Element[] {
             value={current}
             tooltip={(
               <p>
-                {`Set: ${rawMax}. Usage: ${usage}%`}
+                {`Set: ${rawMax}. Usage: ${+usage.toFixed(2)}%`}
```
jim-docker, line 70: Is this better? It will show 100 as 100.00? I guess otherwise anything under 1 will show as 1?

Nokel81: So this will show `100` as `100` but will correctly show `0.05` as `0.05` instead of what it currently does which is `1`

src/common/utils/convertCpu.ts
Nokel81: Sure

src/common/utils/convertCpu.ts
```diff
@@ -3,18 +3,38 @@
...
 
-  if (cpu.includes("m")) return cpuNum / thousand;
-  if (cpu.includes("u")) return cpuNum / million;
-  if (cpu.includes("n")) return cpuNum / shortBillion;
+  if (conversion === undefined) {
+    return undefined;
+  }
 
-  return parseFloat(cpu);
+  return parseInt(digits) * conversion;
```
jim-docker: Oh, couldn't the number be a float? You can have 0.5 for cpu, no? https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu:~:text=Fractional%20requests%20are%20allowed

Nokel81: Maybe, but having tried it kubernetes seems to convert a YML file with `0.5` for the CPU into `500m`

ixrock (dismissed):

src/common/utils/convertCpu.ts
Nokel81: Yes parsing float would probably be better. Will change.

jim-docker (approved):

PR #5912, opened 2022-07-26
Fix AddRoleDialog's create button being enabled before entering data

### LINKED ISSUE #5898: Create Role Dialog
1. Create button is active even if the namespace is not selected (causing a nebulous error notification)

Nokel81:

fixes https://github.com/lensapp/lens/issues/5898


jim-docker:
This looks rather big for a patch release...

Nokel81:
@jim-docker I don't think so, the vast majority of the lines changed are within the snapshots.

jim-docker:
> @jim-docker I don't think so, the vast majority of the lines changed are within the snapshots.

I filtered out the snapshots first

src/renderer/components/button/button.tsx
```diff
@@ -49,7 +50,11 @@ export const Button = withTooltip((props: ButtonProps) => {
 
   // render as button
   return (
-    <button type="button" {...btnProps}>
+    <button
+      type="button"
+      {...btnProps}
+      data-waiting={`${waiting}`}
```
aleksfront: Is this necessary? Dialog button can be disabled with native `disabled` prop.

src/renderer/components/wizard/wizard.tsx
```diff
@@ -190,7 +190,9 @@ export class WizardStep<D> extends React.Component<WizardStepProps<D>, WizardSte
     if (this.form.noValidate || this.form.checkValidity()) {
       this.next();
     }
-  }, 100);
+  }, 100, {
+    leading: true,
```
aleksfront: Can you explain what it does?

src/renderer/frames/cluster-frame/cluster-frame-layout-child-component.injectable.tsx
```diff
@@ -30,25 +30,32 @@ const clusterFrameLayoutChildComponentInjectable = getInjectable({
...
+                    <Redirect to={startUrl.get()} />
+                  )
+                  : (
+                    <div className={styles.centering}>
+                      <div className="error">
+                        An error has occured. No route can be found matching the
+                        current route, which is also the starting route.
+                      </div>
+                    </div>
+                  )
```
aleksfront: Seems that this changes completely unrelated to what PR states. Should be moved to separate PR.

aleksfront (changes requested):
This PR does more that it states in the title and description. Also, we should find a way to disable a button without new `data-waiting` prop since it changes a huge amount of snapshots. If needed, it can be added to the specified button only.

src/renderer/components/wizard/wizard.tsx
Nokel81: Sure.

This configures how the `debounce`-ed function is called when `this.submit` is called. Namely, it now calls the inner function immediately and then won't call it again for at least another 100ms. Whereas it used to wait for the first 100ms before calling the inner function.

This article (at the given heading and the next heading of "leading edge") explains it very well.

https://css-tricks.com/debouncing-throttling-explained-examples/#aa-debounce 

---

As to why I needed to do this:

It was for getting the unit tests to pass. If the inner function is only called after 100ms then I would have to add that amount of delay into the test (which cases other problems), but with this configuration change the action happens immediately (and subsequent calls within 100ms are then just ignored) makes the tests easier to write and understand.

src/renderer/frames/cluster-frame/cluster-frame-layout-child-component.injectable.tsx
Nokel81: Yes probably, I did this so that the `AddRoleDialog` could be its own registered component instead of being tagged on to the `Roles` registered component.

src/renderer/components/button/button.tsx
Nokel81: This is not really for having it be disabled but instead being able to write a test for the button to show that it is in the loading state. Similar to other tests that makes sure that a spinner is shown while something is happening.

Nokel81: I will move this change to a separate PR to make it easier to review

src/renderer/frames/cluster-frame/cluster-frame-layout-child-component.injectable.tsx
Nokel81: I will move this to a separate PR to make it easier to review

src/renderer/components/button/button.tsx
Nokel81: I have made https://github.com/lensapp/lens/pull/6040

src/renderer/frames/cluster-frame/cluster-frame-layout-child-component.injectable.tsx
Nokel81: Removed, we can do this in the future.

src/renderer/components/wizard/wizard.tsx
jansav: > It was for getting the unit tests to pass. 

Test shouldn't require this. In case you want to make tests pass without this, then you would have to use `useFakeTime` and `advanceFakeTime`. Here, you are changing behaviour to make test pass, which is wrong.

Nokel81: It is not wrong if the behaviour was incorrect. The point of this denouncing is to make sure that we don't accidentally double submit.

PR #5948, opened 2022-07-29
Open application only if new update is available when using tray to check for updates

### LINKED ISSUE #5941: 'Check for updates' opens Lens unnecessarily
**To Reproduce**
1. Close Lens but have it running in the tray
2. Open tray menu and choose 'Check for updates'

**Expected behavior**
Lens should not open (especially if there is no update available).

**Actual behavior**
Lens open (but there is no update available).

- Lens Version:6.0.0-latest.20220728.2
- OS: OSX

jansav:
Closes #5941

### DIFF
--- a/src/behaviours/application-update/installing-update-using-topbar-button.test.tsx
+++ b/src/behaviours/application-update/installing-update-using-topbar-button.test.tsx
@@ -73,7 +73,7 @@ describe("encourage user to update when sufficient time passed since update was 
     });
 
     describe("given the update check", () => {
-      let processCheckingForUpdates: (source: string) => Promise<{ updateWasDiscoveredAndDownloaded: boolean }>;
+      let processCheckingForUpdates: (source: string) => Promise<{ updateIsReadyToBeInstalled: boolean }>;
 
       beforeEach(async () => {
         processCheckingForUpdates = applicationBuilder.dis.mainDi.inject(
--- a/src/behaviours/application-update/installing-update-using-topbar-button.test.tsx
+++ b/src/behaviours/application-update/installing-update-using-topbar-button.test.tsx
@@ -105,6 +105,11 @@ describe("encourage user to update when sufficient time passed since update was 
           expect(button).toHaveAttribute("data-warning-level", "light");
         });
 
+        // TODO: Implement after starting main and renderer is separated in ApplicationBuilder
+        xit("given closing the application window, when starting the application window again, still shows the button", () => {
+          expect(button).toBeInTheDocument();
+        });
+
         describe("given some time passes, when checking for updates again", () => {
           beforeEach(() => {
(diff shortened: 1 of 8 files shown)


src/renderer/utils/sync-box/provide-initial-values-for-sync-boxes.injectable.ts
```diff
@@ -7,21 +7,35 @@ import { beforeFrameStartsInjectionToken } from "../../before-frame-starts/befor
...
-        initialValues.forEach(({ id, value }) => {
-          setSyncBoxState(id, value);
+        runInAction(() => {
+          initialValues.forEach(({ id, value }) => {
+            const syncBox = syncBoxes.find((box) => box.id === id);
+
+            assert(syncBox);
+
+            setSyncBoxState(syncBox, value);
+          });
```
jansav, line 38: Note: This change here would be covered by that skipped unit test having TODO. Importance of that rises because this is something that we would have seen.

panuhorsmalahti (approved):

src/behaviours/application-update/installing-update-using-topbar-button.test.tsx
```diff
@@ -107,6 +105,11 @@ describe("encourage user to update when sufficient time passed since update was
           expect(button).toHaveAttribute("data-warning-level", "light");
         });
 
+        // TODO: Implement after starting main and renderer is separated in ApplicationBuilder
+        xit("given closing the application window, when starting the application window again, still shows the button", () => {
+          expect(button).toBeInTheDocument();
+        });
```
aleksfront, line 111: Nit: maybe better to use `it.todo` for such cases? https://jestjs.io/docs/api#testtodoname

src/main/application-update/check-for-updates-tray-item.injectable.ts
```diff
@@ -59,9 +59,11 @@ const checkForUpdatesTrayItemInjectable = getInjectable({
 
       click: pipeline(
         async () => {
-          await processCheckingForUpdates("tray");
+          const { updateIsReadyToBeInstalled } = await processCheckingForUpdates("tray");
 
-          await showApplicationWindow();
+          if (updateIsReadyToBeInstalled) {
+            await showApplicationWindow();
```
aleksfront, line 65: There is no point to show just the application window after user intentionally checked for updates. We should show message at least or dedicated page says `New update is available` with some update info and an action button ("Update now"). Could be done as a separate PR though.

jansav: Hmm. I'd say it's not that bad, since first thing you see when the application opens in this case is the "Update" button in the top bar. :)

aleksfront (approved):

PR #5949, opened 2022-07-29
Fix ingress list line heights

### LINKED ISSUE #5902: Ingress list can contain line gaps
Some lines in the `Ingress` list may be empty. Probably multiline table functionality needs to be checked.
ingress list

aleksfront:
Fixes #5902 

https://user-images.githubusercontent.com/9607060/181698918-ace51abb-d57c-4d02-a34e-c54c2e0c9b98.mov

### DIFF
--- a/src/renderer/components/+network-ingresses/ingresses.scss
+++ b/src/renderer/components/+network-ingresses/ingresses.scss
@@ -5,6 +5,10 @@
 
 .Ingresses {
   .TableCell {
+    &.checkbox .Checkbox {
+      align-items: flex-start;
+    }
+
     &.rules {
       flex-grow: 3.0;
       overflow-x: scroll;
@@ -14,8 +18,13 @@
         display: none;
       }
 
-      span:not(:last-of-type) {
-        margin-right: 1em;
+      > div {
+        overflow: hidden;
+        text-overflow: ellipsis;
+      }
+
+      > div + div {
+        margin-top: 6px;
       }
     }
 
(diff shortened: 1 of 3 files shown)


src/renderer/components/virtual-list/virtual-list.tsx
```diff
@@ -87,7 +87,7 @@ function VirtualListInner<T extends { getId(): string } | string>({
   useEffect(() => {
     try {
       if (prevItems.current.length !== items.length || !isEqual(prevRowHeights.current, rowHeights)) {
-        listRef.current?.resetAfterIndex(0, false);
+        listRef.current?.resetAfterIndex(0);
```
aleksfront, line 90: Removing force render delay attribute allows to update table properly after changing sorting order.

Nokel81 (dismissed):

src/renderer/components/+network-ingresses/ingresses.scss
```diff
@@ -14,8 +18,13 @@
...
-      span:not(:last-of-type) {
-        margin-right: 1em;
+      > div {
+        overflow: hidden;
+        text-overflow: ellipsis;
+      }
+
+      > div + div {
+        margin-top: 6px;
       }
```
jansav: These are very easy to break. Add classes.

ixrock: `+` it would be easier to refactor when converting to css-modules

ixrock (dismissed):
Why this happens only with Ingress List view? 
It feels like should be fixed at more global level (e.g. table/virtual-list).

Also did you check if it works correctly when table has a scroll with more items? _(didn't notice this in the video)_

Otherwise LGTM.

Nokel81 (approved):

ixrock (approved):

PR #5950, opened 2022-07-29
Fix No metrics error

gjkamstra:
Issue occurs when using Prometheus helm chart with default configuration


gjkamstra:
too early....

PR #5954, opened 2022-07-29
Fix KubeObjectDetails drawer not working for Cluster scoped resources

### LINKED ISSUE #5903: PersistenVolumeClass list should load StorageClasses internally
Storage Classes should be loaded along with the Persistent Volume Class list to show proper information in Storage Class details.

For now, Storage Class details would be empty until `Storage Classes` tab clicked.

https://user-images.githubusercontent.com/9607060/180964793-f30f873c-180c-4e22-b4f5-2bc094c7a763.mov

### LINKED ISSUE #5917: Overview page Pod details are empty
Clicking on some pod in `Involved Object` column shows empty details panel even if pod store items loaded in memory.

Might be related to #5903 

https://user-images.githubusercontent.com/9607060/181170911-db04f352-5b7b-45b4-9715-21ddd257ee7d.mov

### LINKED ISSUE #5936: Accessing the cluster role which I just created leads to blank details
**Screenshots**
https://user-images.githubusercontent.com/3083083/181453844-527e5bce-e517-44dd-b3a1-b60f2a6c7c88.mov

Nokel81:
- Fix not displaying spinner while object is not defined

- Fix not displaying loading error if object is not defined

fixes https://github.com/lensapp/lens/issues/5917
fixes https://github.com/lensapp/lens/issues/5903
fixes https://github.com/lensapp/lens/issues/5936

### DIFF
--- a/src/common/k8s-api/kube-object.store.ts
+++ b/src/common/k8s-api/kube-object.store.ts
@@ -355,7 +355,7 @@ export abstract class KubeObjectStore<
   async loadFromPath(resourcePath: string) {
     const { namespace, name } = parseKubeApi(resourcePath);
 
-    assert(name && namespace, "Both name and namesapce must be part of resourcePath");
+    assert(name, "name must be part of resourcePath");
 
     return this.load({ name, namespace });
   }
--- a/src/renderer/components/kube-object-details/kube-object-details.tsx
+++ b/src/renderer/components/kube-object-details/kube-object-details.tsx
@@ -8,7 +8,7 @@ import "./kube-object-details.scss";
 import React from "react";
 import { disposeOnUnmount, observer } from "mobx-react";
 import type { IComputedValue } from "mobx";
-import { computed, observable, reaction, makeObservable } from "mobx";
+import { observable, reaction, makeObservable } from "mobx";
 import { Drawer } from "../drawer";
 import type { KubeObject } from "../../../common/k8s-api/kube-object";
 import { Spinner } from "../spinner";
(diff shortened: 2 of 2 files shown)


src/common/k8s-api/kube-object.store.ts
```diff
@@ -355,7 +355,7 @@ export abstract class KubeObjectStore<
   async loadFromPath(resourcePath: string) {
     const { namespace, name } = parseKubeApi(resourcePath);
 
-    assert(name && namespace, "Both name and namesapce must be part of resourcePath");
+    assert(name, "name must be part of resourcePath");
```
jim-docker, line 358: was this just wrong, or do your other changes here necessitate this?

jim-docker (approved):
This just looks like refactoring, what is the actual fix?

src/common/k8s-api/kube-object.store.ts
Nokel81: No this was the main fix. For cluster scoped resources there will never be a namespace and for namespace scoped resources the server will return an error. In theory we could check if the namespace was present if we know that the API would require it.

Nokel81:
@jim-docker The fix is in Kube object store.

The rest of the changes are there is make the component not empty when it fails. They were what I did to figure out what the error was. I would consider them big fixes and not just refactorings.

jim-docker:
> @jim-docker The fix is in Kube object store.
> 
> The rest of the changes are there is make the component not empty when it fails. They were what I did to figure out what the error was. I would consider them big fixes and not just refactorings.

Ok, but wasn't there already code to make sure the component was not empty (`isLoading`, `loadingError`)?

Nokel81:
> Ok, but wasn't there already code to make sure the component was not empty (isLoading, loadingError)?

Yes but they were only being called if `this.object` was already defined which is why in practice the drawer was blank.

jansav (approved):

jim-docker:
> > Ok, but wasn't there already code to make sure the component was not empty (isLoading, loadingError)?
> 
> Yes but they were only being called if `this.object` was already defined which is why in practice the drawer was blank.

So that was the bug fix

PR #5962, opened 2022-08-01
Replace individual mocks for withTooltip with global mock

jansav:
**Motivation**
1. Remove duplication
2. Prevent situation where you accidentally forgot to mock it, which leads to introduction of tests that fail for wrong reason.
3. One place to control how we want tooltips to show in unit tests

This PR contains changes only to tests.

### DIFF
--- a/src/behaviours/add-cluster/__snapshots__/navigation-using-application-menu.test.tsx.snap
+++ b/src/behaviours/add-cluster/__snapshots__/navigation-using-application-menu.test.tsx.snap
@@ -288,7 +288,11 @@ exports[`add-cluster - navigation using application menu when navigating to add 
             </p>
             <div
               class="flex column"
-            />
+            >
+              <textarea
+                data-testid="monaco-editor-for-undefined"
+              />
+            </div>
             <div
               class="actions-panel"
             >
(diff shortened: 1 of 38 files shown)


Nokel81 (approved):

src/behaviours/add-cluster/__snapshots__/navigation-using-application-menu.test.tsx.snap
```diff
@@ -288,7 +288,11 @@ exports[`add-cluster - navigation using application menu when navigating to add
             </p>
             <div
               class="flex column"
-            />
+            >
+              <textarea
+                data-testid="monaco-editor-for-undefined"
```
aleksfront, line 293: Some id is broken or missing? Are we sure it should end with `undefined`?

aleksfront (approved):
If `withTooltip` decorator causing trouble for tests, we should consider removing it. Only 5 components using it, maybe it's easier to just stop using `withTooltip` HOC.

Nokel81:
@aleksfront Maybe but that would be a breaking API change since those are exported to the extension API.

src/behaviours/add-cluster/__snapshots__/navigation-using-application-menu.test.tsx.snap
jansav: Nah, decided not to add ID for that editor here since it would be change in implementation.

PR #5964, opened 2022-08-01
fix(ci): make appBundleId configurable.

jkroepke:
Allow to notarize 3rd party builds without manually patch build files.


jkroepke:
Not sure, how I could add the label 'chore' here.

jansav:
@jakolehm @Nokel81 I think that the default value shouldn't be here either.

jakolehm (approved):
LGTM (default value should be removed in a separate PR)

jansav (approved):

PR #5970, opened 2022-08-02
Add option to download all logs from the container

### LINKED ISSUE #2677: Download/copy pod logs
download/copy logs

**Why is this needed**:
highlight specific logs always refreshed after few seconds scrolling, so it's better to have option to download whole pod logs, or at least make the select logs less buggy

Linux

- Kubernetes distribution: not sure, it was exists way before i joined the company
- Desktop OS: Ubuntu 20.04 64-bit

btw awesome job making this `kontenta-lens`, i think this the best GUI for kubernetes [i've ever tried](http://kokizzu.blogspot.com/2021/03/kubernetes-gui.html).

### LINKED ISSUE #4803: Cannot find "Download All Logs" button for pod logs
I'm using Lens v.5.3.4. I would like to download(or search) entire pod log but Lens can only download logs that is retrieved on the screen. So I have not choice but to scroll up to top of the logs manually.

On #1722, it says Lens has "Download All Logs" button for pod logs but I couldn't find it.

Could you help me navigate to "Download All Logs" or retrieve all available logs at once?

aleksfront:
Adding dropdown with 2 options: `Download visible logs` and `Download all logs`. Download all logs option will request container logs without any `sinceTime` or `tailLine` attributes which allows to download file with the logs from the start.

download logs menu

https://user-images.githubusercontent.com/9607060/182360081-e187df45-9ad5-4d09-92e9-02fa403c6b5d.mov

Fixes https://github.com/lensapp/lens/issues/4803


src/renderer/components/dock/logs/__test__/download-logs.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+    it("contains download dropdown button", () =>  {
+      expect(rendered.getByTestId("download-logs-dropdown")).toBeInTheDocument();
+    });
+
+    describe("when clicking on button", () => {
+      beforeEach(() => {
+        const button = rendered.getByTestId("download-logs-dropdown");
+
+        act(() => button.click());
```
Nokel81: Is this act call required?

src/renderer/components/dock/logs/controls.tsx
```diff
@@ -44,13 +47,26 @@ const NonInjectedLogControls = observer(({ openSaveFileDialog, model }: Dependen
...
+    return new Promise((resolve) => {
+      const fileName = pod.getName();
+      const logsToDownload: string[] = showTimestamps
+        ? model.logs.get()
+        : model.logsWithoutTimestamps.get();
 
-    openSaveFileDialog(`${fileName}.log`, logsToDownload.join("\n"), "text/plain");
+      openSaveFileDialog(`${fileName}.log`, logsToDownload.join("\n"), "text/plain");
+      resolve(true);
+    });
```
Nokel81: Why does this need to be async?

src/renderer/components/dock/logs/controls.tsx
```diff
@@ -44,13 +47,26 @@ const NonInjectedLogControls = observer(({ openSaveFileDialog, model }: Dependen
...
 
+  const downloadAllLogs = async () => {
+    const pod = model.pod.get();
+    
+    if (pod) {
+      const logs = await callForAllLogs(pod, { timestamps: showTimestamps, previous });
+      
+      openSaveFileDialog(`${pod.getName()}.log`, logs, "text/plain");
+    }
+  };
```
Nokel81: Wouldn't it be a better idea to have this be part of the model directly?

Nokel81 (commented):
Looks good

jim-docker (dismissed):
minor point, it seems when "saving all" the file is terminated with a line-feed while when "saving visible" it is not.

src/renderer/components/dock/logs/controls.tsx
```diff
@@ -77,11 +93,10 @@ const NonInjectedLogControls = observer(({ openSaveFileDialog, model }: Dependen
           onChange={togglePrevious}
           className="show-previous"
         />
-        <Icon
```
jim-docker, line 80: couldn't keep the icon? "Download" takes up more space where space is at a premium

src/renderer/components/dock/logs/__test__/download-logs.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+    builder.setEnvironmentToClusterFrame();
+
+    builder.beforeApplicationStart(({ rendererDi }) => {
+      rendererDi.override(callForAllLogsInjectable, () => () => Promise.resolve("all-logs"));
+      rendererDi.override(logsViewModelInjectable, () => getOnePodViewModel("foobar"));
+      rendererDi.override(getRandomIdForPodLogsTabInjectable, () =>
+        jest
+          .fn(() => "some-irrelevant-random-id")
+          .mockReturnValueOnce("some-first-tab-id")
+          .mockReturnValueOnce("some-second-tab-id"),
```
jansav: Are these used at all?

src/renderer/components/dock/logs/__test__/download-logs.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+  let builder: ApplicationBuilder;
+  let openSaveFileDialogMock: jest.MockedFunction<() => void>;
+
+  beforeEach(() => {
+    builder = getApplicationBuilder();
+
+    builder.setEnvironmentToClusterFrame();
+
+    builder.beforeApplicationStart(({ rendererDi }) => {
+      rendererDi.override(callForAllLogsInjectable, () => () => Promise.resolve("all-logs"));
```
jansav: I would prefer if you used `asyncFn` instead of `Promise.resolve()` because now you cannot test what happens when call has made but not resolved yet. Or what happens if call rejects.

src/renderer/components/dock/logs/__test__/download-logs.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+      rendererDi.override(logsViewModelInjectable, () => getOnePodViewModel("foobar"));
```
jansav: How about not overriding the LogsViewModel, but it's dependencies? That's a way to get rid of `getOnePodViewModel` and `mockLogTabViewModel`

src/renderer/components/dock/logs/__test__/download-logs.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+      it("shows download all logs menu item", () => {
+        expect(rendered.getByTestId("download-all-logs")).toBeInTheDocument();
+      });
+
+      it("when selected 'download visible logs', shows save dialog with proper attributes", async () => {
+        const downloadMenuItem = rendered.getByTestId("download-visible-logs");
+
+        downloadMenuItem.click();
+        
+        await waitFor(() =>
```
jansav: `waitFor` is not good. E.g. it causes test to not fail before timeout if `openSavefileDialog` is not called at all. 

I bet that changing `callForAllLogsMock` to be `asyncFn` and then calling `callForAllLogsMock.resolve('some-logs')` somewhere would be enough to get rid of this. 

If you still prefer using the `Promise.resolve(...)` type mock for it, then `await flushPromises()` should be enough so that you can get rid of `waitFor`.

src/renderer/components/dock/logs/__test__/download-logs.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+    });
+
+    it("renders", () => {
+      expect(rendered.baseElement).toMatchSnapshot();
+    });
+
+    it("contains download dropdown button", () =>  {
+      expect(rendered.getByTestId("download-logs-dropdown")).toBeInTheDocument();
+    });
+
```
jansav: This behaviour is missing tests:

1. "calls the logs with query"
2. "when call for logs resolve with logs, ..."
3. "when call for logs resolve without logs, ..."
4. "when call for logs reject, ..."

src/renderer/components/dock/logs/__test__/download-logs.test.tsx
```diff
@@ -0,0 +1,186 @@
+/**
```
jansav: This test suite is a behaviour, so lets move it to `behaviours`

src/renderer/components/dock/logs/__test__/download-logs.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+          expect(openSaveFileDialogMock).toHaveBeenCalledWith("dockerExporter.log", "some-logs", "text/plain"),
+        );
+      });
+
+      it("when selected 'download all logs', shows save dialog with proper attributes", async () => {
+        const downloadMenuItem = rendered.getByTestId("download-all-logs");
+
+        downloadMenuItem.click();
+
+        await waitFor(() =>
```
jansav: Read above: `waitFor`.

src/renderer/components/dock/logs/__test__/download-logs.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+          expect(openSaveFileDialogMock).toHaveBeenCalledWith("dockerExporter.log", "all-logs", "text/plain"),
+        );
+      });
+
+      it("when selected 'download all logs', block download dropdown for interaction", async () => {
+        const downloadMenuItem = rendered.getByTestId("download-all-logs");
+        
+        downloadMenuItem.click();
+
+        await waitFor(() =>
```
jansav: Read above: `waitFor`

src/renderer/components/dock/logs/__test__/download-logs.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+          expect(rendered.getByTestId("download-logs-dropdown")).toHaveAttribute("disabled"),
+        );
+      });
+
+      it("when save file dialog opens, restore download button for interaction", async () => {
+        const downloadMenuItem = rendered.getByTestId("download-all-logs");
+
+        downloadMenuItem.click();
+
+        await waitFor(() =>
```
jansav: Read above: `waitFor`

src/renderer/components/dock/logs/call-for-all-logs.injectable.ts
```diff
@@ -0,0 +1,25 @@
+/**
```
jansav: This is unnecessary indirection to `callForLogs`. It shouldn't be coupled with `Pod`.

Replace usage of this with: `callForLogs({ name: pod.getName(), namespace:  pod.getNs() }, query)`

src/renderer/components/dock/logs/controls.tsx
```diff
@@ -44,13 +47,26 @@ const NonInjectedLogControls = observer(({ openSaveFileDialog, model }: Dependen
     model.reloadLogs();
   };
 
+  const downloadAllLogs = async () => {
```
jansav: This logic belongs in to model.

src/renderer/components/dock/logs/controls.tsx
```diff
@@ -44,13 +47,26 @@ const NonInjectedLogControls = observer(({ openSaveFileDialog, model }: Dependen
...
+    const pod = model.pod.get();
+    
+    if (pod) {
+      const logs = await callForAllLogs(pod, { timestamps: showTimestamps, previous });
```
jansav: `const logs = callForLogs({ name: pod.getName(), namespace: pod.getNs() }, { timestamps: showTimestamps, previous })`

src/renderer/components/dock/logs/controls.tsx
```diff
@@ -44,13 +47,26 @@ const NonInjectedLogControls = observer(({ openSaveFileDialog, model }: Dependen
...
+      
+      openSaveFileDialog(`${pod.getName()}.log`, logs, "text/plain");
+    }
+  };
+
   const downloadLogs = () => {
```
jansav: This logic belongs to model

src/renderer/components/dock/logs/controls.tsx
```diff
@@ -77,11 +93,10 @@ const NonInjectedLogControls = observer(({ openSaveFileDialog, model }: Dependen
...
         />
-        <Icon
-          material="get_app"
-          onClick={downloadLogs}
-          tooltip="Download"
-          className="download-icon"
+
+        <DownloadLogsDropdown
+          downloadVisibleLogs={downloadLogs}
+          downloadAllLogs={downloadAllLogs}
```
jansav: How about:

```jsx
<div>
  <DropdownToggle dropdownId="download-logs-dropdown">
    <Button>Download...</Button>
  </DropdownToggle>

  <DropdownContents dropdownId="download-logs-dropdown">
    {() => [
       <Button key="download-all" onClick={model.downloadAll}>Download all</Button>,
       <Button key="download visible" onClick={model.downloadVisible}>Download visible</Button>
    ]}
  </DropdownContents>
</div>
```

In this way: 
1. We would have reusable `Dropdown` component
2. Button that toggles visibility of dropdown can be anywhere
3. Business logic for closing dropdown when item is clicked can be written to Dropdown, and not to DownloadLogs.

Actually what I'm defining here is something even more reusable. It doesn't have anything to do with Dropdown. Dropdown is just one usage of it.

So maybe even:

```jsx
const Dropdown = ({ id, contentForToggle, contents }) => (
  <div class={styles.dropdown}>
    <ActivatableToggle activationId={id}>
      <Button>Download...</Button>
    </ActivatableToggle>

    <ActivatableContent activationId="download-logs-dropdown">
      {() => [
         <Button key="download-all" onClick={model.downloadAll}>Download all</Button>,
         <Button key="download visible" onClick={model.downloadVisible}>Download visible</Button>
      ]}
    </ActivatableContent>
  </div>
);
```
```jsx
<Dropdown
  id="download-logs-dropdown"
  contentForToggle={<Button>Download...</Button>}
  contents={
    {() => [
      <Button key="download-all" onClick={model.downloadAll}>Download all</Button>,
      <Button key="download visible" onClick={model.downloadVisible}>Download visible</Button>
    ])
  } 
/>
```

src/renderer/components/dock/logs/download-logs-dropdown.tsx
```diff
@@ -0,0 +1,68 @@
+/**
```
jansav, line 1: Everything here could be done with reusable dropdown / activatable component. Read above

src/renderer/components/dock/logs/download-logs-dropdown.module.scss
```diff
@@ -0,0 +1,37 @@
+.dropdown {
```
jansav, line 1: Styles for reusable dropdown?

src/renderer/components/dock/logs/controls.tsx
jansav: I agree, it doesn't look async at all. `openSaveFileDialog` in the other hand does more than name says. Maybe it could be renamed to `askDirectoryAndSaveToFile` and made async.

src/renderer/components/dock/logs/download-logs-dropdown.module.scss
ixrock: Depends on what do you mean. This is is css-module so the class name would be unique.
If you're about customization of component I would add some more `--css-variables` to manage most important parts.

ixrock (dismissed):

src/renderer/components/dock/logs/download-logs-dropdown.module.scss
jansav: I mean that is it something that could be used for re-usable dropdown component that I mentioned on other comments. I know that this is CSS modules and and the class names wont collide.

jansav (dismissed):
Left individual comments.

src/renderer/components/dock/logs/__test__/download-logs.test.tsx
aleksfront: Yes, it removes error notification (https://reactjs.org/docs/test-utils.html#act):
```
Warning: An update to DownloadLogsDropdown inside a test was not wrapped in act(...).
When testing, code that causes React state updates should be wrapped into act(...):
    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */
This ensures that you're testing the behavior the user would see in the browser.
```

src/renderer/components/dock/logs/__test__/download-logs.test.tsx
aleksfront: Removed unused lines.

src/renderer/components/dock/logs/controls.tsx
aleksfront: Moved to injectables.

src/renderer/components/dock/logs/controls.tsx
aleksfront: Moved `downloadLogs` part to logs model and removed a Promise "wrapper".

src/renderer/components/dock/logs/__test__/download-logs.test.tsx
aleksfront: Added missing tests.

src/renderer/components/dock/logs/controls.tsx
aleksfront: Leaving only 1 icon won't show us that there more than 1 option hidden in the dropdown.

src/renderer/components/dock/logs/__test__/download-logs.test.tsx
aleksfront: `asyncFn` didn't work for me. Used native jest `mockResolvedValue()` instead.

src/renderer/components/dock/logs/download-logs-dropdown.module.scss
aleksfront: Created a `<Dropdown/>` component but without any styles provided. It's better to keep components "headless" to avoid CSS specificity issues and overwrite style attempts in the future.

aleksfront:
Made changes to the PR, please review again.

https://user-images.githubusercontent.com/9607060/184640287-6ccb80f7-b3b8-4085-8a06-128155e1f930.mov

src/renderer/components/dock/logs/__test__/download-logs.test.tsx
aleksfront: Moved into `behaviours`

Nokel81 (approved):
minor point, it would be nice if the dropdown selector had some loading animation while the logs were getting downloaded.

src/behaviours/pod-logs/download-logs.test.tsx
```diff
@@ -0,0 +1,269 @@
...
+            const downloadMenuItem = rendered.getByTestId("download-visible-logs");
+
+            act(() => downloadMenuItem.click());
+
+            expect(rendered.getByTestId("download-logs-dropdown")).not.toHaveAttribute("disabled");
+          });
+        });
+      });
+
+      describe("when call for logs resolves with no logs", () => {
```
jim-docker, line 195: Why show he save dialog for "visible logs" (and allow creating/saving an empty file), but show no save dialog for "all logs" (and show no message about what happened). Both seem odd behaviours.

src/renderer/components/dock/logs/controls.tsx
jim-docker: I guess the dropdown title(?) can't be an icon, only text?

jim-docker (approved):

src/behaviours/pod-logs/download-logs.test.tsx
aleksfront: Rearranging test cases in the upcoming PR.

anagami:
Hi!
Don't work on windows 10 machine

when try download - get error
{
    "error": {
        "kind": "Status",
        "apiVersion": "v1",
        "metadata": {},
        "status": "Failure",
        "message": "a container name must be specified for pod 7fb6d7fbfc-lz6mk, choose one of: [A B C]",
        "reason": "BadRequest",
        "code": 400
    },
    "messages": [
        "a container name must be specified for pod 7fb6d7fbfc-lz6mk, choose one of: [A B C]"
    ],
    "isUsedForNotification": false
}
using latest version

image

Issue #5968, opened 2022-08-02
Allow multiple application windows


jansav:
**What would you like to be added**:
Allow user to start as many application windows they wants.

**Why is this needed**:
There is no reason to limit amount of application windows. User might want to have separate windows for different clusters. 

**What is needed**:
- [x] Figure out when kubeconfig synchronization should start & stop #5982 
- [ ] Figure out when catalog synchronization should start & stop
- [x] Move starting and stopping check for updates to application start/stop #5981
- [ ] Figure out why `lensProtocolRouterMain.rendererLoaded` is needed and replace it with something more elegant
- [ ] Make sure that navigating from e.g. tray, application menu will work properly with multiple windows
  - [x] Replace hard-coded `applicationWindow` with `createApplicationWindow()` which is called once on application start. (`applicationWindow` -> `getCurrentApplicationWindow`) #5979
  - [ ] Introduce a way to detect which application window is "current"
- [x] Allow behavioural testing for multiple windows (remember that there can only be one `document`) #5979
- [ ] Add tray menu items for controlling application windows
- [ ] Add application menu items for controlling application windows

Original: https://github.com/lensapp/lens/issues/431

arash-bizcover:
> Allow user to start as many application windows she wants.

The user is not a she, he uses a male avatar. How rude are you

PR #5979, opened 2022-08-04
Replace static application window with ability to create many

jansav:
What we are doing here:
1. Instead of having static `applicationWindowInjectable` which is always present, we switch to using `createApplicationWindowInjectable` which can be called as many times as you wish. This is to allow creation of multiple application windows in the future.
2. Separate starting main application and starting windows in the behaviours. Before it wasn't possible to behavioural test what happens in main when window hasn't started yet. 

**Notice that there are two commits, first one makes the implementation change and second one makes the required changes in tests. I'd suggest to review them one by one.**

This PR tackles some requirements for #5968. 


jansav:
Don't start review yet. I'm using this PR to figure out some slow tests in CI and will clean up it after that.

jansav:
This is now ready to be reviewed. @jim-docker @Nokel81

jim-docker:
> This is now ready to be reviewed. @jim-docker @Nokel81

FYI I have started looking...

src/main/start-main-application/lens-window/application-window/get-current-application-window.injectable.ts
```diff
@@ -0,0 +1,16 @@
...
+import { getInjectable } from "@ogre-tools/injectable";
+import { first } from "lodash/fp";
+import { applicationWindowInjectionToken } from "./application-window-injection-token";
+
+const getCurrentApplicationWindowInjectable = getInjectable({
+  id: "get-current-application-window",
+
+  instantiate: (di) => () =>
+    first(di.injectMany(applicationWindowInjectionToken)),
+});
```
Nokel81, line 14: Wouldn't it be better if this was a computed?

src/main/start-main-application/lens-window/show-application-window.injectable.ts
```diff
@@ -4,22 +4,31 @@
...
     const focusApplication = di.inject(focusApplicationInjectable);
 
     return async () => {
       focusApplication();
 
+      let applicationWindow: LensWindow | undefined = getApplicationWindow();
+
+      if (!applicationWindow) {
+        applicationWindow = createFirstApplicationWindow();
+      }
```
Nokel81: Couldn't this easily be?

```
const applicationWindow = getApplicationWIndow() ?? createFirstApplicationWindow();
```

src/main/start-main-application/lens-window/application-window/create-first-application-window.injectable.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import createApplicationWindowInjectable from "./create-application-window.injectable";
+
+// Note: Motivation is to create the window with same ID to use stored dimensions
```
Nokel81, line 8: Why would not having a more general `createNextApplicationWindow` be better which picked the correct ID?

src/renderer/components/test-utils/get-application-builder.tsx
```diff
@@ -62,34 +57,50 @@ import activeKubernetesClusterInjectable from "../../cluster-frame-context/activ
...
+    get: (id: string) => {
+      main: LensMainExtension;
 
-    main: {
-      enable: EnableExtensions<LensMainExtension>;
-      disable: DisableExtensions<LensMainExtension>;
+      applicationWindows: Record<string, LensRendererExtension> & {
+        only: LensRendererExtension;
+      };
     };
```
Nokel81, line 88: This is a strange name or is is for getting the individual instance of an extension?

Also what does `only` mean here?

src/renderer/components/test-utils/get-application-builder.tsx
```diff
@@ -209,60 +173,135 @@ export const getApplicationBuilder = () => {
...
-  const disableRendererExtension = disableExtensionsFor(rendererExtensionsState, rendererDi);
-  const disableMainExtension = disableExtensionsFor(mainExtensionsState, mainDi);
+    windowDi.register(rendererExtensionsStateInjectable);
 
-  const selectOptionFor = (menuId: string) => (labelText: string) => {
-    const menuOptions = rendered.baseElement.querySelector<HTMLElement>(
-      `.${menuId}-options`,
+    windowDi.override(
+      currentlyInClusterFrameInjectable,
+      () => environment === environments.clusterFrame,
```
Nokel81, line 191: This seems to imply that all the windows will be the cluster frame at the same time.

src/renderer/components/test-utils/get-application-builder.tsx
```diff
@@ -391,81 +444,129 @@ export const getApplicationBuilder = () => {
...
+          applicationWindows: {
+            get only() {
+              return findExtensionInstance(
+                builder.applicationWindow.only.di,
+                rendererExtensionsInjectable,
+                id,
+              );
+            },
+
+            ...windowInstances,
```
Nokel81, line 522: If a test names an `ID` as "only" this will probably break something.

src/main/start-main-application/lens-window/application-window/get-current-application-window.injectable.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { first } from "lodash/fp";
+import { applicationWindowInjectionToken } from "./application-window-injection-token";
+
+const getCurrentApplicationWindowInjectable = getInjectable({
+  id: "get-current-application-window",
+
+  instantiate: (di) => () =>
+    first(di.injectMany(applicationWindowInjectionToken)),
```
jim-docker, line 13: this will change when there is more than one application window?

src/main/start-main-application/lens-window/reload-current-application-window.injectable.ts
```diff
@@ -0,0 +1,38 @@
...
+import { IpcRendererNavigationEvents } from "../../../renderer/navigation/events";
+import currentClusterFrameInjectable from "./current-cluster-frame/current-cluster-frame.injectable";
+import getCurrentApplicationWindowInjectable from "./application-window/get-current-application-window.injectable";
+
+const reloadCurrentApplicationWindowInjectable = getInjectable({
+  id: "reload-current-application-window",
+
+  instantiate: (di) => {
+    const getCurrentApplicationWindow = di.inject(getCurrentApplicationWindowInjectable);
+    const currentClusterIframe = di.inject(currentClusterFrameInjectable);
```
jim-docker, line 15: These 2 look independent here but they are actually coupled somewhere?

src/behaviours/application-update/selection-of-update-stability.test.ts
```diff
@@ -22,27 +22,27 @@ import setUpdateOnQuitInjectable from "../../main/electron-app/features/set-upda
...
-    applicationBuilder = getApplicationBuilder();
+    builder = getApplicationBuilder();
 
-    applicationBuilder.beforeApplicationStart(({ mainDi, rendererDi }) => {
+    builder.beforeApplicationStart((mainDi) => {
       quitAndInstallUpdateMock = jest.fn();
       checkForPlatformUpdatesMock = asyncFn();
       downloadPlatformUpdateMock = asyncFn();
       setUpdateOnQuitMock = jest.fn();
       showInfoNotificationMock = jest.fn(() => () => {});
```
jim-docker: can delete this line?

src/main/start-main-application/lens-window/application-window/get-current-application-window.injectable.ts
jansav: Yes.

src/main/start-main-application/lens-window/reload-current-application-window.injectable.ts
jansav: From technical perspective they are completely independent from each other. But I'd say that there is logical dependency (which isn't good thing). `currentClusterIframe` will be removed when we get rid of iframes anyway.

src/behaviours/application-update/selection-of-update-stability.test.ts
jansav: Yep, good catch

src/main/start-main-application/lens-window/application-window/get-current-application-window.injectable.ts
jansav: Idea is simple, there will be some kind of state which we keep track of window which has been focused last. Then this implementation will just get the window from that state.

src/main/index.ts
```diff
@@ -10,12 +10,15 @@ import * as LensExtensionsCommonApi from "../extensions/common-api";
...
+import shouldStartHiddenInjectable from "./electron-app/features/should-start-hidden.injectable";
 
 const di = getDi();
 
+const shouldStartHidden = di.inject(shouldStartHiddenInjectable);
+
 const startApplication = di.inject(startMainApplicationInjectable);
 
-void startApplication();
+void startApplication(!shouldStartHidden);
```
Nokel81, line 21: This seems strange, why is it not just a dependency of `startMainApplication`?

src/main/start-main-application/lens-window/application-window/get-current-application-window.injectable.ts
jansav: Most of the usages are not reactive, so I wouldn't do it yet. If needed, we can add `currentApplicationWindow` which is `IComputedValue<LensWindow | undefined>`, but that's for later.

src/renderer/components/test-utils/get-application-builder.tsx
jansav: In 99,9% cases we unit test situations where there is only one application window. That's why I introduced these `only` getters which can be utilized so that you don't have to specify the ID. 

There is one instance of LensRendererExtension for each window, so that's why we need something like this.

src/renderer/components/test-utils/get-application-builder.tsx
jansav: That's limitation for now, yes. Not worth of the time investment since most of the unit tests we write are for one window.

src/renderer/components/test-utils/get-application-builder.tsx
jansav: Probably yes, but I wouldn't be worried about that. I can do similar stuff with thousand other places as well.

src/main/index.ts
jansav: It feels more natural to be an argument.

src/main/start-main-application/lens-window/show-application-window.injectable.ts
jansav: Might be yeah. Tbh, I haven't used `??` a lot and that's the reason why I didn't even think about it.

src/main/start-main-application/lens-window/application-window/create-first-application-window.injectable.ts
jansav: So it would automatically keep in mind if there's already windows and in case there is, it would use `getRandomId()` and if not, it would default to `first-application-window`?

Interesting.

src/main/start-main-application/lens-window/show-application-window.injectable.ts
jansav: Changed.

src/main/start-main-application/lens-window/application-window/create-first-application-window.injectable.ts
Nokel81: Yes that is sort of what I was suggesting.

src/renderer/components/test-utils/get-application-builder.tsx
Nokel81: ok

Nokel81 (approved):

Iku-turso (approved):

PR #6013, opened 2022-08-10
Optimize getting injectable paths only once for all test suites

jansav:
Optimization to make tests faster


Iku-turso (approved):
Had some ideas about not requiring this imitation of auto-registering for unit testing, but let's do that elsewhere.

Nokel81 (approved):

PR #6015, opened 2022-08-10
Decision for organizing code under features

jansav:

Preview: https://github.com/lensapp/lens/blob/2e67d08897cc6a481e34f31cc2b6a57f60c55319/docs/architecture/decisions/README.md

Steps after this PR has been merged (I'm already working on these):

1. Rename "behaviours" to "features"
2. Allow injectables in "features"
3. Move features that have already been built in this way under "features". Using tray and application-update as an example
4. Refactor last remaining parts of code that directly refers to example features. (application menu, getDiForUnitTesting, top-bar, etc. minor stuff. )

### DIFF
--- a/docs/architecture/decisions/0001-organizing-code-under-features.md
+++ b/docs/architecture/decisions/0001-organizing-code-under-features.md
@@ -6,6 +6,8 @@ Date: 2022-07-11
 
 2022-07-11 proposed
 
+2022-08-10 accepted
+
 ## Context
 
 Previously we have organized code under `main`, `renderer` and `common` top level directories. It means that given a feature which has code that needs to run in both environments, we had to split the feature in three different directories.
--- a/docs/architecture/decisions/0001-organizing-code-under-features.md
+++ b/docs/architecture/decisions/0001-organizing-code-under-features.md
@@ -1,0 +1,152 @@
+# 1. Organizing code under features
+
+Date: 2022-07-11
+
+## Status
+
+2022-07-11 proposed
+
+## Context
+
+Previously we have organized code under `main`, `renderer` and `common` top level directories. It means that given a feature which has code that needs to run in both environments, we had to split the feature in three different directories.
+
+**Example:** 
+
+```
+src/
+  behaviours/
+    application-update/
+      installing-update-using-tray.test.ts
+      ...
+      
+  main/
+    application-update
+      process-checking-of-updates.injectable.ts
+      ...
+      
+    tray/
+      tray-menu-item-injection-token.ts
+      check-for-updates-tray-menu-item.injectable.ts
+      ...
+     
+  common/
+    application-update
+      update-channels.ts
+      update-is-being-downloaded.injectable.ts
+      ...
+      
+    front-end-routes/
+      preferences/
+        app/
+          app-preferences-route.injectable.ts
+          navigate-to-app-preferences.injectable.ts
+
+  renderer/
+    top-bar/
+      top-bar.tsx (UpdateButton)
+    
+    preferences/
+      application.tsx (UpdateChannel)
+```
+
+## Decision
+
+Start organizing code under features, meaning that all code that feature needs to exist in a single directory. Features might have child features which have dependency to the parent. We want to limit the coupling of features, which means that feature is not allowed to touch any other feature without creating dependency to it. We want to have visibility for dependency structures of features.
+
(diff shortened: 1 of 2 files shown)


Nokel81 (dismissed):

Nokel81 (approved):

docs/architecture/decisions/0001-organizing-code-under-features.md
```diff
@@ -0,0 +1,154 @@
...
+    
+    preferences/
+      application.tsx (UpdateChannel)
+```
+
+## Decision
+
+Start organizing code under features, meaning that all code that feature needs to exist in a single directory. Features might have child features which have dependency to the parent. We want to limit the coupling of features, which means that feature is not allowed to touch any other feature without creating dependency to it. We want to have visibility for dependency structures of features.
+
+1. Code that belongs to the feature lives in the feature directory. If I want to remove the feature, I can just delete the directory.
```
jim-docker, line 58: In your example below it looks like to remove the tray feature you'd have to delete 3 directories?

docs/architecture/decisions/0001-organizing-code-under-features.md
```diff
@@ -0,0 +1,154 @@
...
+      main/
+        process-checking-of-updates.injectable.ts
+        ...
+      
+      common/
+        update-channels.ts
+        update-is-being-downloaded.injectable.ts
+        ...
+      
+      child-features
```
jim-docker, line 80:
```suggestion
      dependent-features
```
I don't think child conveys the intended meaning here.

jakolehm (approved):

docs/architecture/decisions/0001-organizing-code-under-features.md
jansav: Well of course, because there are features that depend from e.g tray. But let's say you want to delete application-update. You can just delete the directory.

docs/architecture/decisions/0001-organizing-code-under-features.md
jansav: There was discussion about this, and there's the other side as well. Parent/child works better for e.g "Kubernetes/Cluster/Helm/Install Chart".

docs/architecture/decisions/0001-organizing-code-under-features.md
jim-docker: then the argument is not exactly correct...

docs/architecture/decisions/0001-organizing-code-under-features.md
jim-docker: I thought this was purely about code organization, not sure how that is relevant

docs/architecture/decisions/0001-organizing-code-under-features.md
```diff
@@ -0,0 +1,154 @@
...
+      application.tsx (UpdateChannel)
+```
+
+## Decision
+
+Start organizing code under features, meaning that all code that feature needs to exist in a single directory. Features might have child features which have dependency to the parent. We want to limit the coupling of features, which means that feature is not allowed to touch any other feature without creating dependency to it. We want to have visibility for dependency structures of features.
+
+1. Code that belongs to the feature lives in the feature directory. If I want to remove the feature, I can just delete the directory.
+2. Code that is shared between features lives in the closest parent feature. E.g. child features are using code from parent features.
+3. Parent features should never depend on child features. If the dependency is required, the child feature is the one that depends on the parent feature.
```
Nokel81, line 60: This point also seems too prescriptive and backwards since it feels strange to go to the leaves of a feature directory to see the "actual implementation".

docs/architecture/decisions/0001-organizing-code-under-features.md
```diff
@@ -0,0 +1,154 @@
...
+        tray/
+          main/
+            navigate-to-preferences-tray-menu-item.injectable.ts 
+            ...
+```
+
+## Consequences
+
+#### Pros
+1. We will start to focus on the features, not the technical details.
```
Nokel81, line 140: I guess we should also discuss where technical details should live like "readFileInjectable"?

Nokel81 (changes requested):

Nokel81:
Closing as this branch hasn't been touched in months. Can be reopened when the time comes.

PR #6022, opened 2022-08-11
Making apiBase injectable

### LINKED ISSUE #5678: Error on Helm Release Upgrade tab close
```
Uncaught (in promise) AssertionError [ERR_ASSERTION]: cannot reload values if no data
    at UpgradeChartTabStore.reloadValues (store.ts:44)
    at executeAction (mobx.esm.js:1255)
    at UpgradeChartTabStore.reloadValues (mobx.esm.js:1239)
    at NonInjectedUpgradeChart.reloadValues (view.tsx:86)
    at fireImmediately (view.tsx:59)
    at executeAction (mobx.esm.js:1255)
    at Reaction@1477 (mobx.esm.js:1239)
    at reactionRunner (mobx.esm.js:2947)
    at Reaction.onInvalidate_ (mobx.esm.js:2920)
    at Reaction.runReaction_ (mobx.esm.js:2501)
```

https://user-images.githubusercontent.com/9607060/174545057-fb72d957-bbc3-401f-a1fc-839551c85f1b.mov

Nokel81:

Please review commit by commit.

fixes https://github.com/lensapp/lens/issues/5678


src/common/k8s-api/api-base.ts
```diff
@@ -3,38 +3,15 @@
...
-  }, {
-    headers: {
-      "Host": window.location.host,
-    },
-  });
-}
+/**
+ * @deprecated switch to using apiBaseInjectionToken instead
+ */
+export const apiBase = asLegacyGlobalForExtensionApi(apiBaseInjectionToken);
```
Iku-turso: This is the first injectionToken-based legacyGlobal in the codebase. Interesting :)

src/renderer/k8s/api-base.injectable.ts
```diff
@@ -0,0 +1,32 @@
...
+const apiBaseInjectable = getInjectable({
+  id: "api-base",
+  instantiate: (di) => {
+    const createJsonApi = di.inject(createJsonApiInjectable);
+
+    return createJsonApi(
+      {
+        serverAddress: `http://127.0.0.1:${window.location.port}`,
+        apiBase: apiPrefix,
+        debug: di.inject(isDevelopmentInjectable) || isDebugging,
```
Iku-turso, line 24: Hmm. Extract stuff like `isDebugging`, `window.location.host`, `window.location.port` and `apiPrefix` as injectables with side-effect.

src/renderer/k8s/api-kube.injectable.ts
```diff
@@ -5,17 +5,20 @@
...
 const apiKubeInjectable = getInjectable({
   id: "api-kube",
   instantiate: (di) => {
     assert(di.inject(storesAndApisCanBeCreatedInjectionToken), "apiKube is only available in certain environments");
+    const createKubeJsonApi = di.inject(createKubeJsonApiInjectable);
+    const isDevelopment = di.inject(isDevelopmentInjectable);
 
-    const apiKube = new KubeJsonApi({
+    const apiKube = createKubeJsonApi({
       serverAddress: `http://127.0.0.1:${window.location.port}`,
```
Iku-turso: `window.location.port` and such are side-effects :)

src/renderer/k8s/api-base.injectable.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { apiBaseInjectionToken } from "../../common/k8s-api/api-base";
+import { JsonApi } from "../../common/k8s-api/json-api";
+import { apiPrefix, isDebugging } from "../../common/vars";
+import isDevelopmentInjectable from "../../common/vars/is-development.injectable";
+
+const apiBaseInjectable = getInjectable({
```
Iku-turso, line 13: Idea: everything related to `apiBase` strikes to me as overly complicated and could use some simpler competition somewhere else.

src/common/k8s-api/api-kube.ts
```diff
@@ -11,4 +11,7 @@ export const apiKubeInjectionToken = getInjectionToken<KubeJsonApi>({
   id: "api-kube-injection-token",
 });
 
+/**
+ * @deprecated Switch to using apiKubeInjectionToken instead
+ */
 export const apiKube = asLegacyGlobalForExtensionApi(apiKubeInjectionToken);
```
Iku-turso: Note: Looks like this isn't directly exposed in extension API, and it shouldn't therefore be a legacy global at all.

src/common/k8s-api/endpoints/helm-charts.api/get-readme.injectable.ts
```diff
@@ -0,0 +1,24 @@
...
+  instantiate: (di): RequestHelmChartReadme => {
+    const apiBase = di.inject(apiBaseInjectionToken);
+
+    return (repo, name, version) => (
+      apiBase.get(requestReadmeEndpoint.compile({ name, repo }, { version }))
+    );
+  },
+});
+
+export default requestHelmChartReadmeInjectable;
```
Iku-turso: Note: this injectable is used only in renderer, yet it is in common. Stuff like this will probably organize more nicely later with that one ADR for feature-directories.

src/common/k8s-api/endpoints/helm-charts.api/get-readme.injectable.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { urlBuilderFor } from "../../../utils/buildUrl";
+import { apiBaseInjectionToken } from "../../api-base";
+
+const requestReadmeEndpoint = urlBuilderFor("/v2/charts/:repo/:name/readme");
+
+export type RequestHelmChartReadme = (repo: string, name: string, version?: string) => Promise<string>;
+
+const requestHelmChartReadmeInjectable = getInjectable({
```
Iku-turso: Variable names are not in sync with file names. This hurts navigation and discoverability.

src/common/k8s-api/endpoints/helm-releases.api/get-configuration.injectable.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  instantiate: (di): RequestHelmReleaseConfiguration => {
+    const apiBase = di.inject(apiBaseInjectionToken);
+
+    return (name, namespace, all: boolean) => (
+      apiBase.get(requestConfigurationEnpoint.compile({ name, namespace }, { all }))
+    );
+  },
+
+  causesSideEffects: true,
```
Iku-turso: Hmm. Shouldn't be so for this injectable? Maybe for `apiBase`?

src/common/utils/buildUrl.ts
```diff
@@ -35,3 +35,24 @@ export function buildURLPositional<P extends object = {}, Q extends object = {}>
...
+export type UrlParamsFor<Pathname extends string> =
+  Pathname extends `${string}/:${infer A}?/${infer Tail}`
+    ? Partial<Record<A, string>> & UrlParamsFor<`/${Tail}`>
+    : Pathname extends `${string}/:${infer A}/${infer Tail}`
+      ? Record<A, string> & UrlParamsFor<`/${Tail}`>
+      : Pathname extends `${string}/:${infer A}?`
+        ? Partial<Record<A, string>>
+        : Pathname extends `${string}/:${infer A}`
+          ? Record<A, string>
+          : {};
```
Iku-turso, line 48: Ew. Wasn't there a lib somewhere which externalised icky stuff like this? :)

src/renderer/components/+helm-releases/to-helm-release.injectable.ts
```diff
@@ -0,0 +1,87 @@
...
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { capitalize } from "lodash";
+import helmChartStoreInjectable from "../+helm-charts/store.injectable";
+import type { HelmRelease, HelmReleaseDto } from "../../../common/k8s-api/endpoints/helm-releases.api";
+import { formatDuration } from "../../utils";
+
+export type ToHelmRelease = (release: HelmReleaseDto) => HelmRelease;
+
+const toHelmReleaseInjectable = getInjectable({
```
Iku-turso, line 15: Note: this old mapper code creates a god-object.

src/renderer/port-forward/port-forward-store/port-forward-store.ts
```diff
@@ -9,19 +9,22 @@ import type { StorageLayer } from "../../utils";
...
   notifyErrorPortForwarding: (message: string) => void;
+  readonly apiBase: JsonApi; // TODO: replace with individual dependencies
+  requestActivePortForward: RequestActivePortForward;
 }
 
 export class PortForwardStore extends ItemStore<PortForwardItem> {
   @observable portForwards: PortForwardItem[] = [];
 
-  constructor(private dependencies: Dependencies) {
+  constructor(protected readonly dependencies: Dependencies) {
```
Iku-turso, line 27: Hmm. Why protected?

src/renderer/port-forward/port-forward-store/port-forward-store.ts
```diff
@@ -322,7 +325,7 @@ export class PortForwardStore extends ItemStore<PortForwardItem> {
 
     try {
       // check if the port-forward is active, and if so check if it has the same local port
-      const pf = await getActivePortForward(portForward);
+      const pf = await this.dependencies.requestActivePortForward(portForward);
 
       if (pf?.forwardPort && pf.forwardPort !== portForward.forwardPort) {
```
Iku-turso, line 330: Notice: `pf` and `portForward` are the same reference, and therefore `pf.forwardPort !== portForward.forwardPort` is always `false`.

src/renderer/port-forward/port-forward-store/request-active-port-forward.injectable.ts
```diff
@@ -0,0 +1,42 @@
...
+
+      try {
+        response = await apiBase.get(requestActiveEndpoint.compile(portForward), { query: { port, forwardPort }});
+      } catch (error) {
+        logger.warn(`[PORT-FORWARD-STORE] Error getting active port-forward: ${error}`, portForward);
+
+        return undefined;
+      }
+
+      portForward.status = response?.port ? "Active" : "Disabled";
```
Iku-turso: Generally, mutation of an input parameter is a pretty bad idea, as it's easy to cause bugs that way. There's even a lint for this.

src/common/k8s-api/endpoints/metrics.api/get.injectable.ts
```diff
@@ -0,0 +1,74 @@
...
+  id: "request-metrics",
+  instantiate: (di) => {
+    const apiBase = di.inject(apiBaseInjectionToken);
+
+    return (async (query: object, params: RequestMetricsParams = {}) => {
+      const { range = 3600, step = 60, namespace } = params;
+      let { start, end } = params;
+
+      if (!start && !end) {
+        const timeNow = Date.now() / 1000;
```
Iku-turso: Notice: Side-effect, but explicitly overwritten as global. This is rare, and makes sense mostly to mock a consistent date wherever it's used.

src/renderer/components/resource-metrics/resource-metrics-text.tsx
```diff
@@ -4,22 +4,38 @@
...
+  metrics: ResourceMetricsTextMetrics | null | undefined;
 }
 
 export function ResourceMetricsText({ metrics }: ResourceMetricsTextProps) {
   if (!metrics) {
     return null;
   }
 
-  const { cpuUsage, cpuRequests, cpuLimits, memoryUsage, memoryRequests, memoryLimits } = getMetricLastPoints(metrics);
+  const {
```
Iku-turso, line 31: Are the added default values here to fix a bug or something? :)

src/common/k8s-api/kube-object.ts
```diff
@@ -375,6 +379,12 @@ export type ScopedNamespace<Namespaced extends KubeObjectScope> = (
       : string | undefined
 );
 
+const resourceApplierAnnotationsForFiltering = [
+  "kubectl.kubernetes.io/last-applied-configuration",
```
Iku-turso, line 383: Is this an added functionality?

src/main/routes/resource-applier/apply-resource-route.injectable.ts
```diff
@@ -5,17 +5,19 @@
...
 
-const applyResourceRouteInjectable = getRouteInjectable({
+const createResourceRouteInjectable = getRouteInjectable({
   id: "apply-resource-route",
 
-  instantiate: () => clusterRoute({
+  instantiate: () => payloadValidatedClusterRoute({
     method: "post",
     path: `${apiPrefix}/stack`,
+    payloadValidator: Joi.string(),
```
Iku-turso, line 17: This validator is an added functionality?

src/renderer/components/kubeconfig-dialog/open.injectable.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    return ({ title, loader }) => {
+      (async () => {
+        try {
+          const config = await loader();
+
+          state.set({ title, config });
+        } catch (error) {
+          showCheckedErrorNotification(error, "Failed to retrive config for dialog");
+          logger.warn("[KUBEOCONFIG-DIALOG]: failed to retrived config for dialog", error);
```
Iku-turso, line 33: Orphan promise is not required? Such are pretty difficult to unit-test, and should not be used without a very good cause.

src/common/k8s-api/__tests__/kube-api.test.ts
```diff
@@ -179,10 +155,30 @@ describe("KubeApi", () => {
       checkPreferredVersion: true,
     });
 
+
+    fetchMock.mockImplementation(async (url) => {
```
Iku-turso: Nit: using `asyncFn` here would make test easier to read, as then code would be written in same order as it is ran.

src/renderer/getDiForUnitTesting.tsx
```diff
@@ -160,6 +161,7 @@ export const getDiForUnitTesting = (opts: { doGeneralOverrides?: boolean } = {})
       }));
     });
 
+    di.override(environmentVariablesInjectable, () => ({}));
```
Iku-turso, line 157: Instead of accumulating global overrides here, we can now use global override files.

src/common/k8s-api/create-kube-json-api-for-cluster.injectable.ts
```diff
@@ -0,0 +1,34 @@
...
+  id: "create-kube-json-api-for-cluster",
+  instantiate: (di): CreateKubeJsonApiForCluster => {
+    const apiBase = di.inject(apiBaseInjectionToken);
+    const createKubeJsonApi = di.inject(createKubeJsonApiInjectable);
+
+    return (clusterId) => createKubeJsonApi(
+      {
+        serverAddress: apiBase.config.serverAddress,
+        apiBase: apiKubePrefix,
+        debug: isDebugging,
```
Iku-turso, line 25: Side-effect :)

src/behaviours/helm-releases/showing-details-for-helm-release.test.ts
```diff
@@ -65,25 +81,30 @@ describe("showing details for helm release", () => {
...
       );
 
       rendererDi.override(
-        callForHelmReleaseUpdateInjectable,
-        () => callForHelmReleaseUpdateMock,
+        requestHelmReleaseUpdateInjectable,
+        () => requestHelmReleaseUpdateMock,
       );
 
+      rendererDi.override(requestHelmChartsInjectable, () => requestHelmChartsMock);
```
Iku-turso: These overrides are not used for anything. They should probably be global overrides instead?

src/renderer/components/+helm-releases/to-helm-release.injectable.ts
```diff
@@ -0,0 +1,88 @@
...
+
+      // Helm does not store from what repository the release is installed,
+      // so we have to try to guess it by searching charts
+      async getRepo() {
+        const versionsComputed = helmChartVersions(this);
+
+        await when(() => !versionsComputed.pending.get());
+
+        const version = this.getVersion();
+        const versions = versionsComputed.value.get();
```
Iku-turso: One should use something like `getVersions` instead of `versionsComputed`, as seen here is a non-reactive context, and `versionsComputed` is reactive. Such clash causes complexity, and probably bugs. If there really is need to optimize, `getVersions` can be throttled. I doubt we need that.

src/renderer/components/dock/upgrade-chart/create-upgrade-chart-tab.injectable.ts
```diff
@@ -3,23 +3,27 @@
...
+  upgradeChartStore: DockTabStore<IChartUpgradeData>;
   dockStore: DockStore;
   getRandomId: () => string;
 }
 
 const createUpgradeChartTab = ({ upgradeChartStore, dockStore, getRandomId }: Dependencies) => (release: HelmRelease, tabParams: DockTabCreateSpecific = {}): TabId => {
-  const tabId = upgradeChartStore.getTabIdByRelease(release.getName());
+  const tabId = upgradeChartStore.findTabIdFromData(val => (
+    val.releaseName === release.getName()
+    && val.releaseNamespace === release.getNs()
```
Iku-turso, line 25: Is this more specific find a new functionality?

src/renderer/components/dock/upgrade-chart/upgrade-chart-model.injectable.ts
```diff
@@ -0,0 +1,121 @@
...
+        await updateRelease(
+          release.getName(),
+          release.getNs(),
+          {
+            chart: release.getChart(),
+            values: configration.value.get(),
+            ...version,
+          },
+        );
+        storedConfigration.invalidate();
```
Iku-turso: Typo.

src/renderer/components/dock/upgrade-chart/upgrade-chart-model.injectable.ts
```diff
@@ -0,0 +1,121 @@
...
+    set: (value: SingleValue<SelectOption<ChartVersion>>) => void;
+  };
+  submit: () => Promise<UpgradeChartSubmitResult>;
+}
+
+export interface UpgradeChartSubmitResult {
+  completedSuccessfully: boolean;
+}
+
+const upgradeChartModelInjectable = getInjectable({
```
Iku-turso, line 41: Reminder: asked elsewhere to split this PR starting around this commit.

src/renderer/port-forward/port-forward-store/port-forward-store.ts
Nokel81: This one is easy to answer because in general classes can be extended and those child classes can have dependencies too and it is best practise if they can all use the same `dependencies` field so for dependencies they should be `protected`. This is easier to see in places like `src/renderer/components/dock/install-chart/store.ts`

src/renderer/port-forward/port-forward-store/port-forward-store.ts
Nokel81: No they are not, `pf` is the return value of `requestActivePortForward`, they might be the same type but they aren't the same reference.

src/renderer/components/resource-metrics/resource-metrics-text.tsx
Nokel81: Slightly, the new and better return type of `getMetricLastPoints` is 

```
function getMetricLastPoints<Keys extends string>(metrics: Partial<Record<Keys, MetricData>>): Partial<Record<Keys, number>>
```

which correctly encodes that some of the metrics might not have last values. Typescript then complains that `number | undefined` shouldn't be used with the `>` operator.

src/common/k8s-api/kube-object.ts
Nokel81: No I moved this from the `src/common/k8s-api/endpoints/resource-applier.api.ts` when I deleted that file.

src/main/routes/resource-applier/apply-resource-route.injectable.ts
Nokel81: More like clarifying functionality. The `create-resource-route` really only wants to be used with the create resource dock tabs (but is currently erroneously also used by `KubeObject.update` which is marked as deprecated currently). And since it will only be used for create resources dock tab (the edit resource dock tab uses patch resource instead as it is more robust) there is no need for it accept anything other than manifests (so strings)

src/common/k8s-api/__tests__/kube-api.test.ts
Nokel81: Probably it would be easier to read. I even started trying to use it but the tests are much harder to write because `await fetchMock.resolve()` throws if a call is not currently waiting for it, which is rather annoying.

src/behaviours/helm-releases/showing-details-for-helm-release.test.ts
Nokel81: Sure, but in this specific case I used them in a future commit to fix the tests.

src/renderer/components/dock/upgrade-chart/create-upgrade-chart-tab.injectable.ts
Nokel81: No, this is a refactor (that method has existed for a long time) because I removed the `UpgradeChartDockTabStore` class in favour of just using `DockTabStore<UpgradeChartTabData>` directly.

src/renderer/port-forward/port-forward-store/request-active-port-forward.injectable.ts
Nokel81: I completely agree but I mostly just moved this code around

src/renderer/k8s/api-base.injectable.ts
Nokel81: Because it is essentially a wrapper of `fetch`?

src/renderer/k8s/api-base.injectable.ts
Nokel81: `apiPrefix` is not a side effect but will handle the others.

src/common/k8s-api/api-base.ts
Nokel81: And I have now removed it.

src/common/k8s-api/endpoints/helm-releases.api/get-configuration.injectable.ts
Nokel81: Done

src/common/utils/buildUrl.ts
Nokel81: I think I remember that there was, but I don't know if it is worth it for just this type.

src/renderer/port-forward/port-forward-store/port-forward-store.ts
Nokel81: But I completely agree with you that we should never modify the inputs as that is very surprising.

src/renderer/components/kubeconfig-dialog/open.injectable.ts
Nokel81: What do you mean by "not required"? This is more explicit in the fact that the loader has to run before the dialog is open since previously the loader was run in the component itself.

src/renderer/components/+helm-releases/to-helm-release.injectable.ts
Nokel81: Okay will introduce a getVersions too

src/common/k8s-api/endpoints/helm-charts.api/get-readme.injectable.ts
Nokel81: Unless what we really wanted was to export these to the extension API. I think that they should be in common because they are implemented in a way that anyone could use them.

src/renderer/k8s/api-base.injectable.ts
Iku-turso: Something like that. Idea resolved anyway!

src/common/k8s-api/endpoints/helm-charts.api/get-readme.injectable.ts
Iku-turso: That would be YAGNI I think :) Note resolved.

src/common/utils/buildUrl.ts
Iku-turso: Sure. Maybe let's not go and duplicate something like this in the future :) Resolved.

src/renderer/port-forward/port-forward-store/port-forward-store.ts
Iku-turso: But would it not be a better policy to prefer encapsulation over visibility for speculation? To make properties more visible means a curious coder will try to find out why something is protected, find nothing, and be perplexed.

Here I see no reason to not wait for the first inheritor to require protected dependencies, and only then adjust.

Interesting conversation, but pretty trivial here. I suspect somewhere else it might not be :) Resolved.

src/renderer/components/resource-metrics/resource-metrics-text.tsx
Iku-turso: Ah. Maybe the commit could communicate something like this in the future? Resolved :)

src/common/k8s-api/kube-object.ts
Iku-turso: Ok. Resolved.

src/main/routes/resource-applier/apply-resource-route.injectable.ts
Iku-turso: It's pretty hard for me to be sure about this. Is the change something that should be accompanied with an explanatory test for the requirement? Seems to not belong in this PR anyway :/

Nokel81: I can change it with a cast if you would like, or do you mean the change of how the route is interpreted? It is a POST after all.

src/common/k8s-api/create-kube-api-for-remote-cluster.injectable.ts
```diff
@@ -0,0 +1,104 @@
...
+  instantiate: (di): CreateKubeApiForRemoteCluster => {
+    const isDevelopment = di.inject(isDevelopmentInjectable);
+    const createKubeJsonApi = di.inject(createKubeJsonApiInjectable);
+
+    return (config: CreateKubeApiForRemoteClusterConfig, kubeClass: KubeObjectConstructor<KubeObject, KubeJsonApiDataFor<KubeObject>>, apiClass = KubeApi) => {
+      const reqInit: RequestInit = {};
+      const agentOptions: AgentOptions = {};
+
+      if (config.cluster.skipTLSVerify === true) {
+        agentOptions.rejectUnauthorized = false;
```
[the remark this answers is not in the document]
Iku-turso: You know what this is? Sounds serious.

src/renderer/components/kubeconfig-dialog/open.injectable.ts
Iku-turso: I mean avoidable. To avoid orphan promises whenever possible has served pretty well thus far :)

Iku-turso: Resolved for being pretty trivial here.

src/common/k8s-api/__tests__/kube-api.test.ts
Iku-turso: This we talked and studied somewhere else with team? Resolved :)

src/renderer/getDiForUnitTesting.tsx
Iku-turso: Resolved, this stuff will sort out eventually, bit by bit.

src/renderer/components/dock/upgrade-chart/create-upgrade-chart-tab.injectable.ts
Iku-turso: Ok, resolved.

src/common/k8s-api/__tests__/kube-api.test.ts
```diff
@@ -104,167 +134,434 @@ describe("KubeApi", () => {
...
+        });
 
-    await kubeApi.get({
-      name: "foo",
-      namespace: "default",
-    });
+        it("sets fields in the api instance", () => {
+          expect(ingressApi.apiVersionPreferred).toBe("v1");
+          expect(ingressApi.apiPrefix).toBe("/apis");
+          expect(ingressApi.apiGroup).toBe("networking.k8s.io");
```
Iku-turso: Try some time `expect(ingressApi).toEqual(expect.objectContaining({ apiVersionPreferred: "v1", ... }));` for just one expect, and a better error report with a failing test :)

src/renderer/components/+helm-releases/to-helm-release.injectable.ts
Iku-turso: Resolved for being a note.

src/main/routes/resource-applier/apply-resource-route.injectable.ts
Iku-turso: It looks like introduction of an untested run-time validator for something irrelevant to the PR. I'm pushing for keeping PR's small and true to description :) And the usual unit testing.

Let's roll with this for now, if you are dead-sure about this? :) Resolve on will.

src/common/k8s-api/api-base.ts
```diff
@@ -3,38 +3,9 @@
...
+import { getInjectionToken } from "@ogre-tools/injectable";
 
-export let apiBase: JsonApi;
-
-if (typeof window === "undefined") {
-  appEventBus.addListener((event) => {
-    if (event.name !== "lens-proxy" && event.action !== "listen") return;
-
-    const params = event.params as { port?: number };
-
```
Iku-turso, line 17: Ah! So nice to get rid of this! :)

src/common/k8s-api/endpoints/helm-releases.api/get-configuration.global-override-for-injectable.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import { getGlobalOverride } from "../../../test-utils/get-global-override";
+import requestHelmReleaseConfigurationInjectable from "./get-configuration.injectable";
+
+export default getGlobalOverride(requestHelmReleaseConfigurationInjectable, () => () => {
```
Iku-turso: Check out: there's soon something called `getGlobalOverrideForFunction` to reduce boilerplate for stuff like this :)

src/common/utils/date/get-milliseconds-from-unix-epoch.injectable.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+
+const getMillisecondsFromUnixEpochInjectable = getInjectable({
+  id: "get-current-time",
+  instantiate: () => () => Date.now(),
```
Iku-turso: Hmm. There's a conversation to be had on how we want to handle the side-effect of JavasScript's Date global so that it is consistent, even in 3rd party libs.

src/renderer/components/+helm-charts/helm-charts/versions.injectable.ts
```diff
@@ -0,0 +1,29 @@
...
+import requestVersionsOfHelmChartInjectable from "./request-versions-of-chart-for-release.injectable";
+
+const helmChartVersionsInjectable = getInjectable({
+  id: "helm-chart-versions-loader",
+  instantiate: (di, release) => {
+    const helmCharts = di.inject(helmChartsInjectable);
+    const requestVersionsOfHelmChart = di.inject(requestVersionsOfHelmChartInjectable);
+
+    return asyncComputed(async () => {
+      await when(() => !helmCharts.pending.get());
```
Iku-turso, line 19: FYI: we are thinking of more elegant way to compose `asyncComputed`s from other `asyncComputed`s.

src/common/k8s-api/kube-api.ts
```diff
@@ -370,7 +239,7 @@ export class KubeApi<
   constructor(opts: KubeApiOptions<Object, Data>) {
     const {
       objectConstructor,
-      request = apiKube,
+      request = asLegacyGlobalForExtensionApi(apiKubeInjectionToken),
```
Iku-turso, line 261: Venting: This line is as it was before, kinda, but still it screams considerable design problems. This `KubeApi` will need some love pretty soon :/

src/common/k8s-api/__tests__/kube-api.test.ts
```diff
@@ -2,36 +2,42 @@
  * Copyright (c) OpenLens Authors. All rights reserved.
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
-
-import { forRemoteCluster, KubeApi } from "../kube-api";
-import { KubeJsonApi } from "../kube-json-api";
-import { KubeObject } from "../kube-object";
-import { delay } from "../../utils/delay";
+import type { KubeApiWatchCallback } from "../kube-api";
```
Iku-turso, line 5: Nice commit! :)

src/renderer/components/test-utils/get-application-builder.tsx
```diff
@@ -240,10 +247,22 @@ export const getApplicationBuilder = () => {
...
   let applicationHasStarted = false;
+  let currentTimeMs = Date.parse("2000-01-01 12:00:00am");
+
+  mainDi.override(getMillisecondsFromUnixEpochInjectable, () => () => currentTimeMs);
 
   const builder: ApplicationBuilder = {
     mainDi,
-
+    time: {
+      advanceBy: {
```
Iku-turso: Hmm. This is a way to advance one fake time, but it only applies to our own code. We should use `jest.advanceFakeTime` instead for consistent behaviour, even in 3rd party libraries.

src/common/k8s-api/kube-api.ts
```diff
@@ -694,22 +591,29 @@ export class KubeApi<
...
-    const { watchId = `${this.kind.toLowerCase()}-${this.watchId++}` } = opts;
+    const {
+      namespace,
+      callback = noop as KubeApiWatchCallback<Data>,
+      retry = false,
+      timeout,
+      watchId = `${this.kind.toLowerCase()}-${this.watchId++}`,
+    } = opts ?? {};
+
+    console.log(`watching ${this.apiBase} in namespace="${namespace}"`);
```
Iku-turso: Forgotten debug log?

src/common/k8s-api/create-json-api.injectable.ts
```diff
@@ -0,0 +1,26 @@
...
+import type { JsonApiConfig, JsonApiData, JsonApiDependencies, JsonApiParams } from "./json-api";
+import { JsonApi } from "./json-api";
+
+export type CreateJsonApi = <Data = JsonApiData, Params extends JsonApiParams<Data> = JsonApiParams<Data>>(config: JsonApiConfig, reqInit?: RequestInit) => JsonApi<Data, Params>;
+
+const createJsonApiInjectable = getInjectable({
+  id: "create-json-api",
+  instantiate: (di): CreateJsonApi => {
+    const deps: JsonApiDependencies = {
+      fetch: di.inject(fetchInjectable),
```
Iku-turso, line 18: Note: At this point, `fetch` should be hidden behind an abstraction, or else a volatile, complex and uncontrollable model from an external library gets coupled with our internal domain. Basically this breaks Dependency Inversion Principle.

Iku-turso: No action now, but expect change soon :) Resolve at will.

src/common/vars/is-debugging.injectable.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+
+const isDebuggingInjectable = getInjectable({
+  id: "is-debugging",
+  instantiate: () => ["true", "1", "yes", "y", "on"].includes((process.env.DEBUG ?? "").toLowerCase()),
+  causesSideEffects: true,
```
Iku-turso, line 10: Trivial: `environmentVariablesInjectable` should cause side-effects, not this injectable. It also exposes in typing what variables are interesting for Lens, on top of being the single source of truth for that.

src/renderer/components/dock/upgrade-chart/upgrade-chart-model.injectable.ts
```diff
@@ -0,0 +1,123 @@
...
+import releasesInjectable from "../../+helm-releases/releases.injectable";
+import updateReleaseInjectable from "../../+helm-releases/update-release/update-release.injectable";
+import type { HelmRelease } from "../../../../common/k8s-api/endpoints/helm-releases.api";
+import requestHelmReleaseConfigurationInjectable from "../../../../common/k8s-api/endpoints/helm-releases.api/request-configuration.injectable";
+import { waitUntilDefined } from "../../../utils";
+import type { SelectOption } from "../../select";
+import type { DockTab } from "../dock/store";
+import upgradeChartTabStoreInjectable from "./store.injectable";
+
+export interface UpgradeChartModel {
```
Iku-turso, line 21: Uh. Is this model new code? ViewModels are nice when they are mere facades, but this is a god-class, full of different responsibilities.

Iku-turso (changes requested):
The unit tests with asyncFn are nice! Also the individual abstractions for different requests (or did this exist before? No?).

I think the handling of fake time should be put to right track, and `upgradeChartModelInjectable` should be tidied.

Iku-turso (dismissed):

src/common/k8s-api/create-json-api.injectable.ts
Nokel81: That is what json api etc is for...

Nokel81: How is it volatile? There is a standard?

And claiming that something is complex just means that it does a lot. Fetch IS the abstraction over communicated with the internet.

Iku-turso (dismissed):

Iku-turso (dismissed):

src/common/k8s-api/kube-api.ts
Nokel81: Yes I agree, but because it is an exported type we have to keep the constructor the same

Nokel81: Though this line does change the initialisation order requirements.

Iku-turso (dismissed):

PR #6029, opened 2022-08-12
Fix badge not expanding on click when appropriate

Nokel81:

Before (master):

https://user-images.githubusercontent.com/8225332/184399113-dde039d7-cf3f-4f73-abc4-42f1ab473b3e.mov

---

After:

https://user-images.githubusercontent.com/8225332/184399152-7b13d89d-f555-4039-af96-78915291afe4.mov

### DIFF
--- a/src/renderer/components/badge/badge.tsx
+++ b/src/renderer/components/badge/badge.tsx
@@ -5,7 +5,7 @@
 
 import styles from "./badge.module.scss";
 
-import React, { createRef, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import { action, observable } from "mobx";
 import { observer } from "mobx-react";
 import { cssNames } from "../../utils/cssNames";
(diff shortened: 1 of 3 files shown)


src/renderer/components/badge/badge.tsx
```diff
@@ -40,12 +40,26 @@ export const Badge = withTooltip(observer(({
...
-    ? elem.current.clientWidth < elem.current.scrollWidth
-    : false;
+  useEffect(() => {
+    if (elemHasBeenSet.current) {
+      return;
+    }
+
+    if (elem.current) {
+      elemHasBeenSet.current = true;
+    }
```
Nokel81: This check is here because we only want to compute if a badge is expandable once (when it is closed). Otherwise it will be re-computed when it is clicked and then expanded.

src/renderer/components/badge/badge.tsx
```diff
@@ -40,12 +40,26 @@ export const Badge = withTooltip(observer(({
...
+
+    setIsExpandable((
```
Nokel81: This has to be state because we need to trigger a re-render if it is now `true` (to add the class names)

src/renderer/components/badge/badge.tsx
```diff
@@ -40,12 +40,24 @@ export const Badge = withTooltip(observer(({
...
-    ? elem.current.clientWidth < elem.current.scrollWidth
-    : false;
+  useEffect(() => {
+    if (elemHasBeenSet.current) {
+      return;
+    }
+
+    if (elem.current) {
+      elemHasBeenSet.current = true;
+    }
```
aleksfront: Seems to work absolutely fine without these lines. Probably there's no need for this checks.

aleksfront (commented):
Need to clean up unused code a bit. Otherwise, looks good.

aleksfront (approved):

jweak (approved):
LGTM

Issue #6031, opened 2022-08-14
lens spins infinitely viewing certain helm releases


mipearson:
**Describe the bug**
Clicking on the details of a helm release causes lens to show the pane and a "loading" spinner infinitely.

Issue is present in 6.0.0 & 6.0.1, but not 5.x.x.

This doesn't occur for all helm releases: it seems specific to helm releases from our internal charts. eg, opening the helm release from a community datadog chart works fine.

The error log below reports that it can't load certain resources, but I can load them just fine with lens if I track down the resource manually (confirmed by opening up `guru-web-app-staging-reverse-proxy-configmap` and viewing its contents). Suspect a namespace issue of some kind ..

The RBAC role in use has no restrictions around secrets, resources, etc.

**To Reproduce**

1. Click on "Helm -> Releases"
2. Click on a release
3. Wait

**Expected behavior**

Helm release information is shown

**Screenshots**

![image](https://user-images.githubusercontent.com/97317/184527133-e62a6f6a-9712-4db9-b4bf-22e2a78b6b9f.png)

**Environment (please complete the following information):**
- Lens Version: OpenLens (6.0.1-latest.1660302598735)
- OS: OSX 12.3.1 ARM
- Installation method: .dmg download from https://github.com/MuhammedKalkan/OpenLens/releases/tag/v6.0.1

Note that I observed this behaviour as well with the 6.0.0 binary downloaded via k8slens.dev.

**Logs:**
```
BrowserConsole.js:41 [KUBE-API] watch (secret-1) started /api/v1/secrets?watch=1&resourceVersion=140771938
[...]

PR #6039, opened 2022-08-16
Don't error out if we can't find resources from helm in our cluster, and handle "blank" output from kubectl gracefully.

mipearson:
Temporary workaround for #6031, as supplying `--namespace` on the kubectl command line breaks helm releases that use multiple namespaces.


aleksfront (approved):
Works great with missing resources.

Nokel81 (approved):

PR #6043, opened 2022-08-17
Remove mac-ca usage since it was only in tests

Nokel81:
- Fully make injecting CAs injectable

resolves https://github.com/lensapp/lens/security/dependabot/45


aleksfront (dismissed):

jansav (dismissed):
I don't see how PR name matches the implementation. Also this test seems to be something else than unit test? 

However, the change here is minimal, so approving.

Nokel81:
@jansav I had other stuff here... will try and investigate where it went

Nokel81:
@jansav PTALA, I had to recreate the change oddly enough.

src/common/certificate-authorities/inject-system-cas-on-macos.injectable.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    const execSecurity = async (...args: string[]) => {
+      const output = await execFile("/usr/bin/security", args);
+
+      return output.split(certSplitPattern);
+    };
+
+    if (!isMac) {
+      return async () => {};
+    }
```
Iku-turso: Musings for future: Hmm. That one idea about being able to have eg. `system-cas.injectable.mac.ts` and `system-cas.injectable.win.ts` instead of code like this could be pretty good.

src/common/certificate-authorities/inject-system-cas.injectable.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import { getInjectable } from "@ogre-tools/injectable";
+import { systemCAsInjectorInjectionToken } from "./system-cas-injector-token";
+
+const injectSystemCAsInjectable = getInjectable({
```
Iku-turso, line 19: Trivial: This level of abstraction is the correct place for responsibility of filtering ca's from wrong environments, not the ca-injectables itself (by producing empty arrays like now).

This is trivial here, but clear responsibilities of modules will carry us far in the future.

src/common/certificate-authorities/inject-system-cas.injectable.ts
```diff
@@ -0,0 +1,23 @@
...
+  id: "inject-system-cas",
+  instantiate: (di) => {
+    const systemCAsInjectors = di.injectMany(systemCAsInjectorInjectionToken);
+
+    return async () => {
+      for (const systemCAsInjector of systemCAsInjectors) {
```
Iku-turso: Could be `Promise.all()`? Not sure about gains for application start-up, but could be totally worth the optimisation.

src/common/certificate-authorities/system-ca.test.ts
```diff
@@ -0,0 +1,63 @@
...
+import https from "https";
+import { DSTRootCAX3 } from "./is-cert-active";
+import assert from "assert";
+import { describeIf } from "../../test-utils/skippers";
+import { platform } from "process";
+import { getDiForUnitTesting } from "../../main/getDiForUnitTesting";
+import injectSystemCAsOnMacOSInjectable from "./inject-system-cas-on-macos.injectable";
+import injectSystemCAsOnWindowsInjectable from "./inject-system-cas-on-windows.injectable";
+
+describe("CA tests", () => {
```
Iku-turso: This integration test only covers some odd corner-cases, and not the happy-path of tested module.

src/common/certificate-authorities/system-ca.test.ts
```diff
@@ -0,0 +1,63 @@
...
+  let injectSystemCAsOnMacOS: () => Promise<void>;
```
Iku-turso: New integration test is set to run with unit tests, making the unit testing suite slower.

src/common/certificate-authorities/system-ca.test.ts
```diff
@@ -0,0 +1,63 @@
...
+  let injectSystemCAsOnWindows: () => Promise<void>;
```
Iku-turso: 1. Would appreciate behavioural unit tests for modernisations such as in this PR. After all, this is the cheapest place to produce them, as now you, the dev, know a lot of all the moving parts being migrated to new injectables.

2. Maybe the said modernisation is even a more meaningful title for this PR :)

3. Remember: just integration testing this shouldn't be enough, as it serves other purpose.

Iku-turso (changes requested):
I think a modernisation such as this (to me the PR name is a bit misleading) should have unit tests. Maybe we could agree on a policy on a team level? :)

The tricky part is that this policy cannot demoralise modernisations, but instead make the code easier to write now, and to maintain in the future.

src/common/certificate-authorities/inject-system-cas-on-macos.injectable.ts
Nokel81: All the yes!!! I was thinking about using folders but that is probably a better idea. Should I do that in this PR?

src/common/certificate-authorities/inject-system-cas.injectable.ts
Nokel81: I think if we moved to having `.injectable.<platform>.ts` then this would change to be `di.inject(systemCAsInjectorInjectionToken)` and not be "many" at all.

src/common/certificate-authorities/system-ca.test.ts
Nokel81: These tests actually existed previously, they just tested to make sure that we had the same behaviour as a dependency.

PR #6046, opened 2022-08-17
Upgrade to latest node-fetch

Nokel81:
- Introduce injection tokens for retriving the implementation of
  node-fetch via await import() calls

- Add webpack file for compiling node-fetch to electron-renderer single
  file format


src/common/fetch/download-json.injectable.ts
```diff
@@ -4,15 +4,14 @@
...
-import type { JsonValue } from "type-fest";
 import type { AsyncResult } from "../utils/async-result";
 import fetchInjectable from "./fetch.injectable";
 
 export interface DownloadJsonOptions {
   signal?: AbortSignal | null | undefined;
 }
 
-export type DownloadJson = (url: string, opts?: DownloadJsonOptions) => Promise<AsyncResult<JsonValue, string>>;
+export type DownloadJson = (url: string, opts?: DownloadJsonOptions) => Promise<AsyncResult<unknown, string>>;
```
jakolehm, line 14: Why `unknown` instead of `JsonValue`?

src/common/k8s-api/kube-api.ts
```diff
@@ -643,11 +643,12 @@ export class KubeApi<
       clearTimeout(timedRetry);
     });
 
+    setTimeoutFor(abortController, 60 * 1000);
```
jakolehm: Could you explain how this matches the previous behaviour.. it is a bit unclear to me (we had 10x timeout before).

src/common/k8s-api/kube-api.ts
```diff
@@ -686,7 +687,19 @@ export class KubeApi<
           }, timeout * 1000 * 1.1);
         }
 
-        ["end", "close", "error"].forEach((eventName) => {
+        if (!response.body) {
```
jakolehm, line 690: Does this fix some case?

webpack/node-fetch.ts
```diff
@@ -0,0 +1,44 @@
...
+    asyncChunks: false, // This is required so that only one file is created
+  },
+  mode: "production",
+  target: "electron-renderer",
+  optimization: {
+    concatenateModules: true,
+    minimize: true,
+  },
+  externals: {
+    "node:buffer": "commonjs node:buffer",
```
jakolehm: Why these specific externals? Could we just use node externals here (meaning everything)?

src/common/fetch/download-json.injectable.ts
Nokel81: Because due to typescript limitations `JsonValue` is actually a rather useless type. Like `unknown` you have to constantly runtime check (or cast) the fields you are dealing with. Furthermore, if you just don't care about the type safety you have to cast to `unknown` first before casting to some other interface because `tsc` complains about "no overlap".

src/common/k8s-api/kube-api.ts
Nokel81: Yes both because the types of `body` now correctly states that it is `| null` (so the check is required by the compiler) and because the body sometimes isn't there and we shouldn't blow up.

src/common/k8s-api/kube-api.ts
Nokel81: Good catch, it should be the same.

webpack/node-fetch.ts
Nokel81: Let me try again, it has been too long since I initially did this so I forget why that wasn't enough.

jakolehm (approved):

PR #6085, opened 2022-08-23
Fix `<Select/>` text overflow below baseline

aleksfront:
Some of the characters inside `<Select/>` value container get cropped for couple of pixels below the baseline. Example - notice the `g` and `y` tails.
selector crop 1
selector crop 2

Fix adding `overflow: visible` to such values to prevent this.

selector no crop

In addition, 2 tiny fixes for the namespace selector and `Working directory` input within cluster settings


jweak (approved):
LGTM

Nokel81 (approved):

PR #6088, opened 2022-08-23
Make opening of release details work properly when release has resources without namespace

### LINKED ISSUE #6031: lens spins infinitely viewing certain helm releases

jansav:
Fixes #6031

Review commit by commit. Please verify with multiple charts and installing them to separate namespaces etc.

Problem:

The issue comes from place where we get resources of a Helm release.

Before this PR, the process of getting resources of a Helm release works like this:
1.  Get manifest of a Helm release e.g. `$ helm get manifest chatwoot-1660899467 --namespace some-namespace --kubeconfig /var/folders/_3/some-kubeconfig`
2. With response of first command, we call `kubectl` to get kube resources. e.g. `$ kubectl get --kubeconfig /var/folders/_3/some-kubeconfig --output json -f -` (notice output of first command as input)

However, latter command doesn't work properly:

1. In case Helm Release has resources in single namespace, it doesn't matter if we specify `--namespace`
2. In case Helm Release has resources in single namespace and resources without namespace, we have to specify `--namespace` to get all resources (it will crash without it, because it cannot find the resource without namespace)
3. In case Helm Release has resources in multiple namespaces, we shouldn't specify `--namespace` (it will crash if namespace doesn't match)

This means that to get all of these combinations to work properly, we have to:
1. Helm manifest of Helm release
2. Split resources from manifest to groups with namespace and default to whatever namespace was set in the 1. command.
3. Get resources from `kubectl` for each group individually and specify the `--namespace`.

---

Also fixed the CSS for the resources tables so that everything lines up even without namespaces and that the names of the resources get the most space:

Screen Shot 2022-09-08 at 8 24 14 AM

### DIFF
--- a/src/main/helm/exec-helm/exec-helm.injectable.ts
+++ b/src/main/helm/exec-helm/exec-helm.injectable.ts
@@ -8,14 +8,16 @@ import helmBinaryPathInjectable from "../helm-binary-path.injectable";
 import type { AsyncResult } from "../../../common/utils/async-result";
 import { getErrorMessage } from "../../../common/utils/get-error-message";
 
+export type ExecHelm = (...args: string[]) => Promise<AsyncResult<string>>;
+
 const execHelmInjectable = getInjectable({
   id: "exec-helm",
 
-  instantiate: (di) => {
+  instantiate: (di): ExecHelm => {
     const execFile = di.inject(execFileInjectable);
     const helmBinaryPath = di.inject(helmBinaryPathInjectable);
 
-    return async (...args: string[]): Promise<AsyncResult<string>> => {
+    return async (...args) => {
       try {
         const response = await execFile(helmBinaryPath, args);
 
--- a/src/main/helm/exec-helm/exec-helm.injectable.ts
+++ b/src/main/helm/exec-helm/exec-helm.injectable.ts
@@ -8,7 +8,7 @@ import helmBinaryPathInjectable from "../helm-binary-path.injectable";
 import type { AsyncResult } from "../../../common/utils/async-result";
 import { getErrorMessage } from "../../../common/utils/get-error-message";
 
-export type ExecHelm = (...args: string[]) => Promise<AsyncResult<string>>;
+export type ExecHelm = (args: string[]) => Promise<AsyncResult<string>>;
 
 const execHelmInjectable = getInjectable({
   id: "exec-helm",
@@ -17,9 +17,11 @@ const execHelmInjectable = getInjectable({
     const execFile = di.inject(execFileInjectable);
     const helmBinaryPath = di.inject(helmBinaryPathInjectable);
 
-    return async (...args) => {
+    return async (args) => {
       try {
-        const response = await execFile(helmBinaryPath, args);
+        const response = await execFile(helmBinaryPath, args, {
+          maxBuffer: 32 * 1024 * 1024 * 1024, // 32 MiB
+        });
 
         return { callWasSuccessful: true, response };
       } catch (error) {
--- a/src/main/helm/helm-service/get-helm-release-resources/call-for-helm-manifest/call-for-helm-manifest.injectable.ts
+++ b/src/main/helm/helm-service/get-helm-release-resources/call-for-helm-manifest/call-for-helm-manifest.injectable.ts
@@ -6,12 +6,7 @@ import { getInjectable } from "@ogre-tools/injectable";
 import type { AsyncResult } from "../../../../../common/utils/async-result";
 import execHelmInjectable from "../../../exec-helm/exec-helm.injectable";
 import yaml from "js-yaml";
-
-export interface HelmResourceManifest {
-  metadata: {
-    namespace: string;
-  };
-}
+import type { KubeJsonApiData } from "../../../../../common/k8s-api/kube-json-api";
 
 const callForHelmManifestInjectable = getInjectable({
   id: "call-for-helm-manifest",
@@ -23,7 +18,7 @@ const callForHelmManifestInjectable = getInjectable({
       name: string,
       namespace: string,
       kubeconfigPath: string,
-    ): Promise<AsyncResult<HelmResourceManifest[]>> => {
+    ): Promise<AsyncResult<KubeJsonApiData[]>> => {
       const result = await execHelm(
         "get",
         "manifest",
@@ -42,7 +37,7 @@ const callForHelmManifestInjectable = getInjectable({
         callWasSuccessful: true,
         response: yaml
           .loadAll(result.response)
-          .filter((manifest) => !!manifest) as HelmResourceManifest[],
+          .filter((manifest) => !!manifest) as KubeJsonApiData[],
       };
     };
 
--- a/src/main/helm/helm-service/get-helm-release-resources/call-for-helm-manifest/call-for-helm-manifest.injectable.ts
+++ b/src/main/helm/helm-service/get-helm-release-resources/call-for-helm-manifest/call-for-helm-manifest.injectable.ts
@@ -19,7 +19,7 @@ const callForHelmManifestInjectable = getInjectable({
       namespace: string,
       kubeconfigPath: string,
     ): Promise<AsyncResult<KubeJsonApiData[]>> => {
-      const result = await execHelm(
+      const result = await execHelm([
         "get",
         "manifest",
         name,
@@ -27,7 +27,7 @@ const callForHelmManifestInjectable = getInjectable({
         namespace,
         "--kubeconfig",
         kubeconfigPath,
-      );
+      ]);
 
       if (!result.callWasSuccessful) {
         return { callWasSuccessful: false, error: result.error };
--- a/src/main/helm/helm-service/get-helm-release-resources/call-for-helm-manifest/call-for-helm-manifest.injectable.ts
+++ b/src/main/helm/helm-service/get-helm-release-resources/call-for-helm-manifest/call-for-helm-manifest.injectable.ts
@@ -1,0 +1,52 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import type { AsyncResult } from "../../../../../common/utils/async-result";
+import execHelmInjectable from "../../../exec-helm/exec-helm.injectable";
+import yaml from "js-yaml";
+
+export interface HelmResourceManifest {
+  metadata: {
+    namespace: string;
+  };
+}
+
+const callForHelmManifestInjectable = getInjectable({
+  id: "call-for-helm-manifest",
+
+  instantiate: (di) => {
+    const execHelm = di.inject(execHelmInjectable);
+
+    return async (
+      name: string,
+      namespace: string,
+      kubeconfigPath: string,
+    ): Promise<AsyncResult<HelmResourceManifest[]>> => {
+      const result = await execHelm(
+        "get",
+        "manifest",
+        name,
+        "--namespace",
+        namespace,
+        "--kubeconfig",
+        kubeconfigPath,
+      );
+
+      if (!result.callWasSuccessful) {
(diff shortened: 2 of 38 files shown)


src/renderer/components/+helm-releases/release-details/release-details-content.tsx
```diff
@@ -36,96 +35,85 @@ interface Dependencies {
 
 const NonInjectedReleaseDetailsContent = observer(({ model }: Dependencies & ReleaseDetailsContentProps) => {
   const isLoading = model.isLoading.get();
+  const failedToLoad = model.failedToLoad.get();
+
+  if (failedToLoad) {
+    return <div data-testid="helm-release-detail-error">Failed to load release</div>;
+  }
```
Nokel81: Shouldn't we display the load error?

src/renderer/components/+helm-releases/release-details/release-details-content.tsx
```diff
@@ -36,96 +35,85 @@ interface Dependencies {
...
-          <DrawerItem
-            name="Status"
-            className="status"
-            labelsOnly>
-            <Badge
-              label={model.release.getStatus()}
-              className={kebabCase(model.release.getStatus())}
+    (isLoading ? (
+      <Spinner center data-testid="helm-release-detail-content-spinner" />
+    ) : (
```
Nokel81: Isn't the point of `getPlaceholder` so that we don't have to have these sorts of things?

src/renderer/components/+helm-releases/release-details/release-details-model/release-details-model.injectable.tsx
```diff
@@ -92,14 +82,13 @@ export interface ConfigurationInput {
...
   targetRelease: TargetHelmRelease;
-  themeStore: ThemeStore;
   callForHelmReleaseConfiguration: CallForHelmReleaseConfiguration;
   getResourceDetailsUrl: GetResourceDetailsUrl;
   updateRelease: CallForHelmReleaseUpdate;
   showCheckedErrorNotification: ShowCheckedErrorNotification;
   showSuccessNotification: ShowNotification;
   createUpgradeChartTab: (release: HelmRelease) => string;
-  navigateToHelmReleases: NavigateToHelmReleases;
+  closeDrawer: () => void;
```
Nokel81: Shouldn't the dependency name be "correct" and then the exposed name be "closeDrawer"?

src/main/helm/helm-service/get-helm-release-resources/call-for-helm-manifest/call-for-helm-manifest.injectable.ts
```diff
@@ -0,0 +1,52 @@
...
+import execHelmInjectable from "../../../exec-helm/exec-helm.injectable";
+import yaml from "js-yaml";
+
+export interface HelmResourceManifest {
+  metadata: {
+    namespace: string;
+  };
+}
+
+const callForHelmManifestInjectable = getInjectable({
```
Nokel81, line 11: nit: most of the rest of these sorts of injectables are prefixed with `request` now, given that you *have* to call a function to use it anyway, using the name "callFor" seems rather redundant.

src/main/helm/helm-service/get-helm-release-resources/call-for-kube-resources-by-manifest/call-for-kube-resources-by-manifest.injectable.ts
```diff
@@ -0,0 +1,73 @@
...
+      kubectlPath,
+      resourceManifests,
+    ) => {
+      const input = pipeline(
+        resourceManifests,
+        map((manifest) => yaml.dump(manifest)),
+        wideJoin("---\n"),
+      );
+
+      const result = await execFileWithInput({
```
Nokel81, line 40: I guess this is what would be simplified if we just used the kube api directly, since we could do (only on renderer within iframe at the moment)...

src/main/helm/helm-service/get-helm-release-resources/get-helm-release-resources.injectable.ts
```diff
@@ -0,0 +1,52 @@
...
+      const result = await callForHelmManifest(name, namespace, kubeconfigPath);
+
+      if (!result.callWasSuccessful) {
+        throw new Error(result.error);
+      }
+
+      const results = await pipeline(
+        result.response,
+
+        groupBy((item) => item.metadata.namespace || namespace),
```
Nokel81, line 36: Do we need to group cluster scoped resources with a namespace?

src/renderer/components/+helm-releases/release-details/release-details-content.tsx
jansav: Good catch. This was actually just leftover code that was supposed to be removed. Fixed.

src/main/helm/helm-service/get-helm-release-resources/call-for-kube-resources-by-manifest/call-for-kube-resources-by-manifest.injectable.ts
Nokel81: Let us leave this as is and we can switch in the future if we want to.

src/main/helm/helm-service/get-helm-release-resources/get-helm-release-resources.injectable.ts
Nokel81: Probably not but it is easier if we do not.

Nokel81 (dismissed):

jim-docker:
I'm seeing this, which doesn't happen on 6.0.1-latest.

https://user-images.githubusercontent.com/40840436/188932440-f7c4bbaf-161c-4a63-824c-83dd362703fe.mov

happens for `kube-prometheus-stack` release but not others I tried

Nokel81:
Thanks I will investigate.

jim-docker:
Installed Rook Ceph, version 1.8.3, and got:

https://user-images.githubusercontent.com/40840436/188957872-10fbed1d-2880-4d9a-a3a6-13a6c360b175.mov

same error after upgrade to Ceph 1.10.0, so version may be irrelevant

Nokel81:
@jim-docker Can you try again now?

src/main/helm/exec-helm/exec-helm.injectable.ts
```diff
@@ -8,7 +8,7 @@ import helmBinaryPathInjectable from "../helm-binary-path.injectable";
 import type { AsyncResult } from "../../../common/utils/async-result";
 import { getErrorMessage } from "../../../common/utils/get-error-message";
 
-export type ExecHelm = (...args: string[]) => Promise<AsyncResult<string>>;
+export type ExecHelm = (args: string[]) => Promise<AsyncResult<string>>;
```
jim-docker, line 11: this seems like a ride-along change

jim-docker (commented):
I still see `Error: Exited without exit code` for `ceph`, and now for `kube-prometheus-stack`. Maybe this only happens with some pre-existing Helm releases?

src/main/helm/helm-service/get-helm-release-resources/call-for-kube-resources-by-manifest/exec-file-with-input/exec-file-with-input.injectable.ts
```diff
@@ -26,7 +26,9 @@ const execFileWithInputInjectable = getInjectable({
         let execution: ChildProcess;
 
         try {
-          execution = execFile(filePath, commandArguments);
+          execution = execFile(filePath, commandArguments, {
+            maxBuffer: 8 * 1024 * 1024 * 1024, // 8 MiB
```
jim-docker, line 30: how do we know this is enough?

jim-docker (approved):

jim-docker (approved):
I think this should be for 6.1 unless we think it is really critical.

src/main/helm/helm-service/get-helm-release-resources/call-for-kube-resources-by-manifest/exec-file-with-input/exec-file-with-input.injectable.ts
Nokel81: I guess we don't. The default is 200KiB so this is about 20x the size.

Nokel81 (approved):

PR #6091, opened 2022-08-24
Add installed extensions to nodejs module paths

jakolehm:
Setup nodejs module paths so that extensions can import things from other extensions.


aleksfront (approved):

jakolehm:
It seems this is not needed, `await import("..")` works well enough.

PR #6096, opened 2022-08-24
Remove global version of appEventBus

### LINKED ISSUE #4888: appEventBus

Nokel81:


src/common/utils/async-result.ts
```diff
@@ -3,5 +3,9 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
 export type AsyncResult<Response, Error = string> =
-  | { callWasSuccessful: true; response: Response }
+  | (
+    Response extends void
+    ? { callWasSuccessful: true; response?: undefined }
+    : { callWasSuccessful: true; response: Response }
+  )
```
jansav, line 10: This doesn't seem right. If the expected success response is undefined, then it should be used as a `Response` parameter. Now this means that we have to always make sure that `response` exists, which isn't fine.

src/main/helm/exec-helm/exec-helm.injectable.ts
```diff
@@ -16,20 +16,9 @@ const execHelmInjectable = getInjectable({
...
-      }
-
-      return {
-        callWasSuccessful: false,
-        error: response.error.stderr || response.error.error.message,
-      };
-    };
+    return async (args) => execFile(helmBinaryPath, args, {
+      maxBuffer: 32 * 1024 * 1024 * 1024, // 32 MiB
+    });
```
jansav, line 22: Why do we want to expose complexity of `stderr` to usages?

src/main/helm/repositories/add-helm-repository/add-helm-repository.injectable.ts
```diff
@@ -54,7 +55,18 @@ const addHelmRepositoryInjectable = getInjectable({
...
+
+      if (result.callWasSuccessful) {
+        return {
+          callWasSuccessful: true,
+        };
+      }
+
+      return {
+        callWasSuccessful: false,
+        error: result.error.stderr || result.error.error.message,
```
jansav: This is the type of complexity that I mean with earlier comment about `stderr`.

src/common/cluster-store/cluster-store.ts
```diff
@@ -115,7 +116,7 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
   }
 
   addCluster(clusterOrModel: ClusterModel | Cluster): Cluster {
-    appEventBus.emit({ name: "cluster", action: "add" });
+    this.dependencies.emitAppEvent({ name: "cluster", action: "add" });
```
Iku-turso, line 119: I'm thinking `emitAppEvent` is too low an abstraction for this context. It could instead be `emitClusterAdd()` or maybe even `emit<ClusterEvent>("add")`. The former would replace the logical dependency with a more trackable one. Eg. we could discover listeners of `ClusterEvent` more easily (if even present in this codebase). Also it could add type safety to the payload of the event, ie. `action: "add"`.

src/main/resource-applier/resource-applier.ts
```diff
@@ -49,23 +57,17 @@ export class ResourceApplier {
...
-        throw error.stderr ?? error;
-      }
+    const result = await this.dependencies.execFile(kubectlPath, args);
 
-      throw error;
+    if (result.callWasSuccessful) {
+      return result.response;
     }
+
+    throw result.error.stderr || result.error.error;
```
jansav: Again, this doesn't make sense. `execFile` should abstract this type of complexity away from usages. Best would be that we define error as an interface, e.g.

```typescript 
interface Error { 
  code: string;
  message: string;
  
  original: unknown // Error, string, whatever
} 
```

In this way, we could make `AsyncResult`

```typescript
type AsyncResult<Response> =  { callWasSuccessful: true, response: Response } | { callWasSuccessful: false, error: Error }
```

src/common/utils/async-result.ts
jansav: Read: https://github.com/lensapp/lens/pull/6096/files#r1000512666

src/main/router/router.test.ts
```diff
@@ -32,6 +33,7 @@ describe("router", () => {
     const di = getDiForUnitTesting({ doGeneralOverrides: true });
 
     mockFs();
+    di.permitSideEffects(fsInjectable);
```
jansav, line 36: Why is this needed?

src/main/electron-app/runnables/setup-ipc-main-handlers/setup-ipc-main-handlers.ts
```diff
@@ -78,11 +82,11 @@ export const setupIpcMainHandlers = ({
   });
 
   ipcMainHandle(clusterKubectlApplyAllHandler, async (event, clusterId: ClusterId, resources: string[], extraArgs: string[]) => {
-    appEventBus.emit({ name: "cluster", action: "kubectl-apply-all" });
+    emitAppEvent({ name: "cluster", action: "kubectl-apply-all" });
     const cluster = ClusterStore.getInstance().getById(clusterId);
 
     if (cluster) {
-      const applier = new ResourceApplier(cluster);
+      const applier = createResourceApplier(cluster);
```
Iku-turso, line 92: As common practice, it could be better to not inject factory functions, if an already "factored" instance can be injected instead. Here: `applier` could be the dependency instead of `createResourceApplier`.

src/main/helm/repositories/add-helm-repository/add-helm-repository.injectable.ts
Iku-turso: I'm seeing this `error.error` duplicated in many places. The funk is a bit too much. Improve naming? One could be "errorObject"? Or the root error could be "rejection"?

src/common/utils/async-result.ts
jansav: > This doesn't seem right. If the expected success response is undefined, then it should be used as a Response parameter. Now this means that we have to always make sure that response exists, which isn't fine.


Misread the code, it seems fine.

Iku-turso (changes requested):
Check out comments from me and @jansav.

src/main/router/router.test.ts
Nokel81: Because the router tests still use `mockFs`

src/main/electron-app/runnables/setup-ipc-main-handlers/setup-ipc-main-handlers.ts
Nokel81: small difference

src/main/helm/repositories/add-helm-repository/add-helm-repository.injectable.ts
Nokel81: Or maybe in this case we could just wrap it better.

src/common/cluster-store/cluster-store.ts
Iku-turso: Resolved for being just food for thought.

src/main/electron-app/runnables/setup-ipc-main-handlers/setup-ipc-main-handlers.ts
Iku-turso: Sometimes, sometimes not :) Resolved for already communicating the general point somewhere else.

src/main/router/router.test.ts
Iku-turso: Answer's good, resolved.

Iku-turso (approved):

PR #6115, opened 2022-08-26
React table with resizable columns in Pods list

aleksfront:
Using new table and sibling components for the Pods page. They remove need to use such components as `KubeObjectListLayout`, `ItemListLayout` and previous `Table` and provides more light-weight alternatives.

As addition, new "@tanstack/react-virtual": "3.0.0-beta.18" dependency were added to replace currently used `react-window`. It's newer, provides hook syntax and has ability to render row with dynamic heights. 

New table is using `react-table` https://tanstack.com/table/v8 under the hood, and columns data composed in Pods list following its API.

https://user-images.githubusercontent.com/9607060/186921775-35d8ba28-ce11-4dd1-b25c-a5b8fa9546b2.mov

Things to do:
- [ ] Multy-line columns
- [ ] Paginated table as an option instead of infinite one
- [ ] Remove button for deleting pods as bulk
- [ ] Colors for the status column
- [ ] Scroll to selected item


Iku-turso (changes requested):
Unit tests are missing.

src/renderer/components/+extensions/installed-extensions.tsx
```diff
@@ -10,12 +10,11 @@ import type {
   InstalledExtension,
 } from "../../../extensions/extension-discovery/extension-discovery";
 import { Icon } from "../icon";
-import { List } from "../list/list";
 import { MenuActions, MenuItem } from "../menu";
 import { Spinner } from "../spinner";
 import { cssNames } from "../../utils";
 import { observer } from "mobx-react";
-import type { Row } from "react-table";
+import { ColumnDef, useReactTable, getCoreRowModel, getSortedRowModel } from "@tanstack/react-table";
```
Iku-turso, line 17: Coupling with a 3rd party lib in this arbitrary implementation of "installed extensions" is not ok. We should be using a "Table" or "List" -component which is under our control, and without exposed burden of 3rd parties.

src/renderer/components/+extensions/installed-extensions.tsx
```diff
@@ -10,12 +10,11 @@ import type {
   InstalledExtension,
 } from "../../../extensions/extension-discovery/extension-discovery";
```
Iku-turso, line 11: Hmm. Why does a PR about Pods change "Installed extensions"? Commits in wrong PR? :)

src/renderer/components/+extensions/installed-extensions.tsx
```diff
@@ -46,62 +46,58 @@ function getStatus(extension: InstalledExtension) {
...
+      }
+    },
+    {
+      id: "actions",
+      enableSorting: false,
+      size: 30,
+      cell: (cell) => {
+        const extension = cell.row.original;
+        const { id, isEnabled, isCompatible } = extension;
+        const isUninstalling = extensionInstallationStateStore.isExtensionUninstalling(id);
```
Iku-turso, line 99: Logic about "isUninstalling" and such does not quite belong here in UI-code, but instead on the model of the row, preferably as `IComputedValue`.

src/renderer/components/+extensions/installed-extensions.tsx
```diff
@@ -46,62 +46,58 @@ function getStatus(extension: InstalledExtension) {
...
+    },
+    {
+      id: "version",
+      header: "Version",
+      cell: (cell) => cell.row.original.manifest.version
+    },
+    {
+      id: "status",
+      header: "Status",
+      accessorFn: (row) => row,
```
Iku-turso, line 80: Hmm. What does "accessor" do actually?

src/renderer/components/+extensions/installed-extensions.tsx
```diff
@@ -136,10 +132,17 @@ const NonInjectedInstalledExtensions = observer(({ extensionDiscovery, extension
...
-        ),
-      };
-    }), [extensions, extensionInstallationStateStore.anyUninstalling],
-  );
+        )
+      }
+    }
+  ]), []);
+
+  const table = useReactTable({
```
Iku-turso, line 140: I think "installed extensions" shouldn't know about technical stuff of table like `useReactTable`.

src/renderer/components/+workloads-pods/pods.tsx
```diff
@@ -28,29 +24,33 @@ import type { PodStore } from "./store";
...
   getDetailsUrl: GetDetailsUrl;
   apiManager: ApiManager;
   eventStore: EventStore;
   podStore: PodStore;
   nodeApi: NodeApi;
+  subscribeToWatchStores: SubscribeStores;
+  toggleDetails: ToggleKubeDetailsPane;
 }
 
+const columnHelper = createColumnHelper<Pod>()
```
Iku-turso, line 44: Import-time side-effect is not ok :)

src/renderer/components/+workloads-pods/pods.tsx
```diff
@@ -99,94 +99,135 @@ class NonInjectedPods extends React.Component<Dependencies> {
...
-              </Badge>
-            )
-              : "",
-            pod.getQosClass(),
-            <KubeObjectAge key="age" object={pod} />,
-            { title: pod.getStatusMessage(), className: kebabCase(pod.getStatusMessage()) },
+            pod => pod.status?.podIP || "",
+            pod => pod.getNodeName() || "",
           ]}
+          onRowClick={(item) => this.props.toggleDetails(item.selfLink)}
```
Iku-turso, line 229: Logic could be extracted from UI with `item.toggleDetails()`.

src/renderer/components/+workloads-pods/pods.tsx
```diff
@@ -201,5 +242,7 @@ export const Pods = withInjectables<Dependencies>(NonInjectedPods, {
     nodeApi: di.inject(nodeApiInjectable),
     eventStore: di.inject(eventStoreInjectable),
     podStore: di.inject(podStoreInjectable),
+    subscribeToWatchStores: di.inject(subscribeToStoresDisposersInjectable),
+    toggleDetails: di.inject(toggleKubeDetailsPaneInjectable),
```
Iku-turso, line 246: We wouldn't need this business logic in UI if we had `item.toggleDetails()` from before :)

src/renderer/components/+workloads-pods/pods.tsx
```diff
@@ -201,5 +242,7 @@ export const Pods = withInjectables<Dependencies>(NonInjectedPods, {
     nodeApi: di.inject(nodeApiInjectable),
     eventStore: di.inject(eventStoreInjectable),
     podStore: di.inject(podStoreInjectable),
+    subscribeToWatchStores: di.inject(subscribeToStoresDisposersInjectable),
```
Iku-turso, line 245: It is pretty difficult to understand what this does. Overly complicated design or bad naming? It would be better if UI didn't need to know if observing something requires "subscription".

src/renderer/components/list/list.tsx
```diff
@@ -4,47 +4,76 @@
...
 import styles from "./list.module.scss";
-import React, { useState } from "react";
-import { SearchInput } from "../input";
+import themeStyles from "./table-theme.module.scss";
 
-import type { UseTableOptions } from "react-table";
-import { ReactTable } from "../table/react-table";
+import React, { useEffect, useState } from "react";
+import { SearchInputUrl } from "../input";
+import type { TableOptions } from '@tanstack/react-table'
```
Iku-turso, line 11: 3rd party lib is coupled with "List" but it should be only coupled with "Table".

src/renderer/components/list/table-theme.module.scss
```diff
@@ -0,0 +1,78 @@
+.tableTheme {
```
Iku-turso, line 1: Hmm. Should something called "list" know of something called "table-theme"? Feels a bit off.

src/renderer/components/list/table-theme.module.scss
```diff
@@ -0,0 +1,78 @@
...
+  
+  td {
+    padding: 0 var(--padding);
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
+
+  // Virtual table styles
+  tbody > div {
```
Iku-turso, line 59: `div` as child of `tbody` is HTML-semantically incorrect.

src/renderer/components/list/table-theme.module.scss
Iku-turso, line 1: All these implicit stylings (eg. no explicit classes, and instead HTML-elements are styled) can be pretty undesirable. There's a lengthy ongoing discussion about this [elsewhere](https://miracloud.slack.com/archives/C01HHUBRAL8/p1669201517580609) with JP. Are you sure about this? :)

src/renderer/components/list/table-theme.module.scss
Iku-turso: This is an interesting example how implicit styling can cause damage. Even if `div` as child of `tbody` was semantically sound, now all the `div` elements are hijacked to unspecified intent with absolute positioning and whatnot. Redesign with explicit intent?

src/renderer/components/list/table-theme.module.scss
```diff
@@ -0,0 +1,78 @@
...
+    width: 100%;
+    display: flex;
+    gap: 8px;
+    padding: 0 calc(var(--padding) * 2);
+
+    &:hover {
+      background-color: var(--tableBgcSelected);
+    }
+
+    > div {
```
Iku-turso, line 72: Deeply nested implicit styling with unknown and undecipherable intent.

src/renderer/components/table-list/table-list.tsx
```diff
@@ -0,0 +1,167 @@
+import type { SortingState, Table as TableType, ColumnDef } from "@tanstack/react-table";
```
Iku-turso, line 1: Another coupling with 3rd part lib.

src/renderer/components/table-list/table-list.tsx
```diff
@@ -0,0 +1,167 @@
...
+}
+
+export function TableList<Data>({
+  columns,
+  data,
+  className,
+  selectable = true,
+  configurable = true,
+  columnsResizable = true,
+  onRowClick,
```
Iku-turso, line 28: Too many responsibilities for something called "TableList". Shouldn't UI code for something like "TableList" only care about what it looks like, and let other code take care of other responsibilities?

src/renderer/components/table/virtual-table-header.module.scss
```diff
@@ -0,0 +1,13 @@
+.header {
+  display: block;
+
+  tr {
```
Iku-turso, line 4: Implicit styling. Eg. what if a use-case requires styling that is different than this? That would require a more specific class to cascade these stylings, whatever they may be at the time. This makes the styling brittle. It would be better to to pick and choose what kind of styling is required at what point of time, and thus make the intent and responsibility of the styling explicit.

src/renderer/kube-watch-api/subscribe-to-stores-disposers.injectable.ts
```diff
@@ -0,0 +1,29 @@
...
+const subscribeToStoresDisposersInjectable = getInjectable({
+  id: "subscribe-to-stores-disposers",
+  causesSideEffects: true,
+  instantiate: (di) => {
+    const subscribe = di.inject(subscribeStoresInjectable);
+
+    return (stores: SubscribableStore[]) => {
+      const disposer: Disposer = subscribe(stores, {
+        onLoadFailure: error => {
+          console.warn("Failed to load store: ", error);
```
Iku-turso, line 19: Uncontrollable side-effect.

Iku-turso (changes requested):
In general, the new components introduced in this PR are confused about the correct abstraction they should operate in, and also their responsibility. I'm very sorry to say that, together with lack of unit testing, a bit of a re-design is required before quality is sufficient to merge :(

PR #6117, opened 2022-08-26
Remove release building pipelines

Nokel81:

Do Not Merge until after `6.0.2-alpha.0` is released.


aleksfront (approved):

jim-docker (approved):

jim-docker:
Should (some?) stuff in `./build` also be moved out?

Nokel81:
I don't think any of build can be moved out as of yet but in the future since of it will no longer be needed.

PR #6133, opened 2022-08-29
Add better type safety around Channels

Nokel81:
- Reallow non-JSON serializable complex types

- Remove unnecessary JSON serialization step to IPC

- Remove unnecesary injectable-ization of MessageChannel and
  RequestChannel declarations

- Add channel listener injectable getter function to remove general need
  for type casting and improve type safety


src/common/utils/channel/message-channel-listener-injection-token.ts
```diff
@@ -2,17 +2,47 @@
...
+  channel: Channel;
+  handlerInjectable: Injectable<MessageChannelHandler<Channel>, unknown, void>;
+}
+
+export function getMessageChannelListenerInjectable<
+  Channel extends MessageChannel<Message>,
+  Message,
+>(info: GetMessageChannelListenerInfo<Channel, Message>) {
+  return getInjectable({
+    id: `${info.channel.id}-listener`,
```
Iku-turso: A channel can have multiple listeners, in which case this id will collide?

Nokel81: Good catch, I will append the handler's ID as well since using the same handler twice for the same channel is certainly a bug

src/features/cluster/delete-dialog/main/clear-as-deleteing-channel-listener.injectable.ts
```diff
@@ -0,0 +1,14 @@
+/**
```
Iku-turso: Typo in filename :)

src/features/cluster/delete-dialog/main/clear-as-deleteing-channel-listener.injectable.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getRequestChannelListenerInjectable } from "../../../../common/utils/channel/request-channel-listener-injection-token";
+import { clearClusterAsDeletingChannel } from "../common/clear-as-deleting-channel";
+import clearClusterAsDeletingHandlerInjectable from "./clear-as-deleting-handler.injectable";
+
+const clearClusterAsDeletingChannelListenerInjectable = getRequestChannelListenerInjectable({
```
Iku-turso: Hmm. The previous solution needed only one injectable to listen and act for a channel, and the new solution requires two, ie., a "listener" and a "handler"? Thinking real hard if this can bite us.

Nokel81: We certainly could but this is the root of the benefit this change gives us. We can inject the handler separately if we want to (which I think we do since IIRC I found instances that we were). But also the type safety is checked with the `getRequestChannelListenerInjectable` function. In theory I could have just accepted a handler function here...

src/features/cluster/delete-dialog/main/clear-as-deleting-handler.injectable.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import type { RequestChannelHandler } from "../../../../common/utils/channel/request-channel-listener-injection-token";
+import clustersThatAreBeingDeletedInjectable from "../../../../main/cluster/are-being-deleted.injectable";
+import type { ClearClusterAsDeletingChannel } from "../common/clear-as-deleting-channel";
+
+const clearClusterAsDeletingHandlerInjectable = getInjectable({
+  id: "clear-cluster-as-deleting-handler",
+  instantiate: (di): RequestChannelHandler<ClearClusterAsDeletingChannel> => {
```
Iku-turso: Maybe "the bite" is this: Here typing of instantiate is used to define the "contract" of injectable. If listener and handler of a channel were the same thing, the contract would be the familiar `injection token`, and this would look uniform.

Nokel81: The contract of injection tokens are fine (and even used) but the problem with relying on them is that they are not type safe if the handler is supposed to have some inferred types from the channel

src/features/cluster/delete-dialog/main/delete-cluster-handler.injectable.ts
```diff
@@ -0,0 +1,52 @@
...
+      // Remove from the cluster store as well, this should clear any old settings
+      clusterStore.clusters.delete(cluster.id);
+
+      try {
+        // remove the local storage file
+        const localStorageFilePath = joinPaths(directoryForLensLocalStorage, `${cluster.id}.json`);
+
+        await deleteFile(localStorageFilePath);
+      } catch {
+        // ignore error
```
Iku-turso: Pretty scary to see these wildcard catches 😰

Onwards.

src/features/helm-charts/add-helm-repository-from-list-in-preferences.test.ts
```diff
@@ -129,9 +128,13 @@ describe("add helm repository from list in preferences", () => {
 
           describe("when adding rejects", () => {
             beforeEach(async () => {
-              await execFileMock.reject(
-                "Some error",
-              );
+              await execFileMock.resolve({
+                callWasSuccessful: false,
+                error: {
```
Iku-turso, line 133: `error.error` looks funky? :)

Nokel81: Yes I agree but sometimes it is useful to get the error object itself, and sometimes it is useful to get the stderr output.

Iku-turso: We could have that, and less funk with eg. `error.errorObject` or `error.rejection`? 

Also it feels like something called `stderr` doesn't belong in this layer of abstraction. Should it be `error.errorMessage` or `error.message`?

src/features/cluster/delete-dialog/main/clear-as-deleting-handler.injectable.ts
Iku-turso: > they are not type safe

For understanding: types in handler were inferred/derived from channel before, how were they not type safe?

src/features/cluster/delete-dialog/main/clear-as-deleteing-channel-listener.injectable.ts
Iku-turso: Isn't type safety worse if it relies on usage of `getRequestChannelListenerInjectable`, because injectables created outside of it can be unsafe?

Iku-turso: I'm not sure if ability to "inject the handler separately" is aspirational, as it messages multiple responsibilities (ie. not SRP). However, if the handler does something that other places might be interested in, we can extract that to be injected by the other places. They probably didn't need a "handler" anyway, but instead something more specific.

src/features/helm-charts/add-helm-repository-from-list-in-preferences.test.ts
panuhorsmalahti: Why not always pass object of type `Error`, and the caller can access `error.message` as normal. If stderr is required it could be a property of error object (and we could extend the `Error` class to have type safety).

src/features/cluster/delete-dialog/main/clear-as-deleting-handler.injectable.ts
Nokel81: > types in handler were inferred/derived from channel before, how were they not type safe?

Incorrect, the shape of the injectable was inferred but not the param types or even the return type. Those had to be done by hand.

Iku-turso: That sounds like something we would not do. Weird. Maybe something happened in between.

src/features/helm-charts/add-helm-repository-from-list-in-preferences.test.ts
Nokel81: We are passing the `Error` object too. We will certain work to make our error handling story better.

panuhorsmalahti: I know, I was just making a suggestion to fix the "error.error seems odd" and also simplifying things. But I understand this could be a separate change.

src/main/utils/channel/channel-listeners/handling-of-channels.injectable.ts
```diff
@@ -0,0 +1,36 @@
...
+          if (seenChannels.has(listener.channel)) {
+            throw new Error(`Trying to register a multiple channel handlers for "${listener.channel.id}", which is an error`);
+          }
+
+          seenChannels.add(listener.channel);
+        })
+        .map(enlistRequestChannelListener)
+        .collect(v => Array.from(v));
+
+      return disposer(...requestChannelDisposers);
```
Iku-turso: This is fertile ground for [Split Loop/Map-refactoring](https://devlead.io/DevTips/SplitLoop).

src/main/utils/channel/channel-listeners/handling-of-channels.injectable.ts
```diff
@@ -0,0 +1,36 @@
...
+  instantiate: (di) => {
+    const enlistRequestChannelListener = di.inject(enlistRequestChannelListenerInjectable);
+    const requestChannelListeners = di.injectMany(requestChannelListenerInjectionToken);
+
+    return getStartableStoppable("listening-of-request-channels", () => {
+      const seenChannels = new Set<RequestChannel<unknown, unknown>>();
+      const requestChannelDisposers = iter.pipeline(requestChannelListeners.values())
+        .tap(listener => {
+          if (seenChannels.has(listener.channel)) {
+            throw new Error(`Trying to register a multiple channel handlers for "${listener.channel.id}", which is an error`);
```
Iku-turso: Untested **new code**.

src/main/utils/channel/channel-listeners/handling-of-channels.injectable.ts
```diff
@@ -0,0 +1,36 @@
...
+import type { RequestChannel } from "../../../../common/utils/channel/request-channel-listener-injection-token";
+import { getStartableStoppable } from "../../../../common/utils/get-startable-stoppable";
+import enlistRequestChannelListenerInjectable from "./enlist-request-channel-listener.injectable";
+import { requestChannelListenerInjectionToken } from "./listener-tokens";
+
+const listenerOfRequestChannelsInjectable = getInjectable({
+  id: "listener-of-request-channels",
+  instantiate: (di) => {
+    const enlistRequestChannelListener = di.inject(enlistRequestChannelListenerInjectable);
+    const requestChannelListeners = di.injectMany(requestChannelListenerInjectionToken);
```
Iku-turso: Note for conversation from elsewhere: this list of `requestChannelListeners` isn't reactive, and new listeners will not get "enlisted" if they pop into existence (ie. get registered) after `listenerOfRequestChannelsInjectable` has already `started` (as in `startableStoppable`). I'd say this is a bug waiting to happen.

How about a `startableStoppable` which reacts (eg. with MobX `autorun`, `reaction` or `observe`) to `computedInjectMany` of `requestChannelListenerInjectionToken` to coordinate enlisting and de-enlisting of listeners?

Iku-turso (changes requested):

src/main/utils/channel/channel-listeners/handling-of-channels.injectable.ts
Nokel81: Sure

src/main/utils/channel/channel-listeners/handling-of-channels.injectable.ts
Nokel81: I am going to do this in a separate PR because it is quite technical and can be done separately.

Iku-turso (approved):

PR #6139, opened 2022-08-30
Update metric queries used for Helm installed Prometheus

nevalla:
This PR updates metrics to work with latest versions of community provided Prometheus Helm chart. Tested with the chart versions 15.12.0 and 15.5.3. The main difference is that queries uses `node` label instead of `kubernetes-node` that is not present anymore. Because of this change we cannot extend `PrometheusLens` anymore.

image

image


Nokel81 (approved):
LGTM

aleksfront (approved):

PR #6149, opened 2022-08-31
Refactoring dock tab download logs tests

aleksfront:
Refactoring tests for download logs feature
image

Following this PR https://github.com/lensapp/lens/pull/5970 and specifically this comment https://github.com/lensapp/lens/pull/5970#discussion_r957760701

Since `Download visible logs` behaves differently than `Download all logs` option which makes and api fetch request, both options being separated in tests in a more predictable manner.

For now if no logs available in the dock tab, download dropdown will be disabled.

refactored download behaviour


src/behaviours/pod-logs/download-logs.test.tsx
```diff
@@ -41,14 +54,16 @@ describe("download logs options in pod logs dock tab", () => {
...
+    getLogsMock = jest.fn();
+    getSplittedLogsMock = jest.fn();
 
     builder.beforeWindowStart((windowDi) => {
       windowDi.override(callForLogsInjectable, () => callForLogsMock);
 
       // Overriding internals of logsViewModelInjectable
-      windowDi.override(getLogsInjectable, () => () => ["some-logs"]);
-      windowDi.override(getLogsWithoutTimestampsInjectable, () => () => ["some-logs"]);
-      windowDi.override(getTimestampSplitLogsInjectable, () => () => [...logs]);
```
Nokel81: I think it would be better if we overrode `callForLogs` (which is the dependency of `logStore`) instead. That way we can have better confidence that the business logic within LogStore is correct.

src/behaviours/pod-logs/download-logs.test.tsx
```diff
@@ -74,58 +89,76 @@ describe("download logs options in pod logs dock tab", () => {
...
+        const createLogsTab = windowDi.inject(createPodLogsTabInjectable);
+    
+        getLogsMock.mockReturnValue([]);
+        getSplittedLogsMock.mockReturnValue([]);
+  
+        createLogsTab({
+          selectedPod: pod,
+          selectedContainer: container,
+        });
       });
```
Nokel81: I think these improvements are great! I think that if we used `AsyncFnMock` then we would be able to get more fine grained specification of the behaviour.

For example:

```ts
describe("when a pod logs tab is created", () => {
  beforeEach(() => {
    const createLogsTab = windowDi.inject(createPodLogsTabInjectable);

    createLogsTab({
      selectedPod: pod,
      selectedContainer: container,
    });
  });

  it("should request pod logs", () => {
    expect(callForLogsMock).toBeCalledWith(...);
  });

  it.todo("should open the dock tab");

  it.todo("should show a spinner");

  describe("when the pod logs request resolves to empty", () => {
    beforeEach(async () => {
      await callForLogsMock.resolve("");
    });
    
    it.todo("the download pod logs dropdown should be disabled");

    ...
  });

  describe("when the pod logs request resolves to some lines", () => {
    beforeEach(async () => {
      await callForLogsMock.resolve("a log line");
    });

    it.todo("the download pod logs dropdown should be enabled");

    ...
  });
});
```

Nokel81 (dismissed):

src/behaviours/pod-logs/download-logs.test.tsx
aleksfront: I don't see why using AsyncFnMock would be more useful in our case. For now mocking resulted value done using following form:
```
beforeEach(async () => {
  callForLogsMock.mockResolvedValue("all-logs");
});
```

this is still quite readable and not very different from this form

```
beforeEach(async () => {
  await callForLogsMock.resolve("all-logs");
});
```

Nokel81: The difference is the ability to specify in the tests the expected behaviour while the request has not yet resolved.

That is not possible with the `callForLogsMock.mockResolvedValue` version as it will throw if such a mock is not in place (or just immediately return with `undefined`).

ixrock (approved):
LGTM

Nokel81 (approved):

PR #6164, opened 2022-09-02
Introduce a centralised way to emit telemetry from calls of injectable functions

Iku-turso:
The PR also includes the first actual source of telemetry: navigate-to-route.injectable.ts. This means that all (or most) navigations now emit telemetry.

Notable: in scope of this PR, telemetry is emitted to an event bus, for which further processing is done somewhere else.

Notable2: this PR bumps `injectable` for its own purposes. An indirectly related byproduct of this is that now calls to `di.register()` require an explicit MobX transaction, eg. `runInAction(() => { di.register(...); });`.

Notable3: this PR uses `decorable: false` as a blunt object to prevent recursive decoration of decorators. A better solution is in the horizon.

This is a new requirement from `injectable-extension-for-mobx`, as otherwise registrations could surprisingly cause untimely, and too numerous observations by `computedInjectMany`. The bad thing is the duplication, but I do currently consider this as favorable, as now a dev will always know that registrations of injectables with tokens may cause triggering of observations.


src/features/telemetry/renderer/telemetry-decorator.injectable.ts
```diff
@@ -0,0 +1,65 @@
...
+    };
+  },
+
+  decorable: false,
+  // Todo: this is required because of imperfect typing in injectable.
+  lifecycle: lifecycleEnum.singleton,
+  injectionToken: instantiationDecoratorToken,
+});
+
+const shouldEmitTelemetry = (currentContext: any, whiteList: any) => (
```
jweak: We can probably make this whiteList at least an Array type if not string[]?

Iku-turso: True, one sec.

src/features/telemetry/renderer/telemetry-white-list-for-functions.injectable.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+
+const telemetryWhiteListForFunctionsInjectable = getInjectable({
+  id: "telemetry-white-list-for-functions",
+  instantiate: () => ["some-placeholder-injectable-id"],
```
jweak, line 9: Do we need this placeholder?

src/features/telemetry/renderer/telemetry-decorator.injectable.ts
Iku-turso: Fixed, also simplified the related code a bit.

src/features/telemetry/renderer/telemetry-white-list-for-functions.injectable.ts
Iku-turso: It's there for documentation and inferred typing. I'd say no harm?

jweak: Yeah, no harm.

jweak (approved):

src/common/app-event-bus/app-event-bus.injectable.ts
```diff
@@ -9,6 +9,7 @@ const appEventBusInjectable = getInjectable({
   id: "app-event-bus",
   instantiate: () => appEventBus,
   causesSideEffects: true,
+  decorable: false,
```
Nokel81, line 12: I still don't quite understand what this means, we don't use decorators on `getInjectable` anyway...

src/common/catalog-entities/__tests__/kubernetes-cluster.test.ts
```diff
@@ -12,7 +12,7 @@ describe("kubernetesClusterCategory", () => {
   let kubernetesClusterCategory: KubernetesClusterCategory;
 
   beforeEach(() => {
-    const di = getDiForUnitTesting();
+    const di = getDiForUnitTesting({ doGeneralOverrides: true });
```
Nokel81, line 15: Can we just make this the default now anyway (or remove the option entirely)?

src/features/cluster/workloads/overview/extension-api/order-of-workload-overview-details.test.tsx
```diff
@@ -20,11 +20,13 @@ describe("order of workload overview details", () => {
...
-        someCoreItemWithLowOrderNumberInjectable,
-        someCoreItemWithHighOrderNumberInjectable,
-        someCoreItemWithDefaultOrderNumberInjectable,
-      );
+      runInAction(() => {
+        windowDi.register(
+          someCoreItemWithLowOrderNumberInjectable,
+          someCoreItemWithHighOrderNumberInjectable,
+          someCoreItemWithDefaultOrderNumberInjectable,
+        );
```
Nokel81, line 28: Given that `.register` already takes multiple items. Wouldn't it just make sense to mark it as `action` instead?

src/features/quitting-and-restarting-the-app/opening-application-window-using-tray.test.ts
```diff
@@ -61,10 +62,12 @@ describe("opening application window using tray", () => {
           return browserWindow;
         });
 
-        (mainDi as any).decorateFunction(
-          createElectronWindowInjectable,
-          createElectronWindowMock,
-        );
+        runInAction(() => {
+          (mainDi as any).decorateFunction(
```
Nokel81, line 66: Given that we just updated to a new major version, shouldn't `decorateFunction` be exported now?

Nokel81: Also, can we get some documentation about what this does?

Nokel81 (approved):

PR #6178, opened 2022-09-06
Allow extensions to control status bar item visibility

panuhorsmalahti:
* Status bar items already had `visible` property, which defaults to true. This exposes the property for extensions.
* Set for 6.1.0 as an enhancement

### DIFF
--- a/src/renderer/components/status-bar/status-bar-registration.ts
+++ b/src/renderer/components/status-bar/status-bar-registration.ts
@@ -25,11 +25,6 @@ export interface StatusBarComponents {
    * @default "right"
    */
   position?: "left" | "right";
-
-  /**
-   * If specified, controls item visibility
-   */
-  visible?: IComputedValue<boolean>;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


Nokel81 (dismissed):

src/renderer/components/status-bar/status-bar-registration.ts
```diff
@@ -23,6 +25,11 @@ export interface StatusBarComponents {
    * @default "right"
    */
   position?: "left" | "right";
+
+  /**
+   * If specified, controls item visibility
+   */
+  visible?: IComputedValue<boolean>;
```
jim-docker: Is this actually the best place for the `visible` field? Should it be a field in `StatusBarRegistration`, to be consistent with the other registrations (like `ClusterPageMenuRegistration`, `KubeObjectMenuRegistration`, etc)

Nokel81: Good point

panuhorsmalahti: I'll change it

panuhorsmalahti: Changed, please take a look.

aleksfront (approved):
LGTM

Nokel81 (approved):

PR #6183, opened 2022-09-06
Readd space between key and value in ENV display

### LINKED ISSUE #5656: Space missing between key and value in Environment

Nokel81:

fixes https://github.com/lensapp/lens/issues/5656

### DIFF
--- a/src/renderer/components/+workloads-pods/__tests__/pod-container-env.test.tsx
+++ b/src/renderer/components/+workloads-pods/__tests__/pod-container-env.test.tsx
@@ -1,0 +1,252 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import React from "react";
+import type { ConfigMapStore } from "../../+config-maps/store";
+import configMapStoreInjectable from "../../+config-maps/store.injectable";
+import type { SecretStore } from "../../+config-secrets/store";
+import secretStoreInjectable from "../../+config-secrets/store.injectable";
+import type { Container } from "../../../../common/k8s-api/endpoints";
+import { Secret, ConfigMap, Pod, SecretType } from "../../../../common/k8s-api/endpoints";
+import { getDiForUnitTesting } from "../../../getDiForUnitTesting";
+import type { DiRender } from "../../test-utils/renderFor";
+import { renderFor } from "../../test-utils/renderFor";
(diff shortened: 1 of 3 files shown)


src/renderer/components/+workloads-pods/pod-container-env.tsx
```diff
@@ -83,16 +83,16 @@ const NonInjectedContainerEnvironment = observer((props: Dependencies & Containe
...
+            ? configMap.data[key]
+            : `configMapKeyRef(${name}${key})`;
         }
       }
 
       return (
         <div className="variable" key={name}>
           <span className="var-name">{name}</span>
-          :
+          {` : `}
```
jim-docker, line 95: Staying with `:` (you had changed it to `=` in https://github.com/lensapp/lens/pull/5657)?

src/renderer/components/+workloads-pods/__tests__/pod-container-env.test.tsx
```diff
@@ -0,0 +1,252 @@
...
+        selfLink: "/api/v1/pods/default/my-pod",
+        uid: "1234",
+      },
+      spec: {
+        containers: [container],
+      },
+    });
+    const result = render(<ContainerEnvironment container={container} namespace={pod.getNs()} />);
+
+    expect(result.baseElement).toMatchSnapshot();
```
jim-docker, line 70: shouldn't we expect a bit more than just matching snapshots in these tests?

jim-docker (approved):

src/renderer/components/+workloads-pods/pod-container-env.tsx
Nokel81: Yeah. I figured that would be less controversial.

src/renderer/components/+workloads-pods/__tests__/pod-container-env.test.tsx
Nokel81: I don't think we have to since any changes that change the rendering will cause the tests to fail.

jim-docker: This assumes people look at the snapshot differences, and don't update them blindly when changes happen...

aleksfront (approved):

PR #6189, opened 2022-09-06
Fix beta->latest->beta upgrade loop

### LINKED ISSUE #6007: On Alpha channel Lens constantly wants to update between 6.0.0-latest and 6.0.1-beta and back
Screen Shot 2022-08-09 at 11 54 15 AM

If you have the Update Channel set to Alpha Lens will keep reporting update available, and keep updating on exit, back and forth between 6.0.0-latest and 6.0.1-beta

Beta and Stable channels work as expected.

Nokel81:

fixes #6007

### DIFF
--- a/src/common/vars/release-channel.injectable.ts
+++ b/src/common/vars/release-channel.injectable.ts
@@ -1,0 +1,26 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import type { UpdateChannelId } from "../application-update/update-channels";
+import appSemanticVersionInjectable from "./app-semantic-version.injectable";
+
+const currentReleaseIdInjectable = getInjectable({
+  id: "release-channel",
+  instantiate: (di): UpdateChannelId | undefined => {
+    const appSemanticVersion = di.inject(appSemanticVersionInjectable);
+    const currentReleaseChannel = appSemanticVersion.prerelease[0];
+
+    switch (currentReleaseChannel) {
+      case "latest":
+      case "beta":
+      case "alpha":
+        return currentReleaseChannel;
+      default:
+        return undefined;
+    }
+  },
+});
+
+export default currentReleaseIdInjectable;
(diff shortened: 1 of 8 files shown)


src/common/vars/release-channel.injectable.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import type { UpdateChannelId } from "../application-update/update-channels";
+import appSemanticVersionInjectable from "./app-semantic-version.injectable";
+
+const currentReleaseIdInjectable = getInjectable({
+  id: "release-channel",
```
jim-docker, line 10: naming seems a bit loose here. The id should be `"current-release-id"`? and the file name similarly changed?

jim-docker (commented):
testing...

src/common/vars/release-channel.injectable.ts
Nokel81: Good catch

jim-docker:
This may be a pre-existing issue, where there is a 6.0.1-latest, 6.0.2-alpha, and 6.0.2-beta available on S3 (note the Lens binaries are all the same, built (`yarn dist`) from this PR branch, except for the version set in the package.json):

- launched 6.0.0-latest (also same binary), on alpha channel 
- update is available (6.0.2-alpha)
- relaunch to update
- version is now 6.0.2-alpha
- update is available (6.0.2-beta)

The first update should be straight to 6.0.2-beta? Or it shouldn't immediately update to beta after updating to alpha.

Same result starting from a binary built from master (which I called 6.0.1-latest)

Nokel81:
@jim-docker That is a good catch. I think it is a pre-existing issue which should also be fixed.

I think the correct solution would be go straight to `6.0.2-beta` ignoring the alpha.

I don't think that it should be fixed in this PR though.

jim-docker:
> I don't think that it should be fixed in this PR though.

I can make a new issue

jim-docker (approved):
tested various updates from alpha, beta, stable channels. Can no longer reproduce beta->latest->beta loop bug.

jim-docker:
> I can make a new issue

https://github.com/lensapp/lens/issues/6251

Iku-turso (approved):

PR #6195, opened 2022-09-07
Assume that bundled extensions are always compatible

Nokel81:


jim-docker (approved):
This means developer beware? (or beware more)

Nokel81:
Well the check was that the versions were the same. Nothing to do with the engine version.

Plus we update the bundled extension versions during build to be equal. So it doesn't do much at all.

aleksfront (approved):

PR #6223, opened 2022-09-09
Download arm64 linux binaries

jkroepke:
Running make build on a linux arm64 machine results into error like

```
  • file source doesn't exist  from=.../lens/binaries/client/linux/arm64/kubectl
  • file source doesn't exist  from=.../lens/binaries/client/linux/arm64/lens-k8s-proxy
  • file source doesn't exist  from=.../lens/binaries/client/linux/arm64/helm
```

### DIFF
--- a/build/download_binaries.ts
+++ b/build/download_binaries.ts
@@ -194,7 +194,7 @@ async function main() {
     }, multiBar),
   ];
 
-  if (normalizedPlatform === "darwin") {
+  if (normalizedPlatform !== "windows") {
     downloaders.push(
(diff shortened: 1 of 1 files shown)


Nokel81 (approved):
@jakolehm Do you see any reason not to have this? Just means that it is possible to release onto arm64 based linux.

jkroepke:
Can I help here to move forward here? Currently, I have to modify the script locally. All binaries - including lens-proxy - are available for linux/arm64

jakolehm (approved):

PR #6232, opened 2022-09-13
Make welcome page configurable

jim-docker:
- The route to the welcome page is now read from the `package.json` `config.welcomeRoute` field. 
- `config.welcomeRoute` is set to "/welcome" in `package.json`
- "/welcome" is now treated as the default route for the welcome page (and the default page is the `Welcome` component) 
- If the `package.json` `config.welcomeRoute` is modified with a custom route then the default route is disabled and there is expected to be another route component that corresponds to the custom route.
- This can be facilitated by a global page in an extension. See the `setting-welcome-page` test for determining the custom route given the extension and global page id.


src/features/welcome/setting-welcome-page.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+describe("setting-welcome-page", () => {
+  let applicationBuilder: ApplicationBuilder;
+  let rendered : RenderResult;
+  let welcomeRoute: Route;
+
+  beforeEach(() => {
+    applicationBuilder = getApplicationBuilder();
+  });
+
+  describe("given no extension sets the welcome page", () => {
```
Nokel81: I would change this to read "given configuration of welcome page route is the default" because it is not necessarily an extension changing it and more the process of building.

src/features/welcome/setting-welcome-page.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+    });
+
+    it("navigates to the default welcome page", () => {
+      const welcomePage = rendered.getByTestId("welcome-page");  // from the Welcome component (welcome.tsx)
+
+      expect(welcomePage).toBeInTheDocument();
+    });
+  });
+
+  describe("given an extension specifies a custom welcome page", () => {
```
Nokel81: And here

Nokel81 (commented):
Just some minor test naming comments.

Nokel81 (approved):

Nokel81:
LGTM

aleksfront (approved):
LGTM

PR #6239, opened 2022-09-13
Stop overriding intree extension versions on build

Nokel81:


Makefile
```diff
@@ -63,10 +62,6 @@ ifeq "$(DETECTED_OS)" "Windows"
 endif
 	yarn run electron-builder --publish onTag $(ELECTRON_BUILDER_EXTRA_ARGS)
 
-.PHONY: update-extension-locks
-update-extension-locks:
-	$(foreach dir, $(extensions), (cd $(dir) && rm package-lock.json && ../../node_modules/.bin/npm install --package-lock-only);)
```
Nokel81, line 68: This is unused as we no longer keep those package-lock files in git

Makefile
```diff
@@ -92,7 +87,7 @@ src/extensions/npm/extensions/dist: src/extensions/npm/extensions/node_modules
 	yarn compile:extension-types
 
 src/extensions/npm/extensions/node_modules: src/extensions/npm/extensions/package.json
-	cd src/extensions/npm/extensions/ && ../../../../node_modules/.bin/npm install --no-audit --no-fund
```
Nokel81, line 95: We aren't really using the package-lock file anyway since we override it whenever we do a `make build-extension-types` invocation.

extensions/kube-object-event-status/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kube-object-event-status",
-  "version": "0.0.1",
```
Nokel81, line 3: This version needs to be higher than any version currently in the wild but we also don't need to set it on every build. Only when we change the extension.

src/extensions/npm/extensions/package.json
```diff
@@ -16,8 +16,8 @@
     "name": "OpenLens Authors"
   },
   "dependencies": {
-    "@material-ui/core": "4.12.3",
-    "@types/node": "14.17.14",
+    "@material-ui/core": "^4.12.3",
+    "@types/node": "^16.11.58",
```
Nokel81, line 20: These are changed to make them the same as what we have in the root package.json

Nokel81:
Marking this as `6.2.0` so that it has more time to bake. Also do not merge until after we cut `6.1.0`

extensions/metrics-cluster-feature/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "lens-metrics-cluster-feature",
-  "version": "0.0.1",
+  "version": "6.1.0",
```
aleksfront, line 3: Can you tell more about reason to set extension version numbers same as Lens version itself? Also, by your design, should we change those versions on every app release? For example, next release would be `6.2.0` and extensions will stay with `6.1.0`?

Nokel81: The reason that was to make sure that any new updates to the extensions ensure that they are always up to date (because newer versions will be overridden during update). But if there are no changes then I don't see the point of ensuring that.

If we make a change to the extensions then we should bump the versions.

aleksfront (approved):

Nokel81:
@jakolehm PTAL

Makefile
jakolehm: Just wondering should we include package-locks... anyway that is a story for an another PR.

jakolehm (approved):

Makefile
Nokel81: Yeah probably. Once these are libraries and build themselves for the releases.

PR #6250, opened 2022-09-14
Fix 'time of inject' footgun regarding AppPaths

Nokel81:

Current there is a footgun regarding the injectables around `AppPaths` (including the derivative ones eg `directoryForUserDataInjectable`). That being that merely calling `di.inject()` can lead to an error being thrown.

To resolve this I have introduced a new injectable construct called `LazyInitializableState` which needs an `init: (di) => T` function and then initialises itself the first time that `get(): T` is called. This is an "invasive" fix, because all the injetables that depend on these need to change there uses, but it better encapsulates the usage contract and resolves this footgun.


Nokel81:
This needs to be rebased again once #6338 has been merged

src/common/initializable-state/create.ts
```diff
@@ -5,11 +5,13 @@
 
 import type { DiContainerForInjection, Injectable, InjectionToken } from "@ogre-tools/injectable";
 import { getInjectable } from "@ogre-tools/injectable";
+import type { Runnable } from "../runnable/run-many-for";
 
 export interface CreateInitializableStateArgs<T> {
   id: string;
   init: (di: DiContainerForInjection) => Promise<T> | T;
   injectionToken?: InjectionToken<InitializableState<T>, void>;
+  when: InjectionToken<Runnable<void>, void>;
```
Iku-turso, line 15: Nit: `void` should be default value in both type parameters using it.

src/common/initializable-state/create.test.ts
```diff
@@ -24,10 +24,11 @@ describe("InitializableState tests", () => {
```
Iku-turso, line 26: This entire commit is very interesting, and introduces what could be repeatable patterns for benefit of many other scenarios.

src/common/initializable-state/create.ts
```diff
@@ -5,11 +5,13 @@
 
 import type { DiContainerForInjection, Injectable, InjectionToken } from "@ogre-tools/injectable";
 import { getInjectable } from "@ogre-tools/injectable";
+import type { Runnable } from "../runnable/run-many-for";
 
 export interface CreateInitializableStateArgs<T> {
```
Iku-turso, line 11: Could `InitializableState` and "`TemporallyInitializableState`" (utilising the `when`) be two different responsibilities?

src/main/app-paths/impl.injectable.ts
```diff
@@ -8,11 +8,15 @@ import { createInitializableState } from "../../common/initializable-state/creat
...
+import { beforeElectronIsReadyInjectionToken } from "../start-main-application/runnable-tokens/before-electron-is-ready-injection-token";
 import directoryForIntegrationTestingInjectable from "./directory-for-integration-testing/directory-for-integration-testing.injectable";
 import getElectronAppPathInjectable from "./get-electron-app-path/get-electron-app-path.injectable";
 import setElectronAppPathInjectable from "./set-electron-app-path/set-electron-app-path.injectable";
 
-const appPathsInjectable = createInitializableState({
+const {
+  value: appPathsInjectable,
+  initializer: initAppPathsInjectable,
+} = createInitializableState({
```
Iku-turso, line 19: Elsewhere: still considering your initiative for "sibling injectables" to reduce boilerplate here, and the exports. That being said, what we have here is nice already :)

src/common/app-paths/directory-for-binaries.injectable.ts
```diff
@@ -2,18 +2,32 @@
  * Copyright (c) OpenLens Authors. All rights reserved.
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
+import { initAppPathsOnMainInjectable } from "../../main/app-paths/impl.injectable";
+import { initAppPathsOnRendererInjectable } from "../../renderer/app-paths/impl.injectable";
```
Iku-turso: Code in common cannot be allowed to import from main or renderer.

src/common/app-paths/directory-for-binaries.injectable.ts
```diff
@@ -2,18 +2,32 @@
...
+  ],
+} = createDependentInitializableState({
   id: "directory-for-binaries",
   init: (di) => {
     const joinPaths = di.inject(joinPathsInjectable);
     const directoryForUserData = di.inject(directoryForUserDataInjectable);
 
     return joinPaths(directoryForUserData.get(), "binaries");
   },
+  initAfter: [initAppPathsOnMainInjectable, initAppPathsOnRendererInjectable],
```
Iku-turso: This level of abstraction should not know about stuff like "main" or "renderer". Problem in design?

src/common/app-paths/directory-for-binaries.injectable.ts
```diff
@@ -2,18 +2,32 @@
...
+import { initAppPathsOnRendererInjectable } from "../../renderer/app-paths/impl.injectable";
 import directoryForUserDataInjectable from "./directory-for-user-data.injectable";
-import { createLazyInitializableState } from "../initializable-state/create-lazy";
 import joinPathsInjectable from "../path/join-paths.injectable";
+import { createDependentInitializableState } from "../initializable-state/create-dependent";
 
-const directoryForBinariesInjectable = createLazyInitializableState({
+const {
+  value: directoryForBinariesInjectable,
+  initializers: [
```
Iku-turso, line 11: Multiple initializers? Should such scenario ever exist? I mean, why would there be an initializer which only has partial responsibility of initializing something?

src/common/initializable-state/create-dependent.ts
```diff
@@ -0,0 +1,80 @@
...
+          }
+
+          return box.value;
+        },
+      };
+    },
+    injectionToken,
+  });
+
+  const initializers = initAfter.map(runnableInjectable => getInjectable({
```
Iku-turso: This map creates a new runnable for each `initAfter`, which is not the implied usage of `initAfter`, and can cause duplicate inits.

src/common/initializable-state/create-dependent.ts
```diff
@@ -0,0 +1,80 @@
...
+  injectionToken?: InjectionToken<InitializableState<T>, void>;
+  initAfter: Injectable<Runnable<void>, Runnable<void>, void>[];
+}
+
+export interface CreateDependentInitializableStateResult<T> {
+  value: Injectable<InitializableState<T>, unknown, void>;
+  initializers: Injectable<Runnable<void>, Runnable<void>, void>[];
+}
+
+export function createDependentInitializableState<T>(args: CreateDependentInitializableStateArgs<T>): CreateDependentInitializableStateResult<T> {
```
Iku-turso: Integral technical function does not have technical unit tests.

src/common/runnable/run-many-for.ts
```diff
@@ -2,13 +2,10 @@
  * Copyright (c) OpenLens Authors. All rights reserved.
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
-import { pipeline } from "@ogre-tools/fp";
-import type {
-  DiContainerForInjection,
-  InjectionToken,
-} from "@ogre-tools/injectable";
-import { filter, forEach, map, tap } from "lodash/fp";
-import { throwWithIncorrectHierarchyFor } from "./throw-with-incorrect-hierarchy-for";
```
Iku-turso, line 11: Did this file get removed as well? Not in this commit? :)

src/common/runnable/run-many-for.ts
```diff
@@ -18,35 +15,34 @@ export interface Runnable<TParameter = void> {
...
-  injectionToken: InjectionToken<Runnable<Param>, void>
-) => Run<Param>;
+export type RunMany = <Param>(injectionToken: InjectionToken<Runnable<Param>, void>) => Run<Param>;
+
+async function runCompositeRunnables<Param>(param: Param, composite: Composite<Runnable<Param>>) {
+  await composite.value.run(param);
+  await Promise.all(composite.children.map(composite => runCompositeRunnables(param, composite)));
+}
 
 export function runManyFor(di: DiContainerForInjection): RunMany {
```
Iku-turso, line 25: Using `composite` for this algorithm is very cool! Nice :)

src/common/runnable/run-many-sync-for.ts
```diff
@@ -16,33 +15,34 @@ export interface RunnableSync<TParameter = void> {
 
 type RunSync<Param> = (parameter: Param) => void;
 
-export type RunManySync = <Param>(
-  injectionToken: InjectionToken<Runnable<Param>, void>
-) => RunSync<Param>;
+export type RunManySync = <Param>(injectionToken: InjectionToken<RunnableSync<Param>, void>) => RunSync<Param>;
+
+function runCompositeRunnableSyncs<Param>(param: Param, composite: Composite<RunnableSync<Param>>) {
```
Iku-turso, line 21: Hmm. The same thing could be achieved without recursion using `getCompositeNormalization`.

src/common/__tests__/user-store.test.ts
```diff
@@ -47,30 +49,32 @@ describe("user store tests", () => {
...
-      get: () => "latest" as const,
-      init: async () => {},
-    }));
-
     di.unoverride(userStoreInjectable);
+
+    const runManySync = runManySyncFor(di);
+    const runAllBeforeElectronIsReady = runManySync(beforeElectronIsReadyInjectionToken);
+
+    runAllBeforeElectronIsReady();
```
Iku-turso, line 61: Strictly, this is a duplication of implementation details for what happens at start of application. It would be safer to make the test even more like "real application" by running this logic from the real application instead of reimagining it here. Stuff like this is likely to grow out of sync making the test unreliable and brittle.

src/extensions/extension-discovery/extension-discovery.test.ts
```diff
@@ -87,27 +93,27 @@ describe("ExtensionDiscovery", () => {
...
-      });
-      runInAction(() => letTestFinish.set(true));
-    });
+    const onAdd = jest.fn();
+
+    extensionDiscovery.events.on("add", onAdd);
 
     addHandler(joinPaths(extensionDiscovery.localFolderPath, "/my-extension/package.json"));
-    await when(() => letTestFinish.get());
+    await delay(10);
```
Iku-turso, line 101: Arbitrary sleep in test is a design smell. Could `await flushPromises()` make it good?

src/extensions/extension-discovery/extension-discovery.test.ts
```diff
@@ -28,15 +30,20 @@ describe("ExtensionDiscovery", () => {
...
     di.override(installExtensionInjectable, () => () => Promise.resolve());
     di.override(extensionApiVersionInjectable, () => "5.0.0");
 
+    const runManySync = runManySyncFor(di);
+    const runMany = runManyFor(di);
+    const runAllBeforeElectronIsReady = runManySync(beforeElectronIsReadyInjectionToken);
+    const runAllBeforeApplicationIsLoading = runMany(beforeApplicationIsLoadingInjectionToken);
+
+    runAllBeforeElectronIsReady();
+    await runAllBeforeApplicationIsLoading();
```
Iku-turso, line 45: This is another reimagination of details in actual application start.

src/main/catalog-sources/__test__/kubeconfig-sync.test.ts
```diff
@@ -46,16 +48,18 @@ describe("kubeconfig-sync.source tests", () => {
 
     clusters = new Map();
     di.override(getClusterByIdInjectable, () => id => clusters.get(id));
-    di.override(directoryForUserDataInjectable, () => ({
-      get: () => "some-directory-for-user-data",
-    }));
```
Iku-turso: Now that I think of it, getting rid of all these boilerplate overrides for directories is good. The stubbed directories of global overrides are well enough to communicate all relevancy in this.

src/main/router/router.test.ts
```diff
@@ -39,13 +42,18 @@ describe("router", () => {
...
     di.override(normalizedPlatformInjectable, () => "darwin");
 
+    const runManySync = runManySyncFor(di);
+    const runMany = runManyFor(di);
+    const runAllBeforeElectronIsReady = runManySync(beforeElectronIsReadyInjectionToken);
+    const runAllBeforeApplicationIsLoading = runMany(beforeApplicationIsLoadingInjectionToken);
+
+    runAllBeforeElectronIsReady();
+    await runAllBeforeApplicationIsLoading();
+
```
Iku-turso, line 56: Another duplicated reimagination.

src/main/shell-session/local-shell-session/techincal.test.ts
```diff
@@ -20,17 +22,18 @@ import openLocalShellSessionInjectable from "./open.injectable";
...
-    di.override(buildVersionInjectable, () => ({
-      get: () => "1.1.1",
-    }));
+    const runManySync = runManySyncFor(di);
+    const runMany = runManyFor(di);
+    const runAllBeforeElectronIsReady = runManySync(beforeElectronIsReadyInjectionToken);
+    const runAllBeforeApplicationIsLoading = runMany(beforeApplicationIsLoadingInjectionToken);
+
+    runAllBeforeElectronIsReady();
+    await runAllBeforeApplicationIsLoading();
```
Iku-turso, line 36: Reimagination.

src/common/runnable/run-many-sync-for.ts
```diff
@@ -19,7 +19,7 @@ type RunSync<Param> = (parameter: Param) => void;
 export type RunManySync = <Param>(injectionToken: InjectionToken<RunnableSync<Param>, void>) => RunSync<Param>;
 
 function runCompositeRunnableSyncs<Param>(param: Param, composite: Composite<RunnableSync<Param>>) {
-  assert(!((composite.value.run(param) as any) instanceof Promise), "Cannot be an async function for runnable sync");
+  assert(!((composite.value.run(param) as unknown) instanceof Promise), `Cannot be an async function for runnable sync`);
```
Iku-turso, line 22: Technical unit test for integral technical function is missing (I think). As policy, our team could **at least** match the quality of code we encounter :) Detour to the [Boy Scout Rule](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html) we talked about elsewhere :)

Eg. this function is otherwise unit tested, so let's not damage it with addition of non-tested stuff.

Iku-turso, line 22: Bad english.

Iku-turso, line 22: I wonder if there's a way to enforce "no async" with typing instead of at runtime. I do remember it being a bit tricky with TypeScript, but still.

integration/__tests__/app-preferences.tests.ts
```diff
@@ -24,9 +24,10 @@ describe("preferences page tests", () => {
 
     await app.evaluate(async ({ app }) => {
       await app.applicationMenu
-        .getMenuItemById(process.platform === "darwin" ? "mac" : "file")
-        .submenu.getMenuItemById("navigate-to-preferences")
-        .click();
+        ?.getMenuItemById(process.platform === "darwin" ? "mac" : "file")
+        ?.submenu
+        ?.getMenuItemById("navigate-to-preferences")
+        ?.click();
```
Iku-turso, line 30: Not finding these items should be an error instead of something that silently fails?

src/common/utils/get-startable-stoppable.test.ts
```diff
@@ -233,3 +234,59 @@ describe("getStartableStoppable", () => {
...
+
+  it("when stopping before ever starting, throws", () => {
+    expect(() => actual.stop()).toThrow("Tried to stop \"some-id\", but it is already stopped.");
+  });
+
+  it("is not started", () => {
+    expect(actual.started).toBe(false);
+  });
+
+  describe("when started", () => {
```
Iku-turso, line 265: How about "when started again"?

src/common/utils/get-startable-stoppable.test.ts
```diff
@@ -233,3 +234,59 @@ describe("getStartableStoppable", () => {
...
+
+    it("calls start function", () => {
+      expect(startMock).toHaveBeenCalled();
+    });
+
+    it("is started", () => {
+      expect(actual.started).toBe(true);
+    });
+
+    describe("when stopped", () => {
```
Iku-turso, line 278: How about "when stopped again"?

src/common/utils/get-startable-stoppable.ts
```diff
@@ -3,58 +3,85 @@
...
+      return state === "started";
+    },
+
+    start: async () => {
+      if (state !== "stopped") {
+        throw new Error(`Tried to start "${id}", but it is already ${state}.`);
       }
 
-      if (!started) {
-        throw new Error(`Tried to stop "${id}", but it has not started yet.`);
```
Iku-turso, line 51: Nitty opinion: Non-OCP state-string is cognitively more complex than explicit boolean flags. Not a deal-breaker :)

src/common/utils/get-startable-stoppable.ts
```diff
@@ -3,58 +3,85 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
 
-type Stopper = () => Promise<void> | void;
-type Starter = () => Promise<Stopper> | Stopper;
+export type Stopper = () => Promise<void> | void;
+export type Starter = () => Promise<Stopper> | Stopper;
```
Iku-turso, line 7: Do these internal types need to be exported?

src/common/utils/get-startable-stoppable.ts
```diff
@@ -3,58 +3,85 @@
...
 
-export const getStartableStoppable = (
```
Iku-turso, line 9: Refactoring this existing functionality is not implied in name of commit. Add explicit commits in future? :)

Iku-turso:
Let's discuss this footgun a little bit:

> ...merely calling di.inject() can lead to an error being thrown.

I'm considering the possibility of this error actually being a **good thing** instead of **footgun**. We are talking about throwing when attempting to inject something that is not guaranteed ready for consumption. The expected responsibility of dependency injection is to inject dependencies that are ready for consumption. Instead, it could be wrong to load the dependee with an **additional responsibility** of surviving the case of a dependency that is not ready for consumption. That's a breach of Single Responsibility Principle, which often communicates a more severe design problem elsewhere.

Iku-turso:
> ...merely calling di.inject() can lead to an error being thrown.

Note: in general, a `di.inject()` may indeed throw if there is a good reason to do so. This is not to be considered inherently evil.

Iku-turso:
I'd humbly like to understand the footgun better. 

1. What is the damage of it? 
2. What is the sequence of actions leading to it?

Iku-turso:
Argument: deferring logic of initialization from `injection-time` to `run-time` (there's a difference) creates logical/temporal dependencies and thus exposes us to a difficult subset of bugs.

src/common/initializable-state/create.ts
Iku-turso: This is also seen in many other places in the PR.

Iku-turso (changes requested):
For this PR, I worry about some fundamental design smells described in the comments. Let's be extra-cautious, yes? :/

src/common/initializable-state/create.ts
Nokel81: Yes, I prefer just using it here so that I don't have to go searching for a different file to figure out what the type param is.

src/common/initializable-state/create.ts
Nokel81: Maybe in the future...

However, it has been the case that I have always wanted to specify that something is initialised at a specific time.

src/common/runnable/run-many-sync-for.ts
Nokel81: Not if we want implicit returns to be allowed. Typescript treats the `void` return type as special allowing all functions to coerce to it.

We could change it to be `undefined` but then the functions would need an explicit `return undefined` to be typed correctly.

src/common/utils/get-startable-stoppable.ts
Nokel81: 100% disagree. Having to keep multiple variables in sync is the worst kind of code smell

src/common/utils/get-startable-stoppable.ts
Nokel81: They are not internal, they form part of the exported function signature

Nokel81:
> deferring logic of initialization from injection-time to run-time (there's a difference) creates logical/temporal dependencies and thus exposes us to a difficult subset of bugs.

1. You need to give an example or this argument is not very persuasive.
2. Saying that the deferring creates the "logical/temporal dependencies" is false. They already exist, this change just makes them front and centre.

Nokel81:
Closing after discussion

PR #6289, opened 2022-09-26
Switch to shared "runnable" instead of local

Iku-turso:
The motivation for this is to also let extensions do stuff with the `runnable`. This was requested by extension development. The `runnable` is now shared as a new library in `@ogre-tools/injectable-utils`.

Related: Stuff like `startable-stoppable` could also be on its way there later.

Note: not sure at all what is the correct milestone for this.


Nokel81 (approved):

jakolehm (approved):

Iku-turso:
This PR is blocked by organize-feature-for-application-update-in-feature-directory, as new changes in runnable will need to be synced to the "shared" runnable in ogre-tools, which has seen a major version bump since.

Nokel81:
Closing as this is quite out of date now.

PR #6196, opened 2022-09-07
Clean up versioning and packageJson uses

Nokel81:
- Rename packageJsonInjectable -> applicationInformationInjectable

- Add injectables for all sub items

- Introduce storeMigrationVersionInjectable for BaseStores

- Introduce extensionApiVersionInjectable

- Use buildVersionInjectable in more places, such as in telemetry

Signed-off-by: Sebastian Malton <sebastian@malton.name>

This work is required to make sure we use `buildVersionInjectable` in all the correct places. I also took the opportunity to clean up the rest of our uses of `packageJsonInjectable` and to only use a global override.

PR #6291, opened 2022-09-27
Fix Singleton type in exports

panuhorsmalahti:
It seems that https://github.com/lensapp/lens/pull/6196 had a regression in the exported Singleton type in `@k8slens/extensions` version 6.1.1 (6.0.0 worked correctly). This causes a breaking API change with error `Type 'Singleton' is not a constructor function type` for code `export class FooBar extends Common.Util.Singleton`

### DIFF
--- a/src/extensions/common-api/utils.ts
+++ b/src/extensions/common-api/utils.ts
@@ -13,7 +13,7 @@ import type { IClassName } from "../../renderer/utils/cssNames";
 import { cssNames } from "../../renderer/utils/cssNames";
 
 export interface UtilsExtensionItems {
-  Singleton: Singleton;
+  Singleton: typeof Singleton;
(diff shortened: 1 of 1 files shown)


src/extensions/common-api/utils.ts
```diff
@@ -13,7 +13,7 @@ import type { IClassName } from "../../renderer/utils/cssNames";
 import { cssNames } from "../../renderer/utils/cssNames";
 
 export interface UtilsExtensionItems {
-  Singleton: Singleton;
+  Singleton: typeof Singleton;
```
jim-docker, line 16: The rest here are functions so they are ok(?)

jim-docker (approved):

src/extensions/common-api/utils.ts
panuhorsmalahti: I believe so.

chenhunghan (approved):

PR #6294, opened 2022-09-27
Add Runtime Classes views

dex4er:
Signed-off-by: Piotr Roszatycki <piotr.roszatycki@gmail.com>

**Description of changes:**

Additional "Runtime Classes" view:

image

Detailed view shows Node Selector and Tolerations like in Pods view:

image

Detailed view for Pods shows now Runtime Class (only when it is set):

image

Additionally this PR hides Priority Class when it is not set.


Nokel81 (approved):
LGTM

PR #6318, opened 2022-10-04
Fix loading helm release details

Nokel81:
- The helm manifest can sometimes contain KubeJsonApiDataLists
  instead of just KubeJsonApiData entries

- Add additional logging to main for when a route handler throws so that
  we can gain more context in the future

### DIFF
--- a/src/main/helm/helm-service/get-helm-release-resources/get-helm-release-resources.injectable.ts
+++ b/src/main/helm/helm-service/get-helm-release-resources/get-helm-release-resources.injectable.ts
@@ -3,48 +3,38 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
 import { getInjectable } from "@ogre-tools/injectable";
-import callForKubeResourcesByManifestInjectable from "./call-for-kube-resources-by-manifest/call-for-kube-resources-by-manifest.injectable";
-import { groupBy, map } from "lodash/fp";
-import type { JsonObject } from "type-fest";
-import { pipeline } from "@ogre-tools/fp";
 import callForHelmManifestInjectable from "./call-for-helm-manifest/call-for-helm-manifest.injectable";
+import type { KubeJsonApiData, KubeJsonApiDataList } from "../../../../common/k8s-api/kube-json-api";
+import type { AsyncResult } from "../../../../common/utils/async-result";
 
 export type GetHelmReleaseResources = (
   name: string,
   namespace: string,
   kubeconfigPath: string,
   kubectlPath: string
-) => Promise<JsonObject[]>;
+) => Promise<AsyncResult<KubeJsonApiData[], string>>;
 
 const getHelmReleaseResourcesInjectable = getInjectable({
   id: "get-helm-release-resources",
 
   instantiate: (di): GetHelmReleaseResources => {
     const callForHelmManifest = di.inject(callForHelmManifestInjectable);
-    const callForKubeResourcesByManifest = di.inject(callForKubeResourcesByManifestInjectable);
 
-    return async (name, namespace, kubeconfigPath, kubectlPath) => {
+    return async (name, namespace, kubeconfigPath) => {
       const result = await callForHelmManifest(name, namespace, kubeconfigPath);
 
       if (!result.callWasSuccessful) {
-        throw new Error(result.error);
+        return result;
       }
 
-      const results = await pipeline(
-        result.response,
-
-        groupBy((item) => item.metadata.namespace || namespace),
-
-        (x) => Object.entries(x),
-
-        map(([namespace, manifest]) =>
-          callForKubeResourcesByManifest(namespace, kubeconfigPath, kubectlPath, manifest),
-        ),
-
-        promises => Promise.all(promises),
-      );
-
-      return results.flat(1);
+      return {
+        callWasSuccessful: true,
+        response: result.response.flatMap(item => (
+          Array.isArray(item.items)
+            ? (item as KubeJsonApiDataList).items
+            : item as KubeJsonApiData
+        )),
+      };
     };
   },
 });
--- a/src/main/helm/helm-service/get-helm-release-resources/get-helm-release-resources.injectable.ts
+++ b/src/main/helm/helm-service/get-helm-release-resources/get-helm-release-resources.injectable.ts
@@ -11,7 +11,6 @@ export type GetHelmReleaseResources = (
   name: string,
   namespace: string,
   kubeconfigPath: string,
-  kubectlPath: string
 ) => Promise<AsyncResult<KubeJsonApiData[], string>>;
 
 const getHelmReleaseResourcesInjectable = getInjectable({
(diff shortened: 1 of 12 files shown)


src/main/helm/helm-service/get-helm-release-resources/get-helm-release-resources.test.ts
```diff
@@ -66,150 +65,9 @@ describe("get helm release resources", () => {
 
       const actual = await actualPromise;
 
-      expect(actual).toEqual([]);
```
Iku-turso, line 68: These unit tests are completely removed instead of "updated" as proposed in commit! Making tested code untested code is a pretty big hit on quality, right? :)

Nokel81: These unit tests were removed because the function they mock is not called anymore.

Iku-turso: Are you sure? I can see no unit test resulting to other than:

`response: []`

The interesting test would describe how to get some content therein?

src/main/helm/helm-service/get-helm-release-resources/get-helm-release-resources.test.ts
```diff
@@ -65,150 +65,9 @@ describe("get helm release resources", () => {
...
-                resourceVersion: "",
-                selfLink: "",
-              },
-            }),
-          },
-        );
-
-        const actual = await actualPromise;
-
-        expect(actual).toEqual([{ some: "item" }, { some: "other-item" }]);
```
Iku-turso, line 177: About the missing tests: I think there is no new substitute for this removed test.

src/main/router/create-handler-for-route.injectable.ts
```diff
@@ -0,0 +1,80 @@
...
+    serverResponse.write(content);
+    serverResponse.end();
+  } else if (content) {
+    serverResponse.end(content);
+  } else {
+    serverResponse.end();
+  }
+};
+
+const createHandlerForRouteInjectable = getInjectable({
```
Iku-turso, line 44: Hmm. For sake of conversation: I wonder, in general, if instead of "creating" a "handler" with a factory-function, we could have something that "is" a handler (or handlers). That would have better chance of being reactive when required, and it would comply to Hollywood. Later for something else maybe? :)

Iku-turso (dismissed):
LGTM, but I would check with @jansav, as this PR touches a pretty complicated bug fix he did before.

src/main/router/create-handler-for-route.injectable.ts
Nokel81: Fair, I mostly just moved this around because I wanted to add some logging.

This is here so that we can catch exceptions thrown accidentally and return 500s automatically as well as mapping the correct types and serialization

src/main/helm/helm-service/get-helm-release-resources/get-helm-release-resources.injectable.ts
```diff
@@ -3,48 +3,37 @@
...
     const callForHelmManifest = di.inject(callForHelmManifestInjectable);
-    const callForKubeResourcesByManifest = di.inject(callForKubeResourcesByManifestInjectable);
 
-    return async (name, namespace, kubeconfigPath, kubectlPath) => {
+    return async (name, namespace, kubeconfigPath) => {
       const result = await callForHelmManifest(name, namespace, kubeconfigPath);
 
       if (!result.callWasSuccessful) {
-        throw new Error(result.error);
+        return result;
```
jansav, line 26: Hmm. Any particular reason why we are not throwing anymore?

src/main/helm/helm-service/get-helm-release-resources/get-helm-release-resources.injectable.ts
```diff
@@ -3,48 +3,37 @@
...
-
-        map(([namespace, manifest]) =>
-          callForKubeResourcesByManifest(namespace, kubeconfigPath, kubectlPath, manifest),
-        ),
-
-        promises => Promise.all(promises),
-      );
-
-      return results.flat(1);
+      return {
```
jansav, line 29: Could you explain a bit why this is enough. Previously we had to get resources from manifest and this was the only known solution.

src/main/helm/helm-service/get-helm-release-resources/get-helm-release-resources.test.ts
```diff
@@ -65,150 +65,80 @@ describe("get helm release resources", () => {
 
       const actual = await actualPromise;
 
-      expect(actual).toEqual([]);
+      expect(actual).toEqual({
```
jansav, line 68: I do have to agree with Mikko here. It seems we are not testing the happy case anymore at all. Could add test case for it?

src/main/helm/helm-service/get-helm-release-resources/get-helm-release-resources.injectable.ts
Nokel81: Because the return type was changed to `AsyncResult`

src/main/router/create-handler-for-route.injectable.ts
```diff
@@ -0,0 +1,80 @@
...
+
+interface LensServerResponse {
+  statusCode: number;
+  content: any;
+  headers: {
+    [name: string]: string;
+  };
+}
+
+const writeServerResponseFor = (serverResponse: ServerResponse) => ({
```
jansav, line 23: I thought back end routing handles this already? `main/router/router.ts` `handleRoute`

src/main/helm/helm-service/get-helm-release-resources/get-helm-release-resources.injectable.ts
Nokel81: The only reason that we needed to get the resources is so that we could determine their age. I felt that it would be better UX if that even in the case where they don't exist the helm details panel still is displayed I removed that column which then removes the need to get the more detailed results from `kubectl`.

The user can still view the age if they want to by clicking through the links on the release details page

src/main/helm/helm-service/get-helm-release-resources/get-helm-release-resources.test.ts
Nokel81: I thought I added one...

src/main/router/create-handler-for-route.injectable.ts
Nokel81: I moved that exact function here to a new file so that I could add `logger` as a dependency.

src/main/helm/helm-service/get-helm-release-resources/get-helm-release-resources.test.ts
Nokel81: Yeah I did, on line 74 bellow

jansav: Cool, saw it now.

src/main/router/create-handler-for-route.injectable.ts
jansav: Ah.

jansav (approved):

PR #6321, opened 2022-10-04
Make systems CAs platform-independent using injectables

Nokel81:
* Make injecting CAs injectable, remove mac-ca as dependency
* Fix win-ca failing on electron renderer on windows
* Fix the matcher under features/ for main

~This is a reintroduction of the PR which I will work on to see why it causes windows integration tests to fail.~


webpack/renderer.ts
```diff
@@ -85,6 +87,10 @@ export function webpackLensRenderer({ showVars = true } = {}): webpack.Configura
     },
 
     plugins: [
+      new DefinePlugin({
+        CONTEXT_MATCHER_FOR_NON_FEATURES: `/\\.injectable(\\.${platform})?\\.tsx?$/`,
+        CONTEXT_MATCHER_FOR_FEATURES: `/\\/(renderer|common)\\/.+\\.injectable(\\.${platform})?\\.tsx?$/`,
+      }),
```
jansav, line 93: So this means that we don't have to use `isMac`, `isWindows` etc. anymore? We can now introduce injection token and then specify the implementation in `something.injectable.mac.ts`, `something.injectable.windows.ts`?

Nokel81: > So this means that we don't have to use isMac, isWindows etc. anymore?

Yes we should be able to move towards that.

> We can now introduce injection token and then specify the implementation in something.injectable.mac.ts, something.injectable.windows.ts?

Correct. However, something about this PR breaks on windows CI (not locally on my windows machine though) so I need to figure out why.

jansav: Ok, cool.

src/common/certificate-authorities/inject-system-cas.injectable.ts
```diff
@@ -0,0 +1,39 @@
...
+  instantiate: (di) => {
+    const requestSystemCAs = di.inject(requestSystemCAsInjectionToken);
+
+    return async () => {
+      for (const cert of await requestSystemCAs()) {
+        if (isCertActive(cert)) {
+          if (Array.isArray(globalAgent.options.ca) && !globalAgent.options.ca.includes(cert)) {
+            globalAgent.options.ca.push(cert);
+          } else {
+            globalAgent.options.ca = [cert];
```
jim-docker, line 30: This looks right for when `globalAgent.options.ca` is not an array but it will also overwrite an existing array that already has `cert` plus possibly more certs. Is this OK?

I see this is old code. Maybe I made this comment before? :)

src/common/fs/exec-file.injectable.ts
```diff
@@ -7,28 +7,50 @@ import type { ExecFileException, ExecFileOptions } from "child_process";
...
-      if (error) {
-        resolve({
-          callWasSuccessful: false,
-          error: Object.assign(error, { stderr }),
-        });
-      } else {
-        resolve({
-          callWasSuccessful: true,
-          response: stdout,
+  instantiate: (): ExecFile => {
```
jim-docker, line 21: not sure why this needs to be refactored here

jim-docker (approved):
Seems the title should be more like "Make systems CAs platform-independent using injectables"

There are a few ride-alongs here

src/common/fs/exec-file.injectable.ts
Nokel81: Because there are some cases in this PR where I need to pass options but not args

src/common/certificate-authorities/inject-system-cas.injectable.ts
Nokel81: That is a good point, I will make a follow up PR to fix this.

Nokel81: https://github.com/lensapp/lens/pull/6628

src/common/fs/exec-file.injectable.ts
jim-docker: I looked for that, didn't see it. Oh, guess it is for the win32 case

Nokel81: Yes

PR #6351, opened 2022-10-07
Improve shellEnv

jakolehm:
Replaces `shell-env` npm package with a custom implementation which has similar logic as `vscode`.


src/main/utils/shell-env.ts
```diff
@@ -16,20 +90,18 @@ let shellSyncFailed = false;
  * Subsequent calls after such a timeout simply log an error message without trying
  * to get the environment, unless forceRetry is true.
  * @param shell the shell to get the environment from
- * @param forceRetry if true will always try to get the environment, otherwise if
- * a previous call to this function failed then this call will fail too.
  * @returns object containing the shell's environment variables. An empty object is
  * returned if the call fails.
  */
-export async function shellEnv(shell?: string, forceRetry = false) : Promise<EnvironmentVariables> {
```
jakolehm, line 24: Removed `forceRetry` because it was not used anywhere (dead code is bad).

src/main/utils/shell-env.ts
```diff
@@ -3,11 +3,85 @@
...
+  let command: string;
+  let shellArgs: string[];
+  
+  if (/^pwsh(-preview)?$/.test(shellName)) {
+    // Older versions of PowerShell removes double quotes sometimes so we use "double single quotes" which is how
+    // you escape single quotes inside of a single quoted string.
+    command = `& '${process.execPath}' -p '''${mark}'' + JSON.stringify(process.env) + ''${mark}'''`;
+    shellArgs = ["-Login", "-Command"];
+  } else {
+    command = `'${process.execPath}' -p '"${mark}" + JSON.stringify(process.env) + "${mark}"'`;
```
jakolehm, line 35: Basically we outsource environment variable parsing to a nodejs process (which is started using shell environment). Markers are needed because shell might output some intro text that would confuse `JSON.parse`.

Nokel81 (approved):
LGTM

PR #6352, opened 2022-10-07
Clickable links in pods details for ServiceAccount, PriorityClass and RuntimeClass

dex4er:
This is a small improvement for pod details:

image

Links to ServiceAccount, PriorityClass and RuntimeClass are clickable and move to the details of the chosen resource.

### DIFF
--- a/src/renderer/components/+workloads-pods/pod-details.tsx
+++ b/src/renderer/components/+workloads-pods/pod-details.tsx
@@ -100,6 +100,7 @@ class NonInjectedPodDetails extends React.Component<PodDetailsProps & Dependenci
     const initContainers = pod.getInitContainers();
     const containers = pod.getContainers();
 
+    const namespace = pod.getNs();
     const priorityClassName = pod.getPriorityClassName();
     const runtimeClassName = pod.getRuntimeClassName();
     const serviceAccountName = pod.getServiceAccountName();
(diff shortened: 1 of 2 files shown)


src/renderer/components/+workloads-pods/pod-details.tsx
```diff
@@ -94,6 +94,20 @@ class NonInjectedPodDetails extends React.Component<PodDetailsProps & Dependenci
...
+
+    const priorityClassDetailsUrl = getDetailsUrl(priorityClassApi.getUrl({
+      name: priorityClassName,
+    }));
+    const runtimeClassDetailsUrl = getDetailsUrl(runtimeClassApi.getUrl({
+      name: runtimeClassName,
+    }));
+    const serviceAccountDetailsUrl = getDetailsUrl(serviceAccountApi.getUrl({
+      name: serviceAccountName,
+    }));
```
Nokel81: Given that `PodDetails` already has been made to use injectables via `withInjectables`. I would prefer if these were not introduced a globals but instead passed in using `withInjectables`.

To do that you should:
1. Add these three apis to the `Dependencies` type above
2. Add `priorityClassApi: di.inject(priorityClassApiInjectable)`, etc... to the `withInjectables` call below
3. Change the usage to `this.props.priorityClassApi`, etc ...

Nokel81 (changes requested):
Thanks for doing this work, just a minor request.

src/renderer/components/+workloads-pods/pod-details.tsx
dex4er: Ok. Done.

Nokel81:
@dex4er Please resolve lint issues and then this LGTM

Nokel81 (approved):

Nokel81:
@dex4er Thanks again!

dex4er:
@Nokel81 Linter clean. I forgot there is `npm run lint:fix`

PR #6378, opened 2022-10-10
Leases view

dex4er:
This is a view for Leases:

image

image

### DIFF
--- a/src/common/front-end-routing/routes/cluster/config/leases/leases-route.injectable.ts
+++ b/src/common/front-end-routing/routes/cluster/config/leases/leases-route.injectable.ts
@@ -1,0 +1,25 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import isAllowedResourceInjectable from "../../../../../utils/is-allowed-resource.injectable";
+import { frontEndRouteInjectionToken } from "../../../../front-end-route-injection-token";
+
+const priorityClassesRouteInjectable = getInjectable({
+  id: "leases",
+
+  instantiate: (di) => {
+    const isAllowedResource = di.inject(isAllowedResourceInjectable, "leases");
+
+    return {
+      path: "/leases",
+      clusterFrame: true,
+      isEnabled: isAllowedResource,
+    };
+  },
+
+  injectionToken: frontEndRouteInjectionToken,
+});
+
+export default priorityClassesRouteInjectable;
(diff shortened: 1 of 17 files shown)


src/common/front-end-routing/routes/cluster/config/leases/leases-route.injectable.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import isAllowedResourceInjectable from "../../../../../utils/is-allowed-resource.injectable";
+import { frontEndRouteInjectionToken } from "../../../../front-end-route-injection-token";
+
+const priorityClassesRouteInjectable = getInjectable({
+  id: "leases",
```
Nokel81, line 10: 1. This variable name probably should be something like "leasesRouteInjectable"
2. This ID (which has to be globally unique) is a bit too generic, could you change it to be "leases-route"?

src/common/front-end-routing/routes/cluster/config/leases/navigate-to-leases.injectable.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import leasesRouteInjectable from "./leases-route.injectable";
+import { navigateToRouteInjectionToken } from "../../../../navigate-to-route-injection-token";
+
+const navigateToPriorityClassesInjectable = getInjectable({
```
Nokel81: Same with this variable name

Nokel81 (commented):
Two minor things and then this LGTM!

src/common/front-end-routing/routes/cluster/config/leases/leases-route.injectable.ts
dex4er: Fixed

src/common/front-end-routing/routes/cluster/config/leases/navigate-to-leases.injectable.ts
dex4er: Fixed

dex4er:
BTW, I'm not sure if this view fits to Config menu but there is nothing better...

Nokel81:
> BTW, I'm not sure if this view fits to Config menu but there is nothing better...

No I think it being part of Config makes sense

Nokel81 (approved):

PR #6385, opened 2022-10-10
Expose a way to add cluster frame components in Extension API

aleksfront:
~New registry for registering html elements into the top of the cluster views. Should be used for modals and dialogs.~

Expose a way to add cluster frame components in Extensions. Usable for e.g. modals and dialogs.


src/extensions/registries/index.ts
```diff
@@ -10,3 +10,4 @@ export * from "./page-menu-registry";
 export * from "./entity-setting-registry";
 export * from "./catalog-entity-detail-registry";
 export * from "./protocol-handler";
+export * from "./modal-registry";
```
Nokel81: This will export the injection token, which isn't usable yet in the extension API. Should we instead just export the `ClusterModalRegistration` type?

src/renderer/cluster-modals/cluster-modals.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+interface Dependencies {
+  clusterModals: ClusterModalRegistration[];
+}
+
+export const NonInjectedClusterModals = ({ clusterModals }: Dependencies) => {
+  return (
+    <div className={styles.clusterModals} style={{ height: 0 }}>
+      {clusterModals.map((modal) => {
+        return modal.visible ? <modal.Component key={modal.id} /> : null;
```
Nokel81: This is wrong, since `visible` is `IComputedValue<boolean>`

Nokel81: It would be great if we had unit tests to verify this behaviour too

Nokel81 (changes requested):

src/renderer/cluster-modals/cluster-modals.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { withInjectables } from "@ogre-tools/injectable-react";
+import React from "react";
+import type { ClusterModalRegistration } from "../../extensions/registries";
+import clusterModalsInjectable from "./cluster-modals.injectable";
+
+interface Dependencies {
+  clusterModals: ClusterModalRegistration[];
+}
+
+export const NonInjectedClusterModals = ({ clusterModals }: Dependencies) => {
```
jansav: Why do we need this? Could we just expose a way to introduce new implementations for `clusterFrameChildComponentInjectionToken`? It would be reusable and has nothing to do with modals.

src/renderer/cluster-modals/cluster-modals-registrator.injectable.ts
```diff
@@ -0,0 +1,25 @@
...
+  instantiate: (): ExtensionRegistrator => {
+    return (ext) => {
+      const extension = ext as LensRendererExtension;
+
+      return extension.clusterModals.map((registration) => {
+        return {
+          id: registration.id,
+          Component: registration.component,
+          visible: registration.visible,
+        }
```
Iku-turso: Have you tried that this really works? Seems to me this map should return an array of injectables, but this is an array of unrelated objects.

Additionally: I really wonder why typing in `extensionRegistratorInjectionToken` doesn't guard against this. Check?

src/features/cluster/cluster-modal-items.test.tsx
```diff
@@ -0,0 +1,63 @@
+/**
```
Iku-turso: This file contains only snapshot tests, which is not enough to unit test this.

src/features/cluster/cluster-modal-items.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+ * Copyright (c) OpenLens Authors. All rights reserved.
```
Iku-turso: The behavioural unit tests do not document behaviour as test names.

src/extensions/registries/index.ts
aleksfront: Removed injection token export from `./modal-registry`

src/renderer/cluster-modals/cluster-modals.tsx
aleksfront: Computed value now respected. Added test case with variable `visible` property.

src/renderer/cluster-modals/cluster-modals.tsx
aleksfront: This new registration basically relates to any components placed right inside `<body>` which is quite useful for any HTML elements with arbitrary position such as tooltips, modals, dialogs etc. Currently there is no good way to preserve component around within various cluster view "snapshots".

src/features/cluster/cluster-modal-items.test.tsx
aleksfront: Made some refactoring and added few more test cases as well.

src/renderer/cluster-modals/cluster-modals-registrator.injectable.ts
aleksfront: Fixed behaviour. Tested both in real-world and in tests as well.

src/renderer/cluster-modals/cluster-modals.tsx
jansav: Yeah, but that's exactly what `clusterFrameChildComponentInjectionToken` does. So we shouldn't create another one?

aleksfront: Thanks @jansav for going thought this case.

Nokel81 (dismissed):

jansav:
Pushed 3 commits:
- Fixed linting
- Update snapshots
- Renamed test file

After these, the implementation didn't match the name of PR anymore, so I updated the PR. 

As you can see, the current implementation is a lot smaller and does exactly same thing.

jansav (approved):

PR #6400, opened 2022-10-12
Add data-origin to status bar items for better observability

aleksfront:
It's now easier to use `[data-origin=myextension]` selectors to find status bar elements with `document.querySelector()` or similar.

data-origin DOM example


Nokel81 (dismissed):

src/renderer/components/status-bar/status-bar-item-registrator.injectable.tsx
```diff
@@ -74,6 +74,7 @@ const toItemInjectableFor = (extension: LensRendererExtension, getRandomId: () =
       id,
 
       instantiate: () => ({
+        id: extension.sanitizedExtensionId,
```
jansav: There's something wrong here. This makes all of the status bar items from extension to have same ID?

jansav: If that's the intended behaviour, then ID is not correct term for it.

jansav (changes requested):
Left single comment

src/renderer/components/status-bar/status-bar-item-registrator.injectable.tsx
aleksfront: Yes, indeed. It should be called an `origin` not `id`.

Nokel81 (approved):

src/renderer/components/status-bar/status-bar-items.injectable.tsx
```diff
@@ -10,9 +10,14 @@ import type { StatusBarItem } from "./status-bar-item-injection-token";
 import { statusBarItemInjectionToken } from "./status-bar-item-injection-token";
 import { computedInjectManyInjectable } from "@ogre-tools/injectable-extension-for-mobx";
 
+interface StatusItem {
+  origin?: string;
+  component: React.ComponentType<StatusBarItemProps>;
```
jansav, line 15: Small things: I would expect this to be `Component` as it's a React.Component.

src/renderer/components/status-bar/status-bar.tsx
```diff
@@ -22,18 +22,24 @@ const NonInjectedStatusBar = observer(({ items }: Dependencies & StatusBarProps)
   const { left, right } = items.get();
 
   return (
-    <div className={styles.StatusBar}>
+    <div className={styles.StatusBar} data-testid="status-bar">
       <div className={styles.leftSide} data-testid="status-bar-left">
         {left.map((Item, index) => (
```
jansav, line 27: Same here, `Item` was capitalized to describe that it is a React.Component (or a class). However, now it's not anymore.

jansav (approved):

PR #6402, opened 2022-10-12
Fix windows shell not having all environment variables

### LINKED ISSUE #6401: Problem with kubectl command
**Describe the bug**
The version of the Lens software I installed is Lens Setup **2022.10.111653-latest.exe**, but there is a problem with the kubectl command

**Expected behavior**
When I execute the shell of the POD the kubectl command is not used

- Lens Version: 2022.10.111653-latest
- OS: Windows 10
exe in Windows 10
**Logs:**
```
PS C:\Users\sumit> kubectl exec -i -t -n default gus-kube-state-metrics-75b4df5c94-fm9z7 -c kube-state-metrics -- sh -c "clear; (bash || ash || sh)"
kubectl : The term 'kubectl' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:1
+ kubectl exec -i -t -n default gus-kube-state-metrics-75b4df5c94-fm9z7 ...
+ ~~~~~~~
    + CategoryInfo          : ObjectNotFound: (kubectl:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```

Nokel81:


src/main/utils/shell-env/compute-shell-environment.injectable.ts
```diff
@@ -0,0 +1,60 @@
...
+  instantiate: (di): ComputeShellEnvironment => {
+    const isWindows = di.inject(isWindowsInjectable);
+    const computeUnixShellEnvironment = di.inject(computeUnixShellEnvironmentInjectable);
+
+    if (isWindows) {
+      return async () => ({
+        callWasSuccessful: true,
+        response: undefined,
+      });
+    }
```
Nokel81, line 26: This is the main part of the code. In #6351 it was changed to return `{}` instead of `process.env` which breaks the usage of this injectable within `ShellSession`.

src/main/shell-session/shell-session.ts
```diff
@@ -331,19 +336,27 @@ export abstract class ShellSession {
...
-    const pathStr = [await this.kubectlBinDirP, ...this.getPathEntries(), process.env.PATH].join(path.delimiter);
+    const shell = this.dependencies.resolvedShell || userInfo().shell;
+    const result = await this.dependencies.computeShellEnvironment(shell);
+    const rawEnv = (() => {
+      if (result.callWasSuccessful) {
+        return result.response ?? process.env;
+      }
+
+      return process.env;
+    })();
```
Nokel81, line 347: Unlike in the `setupShellInjectable` case where we don't do anything if there are no new env vars to handle, here we need to continue.

src/main/utils/shell-env/compute-shell-environment.injectable.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    return async (shell) => {
+      const controller = new AbortController();
+      const shellEnv = computeUnixShellEnvironment(shell, { signal: controller.signal });
+      const cleanup = disposer();
+
+      const timeoutHandle = setTimeout(() => controller.abort(), 30_000);
+
+      cleanup.push(() => clearTimeout(timeoutHandle));
```
jim-docker, line 35: call `cleanup()` somewhere?

PR #6410, opened 2022-10-13
Clickable StorageClass

dex4er:
Hi.

This patch enabled clickable storageClassName in PV and PVC detailed view.

### DIFF
--- a/src/renderer/components/+storage-volume-claims/volume-claim-details.tsx
+++ b/src/renderer/components/+storage-volume-claims/volume-claim-details.tsx
@@ -14,7 +14,7 @@ import { Link } from "react-router-dom";
 import { ResourceMetrics } from "../resource-metrics";
 import { VolumeClaimDiskChart } from "./volume-claim-disk-chart";
 import type { KubeObjectDetailsProps } from "../kube-object-details";
-import { PersistentVolumeClaim } from "../../../common/k8s-api/endpoints";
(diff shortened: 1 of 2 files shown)


src/renderer/components/+storage-volume-claims/volume-claim-details.tsx
```diff
@@ -89,15 +94,21 @@ class NonInjectedPersistentVolumeClaimDetails extends React.Component<Persistent
...
         )}
-        <KubeObjectMeta object={volumeClaim}/>
+        <KubeObjectMeta object={volumeClaim} />
         <DrawerItem name="Access Modes">
           {accessModes?.join(", ")}
         </DrawerItem>
         <DrawerItem name="Storage Class Name">
-          {storageClassName}
+          <Link
+            key="link"
```
aleksfront, line 106: We do not need `key` prop here for a standalone `<Link/>` component.

src/renderer/components/+storage-volumes/volume-details.tsx
```diff
@@ -57,10 +62,16 @@ export class PersistentVolumeDetails extends React.Component<PersistentVolumeDet
           {persistentVolumeReclaimPolicy}
         </DrawerItem>
         <DrawerItem name="Storage Class Name">
-          {storageClassName}
+          <Link
+            key="link"
```
aleksfront, line 66: We do not need `key` prop here for a standalone `<Link/>` component.

aleksfront (approved):

Issue #4895, opened 2022-02-17
Lens metrics fails if the prometheus service address was set


jim-docker:
Under cluster settings => Metrics, if Prometheus is set to other than Auto detect or Lens, the Prometheus service address field is available. If this field is specified and you at some point install/turn on Lens metrics, Lens metrics will fail because it appears to use the previously set Prometheus service address (even though the field is not shown/applicable for Lens or Auto detect).

5.4.0-beta.2 (probably in 5.3.4 too)

This can lead to major confusion when experimenting with metrics setups.

```
error:   ┏ [Metrics]: metrics not available services "no-such-service" not found +3ms
error:   ┃ [1] {
error:   ┃ [2]   kind: 'Status',
error:   ┃ [3]   apiVersion: 'v1',
error:   ┃ [4]   metadata: {},
error:   ┃ [5]   status: 'Failure',
error:   ┃ [6]   reason: 'NotFound',
error:   ┃ [7]   details: { name: 'no-such-service', kind: 'services' },
error:   ┃ [8]   code: 404
error:   ┗ [9] }
```

jasonmacdonald:
I just hit this same bug. Is there any way to clear that out as a short-term fix?

jim-docker:
> I just hit this same bug. Is there any way to clear that out as a short-term fix?

If you are using Lens metrics, on the cluster settings => Metrics page, just switch Prometheus to "Helm", clear the Prometheus service address field and reset Prometheus back to "Lens" or "Auto detect" (whatever it was)

PR #6435, opened 2022-10-20
Fix manually set prometheus service address to work after re-connect

nevalla:
When disconnecting the cluster, prometheus configuration is reset on context handler. However, that is never setup again and when reconnecting the cluster manually set prometheus service address is ignored and auto-detected is used instead. This PR will read prometheus preferences every time prometheus service is resolved.

### DIFF
--- a/src/main/context-handler/context-handler.ts
+++ b/src/main/context-handler/context-handler.ts
@@ -53,6 +53,7 @@ export class ContextHandler implements ClusterContextHandler {
 
   constructor(private dependencies: Dependencies, protected cluster: Cluster) {
     this.clusterUrl = url.parse(cluster.apiUrl);
+    this.setupPrometheus(cluster.preferences);
   }
 
   public setupPrometheus(preferences: ClusterPrometheusPreferences = {}) {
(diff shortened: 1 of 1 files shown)


jim-docker:
https://github.com/lensapp/lens/issues/4895 may be related? I'll test...

nevalla:
> #4895 may be related? I'll test...

looks slightly different although the symptoms are the same.

src/main/context-handler/context-handler.ts
```diff
@@ -53,7 +53,6 @@ export class ContextHandler implements ClusterContextHandler {
 
   constructor(private dependencies: Dependencies, protected cluster: Cluster) {
     this.clusterUrl = url.parse(cluster.apiUrl);
-    this.setupPrometheus(cluster.preferences);
```
jim-docker: Would it hurt to leave this in?

jim-docker (dismissed):
Fixes described issue but not https://github.com/lensapp/lens/issues/4895 :(

src/main/context-handler/context-handler.ts
nevalla: probably not. I'll put it back.

nevalla:
@jim-docker PTAL again

jim-docker (approved):

PR #6444, opened 2022-10-20
Fix AppImage windows not showing the application icon

### LINKED ISSUE #6443: Task bar icon missing on U22.04 / XFCE
**To Reproduce**
Launch lens on Ubuntu 22.04 under XFCE

**Expected behavior**
I'd expect to see lens' icon :)

~A PR is on it's way.~ PR available #6444

**Screenshots**
Current
image

Expected
image

- Lens Version: 6.1.12
- OS: Ubuntu 22.04 amd64 / XFCE

babs:
Fixes #6443, missing task bar icon on some environments (e.g. U22.04 / XFCE)

Might be interesting to use a better quality icon than the tray's one but it's beyond my knowledge

image => image

### DIFF
--- a/src/main/start-main-application/lens-window/application-window/create-electron-window.injectable.ts
+++ b/src/main/start-main-application/lens-window/application-window/create-electron-window.injectable.ts
@@ -9,6 +9,12 @@ import { BrowserWindow } from "electron";
 import type { ElectronWindow } from "./create-lens-window.injectable";
 import type { RequireExactlyOne } from "type-fest";
 import openLinkInBrowserInjectable from "../../../../common/utils/open-link-in-browser.injectable";
+import getAbsolutePathInjectable from "../../../../common/path/get-absolute-path.injectable";
+import lensResourcesDirInjectable from "../../../../common/vars/lens-resources-dir.injectable";
+import isLinuxInjectable from "../../../../common/vars/is-linux.injectable";
+import fsInjectable from "../../../../common/fs/fs.injectable";
+import applicationInformationInjectable from "../../../../common/vars/application-information.injectable";
+
 
 export type ElectronWindowTitleBarStyle = "hiddenInset" | "hidden" | "default" | "customButtonsOnHover";
 
@@ -45,6 +51,10 @@ const createElectronWindowInjectable = getInjectable({
   instantiate: (di): CreateElectronWindow => {
     const logger = di.inject(loggerInjectable);
     const openLinkInBrowser = di.inject(openLinkInBrowserInjectable);
+    const getAbsolutePath = di.inject(getAbsolutePathInjectable);
+    const lensResourcesDir = di.inject(lensResourcesDirInjectable);
+    const isLinux = di.inject(isLinuxInjectable);
+    const applicationInformation = di.inject(applicationInformationInjectable);
 
     return (configuration) => {
       const applicationWindowState = di.inject(
@@ -78,6 +88,23 @@ const createElectronWindowInjectable = getInjectable({
           contextIsolation: false,
         },
       });
+
+      if (isLinux) {
+        const iconFileName = [
+          getAbsolutePath(lensResourcesDir, `../${applicationInformation.name}.png`),
+          `/usr/share/icons/hicolor/512x512/apps/${applicationInformation.name}.png`,
+        ].find(di.inject(fsInjectable).existsSync);
+
+        if (iconFileName != null) {
+          try {
+            browserWindow.setIcon(iconFileName);
+          } catch (err) {
+            logger.warn(`Error while setting window icon ${err}`);
+          }
+        } else {
+          logger.warn(`No suitable icon found for task bar.`);
+        }
+      }
 
       applicationWindowState.manage(browserWindow);
 
--- a/src/common/vars/application-information.global-override-for-injectable.ts
+++ b/src/common/vars/application-information.global-override-for-injectable.ts
@@ -7,6 +7,7 @@ import { getGlobalOverride } from "../test-utils/get-global-override";
 import applicationInformationInjectable from "./application-information.injectable";
 
 export default getGlobalOverride(applicationInformationInjectable, () => ({
+  name: "some-product-name",
   productName: "some-product-name",
   version: "6.0.0",
   build: {},
--- a/src/common/vars/application-information.injectable.ts
+++ b/src/common/vars/application-information.injectable.ts
@@ -5,16 +5,16 @@
 import { getInjectable } from "@ogre-tools/injectable";
 import packageJson from "../../../package.json";
 
-export type ApplicationInformation = Pick<typeof packageJson, "version" | "config" | "productName" | "copyright" | "description"> & {
+export type ApplicationInformation = Pick<typeof packageJson, "version" | "config" | "productName" | "copyright" | "description" | "name"> & {
   build: Partial<typeof packageJson["build"]> & { publish?: unknown[] };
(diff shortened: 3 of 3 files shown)


Nokel81:
Thanks for this PR. However, it is subtly incorrect because `trayIcon` could be one of several icons (checking for updates, etc...).

Furthermore, I think we would only want to set this on linux (and not on macOS or Windows). I will do some investigation to see how easy this is to accommodate.

babs:
I'm still looking around, i'm following the track of `getAbsolutePath(lensResourcesDir, "../open-lens.png")`
I'll take a look to add the is-linux / is-windows / is-mac.

babs:
Got something, give me a moment :)

babs:
Promising...
image

babs:
ok, @Nokel81, I think it's good:
 * uses the bundled open-lens.png
 * only set on linux version
 * check that the file actually exists (in case of icon name change without notice)

src/main/start-main-application/lens-window/application-window/create-electron-window.injectable.ts
```diff
@@ -79,6 +87,14 @@ const createElectronWindowInjectable = getInjectable({
         },
       });
 
+      if (isLinux) {
+        const iconFileName = getAbsolutePath(lensResourcesDir, "../open-lens.png");
+
+        if (existsSync(iconFileName)) {
+          browserWindow.setIcon(iconFileName);
+        }
```
Nokel81: Don't know if this with pass CI...

In any case I think using `pathExistsInjectable` and then having this be

```ts
pathExists(iconFileName)
  .then(iconExists => {
    if (iconExists) {
      browserWindow.setIcon(iconFileName);
    }
  });
```

src/main/start-main-application/lens-window/application-window/create-electron-window.injectable.ts
```diff
@@ -79,6 +87,14 @@ const createElectronWindowInjectable = getInjectable({
         },
       });
 
+      if (isLinux) {
+        const iconFileName = getAbsolutePath(lensResourcesDir, "../open-lens.png");
```
Nokel81: This files doesn't exist in the repo, where does it get generated?

src/main/start-main-application/lens-window/application-window/create-electron-window.injectable.ts
babs: I tried async/promise but in the (unlikely) eventuality of nag screen going away before promise resolution, it might generate an error message but I can encolse it in a try / catch

src/main/start-main-application/lens-window/application-window/create-electron-window.injectable.ts
babs: good question, I thinks it's generated at build time from a svg ?

image

Nokel81: Do you just run `make build`?

babs: I did a `make build`, then run the generated dist AppImage

Nokel81: Ah I think I understand. You are inside the mount that the AppImage produces.

babs: yes, I'm unable to find a "icon asset" injectable :/

src/main/start-main-application/lens-window/application-window/create-electron-window.injectable.ts
babs: Code is going to be

```node
      if (isLinux) {
        const iconFileName = getAbsolutePath(lensResourcesDir, "../open-lens.png");

        pathExists(iconFileName)
          .then(iconExists => {
            if (iconExists) {
              try {
                browserWindow.setIcon(iconFileName);
              } catch (err) {
                logger.warn(`Error while setting window icon ${err}`);
              }
            }
          });
      }
```

src/main/start-main-application/lens-window/application-window/create-electron-window.injectable.ts
babs: I can also test `/usr/share/icons/hicolor/512x512/apps/open-lens.png` should be ok on packages versions but not if the file name itself changes (I'm thinking of the commercial edition)

src/main/start-main-application/lens-window/application-window/create-electron-window.injectable.ts
babs: This part is working. Remains the icon location mystery :)

src/main/start-main-application/lens-window/application-window/create-electron-window.injectable.ts
babs: `${packager.executableName}` in `./node_modules/app-builder-lib/out/targets/fpm.js`, still digging

Nokel81: `const iconFileName = getAbsolutePath(lensResourcesDir, "../open-lens.png");` should be fine.

The only thing that needs to be changed is the file name itself. To get that please do the following:

1. In `src/common/vars/application-information.injectable.ts` change the `ApplicationInformation` type to include the `"name"` field.
2. Add a `name` destructure to fix the type error within that file.
3. Use `applicationInformationInjectable` in `src/main/start-main-application/lens-window/application-window/create-electron-window.injectable.ts` to produce the file name, which should be `../${applicationInformation.name}.png`

babs: I have two paths to test and several options:
 - first is to set one icon location based on `process.env["APPIMAGE"] == null`
 - second is to test two paths but I end up with nested promise which is quite inelegant

Any suggestions ?

babs: Or I go with `di.inject(fsInjectable).existsSync` ?

babs:
```node
        const iconFileName = [
          getAbsolutePath(lensResourcesDir, `../${applicationInformation.name}.png`),
          `/usr/share/icons/hicolor/512x512/apps/${applicationInformation.name}.png`,
        ].find(di.inject(fsInjectable).existsSync);

        if (iconFileName != null) {
          try {
            browserWindow.setIcon(iconFileName);
          } catch (err) {
            logger.warn(`Error while setting window icon ${err}`);
          }
        } else {
          logger.warn(`No suitable icon found for task bar.`);
        }
```
 @Nokel81 what about this one ?

babs:
[80bd3ee](https://github.com/lensapp/lens/pull/6444/commits/80bd3eec7d25af116a57d63664ca7451faa862f4) has been tested with AppImage and deb install

babs:
removed review request is the result of a misclick

src/main/start-main-application/lens-window/application-window/create-electron-window.injectable.ts
Nokel81: That looks file, I would just add some prefix to the log calls, something like `[ELECTRON-WINDOW]:`

babs: Done :)

Nokel81 (approved):
Let's try CI

PR #6441, opened 2022-10-20
Fix links in Readme

Nokel81:
Signed-off-by: Sebastian Malton <sebastian@malton.name>

fixes #6436

PR #6450, opened 2022-10-21
Fix broken links in README

VladMasarik:
I noticed that the links in main REAME point to 404 pages.


Nokel81:
Thanks, but we already have #6441 for that.

PR #6462, opened 2022-10-24
Change log level to file always be debug

Nokel81:

### DIFF
--- a/src/common/logger.ts
+++ b/src/common/logger.ts
@@ -55,7 +55,7 @@ if (ipcMain) {
     transports.push(
       new winston.transports.File({
         handleExceptions: false,
-        level: logLevel,
+        level: "debug",
         filename: "lens.log",
         /**
          * SAFTEY: the `ipcMain` check above should mean that this is only


jim-docker:
Should all `logger.debug()` calls end up in the file? I noticed that [this logging](https://github.com/lensapp/lens/blob/1aabadd60fd6a9b9b1726371c8ffeff262099670/src/main/routes/port-forward/functionality/port-forward.ts#L83) does not appear in the file (unless I start the app in the terminal with `DEBUG=1` - tested on linux)

Nokel81:
@jim-docker Yes that should show up. I found that the file logs show up in `OpenLens` folder (not `OpenLensDev`).

jim-docker:
Hmm, that's where I was looking (I did `make build` and ran the AppImage)

Nokel81:
Are you sure that we were triggering that log line? Because when I do a `make build` and then run it (on macOS) I do see some `debug` lines.

jim-docker:
Yeah, that's what is weird. I saw some debug logging, but that particular port-forward debug logging did not show up in the file till I started the AppImage with DEBUG=1

jim-docker (approved):

Issue #3770, opened 2021-09-10
Events / describe for custom resources


xeor:
**What would you like to be added**:
Events and `describe` info about an customresource

**Why is this needed**:
Many custom resources, like flux2's `helmrelease` have a bunch of useful info in their custom resources.. In the UI currently, there are no way to see these.

- Kubernetes distribution: RKE
- Desktop OS: OSX
- Lens version: 5.2.0-latest.20210908.1

gaeljw:
Indeed, this would be a really useful feature.

We are missing useful info without these events in Lens.

Example:

```
$ kubectl describe ExternalSecret prod-gitlab-token

Name:         prod-gitlab-token
API Version:  external-secrets.io/v1beta1
Kind:         ExternalSecret
...
Events:
  Type    Reason   Age                 From              Message
  ----    ------   ----                ----              -------
  Normal  Updated  45m (x22 over 21h)  external-secrets  Updated Secret
```

But in Lens, nothing displayed.

PR #6468, opened 2022-10-25
Refactor kube detail components and add KubeEvents automatically to all custom resources

### LINKED ISSUE #6452: Custom Resource views should include related events like all inbuilt resources do
Events can be attached to any resource, including custom resources. We use this heavily. We're stuck doing descibes or filtering the event blade to get what we want but only after losing context with the CR. 

**Why is this needed**:
Events for the CR are important to us and maybe others. I would think that they should be shown with any and all resources.

- Kubernetes distribution: [EKS, AKS, GKE, Rancher, ...] AKS
- Desktop OS: <the OS Lens app runs on in your case> Windows 10

Nokel81:
resolves https://github.com/lensapp/lens/issues/6452

Partial fix for #3770


jansav (approved):
Nice.

PR #6502, opened 2022-11-02
Fix syncing shell environment when using fish

### LINKED ISSUE #6492: Unable to connect to my K8s Cluster w/ latest Arm MACOS version
**Describe the bug**

After updating the following error message pops up on attempting to connect to an EKS cluster `E1031 20:32:13.743311 23106 proxy_server.go:147] Error while proxying request: getting credentials: exec: executable aws not found`. Also worth noting that the aws executable is in my path, but not sure why it would be using it to connect to my cluster.

**To Reproduce**
1. Update to latest version on Mac OSX

**Expected behavior**

Connect to the cluster

**Screenshots**

- Lens Version: Lens: 2022.10.311317-latest
- OS: OSX 12.6 
- Installation method (e.g. snap or AppImage in Linux): Mac Install, in app update

**Logs:**
```
E1031 20:32:13.743311 23106 proxy_server.go:147] Error while proxying request: getting credentials: exec: executable aws not found

It looks like you are trying to use a client-go credential plugin that is not installed.

To learn more about this feature, consult the documentation available at:
https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins

getting credentials: exec: executable aws not found

It looks like you are trying to use a client-go credential plugin that is not installed.

To learn more about this feature, consult the documentation available at:
https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins
```

**Kubeconfig:**

I am able to connect fine via any other connection mechanism using my kube config.

**Additional context**
Same issue was reported here in #5676, however, this solution is not applicable to me as my aws executable is in my PATH.

```
# aws version
aws-cli/2.7.31 Python/3.10.8 Darwin/21.6.0 source/arm64 prompt/off
```

Nokel81:
- Add some better logging for the future


src/main/utils/shell-env/compute-unix-shell-environment.injectable.ts
```diff
@@ -38,67 +53,81 @@ const computeUnixShellEnvironmentInjectable = getInjectable({
...
           // zsh (at least, maybe others) don't load RC files when in non-interactive mode, even when using -l (login) option
           : ["-li"],
       };
     };
 
+
     return async (shellPath, opts = {}) => {
-      const runAsNode = process.env["ELECTRON_RUN_AS_NODE"];
-      const noAttach = process.env["ELECTRON_NO_ATTACH_CONSOLE"];
+      const resetEnvPairs = getResetProcessEnv(process.env, [
```
Iku-turso: Side-effect: `process.env`.

src/main/utils/shell-env/env.injectable.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import process from "process";
+
+const processEnvInjectable = getInjectable({
+  id: "process-env",
+  instantiate: () => process.env,
+  causesSideEffects: true,
```
panuhorsmalahti, line 11: Why is this true? Because process.env can be modified? So every value that can be mutated should have `causesSideEffects` true?

src/main/utils/shell-env/compute-unix-shell-environment.injectable.ts
```diff
@@ -7,96 +7,149 @@ import { getInjectable } from "@ogre-tools/injectable";
...
+  const originals = object.entries(overrides).map(([name]) => [name, src[name]] as const);
+
+  return {
+    env: {
+      ...src,
+      ...overrides,
+    },
+    resetEnvPairs: (target) => {
+      for (const [name, orginalValue] of originals) {
+        if (orginalValue) {
```
panuhorsmalahti: I wonder if we should check something like `if (originalValue !== undefined)` to handle falsy values correctly?

src/main/utils/shell-env/env.injectable.ts
Nokel81: `causesSideEffects` is a hint for unit testing only. It means that if this injectable is not overridden in a unit test then it could cause subtle test failures by accidentally sharing data between tests.

src/main/utils/shell-env/compute-unix-shell-environment.injectable.ts
Nokel81: I guess `""` is falsy too, yeah can change that.

src/main/utils/shell-env/compute-unix-shell-environment.injectable.ts
```diff
@@ -7,96 +7,149 @@ import { getInjectable } from "@ogre-tools/injectable";
...
+import type { AsyncResult } from "../../../common/utils/async-result";
 
 export interface UnixShellEnvOptions {
   signal: AbortSignal;
 }
 
-export type ComputeUnixShellEnvironment = (shell: string, opts: UnixShellEnvOptions) => Promise<EnvironmentVariables>;
+export type ComputeUnixShellEnvironment = (shell: string, opts: UnixShellEnvOptions) => Promise<AsyncResult<EnvironmentVariables, string>>;
+
+const getResetProcessEnv = (src: Partial<Record<string, string>>, overrides: Partial<Record<string, string>>): {
```
panuhorsmalahti, line 27: Short description for the function would make this easier to read.

panuhorsmalahti (dismissed):

src/main/utils/shell-env/compute-unix-shell-environment.injectable.ts
Nokel81: Fair, I can add that quickly

PR #6577, opened 2022-11-15
Do not mark terminating namespaces as disabled

### LINKED ISSUE #6560: terminating object must not be disabled
image (1)

When an object is terminating, it must not be disabled because the object still exists and has events etc that are currently not accessible because the details pane can not be opened. workaround: use kubectl and not lens.

Nokel81:

fixes https://github.com/lensapp/lens/issues/6560


jim-docker (approved):

aleksfront (approved):
LGTM

PR #6580, opened 2022-11-15
Remove spdy support 

Nokel81:

The `spdy` library has not been updated in years and we do not need to support it for our functionality.

We cannot upgrade to HTTP2 easily without splitting off our use of WS since HTTP2 does not have UPGRADE support for WS


jakolehm (approved):

Nokel81:
NOTE: hold off on merging until after 6.2.0 is cut

PR #6591, opened 2022-11-16
Make ResourceStack fully injectable 

Nokel81:


src/renderer/kubectl/apply-all.injectable.ts
```diff
@@ -0,0 +1,19 @@
+/**
```
jim-docker, line 1: Nit: name should be `src/renderer/kubectl/kubectl-apply-all.injectable.ts` since the main partner is `src/main/kubectl/kubectl-apply-all.injectable.ts`

jim-docker (approved):

PR #6597, opened 2022-11-17
Fix not being able to install Lens metrics

Nokel81:

### DIFF
--- a/src/main/resource-applier/resource-applier.ts
+++ b/src/main/resource-applier/resource-applier.ts
@@ -129,9 +129,9 @@ export class ResourceApplier {
 
     const args = [
       subCmd,
-      "--kubeconfig", `"${proxyKubeconfigPath}"`,
+      "--kubeconfig", proxyKubeconfigPath,
       ...parentArgs,
-      "-f", `"${tmpDir}"`,
+      "-f", tmpDir,
     ];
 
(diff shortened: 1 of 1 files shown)


src/main/resource-applier/resource-applier.ts
```diff
@@ -129,9 +129,9 @@ export class ResourceApplier {
 
     const args = [
       subCmd,
-      "--kubeconfig", `"${proxyKubeconfigPath}"`,
+      "--kubeconfig", proxyKubeconfigPath,
       ...parentArgs,
-      "-f", `"${tmpDir}"`,
+      "-f", tmpDir,
```
jim-docker, line 134: I guess this explains this when I try to uninstall Lens metrics? (Just confirmed it)
Screen Shot 2022-11-17 at 9 31 36 AM

jim-docker (approved):
this would impact anything that uses resource-applier, not just Lens Metrics

src/main/resource-applier/resource-applier.ts
Nokel81: It does yes

Nokel81:
> this would impact anything that uses resource-applier, not just Lens Metrics

Fair, though I think the most common case would be this

PR #6611, opened 2022-11-18
Upgrade electron to v21.3.0

Nokel81:


Nokel81:
Closing as we don't need to upgrade yet and node-pty is failing (and vscode is still on v19 like we are)

PR #6628, opened 2022-11-22
Handle globalAgent having a single non-array CA

Nokel81:

### DIFF
--- a/src/common/certificate-authorities/inject-system-cas.injectable.ts
+++ b/src/common/certificate-authorities/inject-system-cas.injectable.ts
@@ -22,15 +22,30 @@ const injectSystemCAsInjectable = getInjectable({
     const requestSystemCAs = di.inject(requestSystemCAsInjectionToken);
 
     return async () => {
-      for (const cert of await requestSystemCAs()) {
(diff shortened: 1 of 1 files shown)


src/common/certificate-authorities/inject-system-cas.injectable.ts
```diff
@@ -22,15 +22,30 @@ const injectSystemCAsInjectable = getInjectable({
...
+      if (certs.length === 0) {
+        // Leave the global option alone
+        return;
+      }
+
+      const cas = Array.isArray(globalAgent.options.ca)
+        ? globalAgent.options.ca
+        : globalAgent.options.ca
+          ? [globalAgent.options.ca]
+          : [];
```
jim-docker: Are you sure this does what you want? I never remember the precedence rule for `? :` used with itself and I've seen this construct used incorrectly before, causing a bug. Why not just write it more explicitly?

Nokel81: Yes I am sure that this is correct but I can rewrite it if you want.

jim-docker (approved):

PR #6644, opened 2022-11-24
Improve UX for shell env sync failure

Nokel81:

1. Add a notification when it fails
2. Add a case for when the output does not have the expected shape.


aleksfront (approved):
LGTM

PR #6657, opened 2022-11-28
Fix allowed resources checks on GKE 

Nokel81:

NOTE: This is a rather large PR that fixes the GKE issue but also does some much need refactoring with fixing some circular dependencies.

The first commit fixes the bug on GKE and the rest of the commits work to do the following refactorings:

1. Convert and simplify the retrival of known and allowed resources on the `Cluster` type so that the `shouldShowResource` method is more correct. Namely, just the `apiName` of a resource is not enough to uniquely specify it. Furthermore, this updated naming better describes how the information should be used (it is also what the Kube docs say to use for it).
    1. Furthermore, this leads to a natural addition to disable this check and just show all resources.
1. Expands the `ClusterContext` type to include some helpful functions that are used in several places (more DRY)
3. Make two different contexts, once for namespaced resources (which can work with cluster scoped resources as well) and one just for cluster scoped resources (which is much simpler in implementation). The benefit of this is that it removes a circular dependency with `NamespaceStore`
4. Moves the `KubeObjectStore.context` static to the legacy extension API type and instead passes the context in as a dependency. This removed some more global shared state and permitted the removal of the circular dependency.
5. Remove the last uses of the legacy global `isMac`, etc... 
6. Start moving more things into runnables for bootstrap


src/common/cluster/request-api-resources.injectable.ts
```diff
@@ -0,0 +1,83 @@
...
+  path: string;
+}
+
+const requestApiResourcesInjectable = getInjectable({
+  id: "request-api-resources",
+  instantiate: (di): RequestApiResources => {
+    const k8sRequest = di.inject(k8SRequestInjectable);
+    const logger = di.inject(loggerInjectable);
+
+    return async (cluster) => {
```
Iku-turso, line 27: Code as seen here could probably benefit from [reducing its nesting](https://www.youtube.com/watch?v=CFRhGnuXG-4). Eg. instead of mutating `resourceListGroups` and `kubeApiResources` inside for-loops inside async-IIFEs inside concurrency protected decorator, inside Promise.all, it could use some functional patterns to get something like. `const resourceListGroups = await pipeline(clusters, map(toK8ApiRequestPromise), awaitAll, flatMap(toResourceListGroup));` and `const kubeApiResources = await pipeline(resourceListGroups, map(toGeneralK8RequestCall), map(withApiLimit), map(toCalledFunction), awaitAll, flatMap(toKubeApiResources));`.

Probably made an error somewhere there, but the point is there :)

Iku-turso: That would be a conversation-starter :)

src/common/k8s-api/kube-object.store.ts
```diff
@@ -208,7 +208,11 @@ export abstract class KubeObjectStore<
         }
       }
 
-      return await res ?? [];
+      const items = await res ?? [];
+
+      console.trace("loadItems", this.api, [...items]);
```
Iku-turso: Forgotten debug?

src/common/k8s-api/kube-object.store.ts
```diff
@@ -249,6 +253,8 @@ export abstract class KubeObjectStore<
     try {
       const items = await this.loadItems({ namespaces, reqInit, onLoadFailure });
 
+      console.log("loadAll", this.api, [...items]);
```
Iku-turso: Forgotten debug?

src/common/k8s-api/kube-object.store.ts
```diff
@@ -281,6 +287,8 @@ export abstract class KubeObjectStore<
   protected mergeItems(partialItems: K[], { merge = true, updateStore = true, sort = true, filter = true, namespaces }: MergeItemsOptions): K[] {
     let items = partialItems;
 
+    console.log("mergeItems", this.api, [...partialItems]);
```
Iku-turso: Forgotten debug?

src/extensions/common-api/k8s-api.ts
```diff
@@ -82,6 +83,17 @@ export abstract class KubeObjectStore<
...
 > extends InternalKubeObjectStore<K, A, D> {
+  /**
+   * @deprecated This is no longer used and shouldn't have been every really used
+   */
+  static readonly context = {
+    set: (ctx: ClusterContext) => {
+      console.warn("Setting KubeObjectStore.context is no longer supported");
+      void ctx;
+    },
+    get: () => asLegacyGlobalForExtensionApi(clusterFrameContextForNamespacedResourcesInjectable),
```
Iku-turso, line 94: One day we'll get rid of these.

Iku-turso (approved):

src/common/rbac.ts
```diff
@@ -11,51 +11,190 @@ export type KubeResource =
...
+    namespaced: true,
+  },
+  networkpolicies: {
+    kind: "NetworkPolicy",
+    group: "networking.k8s.io",
+    namespaced: true,
+  },
+  nodes: {
+    kind: "Node",
+    group: "v1",
```
farodin91, line 117: @Nokel81 Should the group be an empty string? It seems only v1 stuff is not showing up for me.

Nokel81: Actually this probably isn't the place that you need to fix but instead the uses of `shouldShowResourceInjectionToken`

Nokel81: But yes they probably should be empty here too

PR #6670, opened 2022-11-29
Fix back navigation in entity settings

Nokel81:
- The bug was that the user would cycle through all the settings viewed


aleksfront (approved):
LGTM

PR #6690, opened 2022-12-01
Make base store non Singleton

### LINKED ISSUE #4884: BaseStore

Nokel81:
resolves https://github.com/lensapp/lens/issues/4884

This PR does the following:
- Removes `Singleton` as a super class of `BaseStore`
- Makes sure extension API doesn't break, while still deprecating the `SIngleton` style static methods
- Removes all the rest of the uses of `Singleton` that are not related to the extension API
- Makes the Store migrations injectable and a dependency of the stores
- Adds support for multiple `runAfter`'s (to fix a exception in renderer)
- Removes legacy global `execHelm` in favour of explicitly using the injectable version.


src/common/utils/random-bytes.injectable.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { randomBytes } from "crypto";
+import { promisify } from "util";
+
+export type RandomBytes = (size: number) => Promise<Buffer>;
+
+const randomBytesInjectable = getInjectable({
```
Iku-turso, line 11: Throwing dice means `causesSideEffects: true` :)

src/common/utils/random-bytes.injectable.ts
```diff
@@ -0,0 +1,16 @@
...
+  id: "random-bytes",
+  instantiate: (): RandomBytes => promisify(randomBytes),
```
Iku-turso, line 13: Is there a reason for this to be promisified? Could be just sync?

Nokel81: No because the interface of `randomBytes` returns its result in a callback.

src/common/utils/random-bytes.injectable.ts
Nokel81: okay

src/extensions/extensions-store/extensions-store.injectable.ts
```diff
@@ -7,14 +7,7 @@ import { ExtensionsStore } from "./extensions-store";
...
 const extensionsStoreInjectable = getInjectable({
   id: "extensions-store",
-
-  instantiate: () => {
-    ExtensionsStore.resetInstance();
-
-    return ExtensionsStore.createInstance();
-  },
-
-  causesSideEffects: true,
```
Iku-turso, line 17: In wonder if there's still side-effects in `ExtensionStore`.

src/common/catalog-entities/web-link.ts
```diff
@@ -34,12 +34,13 @@ export class WebLink extends CatalogEntity<CatalogEntityMetadata, WebLinkStatus,
     // NOTE: this is safe because `onContextMenuOpen` is only supposed to be called in the renderer
     const di = getEnvironmentSpecificLegacyGlobalDiForExtensionApi(Environments.renderer);
```
Iku-turso, line 35: Side-note: `getEnvironmentSpecificLegacyGlobalDiForExtensionApi` -> the part about "ForExtensionApi" hasn't been very true for some time, and we should remove it ;)

src/extensions/extensions-store/extensions-store.injectable.ts
Nokel81: Not anymore, or at least not by the time the PR is finished everything will be injected

src/main/helm/helm-release-manager.ts
```diff
@@ -107,25 +128,41 @@ export async function getValues(name: string, { namespace, all = false, kubeconf
...
+export async function rollback(name: string, namespace: string, revision: number, kubeconfigPath: string): Promise<void> {
+  const result = await execHelm([
     "rollback",
     name,
     `${revision}`,
     "--namespace", namespace,
     "--kubeconfig", kubeconfigPath,
   ]);
+
+  if (!result.callWasSuccessful) {
```
Iku-turso, line 165: Note: there's something brewing elsewhere to remove duplication from error handling, and formalise it. Let's look at that later.

src/common/fs/read-json-sync.injectable.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import fsInjectable from "./fs.injectable";
+
+const readJsonSyncInjectable = getInjectable({
+  id: "read-json-sync",
+  instantiate: (di) => di.inject(fsInjectable).readJsonSync,
```
Iku-turso, line 10: Is this is detached from current `override-fs-with-fakes`? That can throw someone off.

Nokel81: I don't understand what you are saying.

src/common/catalog-entities/web-link.ts
Nokel81: Sure, future

src/common/utils/random-bytes.injectable.ts
Nokel81: The reason why we don't want the sync version is because it will block the entire application until enough entropy is present.

src/features/helm-charts/child-features/preferences/renderer/adding-of-custom-helm-repository/custom-helm-repo.injectable.ts
```diff
@@ -4,11 +4,12 @@
  */
 import { getInjectable, lifecycleEnum } from "@ogre-tools/injectable";
 import { observable } from "mobx";
+import type { HelmRepo } from "../../../../../../common/helm/helm-repo";
 
 const customHelmRepoInjectable = getInjectable({
   id: "custom-helm-repo",
 
-  instantiate: () => observable({
+  instantiate: () => observable.object<HelmRepo>({
```
Iku-turso, line 12: Is this hack required because of `injectable`? Does it mess up typings of transients?

src/main/index.ts
```diff
@@ -7,14 +7,22 @@
...
+import process from "process";
 import * as LensExtensionsCommonApi from "../extensions/common-api";
 import * as LensExtensionsMainApi from "../extensions/main-api";
 import { getDi } from "./getDi";
 import startMainApplicationInjectable from "./start-main-application/start-main-application.injectable";
 
 const di = getDi();
 
-void di.inject(startMainApplicationInjectable);
+(async () => {
```
Iku-turso, line 19: Curiosity: Root-level `await` should work with recent JS/TS without IIFE, probably.

src/main/index.ts
```diff
@@ -7,14 +7,22 @@
...
+  try {
+    await di.inject(startMainApplicationInjectable);
+  } catch (error) {
+    console.error(error);
```
Iku-turso, line 23: Could be `di.inject(logErrorInjectable)(error);`?

src/main/index.ts
```diff
@@ -7,14 +7,22 @@
...
 import * as LensExtensionsMainApi from "../extensions/main-api";
 import { getDi } from "./getDi";
 import startMainApplicationInjectable from "./start-main-application/start-main-application.injectable";
 
 const di = getDi();
 
-void di.inject(startMainApplicationInjectable);
+(async () => {
+  try {
+    await di.inject(startMainApplicationInjectable);
```
Iku-turso: Nit for elsewhere: starting of application should not be inject-time side-effect, but instead `di.inject(startMainApplicationInjectable)` should return a function which deliberately starts the application.

src/renderer/sentry/initialize.injectable.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const initializeSentryReportingInjectable = getInjectable({
+  id: "initialize-sentry-reporting",
+  instantiate: (di) => ({
+    id: "initialize-sentry-reporting",
+    run: () => {
+      // Have to inject this here instead of above so that its dependency on `setupAppPathsInjectable` doesn't throw
+      const initializeSentryReportingWith = di.inject(initializeSentryReportingWithInjectable);
+
+      if (process.isMainFrame) {
```
Iku-turso: Side-effect, and intent of a technicality is unclear. We do have `currently-in-cluster-frame.injectable.ts` which could be the single source of truth for handling `process.isMainFrame`. Be cautious that currently it either doesn't advertise causing side-effects.

src/renderer/sentry/initialize.injectable.ts
```diff
@@ -0,0 +1,28 @@
...
+import setupAppPathsInjectable from "../app-paths/setup-app-paths.injectable";
+import { beforeFrameStartsInjectionToken } from "../before-frame-starts/before-frame-starts-injection-token";
+import { init } from "@sentry/electron/renderer";
+
+const initializeSentryReportingInjectable = getInjectable({
+  id: "initialize-sentry-reporting",
+  instantiate: (di) => ({
+    id: "initialize-sentry-reporting",
+    run: () => {
+      // Have to inject this here instead of above so that its dependency on `setupAppPathsInjectable` doesn't throw
```
Iku-turso: Generally, these kind of temporal dependencies, and hacks around them are undesirable. In future, we could improve this design by using more async dependencies, ie. `injectables` which have `async instantiate`. That would mean that eg. `asyncAppPathsInjectable` (doesn't exist yet) would resolve with paths that are guaranteed to be ready for consumption, without requirement of temporal initialisation hopefully before said consumption.

Reminder: currently we keep doing this temporal sync hack to comply with some old code, which is difficult to change. That might have changed with the introduction of all these new injectables :)

src/common/app-paths/app-paths-state.injectable.ts
```diff
@@ -21,6 +21,8 @@ const appPathsStateInjectable = getInjectable({
       },
 
       set: (newState: AppPaths) => {
+        console.trace("whereamI?");
```
Iku-turso: Forgotten debug log :)

src/common/app-paths/directory-for-user-data/directory-for-user-data.injectable.ts
```diff
@@ -3,11 +3,15 @@
...
  */
 import { getInjectable } from "@ogre-tools/injectable";
-import { appPathsInjectionToken } from "../app-path-injection-token";
+import appPathsInjectable from "../app-paths.injectable";
 
 const directoryForUserDataInjectable = getInjectable({
   id: "directory-for-user-data",
-  instantiate: (di) => di.inject(appPathsInjectionToken).userData,
+  instantiate: (di) => {
+    console.trace("whereamI?");
```
Iku-turso: Forgotten debug :)

src/common/app-paths/app-path-injection-token.ts
```diff
@@ -2,11 +2,6 @@
  * Copyright (c) OpenLens Authors. All rights reserved.
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
-import { getInjectionToken } from "@ogre-tools/injectable";
 import type { PathName } from "./app-path-names";
 
 export type AppPaths = Record<PathName, string>;
-
-export const appPathsInjectionToken = getInjectionToken<AppPaths>({ id: "app-paths-token" });
```
Iku-turso, line 10: Very interesting that this isn't used in renderer any more. I wonder what made it obsolete.

src/common/runnable/run-many-for.ts
```diff
@@ -3,46 +3,81 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
```
Iku-turso, line 3: This commit for a new feature of `Runnable` is unnecessarily difficult to read without accompanying unit tests.

src/common/runnable/run-many-for.ts
```diff
@@ -3,46 +3,81 @@
...
+};
+
+const executeRunnableWith = <Param>(param: Param) => {
+  const finishedIds = observable.set<string>();
+
+  return async (runnable: Runnable<Param>): Promise<void> => {
+    const parentRunnables = [runnable.runAfter].flat().filter(isDefined);
+
+    for (const parentRunnable of parentRunnables) {
+      await when(() => finishedIds.has(parentRunnable.id));
```
Iku-turso: For this algorithm, using `MobX` is a bit too hacky for my preference, and most likely is the wrong tool for the job. Instead of `when`, could there instead be eg. a `Map` with `runnable.id` as key and the related promise as value?

src/common/runnable/run-many-for.ts
```diff
@@ -3,46 +3,81 @@
...
+    finishedIds.add(runnable.id);
+  };
+};
 
 export function runManyFor(di: DiContainerForInjection): RunMany {
   return <Param>(injectionToken: InjectionToken<Runnable<Param>, void>) => async (param: Param) => {
+    const executeRunnable = executeRunnableWith(param);
     const allRunnables = di.injectMany(injectionToken);
-    const rootId = uuid.v4();
-    const getCompositeRunnables = getCompositeFor<Runnable<Param>>({
```
Iku-turso, line 29: Hmm. Did you consider adding support for multiple `parentId`s in `composite`?

src/common/runnable/run-many-for.ts
```diff
@@ -3,46 +3,81 @@
...
+  }
+};
+
+const verifyRunnablesAreDAG = <Param>(runnables: Runnable<Param>[]) => {
+  const rootId = uuid.v4();
+  const runnableGraph = new Map<string, Set<string>>();
+
+  // Build the Directed graph
+  for (const runnable of runnables) {
+    getOrInsertSet(runnableGraph, runnable.id);
```
Iku-turso: `runnableGraph` could be configuration parameter of something called `getOrInsertSetFor` to reduce duplication.

src/common/runnable/run-many-for.ts
```diff
@@ -3,46 +3,81 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
```
Iku-turso, line 4: Blocker: checked the unit tests, and they are not sufficient to cover this key technical feature of Lens.

src/renderer/stores/init-user-store.injectable.ts
```diff
@@ -4,20 +4,25 @@
...
 import initDefaultUpdateChannelInjectable from "../vars/default-update-channel/init.injectable";
 
 const initUserStoreInjectable = getInjectable({
   id: "init-user-store",
-  instantiate: (di) => {
-    const userStore = di.inject(userStoreInjectable);
+  instantiate: (di) => ({
+    id: "init-user-store",
+    run: () => {
+      // This must be done here so as to actually only be instantiated after the dependencies are
```
Iku-turso, line 15: Hmm. The smell of temporal dependency is here again :/ This is tech debt, and the proper fix is most probably to make `initDefaultUpdateChannelInjectable` async along with its own dependencies.

src/common/app-paths/app-paths-state.injectable.ts
```diff
@@ -21,8 +21,6 @@ const appPathsStateInjectable = getInjectable({
       },
 
       set: (newState: AppPaths) => {
-        console.trace("whereamI?");
```
Iku-turso: Minimal nit: stuff like this could be fixupped/interactive-rebased to not appear in PR.

src/common/get-configuration-file-model/get-configuration-file-model.global-override-for-injectable.ts
```diff
@@ -0,0 +1,50 @@
...
+import assert from "assert";
+import { action } from "mobx";
+import path from "path";
+import { getGlobalOverride } from "../test-utils/get-global-override";
+import getConfigurationFileModelInjectable from "./get-configuration-file-model.injectable";
+import type Config from "conf";
+import readJsonSyncInjectable from "../fs/read-json-sync.injectable";
+import writeJsonSyncInjectable from "../fs/write-json-sync.injectable";
+
+export default getGlobalOverride(getConfigurationFileModelInjectable, (di) => {
```
Iku-turso, line 14: That's one complicated fake as global override. It also hides thrown errors with `try/catch`. Sure about this? :)

src/common/fs/write-file-sync.injectable.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import getDirnameOfPathInjectable from "../path/get-dirname.injectable";
+import fsInjectable from "./fs.injectable";
+
+export type WriteFileSync = (filePath: string, contents: string) => void;
+
+const writeFileSyncInjectable = getInjectable({
```
Iku-turso, line 11: Thought: should we permit sync writes and reads at all, as it can cause all kinds of problems (I trust you know)?

We could at least deprecate these new `injectable`s to discourage new use?

src/test-utils/override-fs-with-fakes.ts
```diff
@@ -50,6 +58,8 @@ export const getOverrideFsWithFakes = () => {
       rm: root.promises.rm,
       access: root.promises.access,
       copy: async (src, dest) => { throw new Error(`Tried to copy '${src}' to '${dest}'. Copying is not yet supported`); },
+      ensureDir: async (path, opts) => ensureDirSync(path, opts),
+      ensureDirSync,
```
Iku-turso, line 62: I'm not sure if these should be deprecated, or even left out of `fs.injectable`, as using them is mostly sign of trouble, and/or the responsibility of other abstraction. WDYT?

src/common/get-configuration-file-model/get-configuration-file-model.global-override-for-injectable.ts
```diff
@@ -21,17 +20,20 @@ export default getGlobalOverride(getConfigurationFileModelInjectable, (di) => {
     assert(options.configName, "Missing options.configName");
 
     const configFilePath = path.posix.join(options.cwd, options.configName);
+    let store: object = {};
+
+    try {
+      store = readJsonSync(configFilePath);
+    } catch {
+      // ignore
+    }
```
Iku-turso, line 29: These make me uneasy. You? :)

src/main/base-store/disable-sync-in-ipc-listener.injectable.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { shouldBaseStoreDisableSyncInIpcListenerInjectionToken } from "../../common/base-store/disable-sync";
+
+const shouldBaseStoreDisableSyncInIpcListenerInjectable = getInjectable({
+  id: "should-base-store-disable-sync-in-ipc-listener",
+  instantiate: () => false,
+  injectionToken: shouldBaseStoreDisableSyncInIpcListenerInjectionToken,
```
Iku-turso, line 11: Curious, but I think I get it.

src/renderer/before-frame-starts/runnables/setup-auto-registration.injectable.ts
```diff
@@ -0,0 +1,84 @@
...
+
+        if (!apiManager.getStore(api)) {
+          apiManager.registerStore(new CustomResourceStore(deps, api));
+        }
+      };
+      const autoInitKubeApi = (api: KubeApi) => {
+        apiManager.registerApi(api);
+      };
+
+      autoRegistrationEmitter
```
Iku-turso, line 53: Thought: in a sense, `emitters` and their `listeners`, such as seen here, do something that `Runnable` could do, but without the burden of having to make a big deal about getting listeners "enlisted". That would be the same thing as `register`ing the `injectable`, which is more like a convention.

src/renderer/before-frame-starts/runnables/setup-root-mac-class.injectable.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const setupRootMacClassnameInjectable = getInjectable({
+  id: "setup-root-mac-classname",
+  instantiate: (di) => {
+    const isMac = di.inject(isMacInjectable);
+
+    return {
+      id: "setup-root-mac-classname",
+      run: () => {
+        const rootElem = document.getElementById("app");
```
Iku-turso: Side-effect.

src/main/index.ts
ixrock: BTW if it's not **yet** supported with TS-compiler there is some also workaround with `webpack.config.js`:
```js
return {
   output: {/**/},
    experiments: {topLevelAwait: true },
},
```

src/common/get-configuration-file-model/get-configuration-file-model.global-override-for-injectable.ts
ixrock: _Offtopic_: such a long filename with many meanings kinda hard to perceive `get-configuration-file-model.global-override-for-injectable.ts`. Could we have more clear/simple `(file|class?)` names?

src/renderer/before-frame-starts/tokens.ts
```diff
@@ -0,0 +1,25 @@
...
+
+// NOTE: these are only run when process.isMainFrame === true
+export const evenBeforeMainFrameStartsInjectionToken = getInjectionToken<Runnable>({
+  id: "even-before-main-frame-starts",
+});
+
+// NOTE: these are only run when process.isMainFrame === false
+export const evenBeforeClusterFrameStartsInjectionToken = getInjectionToken<Runnable>({
+  id: "even-before-cluster-frame-starts",
+});
```
ixrock: Since all these tokens located in folder `before-frame-starts` shouldn't we have a convention about their names starting from `beforeFrameStarts[a-Z]+` for example? _(like below one)._ 
Also as benefit would be easier to (auto)import these tokens with IDE by start typing known prefixed names.

src/features/helm-charts/child-features/preferences/renderer/adding-of-custom-helm-repository/custom-helm-repo.injectable.ts
Nokel81: 1. I wanted to change the field `cacheFilePath` to be required instead of optional but I noticed that this injectable wasn't typed as such so the type error was in the wrong place.
2. The typings of `observable` are wonky and just specifying a generic param (such as the code `observable<HelmRepo>(...)` types as `IObservableArray<HelmRepo>` for some reason.

src/main/index.ts
Nokel81: Furthermore, it is only really supported with ES modules and an electron application's entry point cannot be an ESM.

src/main/index.ts
Nokel81: Maybe but since this is failing to start I figured that we don't actually know if logger could be available. This is mostly a last ditch effort to get some information back before exiting.

src/main/index.ts
Nokel81: okay I can change that.

src/renderer/sentry/initialize.injectable.ts
Nokel81: Already changed in future commit.

src/common/app-paths/app-path-injection-token.ts
Nokel81: Because we have a single State implementation that is in common and used to create all the dependent injectables.

src/common/runnable/run-many-for.ts
Nokel81: Will try other solutions

src/common/runnable/run-many-for.ts
Nokel81: No I didn't because Composite (which needs a name change ASAP) is for trees. And I am no longer building trees.

I don't think we should try and make composite support DAGs as that would make it to complex.

src/common/runnable/run-many-for.ts
Nokel81: sounds good

src/renderer/stores/init-user-store.injectable.ts
Nokel81: Nope. It already is async...

That shows that this concern is not correct (at least not fully correct). Furthermore, stating that you want to make it "Async which resolves when complete" is very similar to my previous solution of "InitializableState".

Given that the shape of a `Runnable` is very close to an `Injectable` anyway, (it has an ID, it has a function that gets run at some point). I think it is entirely reasonable to say that the best practice is to just treat it has a way to provide more metadata to an injectable and then say "always inject in the run()".

src/common/get-configuration-file-model/get-configuration-file-model.global-override-for-injectable.ts
Nokel81: Yup, this behaves similar to how we use the actual dependency. Sure I could write is slightly differently, but it is correct enough to have the tests behave as expected.

src/common/fs/write-file-sync.injectable.ts
Nokel81: Yes we have to (for the time being) because all of `getConfigurationFileModel` is `sync` and so are the migrations.

I think it would be good to write our own, but that is for the future.

src/test-utils/override-fs-with-fakes.ts
Nokel81: I think you are confused because our own abstractions of `writeFile`, etc... use these.... so they are certainly required.

src/renderer/before-frame-starts/runnables/setup-auto-registration.injectable.ts
Nokel81: Reminder: this is a hack to preserve behaviour while making sure circular dependencies don't exist.

src/renderer/before-frame-starts/runnables/setup-root-mac-class.injectable.ts
Nokel81: By that definition, so it `render`. So I think your point is moot.

src/renderer/before-frame-starts/tokens.ts
Nokel81: Good point. Will try and make better names

src/common/runnable/run-many-for.ts
Nokel81: Oh okay, will add more then. Thought they were

src/common/runnable/run-many-for.ts
Nokel81: That by itself isn't enough because the order of the runnables are not guaranteed to be in the correct order for that insertion to be guaranteed to happen.

Nokel81: Done

src/common/runnable/run-many-for.ts
Nokel81: Added more tests

src/renderer/before-frame-starts/tokens.ts
Nokel81: How about now?

src/common/fs/write-json-sync.injectable.ts
```diff
@@ -3,11 +3,29 @@
...
+      ensureDirSync,
+    } = di.inject(fsInjectable);
+    const getDirnameOfPath = di.inject(getDirnameOfPathInjectable);
+
+    return (filePath, content) => {
+      ensureDirSync(getDirnameOfPath(filePath), { mode: 0o755 });
+
+      writeJsonSync(filePath, content, {
+        encoding: "utf-8",
+        spaces: 2,
```
Iku-turso, line 25: These could be always set in `writeJsonSync`.

src/renderer/start-frame/start-frame.injectable.ts
```diff
@@ -13,19 +13,19 @@ const startFrameInjectable = getInjectable({
...
-    const evenBeforeClusterFrameStarts = runMany(evenBeforeClusterFrameStartsInjectionToken);
+    const beforeFrameStartsFirst = runMany(beforeFrameStartsFirstInjectionToken);
+    const beforeMainFrameStarts = runMany(beforeMainFrameStartsInjectionToken);
+    const beforeClusterFrameStarts = runMany(beforeClusterFrameStartsInjectionToken);
     const beforeFrameStarts = runMany(beforeFrameStartsInjectionToken);
     const currentlyInClusterFrame = di.inject(currentlyInClusterFrameInjectable);
 
     return async () => {
-      await evenBeforeFrameStarts();
+      await beforeFrameStartsFirst();
```
ixrock, line 23: Which frame starts? _(since below we're seeing distinguishing btw cluster/main frame)._ 
Why we should wait for?  _(just random question for the first time looking at the code)_

src/renderer/before-frame-starts/tokens.ts
ixrock: Looking good.

src/renderer/components/test-utils/get-application-builder.tsx
```diff
@@ -179,6 +182,15 @@ export const getApplicationBuilder = () => {
 
   overrideFsWithFakes(mainDi);
 
+  // Set up ~/.kube as existing as a folder
+  {
+    const { ensureDirSync } = mainDi.inject(fsInjectable);
+    const joinPaths = mainDi.inject(joinPathsInjectable);
+    const homeDirectoryPath = mainDi.inject(homeDirectoryPathInjectable);
+
+    ensureDirSync(joinPaths(homeDirectoryPath, ".kube"));
```
Iku-turso, line 191: Does this access actual filesystem or the fake one? The actual one is not to be accessed in a unit test.

src/main/index.ts
Iku-turso: Roger.

src/common/runnable/run-many-for.test.ts
```diff
@@ -340,4 +342,315 @@ describe("runManyFor", () => {
...
+    });
+
+    it("should run 'runnable-1'", () => {
+      expect(runMock).toBeCalledWith("runnable-1");
+    });
+
+    it("should run 'runnable-2'", () => {
+      expect(runMock).toBeCalledWith("runnable-2");
+    });
+
```
Iku-turso, line 461: Missing test: "should not run 'runnable-3' yet".

src/common/fs/write-json-sync.injectable.ts
Nokel81: What do you mean? This is `wrieJsonSync`

src/renderer/start-frame/start-frame.injectable.ts
Nokel81: Either frame, this function is here to consolidate all the initialisation steps (and the temporal dependencies between them). The reason why we split up these `Runnable`'s (which is the type we created to describe each of these steps) into separate phases (which is what each of these `runMany` calls represents) is because sometimes there is a general need to have a step complete before anything else is run.

We could have instead specified a `runAfter` for every other runnable (which is how we specify the temporal ordering dependencies) but that seemed to be too repetitive.

src/renderer/components/test-utils/get-application-builder.tsx
Nokel81: This accesses the fake one because we have already overridden `fsInjectable` for `mainDi` above on line 183

src/common/get-configuration-file-model/get-configuration-file-model.global-override-for-injectable.ts
Iku-turso: At least global-override-for-yaddayaddayadda will get shorter soon enough. Outside of that, naming should follow common sense, or what is the proposal? :)

Iku-turso (approved):

PR #6693, opened 2022-12-01
Add authentication requirement for using LensProxy

Nokel81:


Nokel81:
NOTE: This PR is not yet ready for review as I will need to also add HTTPS support since https://github.com/kubernetes/kubectl/issues/744 blocks using authentication of any sort over HTTP

PR #6715, opened 2022-12-08
Include bundled extensions in package.json

jakolehm:
- enhanced security because bundled extensions cannot be tampered
- faster boot

Related to #6749


Nokel81 (dismissed):

package.json
```diff
@@ -290,7 +287,11 @@
     "winston": "^3.8.2",
     "winston-transport-browserconsole": "^1.0.5",
     "ws": "^8.11.0",
-    "xterm-link-provider": "^1.3.1"
+    "xterm-link-provider": "^1.3.1",
+    "kube-object-event-status": "file:./extensions/kube-object-event-status",
+    "metrics-cluster-feature": "file:./extensions/metrics-cluster-feature",
+    "node-menu": "file:./extensions/node-menu",
+    "pod-menu": "file:./extensions/pod-menu"
```
ixrock: I'm wondering why those are still extensions if needs to be always bundled?

ixrock (dismissed):

ixrock (approved):

PR #6717, opened 2022-12-08
Unify build fs layout

jakolehm:
Simplifies things.

### DIFF
--- a/src/common/vars/static-files-directory.global-override-for-injectable.ts
+++ b/src/common/vars/static-files-directory.global-override-for-injectable.ts
@@ -1,0 +1,9 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getGlobalOverride } from "../test-utils/get-global-override";
+import staticFilesDirectoryInjectable from "./static-files-directory.injectable";
+
+
+export default getGlobalOverride(staticFilesDirectoryInjectable, () => "some-static-directory");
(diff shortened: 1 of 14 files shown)


src/common/vars/static-files-directory.injectable.ts
```diff
@@ -4,16 +4,14 @@
...
 
 const staticFilesDirectoryInjectable = getInjectable({
   id: "static-files-directory",
 
   instantiate: (di) => {
     const joinPaths = di.inject(joinPathsInjectable);
-    const lensResourcesDir = di.inject(lensResourcesDirInjectable);
 
-    return joinPaths(lensResourcesDir, "static");
+    return joinPaths(__dirname, "..");
```
Iku-turso: Side-effect :)

jakolehm: Right, you mean `__dirname` here? Just wondering what would be the proper injectable to get the same result.

jakolehm: Fixed.

src/common/vars/static-files-directory.global-override-for-injectable.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getGlobalOverride } from "../test-utils/get-global-override";
+import staticFilesDirectoryInjectable from "./static-files-directory.injectable";
+
+
+export default getGlobalOverride(staticFilesDirectoryInjectable, () => "some-static-directory");
```
Nokel81: This needs to be an absolute path

Nokel81 (changes requested):

src/common/vars/static-files-directory.global-override-for-injectable.ts
Nokel81: It could just be `/some-static-directory`

jakolehm: Fixed.

Nokel81 (approved):

Nokel81 (approved):

PR #6722, opened 2022-12-09
Allow to import app as a library

jakolehm:
Allows to use OpenLens as a library (via npm).

### Example

```ts
// src/main/index.ts

import { createContainer } from "@ogre-tools/injectable";
import { createApp, extensionApi } from "@k8slens/open-lens/main";

const di = createContainer("main");
const app = createApp({
  di,
  mode: process.env.NODE_ENV || "development"
});

app.start().catch((error) => {
  console.error(error);
  process.exit(1);
})

const { Mobx, LensExtensions, Pty } = extensionApi;

export {
  Mobx,
  LensExtensions,
  Pty
};
```

```ts
// src/renderer/index.ts

import "@k8slens/open-lens/styles";
import { createContainer } from "@ogre-tools/injectable";
import { createApp, extensionApi } from "@k8slens/open-lens/renderer";

const di = createContainer("renderer");
const app = createApp({
  di,
  mode: process.env.NODE_ENV || "development"
});

app.start();

const {
  React,
  ReactDOM,
  ReactRouter,
  ReactRouterDom,
  Mobx,
  MobxReact,
  LensExtensions
} = extensionApi;

export {
  React,
  ReactDOM,
  ReactRouter,
  ReactRouterDom,
  Mobx,
  MobxReact,
  LensExtensions
};

```

Requires #6717, #6715 and #6775


src/main/register-injectables.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import type { DiContainer } from "@ogre-tools/injectable";
+import { autoRegister } from "@ogre-tools/injectable-extension-for-auto-registration";
+import { registerMobX } from "@ogre-tools/injectable-extension-for-mobx";
+import { runInAction } from "mobx";
+import { Environments, setLegacyGlobalDiForExtensionApi } from "../extensions/as-legacy-globals-for-extension-api/legacy-global-di-for-extension-api";
+
+export function registerInjectables(di: DiContainer) {
+  setLegacyGlobalDiForExtensionApi(di, Environments.main);
```
Iku-turso, line 12: Should there also be a counterpart for this in renderer?

webpack/library.ts
```diff
@@ -0,0 +1,116 @@
+/**
```
Iku-turso, line 1: More explanatory name than "library" would serve well? `build-of-extensions.ts`? Also it would be better to make the extensions responsible for getting built, if I'm not missing something.

src/library.ts
```diff
@@ -0,0 +1,35 @@
...
+  return di.inject(startMainApplicationInjectable);
+}
+
+export { 
+  registerInjectables,
+  startApp,
+  afterApplicationIsLoadedInjectionToken,
+  beforeApplicationIsLoadingInjectionToken,
+  beforeElectronIsReadyInjectionToken,
+  onLoadOfApplicationInjectionToken,
```
Iku-turso: Note: interestingly, these `injectionTokens` will later be exports of more specific `Features`, probably "application-life-cycle". Before that, what is here is perfectly fine.

src/library.ts
```diff
@@ -0,0 +1,35 @@
...
+
+interface AppConfig {
+  di: DiContainer;
+  staticPath: string;
+}
+
+function startApp(conf: AppConfig) {
+  const { di, staticPath } = conf;
+
+  di.override(staticFilesDirectoryInjectable, () => staticPath);
```
Iku-turso: Until now, `override` has been a concept only relevant to unit testing. I'm not sure if we want to enforce that. I think this line of code is ok for now, but it is likely to change when a better way to have correct `staticFilesDirectory` emerges. Let's omit thinking that for now, as there's bigger fish to fry.

src/main/register-injectables.ts
jakolehm: Probably but it is going to be a bit tricky because main process spawns renderer (== renderer gets initialized in a separate process). Basically to make it happen library should not bundle renderer at all, the importing application would need to build custom version of a renderer bundle and give that as source to app window.

src/library.ts
jakolehm: Fixed.

src/main/index.ts
```diff
@@ -5,29 +5,12 @@
...
-import { spawn } from "node-pty";
-import * as LensExtensionsCommonApi from "../extensions/common-api";
-import * as LensExtensionsMainApi from "../extensions/main-api";
 import { getDi } from "./getDi";
 import startMainApplicationInjectable from "./start-main-application/start-main-application.injectable";
+import { Mobx, LensExtensions, Pty } from "./extension-api";
 
 const di = getDi();
 
 void di.inject(startMainApplicationInjectable);
```
Iku-turso: Looks like an inject-time side-effect, because function call is missing. Also, shouldn't we invoke the new `startApp` here in `index.ts` to guarantee identical behaviour with clients of "library"?

src/renderer/library.ts
```diff
@@ -0,0 +1,15 @@
+/**
```
Iku-turso, line 1: This `library.ts` could share interface with its main counterpart.

jakolehm: They are now similar although still not exactly the same.

package.json
```diff
@@ -22,12 +44,12 @@
     "dev-run": "nodemon --watch ./static/build/main.js --exec \"electron --remote-debugging-port=9223 --inspect .\"",
     "dev:main": "yarn run compile:main --watch --progress",
     "dev:renderer": "yarn run ts-node webpack/dev-server.ts",
+    "compile-library": "env NODE_ENV=production yarn run webpack --config webpack/library-bundle.ts",
     "compile": "env NODE_ENV=production concurrently yarn:compile:*",
     "compile:main": "yarn run webpack --config webpack/main.ts",
     "compile:renderer": "yarn run webpack --config webpack/renderer.ts",
     "compile:extension-types": "yarn run webpack --config webpack/extensions.ts",
     "compile:node-fetch": "yarn run webpack --config ./webpack/node-fetch.ts",
-    "postinstall": "yarn run compile:node-fetch",
```
Nokel81, line 30: Why was this removed?

src/common/vars/application-information-token.injectable.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+import { getInjectionToken } from "@ogre-tools/injectable";
+import type packageJson from "../../../package.json";
+
+export type ApplicationInformation = Pick<typeof packageJson, "version" | "config" | "productName" | "copyright" | "description" | "name"> & {
+  build: { publish?: unknown[] };
+};
+
+const applicationInformationToken = getInjectionToken<ApplicationInformation>({
```
Nokel81: Note: token files don't need to be named `*.injectable.*`

src/renderer/frames/cluster-frame/cluster-frame-layout-child-component.injectable.tsx
```diff
@@ -9,51 +9,62 @@ import { MainLayout } from "../../components/layout/main-layout";
...
+    </MainLayout>
+  );
+});
+
+const ClusterFrameLayout = withInjectables<Dependencies>(NonInjectedClusterFrameLayout, {
+  getProps: (di, props) => ({
+    ...props,
+    currentRouteComponent: di.inject(currentRouteComponentInjectable),
+    startUrl: di.inject(startUrlInjectable),
+    currentPath: di.inject(currentPathInjectable),
```
Iku-turso, line 56: Ok, so this fixes the injection cycle by deferring injection to only until the component is rendered, which is correct. We remain a bit baffled why the cycle happened only now, and not long before. Curious.

Iku-turso: I think this should be merged to master as well in separate PR.

Nokel81: I am especially curious because the error states a cycle with an edge that didn't actually exist.

package.json
jakolehm: `postinstall` does not really work when lens is consumed via a npm package. I need to figure out where this can be executed automatically.

src/main/index.ts
jakolehm: I believe this is already fixed.

src/common/vars/application-information-token.injectable.ts
jakolehm: Good catch, will fix.

src/main/register-injectables.ts
jakolehm: Already fixed.

package.json
jakolehm: Now it's done via `prepare` hook.

src/renderer/frames/cluster-frame/cluster-frame-layout-child-component.injectable.tsx
jakolehm: @Nokel81 are we going to split this into a separate PR?

src/renderer/start-app.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export function startApp(conf: AppConfig) {
+  const { di, mode } = conf;
+
+  if (mode) {
+    const environmentVariables = di.inject(environmentVariablesInjectable);
+
+    environmentVariables.NODE_ENV = mode;
+  }
```
jansav: This doesn't seem right. I don't think we should be mutating environment variables in this sense. 

As a temporary solution, how about making `isProduction` and `isDevelopment`  inject `environmentInjectionToken` which is registered here? 

In the end, we should remove `isProduction` and `isDevelopment` for good. Those are used in places where we need some kind of configuration from the client. 

E.g. `createKubeApiForClusterInjectable` uses `isDevelopment` to determinate whether debugging should be enabled or not. 

How about we introduce `kubeApiDebuggingIsEnabledInjectionToken` in the library, but require client to register implementation for it. 

In that way, the client has full control over the configuration and we as a library do not depend from arbitrary environment variables at all.

jakolehm: > As a temporary solution, how about making isProduction and isDevelopment inject environmentInjectionToken which is registered here?

I actually had it like that first, it felt also really weird... anyway I can change this.

package.json
```diff
@@ -411,5 +446,64 @@
...
+    "@types/react-beautiful-dnd": "^13.1.3",
+    "@types/react-dom": "^17.0.16",
+    "@types/react-router": "^5.1.19",
+    "@types/react-router-dom": "^5.3.3",
+    "@types/react-table": "^7.7.12",
+    "@types/react-virtualized-auto-sizer": "^1.0.1",
+    "@types/react-window": "^1.8.5",
+    "@types/readable-stream": "^2.3.13",
+    "@types/request": "^2.48.7",
+    "@types/request-promise-native": "^1.0.18",
```
jansav: Hard to see that all of these are actually `peerDependencies`. `peerDependency` is something that is required to be installed when using already compiled library and e.g. `@async-fn/jest` is only used for development. 

There might be something that I don't see here and I don't think this is that important yet, so let's figure out these later.

src/main/start-app.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export async function startApp(conf: AppConfig) {
+  const { di, mode } = conf;
+
+  if (mode) {
+    const environmentVariables = di.inject(environmentVariablesInjectable);
+
+    environmentVariables.NODE_ENV = mode;
+  }
```
jansav: Same comment here than in `src/renderer/start-app.ts`

package.json
jakolehm: The thing is that when a library is compiled it won't include dependent declarations (https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#packaging-dependent-declarations). I can try to narrow down which of the types can be removed from peerDependencies. It's also a bit unfortunate that we need to duplicate these (dev vs peer), switching from yarn to npm might make this a bit cleaner.

jansav: I'd say it's fine for now. Let's clean these later.

Iku-turso: > switching from yarn to npm might make this a bit cleaner.

How so? For curious am I ;)

jakolehm: > > switching from yarn to npm might make this a bit cleaner.
> 
> How so? For curious am I ;)

NPM >= 7 installs peerDependencies automatically, yarn does not allow to install local peerDeps. I think this would mean that we would not need to duplicate packages to both dev & peer.

jakolehm: Fixed to include minimal list of peerDependencies (without these client/consumer will start to spit type errors).

src/common/utils/environment-variables.injectable.ts
```diff
@@ -3,14 +3,15 @@
...
 
 const environmentVariablesInjectable = getInjectable({
   id: "environment-variables",
 
-  instantiate: () => {
+  instantiate: (di) => {
     // IMPORTANT: The syntax needs to be exactly this in order to make environment variable values
     // hard-coded at compile-time by Webpack.
-    const NODE_ENV = process.env.NODE_ENV;
+    const NODE_ENV = di.inject(nodeEnvInjectionToken);
```
Iku-turso: This makes the above "IMPORTANT" comment inaccurate, as the line is no longer handled in a special way by Webpack.

jansav: I think this should be removed from here and used directly from `di.inject(nodeEnvInjectionToken)`. Since the source of `nodeEnv` is no longer environment variable, which is against of what this injectable does.

src/renderer/frames/cluster-frame/cluster-frame-layout-child-component.injectable.tsx
Nokel81: Sure, I can make this a separate PR

src/common/utils/environment-variables.injectable.ts
jakolehm: Fixed.

jakolehm: Fixed.

src/renderer/start-app.ts
jakolehm: Fixed.

src/renderer/frames/cluster-frame/cluster-frame-layout-child-component.injectable.tsx
Iku-turso: > I am especially curious because the error states a cycle with an edge that didn't actually exist.

This is true, and we now have some understanding about why this is, and the solution is in the works.

Iku-turso (approved):

PR #6733, opened 2022-12-12
Using custom properties for z-index values

aleksfront:
Create new and manage available stacking contexts for new elements became difficult. It's hard to select proper `z-index` value for the element which cover other elements because there are many `z-indexes` may overlap each other.

Providing css custom properties in `app.scss` in order to fight with arbitrary `z-index` values spread across codebase. It is encouraged to use these properties for any new pages or elements if needed.

```scss
// app.scss
--z-index-base: 0;
--z-index-above: 1;
--z-index-modals: 2;
--z-index-menus: 3;
--z-index-topmost: 9999;
```
settings
hotbar

https://user-images.githubusercontent.com/9607060/207019704-55fee782-3cd7-4fcc-b343-1b1256a3d641.mov

namespace filter

### DIFF
--- a/src/renderer/components/+namespaces/namespace-select-filter.scss
+++ b/src/renderer/components/+namespaces/namespace-select-filter.scss
@@ -33,7 +33,6 @@
       &::before, &::after {
         content: ' ';
         position: absolute;
-        z-index: 20;
         display: block;
         width: 8px;
         height: var(--font-size);
--- a/src/renderer/components/+welcome/welcome.scss
+++ b/src/renderer/components/+welcome/welcome.scss
@@ -7,7 +7,7 @@
   text-align: center;
   width: 100%;
   height: 100%;
-  z-index: 1;
+  z-index: var(--z-index-base);
 
   h2 {
     color: var(--textColorAccent);
--- a/src/renderer/components/cluster-manager/cluster-manager.scss
+++ b/src/renderer/components/cluster-manager/cluster-manager.scss
@@ -27,7 +27,7 @@
   }
 
   .error {
-    z-index: 1;
+    z-index: var(--z-index-base);
   }
 
   #lens-views {
--- a/src/renderer/components/dialog/dialog.scss
+++ b/src/renderer/components/dialog/dialog.scss
@@ -12,7 +12,7 @@
   width: 100%;
   height: 100%;
   padding: $unit * 5;
-  z-index: $zIndex-dialog;
+  z-index: var(--z-index-middle);
   overscroll-behavior: none; // prevent swiping with touch-pad on MacOSX
   overflow: auto;
 
--- a/src/renderer/components/dialog/dialog.scss
+++ b/src/renderer/components/dialog/dialog.scss
@@ -12,7 +12,7 @@
   width: 100%;
   height: 100%;
   padding: $unit * 5;
-  z-index: var(--z-index-middle);
+  z-index: var(--z-index-modals);
   overscroll-behavior: none; // prevent swiping with touch-pad on MacOSX
   overflow: auto;
 
--- a/src/renderer/components/dock/dock.scss
+++ b/src/renderer/components/dock/dock.scss
@@ -26,7 +26,7 @@
       right: 0;
       left: 0;
       bottom: 0;
-      z-index: 100;
+      z-index: var(--z-index-base);
     }
   }
 
--- a/src/renderer/components/hotbar/hotbar-menu.scss
+++ b/src/renderer/components/hotbar/hotbar-menu.scss
@@ -12,6 +12,7 @@
   padding-top: 1px;
   width: var(--hotbar-width);
   overflow: hidden;
+  isolation: isolate;
 
   &.draggingOver::after {
     content: " ";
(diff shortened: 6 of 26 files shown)


src/renderer/components/+namespaces/namespace-select-filter.scss
```diff
@@ -33,7 +33,6 @@
       &::before, &::after {
         content: ' ';
         position: absolute;
-        z-index: 20;
```
ixrock, line 36: Shouldn't this to be replaced with something like:

```css
z-index: var(--z-index-menus);
```

src/renderer/components/+welcome/welcome.scss
```diff
@@ -7,7 +7,7 @@
   text-align: center;
   width: 100%;
   height: 100%;
-  z-index: 1;
+  z-index: var(--z-index-base);
```
ixrock, line 10: What means base? Base should be `z-index: 0` which is by default for all elements?

src/renderer/components/cluster-manager/cluster-manager.scss
```diff
@@ -27,7 +27,7 @@
   }
 
   .error {
-    z-index: 1;
+    z-index: var(--z-index-base);
```
ixrock, line 30: Seems like it's used with `position: relative` and this name probably more fits:

```suggestion
    z-index: var(--z-index-above);
```

src/renderer/components/dialog/dialog.scss
```diff
@@ -12,7 +12,7 @@
   width: 100%;
   height: 100%;
   padding: $unit * 5;
-  z-index: $zIndex-dialog;
+  z-index: var(--z-index-middle);
```
ixrock:
```suggestion
  z-index: var(--z-index-dialog);
```

src/renderer/components/dock/dock.scss
```diff
@@ -26,7 +26,7 @@
       right: 0;
       left: 0;
       bottom: 0;
-      z-index: 100;
+      z-index: var(--z-index-base);
```
ixrock, line 29: Could we avoid using base as keeping it's as default (e.g. `z-index` == 0)?

src/renderer/components/hotbar/hotbar-menu.scss
```diff
@@ -12,6 +12,7 @@
   padding-top: 1px;
   width: var(--hotbar-width);
   overflow: hidden;
+  isolation: isolate;
```
ixrock, line 15: How does it corresponds with `z-index`?

src/renderer/components/layout/main-layout.module.scss
```diff
@@ -11,7 +11,7 @@
   grid-template-rows: [contents] 1fr [footer] auto;
   grid-template-columns: [sidebar] var(--sidebar-width) [contents] 1fr;
   width: 100%;
-  z-index: 1;
+  z-index: var(--z-index-base);
```
ixrock, line 14: Looks again that meaning of this should not correlate with some base, but something that works with `position: relative` to put child above another with `position: static` flow. 
So, maybe better `z-index: var(--z-index-above)` or give a better name to `parent:relative + child:above` combination. Or I'm missing something?

src/renderer/components/layout/setting-layout.scss
```diff
@@ -10,7 +10,7 @@
   display: grid;
   color: var(--settingsColor);
   position: fixed;
-  z-index: 13!important;
+  z-index: var(--z-index-middle);
```
ixrock: Where is correlation between some `Layout` and visibility in the middle? Why the magic number was 13?

ixrock (dismissed):
Idea for better management of UI layers LGTM, some naming changes might require to apply.

src/renderer/components/+namespaces/namespace-select-filter.scss
aleksfront: Works nice without any `z-index` specified.

src/renderer/components/layout/setting-layout.scss
aleksfront: Settings Layout being placed **above** main layout so it should have some bigger `z-index` number.

Number `13` previously used to overwrite specified z-indexes for hotbar items. Hotbar uses max value `12` (This is used to drag-n-drop purposes).

Adding some very low `z-index` + `isolation: isolate` over hotbar solves the issue.

src/renderer/components/hotbar/hotbar-menu.scss
aleksfront: It creates a new [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) for the hotbar items which means any `z-index` values set for elements **within** hotbar menu don't affect anything outside of it. This is basically needed for showing `.SettingsLayout` properly by covering hotbar items.

src/renderer/components/+welcome/welcome.scss
aleksfront: Default value for all elements is `z-index: auto`. Created new set of variables. Now `--z-index-base: 0`.

ixrock (approved):
LGTM

PR #6734, opened 2022-12-12
Remove blue borders in light themed tables

### LINKED ISSUE #6711: Lens TableHead component renders with blue bars top/bottom in light theme only
**Describe the bug**

The native Lens `TableHead` component produces blue bars only in light theme mode:

image

Looks good in dark theme:

image

We're wanting to use this component for a custom cluster page in https://github.com/Mirantis/lens-extension-cc

**To Reproduce**

1. Open a cluster.
2. Go to the Nodes page.
3. Switch to light theme.

**Expected behavior**

There are no blue bars in light theme.

- Lens Version: 2022.12.11410-latest
- OS: macOS 12.6.1 (21G217) (M1)

**Additional context**

Looks like these are the offending styles in light theme:

image

vs in dark theme:

image

aleksfront:
Resolves #6711

Making `TableHead` light and dark coloring more consistent.

table dark
table light

### DIFF
--- a/src/renderer/themes/lens-dark.ts
+++ b/src/renderer/themes/lens-dark.ts
@@ -45,8 +45,6 @@ const lensDarkTheme: LensTheme = {
     "tableBgcStripe": "#2a2d33",
     "tableBgcSelected": "#383c42",
     "tableHeaderBackground": "#262b2f",
-    "tableHeaderBorderWidth": "1px",
-    "tableHeaderBorderColor": "#36393e",
     "tableHeaderColor": "#ffffff",
     "tableSelectedRowColor": "#ffffff",
     "helmLogoBackground": "#ffffff",
--- a/src/renderer/components/table/table-head.scss
+++ b/src/renderer/components/table/table-head.scss
@@ -4,18 +4,14 @@
  */
 
 .TableHead {
-  $border: 1px solid var(--tableHeaderBorderColor);
-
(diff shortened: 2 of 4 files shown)


ixrock (approved):

src/renderer/themes/lens-dark.ts
```diff
@@ -45,8 +45,6 @@ const lensDarkTheme: LensTheme = {
     "tableBgcStripe": "#2a2d33",
     "tableBgcSelected": "#383c42",
     "tableHeaderBackground": "#262b2f",
-    "tableHeaderBorderWidth": "1px",
-    "tableHeaderBorderColor": "#36393e",
```
stefcameron, line 49: @aleksfront I'm concerned here you just eliminated 2 theme CSS vars, which would technically be a breaking change for any extension builders like us out there who might be using these variables for whatever purpose. While you don't need them in the TableHead component anymore, I don't think it's right to just remove them from theme altogether, at least not until Lens 7, assuming Lens is still following semantic versioning rules for the sake of extension developers...

stefcameron:
Thanks for the fix! Just concerned about a possible breaking change...

src/renderer/themes/lens-dark.ts
aleksfront: I believe usage of these variables might be extremely rare. However, I agree this might be considered as a breaking change and we should restore them.

PR #6740, opened 2022-12-13
Renderer logging to file

samitiilikainen:
Related to https://github.com/lensapp/support-lens-extension/issues/118

### DIFF
--- a/src/renderer/logger/add-file-logging.injectable.ts
+++ b/src/renderer/logger/add-file-logging.injectable.ts
@@ -1,0 +1,29 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { transports } from "winston";
+import type winston from "winston";
+import directoryForLogsInjectable from "../../common/app-paths/directory-for-logs.injectable";
+import loggerInjectable from "../../common/logger.injectable";
+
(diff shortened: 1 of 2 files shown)


src/renderer/logger/add-file-logging.injectable.ts
```diff
@@ -0,0 +1,29 @@
...
+import directoryForLogsInjectable from "../../common/app-paths/directory-for-logs.injectable";
+import loggerInjectable from "../../common/logger.injectable";
+
+const addFileLoggingInjectable = getInjectable({
+  id: "add-renderer-file-logger-transport",
+  causesSideEffects: true,
+  instantiate: (di) => {
+    const logger = di.inject(loggerInjectable) as winston.Logger;
+
+    return logger.add(new transports.File({
```
jakolehm, line 17: Just a note that writing logs to filesystem from renderer caused issues in the past. Not sure if it is still the case...

Nokel81: That is a good point, do we know if this transport uses async or sync FS calls?

samitiilikainen: Winston file transport is using async `fs.createWriteStream`

Nokel81: I think after #6690 is merged this will be simpler.

Also, I think that we should include the `clusterId` in the file name so that we don't have multiple frames writing to the same file.

PR #6764, opened 2022-12-15
Fix typing of ExtensionStore.createInstance static

Nokel81:

Regression from #6690

### DIFF
--- a/src/extensions/extension-store.ts
+++ b/src/extensions/extension-store.ts
@@ -27,12 +27,12 @@ export interface ExtensionStoreParams<T extends object> extends BaseStoreParams<
 }
 
 export abstract class ExtensionStore<T extends object> extends BaseStore<T> {
-  private static readonly instances = new WeakMap<object, ExtensionStore<object>>();
+  private static readonly instances = new WeakMap<object, any>();
 
   /**
    * @deprecated This is a form of global shared state. Just call `new Store(...)`
    */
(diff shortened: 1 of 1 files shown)


src/extensions/extension-store.ts
```diff
@@ -27,12 +27,12 @@ export interface ExtensionStoreParams<T extends object> extends BaseStoreParams<
...
 
 export abstract class ExtensionStore<T extends object> extends BaseStore<T> {
-  private static readonly instances = new WeakMap<object, ExtensionStore<object>>();
+  private static readonly instances = new WeakMap<object, any>();
 
   /**
    * @deprecated This is a form of global shared state. Just call `new Store(...)`
    */
-  static createInstance<T extends ExtensionStore<object>, R extends any[]>(this: StaticThis<T, R>, ...args: R): T {
+  static createInstance<T, R extends any[]>(this: StaticThis<T, R>, ...args: R): T {
```
jim-docker, line 35: Is this just reverting changes made in #6690, or are they typing changes needed because of refactoring there?

jim-docker (approved):

src/extensions/extension-store.ts
Nokel81: So in that PR I removed the `extends Singleton` from `Basestore` so to keep the same API of `ExtensionStore` I added the static methods here (because JS doesn't have multiple inheritance).

The ones on `Singleton` have `T extends Singleton` which is why I initially added the `T extends ExtensionStore<object>`, just it wasn't actually needed.

PR #6771, opened 2022-12-16
Split out use of httpsProxy preference

Nokel81:


src/common/fetch/proxy-fetch.injectable.ts
```diff
@@ -0,0 +1,30 @@
...
+import fetchInjectable from "./fetch.injectable";
+
+const proxyFetchInjectable = getInjectable({
+  id: "proxy-fetch",
+  instantiate: (di): Fetch => {
+    const fetch = di.inject(fetchInjectable);
+    const { httpsProxy, allowUntrustedCAs } = di.inject(userStoreInjectable);
+    const agent = httpsProxy
+      ? new HttpsProxyAgent({
+        proxy: httpsProxy,
```
jakolehm, line 18: If user has not specified any proxy value we probably should check what `resolveSystemProxyInjectable` gives us (per url).

jakolehm (approved):

PR #6787, opened 2022-12-19
Massively simplify bundled extension loading

Nokel81:

What this PR does:
1. Switch to using the standard `exports` field for the bundled extension entry points.
    - NOTE: this does mean that each extension is required to list all of `./package.json`, `./main`, and `./renderer` entry points. But since the default export can be `null` it is easier to have that requirement.
3. Makes the bundled extensions use an `injectionToken` so that it is fully OCP


.github/workflows/check-docs.yml
```diff
@@ -23,9 +23,9 @@ jobs:
 
       - name: Generate Extensions API Reference using typedocs
         run: |
-          yarn install
-          yarn typedocs-extensions-api
+          npm install
```
jakolehm: Just wondering what is the reason to change yarn to npm in this PR? What is the advantage?

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -21,19 +21,28 @@ import type { Extension } from "./extension/extension.injectable";
...
   readonly logger: Logger;
+  readonly extensionEntryPointName: "main" | "renderer";
   updateExtensionsState: (extensionsState: Record<LensExtensionId, LensExtensionState>) => void;
   createExtensionInstance: CreateExtensionInstance;
   getExtension: (instance: LensExtension) => Extension;
   joinPaths: JoinPaths;
   getDirnameOfPath: GetDirnameOfPath;
 }
 
+interface SemiLoadedExtension {
```
jakolehm: Name feels a bit confusing to me. Could you describe what is a "semi loaded extension"?

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -248,14 +257,84 @@ export class ExtensionLoader {
...
+      .map(extension => {
+        try {
+          const LensExtensionClass = extension[this.dependencies.extensionEntryPointName]();
+
+          if (!LensExtensionClass) {
+            return null;
+          }
+
+          const installedExtension: InstalledExtension = {
+            absolutePath: "irrelavent",
```
jakolehm:
```suggestion
            absolutePath: "irrelevant",
```

src/extensions/extension-loader/extension-loader.ts
```diff
@@ -248,14 +257,84 @@ export class ExtensionLoader {
...
+            id: extension.manifest.name,
+            isBundled: true,
+            isCompatible: true,
+            isEnabled: true,
+            manifest: extension.manifest,
+            manifestPath: "irrelavent",
```
jakolehm:
```suggestion
            manifestPath: "irrelevant",
```

src/extensions/extension-loader/extension-loader.ts
Nokel81: Tried to improve the name.

src/main/extension-loader/entry-point-name.injectable.ts
```diff
@@ -0,0 +1,14 @@
...
+import { getInjectable } from "@ogre-tools/injectable";
+import { extensionEntryPointNameInjectionToken } from "../../extensions/extension-loader/entry-point-name";
+
+const extensionEntryPointNameInjectable = getInjectable({
+  id: "extension-entry-point-name",
+  instantiate: () => "main" as const,
+  injectionToken: extensionEntryPointNameInjectionToken,
+});
+
+export default extensionEntryPointNameInjectable;
```
jakolehm, line 14: filename does not match to exported const name, maybe it should for consistency?

src/renderer/extension-loader/entry-point-name.injectable.ts
```diff
@@ -0,0 +1,14 @@
...
+import { getInjectable } from "@ogre-tools/injectable";
+import { extensionEntryPointNameInjectionToken } from "../../extensions/extension-loader/entry-point-name";
+
+const extensionEntryPointNameInjectable = getInjectable({
+  id: "extension-entry-point-name",
+  instantiate: () => "renderer" as const,
+  injectionToken: extensionEntryPointNameInjectionToken,
+});
+
+export default extensionEntryPointNameInjectable;
```
jakolehm, line 14: filename does not match to exported const name, maybe it should for consistency?

src/extensions/extension-loader/entry-point-name.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import { getInjectionToken } from "@ogre-tools/injectable";
+
+export const extensionEntryPointNameInjectionToken = getInjectionToken<"main" | "renderer">({
```
jakolehm, line 8: filename does not match to exported const name, maybe it should for consistency?

src/extensions/extension-discovery/bundled-extension-token.ts
```diff
@@ -0,0 +1,17 @@
...
+import { getInjectionToken } from "@ogre-tools/injectable";
+import type { LensExtensionConstructor, LensExtensionManifest } from "../lens-extension";
+
+export interface BundledExtension {
+  readonly manifest: LensExtensionManifest;
+  main: () => LensExtensionConstructor | null;
+  renderer: () => LensExtensionConstructor | null;
+}
+
+export const bundledExtensionInjectionToken = getInjectionToken<BundledExtension>({
```
jakolehm, line 15: filename does not match to exported const name, maybe it should for consistency?

src/renderer/extension-loader/entry-point-name.injectable.ts
Nokel81: Sure it can. I sometimes I include some or all of the folder name as part of what I am reading so that the file names don't get extremely long.

Iku-turso (approved):

PR #6757, opened 2022-12-15
Fix: node details not showing all pods

aleksfront:
Node details: ignore global namespace selector and always load pods from all namespaces.

Also, added tests for `<PodDetailsList/>` and moved styles to css modules.

https://user-images.githubusercontent.com/9607060/207845745-957778d8-30ae-46ff-92d8-1f6feb04090e.mov

Fixes #6739

PR #6794, opened 2022-12-20
Fix: load pods from all namespaces in node details

### LINKED ISSUE #6739: Node pod listing filtered by namespace selector
When Lens lists pods on the node in the node describe pane its filtered by the namespace selector.

a) Ignore namespace selector
b) Add namespace selector in this list

aleksfront:
Fixes #6739 

Instead of #6757 (Pull request 6757 contained mixed unrelated things).

https://user-images.githubusercontent.com/9607060/208654277-4b657b66-9086-4d4d-9d9d-48a218326c0c.mov

### DIFF
--- a/src/renderer/components/+workloads-pods/load-pods-from-all-namespaces.injectable.ts
+++ b/src/renderer/components/+workloads-pods/load-pods-from-all-namespaces.injectable.ts
@@ -21,8 +21,8 @@ const loadPodsFromAllNamespacesInjectable = getInjectable({
         onLoadFailure: error =>
           showErrorNotification(`Can not load Pods. ${String(error)}`),
       });
-    }
-  }
-})
+    };
+  },
+});
 
 export default loadPodsFromAllNamespacesInjectable;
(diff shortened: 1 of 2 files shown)


src/renderer/components/+workloads-pods/load-pods-from-all-namespaces.injectable.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import { getInjectable } from "@ogre-tools/injectable";
+import namespaceStoreInjectable from "../+namespaces/store.injectable";
+import showErrorNotificationInjectable from "../notifications/show-error-notification.injectable";
+import podStoreInjectable from "./store.injectable";
+
+const loadPodsFromAllNamespacesInjectable = getInjectable({
```
ixrock, line 11: Why this needs to be injectable? How previously it was worked?

ixrock (approved):
LGTM

src/renderer/components/+workloads-pods/load-pods-from-all-namespaces.injectable.ts
aleksfront: Previously, this function would live inside React component or in separate file with `namespaceStore` and `podStore` imported at the top of the file.

PR #6795, opened 2022-12-20
Renderer file logging transport

samitiilikainen:
Add file logging to renderer, writing separate log files for renderer main frame and each cluster frame.

Closes file handles (otherwise left open by Winston) on main and cluster frame refresh and when cluster frame is closed through cluster disconnect.

Related to lensapp/support-lens-extension#118

### DIFF
--- a/packages/core/src/renderer/frames/cluster-frame/init-cluster-frame/init-cluster-frame.ts
+++ b/packages/core/src/renderer/frames/cluster-frame/init-cluster-frame/init-cluster-frame.ts
@@ -2,6 +2,7 @@
  * Copyright (c) OpenLens Authors. All rights reserved.
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
+import { once } from "lodash";
 import type { Cluster } from "../../../../common/cluster/cluster";
 import type { CatalogEntityRegistry } from "../../../api/catalog/entity/registry";
 import type { ShowNotification } from "../../../components/notifications";
@@ -18,6 +19,7 @@ interface Dependencies {
   emitAppEvent: EmitAppEvent;
   logger: Logger;
   showErrorNotification: ShowNotification;
+  closeFileLogging: () => void;
 }
 
 const logPrefix = "[CLUSTER-FRAME]:";
@@ -30,6 +32,7 @@ export const initClusterFrame = ({
   emitAppEvent,
   logger,
   showErrorNotification,
+  closeFileLogging,
 }: Dependencies) =>
   async (unmountRoot: () => void) => {
     // TODO: Make catalogEntityRegistry already initialized when passed as dependency
@@ -73,11 +76,14 @@ export const initClusterFrame = ({
       });
     });
 
-    window.onpagehide = () => {
+    const onCloseFrame = once(() => {
       logger.info(
         `${logPrefix} Unload dashboard, clusterId=${(hostedCluster.id)}, frameId=${frameRoutingId}`,
       );
-
+      closeFileLogging();
       unmountRoot();
-    };
+    });
+
+    window.addEventListener("beforeunload", onCloseFrame, { capture: true });
+    window.addEventListener("pagehide", onCloseFrame, { capture: true });
   };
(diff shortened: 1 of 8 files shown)


Nokel81 (dismissed):

Nokel81:
@samitiilikainen Please sign your commit

packages/core/src/renderer/logger/file-transport.injectable.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import { transports } from "winston";
+import directoryForLogsInjectable from "../../common/app-paths/directory-for-logs.injectable";
+import { loggerTransportInjectionToken } from "../../common/logger/transports";
+import windowLocationInjectable from "../../common/k8s-api/window-location.injectable";
+import currentlyInClusterFrameInjectable from "../routes/currently-in-cluster-frame.injectable";
+import { getClusterIdFromHost } from "../utils";
+
+const rendererFileLoggerTransportInjectable = getInjectable({
```
jim-docker, line 13: usually the filename matches the injectable name/id ?

packages/core/src/renderer/frames/cluster-frame/init-cluster-frame/init-cluster-frame.ts
```diff
@@ -73,11 +76,14 @@ export const initClusterFrame = ({
...
         `${logPrefix} Unload dashboard, clusterId=${(hostedCluster.id)}, frameId=${frameRoutingId}`,
       );
-
+      closeFileLogging();
       unmountRoot();
-    };
+    });
+
+    window.addEventListener("beforeunload", onCloseFrame, { capture: true });
+    window.addEventListener("pagehide", onCloseFrame, { capture: true });
```
jim-docker: Does this mean logging to file stops if the cluster frame is not in view?  If I disconnect a cluster while on the catalog cluster list, any disconnect logging is missed?

packages/core/src/renderer/logger/file-transport.injectable.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    return new transports.File({
+      handleExceptions: false,
+      level: "info",
+      filename: `lens-renderer-${frameId}.log`,
+      dirname: di.inject(directoryForLogsInjectable),
+      maxsize: 1024 * 1024,
+      maxFiles: 0,
+      tailable: true,
```
jim-docker: Not sure what the expectation is here, but I tested with a file that was just under 1M in size and observed that when the file reached 1M nothing more was written to the file or any other file, and nothing more was written to the developer tools console either??

https://user-images.githubusercontent.com/40840436/215165042-11ae55cf-3529-4fb0-91a4-9baa9be15dea.mov

This is on Ubuntu 22.04. I artificially padded the file with logs to get it near 1M in size.

jim-docker (commented):
I noticed we have many `console.warn` etc calls still in the core code. They show up in the dev tools console but not in the log file of course. Out of scope for this PR, just noting it.

packages/core/src/renderer/logger/file-transport.injectable.ts
samitiilikainen: Thanks! Expectation was that it would keep writing logs and older lines would be lost but I was mistaken. I'll increase the maxFiles to allow it to properly rotate files.

packages/core/src/renderer/frames/cluster-frame/init-cluster-frame/init-cluster-frame.ts
samitiilikainen: Not all logging. It stops cluster frame logging right before the iframe is removed from DOM (when also all code execution in that iframe should stop). Stopping logging of the cluster frame does not affect logging done by the renderer main frame (catalog page).

jim-docker (dismissed):
Thanks for doing this!

(users with many many clusters may notice large disk usage by Lens due to all the new log files)

Nokel81 (approved):
Finally looks like this is green!

PR #6802, opened 2022-12-20
Make request-api-resources flatter in implementation

Nokel81:

Follow up to #6657


src/common/cluster/request-api-resources.injectable.ts
```diff
@@ -24,58 +25,49 @@ const requestApiResourcesInjectable = getInjectable({
...
-                namespaced: resource.namespaced,
-              });
-            }
-          })),
+        const resources = await Promise.all(
+          resourceListGroups
+            .map(withApiLimit(requestKubeApiResources)),
         );
+
+        return resources.flat();
```
Iku-turso: Could this be even flatter, consistent and more SRP with:

```typescript
const resources = await pipeline(resourceListGroups)
  .map(toCallForKubeApiResources)
  .map(withApiLimit)
  // Note: this could be .callAll() in iter.
  .map(call => call())
  // Not sure if these exist in iter yet?
  .awaitAll()
  .flat()
  .asArray();
```

src/common/cluster/request-api-resources.injectable.ts
```diff
@@ -24,58 +25,49 @@ const requestApiResourcesInjectable = getInjectable({
...
-          resourceListGroups.map(({ group, path }) => apiLimit(async () => {
-            const { resources } = await k8sRequest(cluster, path) as V1APIResourceList;
+      try {
+        const resourceListGroups: KubeResourceListGroup[] = [
+          ...(await requestApiVersions(cluster))
+            .map(version => ({
+              group: version,
+              path: `/api/${version}`,
+            })),
+          ...pipeline((await requestApisVersions(cluster)).values())
```
Iku-turso: Side-note: `pipeline`, as used here, could be more accurately called `chain`, as it implements collection chaining -pattern, and not the pipeline-pattern.

Iku-turso (commented):
Note: not reviewing this as normal PR to be merged when good enough, but instead as means to converse some coding patterns.

src/common/cluster/request-api-resources.injectable.ts
Nokel81: ah good point. That is probably a better name.

src/common/cluster/request-api-resources.injectable.ts
Nokel81: Given that `awaitAll` doesn't exist. I think that just splitting the maps would be good enough to make this flat.

Nokel81: I attempted that and the typings are not working out. Will push what I have so far since I think it is better. But I don't think that flatness is a quality indicator that must be pursued forever.

Iku-turso: Note: further pursuit will probably reveal some pretty cool stuff. Let's do that later, when time comes :)

Iku-turso (approved):

PR #6830, opened 2022-12-26
Use Monaco editor in ConfigMap details

### LINKED ISSUE #6815: Quick-view panel editors should have a fixed font, and a better editor
I would like the editor used in the bottom panels to be used in the quick view side-panel.

**Why is this needed**:
The side panel is brilliant for quick editing files in the config-map, but the editor is very limited, requires "shift-return" to enter newlines, and is not fixed-width font, which makes editing technical files more difficult.

I have included a screen shot for clarity about which editor I am talking about.

Screenshot 2022-12-21 at 10 05 33 PM

You could try to improve the editor, but it seems like dropping in the editor that is used for resources would be better and easier.

- Kubernetes distribution: multiple
- Desktop OS: osx Ventura

aleksfront:
Fixes #6815 providing more useful editor with `Tab` key support for `ConfigMap` data entries.

https://user-images.githubusercontent.com/9607060/209536408-63706305-c9ae-45bb-bb75-5e54afbe12f6.mov


src/renderer/components/monaco-editor/monaco-editor.tsx
```diff
@@ -37,11 +38,13 @@ export interface MonacoEditorProps {
   onDidContentSizeChange?(evt: editor.IContentSizeChangedEvent): void;
   onModelChange?(model: editor.ITextModel, prev?: editor.ITextModel): void;
   innerRef?: React.ForwardedRef<MonacoEditorRef>;
+  setInitialHeight?: boolean;
 }
 
 interface Dependencies {
   userStore: UserStore;
   activeTheme: IComputedValue<LensTheme>;
+  getEditorHeightFromLinesCount: (linesNumber: number) => "small" | "medium" | "large";
```
ixrock: Probably better to extract to some `type MonacoEditorSize = "small" | "medium" | "large"`. Wdyt?

src/renderer/components/monaco-editor/monaco-editor.tsx
```diff
@@ -288,13 +291,19 @@ class NonInjectedMonacoEditor extends React.Component<MonacoEditorProps & Depend
   // avoid excessive validations during typing
   validateLazy = debounce(this.validate, 250);
 
+  get initialHeightClassName() {
+    return this.props.setInitialHeight ?
+      styles[this.props.getEditorHeightFromLinesCount(this.model.getLineCount())] :
+      null;
+  }
+
```
ixrock: Why this value needs to setup for css besides predefined css-names? (e.g. "large" from injectable).
Also probably better name required for the getter + add returned `type`.

src/renderer/components/monaco-editor/get-editor-height-from-lines-number.injectable.ts
```diff
@@ -0,0 +1,30 @@
...
+
+      if (linesNumber < 10) {
+        return "small";
+      }
+
+      if (linesNumber < 20) {
+        return "medium";
+      }
+
+      return "large";
```
ixrock: Why do we return predefined names instead real numbers? Otherwise these built-in constants locked with css-names (e.g. "small")? And what the benefit then from your injectable when it receives current editor height-number and should return predefined css-class names?

ixrock (dismissed):
Some questions.. Otherwise LGTM.

src/renderer/components/+config-maps/config-map-details.tsx
```diff
@@ -102,15 +102,18 @@ class NonInjectedConfigMapDetails extends React.Component<ConfigMapDetailsProps
...
-                    </div>
+                    <MonacoEditor
+                      id={`config-map-data-${name}`}
+                      style={{
+                        resize: "vertical",
+                        overflow: "hidden",
+                        border: "1px solid var(--borderFaintColor)",
+                        borderRadius: "4px",
+                      }}
+                      value={value || ""}
```
ixrock: _nit:_ probably better would be to setup default value while extracting fields, e.g. `data.map(([name, value = ""])`

src/renderer/components/monaco-editor/get-editor-height-from-lines-number.injectable.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import { getInjectable } from "@ogre-tools/injectable";
+
+export type MonacoEditorSize = (linesCount: number) => number;
+
+const getEditorHeightFromLinesCountInjectable = getInjectable({
```
ixrock, line 8: So why in general do we need this **function/feature** to be `*.injectable.ts`? 
Any useful examples of how it would be utilized later in other envs? Or what's purpose in other words?

ixrock (dismissed):
LGTM

src/renderer/components/monaco-editor/monaco-editor.tsx
```diff
@@ -37,11 +39,13 @@ export interface MonacoEditorProps {
   onDidContentSizeChange?(evt: editor.IContentSizeChangedEvent): void;
   onModelChange?(model: editor.ITextModel, prev?: editor.ITextModel): void;
   innerRef?: React.ForwardedRef<MonacoEditorRef>;
+  setInitialHeight?: boolean;
 }
 
 interface Dependencies {
   userStore: UserStore;
   activeTheme: IComputedValue<LensTheme>;
+  getEditorHeightFromLinesCount: MonacoEditorSize;
```
ixrock: _btw_: looks a bit misleading type name `MonacoEditorSize` for some **function/action**, it feels more like a `number` or set of constants / enum by it's current name. Wdyt?

src/renderer/components/monaco-editor/monaco-editor.tsx
```diff
@@ -288,14 +292,23 @@ class NonInjectedMonacoEditor extends React.Component<MonacoEditorProps & Depend
...
+    return this.props.getEditorHeightFromLinesCount(this.model.getLineCount());
+  }
+
   render() {
-    const { className, style } = this.props;
+    const { className, style, setInitialHeight } = this.props;
+
+    const css: React.CSSProperties = {
+      ...style,
+      height: setInitialHeight ? this.initialHeight : style?.height,
```
ixrock: Shouldn't own `style?.height` have more priority over some initial height? E.g.

```suggestion
      height: style?.height ?? this.initialHeight,
```

src/renderer/components/monaco-editor/monaco-editor.tsx
aleksfront: Set an inline type 
```
interface Dependencies {
  userStore: UserStore;
  activeTheme: IComputedValue<LensTheme>;
  getEditorHeightFromLinesCount: (linesCount: number) => number;
}
```

src/renderer/components/monaco-editor/get-editor-height-from-lines-number.injectable.ts
aleksfront: It is implemented as injectable to follow de-facto standard and being able to inject it as dependency into `MonacoEditor`:

```
export const MonacoEditor = withInjectables<Dependencies, MonacoEditorProps, MonacoEditorRef>(
  React.forwardRef<MonacoEditorRef, MonacoEditorProps & Dependencies>((props, ref) => <NonInjectedMonacoEditor innerRef={ref} {...props} />),
  {
    getProps: (di, props) => ({
      ...props,
      userStore: di.inject(userStoreInjectable),
      activeTheme: di.inject(activeThemeInjectable),
      getEditorHeightFromLinesCount: di.inject(getEditorHeightFromLinesCountInjectable),
    }),
  },
);
```

Without `injectable` format it would be used from the shared state (eg. imported module) which is considered as bad practice.

ixrock: Alright, but why you've decided to control only height from MonacoEditor and this is why it's moved to `injectable.ts`?
You still didn't answer my question: if you able to control height of the editor, when/where would be reused? 
Some examples? Maybe builds for `lens-ide` or `lens-some-build` with different initial heights for the editor? This would make sense?

ixrock (approved):

PR #6837, opened 2022-12-28
Fix kubeconfig sync not saving data

### LINKED ISSUE #6834: Kubeconfig sync not working as expected
Kubeconfig sync in Preferences does not update changes. Shows only 1 synced folder `/Users/user/kube` even if it was removed or any other files added.

catalog clusters
synced items

### To reproduce:
1. Note clusters in `Catalog`
2. Go to Preferences -> Kubernetes -> Kubeconfig Syncs
3. Remove or add any entries
4. Close Preferences
5. See same clusters
6. If you go back to `Kubecongif Sync` you'll see `/Users/user/kube` as the only 1 item

aleksfront:
Fixing regression from #6690 

Fixes #6834 
kubeconfig sync

### DIFF
--- a/src/features/preferences/renderer/preference-items/kubernetes/kubeconfig-sync/kubeconfig-sync.tsx
+++ b/src/features/preferences/renderer/preference-items/kubernetes/kubeconfig-sync/kubeconfig-sync.tsx
@@ -3,7 +3,7 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
 import { withInjectables } from "@ogre-tools/injectable-react";
-import { action, computed, makeObservable, observable, reaction } from "mobx";
+import { computed, makeObservable, observable, reaction } from "mobx";
 import { disposeOnUnmount, observer } from "mobx-react";
 import React from "react";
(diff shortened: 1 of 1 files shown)


src/features/preferences/renderer/preference-items/kubernetes/kubeconfig-sync/kubeconfig-sync.tsx
```diff
@@ -59,11 +59,7 @@ class NonInjectedKubeconfigSync extends React.Component<Dependencies> {
       reaction(
         () => Array.from(this.syncs.entries(), ([filePath, kind]) => tuple.from(filePath, kind)),
         syncs => {
-          action(() => {
-            for (const [path] of syncs) {
-              this.props.userStore.syncKubeconfigEntries.set(path, {});
```
jansav, line 64: Was this empty object by accident? Was that the root cause? 

FYI: In the long run, we will be moving out from "saving happening in reaction" to mitigate subset of reaction bugs for good. The movement starts when first "Feature" wants to introduce new preferences (so very soon..) That will also make `UserStore` redundant. More on this topic when we have something to show.

jansav (approved):

PR #6851, opened 2022-12-30
Lens proxy with TLS

jakolehm:
Splitted and simplified from #6693 which has grown too big (imho).


src/main/lens-proxy/lens-proxy.injectable.ts
```diff
@@ -27,6 +28,7 @@ const lensProxyInjectable = getInjectable({
     contentSecurityPolicy: di.inject(contentSecurityPolicyInjectable),
     emitAppEvent: di.inject(emitAppEventInjectable),
     logger: di.inject(loggerInjectable),
+    certificate: di.inject(lensProxyCertificateInjectable).get(),
```
ixrock, line 31: If `set()` always throws why not return `cert` directly without extra `.get()`?

src/main/start-main-application/lens-window/application-window/create-electron-window.injectable.ts
```diff
@@ -56,6 +57,8 @@ const createElectronWindowInjectable = getInjectable({
     const isLinux = di.inject(isLinuxInjectable);
     const applicationInformation = di.inject(applicationInformationToken);
     const pathExistsSync = di.inject(pathExistsSyncInjectable);
+    const lensProxyCertificate = di.inject(lensProxyCertificateInjectable).get();
```
ixrock, line 60: `.get()` seems like redundant.

src/main/start-main-application/runnables/setup-hostnames.injectable.ts
```diff
@@ -0,0 +1,32 @@
...
+          "MAP lens.app 127.0.0.1",
+          "MAP *.lens.app 127.0.0.1",
+        ].join());
+
+        return undefined;
+      },
+    };
+  },
+
+  injectionToken: beforeElectronIsReadyInjectionToken,
```
ixrock: Since this injectable modifying global `app.commandLine.*` things maybe this should be marked with `causesSideEffects: true`? cc @Iku-turso @jansav

src/renderer/certificate/lens-proxy-certificate.injectable.ts
```diff
@@ -0,0 +1,39 @@
...
+        return certState;
+      },
+      set: (certificate: SelfSignedCert) => {
+        certState = certificate;
+      },
+    };
+    
+    requestFromChannel(lensProxyCertificateChannel).then((value) => {
+      cert.set(value);
+    });
```
ixrock: Shouldn't this to be `await`ed before returning `cert`?

ixrock (dismissed):
Haven't tested within app, code LGTM.

src/main/lens-proxy/lens-proxy.injectable.ts
jakolehm: Because the renderer needs to fetch it async from the main (and the injection token aka interface is the same for both sides). I decided to _not_ use async/await here because it would have caused a big refactoring.

src/renderer/certificate/lens-proxy-certificate.injectable.ts
jakolehm: We cannot await because `lensProxyCertificateInjectionToken` interface does not return a promise. If it would be async then we would need to refactor a big chunk of logic in `main` process because it currently does not allow to await. IMHO we can do that in a separate PR if there is a bigger need.

Nokel81: Actually I think this is the wrong place for this entirely. This should be a runnable. Probably under the `beforeFrameStartsFirstInjectionToken` phase

src/common/k8s-api/create-json-api.injectable.ts
```diff
@@ -18,8 +20,21 @@ const createJsonApiInjectable = getInjectable({
       fetch: di.inject(fetchInjectable),
       logger: di.inject(loggerInjectable),
     };
+    const lensProxyCert = di.inject(lensProxyCertificateInjectionToken);
 
-    return (config, reqInit) => new JsonApi(deps, config, reqInit);
+    return (config, reqInit) => {
+      config.getRequestOptions = async () => {
```
Nokel81: We shouldn't be overriding something that might be set

src/main/kubeconfig-manager/kubeconfig-manager.ts
```diff
@@ -101,16 +103,19 @@ export class KubeconfigManager {
...
         {
           name: contextName,
           server: this.resolveProxyUrl,
+          skipTLSVerify: false,
+          caData: Buffer.from(certificate.cert).toString("base64"),
         },
       ],
       users: [
-        { name: "proxy" },
+        { name: "proxy", username: "lens", password: "fake" },
```
Nokel81, line 118: I guess my work will actually add a correct password here too

src/main/lens-proxy/lens-proxy-certificate.global-override-for-injectable.ts
```diff
@@ -0,0 +1,20 @@
...
+import { getGlobalOverride } from "../../common/test-utils/get-global-override";
+import lensProxyCertificateInjectable from "./lens-proxy-certificate.injectable";
+
+export default getGlobalOverride(lensProxyCertificateInjectable, () => {
+  return {
+    get: () => ({
+      public: "<public-data>",
+      private: "<private-data>",
+      cert: "<ca-data>",
+    }) as SelfSignedCert,
```
Nokel81: I think it would be better to not override these or maybe override the runnable instead

src/main/start-main-application/runnables/setup-hostnames.injectable.ts
Nokel81: It doesn't cause side effects because side effects are for global shared state. `electronAppInjectable` is overridden in tests so there is not shared state between tests.

src/main/start-main-application/runnables/setup-hostnames.injectable.ts
```diff
@@ -0,0 +1,32 @@
...
+const setupHostnamesInjectable = getInjectable({
+  id: "setup-hostnames",
+
+  instantiate: (di) => {
+    const app = di.inject(electronAppInjectable);
+
+    return {
+      id: "setup-hostnames",
+      run: () => {
+        app.commandLine?.appendSwitch("host-rules", [
```
Nokel81: Is this `?` here because our override is incomplete?

src/main/lens-proxy/lens-proxy-certificate.global-override-for-injectable.ts
jakolehm: Why it would be better?

src/main/kubeconfig-manager/kubeconfig-manager.ts
jakolehm: Yes, I just had to add placeholder values here so that kubectl is happy.

src/common/k8s-api/create-json-api.injectable.ts
jakolehm: Good catch, will fix.

src/main/lens-proxy/lens-proxy-certificate.global-override-for-injectable.ts
Nokel81: The general preference to override as little business logic as possible. That way if a test tries to call `set` multiple times it fails the test instead of succeeding silently.

src/main/start-main-application/runnables/setup-hostnames.injectable.ts
jakolehm: This PR does not override `setupHostnamesInjectable`, what would be a proper way to override it?

Nokel81: I would just change the global override for `electronApp` to also include `commandLine: []` or better yet add a new injectable `appendCommandLineSwitch` which then is overridden as a noop.

src/main/lens-proxy/lens-proxy-certificate.global-override-for-injectable.ts
jakolehm: You mean that generating a LensProxy certificate should happen in a runnable and then we override the runnable? I have hard time to see how it would be better (the override would look exactly the same but in a different place) 🤔

src/main/start-main-application/runnables/setup-hostnames.injectable.ts
jakolehm: Fixed.

src/common/k8s-api/create-json-api.injectable.ts
jakolehm: Fixed.

src/renderer/certificate/lens-proxy-certificate.injectable.ts
jakolehm: > Actually I think this is the wrong place for this entirely. This should be a runnable. Probably under the `beforeFrameStartsFirstInjectionToken` phase

I agree, will refactor.

jakolehm: Fixed.

src/main/lens-proxy/lens-proxy-certificate.global-override-for-injectable.ts
Nokel81: I would argue that it is better because it more closely represents what happens in the application. If the runnable never gets called then the certificate is not set and should throw and thus causing test failures. Furthermore, if we accidently run the runnable twice then it should also throw.

src/renderer/certificate/lens-proxy-certificate.injectable.ts
```diff
@@ -26,10 +22,6 @@ const lensProxyCertificateInjectable = getInjectable({
         certState = certificate;
```
Nokel81: Probably should also verify that set is not called multiple times

src/main/lens-proxy/lens-proxy-certificate.global-override-for-injectable.ts
jakolehm: Ok, moved to runnables.

src/renderer/certificate/lens-proxy-certificate.injectable.ts
jakolehm: Good point... and we can probably have just a single implementation under "common" now.

Nokel81: Yes

jakolehm: Done.

Nokel81 (approved):

PR #6867, opened 2023-01-03
Remove kludge from KubeApi

Nokel81:

### DIFF
--- a/src/common/k8s-api/api-manager/api-manager.ts
+++ b/src/common/k8s-api/api-manager/api-manager.ts
@@ -5,11 +5,12 @@
 
 import type { KubeObjectStore } from "../kube-object.store";
 
-import { action, observable, makeObservable } from "mobx";
-import { autoBind, isDefined, iter } from "../../utils";
+import type { IComputedValue } from "mobx";
+import { action, observable } from "mobx";
 import type { KubeApi } from "../kube-api";
-import type { KubeJsonApiDataFor, KubeObject, ObjectReference } from "../kube-object";
+import type { KubeObject, ObjectReference } from "../kube-object";
 import { parseKubeApi, createKubeApiURL } from "../kube-api-parse";
+import { chain } from "../../utils/iter";
 
 export type RegisterableStore<Store> = Store extends KubeObjectStore<any, any, any>
   ? Store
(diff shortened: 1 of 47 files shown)


src/common/utils/iter.ts
```diff
@@ -236,3 +238,13 @@ export function every<T>(src: Iterable<T>, fn: (val: T) => any): boolean {
 
   return true;
 }
+
+export function* concat<T>(src1: IterableIterator<T>, src2: IterableIterator<T>): IterableIterator<T> {
```
Iku-turso: Isn't it starting to get at least a bit scary to produce integral technical functions without accompanying technical unit tests? :)

src/common/k8s-api/api-manager/kube-object-store-token.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectionToken } from "@ogre-tools/injectable";
+import type { KubeObjectStore } from "../kube-object.store";
+
+export const kubeObjectStoreInjectionToken = getInjectionToken<KubeObjectStore<any, any, any>>({
```
Iku-turso, line 8: Nit: `<unknown, unknown, unknown>` is a tad more accurate and safe.

src/common/k8s-api/api-manager/api-manager.ts
```diff
@@ -21,87 +22,71 @@ export type KubeObjectStoreFrom<Api> = Api extends KubeApi<infer KubeObj, infer
...
-
-    if (!this.apis.has(api.apiBase)) {
-      this.stores.forEach((store) => {
-        if (store.api === api) {
-          this.stores.set(api.apiBase, store);
-        }
-      });
-
-      this.apis.set(api.apiBase, api);
+  registerApi<Api>(...args: [RegisterableApi<Api>] | [string, RegisterableApi<Api>]) {
```
Iku-turso, line 79: Nit: Different arguments could be handled cleaner with **method overloading**. Also: why is string-argument present at all if it's not used?

src/common/k8s-api/api-manager/kube-object-store-token.ts
Nokel81: Not possible due to how the type params work

src/common/k8s-api/api-manager/api-manager.ts
Nokel81: backwards compatability and this is method overloading, did you see the diff above.

Reminder: typescript doesn't have true method overloading, but instead it has argument type specifications where the final "concrete" version has to be at least as general as all the "overloads" combined.

src/common/utils/iter.ts
Nokel81: Sure I can do that.

src/common/k8s-api/api-manager/api-manager.ts
```diff
@@ -33,20 +33,38 @@ export class ApiManager {
...
+
+      for (const api of removedApis) {
+        for (const [apiBase, storedApi] of this.apis) {
+          if (storedApi === api) {
+            this.apis.delete(apiBase);
+          }
+        }
+      }
+    });
+  }
```
Iku-turso: Not unit testing code like this strikes terror in my heart.

Nokel81: This is though, all of the tests relating to drawers go through this code.

src/common/k8s-api/api-manager/api-manager.ts
```diff
@@ -33,20 +33,38 @@ export class ApiManager {
   private readonly externalApis = observable.array<KubeApi>();
   private readonly externalStores = observable.array<KubeObjectStore>();
 
-  constructor(private readonly dependencies: Dependencies) {}
+  private readonly apis = observable.map<string, KubeApi>();
+
+  constructor(private readonly dependencies: Dependencies) {
+    // NOTE: this is done to preserve the old behaviour of an API being discoverable using all previous apiBases
+    autorun(() => {
```
Iku-turso, line 40: Could this stateful `autorun` be a stateless `computed` instead?

Nokel81: No, because the fact that it is stateful is the point of being used. To maintain both backwards comparability and the expected behaviour we want to maintain a list of all the `apiBase`'s that were ever used even if they are currently different.

src/common/k8s-api/kube-api.ts
```diff
@@ -217,7 +217,7 @@ export class KubeApi<
   readonly kind: string;
   readonly apiVersion: string;
 
-  @observable apiBase: string;
+  @observable apiBase = "";
```
Iku-turso: This is the fix from earlier, right? Maybe a cleaner way to fix this, without the intermediate empty string, would be to defer `makeObservable` until the real initial string value is set.

Nokel81: Sure

Iku-turso (approved):

PR #6880, opened 2023-01-05
Set correct group in rbac

### LINKED ISSUE #6879: Allow Resource with group set to v1 in rbac.ts are not showing, if your are not cluster admin
**Describe the bug**
Allow Resource with group set to v1 in rbac.ts are not showing, if your are not cluster admin

https://github.com/lensapp/lens/pull/6657/files?show-deleted-files=true&show-viewed-files=true&file-filters%5B%5D=#r1062429006

**To Reproduce**
1. Create a user than can only view a single namespace.
2. Open lens with this context
3. No ConfigMap is showing

**Expected behavior**
I can see the configmap....

- Lens Version: 6.3
- Installation method (e.g. snap or AppImage in Linux): OpenLens

farodin91:
Fixes #6879

Signed-off-by: Jan Jansen <jan.jansen@gdata.de>

### DIFF
--- a/src/renderer/components/+workloads-overview/workloads/implementations/pods-workload.injectable.ts
+++ b/src/renderer/components/+workloads-overview/workloads/implementations/pods-workload.injectable.ts
@@ -21,7 +21,7 @@ const podsWorkloadInjectable = getInjectable({
     return {
       resource: {
         apiName: "pods",
-        group: "v1",
+        group: "",
       },
       open: navigate,
 
--- a/src/common/front-end-routing/routes/cluster/config/config-maps/config-maps-route.injectable.ts
+++ b/src/common/front-end-routing/routes/cluster/config/config-maps/config-maps-route.injectable.ts
@@ -13,7 +13,7 @@ const configMapsRouteInjectable = getInjectable({
     clusterFrame: true,
     isEnabled: di.inject(shouldShowResourceInjectionToken, {
       apiName: "configmaps",
-      group: "v1",
+      group: "",
     }),
   }),
   injectionToken: frontEndRouteInjectionToken,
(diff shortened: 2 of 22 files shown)


Nokel81 (approved):

Nokel81:
Thanks!

Nokel81 (changes requested):
Actually there is slightly more work to do.

Can you also change the relevant uses of `di.inject(shouldShowResourceInjectionToken, { ...`?

For instance there is one in `src/common/front-end-routing/routes/cluster/config/config-maps/config-maps-route.injectable.ts` that should be updated too.

farodin91:
@Nokel81  I hope caught all.

src/renderer/components/+workloads-overview/workloads/implementations/pods-workload.injectable.ts
```diff
@@ -21,7 +21,7 @@ const podsWorkloadInjectable = getInjectable({
     return {
       resource: {
         apiName: "pods",
-        group: "v1",
+        group: ",
```
Nokel81: missing a `"` I think

Nokel81:
Also please sign your commit. Thanks for the fast response

farodin91:
@Nokel81 done

Nokel81 (dismissed):

Nokel81:
Might have to fix the tests too

farodin91:
i will try to fix it tomorrow.

Nokel81:
I think it will be as simple as to just remove the `"v1"` in the tests too

Nokel81 (approved):

farodin91:
@Nokel81 i have no idea why the missing tests are failing.

Nokel81:
I can investigate

Nokel81:
I noticed that this doesn't entirely solve the issue. Will look again on Monday.

Nokel81:
@farodin91 if you run kubectl get --raw /api you should get back a JSON object that has a field called versions. What is its value? On GKE and minikube it is ["v1"]

farodin91:
I get the same value.

PR #6598, opened 2022-11-17
Allow computed tray menu for extensions

panuhorsmalahti:
* Make `trayMenus` optionally `IComputedValue` for extensions to allow extensions to dynamically change e.g. submenus
* Implementation background:
  * Makes `ExtensionRegistrator` optionally `IComputedValue` so that `trayMenuItemRegistratorInjectable` can return a `computed` that reacts to computed extension trayMenus. I originally changed the `trayMenuItemInjectionToken` to be optionally computed, but @Nokel81 suggested to change `ExtensionRegistrator` instead

PR #6904, opened 2023-01-11
Computed appMenus option

samitiilikainen:
* Make `appMenus` optionally `IComputedValue` similarly as `trayMenus` already are (#6598)
* Allows extensions to dynamically show/hide menu items

### DIFF
--- a/src/features/application-menu/main/menu-registration.ts
+++ b/src/features/application-menu/main/menu-registration.ts
@@ -3,7 +3,9 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
 import type { MenuItemConstructorOptions } from "electron";
+import type { IComputedValue } from "mobx";
 
-export interface MenuRegistration extends MenuItemConstructorOptions {
+export type MenuRegistration = {
   parentId: string;
-}
+  visible?: IComputedValue<boolean> | boolean;
+} & Omit<MenuItemConstructorOptions, "visible">;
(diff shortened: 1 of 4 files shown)


aleksfront (approved):

panuhorsmalahti (approved):

src/features/application-menu/main/menu-registration.ts
```diff
@@ -3,7 +3,9 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
 import type { MenuItemConstructorOptions } from "electron";
+import type { IComputedValue } from "mobx";
 
-export interface MenuRegistration extends MenuItemConstructorOptions {
+export type MenuRegistration = {
```
Nokel81, line 8: Why is this required?

Nokel81 (commented):
Sounds good. Though this sort of feature would be best used for when there is an unknown number of menu items instead of for hiding them.

src/features/application-menu/main/menu-registration.ts
samitiilikainen: MenuItemConstructorOptions contains property visibility (boolean) and here we would like to add the option for it to be `IComputedValue<boolean>`

Nokel81: Sounds good, missed that on my phone

Nokel81 (approved):

PR #6906, opened 2023-01-11
Bump Horizontal Pod Autoscaler apiBase version

aleksfront:
* Switched to `v2` version (introduced in [k8s v1.23](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#-strong-api-groups-strong-))
* Added fallbackApiBases: 
```
fallbackApiBases: [
  "/apis/autoscaling/v2beta2/horizontalpodautoscalers",
  "/apis/autoscaling/v2beta1/horizontalpodautoscalers",
  "/apis/autoscaling/v1/horizontalpodautoscalers",
],
```
* Update some types

Now HPAs get loaded using version `v2` for newer k8s. And new fields such as `behavior` and `metrics` presented:

https://user-images.githubusercontent.com/9607060/211752380-6902da32-6d0d-42b5-acaa-5673a344d255.mov

autoscaling:v2

HPAs get loaded using some of the fallback api versions if running older kubernetes:

https://user-images.githubusercontent.com/9607060/211752600-adc3cf58-32a0-425c-b245-3e328a14dafe.mov

Fixes #6168
Fixes #6817 
Fixes #6827

PR #6908, opened 2023-01-11
Add ability for KubeApi to filter server versions

Nokel81:

follow up to #6906

### DIFF
--- a/packages/core/src/common/k8s-api/__tests__/kube-api-version-detection.test.ts
+++ b/packages/core/src/common/k8s-api/__tests__/kube-api-version-detection.test.ts
@@ -736,7 +736,7 @@ describe("KubeApi", () => {
 
     it("requests version list from the api group from the initial apiBase", () => {
       expect(fetchMock.mock.lastCall).toMatchObject([
-        "http://127.0.0.1:9999/api-kube/apis/autoscaling",
+        "https://127.0.0.1:12345/api-kube/apis/autoscaling",
         {
           headers: {
             "content-type": "application/json",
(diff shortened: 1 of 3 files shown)


src/common/k8s-api/kube-api.ts
```diff
@@ -336,7 +357,7 @@ export class KubeApi<
   }
 
   protected async checkPreferredVersion() {
-    if (this.fallbackApiBases && !this.doCheckPreferredVersion) {
+    if (!this.doCheckPreferredVersion && (this.fallbackApiBases || this.allowedUsableVersions)) {
       throw new Error("checkPreferredVersion must be enabled if fallbackApiBases is set in KubeApi");
```
jim-docker:
```suggestion
      throw new Error("checkPreferredVersion must be enabled if fallbackApiBases is set or there are allowed usable versions in KubeApi");
```

jim-docker (commented):
will test further when https://github.com/lensapp/lens/pull/6906 is fixed

src/common/k8s-api/kube-api.ts
```diff
@@ -58,14 +58,32 @@ export interface DerivedKubeApiOptions {
   /**
    * If the API uses a different API endpoint (e.g. apiBase) depending on the cluster version,
    * fallback API bases can be listed individually.
+   *
    * The first (existing) API base is used in the requests, if apiBase is not found.
-   * This option only has effect if checkPreferredVersion is true.
+   *
+   * This option only has effect if {@link DerivedKubeApiOptions.checkPreferredVersion} is `true`.
    */
   fallbackApiBases?: string[];
```
aleksfront: 2 questions:
1. Should we use newly introduced `allowedUsableVersions` for Ingresses as well https://github.com/lensapp/lens/blob/master/src/common/k8s-api/endpoints/ingress-class.api.ts#L16
2. Should we completely remove `fallbackApiBases` entity then?

Nokel81: 1. Yes
2. No, because that is used for when the `apiBase` changes (not just the version). Take `IngressClassApi` as an example. When it was being tested it was under `/apis/extensions` but now it is under `/apis/networking.k8s.io`. That is what `fallbackApiBases` is for.

aleksfront:
Tested on the example of HPA.

Looks good for hpa v2 and newer k8s version. But got an app crash opening HPA `autoscaling/v2beta1` details panel (minikube with kubernetes version of `1.21.0`):


https://user-images.githubusercontent.com/9607060/214316554-271a4b78-f3a8-4269-abfe-eef6bf07b1c9.mov

Nokel81:
Thanks will take a look

Nokel81:
@aleksfront Seems like an issue stemming from your work in #6971 probably because you only have a v2 and a v1 metrics parser.

aleksfront:
> @aleksfront Seems like an issue stemming from your work in #6971 probably because you only have a v2 and a v1 metrics parser.

Fixing it right now. Will point new PR to this one.

aleksfront (approved):

PR #6911, opened 2023-01-11
Remove last usages of request in our code

Nokel81:


src/common/cluster/cluster.ts
```diff
@@ -656,7 +655,7 @@ export class Cluster implements ClusterModel {
       const namespaceList = [ctx?.namespace].filter(isDefined);
 
       if (namespaceList.length === 0 && error instanceof HttpError && error.statusCode === 403) {
-        const { response } = error as HttpError & { response: Response };
+        const { response } = error as HttpError & { response: { body: unknown }};
```
ixrock, line 658: Why `body` is unknown? Shouldn't it be `Response` from `node-fetch` or something (or better to re-export desired response type and re-use from `fetch.injectable`)?

src/main/get-metrics.injectable.ts
```diff
@@ -22,13 +25,16 @@ const getMetricsInjectable = getInjectable({
     ) => {
       const prometheusPrefix = cluster.preferences.prometheus?.prefix || "";
       const metricsPath = `/api/v1/namespaces/${prometheusPath}/proxy${prometheusPrefix}/api/v1/query_range`;
+      const body = new FormData();
+
+      for (const [key, value] of object.entries(queryParams)) {
```
ixrock, line 30: Why `object.entries` and not standard `Object.entries`? What the difference here?

ixrock (approved):

src/common/cluster/cluster.ts
Nokel81: It is still using the `request` package because that is what the `@kubernetes/client` package uses until https://github.com/kubernetes-client/javascript/issues/754 is resolved.

src/main/get-metrics.injectable.ts
Nokel81: Because the `Object.entries` is not typed as usefully. The only difference is their typing.

PR #6913, opened 2023-01-11
Better handling loading terminal fonts

### LINKED ISSUE #5019: RobotoMono font not loaded when terminal opened first time

ixrock:
Moving out from loading fonts only by CSS to JS with [FontFace.load()](https://developer.mozilla.org/en-US/docs/Web/API/FontFace/load) interface due rendering terminal on canvas and since it's more reliable/controllable way of loading font files in general).

https://user-images.githubusercontent.com/6377066/211885321-d993143a-c26b-4d6f-bc06-5c1ccc065795.mov

close #5019


src/renderer/components/dock/terminal/terminal.ts
```diff
@@ -53,7 +54,8 @@ export class Terminal {
     return this.elem.querySelector(".xterm-viewport")!;
   }
 
-  attachTo(parentElem: HTMLElement) {
+  async attachTo(parentElem: HTMLElement) {
+    await this.dependencies.preloadFont(this.fontFamily);
```
Nokel81: wouldn't it be better if we just preloaded the fonts once in a `Runnable`? That way we don't have to add these `preloadFont` calls all over the place.

src/renderer/components/dock/terminal/terminal-fonts.injectable.ts
```diff
@@ -0,0 +1,71 @@
...
+    const preloadFont = di.inject(preloadTerminalFontInjectable);
+
+    return async function (): Promise<any> {
+      return Promise.allSettled(
+        Array.from(terminalFonts.keys()).map(preloadFont),
+      );
+    };
+  },
+
+  causesSideEffects: true,
```
jansav: As @Nokel81 mentioned, to make this runnable, we can do it like: 

```typescript
export const preloadAllTerminalFontsInjectable = getInjectable({
  id: "preloadAllTerminalFontsInjectable",

  instantiate(di) {
    const terminalFonts = di.inject(terminalFontsInjectable);
    const preloadFont = di.inject(preloadTerminalFontInjectable);

    return {
      id: "preload-all-terminal-fonts",
      
      run: async function (): Promise<any> {
        return Promise.allSettled(
          Array.from(terminalFonts.keys()).map(preloadFont)
        );
      },
    };
  },

  injectionToken: beforeFrameStartsInjectionToken,

  causesSideEffects: true,
});
```

This means that before application in frame (renderer, e.g. root frame or cluster frame) starts, we preload all required fonts and then the component it self doesn't have to worry about preloading at all.

src/renderer/components/dock/terminal/terminal.ts
jansav: I agree. Should be easy. Example here: https://github.com/lensapp/lens/pull/6913/files#r1067800560

src/renderer/components/dock/terminal/terminal-fonts.injectable.ts
```diff
@@ -0,0 +1,78 @@
...
+
+      document.fonts.add(font);
+      await font.load();
+    };
+  },
+
+  causesSideEffects: true,
+});
+
+export const preloadAllTerminalFontsInjectable = getInjectable({
```
Nokel81, line 57: This should probably have a `*.global-override-for-injectable.ts` for tests

Nokel81: Another option would be to override the `preloadFontInjectable` so that the runnable isn't overridden.

ixrock: Tests fixed, PTAL

src/features/preferences/renderer/preference-items/terminal/terminal-font-family/terminal-font-family.tsx
```diff
@@ -13,26 +13,29 @@ import { Select } from "../../../../../../renderer/components/select";
...
-  ({ userStore, logger }: Dependencies) => {
-
-    // fonts must be declared in `fonts.scss` and at `template.html` (if early-preloading required)
-    const supportedCustomFonts: SelectOption<string>[] = [
-      "RobotoMono", "Anonymous Pro", "IBM Plex Mono", "JetBrains Mono", "Red Hat Mono",
-      "Source Code Pro", "Space Mono", "Ubuntu Mono",
-    ].map(customFont => {
+  ({ userStore, logger, terminalFonts }: Dependencies) => {
+    const bundledFonts: SelectOption<string>[] = Array.from(terminalFonts.keys()).map(font => {
       const { fontFamily, fontSize } = userStore.terminalConfig;
```
Nokel81, line 29: nit: this line could be moved to outside the `.map(() => {...})` which could simplify the code

src/renderer/components/dock/terminal/create-terminal.injectable.ts
```diff
@@ -9,7 +9,8 @@ import type { TabId } from "../dock/store";
 import type { TerminalApi } from "../../../api/terminal-api";
 import terminalSpawningPoolInjectable from "./terminal-spawning-pool.injectable";
 import terminalConfigInjectable from "../../../../common/user-store/terminal-config.injectable";
-import terminalCopyOnSelectInjectable from "../../../../common/user-store/terminal-copy-on-select.injectable";
+import terminalCopyOnSelectInjectable
```
Nokel81, line 12: this file probably doesn't need to be changed in this PR

src/renderer/components/dock/terminal/terminal-fonts.global-override-for-injectable.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getGlobalOverride } from "../../../../common/test-utils/get-global-override";
+import { preloadAllTerminalFontsInjectable } from "./terminal-fonts.injectable";
+
+export default getGlobalOverride(preloadAllTerminalFontsInjectable, () => {
+  return {
+    id: "",
```
Nokel81, line 10: nit: might be a good idea to keep the same `id`

src/renderer/components/dock/terminal/terminal-fonts.injectable.ts
```diff
@@ -0,0 +1,78 @@
...
+import RedHatMono from "../../../fonts/RedHatMono-Regular.ttf";
+import SourceCodePro from "../../../fonts/SourceCodePro-Regular.ttf";
+import SpaceMono from "../../../fonts/SpaceMono-Regular.ttf";
+import UbuntuMono from "../../../fonts/UbuntuMono-Regular.ttf";
+
+export const terminalFontsInjectable = getInjectable({
+  id: "terminalFontsInjectable",
+
+  instantiate() {
+    return new Map([
```
Nokel81, line 21: nit: either this map should be created from a call to `di.injectMany(terminalFontInjectionToken)` (for some new injection token) and then be fully OCP or it doesn't need to be wrapped in an injectable.

src/renderer/components/dock/terminal/terminal-fonts.injectable.ts
```diff
@@ -0,0 +1,78 @@
...
+        await Promise.allSettled(
+          Array.from(terminalFonts.keys()).map(preloadFont),
+        );
+      },
+    };
+  },
+
+  injectionToken: beforeFrameStartsFirstInjectionToken,
+
+  causesSideEffects: true,
```
Nokel81, line 77: actually on further note, given that `preloadTerminalFont` is `causesSideEffects: true` this one doesn't have to be since it doesn't introduce any new side effects.

Nokel81 (approved):

Nokel81:
Merging, the above comments can be resolved in a future PR

src/renderer/components/dock/terminal/terminal-fonts.global-override-for-injectable.ts
ixrock: I intentionally didn't put any id to avoid duplication and keeping in sync with the injectable in future (it feels like it should not be responsible for this, maybe we should not ask `id` at all and use it from passed injectable automatically) cc @jansav

Nokel81: This ID is different from the Injectable ID because it is part of the data which is created, same as any other field (could have been call `name`).

ixrock: Why it's different? If I duplicate the string it would be the same. Moreover, we can use different ID from returned object (from runnable):

```ts
export const preloadAllTerminalFontsInjectable = getInjectable({
  id: "preloadAllTerminalFontsInjectable",

  instantiate(di) {
    const terminalFonts = di.inject(terminalFontsInjectable);
    const preloadFont = di.inject(preloadTerminalFontInjectable);

    return {
      id: "preload-all-terminal-fonts", // <--- THAT ONE

      async run() {
        await Promise.allSettled(
          Array.from(terminalFonts.keys()).map(preloadFont),
        );
      },
    };
  },

  injectionToken: beforeFrameStartsFirstInjectionToken,

  causesSideEffects: true,
});
```

Nokel81: ` id: "preload-all-terminal-fonts", // <--- THAT ONE` This is the one that I am talking about

ixrock: Yes, I mean this ID should be utilized automatically from passed injectable to make `id` optional.

Nokel81: Fair enough that would be nicer indeed

PR #6926, opened 2023-01-12
Allow specifying ResourceName in KubeObjectListLayout

panuhorsmalahti:
* If an extensions uses `KubeObjectListLayout` for a custom resource type, the `ResourceName` would be missing and `kind` would be used. However, extensions might want to customize the resource name as it's visible in the search input placeholder text.

### DIFF
--- a/src/renderer/components/kube-object-list-layout/kube-object-list-layout.tsx
+++ b/src/renderer/components/kube-object-list-layout/kube-object-list-layout.tsx
@@ -40,6 +40,10 @@ export interface KubeObjectListLayoutProps<
   store: KubeObjectStore<K, A, D>;
   dependentStores?: SubscribableStore[];
   subscribeStores?: boolean;
+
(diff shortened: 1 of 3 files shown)


src/renderer/components/kube-object-list-layout/kube-object-list-layout.tsx
```diff
@@ -40,6 +40,10 @@ export interface KubeObjectListLayoutProps<
   store: KubeObjectStore<K, A, D>;
   dependentStores?: SubscribableStore[];
   subscribeStores?: boolean;
+
+  // Customize resource name for e.g. search input ("Search <ResourceName>..."")
```
Nokel81: nit: this sort of documentation won't be picked up. Please use `/**` and `*/` instead

panuhorsmalahti: Fixed

Nokel81 (approved):

PR #6930, opened 2023-01-12
Add faking LensProxy to ApplicationBuilder

Nokel81:


src/main/router/route-request.injectable.ts
```diff
@@ -0,0 +1,81 @@
...
+  instantiate: (di): RouteRequest => {
+    const router = createRouter(di);
+
+    return async (cluster, req, res) => {
+      const url = new URL(req.url, "https://localhost");
+      const path = url.pathname;
+      const method = req.method.toLowerCase();
+      const matchingRoute = router.route(method, path);
+
+      if (matchingRoute instanceof Error) {
```
Iku-turso, line 52: That's one peculiar way to use `Errors` :)

src/main/router/route-request.injectable.ts
```diff
@@ -0,0 +1,81 @@
...
+        return false;
```
Iku-turso: This scenario could self-document better. What does it mean for `route-request` to return `false`?

src/main/lens-proxy/lens-proxy.injectable.ts
```diff
@@ -5,7 +5,6 @@
 import { getInjectable } from "@ogre-tools/injectable";
```
Iku-turso, line 5: "Simplify router for LensProxy" -> scary to simplify without unit tests, 'innit? :) We could try adding the missing tests first, and then simplify from the safety of that. Who's with me? ⚔️

src/main/lens-proxy/lens-proxy.global-override-for-injectable.ts
```diff
@@ -0,0 +1,19 @@
...
+import lensProxyInjectable from "./lens-proxy.injectable";
+
+export default getGlobalOverride(lensProxyInjectable, (di) => {
+  const lensProxyPort = di.inject(lensProxyPortInjectable);
+
+  return ({
+    close: () => { },
+    listen: async () => {
+      lensProxyPort.set(12345);
+    },
```
Iku-turso: Looks like setup of a temporal dependency in a global override. Is this the case? Some ideas to improve the design:

1. We could instead override the side-effects of `lensProxyInjectable`, and let `lensProxyPort` be set more plausibly.
2. We could consider having `lensProxyPort` add `lensProxy` as dependency. This would mean elimination of the logical/temporal dependency, as then the port would always be true to whatever `lensProxy` says it should be, without state (ie. reactively).
3. Failing at previous stuff (or while waiting), maybe there should also be a `globalOverride` for `lensProxyPort`, instead of this fake implementation.

To know that thing such as `lensProxy` needs to be magically listened before accessing `lensProxyPort` is a big ask.

src/main/lens-proxy/handle-request-for.injectable.ts
```diff
@@ -0,0 +1,62 @@
...
+        delete req.headers.authorization;
+        req.url = req.url.replace(apiKubePrefix, "");
+
+        return contextHandler.getApiTarget(isLongRunningRequest(req.url));
+      }
+
+      return undefined;
+    };
+
+    return (proxy) => async (req, res) => {
```
Iku-turso: Would it be possible for the proxy to be a normal injected dependency instead of one injected using a configuration parameter of a "for" function (eg. `handle-request-for`).

src/renderer/components/test-utils/get-application-builder.tsx
```diff
@@ -213,6 +218,35 @@ export const getApplicationBuilder = () => {
...
+      await handleLensRequest.handle(req, res);
+
+      return new Response(duplex, {
+        headers: new Headers(res._getHeaders() as Record<string, string>),
+        status: res._getStatusCode(),
+        statusText: res._getStatusMessage(),
+      });
+    };
+  });
+
```
Iku-turso: Interesting fake! Thinking real hard about the ramifications of it (note to self).

src/common/test-utils/use-fake-time.ts
```diff
@@ -19,7 +19,13 @@ export const advanceFakeTime = (milliseconds: number) => {
...
   usingFakeTime = true;
 
-  jest.useFakeTimers();
+  jest.useFakeTimers({
+    doNotFake: [
+      "nextTick",
+      "setImmediate",
+      "clearImmediate",
+    ],
+  });
```
Iku-turso: Hmm. What is the motivation of this? Messing with the default fake time can come as a surprise very easily.

Iku-turso (changes requested):
In general, many pieces of new code here suffer from lack of `extract function` -refactoring, making code and its motivation a tad hard to decipher. It's more than tad.

Iku-turso:
Ah! Still thinking of that one note to self :/

src/main/router/route-request.injectable.ts
Nokel81: Oh yes. That is just how the library does exception-less error handling. Sort of like our `AsyncResult` type

src/main/router/route-request.injectable.ts
Nokel81: Good point

src/main/lens-proxy/lens-proxy.injectable.ts
Nokel81: By the end of this PR there are tests that verify this.

src/main/lens-proxy/lens-proxy.global-override-for-injectable.ts
Nokel81: I think instead it shows more confidence in actual business code if we do not do that sort of overriding magic. It means that the `Runnable` phases are setup correctly and verified in unit tests (without the side effects or shared state).

src/main/lens-proxy/handle-request-for.injectable.ts
Nokel81: Will attempt to do so, but I think I did this to prevent a circular dependency.

src/common/test-utils/use-fake-time.ts
Nokel81: So lots of libraries it use those three methods to break causal chains. I think I should be able to revert this if I add the `advance: true` option instead. 

That option automatically advances the fake time 20ms for every real 20ms but does so with an `act()` wrapping.

WDYT?

Nokel81:
> extract function refactoring

What do you mean exactly? Not enough DRY?

src/main/lens-proxy/handle-request-for.injectable.ts
Nokel81: Unfortunately it would cause a circular dep instead I will just consolidate it.

Iku-turso: The circular dep does indicate a probable design error, and this kludges around it. I would advice caution :/

src/main/lens-proxy/lens-proxy.injectable.ts
Iku-turso: Yes, but if the unit tests exist when the production code changes for eg. simplicity, it would be **refactoring** instead of **reworking**, making the act more credible. Something to consider in the future. Resolved.

src/main/lens-proxy/handle-request-for.injectable.ts
```diff
@@ -0,0 +1,62 @@
...
+      if (cluster) {
+        const proxyTarget = await getProxyTarget(req, cluster.contextHandler);
+
+        if (proxyTarget) {
+          return proxy.web(req, res, proxyTarget);
+        }
+      }
+
+      res.setHeader("Content-Security-Policy", contentSecurityPolicy);
+      await routeRequest(cluster, req, res);
```
Iku-turso: This looks wrong, why does handling of a request mutate response from somewhere instead of creating it?

Iku-turso: Type of stuff like `req` and `res` are from a foreign library, and should be abstracted with an internal model. We really shouldn't couple/entangle with a specific technology such as node-fetch, as it would be better to be able to change our mind with that later.

It helps to consider foreign models as bad and poisonous as default.

src/main/lens-proxy/handle-lens-request.injectable.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  return req.headers.host + req.url;
+};
+
+export interface HandleLensRequest {
+  handle: HandleRequest;
+  stopHandling: () => void;
+}
+
+const handleLensRequestInjectable = getInjectable({
```
Iku-turso, line 150: Extract some functions/variables to make sense in all dat spaghetti? 🍝

src/renderer/components/test-utils/get-application-builder.tsx
```diff
@@ -213,6 +218,35 @@ export const getApplicationBuilder = () => {
     },
   }));
 
+  mainDi.override(lensFetchInjectable, (di) => {
```
Iku-turso: Is this "lensFetchFake" supposed to work like `channel-fakes`, ie. with multiple environments/dis collaborating together? Shouldn't something called `lensFetchInjectable` be present only in renderer? Or does main `lensFetch` stuff also? Why isn't the fake present in renderer?

Iku-turso: As a higher priority design problem, `lensFetchInjectable` works at too specific level of abstraction :/ It shouldn't know anything about `http`, `express`, `streams`, `node-fetch`, etc. Coupling/entangling it with these low-level policies means trouble with eg. "fakification" as can be seen with usage of a framework `httpMocks`. 

It is too difficult to at whim reimplement a contract which has been perverted with implementation-specific technical details and bleeding abstractions.

src/renderer/components/test-utils/get-application-builder.tsx
```diff
@@ -213,6 +218,35 @@ export const getApplicationBuilder = () => {
...
+    return async (pathnameAndQuery, init) => {
+      const handleLensRequest = di.inject(handleLensRequestInjectable);
+      const { Headers, Response } = di.inject(nodeFetchModuleInjectable);
```
Iku-turso: It's pretty weird to inject constructor-functions, even more so as here they are specific to foreign, non-contract technical details. 

It would make more sense to inject `factory-functions`, if not instances, when required.

src/main/lens-proxy/handle-request-for.injectable.ts
Nokel81: Yes so I removedthis injectable since it was only used in one place

src/renderer/components/test-utils/get-application-builder.tsx
Nokel81: Main `lensFetch`'s too and this fake is present in renderer. Continue reading.

src/renderer/components/test-utils/get-application-builder.tsx
Nokel81: The only reason `nodeFetchModuleInjectable` exists at all is because we cannot just import it normally.

src/common/test-utils/use-fake-time.ts
Nokel81: I came up with an even better plan. One that using the real clock to do `act(() => {})` calls.

PR #6937, opened 2023-01-13
Cleanup loading terminal fonts

Nokel81:
Follow up #6913 

- Make list of possible fonts fully OCP
- Only mark the loading of a font as 'causesSideEffects'
- Make a view model for the terminal font preferences component
- Move the TTF files next to where they are registered to help with finding them


src/features/terminal/renderer/fonts/preload-fonts.injectable.ts
```diff
@@ -0,0 +1,24 @@
...
+import terminalFontsInjectable from "./fonts.injectable";
+import loadTerminalFontInjectable from "./load-font.injectable";
+
+const preloadTerminalFontsInjectable = getInjectable({
+  id: "preload-terminal-fonts",
+  instantiate: (di) => ({
+    id: "preload-terminal-fonts",
+    run: async () => {
+      const terminalFonts = di.inject(terminalFontsInjectable);
+      const loadTerminalFont = di.inject(loadTerminalFontInjectable);
```
ixrock, line 16: Shouldn't these be outside of `run()`?

src/features/preferences/renderer/preference-items/terminal/terminal-font-family/terminal-font-family.tsx
```diff
@@ -5,66 +5,31 @@
...
+  </section>
+),
 );
 
 export const TerminalFontFamily = withInjectables<Dependencies>(NonInjectedTerminalFontFamily, {
   getProps: (di) => ({
-    userStore: di.inject(userStoreInjectable),
-    logger: di.inject(loggerInjectable),
-    terminalFonts: di.inject(terminalFontsInjectable),
+    model: di.inject(terminalFontPreferenceModelInjectable),
```
ixrock: Where this model would be useful/utilized besides terminal settings?
Currently nowhere and probably would not need anywhere.. So why to make it complex from the beginning and split to maximum amount of files? No benefits currently, imho.

ixrock (dismissed):

src/features/terminal/renderer/fonts/red-hat-mono.injectable.ts
```diff
@@ -0,0 +1,18 @@
...
+const redHatMonoTerminalFontInjectable = getInjectable({
+  id: "red-hat-mono-terminal-font",
+  instantiate: () => ({
+    name: "Red Hat Mono",
+    url: RedHatMono,
+  }),
+  injectionToken: terminalFontInjectionToken,
+});
+
+export default redHatMonoTerminalFontInjectable;
```
ixrock, line 18: _btw:_ why do we need to name some injectable when it's name never used in many files? we could do direct `default` export like this:

```ts
export default getInjectable({
  id: "red-hat-mono-terminal-font",
  instantiate: () => ({
    name: "Red Hat Mono",
    url: RedHatMono,
  }),
  injectionToken: terminalFontInjectionToken,
});
```

src/features/preferences/renderer/preference-items/terminal/terminal-font-family/terminal-font-options.injectable.tsx
```diff
@@ -0,0 +1,45 @@
...
+            {font.name}
+          </span>
+        ),
+        value: font.name,
+        isSelected: userStore.terminalConfig.fontFamily === font.name,
+      }))),
+      current: computed(() => userStore.terminalConfig.fontFamily),
+      set: action(selection => {
+        userStore.terminalConfig.fontFamily = selection?.value ?? defaultTerminalFontFamily;
+      }),
```
ixrock: _nit:_ better readability + `set()` should not know anything about `SelectOption`, better to pass just a font value:

```suggestion
      get fontFamily(){
        return userStore.terminalConfig.fontFamily;
      },
      setFont: action(fontFamily => {
        userStore.terminalConfig.fontFamily = fontFamily ?? defaultTerminalFontFamily;
      }),
```

src/features/preferences/renderer/preference-items/terminal/terminal-font-family/terminal-font-family.tsx
ixrock: For further reading: https://en.wikipedia.org/wiki/Occam's_razor

src/features/terminal/renderer/fonts/preload-fonts.injectable.ts
Nokel81: They could be, however I have found that for `Runnable`'s having the `di.inject` calls within the `run()` causes less headaches. It is not that important here, but when there are `runAfter`'s involved it is very important.

src/features/terminal/renderer/fonts/red-hat-mono.injectable.ts
Nokel81: In this case we could just do a `export default` since this will only ever be `di.injectMany`-ed. The reason to add the `const ...; export default ...;` is to help with intellisense and auto-import when using an injectable with `di.inject()`. This way it can easily be auto-imported if someone ever wants to import just this one injectable.

src/features/preferences/renderer/preference-items/terminal/terminal-font-family/terminal-font-family.tsx
Nokel81: This is an implementation of the https://en.wikipedia.org/wiki/Model-view-presenter pattern. Like similar patterns it helps make the view very passive and simple.

src/features/preferences/renderer/preference-items/terminal/terminal-font-family/terminal-font-options.injectable.tsx
Nokel81: I can update the name of the functions sure but using `computed(() => ...)` better communicates the contract then a getter.

src/features/preferences/renderer/preference-items/terminal/terminal-font-family/terminal-font-family.tsx
ixrock: Yes, it's kinda improvement I agree. 
But maybe done too early before any real some needs/system optimizations requirements, meanwhile making a whole system a bit more complex for nothing.

src/features/preferences/renderer/preference-items/terminal/terminal-font-family/terminal-font-options.injectable.tsx
ixrock: Better for who? Why end-user of this let's say api should care about `computed` that on that side or not?

ixrock: In other words to distinguish `observable` this field or not?

Nokel81: > Why end-user of this let's say api should care about computed that on that side or not?

So that the API user knows that they should only read the field in an observable context or that reading it multiple times outside of a context may result in different values.

aleksfront (approved):

PR #6939, opened 2023-01-13
Allow CatalogEntityDetails to be opened anywhere

### LINKED ISSUE #4963: Allow extensions to toggle the entity details panel
A way for an extension to toggle or view the entity details panel.

**Why is this needed**:
Some extensions would like that to be able to do so.

Nokel81:

Also expose this functionality to the extension API.

resolves #4963


packages/core/src/features/catalog/entity-running.test.tsx
```diff
@@ -0,0 +1,232 @@
...
+      });
+
+      it("onBeforeRun prevents event => onRun wont be triggered", async () => {
+        const onBeforeRunMock = jest.fn((event) => event.preventDefault());
+
+        catalogEntityRegistry.addOnBeforeRun(onBeforeRunMock);
+
+        rendered.getByTestId("detail-panel-hot-bar-icon").click();
+
+        expect(onRun).not.toHaveBeenCalled();
```
jim-docker, line 171: `await onBeforeRunMock.resolve();` before this?

packages/core/src/renderer/api/catalog/entity/registry.ts
```diff
@@ -243,20 +243,22 @@ export class CatalogEntityRegistry {
...
-          this.dependencies.logger.debug(`onBeforeRun for ${entity.getId()} returned false`);
-        }
-      })
-      .catch(error => this.dependencies.logger.error(`[CATALOG-ENTITY-REGISTRY]: entity ${entity.getId()} onRun threw an error`, error));
+  async onRun(entity: CatalogEntity) {
+    try {
+      const doOnRun = await this.onBeforeRun(entity);
+
+      if (!doOnRun) {
+        this.dependencies.logger.debug(`onBeforeRun for ${entity.getId()} returned false`);
```
jim-docker, line 251: Don't you need to return here?

```suggestion
        this.dependencies.logger.debug(`onBeforeRun for ${entity.getId()} returned false`);
        return;
```

packages/core/src/renderer/components/+catalog/entity-details/selected-uid.injectable.ts
```diff
@@ -0,0 +1,19 @@
+/**
```
jim-docker, line 1: injectable creep? So many injectables

jim-docker (commented):
doing some testing

packages/core/src/features/catalog/opening-entity-details.test.tsx
```diff
@@ -129,6 +133,7 @@ describe("opening catalog entity details panel", () => {
     describe("when opening the menu 'some-kubernetes-cluster'", () => {
       beforeEach(() => {
         rendered.getByTestId("icon-for-menu-actions-for-catalog-for-some-entity-id").click();
+        advanceFakeTime(1000);
```
jansav, line 136: Is the 1000ms something special here? The name of describe doesn't really provide any information and I would have to dig deeper in the code. What happens after 999ms?

jansav (approved):
Advancing of the fake time is not clear in the behavioural tests. I would like to see what happens after e.g. 999ms and 1000ms so that I can immediately see what's the time were waiting for. Also it would start bringing up more corner cases, e.g. what happens if I click something twice which does something only after 1000ms.

packages/core/src/features/catalog/opening-entity-details.test.tsx
Nokel81: This is to cover animations

PR #6947, opened 2023-01-16
Fix KubeJsonApi.forCluster missing from extension API and remove class constructors that require dependencies

Nokel81:
- Make sure that dependency constructors are not exported
- Readd `KubeJsonApi.forCluster`


src/extensions/common-api/k8s-api.ts
```diff
@@ -107,11 +111,24 @@ export {
...
 } from "../../common/k8s-api/kube-object";
 
-export {
-  KubeJsonApi,
-  type KubeJsonApiData,
+export type {
+  KubeJsonApiData,
 } from "../../common/k8s-api/kube-json-api";
 
+function KubeJsonApiCstr(config: JsonApiConfig, reqInit?: RequestInit) {
```
panuhorsmalahti, line 118: Does Cstr stand for Constructor? Maybe avoid abbreviation for clarity?

panuhorsmalahti (approved):

src/extensions/common-api/k8s-api.ts
Nokel81: Sure, that function isn't exposed to the extension API at all. Just an an internal implementation to this file.

PR #6951, opened 2023-01-16
Cleanup Cluster to remove environment specific details

Nokel81:

Please review this commit by commit.

By the end of this PR, the `Cluster` type will be no more than a shell which contains the `status` and `preferences` related details. These can be further re-factored in future PRs.

Notes:
- All connection related methods that used to be on `Cluster` are now on a new `ClusterConnection` type
- Many things were factored out into keyed singletons to improve readability.


src/main/kube-auth-proxy/kube-auth-proxy.ts
```diff
@@ -42,19 +44,17 @@ export class KubeAuthProxy {
 
   protected _port?: number;
   protected proxyProcess?: ChildProcess;
-  @observable protected ready = false;
+  protected readonly ready = observable.box(false);
```
ixrock: Removing observable fields/decorators everywhere?) _(in favour of computed values)_

src/main/cluster/manager.ts
```diff
@@ -239,7 +241,16 @@ export class ClusterManager {
...
+
+        void (async () => {
+          try {
+            await this.dependencies
+              .getClusterConnection(cluster)
+              .refreshConnectionStatus();
+          } catch {
+            // ignore
+          }
+        })();
```
ixrock: Weird construction. Could we make `async onNetworkOffline` and `await` at first level function's body?
And actually why `await` needed here at all when in `catch()` nothing anyway?

src/main/kube-auth-proxy/kube-auth-proxy.ts
Nokel81: Yes as an end goal because decorators have still not been finalised.

src/main/cluster/manager.ts
Nokel81: That is a better design, good point.

Nokel81:
This PR is blocked on https://github.com/lensapp/lens/pull/6930 and https://github.com/lensapp/lens/pull/4571

packages/core/src/common/cluster-store/cluster-store.ts
```diff
@@ -64,7 +62,7 @@ export class ClusterStore extends BaseStore<ClusterStoreModel> {
 
     const cluster = clusterOrModel instanceof Cluster
       ? clusterOrModel
-      : this.dependencies.createCluster(
+      : new Cluster(
```
jansav, line 65: It's not good thing to remove the `createCluster` since it abstracts the fact that Cluster is even a class. `new` keyword should never be used in the implementation code. There should always be a factory function to abstract the constructor arguments. This allows us to change the implementation how ever we want, but using `new Cluster` strictly ties us down to `Cluster` class with specific arguments.

packages/core/src/common/cluster/cluster.ts
```diff
@@ -3,164 +3,74 @@
...
+  readonly online = observable.box(false);
+
   /**
-   * Can user access cluster resources
-   *
-   * @observable
+   * Describes if user is able to access cluster resources
    */
-  @observable accessible = false; // if user is able to access cluster resources
+  readonly accessible = observable.box(false);
```
jansav, line 43: Isn't `Cluster` exposed in the Extension API? This would be breaking change.

This affects all properties that have changed here.

packages/core/src/common/cluster/cluster.ts
```diff
@@ -273,454 +148,83 @@ export class Cluster implements ClusterModel {
...
-        await this.reconnect();
-      } catch (error) {
-        this.broadcastConnectUpdate(`Failed to start connection: ${error}`, "error");
+    runInAction(() => {
+      this.kubeConfigPath.set(model.kubeConfigPath);
+      this.contextName.set(model.contextName);
 
-        return;
+      if (model.preferences) {
+        replaceObservableObject(this.preferences, model.preferences);
```
jansav, line 156: Why are we replacing the observable with another object? Why we are not just setting the value? This seems very weird because reference of an observable changes. This would mean that anyone using the old observable in computing doesn't recompute with the new one.

packages/core/src/common/utils/replace-observable-object.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import { runInAction } from "mobx";
+
+export function replaceObservableObject<Key extends string | number | symbol>(target: Partial<Record<Key, unknown>>, source: Partial<Record<Key, unknown>>): void {
```
jansav, line 8: I don't think should ever be used. The pattern that it introduces is troublesome. This will replace existing reference of an property with another. By looking at the use cases and the name of function it seems that it's meant to be used with observables, so it would mean that we replace reference of an observable with another. I don't see any good reason why we should do so, since we could just update the value. This opens up a whole new category of bugs where computing observable doesn't do it's job at all because the observable which it's dereferencing is replaced with another observable. I'd prefer not knowing if something like this works properly in MobX.

jansav (approved):
Needs to be said that this PR touches so many places without confidence of unit tests that it is likely to cause some bug somewhere. 

Seems like an improvement anyway.

PR #6957, opened 2023-01-17
Initial WIP flatpak build

tstoeckler:
This provides an initial work-in-progress flatpak build of Lens. It is still fairly rough, but assuming you have `flatpak-builder` and `flatpak-node-generator` (see https://github.com/flatpak/flatpak-builder-tools/tree/master/node) installed locally and have the `org.electronjs.Electron2.BaseApp` and `org.freedesktop.Sdk.Extension.node16` flatpaks installed (both in version 21.08) this should allow you to build and run a basic Flatpak version of Lens.

What I did:
- Followed https://docs.flatpak.org/en/latest/electron.html and modified the electron example flatpak manifest for lens
- Mucked around with the sandbox permissions until there were less warnings
- Found https://github.com/flatpak/flatpak-builder-tools/issues/173#issuecomment-770247677 (hat tip to @proletarius101) and used that to fix node-gyp being used in a post-install hook
- Modified the electron-builder manifest so that the tray icons are copied into the webpack build to avoid a startup error
- Instead of committing the `generated-sources.json` to the repo like in the example app, added a command to generate that on-the-fly prior to the actual flatpak build

A bunch of things don't work yet, but the hope is that this will get the ball rolling for a proper Lens flatpak and/or that this may already be useful to someone.

Would appreciate any kind of feedback!

EDIT: Add note about flatpak dependencies

### DIFF
--- a/flatpak/dev.k8slens.open-lens.yml
+++ b/flatpak/dev.k8slens.open-lens.yml
@@ -1,0 +1,54 @@
+app-id: dev.k8slens.open-lens
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
+sdk: org.freedesktop.Sdk
+base: org.electronjs.Electron2.BaseApp
+base-version: '21.08'
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node16
+command: run.sh
+separate-locales: false
+finish-args:
+  - --device=dri
+  - --filesystem=~/.k8slens
+  - --filesystem=~/.kube
+  - --share=ipc
+  - --share=network
+  - --socket=x11
+  - --system-talk-name=org.freedesktop.login1
+build-options:
+  append-path: /usr/lib/sdk/node16/bin
+  env:
+    NPM_CONFIG_LOGLEVEL: info
+modules:
+  - name: open-lens
+    buildsystem: simple
+    build-options:
+      env:
+        # DEBUG: electron-builder
+        XDG_CACHE_HOME: /run/build/open-lens/flatpak-node/cache
+        npm_config_cache: /run/build/open-lens/flatpak-node/npm-cache
+        npm_config_nodedir: /run/build/open-lens/flatpak-node/cache/node-gyp/19.0.4
+        npm_config_offline: 'true'
+    build-commands:
+      # Install dependencies
+      - yarn install --offline
+      # Build the app
+      - |
+        . ../flatpak-node/electron-builder-arch-args.sh
+        yarn run --offline build:linux $ELECTRON_BUILDER_ARCH_ARGS --linux --dir
+      # Bundle app and dependencies
+      - cp -a dist/linux*unpacked /app/main
+      # Install app wrapper
+      - install -Dm755 -t /app/bin/ ../run.sh
+    subdir: main
+    sources:
+      - type: dir
+        path: ..
+        dest: main
+      - generated-sources.json
+      # Wrapper to launch the app
+      - type: script
+        dest-filename: run.sh
+        commands:
+          - zypak-wrapper.sh /app/main/open-lens "$@"
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@ binaries/server/
 docs/extensions/api
 site/
 build/webpack/
+flatpak/.flatpak-builder/
+flatpak/builder/
+flatpak/generated-sources.json
--- a/.gitignore
+++ b/.gitignore
@@ -15,4 +15,3 @@ site/
 build/webpack/
 flatpak/.flatpak-builder/
 flatpak/builder/
-flatpak/generated-sources.json
--- a/flatpak/generated-sources.json
+++ b/flatpak/generated-sources.json
@@ -1,0 +1,13406 @@
+[
+    {
+        "type": "archive",
+        "url": "https://electronjs.org/headers/v19.0.4/node-v19.0.4-headers.tar.gz",
+        "strip-components": 1,
+        "sha256": "5d15daf89842ef3a4aa5cd8ee90334aca7c461d10a9d13c0a8ef164cf9737968",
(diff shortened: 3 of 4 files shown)


flatpak/dev.k8slens.open-lens.yml
```diff
@@ -0,0 +1,54 @@
+app-id: dev.k8slens.open-lens
+runtime: org.freedesktop.Platform
+runtime-version: '21.08'
```
harmathy, line 3: Why not use 22.08?

tstoeckler: Good question. For some reason I couldn't get the `node16` extension to install locally for version 22.08, even though it should exist. Not sure if there was something wrong on my end or what the issue there was/is, but, yeah, if that can be made to work, we should use 22.08 instead :+1:

harmathy: I'll have a look into this

proletarius101:
It would be ok not to commit generated-sources.json if Lens wants to host the flatpak repo themselves. Otherwise they should commit it to be included into flathub.org

tstoeckler:
Ahh I hadn't thought about that, but that makes sense. Yeah, I had introduced that build command for now under the assumption that they _don't_ want to maintain the `generated-sources.json` in the repo, but yeah I guess that would mean this should be moved to flathub instead. But I can't make that call, of course.

proletarius101:
@tstoeckler Nice catch. I should build from source to make it licensed under MIT. The official doesn't release the open source binary

harmathy:
How do you build the flatpak? If I run `yarn run flatpak:install`, I get

```
$ yarn run flatpak:install
yarn run v1.22.19
$ cd flatpak && flatpak-node-generator yarn ../yarn.lock && flatpak-builder builder dev.k8slens.open-lens.yml --install --force-clean --user
Reading packages from lockfiles...
1864 packages read.
Generating packages [1864/1864] esbuild @ 0.16.17                                                                                                                                                                                              
Generating headers electron @ 19.0.4
Wrote 1900 source(s).

(flatpak-builder:25783): flatpak-builder-WARNING **: 18:01:14.882: Unknown source type inline

(flatpak-builder:25783): Json-WARNING **: 18:01:14.884: Failed to deserialize "sources" property of type "gpointer" for an object of type "BuilderModule"
Emptying app dir 'builder'
Downloading sources
Initializing build dir
Committing stage init to cache
Starting build of dev.k8slens.open-lens
========================================================================
Building module open-lens in /src/lens/flatpak/.flatpak-builder/build/open-lens-1
========================================================================
Running: yarn install --offline
bwrap: Can't chdir to /run/build/open-lens/main: No such file or directory
Error: module open-lens: Child process exited with code 1
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

tstoeckler:
Hmm... that's strange. The first part seems like there is something wrong with the `generated-sources.json`. At least I didn't get that error. I went ahead and pushed the one I have locally to the repo now, so that it's easier to compare. If something does end up being merged here and it's without that file, we can always squash or whatever, so it shouldn't hurt for now, I think.

The `bwrap` error is strange, not really sure what that's about. For reference attaching the full build log that I get from `yarn run flatpak:install` (which does not generate the sources file as of now). Not sure if that helps...

[yarn-run-flatpak-install-log.txt](https://github.com/lensapp/lens/files/10439494/yarn-run-flatpak-install-log.txt)

harmathy:
The above error appears on a debian 11 machine. Possibly the flatpak-builder version is too old.

Now I tried also on a fedora 36 and Ubuntu 22.04 machine. But it won't build there either. Eventually I get this error:

```
Running: yarn install --offline
yarn install v1.22.19
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
error Can't make a request in offline mode ("https://registry.yarnpkg.com/underscore/-/underscore-1.13.6.tgz")
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Error: module open-lens: Child process exited with code 1
error Command failed with exit code 1.
```

tstoeckler:
Re @harmathy Re-did this from scratch in order to reproduce your error and I did in fact get the same error as you, so that's ... good I guess :shrug: :wink: 

Not sure what was different when I got it to work initially, but will try to figure it out. It's especially weird, because that exact underscore file is in the `generated-sources.json` and I can see that it was downloaded up front. But yeah, will keep you posted.

tstoeckler:
Ahh OK, that was an easy one: You actually need to run `yarn install` locally before running `yarn run flatpak:install`. It works as soon as I do that and I get the same error when I remove the `node_modules` directory again. Not sure why that's the case yet, but yeah, maybe that will you do at least try it out locally.

tstoeckler:
Now that I got this to work again, here's some more in detail instructions for this, if only for my own reference or in case someone else is following along:

1. Make sure you have the following binaries installed:
    - `flatpak`
    - `flatpak-builder`
    - [`flatpak-node-generator`](https://github.com/flatpak/flatpak-builder-tools/tree/master/node)
    - `g++`
    - `make`
    - `yarn`
    
2. Make sure you have the following flatpaks installed:

| Name | Version |
| --- | --- |
| `org.electronjs.Electron2.BaseApp` | 21.08 |
| `org.freedesktop.Platform` | 21.08 |
| `org.freedesktop.Sdk` | 21.08 |
| `org.freedesktop.Sdk.Extension.node16` | 21.08 |

3. Run the following commands:

```bash
yarn install
yarn flatpak:generate-sources # I guess this step is somewhat optional, as the sources are currently committed in the repo
yarn flatpak:install
```

4. Run the Lens flatpak via `flatpak run dev.k8slens.open-lens` (or `yarn run flatpak:run`)!

tstoeckler:
Closing this in favor of https://github.com/flathub/flathub/pull/3811 , thanks @proletarius101 !

PR #6974, opened 2023-01-18
Fix extension API version being specified by library user

Nokel81:


aleksfront (commented):
Can you add a bit more background behind this issue and how it's being solved?

We ignore extension version for some specific reason?

jakolehm (approved):

Nokel81:
> Can you add a bit more background behind this issue and how it's being solved?
> 
> We ignore extension version for some specific reason?

The bug was that after #6722 the extension API version was set as the library's consumer's version. Which is wrong.

PR #6984, opened 2023-01-19
Do not show Created on DetailsPane when there's no timestamp.

jweak:

It would show `<unknown> ago` if `creationTimestamp` was missing.


Nokel81 (dismissed):
Might need to update some snapshots but we will see

jweak:
> Might need to update some snapshots but we will see

Oh yeah looks like. it

jweak:
> Might need to update some snapshots but we will see

Updated snapshots.

Nokel81 (approved):

PR #7002, opened 2023-01-24
Remove `BaseStore<T>` and replace with composition of dependencies

Nokel81:


packages/core/src/common/entity-preferences/store.ts
```diff
@@ -20,30 +20,36 @@ export interface EntityPreferencesStoreModel {
...
   mergePreferences(entityId: string, preferences: PartialDeep<EntityPreferencesModel>): void {
-    this.preferences.set(entityId, merge(this.preferences.get(entityId), preferences));
-  }
-
-  @action
-  protected fromStore(data: EntityPreferencesStoreModel): void {
-    this.preferences.replace(data.entities ?? []);
+    runInAction(() => {
+      this.preferences.set(entityId, merge(this.preferences.get(entityId), preferences));
+    });
```
ixrock: Why do we need to "run in action" that single observable's update? or updating any map/set or single in general observable unit. Afaik `runInAction` needs to optimize update something/optimize reactions/etc? and do it in single "transaction" when updates multiple observables within single function scope, e.g.

```ts
// that would make sense for me for example
runInAction(() => {
      this.preferences.set(entityId, merge(this.preferences.get(entityId), preferences));
      this.somethingObservableToo = "UPDATED!";
    });
```

packages/core/src/extensions/extension-loader/file-system-provisioner-store/file-system-provisioner-store.ts
```diff
@@ -4,36 +4,49 @@
...
+import type { CreateBaseStore } from "../../../common/base-store/create-base-store.injectable";
 
-interface FSProvisionModel {
-  extensions: Record<string, string>; // extension names to paths
+interface FileSystemProvisionerStoreModel {
+  extensions: Partial<Record<string, string>>; // extension names to paths
 }
 
-interface Dependencies extends BaseStoreDependencies {
+interface Dependencies {
```
ixrock: Why not `extends BaseStoreDependencies` anymore? 
It seems we still pass it as dependencies to `this.store`? If so extending base interface would help with auto-completion when editing these fields.

packages/core/src/extensions/extensions-store/extensions-store.ts
```diff
@@ -23,23 +23,35 @@ export interface IsEnabledExtensionDescriptor {
...
     });
-    makeObservable(this);
-    this.load();
+    this.store.load();
   }
 
-  @computed
-  get enabledExtensions() {
-    return Array.from(this.state.values())
+  readonly enabledExtensions = computed(() => (
```
ixrock: _nit:_
```suggestion
  readonly enabledExtensions: IComputedValue<string[]> = computed(() => (
```

packages/core/src/main/cluster/manager.ts
```diff
@@ -246,7 +246,7 @@ export class ClusterManager {
 
   protected onNetworkOnline = () => {
     this.dependencies.logger.info(`${logPrefix} network is online`);
-    this.dependencies.store.clustersList.forEach((cluster) => {
+    this.dependencies.store.clustersList.get().forEach((cluster) => {
```
ixrock: _nit:_ why not? 🤪
anyway.. i don't think there is a big need to convert normal observable fields to computed values on the way _(too much)_.

```suggestion
    this.dependencies.get().store.get().clustersList.get().forEach((cluster) => {
```

ixrock (dismissed):
LGTM

packages/core/src/extensions/extension-loader/file-system-provisioner-store/file-system-provisioner-store.ts
Nokel81: Actually we specifically don't pass the whole of `this.dependencies` when we call `this.dependencies.createBaseStore()`.

The reason for this is the point of this PR. It is to make `BaseStore` compostable instead of inheritable. That way if `BaseStore` needs to add or remove a dependency we don't have to add a `di.inject()` call in all the places where subclasses are instantiated but instead only in the `createBaseStoreInejctable`.

packages/core/src/common/entity-preferences/store.ts
Nokel81: Yes `runInAction` is a way to specify a transaction. The reason why updating a single observable doesn't normally need to also be within an action is because we currently have `enforceActions: false` as part of our mobx configuration.

Iku-turso (approved):
Live-reviewed

PR #7009, opened 2023-01-25
Move node-fetch to separate package

jakolehm:
Moves custom webpack build of node-fetch to a separate package.


nx.json
```diff
@@ -20,6 +21,11 @@
       "dependsOn": [
         "prepare:dev"
       ]
+    },
+    "lint": {
+      "dependsOn": [
+        "^prepare:lint"
```
jakolehm: Added because lint depends on node-fetch types.

Nokel81 (commented):
Probably can remove the injection token too now from core.

Nokel81 (approved):

PR #7011, opened 2023-01-25
Fix getting metrics name for HPA v1

aleksfront:
Fixes https://github.com/lensapp/lens/pull/6908#issuecomment-1402012375

hpa v1 metrics in details


packages/core/src/renderer/components/+config-autoscalers/hpa-details.tsx
```diff
@@ -85,11 +85,13 @@ class NonInjectedHpaDetails extends React.Component<HpaDetailsProps & Dependenci
         }
         case HpaMetricType.External:
           return `${metricName} on ${JSON.stringify(metric.external.metricSelector ?? metric.external.metric?.selector)}`;
+        default:
+          return hpa.spec?.targetCPUUtilizationPercentage ? "CPU Utilization percentage" : "unknown";
```
Nokel81, line 89: Nit: I would check to make sure it is `typeof X === "number"` instead but I doubt that someone would set `targetCPUUtilizationPercentage` as `0`. So probably not important

Nokel81 (approved):

PR #7064, opened 2023-01-31
Use extension storeName to get store and data location

### LINKED ISSUE #7052: Allow extension to specify storeName

apialaviiva:
Resolves #7052 
* Add support for extension `storeName`
* Use storeName to get `extension-store` and `extension_data` locations


packages/core/src/extensions/__tests__/lens-extension.test.ts
```diff
@@ -33,4 +34,10 @@ describe("lens extension", () => {
       expect(ext.name).toBe("foo-bar");
     });
   });
+
+  describe("storeName", () => {
+    it("returns storeName", () => {
+      expect(ext.storeName).toBe("foo-bar-store");
```
jweak: Important tests would be for the `storeLocation` and `dataLocation`

packages/core/src/extensions/lens-extension.ts
```diff
@@ -93,7 +106,7 @@ export class LensExtension<
    * folder name.
    */
   async getExtensionFileFolder(): Promise<string> {
-    return this[lensExtensionDependencies].fileSystemProvisionerStore.requestDirectory(this.id);
+    return this[lensExtensionDependencies].fileSystemProvisionerStore.requestDirectory(this.dataLocation);
```
panuhorsmalahti: The id seems to be absolute path: https://github.com/lensapp/lens/blob/master/packages/core/src/extensions/extension-discovery/extension-discovery.ts#L354

Therefore, if we set storeName to "foo-bar" and we rename "foo-bar" to "@corp/foo-bar", the data location will change.

panuhorsmalahti: Maybe the fix is something like (pseudocode): `fileSystemProvisionerStore.requestDirectory(getExtensionId(this.storeName || this.name))`

apialaviiva: I'll check that out.

jweak: It'd be probably be safer to have the `extension.id` be calculated by `storeName` if it exists. That way you don't need to remember to use `this.storeName || this.name` everywhere you need `extension.id`

panuhorsmalahti: @jweak good point. The only 'drawback' is that it's slightly surprising that `storeName` affects extension id.. but perhaps fine.

apialaviiva: Updated after @jweak's comment, provided I understood it correctly.

jweak: I don't think it's right. So right now the LDK extension data is is `/Users/jheikka/Library/Application Support/Lens/extension_data/5d0651a6f92801496da1fe4fa20e1585ceeb4fa127282df2cf25d4f294e92a8d`. So we need to make sure that even after we update the name of the extension, directory name will still be `5d0651a6f92801496da1fe4fa20e1585ceeb4fa127282df2cf25d4f294e92a8d ` so that we don't lose data. So if it was calculated by `extension.name` we need to make sure it is now calculated by `storeName`

jweak: I think the place is here: 

https://github.com/lensapp/lens/blob/f445bd4eb982bb6683407e20b29858ece97a9f49/packages/core/src/extensions/extension-loader/file-system-provisioner-store/file-system-provisioner-store.ts#L39-L60

jweak: I'm not sure how with the random bytes it would have the same outcome. Anyway it seems this is actually called by `extension.id` and not `extensionName` like the parameter says

https://github.com/lensapp/lens/blob/f445bd4eb982bb6683407e20b29858ece97a9f49/packages/core/src/extensions/lens-extension.ts#L95-L98

jweak: So maybe this is working after all

panuhorsmalahti: I think the result is saved to `this.registeredExtensions` which is ultimately persisted to `lens-filesystem-provisioner-store.json` so the `randomBytes ` are not called again if it has already been calculated before.

packages/core/src/extensions/lens-extension.ts
```diff
@@ -62,7 +63,7 @@ export class LensExtension<
 
   constructor({ id, manifest, manifestPath, isBundled }: InstalledExtension) {
     makeObservable(this);
-    this.id = id;
+    this.id = manifest.storeName || id;
```
panuhorsmalahti, line 66: This doesn't seem right. Currently the id is absolute path to the package.json file.. which is probably arbitrary but it would seem inconsistent if some extensions suddenly had storeName as the id.

PR #7070, opened 2023-01-31
Convert Kubectl versioning map to compile time dynamic

### LINKED ISSUE #7473: Switch to dynamically fetching avaialble kubectl versions at build time
Motivation:
- Removes the need to update the list by hand
- Should be able to factor this out of core

Nokel81:
- Allows for bundled kubectl config to be changed without code changes


packages/core/src/main/kubectl/request-greatest-patch-version.injectable.ts
```diff
@@ -0,0 +1,52 @@
...
+const expectedResponseForm = TypedRegEx("^v(?<version>\\d+\\.\\d+\\.\\d+)$");
+
+const requestGreatestKubectlPatchVersionInjectable = getInjectable({
+  id: "request-greatest-kubectl-patch-version",
+  instantiate: (di): RequestGreatestKubectlPatchVersion => {
+    const fetch = di.inject(fetchInjectable);
+    const logger = di.inject(loggerInjectable);
+
+    return async (majorMinor) => {
+      const response = await fetch(`https://dl.k8s.io/release/stable-${majorMinor}.txt`);
```
ixrock: Maybe we should extract fetching URL to some `*.injectable.ts` or common function?
Btw, is it only `xml`-response for their apis available? No plain `json`?

ixrock (dismissed):

packages/core/src/main/kubectl/request-greatest-patch-version.injectable.ts
Nokel81: Yeah only XML (for error messages). Normal (200) response is just plain text

Nokel81:
@jakolehm PTALA

packages/technical-features/kubectl-versions/README.md
```diff
@@ -0,0 +1,18 @@
+# @k8slens/application
```
jakolehm: Readme needs to be updated.

packages/technical-features/kubectl-versions/build/versions.json
```diff
@@ -0,0 +1,110 @@
+[
```
jakolehm: This file does not belong to version control, right?

jakolehm (dismissed):

Nokel81:
@jakolehm It needs to be part of version control to help make tests pass. It will be unconditionally updated during a release.

packages/technical-features/kubectl-versions/index.ts
```diff
@@ -0,0 +1,3 @@
+import kubectlVersionsInjectable from "./src/versions.injectable";
+
+export { kubectlVersionsInjectable };
```
jakolehm: What is the benefit of exporting an injectable?

jansav: Downsides for exporting `injectable`:
- What would happen when someone decides to `register` or even `deregister` the injectable. This shouldn't be allowed
- Usage would depend from the implementation and not from the contract
- It would be harder to change our mind with technical details because they would have been leaked to the usage

Therefore, I would suggest to always `export` the `injectionToken` instead. This doesn't have same downsides and corner cases.

packages/technical-features/application/agnostic/src/start-application/start-application.injectable.ts
```diff
@@ -1,19 +1,19 @@
+/* eslint-disable prettier/prettier */
```
jansav, line 1: No, we either use the prettier or we don't. Let's not start disabling it file to file.

jansav: Also, this doesn't have anything to do with the contents of this PR. Revert.

packages/core/src/main/getDiForUnitTesting.ts
```diff
@@ -50,6 +49,8 @@ export function getDiForUnitTesting() {
     for (const block of chunk(100)(injectables)) {
       di.register(...block);
     }
+
+    di.register(kubectlVersionsInjectable);
```
jansav: This registration is the responsibility of the `kubectl-versions` feature. Introduce `feature.ts` in the package (check example in the `application` -feature) and use `registerFeature(kubectlFeature)`

packages/technical-features/kubectl-versions/src/versions.injectable.ts
```diff
@@ -0,0 +1,9 @@
+import { getInjectable } from "@ogre-tools/injectable";
+import versions from "../build/versions.json";
+
+const kubectlVersionsInjectable = getInjectable({
+  id: "kubectl-versions",
+  instantiate: (di) => versions as [string, string][],
+});
+
+export default kubectlVersionsInjectable;
```
jansav: Why is this even an injectable?

packages/core/src/main/getDiForUnitTesting.ts
Nokel81: This PR was created before that package existed. So sure

jansav (dismissed):
Check earlier comments

packages/core/src/main/getDiForUnitTesting.ts
jansav: Ok, cool

packages/technical-features/kubectl-versions/src/versions.injectable.ts
Nokel81: Maybe it shouldn't be? Okay

packages/technical-features/kubectl-versions/index.ts
```diff
@@ -1,3 +1,3 @@
-import kubectlVersionsInjectable from "./src/versions.injectable";
+import raw from "./build/versions.json";
 
-export { kubectlVersionsInjectable };
+export const kubectlVersions = raw as [string, string][];
```
Iku-turso: This package does not export a Feature, making it not belong in `technical-features`.

Nokel81: where should it live then? root level?

packages/technical-features/kubectl-versions/src/versions.injectable.ts
Iku-turso: One good reason for `injectable` is to control it for unit testing, even if it isn't a side-effect. Well it kinda is, for being hard-coded, volatile looking data.

packages/technical-features/kubectl-versions/build/compute-versions.ts
Iku-turso, line 1: No unit tests for this new script? :/

packages/technical-features/kubectl-versions/src/versions.injectable.ts
Iku-turso: ...but there's no related unit tests?

packages/technical-features/kubectl-versions/.swcrc
Iku-turso, line 1: Does this introduce another way to build a package? Consolidate?

packages/technical-features/kubectl-versions/README.md
```diff
@@ -0,0 +1,8 @@
+# @k8slens/kubectl-versions
+
+This package contains a injectable for a JSON object between strings of the format "MAJOR.MINOR" and "MAJOR.MINOR.PATCH".
```
Iku-turso: No longer accurate?

packages/core/src/main/kubectl/kubectl.ts
```diff
@@ -302,7 +302,7 @@ export class Kubectl {
   }
 
   protected async writeInitScripts() {
-    const binariesDir = this.dependencies.baseBundeledBinariesDirectory;
+    const binariesDir = this.dependencies.baseBundledBinariesDirectory;
```
Iku-turso, line 305: Fix typo in new commit in future :)

packages/technical-features/kubectl-versions/README.md
```diff
@@ -0,0 +1,8 @@
...
+
+This object represents the current (at time of building each release of this package)
+greatest PATCH version for each "MAJOR.MINOR" version of `kubectl`.
+This is done at compile time so that we don't have to worry about being able to access this data on user machines.
```
Iku-turso: Population of `versions.json` is triggered at compile time? Is there a threat that something not-committed breaks unit tests in CI, or even gets released? Can this make compilation slower? 

Maybe triggering this should be made manual instead of automatic. WDYT?

Nokel81: The whole point was to make it automatic

Nokel81: Maybe we could instead have some CI that triggers this on a daily basis and if there is a change, make a PR. WDYT?

Nokel81: Actually thinking about this more...

I don't think there is a threat that it breaks unit tests in CI. We can discuss making it periodic in the future.

gabriel-mirantis (approved):

PR #7077, opened 2023-02-01
Move resource templates to @k8slens/resource-templates package

jakolehm:
Allows to reuse templates on Lens Desktop Core based distros.


Nokel81 (dismissed):

packages/open-lens/package.json
```diff
@@ -103,7 +103,7 @@
     "afterSign": "build/notarize.js",
     "extraResources": [
       {
-        "from": "templates/",
+        "from": "node_modules/@k8slens/resource-templates/templates/",
```
Nokel81: Are you sure that this still works in dev mode?

Nokel81: Because the code in https://github.com/lensapp/lens/blob/master/packages/core/src/renderer/components/dock/create-resource/lens-templates.injectable.ts hasn't been changed

jakolehm: Now that you said it: it might be already broken (since the monorepo restructure) .. need to verify before merging.

Nokel81 (dismissed):

Nokel81 (approved):

PR #7080, opened 2023-02-01
Namespace details tree view

### LINKED ISSUE #7079: Create a tree view for hierarchical namespaces
Part of #2658

aleksfront:
Tree view supporting Hierarchical Namespaces.

To enable tree view please install Hierarchical Namespace Controller (HNC) https://github.com/kubernetes-sigs/hierarchical-namespaces/tree/1fdc099b5d46fdb691170a4b129ca010b9b90420 and do quick start steps described in https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/quickstart.md

https://user-images.githubusercontent.com/9607060/216025438-5bd7a7bf-6ebf-42e0-bb56-201c50d9c61a.mov

Resolves #7079


Nokel81:
Looks good. Some initial comments about the look and feel

1. The [X] box as part of the tree looks a but weird, I would use a more conventional list icon like a dash or dot
2. Would be great if the user could modify the hierarchy from within this view. Like adding or removing the links.

aleksfront:
> The [X] box as part of the tree looks a but weird, I would use a more conventional list icon like a dash or dot

Replaced `CloseIcon` with semi-transparent `MinusIcon`.

expanded tree view

> Would be great if the user could modify the hierarchy from within this view. Like adding or removing the links.

This would be a very good addition. Added to the list https://github.com/lensapp/lens/issues/7092.

packages/core/src/renderer/components/+namespaces/namespace-tree-view.tsx
```diff
@@ -0,0 +1,144 @@
...
+  group: {
+    marginLeft: 8,
+    paddingLeft: 16,
+    borderLeft: `1px dashed var(--borderColor)`,
+  },
+  label: {
+    fontSize: "inherit",
+    lineHeight: "1.8",
+    cursor: "default",
+    backgroundColor: "transparent!important",
```
ixrock: Can we avoid such inline styles in `*.tsx`? (and apply to components in separated `*.module.scss`

packages/core/src/renderer/components/+namespaces/subnamespace-badge.tsx
```diff
@@ -0,0 +1,36 @@
...
+import type { Namespace } from "../../../common/k8s-api/endpoints";
+import { Tooltip } from "../tooltip";
+
+interface SubnamespaceBadgeProps extends React.HTMLAttributes<HTMLSpanElement> {
+  namespace: Namespace;
+  id: string;
+}
+
+export function SubnamespaceBadge({ namespace, id, ...other }: SubnamespaceBadgeProps) {
+  if (!namespace.getAnnotations().find(annotation => annotation.includes("hnc.x-k8s.io/subnamespace-of"))) {
```
ixrock: Probably better to create `Namespace#isSubnamespace()` method and re-use where it's needed.

```tsx
isSubnamespace(){  
 return this.getAnnotations().find(annotation => annotation.includes("hnc.x-k8s.io/subnamespace-of"));
}
```

packages/core/src/renderer/components/+namespaces/namespace-tree-view.tsx
```diff
@@ -0,0 +1,144 @@
...
+        defaultEndIcon={(<div style={{ opacity: 0.3 }}><MinusSquare /></div>)}
+        expanded={expandedItems}
+      >
+        <StyledTreeItem
+          nodeId={`namespace-${root.getId()}`}
+          label={root.getName()}
+          data-testid={`namespace-${root.getId()}`}
+          onIconClick={(evt) => {
+            toggleNode(`namespace-${root.getId()}`);
+            evt.stopPropagation();
```
ixrock: Extract _(instead of multiple duplications)_:

```tsx
const nodeId = `namespace-${root.getId()}`;
```

packages/core/src/renderer/components/+namespaces/namespace-tree-view.tsx
```diff
@@ -0,0 +1,144 @@
...
+          }}
+        >
+          {renderChildren(root)}
+        </StyledTreeItem>
+      </TreeView>
+    </div>
+  );
+}
+
+function MinusSquare() {
```
ixrock: _nit:_ 

```suggestion
function MinusSquareIcon() {
```

_btw,_ good idea to add custom inline svg icons like this.

packages/core/src/renderer/components/+namespaces/namespace-tree-view.test.tsx
```diff
@@ -0,0 +1,264 @@
...
+    }, {
+      clusterServerUrl: "https://localhost:8080",
+    }));
+
+    namespaceStore = di.inject(namespaceStoreInjectable);
+
+    const subscribeStores = di.inject(subscribeStoresInjectable);
+
+    cleanup = disposer(subscribeStores([namespaceStore]));
+    render = renderFor(di);
```
ixrock: Looks like `<NamespaceTreeView />` needs lot of things to setup for the test. Why? It should not require or even know about some stores imho and just render some sort of data-tree -> dom-tree representations. Or I miss something? 🤔

packages/core/src/renderer/components/+namespaces/namespace-tree-view.test.tsx
```diff
@@ -0,0 +1,264 @@
...
+  "acme-org.tree.hnc.x-k8s.io/depth": "1",
+  "kubernetes.io/metadata.name": "team-a",
+  "team-a.tree.hnc.x-k8s.io/depth": "0",
+});
+
+const teamB = createNamespace("team-b", {
+  "hnc.x-k8s.io/included-namespace": "true",
+  "acme-org.tree.hnc.x-k8s.io/depth": "1",
+  "kubernetes.io/metadata.name": "team-b",
+  "team-b.tree.hnc.x-k8s.io/depth": "0",
```
ixrock, line 62: So, supporting `HNC` is purely based on `labels` and their `parent<->child` relationships?

packages/core/src/renderer/components/+namespaces/namespace-tree-view.tsx
aleksfront: Moved to scss.

packages/core/src/renderer/components/+namespaces/namespace-details.tsx
```diff
@@ -103,6 +104,8 @@ class NonInjectedNamespaceDetails extends React.Component<NamespaceDetailsProps
             </Link>
           ))}
         </DrawerItem>
+
+        <NamespaceTreeView root={namespace}/>
```
ixrock: It would be better if `<NamespaceTreeView namespaces={(string|object)[]} />` component could accept as input just plain array with extracted namespaces data from previous step (e.g. some extra function at api level maybe), that way you don't need to worry about apis, stores, etc. in your component tests. 
Also it's better follows single responsibility principles, UI component just displays some prepared data. Wdyt?

aleksfront: Cleaned up tests and `NamespaceTreeView` from any of the store dependencies. This allows to use much simpler test setup.

packages/core/src/renderer/components/+namespaces/namespace-tree-view.test.tsx
aleksfront: Yes, we can rely on labels to see parent/child relationships. For other use-cases like removing a subnamespace https://github.com/lensapp/lens/issues/7114 there will be some additional work with CRDs.

packages/core/src/renderer/components/+namespaces/subnamespace-badge.module.scss
```diff
@@ -1,10 +1,11 @@
 .subnamespaceBadge {
-  border-radius: 2px;
+  border-radius: 3px;
   padding: 0px 4px;
   border: 1px solid cadetblue;
   color: cadetblue;
-  display: inline-block;
-  font-size: 0.7rem;
+  display: inline-flex;
+  font-size: 0.6rem;
```
ixrock: Why `rem`?

packages/core/src/renderer/components/+namespaces/hierarchical-namespaces.injectable.ts
```diff
@@ -0,0 +1,18 @@
...
+import { getInjectable } from "@ogre-tools/injectable";
+import namespaceStoreInjectable from "./store.injectable";
+
+const hierarchicalNamespacesInjectable = getInjectable({
+  id: "hierarchical-namespaces",
+
+  instantiate: (di) => {
+    const namespaceStore = di.inject(namespaceStoreInjectable);
+
+    return namespaceStore.getByLabel(["hnc.x-k8s.io/included-namespace=true"]);
```
ixrock: Probably we should keep such label / api-related data as `hnc.x-k8s.io/included-namespace=true` closer to api definitions. So we don't need to create new injectable each time just for single method.
So this could be even just a method of `NamespaceStore.getNamespacesTree(optionalParamsForLabel = {})` or similar. Wdyt?

packages/core/src/renderer/components/+namespaces/subnamespace-badge.tsx
```diff
@@ -0,0 +1,36 @@
...
+      <span
+        className={styles.subnamespaceBadge}
+        data-testid={`subnamespace-badge-for-${namespace.getId()}`}
+        id={id}
+        {...other}
+      >
+        S
+      </span>
+      <Tooltip targetId={id}>
+        Subnamespace
```
ixrock, line 27: Looks like useless info (Subnamespace) for every namespace..
Why do we even need this component at all? I would define all these details directly in `<NamespaceTreeView/>`.
Also I don't like `if(!namespace.isSubnamespace()) return;` which looks like limit usages of a whole component in general.

packages/core/src/renderer/components/+namespaces/namespace-tree-view.tsx
```diff
@@ -0,0 +1,138 @@
...
+      >
+        <TreeItem
+          nodeId={nodeId}
+          label={root.getName()}
+          data-testid={nodeId}
+          classes={classes}
+          onIconClick={(evt) => {
+            toggleNode(nodeId);
+            evt.stopPropagation();
+          }}
```
ixrock: _nit:_ 
```suggestion
          onIconClick={prevDefault(() => toggleNode(nodeId))}
```

packages/core/src/renderer/components/+namespaces/namespace-tree-view.tsx
```diff
@@ -0,0 +1,138 @@
...
+}
+
+function NonInjectableNamespaceTreeView({ root, namespaces, getDetailsUrl }: Dependencies & NamespaceTreeViewProps) {
+  const [expandedItems, setExpandedItems] = React.useState<string[]>(namespaces.map(ns => `namespace-${ns.getId()}`));
+  const classes = { group: styles.group, label: styles.label };
+  const nodeId = `namespace-${root.getId()}`;
+
+  function renderChildren(parent: Namespace) {
+    const children = namespaces.filter(ns =>
+      ns.getLabels().find(label => label === `${parent.getName()}.tree.hnc.x-k8s.io/depth=1`),
```
ixrock: Again I'd better to keep such pieces of api as `*.tree.hnc.x-k8s.io/depth=1` closer to namespace api (for easier future maintenance at least)

packages/core/src/renderer/components/+namespaces/subnamespace-badge.tsx
aleksfront: > Looks like useless info (Subnamespace) for every namespace..

It is being shown not for all namespaces but for subnamepaces https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/quickstart.md#subnamespaces-deep-dive. 

> Why do we even need this component at all?

1. To separate it from `NamespaceTreeView` and its styles as well. To reach Single Responsibility principle a bit more.
2. It will be used in other places like namespace list.

> Also I don't like if(!namespace.isSubnamespace()) return; which looks like limit usages of a whole component in general.

Moved this check to parent component.

packages/core/src/renderer/components/+namespaces/hierarchical-namespaces.injectable.ts
aleksfront: Refactored the PR, moved all tree build logic to store and namespace API.

packages/core/src/renderer/components/+namespaces/namespace-tree-view.test.tsx
```diff
@@ -0,0 +1,334 @@
...
+      uid: `${name}`,
+      labels: {
+        ...labels,
+      },
+      annotations: {
+        ...annotations,
+      },
+    },
+  });
+}
```
ixrock, line 37: So why did you still chosen using `Namespace` object for `<NamespaceTreeView root={objectTreeItem}/>`?
Do we need something more than just a `namespace.getName()` from the `Namespace`-object (besides getting access to labels, etc)?

So I did suggested to to use for `objectTreeItem` only plain data types, so the exported tree should  be something like:

```ts
type TreeItem = {namespace: string, expanded?: boolean; children?: TreeItem[] }
```

In that case benefits of it:

- it's already exportable/serializable json format in case of persisting data or sending over network
- simplifies your test with the component so you just operate by namespace names there
- simplifies component itself (NamespaceTreeView) and renders plain data tree

Converting namespace to json-tree we can assign to `NamespaceStore` probably or some separated function-helper that could be easily unit-tested. 
Could be done in separated follow-up PR too. Wdyt?

ixrock (approved):

packages/core/src/renderer/components/+namespaces/namespace-tree-view.test.tsx
aleksfront: Make sense, will do the refactoring.

PR #7088, opened 2023-02-01
Export more things for the extension API

Nokel81:

We need to export these things (and others as we find them) because `@k8slens/extensions` is now a separate package and the older `@k8slens/extensions/dist/src/...` type exports (which we never officially supported anyway) no longer works.

Future work: 
- Turn on `requiredToBeDocumented`
- Turn on `treatWarningsAsErrors`
- Turn on `validation`


packages/core/src/extensions/common-api/utils.ts
```diff
@@ -7,20 +7,23 @@ import openLinkInBrowserInjectable from "../../common/utils/open-link-in-browser
...
-  openBrowser: asLegacyGlobalFunctionForExtensionApi(openLinkInBrowserInjectable),
-  getAppVersion: () => {
-    const di = getLegacyGlobalDiForExtensionApi();
-
-    return di.inject(buildVersionInjectable).get();
-  },
-} as const;
+
+export { Singleton } from "../../common/utils";
+export { prevDefault, stopPropagation } from "../../renderer/utils/prevDefault";
```
jim-docker, line 12: no longer exported as part of `Util`?

.github/workflows/publish-release-npm.yml
```diff
@@ -38,7 +38,9 @@ jobs:
       - name: Publish NPM packages
         run: |
           npm config set '//registry.npmjs.org/:_authToken' "${NPM_TOKEN}"
-          DIST_TAG=$(cat lerna.json | jq '.version' --raw-output | xargs npm exec -- @k8slens/semver --prerelease 0)
+          VERSION=$(cat lerna.json | jq '.version' --raw-output)
+          echo ${VERSION}
```
jim-docker, line 42: did you mean to leave this? Personally, I think it is a good idea, helps with debugging if there are issues in the future

packages/core/src/extensions/common-api/utils.ts
Nokel81: No they are, I changes the use from `{ Util }` to `* as Util`

jim-docker: ah right, in line with other types like this.

.github/workflows/publish-release-npm.yml
Nokel81: Yes I did mean to leave this in.

jakolehm (approved):

Issue #7065, opened 2023-01-31
Dependencies are duplicated in bundles


jansav:
**Describe the bug**
Current size of "open-lens" renderer -bundle is over 60 MB. There's duplicated instances of dependencies.
There must be something wrong with current setup.

**core/static/build/renderer.js**
<img width="1724" alt="image" src="https://user-images.githubusercontent.com/3083083/215991613-beab434c-bb65-4dbf-bd9b-0a22ef4cea84.png">

jansav:
There’s multiple things that need to be fixed in order to make it work properly.

1. Library package core needs to make all of it’s dependencies external, which means that they are not included in the bundle.
2. OpenLens monorepo needs to be changed to use Yarn/NPM workspaces which means that all dependencies are hoisted to root level. This ensures that there’s only one instance of a dependency. Without this, there’s e.g. react installed in packages/core/node_modules and packages/open-lens/node_modules .
3. The second bullet leads to unknown amount of stuff that needs to be fixed, because it seems that building core doesn’t work properly with workspaces  being enabled.

Nokel81:
Moving to 6.5.0 since the rest of this work can be delayed until after 6.4.0

PR #7094, opened 2023-02-02
Set core webpack dependencies as externals

jakolehm:
- configures `core` package dependencies as externals for webpack
- cleans up `core` webpack configuration (it was a mess)
- removes unused webpack plugins from `core`

Related to #7065


jansav (dismissed):

packages/core/webpack/renderer.ts
```diff
@@ -64,10 +45,7 @@ export function webpackLensRenderer({ showVars = true } = {}): webpack.Configura
       ],
     },
     externals: [
-      {
-        "win-ca": "commonjs win-ca",
-      },
-      ...additionalExternals,
+      nodeExternals(),
```
jansav, line 49: This is the fix.

packages/core/webpack/vars.ts
```diff
@@ -21,8 +21,5 @@ export const htmlTemplate = path.resolve(rendererDir, "template.html");
 export const publicPath = "/build/";
 export const sassCommonVars = path.resolve(rendererDir, "components/vars.scss");
 export const webpackDevServerPort = Number(process.env.WEBPACK_DEV_SERVER_PORT) || 9191;
-export const additionalExternals = (process.env.LENS_BUILD_ADDITIONAL_WEBPACK_EXTERNALS || "")
```
Nokel81, line 24: Is this not needed anymore?

jakolehm: I think we don't need it anymore here because everything is already external.

Nokel81 (approved):

PR #7095, opened 2023-02-02
Remove @k8slens/core/common

Nokel81:
Fix crashes with usages of @k8slens/core/common by moving all functionality from common to main/renderer entrypoints. Crashes were only observable with extensions (where state is shared between common-main and common-renderer)


jakolehm (approved):

packages/open-lens/src/renderer/application-information.injectable.ts
```diff
@@ -0,0 +1,17 @@
+
+import { getInjectable } from "@ogre-tools/injectable";
+import { applicationInformationToken, ApplicationInformation } from "@k8slens/core/renderer";
+import packageJson from "../../package.json";
+
+const applicationInformationInjectable = getInjectable({
+  id: "application-information",
+  injectionToken: applicationInformationToken,
+  instantiate: () => {
+    const { version, config, productName, build, copyright, description, name } = packageJson;
```
ixrock, line 10: Maybe we should make getting `package.json` data as injectable? _(instead of direct import)_

ixrock (approved):
LGTM

PR #7099, opened 2023-02-03
Fix common extension api bundling

jakolehm:
We cannot export common extension api through `@k8slens/core/common` because it shares state with main & renderer.

Fixes regression from #7058 .


jakolehm:
Closing because #7095 seems to be better fix.

jansav (approved):

PR #7101, opened 2023-02-03
Fix invalid bin links causing errors on windows

jakolehm:
Should fix windows test hangs on GH Actions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "test:integration": "lerna run --stream test:integration",
     "bump-version": "lerna version --no-git-tag-version --no-push",
     "create-release-pr": "node packages/release-tool",
-    "postinstall": "lerna bootstrap --ignore open-lens && lerna bootstrap --scope open-lens"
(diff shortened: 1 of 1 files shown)


aleksfront (dismissed):

package.json
```diff
@@ -19,7 +19,7 @@
     "test:integration": "lerna run --stream test:integration",
     "bump-version": "lerna version --no-git-tag-version --no-push",
     "create-release-pr": "node packages/release-tool",
-    "postinstall": "lerna bootstrap"
+    "postinstall": "lerna bootstrap --ignore open-lens && lerna bootstrap --scope open-lens --include-dependencies"
```
jakolehm, line 22: `open-lens` package needs to be bootstrapped as last because during install it creates `bin` wrappers from other packages in the monorepo. Without this hack bin wrappers are invalid on Windows on first install.

aleksfront (approved):

PR #7104, opened 2023-02-03
HMR/live-reload fixes (webpack-dev-server)

ixrock:
As part of #6948 since incorrectly rebased from first `monorepo` times.

**Current master:** _only manual page refresh, e.g`Cmd+R` (example of editing `welcome.tsx`)_

https://user-images.githubusercontent.com/6377066/216606292-41b1a294-bd73-462d-b1b0-29dce42532f0.mov

**PR changes:**

Does some auto page-reloading, but doesn't respect cluster frame location, therefore leads to updating page to global catalog page which is not desired, but better then nothing. _(needs to be improved in follow up PRs)_

https://user-images.githubusercontent.com/6377066/216606864-607a2ffd-a088-4329-8350-8be5a1712863.mov

**Current error leading to full page reload instead of HMR** _(caught from video above):_

Screenshot 2023-02-03 at 14 41 29

### DIFF
--- a/packages/open-lens/webpack/dev-server.ts
+++ b/packages/open-lens/webpack/dev-server.ts
@@ -23,20 +23,25 @@ const server = new WebpackDevServer({
   allowedHosts: ".lens.app",
   host: "localhost",
   port: webpackDevServerPort,
-  static: buildDir, // aka `devServer.contentBase` in webpack@4
-  hot: "only", // use HMR only without errors
+  static: {
+    directory: buildDir,
+    serveIndex: true,
+  },
+  hot: true,
   liveReload: false,
+  // historyApiFallback: true,
   devMiddleware: {
     writeToDisk: true,
     index: "index.html",
     publicPath: "/build",
   },
-  proxy: {
-    "^/$": "/build/",
-  },
+  // proxy: {
+  //   "^/$": "/build/",
+  // },
   client: {
     overlay: false, // don't show warnings and errors on top of rendered app view
-    logging: "error",
+    logging: "info",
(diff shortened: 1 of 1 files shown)


packages/open-lens/webpack/dev-server.ts
```diff
@@ -23,20 +23,25 @@ const server = new WebpackDevServer({
...
     writeToDisk: true,
     index: "index.html",
     publicPath: "/build",
   },
-  proxy: {
-    "^/$": "/build/",
-  },
+  // proxy: {
+  //   "^/$": "/build/",
+  // },
```
ixrock, line 40: @jakolehm why do we need this `proxy` at all? any ideas? it seem like works fine without this..

aleksfront (commented):
Not very useful with stated behavior: always reloading the page and moving to Catalog. Much easier to hit Cmd+R while in Cluster view because moving back from Catalog requires "bigger effor". 

Also, HMR works as expected now when changing CSS.

ixrock:
> Not very useful with stated behavior: always reloading the page and moving to Catalog. Much easier to hit Cmd+R while in Cluster view because moving back from Catalog requires "bigger effor".

@aleksfront Actually I think it should do page reloading in case when HMR cannot be applied and it works fine for app's global window (e.g. it keeps latest page, /welcome, /settings) but doesn't work inside cluster iframe properly for some reasons.. 
And that thing should be fixed or at least ignored _(handling HMR inside iframe differently)_, so you could do manual page reloading whenever you want.

**Any ideas why this error happens leading to full page reload?**

image

PR #7107, opened 2023-02-03
Enable extension to specify storeName for persisting data

### LINKED ISSUE #7052: Allow extension to specify storeName

panuhorsmalahti:
Fixes #7052 

* Alternative based on https://github.com/lensapp/lens/pull/7064
* If extensions specifies `storeName`, it is used for specifying extension data locations instead of extension name


packages/core/src/extensions/__tests__/configurable-directories.test.ts
```diff
@@ -0,0 +1,109 @@
...
+import type { ApplicationBuilder } from "../../renderer/components/test-utils/get-application-builder";
+import { getApplicationBuilder } from "../../renderer/components/test-utils/get-application-builder";
+import type { FakeExtensionOptions } from "../../renderer/components/test-utils/get-extension-fake";
+import logErrorInjectable from "../../common/log-error.injectable";
+import getHashInjectable from "../../extensions/extension-loader/file-system-provisioner-store/get-hash.injectable";
+import fsInjectable from "../../common/fs/fs.injectable";
+
+describe("configurable directories for extension files", () => {
+  let builder: ApplicationBuilder;
+  let logErrorMock: jest.Mock;
```
jweak: I this is not needed?

apialaviiva: Do you mean `logErrorMock`? That's right, will remove.

jweak: 👍

packages/core/src/extensions/lens-extension.ts
```diff
@@ -80,6 +87,11 @@ export class LensExtension<
     return this.manifest.description;
   }
 
+  // Name of extension for persisting data
+  get storeName() {
+    return this.manifest.storeName || this.name;
```
ixrock, line 92: _nit:_ 
```suggestion
    return this.manifest.storeName ?? this.name;
```

packages/core/src/extensions/extension-loader/file-system-provisioner-store/get-hash.injectable.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import { getInjectable } from "@ogre-tools/injectable";
+import { SHA256 } from "crypto-js";
+
+const getHashInjectable = getInjectable({
+  id: "get-hash",
+
+  instantiate: () => (text: string) => SHA256(text).toString(),
```
ixrock, line 12: Should we add `causesSideEffects: true` since used external imported package?

ixrock (approved):
LGTM

packages/core/src/extensions/extension-loader/file-system-provisioner-store/get-hash.injectable.ts
Nokel81: In don't think that is necessary because a hash should be stable given its input

packages/core/src/extensions/lens-extension.ts
apialaviiva: I don't think we want this. If `storeName` is set as `""`, we want to fall back to `name`.

jweak (approved):

PR #7116, opened 2023-02-06
Sidebar layout fixes and improved CRD readability

aleksfront:
* More room for item`s name by adjusting paddings, smaller icons and on-hover scrollbar
* Better readability for third level menus (CRDs)
* Consistent hover effects across menus and their submenus

Before:

https://user-images.githubusercontent.com/9607060/216939120-0e2e395d-37ca-4103-8d6a-105387473616.mov

After:

https://user-images.githubusercontent.com/9607060/216939143-265184f6-a492-4d6b-ac8a-17e02a5dce57.mov


Nokel81:
I would prefer about a little bit of padding and the spacing of the CRD names just so that the hierarchy is better visible.

ixrock:
Probably we should keep the blue line highlighting active sidebar's main section (if sub-section was selected, e.g. `Workloads -> Pods`). @miskun wdyt?

PR #7124, opened 2023-02-08
Fix: create/edit resource tab Save & Close button behavior

### LINKED ISSUE #6785: Save & Close should wait the result before closing the tab
Save & Close should wait the result before closing the tab

**Why is this needed**:
Currently when `Save & Close` is clicked, the resource tab closes regardless of its result and I have to rewrite config from the scratch if there are mistakes in config.

This behavior made me copying new config to another editor before clicking `Save & Close` or sometimes clicking `Save` instead of `Save & Close` then manually close the tab. I feel this kind of behavior makes `Save & Close` button unusable in almost every case.

- Desktop OS: Windows

aleksfront:
Currently `Save & Close` button doesn't respect result of save function. It closes tab even if error occurred.

save and close button

This PR adds following rules for `Save & Close`:
* `Create resource tab` closes only when **all resources** are successfully created
* `Edit resource tab` closes if no error happed during save.

https://user-images.githubusercontent.com/9607060/217512905-c7e7cd03-1493-4ed6-9113-562c90a20624.mov

https://user-images.githubusercontent.com/9607060/217513700-01ca1598-d943-42f6-9c91-5103f404804a.mov

Fixes #6785


packages/core/src/renderer/components/dock/create-resource/view.tsx
```diff
@@ -98,7 +98,7 @@ class NonInjectedCreateResource extends React.Component<CreateResourceProps & De
         this.props.logger.warn("Failed to create resource", { resource }, result.error);
         this.props.showCheckedErrorNotification(result.error, "Unknown error occured while creating resources");
 
-        return;
+        throw Error(result.error);
```
ixrock: Or maybe just `throw result.error` since we don't even considering rejected error below?

ixrock (dismissed):

ixrock (approved):

PR #7115, opened 2023-02-06
Switch to using NPM workspaces

jansav:
This PR will turn on NPM workspaces to prevent duplicate instances of same dependency in application bundle.

Fixes #7065

PR #7113, opened 2023-02-06
Switch to using npm over yarn

jansav:
This PR changes all usages of Yarn to NPM. This is done so that in future PR we can enable NPM workspaces. Yarn workspaces seem to install duplicate instances of same dependencies if there is:

- direct dependency to the package (e.g. `@types/react`) with specific version 
- dependency to some other package (e.g. `@types/react-router`) that depends from the same dependency but with `*` (like all other `@types/*` packages.)

Related to #6801 

Notice the negative side, NPM doesn't seem to create `package-lock.json`'s for packages.

PR #7133, opened 2023-02-09
use correct lerna workspace setup

proletarius101:
- `lerna bootstrap` is to be deprecated. The `yarn workspace` should be used instead: https://lerna.js.org/docs/getting-started#package-dependency-management
- `"@k8slens/semver": "./packages/semver"` syntax is not universally supported. (In fact, it's not supported by [flatpak_node_generator](https://github.com/flatpak/flatpak-builder-tools/tree/master/node) that I use to [publish this application to flathub.org](https://github.com/flathub/flathub/pull/3811). Of course, it would be [more standard-compliant](https://yarnpkg.com/features/workspaces#workspace-ranges-workspace) to use `workspace:` when you upgrade to yarn 2.

Note: Although I updated the `yarn.lock` in this PR, I would prefer the repo owner to update it instead, for security concerns.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "$schema": "node_modules/lerna/schemas/lerna-schema.json",
-  "useWorkspaces": false,
+  "useWorkspaces": true,
   "packages": [
     "packages/*"
   ],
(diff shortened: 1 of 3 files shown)


jansav:
Closing for being solved in #7113 and  #7115

PR #7137, opened 2023-02-10
Ability to remove subnamespaces

### LINKED ISSUE #7114: Ability to delete subnamespaces
For better hierarchical namespace controller support we should provide a way to easily remove a subnamespace. It is listed in `Namespaces` page just like a regular namespaces, but associated CRD should be removed first in this case.

https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/quickstart.md#subnamespaces-deep-dive

Part of #2658

aleksfront:
Removing [subnamespaces](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/quickstart.md#subnamespaces-deep-dive) in UI by removing parent anchor which lives as a CRD resource.

Previously there was no clear way to remove a subnamespace:

https://user-images.githubusercontent.com/9607060/218083309-eca4ec75-2150-4039-8f74-0bdc5697c1ea.mov

Currently subnamespaces being removed visually same way as "regular" ones:

https://user-images.githubusercontent.com/9607060/218083457-56847310-355d-4f36-b13c-4eb5ba09e315.mov

**Note**. Cascading deletion (remove subnamespace which contains children) not implemented here.

no cascading deletion

Resolves #7114


packages/core/src/renderer/components/+namespaces/namespace-store.test.ts
```diff
@@ -137,6 +153,10 @@ describe("NamespaceStore", () => {
     ]);
   });
 
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
```
Nokel81: This should not be necessary. This implies that there is shared state between tests.

packages/core/src/renderer/components/+namespaces/namespace-store.test.ts
```diff
@@ -99,17 +106,25 @@ const levelDeepSubChildA = createNamespace("level-deep-subchild-a", {
   "level-deep-subchild-a.tree.hnc.x-k8s.io/depth": "0",
 });
 
+const deleteNamespaceMock = jest.spyOn(NamespaceApi.prototype, "delete")
+  .mockImplementation();
+
+const removeSubnamespaceMock = jest.fn();
```
Nokel81: These things should be initialised in a `beforeEach` hook.

packages/core/src/renderer/components/+namespaces/remove-subnamespace.injectable.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Removing subnamespace controlled by hierarchical namespace controller by deleting
+     * the SubnamespaceAnchor from the parent namespace. Anchor is CRD with the same name
+     * https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/quickstart.md#subnamespaces-deep-dive
+     * @param subnamespaceName namespace name
+     * @returns 
+     */
+    return async (subnamespaceName: string) => {
+      await crdStore.loadAll();
```
Nokel81, line 26: This seems unnecessary to do every single time especially since we are already always subscribed to CRDs for the sidebar

packages/core/src/renderer/components/+namespaces/store.ts
```diff
@@ -220,6 +221,12 @@ export class NamespaceStore extends KubeObjectStore<Namespace, NamespaceApi> {
 
   @action
   async remove(item: Namespace) {
+    if (item.isSubnamespace()) {
+      this.dependencies.removeSubnamespace(item.getName());
```
Nokel81: This should be `await`-ed

packages/core/src/renderer/components/+namespaces/remove-subnamespace.injectable.ts
```diff
@@ -0,0 +1,46 @@
...
+
+      const anchor = store?.getByName(subnamespaceName);
+
+      if (!anchor) {
+        logger.error("[REMOVING-SUBNAMESPACE]: Error removing namespace: can not find subnamespace anchor.");
+        
+        return;
+      }
+
+      store?.remove(anchor);
```
Nokel81: Same here, I think we probably should turn on `no-floating-promises`

packages/core/src/renderer/components/+namespaces/remove-subnamespace.injectable.ts
aleksfront: We can't find proper CRD instance without this.

packages/core/src/renderer/components/+namespaces/namespace-store.test.ts
aleksfront: This is needed to reset

```
deleteNamespaceMock = jest.spyOn(NamespaceApi.prototype, "delete").mockImplementation();
```

Nokel81: A better way to move around this would be to do the following:

```ts
beforeEach(() => {
  deleteNamespaceMock = jest.fn();
  di.override(namespaceApiInjectable, () => ({
    delete: deleteNamespaceMock,
  } as any);
});
```

aleksfront:
After some updates injecting `apiManager` in namespace store causing circular dependency bug:
circular dependency error



Finding a way how to avoid this.

Nokel81:
@aleksfront The way to resolve that error is to remove the `this.dependencies.removeSubnamespace(item.getName());` from `remove` and instead have a new dependency called `deleteNamespace(namespace: Namespace) => Promise<void>` so that `NamespaceStore` does not need to have `removeSubnamespace` as a dependency.

packages/core/src/renderer/components/+namespaces/remove-subnamespace.injectable.ts
Nokel81: When in tests or not?

aleksfront: Not only in tests.

aleksfront:
@Nokel81 After couple of different attempts ended up created a wrapper injectable for `removeSubnamespaceInjectable`. This allows to get from circular dependency error.

packages/core/src/renderer/components/+namespaces/remove-subnamespace.injectable.ts
Nokel81: I really don't think we should be loading all in a delete, especially not twice (like bellow)

packages/core/src/renderer/components/+namespaces/store.ts
```diff
@@ -210,7 +211,12 @@ export class NamespaceStore extends KubeObjectStore<Namespace, NamespaceApi> {
 
   @action
   async remove(item: Namespace) {
-    await super.remove(item);
+    if (item.isSubnamespace()) {
+      this.dependencies.deleteSubnamespace(item.getName());
```
Nokel81: I don't think this belongs here. `removeNamespace` should be an injectable byitself which does this check and then calls `deleteSubnamespace`.

Futhermore, this is a floating promise.

Nokel81 (changes requested):

packages/core/src/renderer/components/+namespaces/store.ts
aleksfront: Removing a regular namespace made by using store's `super.remove()` method. If I move all of the logic above to the separate injectable, I have to call store method again.

aleksfront: Fixed floating promise.

Nokel81: Calling the store method is not the problem, maybe I am misunderstanding what you are meaning.

I would think that we should have the following:

```
const removeNormalNamespaceInjectable = ...; // This is a wrapper around `NamespaceStore.remove`
const removeSubnamespaceMarkerInjectable = ...; // This is the code that removes the subnamespace marker

const removeNamespaceInjectable = (namespace) => {
  if (namespace.isSubnamespace()) {
    await removeSubnamespaceMarker();
  }

  await removeNormaltNamespace();
};
```

PR #7154, opened 2023-02-14
Replication Controllers (new resource view)

### LINKED ISSUE #3998: Add support for ReplicationController

Have the possibility to view some additional k8s resources such as `ReplicationController`

- Kubernetes distribution: K8s 1.21 cluster created using kind tool
- Desktop OS: Macos

ixrock:
Screenshot 2023-02-16 at 13 36 05

Screenshot 2023-02-16 at 13 36 00

**_API Docs:_**

- https://kubernetes.io/docs/concepts/workloads/controllers/replicationcontroller/
- https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#replicationcontroller-v1-core

close https://github.com/lensapp/lens/issues/3998


packages/core/src/common/k8s-api/endpoints/replication-controller.api.ts
```diff
@@ -37,9 +37,25 @@ export class ReplicationControllerApi extends KubeApi<ReplicationController> {
           replicas,
         },
       },
+    }, {
+      headers: {
+        "content-type": "application/strategic-merge-patch+json",
+      },
     });
```
Nokel81: nit: this is an unnecessary change since it is already the default

packages/core/src/renderer/components/+workloads-replicationcontrollers/replicationcontroller-details.tsx
```diff
@@ -5,24 +5,114 @@
...
+    return api.scale({
+      name: resource.getName(),
+      namespace: resource.getNs(),
+    }, replicas);
+  }
+
+  renderReplicasAndScaleDialog() {
+    const { object: resource, showNotificationError, hideDetails } = this.props;
+
+    if (this.showScaleDialog) {
```
Nokel81: I think it would be a better UX to have a slider + scale button instead of a text input field

ixrock: Not sure.. 
It's numeric number input, arrow(s)-up/bottom increases and decreases value with default `step={1}`.
So for me it's very easy to use this UX by changing slightly value with arrows or typing arbitrary number.

Nokel81: Fair I guess. I am thinking mostly for consistency with the other scale dialogs. The nice thing about a slide is that a user cannot accidentally input a letter.

The other issue that I saw was that the buttons seemed a bit big and that the error message made the buttons jump around. (But those don't need to be fixed if we change how the input is entered).

A third option would be to use a slider but allow the user to enter digits after clicking the drag point which would make the drag point move around (with ESC as a reset to old value)

ixrock: Ok, I'll try to play around with slider and see how does it look & feel.

ixrock: Updated with using Slider component, PTAL

Nokel81 (approved):

Nokel81:
> resource object (replication controller) don't gets updated inside details panel after scaling / from watch-api events (but if we close/open then all up to date)

This can be fixed in a follow up PR. I have a feeling that it is not related to replication controllers themselves.

PR #7164, opened 2023-02-15
Introduce clearer boundry between extensions

### LINKED ISSUE #7471: Split internal typings of Bundled and non bundled legacy extensions
Motivation:
- The requirements for creation are different
- The source/discovery mechanisms are different

Nokel81:
- Bundled extensions are always enabled, and are always compatible
- Have bundled extensions be loaded asyncronously to support typescript dynamic import (which is typed) as opposed to require


packages/core/src/common/protocol-handler/router.ts
```diff
@@ -209,7 +209,7 @@ export abstract class LensProtocolRouter {
       return name;
     }
 
-    if (!this.dependencies.extensionsStore.isEnabled(extension)) {
+    if (!extension.isBundled && !this.dependencies.extensionsStore.isEnabled(extension.id)) {
```
jim-docker: If it is bundled and not enabled that would be an internal error? I guess you prevent this elsewhere.

jim-docker:
Still reviewing but noticed this while testing:


https://user-images.githubusercontent.com/40840436/219488889-2457f7f1-d80f-46dc-befd-030d3c9ffc16.mov

Nokel81:
Thanks! Will look into that

jim-docker:
Do we really need to introduce this this close to release?

Nokel81:
Maybe not. Might indeed be better to hold off until after. This was motivated by the need for documentation so it can hold off.

Nokel81:
> Still reviewing but noticed this while testing:

Seems like it is merely a state data issue.

packages/core/src/renderer/components/+extensions/extensions.tsx
```diff
@@ -4,114 +4,66 @@
...
-import extensionInstallationStateStoreInjectable from "../../../extensions/extension-installation-state-store/extension-installation-state-store.injectable";
-import type { ExtensionInstallationStateStore } from "../../../extensions/extension-installation-state-store/extension-installation-state-store";
 import Gutter from "../gutter/gutter";
 
+const ExtensionsNotice = () => (
+  <Notice className={styles.notice}>
+    <p>
+      {"Add new features via Lens Extensions. Check out the "}
+      <a
+        href={`${docsUrl}/extensions/`}
```
jim-docker: I noticed this is 404 now, maybe
```
     href={`${docsUrl}/extensions/lens-extensions`}
```
?

packages/core/src/extensions/extension-loader/extension-loader.ts
```diff
@@ -203,17 +171,19 @@ export class ExtensionLoader {
     const extension = this.extensions.get(lensExtensionId);
 
     assert(extension, `Must register extension ${lensExtensionId} with before enabling it`);
```
jim-docker: I know this is old but there seems to be a word missing or extra?

jim-docker (commented):
just need to test...

packages/core/src/common/protocol-handler/router.ts
Nokel81: If it is `bundled` then it is always enabled and cannot be disabled.

gabriel-mirantis (approved):

PR #7165, opened 2023-02-15
Simplify extension dependency injection 

### LINKED ISSUE #7472: Switch to using global DI references for legacy extension creation
Motivation:
- Simplifies code around creation
- Has better typing experience (less casts)
- Removes use of unusual unique symbol for dependency injection

Nokel81:
- Has better typing
- Removes use of unnusual unique symbol

follow up to https://github.com/lensapp/lens/pull/7164

To get the better unit tests to pass I had to replace the `TreeView` and `TreeItem` from `@material-ui/lab` for a custom one. They look and feel very similar:

https://user-images.githubusercontent.com/8225332/221031316-78f41546-846c-428a-92e4-dd0c47f0c347.mov


packages/core/src/extensions/__tests__/lens-extension.test.ts
```diff
@@ -3,17 +3,18 @@
...
+import { LensMainExtension } from "../lens-main-extension";
 
 describe("lens extension", () => {
+  let ext: LensMainExtension;
+
   beforeEach(async () => {
-    ext = new LensExtension({
+    const builder = getApplicationBuilder();
+
+    await builder.render();
```
ixrock: Why do we need to render application and afterwards do not use the result or `builder` apis? In other words it's not clear if `new LensMainExtension` below depends somehow on that code.

packages/core/src/extensions/ipc/ipc-main.ts
```diff
@@ -4,16 +4,31 @@
...
+  readonly logger: Logger;
+}
 
 export abstract class IpcMain extends IpcRegistrar {
+  private readonly dependencies: Dependencies;
+
   constructor(extension: LensMainExtension) {
     super(extension);
 
+    const di = getEnvironmentSpecificLegacyGlobalDiForExtensionApi(Environments.main);
```
ixrock: Should we still keep using something with `*Legacy*` in a name? Is there any other ways to get access to `di` within the class? Maybe receive it via second argument in `constructor`?

packages/core/src/extensions/lens-main-extension.ts
```diff
@@ -3,19 +3,49 @@
...
+}
+
+export class LensMainExtension extends LensExtension {
   appMenus: MenuRegistration[] | IComputedValue<MenuRegistration[]> = [];
   trayMenus: TrayMenuRegistration[] | IComputedValue<TrayMenuRegistration[]> = [];
 
+  /**
+   * @ignore
+   */
+  declare readonly dependencies: LensMainExtensionDependencies;
```
ixrock, line 35: Looks like a hack.. Why to ignore + declare?

packages/core/src/extensions/lens-extension.ts
```diff
@@ -68,12 +73,12 @@ export class LensExtension<
    */
   [Disposers] = disposer();
 
-  constructor({ id, manifest, manifestPath, isBundled }: InstalledExtension) {
-    // id is the name of the manifest
+  constructor(deps: LensExtensionDependencies, { id, manifest, manifestPath, isBundled }: InstalledExtension) {
+    this.dependencies = deps;
```
ixrock, line 49: _nit:_
```suggestion
  constructor(protected dependencies: LensExtensionDependencies, { id, manifest, manifestPath, isBundled }: InstalledExtension) {
```

packages/core/src/renderer/components/+catalog/catalog-entity-store.injectable.ts
```diff
@@ -0,0 +1,91 @@
...
+
+const catalogEntityStoreInjectable = getInjectable({
+  id: "catalog-entity-store",
+
+  instantiate: (di): CatalogEntityStore => {
+    const entityRegistry = di.inject(catalogEntityRegistryInjectable);
+    const catalogRegistry = di.inject(catalogCategoryRegistryInjectable);
+
+    const activeCategory = observable.box<CatalogCategory | undefined>(undefined);
+    const selectedItemId = observable.box<string | undefined>(undefined);
```
ixrock: _nit:_
```suggestion
    const activeCategory = observable.box<CatalogCategory | undefined>();
    const selectedItemId = observable.box<string | undefined>();
```

packages/core/src/renderer/components/tree-view/tree-view.tsx
```diff
@@ -3,103 +3,131 @@
...
-  children?: NavigationTree[];
+export interface TreeViewClasses {
+  root?: string;
 }
 
-export interface RecursiveTreeViewProps {
-  data: NavigationTree[];
+export interface TreeViewProps {
+  classes?: TreeViewClasses;
+  children: JSX.Element[] | JSX.Element;
```
ixrock:
```suggestion
  children: React.ReactNode;
```

_From `lens/node_modules/@types/react/index.d.ts`_:

```tsx
    type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;
    type ReactChild = ReactElement | ReactText;
    type ReactFragment = {} | Iterable<ReactNode>;
    type ReactText = string | number;
```

packages/core/src/renderer/components/tree-view/tree-view.tsx
```diff
@@ -3,103 +3,131 @@
...
+              ? props.collapseIcon ?? <Icon material="expand_more" />
+              : props.expandIcon ?? <Icon material="chevron_right" />
+          }
+        </div>
+        <div className={props.classes?.label}>
+          {props.label}
+        </div>
+      </div>
+      <ul
+        className={cssNames(props.classes?.contents, styles.contents, {
```
ixrock: `props.classes` is repeated 6 times here. Recommend you to extract it in constructor:

```tsx
export function TreeGroup({classes = {}, children, ...props}: TreeGroupProps)
```

ixrock (approved):

packages/core/src/extensions/__tests__/lens-extension.test.ts
Nokel81: Oh good point. This name probably could be improved. What it actually does is to fake the start of the application and the opening of the first application window.

Though now that you mention it. It is probably not required for this test suite either

packages/core/src/extensions/ipc/ipc-main.ts
Nokel81: So normally we wouldn't want to access DI like this at all. However, this class is exported to the extension API so it's constructor params cannot change.

packages/core/src/extensions/lens-main-extension.ts
Nokel81: The `@ignore` is there so that we don't generate documentation for that field.

The declare is there because it will be assigned in the base class only.

packages/core/src/extensions/ipc/ipc-main.ts
ixrock: So does it mean we are stuck with `getEnvironmentSpecificLegacyGlobalDiForExtensionApi` forever?

Nokel81: Not necessarily. Once we decide to make a major version change to the extension API then we can remove it.

packages/core/src/extensions/lens-extension.ts
Nokel81: I don't think typedoc supports `@ignore` annotations on constructor args.

packages/core/src/extensions/__tests__/lens-extension.test.ts
Nokel81: This is required because we need to setup app paths before creating an extension instance. Will add a comment

packages/core/src/renderer/components/tree-view/tree-view.tsx
```diff
@@ -47,33 +47,41 @@ export interface TreeItemProps {
...
-      data-testid={props["data-testid"]}
-      onClick={props.onClick}
+      data-testid={dataTestId}
+      onClick={onClick}
       onMouseOver={() => setHovering(true)}
       onMouseLeave={() => setHovering(false)}
     >
-      <div className={cssNames(props.classes?.iconContainer, styles.iconContainer)}>
-        {props.icon}
+      <div className={cssNames(classes?.iconContainer, styles.iconContainer)}>
```
ixrock: If you would set up default value `classes={}` in props destruction you wouldn't need to repeat `?` everywhere to check object existence.

ixrock (dismissed):

packages/core/src/extensions/ipc/ipc-main.ts
jansav: @ixrock this is function that should only be used when there's no other way to support the Extension API. There will be point of time that we remove the support for old Extension API and that's the time when we will remove these functions as well. However, there's still a lot of stuff that has to happen before that (e.g. Runtime Features - the Extension API 2.0)

jansav:
How much work we want to do to support old extensions? The path is clear with Features, so I don't think we should do a lot here anymore. It's just a way to bring in regressions. We could use that time to extract Features and therefore broaden the new way of extending the application which then allows extension developers to start the process of migrating.

jansav (dismissed):
Looks fine.

gabriel-mirantis (dismissed):

gabriel-mirantis (approved):

PR #7168, opened 2023-02-15
Fix external extensions not loading initially

### LINKED ISSUE #7139: Enabled extension not fully loading on startup
With the GKE-sync extension, on starting Lens, the GKE status bar item does not appear. Disable and then re-enable the extension (or a different extension) to make the status bar item appear.

**Screenshots**

https://user-images.githubusercontent.com/40840436/218170213-aca2610c-35ec-42ec-bff5-ba0f8b77008c.mov

- Lens Version: 2023.2.101253-alpha

Nokel81:
fixes https://github.com/lensapp/lens/issues/7139

### DIFF
--- a/packages/core/src/renderer/extension-discovery/init.injectable.ts
+++ b/packages/core/src/renderer/extension-discovery/init.injectable.ts
@@ -1,0 +1,24 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import extensionDiscoveryInjectable from "../../extensions/extension-discovery/extension-discovery.injectable";
+import { beforeFrameStartsSecondInjectionToken } from "../before-frame-starts/tokens";
+import initializeExtensionLoaderInjectable from "../extension-loader/init.injectable";
+
+const initializeExtensionDiscoveryInjectable = getInjectable({
+  id: "initialize-extension-discovery",
+  instantiate: (di) => ({
+    id: "initialize-extension-discovery",
+    run: async () => {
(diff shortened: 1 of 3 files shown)


packages/core/src/renderer/extension-discovery/init.injectable.ts
```diff
@@ -0,0 +1,24 @@
...
+  instantiate: (di) => ({
+    id: "initialize-extension-discovery",
+    run: async () => {
+      const extensionDiscovery = di.inject(extensionDiscoveryInjectable);
+
+      await extensionDiscovery.init();
+    },
+    runAfter: di.inject(initializeExtensionLoaderInjectable),
+  }),
+  injectionToken: beforeFrameStartsSecondInjectionToken,
```
jim-docker, line 21: How does one know to use `beforeFrameStartsSecondInjectionToken` over `beforeFrameStartsFirstInjectionToken`?

jim-docker (approved):
tested on mac. ~This might also fix https://github.com/lensapp/lens/issues/7071? Seemed to for me, but I had both our and their pod menu items in the context menu~. Never mind, I can reproduce that problem

packages/core/src/renderer/extension-discovery/init.injectable.ts
Nokel81: If it depends on something that runs in one of the cluster/main frames ifs above in `packages/core/src/renderer/start-frame/start-frame.injectable.ts`

Nokel81: Hopefully once we remove iframes this can be simplified.

Nokel81:
Will recreate this PR simpler to fix just the issue (but not via Runnables) because the necessary work to fix the tests is huge

PR #7176, opened 2023-02-16
Refresh release data on every details panel open

### LINKED ISSUE #6197: Helm release details page is stale after upgrade
https://user-images.githubusercontent.com/40840436/188977419-0757fd40-be9e-4b2a-b9a4-ba20dd73b866.mov

Observed on 6.0.1-latest

aleksfront:
Release data should be refreshed every time details panel opened.

https://user-images.githubusercontent.com/9607060/219301623-c57e9dbb-3af4-43f9-b955-b2c0cf959724.mov

   
In case of load error

Screenshot 2023-02-16 at 10 46 39

Fixes #6197 

Also, fixed bug with non-unique keys for `TableRow`:

Screenshot 2023-02-15 at 11 13 55


packages/core/src/renderer/components/+helm-releases/release-details/release-details-content.tsx
```diff
@@ -35,6 +35,10 @@ interface Dependencies {
 const NonInjectedReleaseDetailsContent = observer(({ model }: Dependencies & ReleaseDetailsContentProps) => {
   const loadingError = model.loadingError.get();
 
+  useEffect(() => {
+    model.load();
+  }, []);
```
Nokel81: Do we not call `model.load()` anywhere else? It seems strange that we should have to call it here again.

aleksfront: We call it in `releaseDetailsModelInjectable` once https://github.com/lensapp/lens/blob/master/packages/core/src/renderer/components/%2Bhelm-releases/release-details/release-details-model/release-details-model.injectable.tsx#L53

Adding `model.load()` here mean that we fetch release details every time this component rendered == release details opened.

Nokel81: A better fix would be to use `asyncComputed` which automatically gets called again when it goes from UNOBSERVED -> OBSERVED. That way this technical details don't get polluted into the component.

aleksfront: Unable to do it with `asyncComputed`.

Moved loading away from model injectable to `getProps` inside `ReleaseDetailsContent`.

aleksfront:
Closing in favor of https://github.com/lensapp/lens/pull/7206

PR #7185, opened 2023-02-17
Refactor electron window setCertificateVerifyProc

jakolehm:
Follow-up to #7180

### DIFF
--- a/packages/core/src/main/start-main-application/lens-window/application-window/create-electron-window.injectable.ts
+++ b/packages/core/src/main/start-main-application/lens-window/application-window/create-electron-window.injectable.ts
@@ -3,7 +3,6 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
 import { getInjectable } from "@ogre-tools/injectable";
-import { timingSafeEqual, X509Certificate } from "crypto";
 import loggerInjectable from "../../../../common/logger.injectable";
 import applicationWindowStateInjectable from "./application-window-state.injectable";
 import { BrowserWindow } from "electron";
@@ -14,8 +13,8 @@ import getAbsolutePathInjectable from "../../../../common/path/get-absolute-path
 import lensResourcesDirInjectable from "../../../../common/vars/lens-resources-dir.injectable";
 import isLinuxInjectable from "../../../../common/vars/is-linux.injectable";
 import pathExistsSyncInjectable from "../../../../common/fs/path-exists-sync.injectable";
-import lensProxyCertificateInjectable from "../../../../common/certificate/lens-proxy-certificate.injectable";
 import { applicationInformationToken } from "@k8slens/application";
+import sessionCertificateVerifierInjectable from "./session-certificate-verifier.injectable";
 
 export type ElectronWindowTitleBarStyle = "hiddenInset" | "hidden" | "default" | "customButtonsOnHover";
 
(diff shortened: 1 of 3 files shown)


packages/core/src/main/start-main-application/lens-window/application-window/session-certificate-verifier.injectable.ts
```diff
@@ -0,0 +1,36 @@
...
+import { timingSafeEqual, X509Certificate } from "crypto";
+import type { Request } from "electron";
+import lensProxyCertificateInjectable from "../../../../common/certificate/lens-proxy-certificate.injectable";
+
+// see https://www.electronjs.org/docs/latest/api/session#sessetcertificateverifyprocproc
+export enum ChromiumNetError {
+  SUCCESS = 0,
+  FAILURE = -2,
+  RESULT_FROM_CHROMIUM = -3,
+}
```
Iku-turso, line 15: I wonder if there's an official `enum` for this from eg. electron or chromium itself.

packages/core/src/main/start-main-application/lens-window/application-window/session-certificate-verifier.injectable.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export type CertificateVerificationCallback = (error: ChromiumNetError) => void;
+
+const sessionCertificateVerifierInjectable = getInjectable({
+  id: "session-certificate-verifier",
+  instantiate: (di) => {
+    const lensProxyCertificate = di.inject(lensProxyCertificateInjectable).get();
+    const lensProxyX509Cert = new X509Certificate(lensProxyCertificate.cert);
+
+    return (request: Request, shouldBeTrusted: CertificateVerificationCallback) => {
```
Iku-turso, line 25: This function could be simpler by being named `requestCanBeTrustedInjectable`, and returning `boolean` instead of bleeding abstraction of `ChromiumNetError`. That would mean that call to `shouldBeTrusted` happens somewhere else, and it would internally translate said boolean to appropriate `ChromiumNetError`. 

The resulting code could be dead-simple such as:
```
const thereIsTrust = requestCanBeTrusted(request);
shouldBeTrusted(thereIsTrust);
```

packages/core/src/main/start-main-application/lens-window/application-window/create-electron-window.injectable.ts
```diff
@@ -119,14 +110,7 @@ const createElectronWindowInjectable = getInjectable({
...
 
-      browserWindow.webContents.session.setCertificateVerifyProc((request, shouldBeTrusted) => {
-        const { certificate } = request;
-        const cert = new X509Certificate(certificate.data);
-        const shouldTrustCert = cert.raw.length === lensProxyX509Cert.raw.length
-          && timingSafeEqual(cert.raw, lensProxyX509Cert.raw);
-
-        shouldBeTrusted(shouldTrustCert ? ChromiumNetError.SUCCESS : ChromiumNetError.RESULT_FROM_CHROMIUM);
-      });
+      browserWindow.webContents.session.setCertificateVerifyProc(sessionCertificateVerifier);
```
Iku-turso, line 113: Responsibility extracted, nice :)

I spent a moment thinking if this responsibility of "certificate verification" should be abstracted to eg. `createLensWindowInjectable`, and to have no knowledge of Electron. But the implementation seems to be very Electron-specific, making this the correct place.

Iku-turso (dismissed):

packages/core/src/main/start-main-application/lens-window/application-window/session-certificate-verifier.injectable.ts
jakolehm: I was wondering the same, feels bad to include error codes here from chromium... but I could not find anything.

packages/core/src/main/start-main-application/lens-window/application-window/session-certificate-verifier.injectable.ts
Nokel81: I disagree. `boolean` is not a good fit because the return type either `YES` or `MAYBE`.

Nokel81 (dismissed):

Nokel81 (approved):

PR #7187, opened 2023-02-17
Fix: stale resource in details panel after update from apis/store

### LINKED ISSUE #7186: Stale resource in details panel after update from apis/store
**Describe the bug**
Opened resource in the details panel don't gets updated after receiving new object from watch-api/store/whatever.
On the video below I've tried to update resource with external terminal and via opened details panel and the resource was stale _(but updated in the items list if i close the panel)_.

**Expected behavior**
Updated resource is coming to props to details panel component and the render is up to date with api changes.

**Demo:**

https://user-images.githubusercontent.com/6377066/219623521-ab3e6ae1-ecee-47b1-90db-fa18da81834e.mov

ixrock:
https://user-images.githubusercontent.com/6377066/219681715-d1696ea9-82ce-458f-90f5-2e7559045d4e.mov


Nokel81:
Another solution would be something like `asyncSubscribableComputed` which @Iku-turso says is already mostly created for `computedChannel`. That way we don't have to fix each details panel individually and the contract of `currentKubeObjectForDetailsViewInjectable` stays the same.

ixrock:
> Another solution would be something like `asyncSubscribableComputed` which @Iku-turso says is already mostly created for `computedChannel`. That way we don't have to fix each details panel individually and the contract of `currentKubeObjectForDetailsViewInjectable` stays the same.

This is more complex solution/more abstraction which is not good. 
All what we need is just get rid of `currentKubeObjectInDetailsInjectable` _(replace to `currentKubeObjectForDetailsViewInjectable2` and rename)_

Nokel81:
You solution doesn't work though. If a `KubeObject` is opened in details without that list view being navigated to then the bug presented still happens.

ixrock:
> You solution doesn't work though. If a `KubeObject` is opened in details without that list view being navigated to then the bug presented still happens.

didn't get what you mean.. can you record some demo? or exact set of instructions how to represent the issue..

Nokel81:
Sure. I think the following will show the error.

1. Go to Pods
2. Click on a pod that is controlled by a deployment
3. Click on the link to the deployment
4. Update a label on that deployment
5. See that the details panel doesn't update.

aleksfront (commented):
LGTM once tests passed.

> Another solution would be something like asyncSubscribableComputed which @Iku-turso says is already mostly created for computedChannel.

Current solution in this PR is enough for now in my opinion for the 6.4.0. Described one probably will be ready for upcoming versions.

@Nokel81 Followed your example by changing labels on a deployment and it works:


https://user-images.githubusercontent.com/9607060/220049063-ff3da156-fe4e-4f43-9729-66e0099f6004.mov

ixrock:
@aleksfront for some reasons I've been able to reproduce the bug Sebastian was described above.. 

@Nokel81 is this the case? If so it's not updated after sequential navigations within opened detail-view, right?
Anyway I think could be even fixed in follow-up PR since now opening details panel works as expected and active object is up-to-date with it's store.

https://user-images.githubusercontent.com/6377066/220101739-eb6d19d6-b23c-461d-a6d0-ae7ff0aad01f.mov

ixrock:
@lensapp/lens-maintainers any ideas how to fix tests?

Nokel81:
Marking as `6.5.0` since this is a longer standing bug and this won't be blocking that release.

Nokel81:
https://github.com/mobxjs/mobx-utils#fromresource is what I think we should use for this case.

The URL seems to be something like: `http://127.0.0.1:52267/e54971f316520ea4fa8d54555edc3054/api/v1/nodes?fieldSelector=metadata.name%3Dlens-desktop-kube&resourceVersion=0&watch=true`

PR #7188, opened 2023-02-17
Fix ensure hashed directory for extension

### LINKED ISSUE #7052: Allow extension to specify storeName

jweak:
Earlier open lens versions stored `extension_data` directory based on path of extension's package.json. This causes problems because extensions have moved to new location. This ensures backward compatibility that extension will get the same directory than before the change and migrates the key to a new implementation that does not depend on the installed directory.

Fixes #7052


packages/core/src/common/utils/find-key-contains.injectable.ts
```diff
@@ -0,0 +1,27 @@
...
+      for (const [key, value] of entries) {
+        if (key.includes(text)) {
+          return value;
+        }
+      }
+
+      return undefined;
+    };
+  },
+});
```
Nokel81: nit: This doesn't need to be an injectable because it has no side effects and has no dependencies.

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.injectable.ts
```diff
@@ -24,8 +25,15 @@ const ensureHashedDirectoryForExtensionInjectable = getInjectable({
     const directoryForExtensionData = di.inject(directoryForExtensionDataInjectable);
     const ensureDirectory = di.inject(ensureDirInjectable);
     const getHash = di.inject(getHashInjectable);
+    const findKeyContains = di.inject(findKeyContainsInjectable)<string>;
 
     return async (extensionName, registeredExtensions) => {
```
Nokel81: Shouldn't this `extensionName` be the value from the `package.json`?

Nokel81 (commented):
I don't understand why this is needed.

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.injectable.ts
```diff
@@ -24,8 +25,15 @@ const ensureHashedDirectoryForExtensionInjectable = getInjectable({
...
+      const legacyDirPath = findKeyContains(registeredExtensions, extensionName);
+
+      if (legacyDirPath) {
+        return legacyDirPath;
```
jweak: instead of returning early here, we should still make sure that `ensureDir` is run after this

packages/core/src/common/utils/find-key-contains.injectable.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const findKeyContainsInjectable = getInjectable({
+  id: "find-key-contains",
+
+  instantiate: (): FindKeyContains => {
+    return <T>(map: Map<string, T>, text: string) => {
+      const entries = map.entries();
+
+      for (const [key, value] of entries) {
+        if (key.includes(text)) {
```
jweak: instead of using `includes` we should try to reconstruct the whole legacy key path.

something like this:
`path.join(directoryForUserData, "extension_data", name, "package.json") `

jweak: done

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.injectable.ts
jweak: done

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.injectable.ts
jweak: yeah this has been changed to reconstruct the legacy key

packages/core/src/common/utils/find-key-contains.injectable.ts
jweak: the file has been removed

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    runInAction(() => {
+      registeredExtensions.set("/some-directory-for-user-data/node_modules/some-extension-name/package.json", "some-directory");
+    });
+
+    await ensureHashedDirectoryForExtension(
+      "some-extension-name",
+      registeredExtensions,
+    );
+
+    expect(ensureDirMock).toHaveBeenCalledWith("some-directory");
```
Iku-turso: A "negation-test" with `expect(ensureDirMock).not.toHaveBeenCalled()` is missing :)

packages/core/src/extensions/extension-loader/file-system-provisioner-store/get-path-to-legacy-package-json.injectable.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import directoryForUserDataInjectable from "../../../common/app-paths/directory-for-user-data/directory-for-user-data.injectable";
+import joinPathsInjectable from "../../../common/path/join-paths.injectable";
+
+export type GetPathToLegacyPackageJson = (extensionName: string) => string;
+
+const getPathToLegacyPackageJson = getInjectable({
```
Iku-turso, line 11: This could be unit tested as part of `ensure-hashed-directory-for-extension.test.ts`, instead of coupling tests too tightly with implementation details.

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  });
+
+  it("given registered extension exists, returns existing directory", async () => {
+    runInAction(() => {
+      registeredExtensions.set("some-extension-name", "some-directory");
+    });
+
+    const actual = await ensureHashedDirectoryForExtension(
+      "some-extension-name",
+      registeredExtensions,
```
Iku-turso: Hmm. I wonder if `registeredExtensions` could be promoted to be a **dependency** instead of **function parameter** for a bit less complexity. There's only one set of those, right?

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  let ensureHashedDirectoryForExtension: EnsureHashedDirectoryForExtension;
+  let ensureDirMock: jest.Mock;
+  let registeredExtensions: ObservableMap<string, string>;
+
+  beforeEach(() => {
+    const di = getDiForUnitTesting({  doGeneralOverrides: true });
+
+    ensureDirMock = jest.fn();
+
+    di.override(ensureDirInjectable, () => ensureDirMock);
```
Iku-turso, line 25: Note: Another way to test `ensureDir` and its effects would be to test it using those **file system fakes**. Generally it leads to bit more credible, but also sometimes easier to write tests.

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    });
+    const actual = await ensureHashedDirectoryForExtension(
+      "some-extension-name",
+      registeredExtensions,
+    );
+
+    expect(actual).toBe("some-directory");
+  });
+
+  it("given extension directory was saved based on extension's package.json path, ensure dir is called with right parameter", async () => {
```
Iku-turso: Instead of "right parameter", one could document what makes the parameter right :)

packages/core/src/extensions/extension-loader/file-system-provisioner-store/get-path-to-legacy-package-json.test.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  beforeEach(() => {
+    const di = getDiForUnitTesting({ doGeneralOverrides: true });
+    const directoryForExtensionDataMock = jest.fn().mockReturnValue("/path/to/user/data");
+
+    di.override(directoryForUserDataInjectable, directoryForExtensionDataMock);
+    getPathToLegacyPackageJson = di.inject(getPathToLegacyPackageJsonInjectable);
+  });
+
+  describe("given extension name", () => {
```
Iku-turso: Describe is not required :)

packages/core/src/extensions/extension-loader/file-system-provisioner-store/get-path-to-legacy-package-json.injectable.ts
```diff
@@ -0,0 +1,24 @@
...
+  id: "get-path-to-legacy-package-json",
+
+  instantiate: (di): GetPathToLegacyPackageJson => {
+    const directoryForUserData = di.inject(directoryForUserDataInjectable);
+    const joinPaths = di.inject(joinPathsInjectable);
+
+    return (extensionName: string) => {
+      return joinPaths(directoryForUserData, "node_modules", extensionName, "package.json");
```
Iku-turso: Nit: `return` can be removed.

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    expect(actual).toBe("some-directory");
+  });
+
+  it("given registered extension does not exist, returns random directory", async () => {
+    const actual = await ensureHashedDirectoryForExtension(
+      "some-extension-name",
+      registeredExtensions,
+    );
+
+    expect(actual).not.toBe("some-directory");
```
Iku-turso: To test eg. `expect(actual).toBe("some-random-directory");` would be more credible.

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.test.ts
```diff
@@ -0,0 +1,82 @@
...
+import type { ObservableMap } from "mobx";
+import { observable, runInAction } from "mobx";
+import { getDiForUnitTesting } from "../../../main/getDiForUnitTesting";
+import type { EnsureHashedDirectoryForExtension } from "./ensure-hashed-directory-for-extension.injectable";
+import ensureHashedDirectoryForExtensionInjectable from "./ensure-hashed-directory-for-extension.injectable";
+import ensureDirInjectable from "../../../common/fs/ensure-dir.injectable";
+import directoryForExtensionDataInjectable from "./directory-for-extension-data.injectable";
+import directoryForUserDataInjectable from "../../../common/app-paths/directory-for-user-data/directory-for-user-data.injectable";
+
+describe("ensure-hashed-directory-for-extension", () => {
```
Iku-turso, line 15: I'm thinking if case of NPM-packages with a namespaced name (eg. `"name": "@some-namespace/some-package-name"`) will require their own tests and maybe implementations. WDYT?

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.injectable.ts
```diff
@@ -24,14 +26,24 @@ const ensureHashedDirectoryForExtensionInjectable = getInjectable({
...
+      let dirPath: string;
+
+      const legacyDirPath = getPathToLegacyPackageJson(extensionName);
+      const hashedDirectoryForLegacyDirPath = registeredExtensions.get(legacyDirPath);
+
+      if (hashedDirectoryForLegacyDirPath) {
+        dirPath = hashedDirectoryForLegacyDirPath;
+      } else {
+        dirPath = await getOrInsertWithAsync(registeredExtensions, extensionName, async () => {
+          const salt = (await randomBytes(32)).toString("hex");
```
Iku-turso: I don't think `randomBytes` should be `async` :)

packages/core/src/extensions/extension-loader/file-system-provisioner-store/get-path-to-legacy-package-json.test.ts
```diff
@@ -0,0 +1,34 @@
...
+    it("given some extension returns path for package.json in user data directory", () => {
+      const pathToExtensionPackageJson = getPathToLegacyPackageJson("some-extension");
+
+      expect(pathToExtensionPackageJson).toEqual("/path/to/user/data/node_modules/some-extension/package.json");
+    });
+
+    it("given some other extension returns path for package.json in user data directory", () => {
```
Iku-turso: Separating Given/When/Then with colons would make names of tests a bit more skimmable :)

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.injectable.ts
jweak: I think this is some kind of optimisation. 

> The crypto.randomBytes() method will not complete until there is sufficient entropy available. This should normally never take longer than a few milliseconds. The only time when generating the random bytes may conceivably block for a longer period of time is right after boot, when the whole system is still low on entropy.
> 
> This API uses libuv's threadpool, which can have surprising and negative performance implications for some applications; see the [UV_THREADPOOL_SIZE](https://nodejs.org/api/cli.html#uv_threadpool_sizesize) documentation for more information.
> 
> The asynchronous version of crypto.randomBytes() is carried out in a single threadpool request. To minimize threadpool task length variation, partition large randomBytes requests when doing so as part of fulfilling a client request.
> 

https://nodejs.org/api/crypto.html#cryptorandombytessize-callback

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.test.ts
jweak: Is there only one? I'm not sure. That is a part of `Store`, does it have some kind main / renderer thing going on with syncing stuff. Can we move `registeredExtensions` from the Store to a dependency?

packages/core/src/extensions/extension-loader/file-system-provisioner-store/get-path-to-legacy-package-json.injectable.ts
jweak: sure we can just get rid off the specific tests for this dependency

packages/core/src/extensions/extension-loader/file-system-provisioner-store/get-path-to-legacy-package-json.test.ts
jweak: this is test removed

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.injectable.ts
jweak: anyway moved to the sync version of the random bytes

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.test.ts
jweak: good catch, fixed

packages/core/src/extensions/extension-loader/file-system-provisioner-store/get-path-to-legacy-package-json.injectable.ts
jweak: done

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.test.ts
jweak: good call. done

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.test.ts
jweak: I'm not sure I follow. `ensureDirMock` is called always with something or am I missing something?

packages/core/src/extensions/extension-loader/file-system-provisioner-store/get-path-to-legacy-package-json.test.ts
jweak: got rid off this one

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.test.ts
Iku-turso: Ah! My mistake :)

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.test.ts
Iku-turso: There indeed is only one, because `fileSystemProvisionerStoreInjectable` is a `singleton` (that being the default lifecycle of an `injectable`). 

This means we could extract eg. `registeredExtensionsInjectable`, and make it a dependency for both `fileSystemProvisionerStoreInjectable` and `ensureHashedDirectoryForExtensionInjectable`.

Being diligent about separating dependencies and eg. function arguments will lead to a cleaner kitchen, as it creates more easily reusable functions, and hides irrelevant details out of sight.

packages/core/src/extensions/extension-loader/file-system-provisioner-store/ensure-hashed-directory-for-extension.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    it("returns existing directory", async () => {
+      const actual = await ensureHashedDirectoryForExtension(
+        "some-extension-name",
+        registeredExtensions,
+      );
+
+      expect(actual).toBe("some-directory");
+    });
+
+    it("ensure dir is called with some directory", async () => {
```
Iku-turso, line 72: "ensure dir" is a technical phenomenon, and in general, unit tests benefit more from describing motivations of the real world. That would mean eg.: `it("ensures that the said extension directory exists", ...)`.

Iku-turso (dismissed):

Iku-turso (approved):

PR #7205, opened 2023-02-21
Finish work to make `bootstrap` only use `startFrame`

Nokel81:

This PR also replaces the work done in #7189.


jansav:
I was thinking that do we want to do this because it opens up a possibility to regressions and even if we would do it, we would have to extract stuff to own Features. So what I'm suggesting, is that we would start extracting immediately, one by one and after everything is extracted, we remove the bootstrap. In my opinion, it would be easier to control and wouldn't lead to massive PRs because we can create one for introducing the Feature and one for removing the old implementation and switching to use the Feature. WDYT?

Nokel81:
Okay sure we can do that. I'll close this PR then.

PR #7206, opened 2023-02-21
Fix helm release data being stale after reopening details

### LINKED ISSUE #6197: Helm release details page is stale after upgrade

Nokel81:
Fixes https://github.com/lensapp/lens/issues/6197

Alternative to and closes https://github.com/lensapp/lens/pull/7176

### DIFF
--- a/packages/core/src/renderer/components/+helm-releases/release-details/release-details-model/release-details-model.injectable.tsx
+++ b/packages/core/src/renderer/components/+helm-releases/release-details/release-details-model/release-details-model.injectable.tsx
@@ -32,11 +32,14 @@ import activeThemeInjectable from "../../../../themes/active.injectable";
 import type { ToHelmRelease } from "../../to-helm-release.injectable";
 import toHelmReleaseInjectable from "../../to-helm-release.injectable";
 import { asyncComputed } from "@ogre-tools/injectable-react";
+import targetHelmReleaseInjectable from "../target-helm-release.injectable";
 
 const releaseDetailsModelInjectable = getInjectable({
   id: "release-details-model",
 
   instantiate: (di, targetRelease: TargetHelmRelease) => {
+    const detailsHelmRelease = di.inject(targetHelmReleaseInjectable);
+
     const model = new ReleaseDetailsModel({
       requestDetailedHelmRelease: di.inject(requestDetailedHelmReleaseInjectable),
       targetRelease,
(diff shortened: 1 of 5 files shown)


jim-docker (dismissed):
Confirmed it works, the version gets updated properly. I noticed that if the release is upgraded outside of lens while the details page is open it does not update, the version becomes stale. If you close and open the details page the details are correct again though, so this is an improvement.

Nokel81:
Yes. I don't know if there is a way to get helm to update us if an update has happened outside of polling though

packages/core/src/renderer/components/+helm-releases/release-details/release-details-model/release-details-model.injectable.tsx
```diff
@@ -50,9 +54,22 @@ const releaseDetailsModelInjectable = getInjectable({
...
+          await model.load();
+        }
+
+        return model;
+      },
+      betweenUpdates: "show-latest-value",
+    });
   },
 
   lifecycle: lifecycleEnum.keyedSingleton({
```
Iku-turso: Would we lose anything substantial if `lifecycle` here was `lifecycleEnum.transient`, and `instantiate` was `async` because of `await model.load()`? Then we wouldn't need `asyncComputed` at all, and `withInjectables` could be responsible of rendering a spinner (using `getPlaceholder`) while an async dependency (ie. a fresh instance of `ReleaseDetailsModel`) is being instantiated.

Iku-turso: Proposed solution would be simpler with less code, and less prone to bugging. Eg. existing solution in PR will likely still show stale data when reopening same "release details" after a long wait, as that will not cause new `model.load()`. And it will briefly show details of wrong release when changing to other release, because of `betweenUpdates: "show-latest-value"`.

Iku-turso (changes requested):
Behavioural unit tests are missing :/

Also [this](https://github.com/lensapp/lens/pull/7206/files#r1113991579) could be worthy endeavour.

Nokel81:
This is now blocked on upgrading the ogre-tools @ 15.1.0

PR #7211, opened 2023-02-22
Extract injection token for application information to separate NPM package 

jansav:
This fixes the issue that the injection token is duplicated in the bundle. 

This is prerequisite for updating ogre-tools packages to the latest version since it will again start using references.


jakolehm (approved):
LGTM

PR #7225, opened 2023-02-23
Export `getPodsByOwnerId` to extension API

apialaviiva:


packages/core/src/renderer/components/+workloads-cronjobs/store.ts
```diff
@@ -6,17 +6,24 @@
...
+  getPodsByOwnerId: GetPodsByOwnerId;
 }
 
 export class CronJobStore extends KubeObjectStore<CronJob, CronJobApi> {
   constructor(protected readonly dependencies: Dependencies, api: CronJobApi, opts?: KubeObjectStoreOptions) {
     super(dependencies, api, opts);
   }
 
+  getChildPods(cronJob: CronJob): Pod[] {
+    return this.dependencies.getPodsByOwnerId(cronJob.getId());
```
Nokel81: Wouldn't it be better to export `getPodsByOwnerId` to the extension API?

Nokel81 (approved):

PR #7235, opened 2023-02-24
Fix ApiManager not handling duplicate apiBases of KubeApis

Nokel81:

### DIFF
--- a/packages/core/src/common/k8s-api/__tests__/api-manager.test.ts
+++ b/packages/core/src/common/k8s-api/__tests__/api-manager.test.ts
@@ -19,6 +19,9 @@ import { KubeObject } from "../kube-object";
 import { KubeObjectStore } from "../kube-object.store";
 import maybeKubeApiInjectable from "../maybe-kube-api.injectable";
 
+// eslint-disable-next-line no-restricted-imports
+import { KubeApi as ExternalKubeApi } from "../../../extensions/common-api/k8s-api";
+
 class TestApi extends KubeApi<KubeObject> {
   protected async checkPreferredVersion() {
     return;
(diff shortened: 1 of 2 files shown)


packages/core/src/common/k8s-api/api-manager/api-manager.ts
```diff
@@ -41,19 +41,22 @@ export class ApiManager {
       const apis = chain(this.dependencies.apis.get().values())
         .concat(this.externalApis.values());
       const removedApis = new Set(this.apis.values());
+      const newState = new Map(this.apis);
```
panuhorsmalahti, line 44: [nit] `const newApis`

Nokel81: Technically it is also `oldApis` since we don't necessarily remove all of them.

jakolehm (approved):
LGTM

packages/core/src/common/k8s-api/__tests__/api-manager.test.ts
```diff
@@ -72,21 +75,48 @@ describe("ApiManager", () => {
...
       Object.defineProperty(kubeApi, "apiBase", { value: fallbackApiBase });
-      apiManager.registerApi(fallbackApiBase, kubeApi);
+      apiManager.registerApi(kubeApi);
 
       // Test that store is returned with new apiBase
       expect(apiManager.getStore(kubeApi)).toBe(kubeStore);
     });
   });
+
+  describe("technical tests for autorun", () => {
```
jim-docker, line 96: could add another test where the second api is removed and then `apiManager.getApi(apiBase)` returns `firstApi`?

jim-docker (approved):

packages/core/src/common/k8s-api/__tests__/api-manager.test.ts
Nokel81: Good idea

Nokel81: Actually this is not possible given that the apis are registered automatically and then never garbage collected.

PR #7237, opened 2023-02-24
Release 6.4.0-beta.17

Nokel81:
## 🚀 Features

- Fix ApiManager not handling duplicate apiBases of KubeApis (**[#7235](https://github.com/lensapp/lens/pull/7235)**) https://github.com/Nokel81
- Allow extensions to opt-out of KubeApi auto registering (**[#7217](https://github.com/lensapp/lens/pull/7217)**) https://github.com/Nokel81

## 🧰 Maintenance

- Various improvements to release-tool (**[#7232](https://github.com/lensapp/lens/pull/7232)**) https://github.com/Nokel81
- Remove all references to slack (**[#7233](https://github.com/lensapp/lens/pull/7233)**) https://github.com/Nokel81

PR #7238, opened 2023-02-24
General fixes for release-tool

Nokel81:
Found while making https://github.com/lensapp/lens/pull/7237

### DIFF
--- a/packages/release-tool/src/index.ts
+++ b/packages/release-tool/src/index.ts
@@ -5,7 +5,7 @@
  */
 import assert from "assert";
 import chalk from "chalk";
-import child_process from "child_process";
+import child_process, { spawn } from "child_process";
 import { readFile } from "fs/promises";
 import inquirer from "inquirer";
 import { createInterface, ReadLine } from "readline";
@@ -16,8 +16,20 @@ type SemVer = semver.SemVer;
 
 const { SemVer } = semver;
 const exec = promisify(child_process.exec);
-const spawn = promisify(child_process.spawn);
 const execFile = promisify(child_process.execFile);
+
+async function pipeExecFile(file: string, args: string[], opts?: { stdin: string }) {
+  const p = execFile(file, args);
+
+  p.child.stdout?.pipe(process.stdout);
+  p.child.stderr?.pipe(process.stderr);
+
+  if (opts) {
+    p.child.stdin?.end(opts.stdin);
+  }
+
+  await p;
+}
 
 interface GithubPrData {
   author: {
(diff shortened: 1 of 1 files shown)


packages/release-tool/src/index.ts
```diff
@@ -153,28 +183,28 @@ function sortExtendedGithubPrData(left: ExtendedGithubPrData, right: ExtendedGit
...
 async function getRelevantPRs(milestone: string, previousReleasedVersion: string): Promise<ExtendedGithubPrData[]> {
-  console.log("retreiving previous 500 PRs...");
+  console.log("retrieving previous 200 PRs...");
 
   const getMergedPrsArgs = [
     "gh",
     "pr",
     "list",
-    "--limit=500", // Should be big enough, if not we need to release more often ;)
+    "--limit=200", // Should be big enough, if not we need to release more often ;)
```
jim-docker: do we gain anything by reducing to 200? Is there an easy way to catch when it's not big enough?

jim-docker (commented):
What does it mean when this appears:
```
Please resolve conflicts in a separate terminal and then press enter here...

The previous cherry-pick is now empty, possibly due to conflict resolution.
If you wish to commit it anyway, use:

    git commit --allow-empty

Otherwise, please use 'git cherry-pick --skip'
On branch more-fixes-release-tool
Your branch is up to date with 'origin/more-fixes-release-tool'.

You are currently cherry-picking commit 2550b1efbd.
  (all conflicts fixed: run "git cherry-pick --continue")
  (use "git cherry-pick --skip" to skip this patch)
  (use "git cherry-pick --abort" to cancel the cherry-pick operation)

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   package-lock.json

no changes added to commit (use "git add" and/or "git commit -a")
Please resolve conflicts in a separate terminal and then press enter here...

```
The cherry-pick succeeded without conflicts? Or the commit was already in the branch? If it is already in the branch then it should probably be removed from the list of commits for the release. Feels like this should proceed to the next cherry-pick without user interaction.

packages/release-tool/src/index.ts
Nokel81: Slight speedup of this step.

Nokel81:
> The cherry-pick succeeded without conflicts? Or the commit was already in the branch? If it is already in the branch then it should probably be removed from the list of commits for the release. Feels like this should proceed to the next cherry-pick without user interaction.

This is just how git works. I don't want to try and parse git output....

I think that `git cherry-pick --continue` would be the correct course of action.

packages/release-tool/src/index.ts
Nokel81: Will revert this.

aleksfront (approved):

PR #7241, opened 2023-02-27
Upgrade and adapt to new version of libraries in ogre-tools

jansav:
This version brings back requirement for exact reference for injection tokens and injectables. It also allows future Features not have conflicting IDs for injectables automatically.


packages/core/src/common/__tests__/cluster-store.test.ts
```diff
@@ -262,6 +272,11 @@ describe("cluster-store", () => {
 
   describe("pre 3.6.0-beta.1 config with an existing cluster", () => {
     beforeEach(() => {
+      di.override(storeMigrationVersionInjectable, () => "3.6.0");
```
jansav, line 275: `override` has to happen before `inject`.

packages/core/src/common/__tests__/user-store.test.ts
```diff
@@ -90,6 +92,8 @@ describe("user store tests", () => {
 
       di.override(storeMigrationVersionInjectable, () => "10.0.0");
 
+      userStore = di.inject(userStoreInjectable);
```
jansav, line 95: `override` has to happen before `inject`

packages/core/src/common/k8s-api/__tests__/api-manager.test.ts
```diff
@@ -43,7 +43,7 @@ describe("ApiManager", () => {
     di.override(directoryForKubeConfigsInjectable, () => "/some-kube-configs");
     di.override(storesAndApisCanBeCreatedInjectable, () => true);
 
-    const createCluster = di.inject(createClusterInjectable);
+    const createCluster = di.inject(createClusterInjectionToken);
```
jansav, line 46: Reference equality is important here. Previously it worked because the implementation for `createClusterInjectionToken ` in `main` and `renderer` had same ID.

packages/core/src/common/utils/channel/channel.test.ts
```diff
@@ -199,21 +202,32 @@ describe("channel", () => {
...
+    const someOtherChannelListenerInjectable = getInjectable({
+      id: "some-other-channel-listener",
+
+      instantiate: () => ({
+        channel: testRequestChannel,
+        handler: () => () => "irrelevant",
+      }),
+
+      injectionToken: requestChannelListenerInjectionToken,
+    });
```
jansav, line 226: Mutating injectable ID is not possible and shouldn't be advertised.

packages/core/src/features/catalog/opening-entity-details.test.tsx
```diff
@@ -25,6 +25,17 @@ describe("opening catalog entity details panel", () => {
...
+      // TODO: remove once ClusterStore can be used without overriding it
+      windowDi.override(getClusterByIdInjectable, () => (clusterId) => {
+        if (clusterId === cluster.id) {
+          return cluster;
+        }
+
+        return undefined;
+      });
+    });
+
```
jansav, line 38: `override` has to happen before `inject`

packages/core/src/features/entity-settings/__snapshots__/showing-settings-for-correct-entity.test.tsx.snap
```diff
@@ -830,7 +830,40 @@ exports[`Showing correct entity settings when navigating to non-local cluster en
                 >
                   Proxy
                 </h2>
-                <section />
+                <section>
```
jansav, line 833: Notice that previously the snapshot was providing wrong output.

Nokel81 (approved):

PR #7242, opened 2023-02-27
Introduce package for Features

jansav:
This `feature-core` package will allow us to start extending the application without relying current Extension API.

Expect future PRs to showcase usage.

### DIFF
--- a/packages/technical-features/feature-core/README.md
+++ b/packages/technical-features/feature-core/README.md
@@ -1,0 +1,41 @@
+# @k8slens/feature-core
+
+Feature is set of injectables that are registered and deregistered simultaneously.
+
+## Install
+```bash
+$ npm install @k8slens/feature-core
+```
+
+## Usage
+
+```typescript
+import { createContainer } from "@ogre-tools/injectable"
+import { getFeature, registerFeature, deregisterFeature } from "@k8slens/feature-core"
+
+// Notice that this Feature is usually exported from another NPM package.
+const someFeature = getFeature({
+  id: "some-feature",
+  
+  register: (di) => {
+    di.register(someInjectable, someOtherInjectable);
+  },
+  
+  // Feature dependencies are automatically registered and 
+  // deregistered when necessary.
+  dependencies: [someOtherFeature] 
+});
+
+const di = createContainer("some-container");
+
+registerFeature(di, someFeature);
+
+// Or perhaps you want to deregister?
+deregisterFeature(di, someFeature);
+```
+
+## Need to know
+
+#### NPM packages exporting a Feature
+- Prefer `peerDependencies` since they are installed from the application and are not allowed to be in the built bundle. 
+- Prefer exporting `injectionToken` instead of `injectable` for not allowing other features to access technical details like the `injectable`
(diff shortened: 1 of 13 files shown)


packages/technical-features/feature-core/src/registration-of-feature.test.ts
```diff
@@ -0,0 +1,147 @@
...
+      );
+    });
+
+    it("given a Feature deregistered, when deregistered again, throws", () => {
+      deregisterFeature(di, someFeature);
+
+      expect(() => {
+        deregisterFeature(di, someFeature);
+      }).toThrow(
+        'Tried to deregister feature "some-feature-1", but it was not registered.'
```
ixrock, line 104: Is there other more robust ways to check it? I guess if I update the error message from  code this will fail?

packages/technical-features/feature-core/src/register-feature.ts
```diff
@@ -0,0 +1,90 @@
...
+  features.forEach((feature) => {
+    registerFeatureRecursed(di, feature);
+  });
+};
+
+const registerFeatureRecursed = (
+  di: DiContainer,
+  feature: Feature,
+  dependedBy?: Feature
+) => {
```
ixrock, line 19: _nit_: this is more readable imho + there is no actual reason to use arrow-function, `this` context is not even considered).

```suggestion
function registerFeatureRecursed(di: DiContainer, feature: Feature, dependedBy?: Feature) {
```

ixrock (dismissed):

packages/technical-features/feature-core/src/registration-of-feature.test.ts
jansav: Yes, this will fail if you update the message. That's kinda important because we want to make sure that it fails just for this reason only.

ixrock: Yes, I mean could we detach the text and the reason? So changing text should not break the test, but still we should be able to verify the case (e.g. via `id` or some `CustomError`). Wdyt?

jansav: I don't think it's needed here. It serves as easy way to find the meaningful unit test with quick search.

packages/technical-features/feature-core/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+// Or perhaps you want to deregister?
+deregisterFeature(di, someFeature);
+```
+
+## Need to know
+
+#### NPM packages exporting a Feature
+- Prefer `peerDependencies` since they are installed from the application and are not allowed to be in the built bundle. 
+- Prefer exporting `injectionToken` instead of `injectable` for not allowing other features to access technical details like the `injectable`
```
Nokel81, line 41: Can the example above show this then. Not the best if the best practice is not shown in the only example code.

jansav: That's just example which shows how Feature is created. We will have scaffolding for Features which will demonstrate both bullet points. And link for that will be updated here.

jakolehm (approved):
LGTM

PR #7243, opened 2023-02-27
Quick fix for store migration version being wrong

Nokel81:
fixes https://github.com/lensapp/lens-ide/issues/529

Better version is in #4571, which makes the migration version specific to each store.

### DIFF
--- a/packages/core/src/common/vars/store-migration-version.injectable.ts
+++ b/packages/core/src/common/vars/store-migration-version.injectable.ts
@@ -2,12 +2,11 @@
  * Copyright (c) OpenLens Authors. All rights reserved.
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
-import { applicationInformationToken } from "@k8slens/application";
 import { getInjectable } from "@ogre-tools/injectable";
 
 const storeMigrationVersionInjectable = getInjectable({
   id: "store-migration-version",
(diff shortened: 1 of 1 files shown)


packages/core/src/common/vars/store-migration-version.injectable.ts
```diff
@@ -2,12 +2,11 @@
  * Copyright (c) OpenLens Authors. All rights reserved.
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
-import { applicationInformationToken } from "@k8slens/application";
 import { getInjectable } from "@ogre-tools/injectable";
 
 const storeMigrationVersionInjectable = getInjectable({
   id: "store-migration-version",
-  instantiate: (di) => di.inject(applicationInformationToken).version,
+  instantiate: () => "6.4.0",
```
jim-docker, line 9: Could this remain as `di.inject(applicationInformationToken).version` unless it is not our usual versioning value. Can you easily detect when `di.inject(applicationInformationToken).version` is unusual?

jim-docker (approved):

packages/core/src/common/vars/store-migration-version.injectable.ts
jakolehm, line 9: `instantiate: (di) => di.inject(extensionApiVersionInjectable)`

PR #7247, opened 2023-02-28
Fix extension install

### LINKED ISSUE #7240: Extension installation fails
How to reproduce:
1. Follow instructions at https://github.com/nevalla/lens-resource-map-extension#installation to install extension

What should happen: extension is installed.

What actually happens:
* Lens is stuck on "Installing extension" state
* Error logs:

```
npacking extension @nevalla/kube-resource-map@1.1.0 {fileName: 'kube-resource-map-1.1.0.tgz', tempFile: '/var/folders/_j/ztr5t9y90jq_jpsz1s2ytvk40000gp/T/lens-extensions/kube-resource-map-1.1.0.tgz'}
renderer.js:26389 Uncaught Error: [ExtensionInstallationStore]: cannot set /Users/phorsmalahti/.k8slens/extensions/nevalla--kube-resource-map/package.json as installing. Is currently installing.
    at ExtensionInstallationStateStore.value [as setInstalling] (renderer.js:26389:1)
    at EventEmitter.<anonymous> (renderer.js:26369:1)
    at EventEmitter.emit (node:events:526:28)
    at Object.onMessage (node:electron/js2c/renderer_init:69:755)
value @ renderer.js:26389
(anonymous) @ renderer.js:26369
emit @ node:events:526
onMessage @ node:electron/js2c/renderer_init:69
renderer.js:26472 Uncaught Error: [ExtensionInstallationStore]: cannot clear INSTALLING state for /Users/phorsmalahti/.k8slens/extensions/nevalla--kube-resource-map/package.json, it is currently idle
    at ExtensionInstallationStateStore.value [as clearInstalling] (renderer.js:26472:1)
    at EventEmitter.<anonymous> (renderer.js:26372:1)
    at EventEmitter.emit (node:events:526:28)
    at Object.onMessage (node:electron/js2c/renderer_init:69:755)
```

Nokel81:
- Remove old bundled extension dependencies
- Make sure external extensions are installed as optional

fixes https://github.com/lensapp/lens/issues/7240


packages/core/src/extensions/install-extension/install-extension.injectable.ts
```diff
@@ -0,0 +1,92 @@
...
+        reject(error);
+      });
+    });
+
+    const packageJsonPath = joinPaths(extensionPackageRootDirectory, "package.json");
+
+    const fixupPackageJson = once(async () => {
+      const packageJson = await readJsonFile(packageJsonPath) as PackageJson;
+
+      delete packageJson.dependencies;
```
panuhorsmalahti: This needs a comment why we're doing it for future reference

panuhorsmalahti: It seems in current stable 3rd party extensions are `dependencies`, so this would delete these also. Will this cause problems?

```
➜ cat package.json
{
  "dependencies": {
    "@mirantis/lens-extension-cc": "file:../../../.k8slens/extensions/mirantis--lens-extension-cc",
```

panuhorsmalahti: Maybe no, because the installed extension is still in `/Users/<username>/Library/Application Support/Lens/node_modules/` and `/Users/phorsmalahti/Library/Application Support/Lens/lens-extensions.json`. But doesn't sound good anyway.

panuhorsmalahti: For consistency we could leave out `--save-optional` if we really don't need the exts in the package.json file.

aleksfront (commented):
LGTM, fixes bug, extensions got installed successfully. Let's test it more in beta.

packages/core/src/extensions/install-extension/install-extension.injectable.ts
Nokel81: I'll add a comment. I was trying to emulate what was removed in #6715 since that is when this was all changed.

jakolehm: I think we need to persist package.json changes because npm will remove packages that are not there (at least npm behaved like that in the past).

Nokel81: Yes that is correct

jakolehm (approved):
LGTM

PR #7253, opened 2023-03-01
Bump versions to 6.5.0-alpha.0

Nokel81:


aleksfront (approved):

packages/infrastructure/jest/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@k8slens/jest",
   "private": false,
-  "version": "0.0.1",
+  "version": "6.5.0-alpha.0",
```
jim-docker, line 4: updates everything in @k8slens?

jim-docker (approved):

packages/infrastructure/jest/package.json
Nokel81: That is what I put it too, maybe this is wrong... but I think it should be fine.

PR #7262, opened 2023-03-02
White-listing of telemetry params

Iku-turso:
Motivation: make unexpected large params not eat resources.

With this, we can have white-list such as:
```typescript
const whiteList = [
  ...
  // old types of white-list items, but now not resulting in any args as params of telemetry
  "navigate-to-extensions",
  "navigate-to-cluster-view",
  "navigate-to-entity-settings",

  // new type of white-list item, now with selectable args as params of telemetry
  {
    id: "navigate-to-route",
    getParams: (route: Route<unknown>) => ({ param: route.path }),
  },
];  
```

Note: also telemetry by tagging is dropped for not being used.

### DIFF
--- a/packages/core/src/common/log-error.injectable.ts
+++ b/packages/core/src/common/log-error.injectable.ts
@@ -8,6 +8,7 @@ import loggerInjectable from "./logger.injectable";
 const logErrorInjectable = getInjectable({
   id: "log-error",
   instantiate: (di) => di.inject(loggerInjectable).error,
+  decorable: false,
 });
 
 export default logErrorInjectable;
--- a/packages/core/src/common/logger.injectable.ts
+++ b/packages/core/src/common/logger.injectable.ts
@@ -26,6 +26,8 @@ const loggerInjectable = getInjectable({
       silly: (message, ...data) => baseLogger.silly(message, ...data),
     };
   },
+
+  decorable: false,
 });
 
 export default loggerInjectable;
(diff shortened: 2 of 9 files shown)


packages/core/src/common/log-error.injectable.ts
```diff
@@ -8,6 +8,7 @@ import loggerInjectable from "./logger.injectable";
 const logErrorInjectable = getInjectable({
   id: "log-error",
   instantiate: (di) => di.inject(loggerInjectable).error,
+  decorable: false,
```
aleksfront, line 11: Can you describe what `decorable` is?

aleksfront (approved):

aleksfront:
> Motivation: make unexpected large params not eat resources.

Any advice where whitelisting params should be implemented very first?

jweak (approved):

packages/core/src/common/log-error.injectable.ts
Iku-turso: It is a solution to make `decorators`, and their affiliates, not decorate themselves infinitely. We'll remove requirement for `decorable` eventually.

Iku-turso:
> Any advice where whitelisting params should be implemented very first?

@aleksfront I suppose we could follow the white-list in `/lens/packages/core/src/features/telemetry/renderer/telemetry-white-list-for-functions.injectable.ts`, and start making some informed guesses :/

Eg. `enable-extension` and friends could benefit from name of extension. `show-details` could tell what details are being shown.

PR #7276, opened 2023-03-03
Update release guide and fix release script

Nokel81:
- Fix for release script boils down to incrementing the patch version for the milestone that we are looking for when picking PRs to use when making a non-prerelease


Iku-turso (approved):
"Fix release script" -> PR name could be more specific :)

Iku-turso (approved):

PR #7277, opened 2023-03-03
Fix clusters not having namespaces in restricted RBAC scenarios

Nokel81:

### DIFF
--- a/packages/core/src/renderer/components/+namespaces/namespace-storage.injectable.ts
+++ b/packages/core/src/renderer/components/+namespaces/namespace-storage.injectable.ts
@@ -1,0 +1,23 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import clusterFrameContextForNamespacedResourcesInjectable from "../../cluster-frame-context/for-namespaced-resources.injectable";
+import { isDefined } from "../../utils";
+import createStorageInjectable from "../../utils/create-storage/create-storage.injectable";
+
+const selectedNamespaceStorageInjectable = getInjectable({
+  id: "selected-namespace-storage",
+  instantiate: (di) => {
+    const createStorage = di.inject(createStorageInjectable);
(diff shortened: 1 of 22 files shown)


packages/core/src/renderer/components/+namespaces/namespace-storage.injectable.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import allNamespacesInjectable from "../../cluster-frame-context/all-namespaces.injectable";
+import { isDefined } from "../../utils";
+import createStorageInjectable from "../../utils/create-storage/create-storage.injectable";
+
+const selectedNamespaceStorageInjectable = getInjectable({
```
Iku-turso, line 10: Is this a duplicated implementation of what `namespaceStoreInjectable` does? Are there problems with the two co-existing, as they might have different opinions on eg. what namespace is selected?

For instance, they both use different "storages" created with same id: `createStorage("selected_namespaces")`. That could be trouble, right?

`namespaceStore` is also exposed in extension API. I wonder if this is in sync with that.

packages/core/src/renderer/components/+namespaces/store.ts
```diff
@@ -28,23 +28,11 @@ export class NamespaceStore extends KubeObjectStore<Namespace, NamespaceApi> {
...
-      this.selectNamespaces(["default"]);
-    } else if (allowedNamespaces.length) {
-      this.selectNamespaces([allowedNamespaces[0]]);
-    } else {
-      this.selectNamespaces([]);
-    }
   }
 
+  /**
+   * @deprecated
```
Iku-turso, line 34: Deprecations should explain the non-deprecated solution.

packages/core/src/renderer/components/+namespaces/namespace-storage.injectable.ts
Nokel81: This is not a duplicate. It actually moves the impl here instead.

PR #7279, opened 2023-03-03
Fix namespaces not respecting accessible namespace config

### LINKED ISSUE #7270: [Windows]After updating the latest version,i cannot see all the pods
**Describe the bug**
[Windows]After updating the latest version,i cannot see all the pod

**To Reproduce**
1、Go to update Lens in application.

 
**Screenshots**
image

### LINKED ISSUE #7271: [Windows] lens keep forgetting last selected namespace after recent update
**Describe the bug**
After latest update of lens app `2023.3.21022-latest` the cluster browser do not remember the last selected namespace. 
image

**To Reproduce**
1. Update the Lens app to latest `2023.3.21022-latest`.
2. Browse any local cluster.
3. Select any `Workloads` or `Config` and chose a namespace (other than `All namespaces`)
4. Quit the Lens app and repeat step `2` and `3` above. But this time in step `3` last selected namespace should be used instead of `All namespaces` again.

**Expected behavior**
As described in Step 4 of Steps to reproduce above.

- Lens Version: 2023.3.21022-latest
- OS: Windows
- Installation method (e.g. snap or AppImage in Linux): Installer

Nokel81:

minimal version of #7277 

### Description of change

- Due to #6657 `NamespaceStore` was changed to no longer include fake loading and fake instances of the namespaces retrieved from the `refreshAccessbility` logic within `Cluster`. Thus `NamespaceStore.items` is not `[]` if the user does not have `list` permissions for `namespaces` (just like all other resources).
- `NamespaceStore`’s dependencies field has a context field. This is set to `clusterFrameContextForClusterScopedResources` (since it is a cluster scoped resource)
- As such `this.dependencies.context.allNamespaces` is always `[]`, so reverting the implementation of `allowedNamespaces` changed nothing

It is preferable to not have a self-referential updating stateful data source as it has made `NamespaceStore` extremely  hard to reason about.

### DIFF
--- a/packages/core/src/common/cluster/cluster.ts
+++ b/packages/core/src/common/cluster/cluster.ts
@@ -498,6 +498,7 @@ export class Cluster implements ClusterModel {
 
     this.allowedResources.replace(await this.getAllowedResources(requestNamespaceListPermissions));
     this.ready = this.knownResources.length > 0;
+    this.dependencies.logger.info(`[CLUSTER]: refreshed accessibility data`, this.getState());
   }
 
   /**
--- a/packages/core/src/common/cluster/cluster.ts
+++ b/packages/core/src/common/cluster/cluster.ts
@@ -498,7 +498,7 @@ export class Cluster implements ClusterModel {
 
     this.allowedResources.replace(await this.getAllowedResources(requestNamespaceListPermissions));
     this.ready = this.knownResources.length > 0;
-    this.dependencies.logger.info(`[CLUSTER]: refreshed accessibility data`, this.getState());
+    this.dependencies.logger.debug(`[CLUSTER]: refreshed accessibility data`, this.getState());
   }
 
   /**
--- a/packages/core/src/renderer/components/+workloads-pods/load-pods-from-all-namespaces.injectable.ts
+++ b/packages/core/src/renderer/components/+workloads-pods/load-pods-from-all-namespaces.injectable.ts
@@ -4,7 +4,7 @@
  */
 
 import { getInjectable } from "@ogre-tools/injectable";
-import namespaceStoreInjectable from "../+namespaces/store.injectable";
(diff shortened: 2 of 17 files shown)


packages/core/src/renderer/components/+workloads-pods/load-pods-from-all-namespaces.injectable.ts
```diff
@@ -4,20 +4,20 @@
...
     const podStore = di.inject(podStoreInjectable);
-    const namespaceStore = di.inject(namespaceStoreInjectable);
+    const context = di.inject(clusterFrameContextForNamespacedResourcesInjectable);
     const showErrorNotification = di.inject(showErrorNotificationInjectable);
-  
+
     return () => {
       podStore.loadAll({
-        namespaces: [...namespaceStore.getItems().map(ns => ns.getName())],
+        namespaces: context.allNamespaces,
```
jakolehm, line 20: Why do we need to pass `namespaces` here, shouldn't it default to all namespaces if none given?

packages/core/src/common/cluster/cluster.ts
```diff
@@ -498,6 +498,7 @@ export class Cluster implements ClusterModel {
 
     this.allowedResources.replace(await this.getAllowedResources(requestNamespaceListPermissions));
     this.ready = this.knownResources.length > 0;
+    this.dependencies.logger.info(`[CLUSTER]: refreshed accessibility data`, this.getState());
```
jakolehm: Maybe `debug`?

jakolehm (approved):

jim-docker:
After random Cmd-click to select various namespaces, the namespaces are no longer sorted alphabetically. Possibly why some tests failed.

Screen Shot 2023-03-06 at 4 18 18 PM

PR #7280, opened 2023-03-03
Add behavioural tests for Cluster Menu K8s Resources in Sidebar menu not being shown

Nokel81:
This PR fixes an oversight in #6657. Namely, I previously assumed that all the applicable resources rules would be merged together on the server side before sending a response back for a given `SelfSubjectRulesReview`. That, however, is not the case and instead it seems that most kubernetes distros (or at least K0s) just find all the applicable rules and then send them all back in a list.

To reproduce this yourself:

1. Create an LDK
2. Share it with a space
3. Pause the updater and remove the rules targeting lens space Members and Admins
4. Add two new `ClusterRoles`'s, one called `namespace-get` which allows the `get` verb on `namespaces` and one called `namespace-list` which allows the `list` verb on `namespaces`
5. Add two new `ClusterRoleBindings`, one for each of the above, targeting your own user.
6. Connect to the shared version of the LDK and run the following command in the terminal.
    - `curl -H "Content-Type: application/json" -X POST 127.0.0.1:8001/apis/authorization.k8s.io/v1/selfsubjectrulesreviews --data '{ "apiVersion": "authorization.k8s.io/v1", "kind": "SelfSubjectRulesReview",  "spec": { "namespace": "default" }}'`
7. See something like the following:

```yml
status:
  resourceRules:
    - apiGroups:
        - ''
      resourceNames: null
      resources:
        - 'namespaces'
      verbs:
        - 'list'
    - apiGroups:
        - 'policy'
      resourceNames:
        - '00-k0s-privileged'
      resources:
        - 'podsecuritypolicies'
      verbs:
         - 'use'
    - apiGroups:
        - 'authorization.k8s.io'
      resourceNames: null
      resources:
        - 'selfsubjectaccessreviews'
        - 'selfsubjectrulesreviews'
      verbs:
        - 'create'
    - apiGroups:
        - ''
      resourceNames: null
      resources:
        - 'namespaces'
      verbs:
        - 'get'
```


packages/core/src/common/cluster/request-namespace-list-permissions.injectable.ts
```diff
@@ -47,23 +47,17 @@ const requestNamespaceListPermissionsForInjectable = getInjectable({
           const { resourceRules } = status;
 
           return (resource) => {
-            const resourceRule = resourceRules.find(({
-              apiGroups = [],
-              resources = [],
+            const rules = resourceRules.filter(({
```
jakolehm: Did I get this right?
- defaults are now `["*"]`
- instead of checking verbs from the first match we check them from all matches

Nokel81: Yes that is correct.

I think that the section·second point is more important but I haven't been able to test this.

jakolehm (commented):
IMHO this PR touches too many places for a hotfix (patch release).

packages/core/src/renderer/components/test-utils/get-application-builder.tsx
```diff
@@ -159,7 +159,18 @@ interface Environment {
...
+   * @default true
+   */
+  fakeTime?: boolean | FakeTimersConfig;
+}
+
+export const getApplicationBuilder = (options: ApplicationBuilderOptions = {}) => {
+  const {
+    fakeTime = true,
+  } = options;
+
```
jansav: I don't think that there's case where we want to support behavioural unit tests without fake time.

packages/core/src/common/cluster/create-can-i.injectable.ts
```diff
@@ -0,0 +1,42 @@
...
+          kind: "SelfSubjectAccessReview",
+          spec: { resourceAttributes },
+        });
+
+        return body.status?.allowed ?? false;
+      } catch (error) {
+        logger.error(`[AUTHORIZATION-REVIEW]: failed to create access review: ${error}`, { resourceAttributes });
+
+        return false;
+      }
```
jim-docker, line 37: If there's an error do we know for sure the answer is `false`? Seems strange that the caller doesn't know there was an error.

packages/core/src/common/cluster/create-request-namespace-list-permissions.injectable.ts
```diff
@@ -0,0 +1,57 @@
...
+        const { body: { status }} = await api.createSelfSubjectRulesReview({
+          apiVersion: "authorization.k8s.io/v1",
+          kind: "SelfSubjectRulesReview",
+          spec: { namespace },
+        });
+
+        if (!status || status.incomplete) {
+          logger.warn(`[AUTHORIZATION-NAMESPACE-REVIEW]: allowing all resources in namespace="${namespace}" due to incomplete SelfSubjectRulesReview: ${status?.evaluationError}`);
+
+          return () => true;
```
jim-docker, line 37: The caller has no idea here, or in the catch below, that `api.createSelfSubjectRulesReview()` failed

jim-docker (approved):
This is all refactoring? (I don't see any tests added)

I guess my comments are for the previous review, and not really an issue if "CreateCanI" is called more than once, so "CanI" can change, if some auth/api error gets corrected.

Nokel81:
packages/core/src/features/cluster/refresh-accessibility-technical.test.ts was added

The diff is large so it is hidden by default.

packages/core/src/common/cluster/create-can-i.injectable.ts
Nokel81: No we don't, but since these are currently only used for optimizations it is fine if they are turned off in case of error.

Plus this has always been the case.

packages/core/src/common/cluster/create-request-namespace-list-permissions.injectable.ts
Nokel81: It is safer to show all the resources of this fails then to show none.

jim-docker: Sure, I'm really questioning whether that decision should be made here or by the caller, which might be able to show a helpful notification to the end-user. Now there is no indication to the user why there are suddenly all these resources listed in the sidebar that perhaps can't be listed.

PR #7282, opened 2023-03-06
Fix request namespace list permissions check if a list rule is not the first match

jakolehm:
Smaller version of #7280 .

### DIFF
--- a/packages/core/src/common/cluster/request-namespace-list-permissions.test.ts
+++ b/packages/core/src/common/cluster/request-namespace-list-permissions.test.ts
@@ -1,0 +1,246 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import type { V1SubjectRulesReviewStatus } from "@kubernetes/client-node";
+import type { DiContainer } from "@ogre-tools/injectable";
+import { getDiForUnitTesting } from "../../main/getDiForUnitTesting";
+import type { RequestNamespaceListPermissionsFor } from "./request-namespace-list-permissions.injectable";
+import requestNamespaceListPermissionsForInjectable from "./request-namespace-list-permissions.injectable";
+
+const createFakeProxyConfig = (statusResponse: Promise<V1SubjectRulesReviewStatus>) => ({
+  makeApiClient: () => ({
+    createSelfSubjectRulesReview: (): Promise<V1SubjectRulesReviewStatus> => statusResponse,
+  }),
+});
+
+describe("requestNamespaceListPermissions", () => {
+  let di: DiContainer;
+  let requestNamespaceListPermissions: RequestNamespaceListPermissionsFor;
+
+  beforeEach(() => {
+    di = getDiForUnitTesting({ doGeneralOverrides: true });
+    requestNamespaceListPermissions = di.inject(requestNamespaceListPermissionsForInjectable);
+  });
+
+  describe("when api returns incomplete data", () => {
+    it("returns truthy function", async () => {
+      const requestPermissions = requestNamespaceListPermissions(createFakeProxyConfig(
+        new Promise((resolve) => resolve({
+          incomplete: true,
+          resourceRules: [],
+          nonResourceRules: [],
+        })),
+      ) as any);
+
+      const permissionCheck = await requestPermissions("fake-namespace");
+
+      expect(permissionCheck({
+        apiName: "pods",
+        group: "",
+        kind: "Pod",
+        namespaced: true,
+      })).toBeTruthy();
+    });
+  });
+
+  describe("when api rejects", () => {
+    it("returns truthy function", async () => {
+      const requestPermissions = requestNamespaceListPermissions(createFakeProxyConfig(
+        new Promise((resolve, reject) => reject("unknown error")),
+      ) as any);
+
+      const permissionCheck = await requestPermissions("fake-namespace");
+
+      expect(permissionCheck({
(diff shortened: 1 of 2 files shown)


packages/core/src/common/cluster/request-namespace-list-permissions.test.ts
```diff
@@ -0,0 +1,336 @@
...
+      const requestPermissions = requestNamespaceListPermissions(createFakeProxyConfig(
+        new Promise((resolve) => resolve({
+          body: {
+            status: {
+              incomplete: true,
+              resourceRules: [],
+              nonResourceRules: [],
+            },
+          },
+        })),
```
Iku-turso, line 38: Using `asyncFn` for stuff like this would remove a bit of duplication :)

jakolehm: Maybe give an example and I'll refactor? :)

packages/core/src/common/cluster/request-namespace-list-permissions.test.ts
```diff
@@ -0,0 +1,336 @@
...
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import type { V1SubjectRulesReviewStatus } from "@kubernetes/client-node";
+import type { DiContainer } from "@ogre-tools/injectable";
+import { getDiForUnitTesting } from "../../main/getDiForUnitTesting";
+import type { RequestNamespaceListPermissionsFor } from "./request-namespace-list-permissions.injectable";
+import requestNamespaceListPermissionsForInjectable from "./request-namespace-list-permissions.injectable";
+
+const createFakeProxyConfig = (statusResponse: Promise<{ body: { status: V1SubjectRulesReviewStatus }}>) => ({
```
Iku-turso: To consolidate communication in future, this is a `stub` instead of `fake` :)

A `fake` is a *trusted simulation* of something with non-trivial implementation. A `stub` is a trivial substitution of value.

jakolehm: Changed.

packages/core/src/common/cluster/request-namespace-list-permissions.test.ts
Iku-turso: An example on adding `asyncFn` to existing code structure. Note that the different parts of code here can be spread in different `beforeEach` to remove said duplication.

```typescript
const createSelfSubjectRulesReviewMock = asyncFn();

const proxyConfigStub = createProxyConfigStub(createSelfSubjectRulesReviewMock);

const requestPermissions = requestNamespaceListPermissions(proxyConfigStub);

const permissionCheckPromise = requestPermissions("some-namespace");

// Note: in different test, this could be `.reject(...)`.
createSelfSubjectRulesReviewMock.resolve({
  body: {
    status: {
      incomplete: true,
      resourceRules: [],
      nonResourceRules: [],
    },
  },
});

const permissionCheck = await permissionCheckPromise;

expect(permissionCheck({
  apiName: "pods",
  group: "",
  kind: "Pod",
  namespaced: true,
})).toBeTruthy();
```

packages/core/src/common/cluster/request-namespace-list-permissions.test.ts
```diff
@@ -0,0 +1,336 @@
...
+            status: {
+              incomplete: true,
+              resourceRules: [],
+              nonResourceRules: [],
+            },
+          },
+        })),
+      ) as any);
+
+      const permissionCheck = await requestPermissions("fake-namespace");
```
Iku-turso: A string value that does not affect test passing or failing should be `"irrelevant"` or `"irrelevant-namespace"`.

packages/core/src/common/cluster/request-namespace-list-permissions.test.ts
```diff
@@ -0,0 +1,336 @@
...
+  let requestNamespaceListPermissions: RequestNamespaceListPermissionsFor;
+
+  beforeEach(() => {
+    di = getDiForUnitTesting({ doGeneralOverrides: true });
+    requestNamespaceListPermissions = di.inject(requestNamespaceListPermissionsForInjectable);
+  });
+
+  describe("when api returns incomplete data", () => {
+    it("returns truthy function", async () => {
+      const requestPermissions = requestNamespaceListPermissions(createFakeProxyConfig(
```
Iku-turso: If injection of dependencies were done using `injectables` instead of configuring curried functions manually, we could get rid of many steps of test initialisation here, namely everything but override of `createSelfSubjectRulesReviewInjectable`, (for which that would `asyncFn` fit nicely).

packages/core/src/common/cluster/request-namespace-list-permissions.test.ts
jakolehm: Changed.

packages/core/src/common/cluster/request-namespace-list-permissions.test.ts
Iku-turso: Example:
```typescript
const createSelfSubjectRulesReviewMock = asyncFn();

di.override(createSelfSubjectRulesReviewInjectable, () => createSelfSubjectRulesReviewMock);

const requestNamespaceListPermissions = di.inject(requestNamespaceListPermissionsInjectable);
```

jakolehm: I agree that the implementation could be simplified a lot but I'm not comfortable doing that in this PR.

msa0311 (approved):

PR #7283, opened 2023-03-06
Configure dependabot to update all packages in monorepo

aleksfront:
Currently it checks only root directory to dependencies. This PR configures to respect other packages as well.

As described in https://github.com/dependabot/dependabot-core/issues/4993#issuecomment-1289133027


aleksfront:
Just figured out that dependabot.yml should be located in `.github` folder, and there is a configuration. Will fine-tune config from there.

PR #7245, opened 2023-02-28
Revert "Renderer file logging transport (#6795)"

samitiilikainen:
Renderer file logging still caused UI freezing (at least on apple silicon macs) when cluster frame was open and main frame was reloaded.

See https://github.com/lensapp/lens-ide/issues/544

This reverts commit ac2d0e46ff9db63bbebb58b47a9a6455fc5ced84.

PR #7300, opened 2023-03-07
Renderer file logging through IPC

samitiilikainen:
Implementation similar to #6795, produces the same log files.

To avoid using file system directly in renderer, now using IPC transport to send log messages from renderer to main. Main then writes messages to files using the file transport of Winston, ensuring there is at most one file handle open for each log file. 

Unfortunately also the same UI reload/quit issues apply here as in #6795 (reverted in #7245).
1. UI freezes if there is any `pagehide` listener in cluster frame - at least in certain situations with multiple clusters open. 
2. This forces to use `beforeunload` as that doesn't freeze the UI, but also doesn't often get called in all cluster frames when main frame is unmounted.
3. Cluster log files are left open when `beforeunload` is not called

To fix issues leaving files open:
1. This implementation creates only one file handle per cluster id in main, so when UI is reloaded, the same unclosed files will be reused.
2. When only the UI is quit (leaving the main process running), there is now `stopIpcLoggingInjectable` with `beforeQuitOfFrontEndInjectionToken` to ensure all renderer log files are closed.


packages/core/src/main/logger/close-ipc-logging-listener.injectable.ts
```diff
@@ -0,0 +1,20 @@
...
+  closeIpcFileLoggerChannel,
+} from "../../common/logger/ipc-file-logger-channel";
+
+const closeIpcFileLoggingListenerInjectable = getMessageChannelListenerInjectable({
+  id: "close-ipc-file-logging",
+  channel: closeIpcFileLoggerChannel,
+  handler: (di) => (fileId) =>
+    di
+      .inject(ipcFileLoggerInjectable)
+      .close(fileId),
```
Nokel81: We should inject `ipcFileLogger` once and then use it afterwards

```suggestion
  handler: (di) => {
    const ipcFileLogger = di.in.inject(ipcFileLoggerInjectable);
    
    return (fileId) => ipcFileLogger.close(fileId);
  },
```

packages/core/src/main/logger/ipc-file-logger.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    const fileTransport = this.createNewFileTransport({
+      ...this.options,
+      filename: `lens-${fileId}.log`,
+    });
+
+    this.fileTransports.set(fileId, fileTransport);
+
+    return fileTransport;
+  }
```
Nokel81: This can be simplified by using the `getOrInsertWith` helper from `@k8slens/utilities`

packages/core/src/main/logger/ipc-file-logger.ts
```diff
@@ -0,0 +1,56 @@
...
+    if (transport) {
+      transport.close?.();
+      this.fileTransports.delete(fileId);
+    }
+  }
+
+  closeAll() {
+    [...this.fileTransports.keys()].forEach((fileId) => {
+      this.close(fileId);
+    });
```
Nokel81: nit: maybe use a `for...each`

packages/core/src/main/logger/ipc-file-logger.ts
```diff
@@ -0,0 +1,56 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import type { LogEntry, transports } from "winston";
+
+type IpcFileLoggerOptions = Omit<transports.FileTransportOptions, "filename">;
+
+class IpcFileLogger {
```
Nokel81: Don't quite see why this needs to be a class, could just be created within the `ipc-file-logger.injectable.ts` file. Especially since there will never bee a need to `extend` this class.

packages/core/src/main/logger/ipc-file-logger.ts
```diff
@@ -0,0 +1,56 @@
...
+    const transport = this.ensureTransportForFile(fileId);
+
+    transport?.log?.(entry, () => {});
+  }
+
+  close(fileId: string) {
+    const transport = this.fileTransports.get(fileId);
+
+    if (transport) {
+      transport.close?.();
```
Nokel81: Is the typing of `transport.close` really `undefined | () => void`?

packages/core/src/main/logger/ipc-logging-listener.injectable.ts
```diff
@@ -0,0 +1,40 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import ipcFileLoggerInjectable from "./ipc-file-logger.injectable";
+import { getMessageChannelListenerInjectable } from "../../common/utils/channel/message-channel-listener-injection-token";
+import type {
+  IpcFileLogObject } from "../../common/logger/ipc-file-logger-channel";
```
Nokel81: nit: this formatting looks really weird.

(If it is a prettier issue, disregard)

packages/core/src/renderer/frames/root-frame/init-root-frame.injectable.ts
```diff
@@ -60,6 +62,7 @@ const initRootFrameInjectable = getInjectable({
       window.addEventListener("beforeunload", () => {
         logger.info("[ROOT-FRAME]: Unload app");
 
+        closeRendererLogFile();
```
Nokel81: Maybe it would be better to factor the registering of this event handler out of these "god-functions" right now. Instead of adding more to these legacy form.

packages/core/src/renderer/logger/close-renderer-log-file-id.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  it("sends the ipc close message with correct log id", () => {
+    const closeLog = di.inject(closeRendererLogFileInjectable);
+
+    closeLog();
+
+    expect(sendIpcMock).toHaveBeenCalledWith(
+      { id: "close-ipc-file-logger-channel" },
+      "some-log-id",
+    );
```
Nokel81, line 44: I think that sending Ipc would be the wrong level of a test.

Suggestion: Write this out as a behavourial test suite (this goes along with the suggestion of factoring out the registering of the "beforeunload" event handler) so that once a window is registered you can emit the "beforeunload" event and then verify that the function on `main` side is called. That way the communication specifics can change without the test failing.

packages/core/src/renderer/logger/renderer-log-file-id.test.ts
```diff
@@ -0,0 +1,34 @@
...
+    di.override(currentlyInClusterFrameInjectable, () => true);
+    di.override(windowLocationInjectable, () => ({
+      host: "some-cluster.lens.app",
+      port: "irrelevant",
+    }));
+    const clusterFileId = di.inject(rendererLogFileIdInjectable);
+
+    expect(clusterFileId).toBe("renderer-cluster-some-cluster");
+  });
+});
```
Nokel81, line 34: This could be replaced by some behavioural tests that verify that a file is written to the mock file system with the correct name when a logging entry is used.

packages/core/src/renderer/logger/renderer-log-file-id.injectable.ts
```diff
@@ -0,0 +1,29 @@
...
+  id: "renderer-log-file-id",
+  instantiate: (di) => {
+    let frameId: string;
+    const currentlyInClusterFrame = di.inject(currentlyInClusterFrameInjectable);
+
+    if (currentlyInClusterFrame) {
+      const { host } = di.inject(windowLocationInjectable);
+      const clusterId = getClusterIdFromHost(host);
+
+      frameId = clusterId ? `cluster-${clusterId}` : "cluster";
```
Nokel81: Given that a `clusterId` should always be present, maybe just trying to compute the `getClusterIdFromHost` is enough to verify that we are currently in the cluster frame

Nokel81 (commented):
Also a general comment:

I recently noticed #7321, do you think that might be easy to fix?

packages/core/src/main/logger/close-ipc-logging-listener.injectable.ts
samitiilikainen: Thanks!

packages/core/src/main/logger/ipc-file-logger.ts
samitiilikainen: Nice!

packages/core/src/main/logger/ipc-file-logger.ts
samitiilikainen: Yes unfortunately it is

packages/core/src/main/logger/ipc-logging-listener.injectable.ts
samitiilikainen: That's a result of running `lint:fix` to please the PR checks, agreed that it does not look ok. Prettier seems to make it nice though.

I've also recently had some issues with VS Code not showing linting issues as the PR github actions.

packages/core/src/renderer/logger/renderer-log-file-id.injectable.ts
samitiilikainen: I removed the ternary condition as it was quite unnecessary. Not sure if you meant `currentlyInClusterFrame` should be replaced with `getClusterIdFromHost`. To me it seems more correct to use `currentlyInClusterFrame` which checks the process frame, getting part of the url host with `getClusterIdFromHost` might accidentally return something for the main frame too.

packages/core/src/main/logger/file-transport.injectable.ts
```diff
@@ -8,7 +8,7 @@ import directoryForLogsInjectable from "../../common/app-paths/directory-for-log
 import { loggerTransportInjectionToken } from "../../common/logger/transports";
 
 const fileLoggerTranportInjectable = getInjectable({
```
Nokel81: nit: can you also fix the injectable name

samitiilikainen: Thanks, that typo seems to be hard to spot

Nokel81 (approved):
Would be great if we could get some behaviour tests too, but these should be fine for the time being

PR #7304, opened 2023-03-07
Navigate to catalog explicitly on cluster disconnect

### LINKED ISSUE #6800: Cluster Disconnect doesn't always navigate back to catalog
For some reason, once in a while, `Disconnect` leads to infinite `Connection` page instead of catalog.

https://user-images.githubusercontent.com/9607060/208675183-7f407435-bfe7-4947-822c-be294bfd1652.mov

aleksfront:
Fixes #6800 

There is a race condition which prevents navigating to catalog after cluster disconnect.

### The race condition
For some reason `when` autorun in `cluster-frame-handler.ts` triggers before `reaction` in `cluster-view.tsx` **if user goes to `Settings` page before cluster disconnect**.

Cluster frame handler https://github.com/lensapp/lens/blob/master/packages/core/src/renderer/components/cluster-manager/cluster-frame-handler.ts#L78
Cluster view https://github.com/lensapp/lens/blob/master/packages/core/src/renderer/components/cluster-manager/cluster-view.tsx#L84

### Solution
Navigating to catalog explicitly from kubernetes cluster menu.

https://user-images.githubusercontent.com/9607060/223437969-78306eea-95b9-44a2-9679-9454875cfa1a.mov

### DIFF
--- a/packages/core/src/common/catalog-entities/kubernetes-cluster.ts
+++ b/packages/core/src/common/catalog-entities/kubernetes-cluster.ts
@@ -127,7 +127,13 @@ export class KubernetesCluster<
         context.menuItems.push({
           title: "Disconnect",
           icon: "link_off",
-          onClick: () => requestClusterDisconnection(this.getId()),
+          onClick: () => {
+            requestClusterDisconnection(this.getId());
+            broadcastMessage(
+              IpcRendererNavigationEvents.NAVIGATE_IN_APP,
+              "/catalog",
+            )
+          },
         });
         break;
       case LensKubernetesClusterStatus.DISCONNECTED:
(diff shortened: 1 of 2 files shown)


packages/core/src/common/catalog-entities/kubernetes-cluster.ts
```diff
@@ -127,7 +127,13 @@ export class KubernetesCluster<
         context.menuItems.push({
           title: "Disconnect",
           icon: "link_off",
-          onClick: () => requestClusterDisconnection(this.getId()),
+          onClick: () => {
+            requestClusterDisconnection(this.getId());
+            broadcastMessage(
+              IpcRendererNavigationEvents.NAVIGATE_IN_APP,
+              "/catalog",
+            );
```
Nokel81, line 135: Thinking aloud, could we instead say "NAVIGATE_BACK_IN_APP"? Or is that not currently supported?

If not, the LGTM

Nokel81 (approved):
LGTM!

We change fine tune the behaviour in the future

PR #7316, opened 2023-03-08
Copy detected metadata from cluster to catalog cluster

jweak:
- https://github.com/lensapp/lens/pull/7255

Fixed the metadata detectors, but not all of the metadata from `Cluster` was copied to `KubernetesCluster`. The KubernetesCluster is the instance extension has access.

I also refactored a bit the copying part and added unit tests to the existing implementation.

### DIFF
--- a/packages/core/src/main/cluster/update-entity-metadata.injectable.ts
+++ b/packages/core/src/main/cluster/update-entity-metadata.injectable.ts
@@ -1,0 +1,42 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import type { KubernetesCluster } from "../../common/catalog-entities";
+import { ClusterMetadataKey } from "../../common/cluster-types";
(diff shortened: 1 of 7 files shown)


packages/core/src/main/cluster/update-entity-metadata.injectable.ts
```diff
@@ -0,0 +1,42 @@
...
+      if (cluster.preferences?.clusterName) {
+        /**
+         * Only set the name if the it is overriden in preferences. If it isn't
+         * set then the name of the entity has been explicitly set by its source
+         */
+        entity.metadata.name = cluster.preferences.clusterName;
+      }
+    };
+  },
+});
```
Nokel81, line 40: nit: this function has no side effects and has no dependencies. It doesn't have to be an injectable.

Nokel81 (approved):

packages/core/src/main/cluster/update-entity-metadata.injectable.ts
jweak: Yeah, it's kind of a hammer for me now. See a problem and use injectable. 

I'd want to make these functions pure future as well so that they don't mutate `entity` but return the combined metadata .

PR #7331, opened 2023-03-09
Better displaying container's info for Pods

ixrock:
**Before**

Screenshot 2023-03-09 at 17 43 32

image

**After**

Screenshot 2023-03-09 at 18 15 13

Screenshot 2023-03-09 at 18 48 25

### DIFF
--- a/packages/core/src/common/k8s-api/endpoints/pod.api.ts
+++ b/packages/core/src/common/k8s-api/endpoints/pod.api.ts
@@ -80,6 +80,8 @@ export interface ContainerState {
   terminated?: ContainerStateTerminated;
 }
 
+export type ContainerStateValues = Partial<ContainerState[keyof ContainerState]>;
+
 export interface PodContainerStatus {
   name: string;
   state?: ContainerState;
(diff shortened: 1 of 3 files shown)


jim-docker (dismissed):
It's a bit unfortunate that for a terminated container the container id is still wrapped, last few characters don't make it on the first line.

ixrock:
> It's a bit unfortunate that for a terminated container the container id is still wrapped, last few characters don't make it on the first line.

Now should be fine:

Screenshot 2023-03-09 at 23 36 44

Nodes:

Screenshot 2023-03-09 at 23 42 00

Nokel81 (dismissed):

jim-docker:
I didn't mean to make more work but this is on linux:

Screen Shot 2023-03-09 at 7 08 07 PM

You probably can't make it perfect.

aleksfront (dismissed):

ixrock:
> I didn't mean to make more work but this is on linux:
> You probably can't make it perfect.

PTAL. now even alignment for the first column should be always to the right.

Screenshot 2023-03-10 at 14 13 16
Screenshot 2023-03-10 at 14 13 27

Nokel81 (approved):

Issue #3846, opened 2021-09-21
Allow overriding icon background color


rittneje:
**What would you like to be added**:

I know I can upload a custom hotbar icon image, but it would be nice for simple use cases to just be able to set the color of the icon used in the catalog and hotbar. Right now it just picks some color derived from the full name.

(Customizing the letters instead of just picking the first letter of the first three words would also be great, as requested in #3498.)

**Why is this needed**:

This would just be a nice quality of life improvement, so developers don't have to go through the effort of making a custom image.

**Environment you are Lens application on:**

- Kubernetes distribution: EKS
- Desktop OS: macos

Nokel81:
Partial duplicate of https://github.com/lensapp/lens/issues/3498 so could you change this issue to just be about the icon colour?

rittneje:
Closing because this project is dead.

PR #7341, opened 2023-03-13
Adding cluster settings icon menu items using injection token

aleksfront:
Using `clusterIconSettingsMenuInjectionToken` for adding Cluster Setting menu items.

Part of #3846 

https://user-images.githubusercontent.com/9607060/224667593-c90fd39a-28c5-4ae9-919e-cc669ef094ef.mov


packages/core/src/extensions/renderer-api/tokens.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+export * from "../../renderer/components/cluster-settings/cluster-settings-menu-injection-token";
```
Nokel81: The idea is not to export these to the legacy extension API, but instead have a new package which just contain these tokens.

That way a new feature extension would just import the token package (declaring it as a `peerDependency`)

aleksfront: Removed legacy exports.

packages/core/src/renderer/components/cluster-settings/__tests__/icon-settings.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+    it("has menu item from build-in registration", async () => {
+      userEvent.click(await screen.findByTestId("icon-for-menu-actions-for-cluster-icon-settings-for-some-entity-id"));
+
+      expect(rendered.getByText("Clear")).toBeInTheDocument();
+    });
+  });
+
+  describe("given external registrations for cluster settings menu injection token", () => {
+    beforeEach(() => {
+      runInAction(() => {
```
ixrock, line 90: Why this needs to be wrapped to run in action? Doesn't sound right at all..
There are no observables inside callback, if we still need to do some optimisations those are should be _within_ `di.register(injectable)` imho.

packages/core/src/renderer/components/cluster-settings/cluster-settings-menu-injection-token.ts
```diff
@@ -0,0 +1,19 @@
...
+import { getInjectionToken } from "@ogre-tools/injectable";
+import type { ClusterPreferences } from "../../../common/cluster-types";
+
+export type ChangedClusterPreference = [keyof ClusterPreferences, any];
+
+export interface ClusterIconMenuItem {
+  id: string;
+  title: string;
+  disabled?: (preferences: ClusterPreferences) => boolean;
+  onClick: (preferences: ClusterPreferences) => ChangedClusterPreference;
```
ixrock: Why do we need to return anything on click? Whos gonna use this data?

packages/core/src/renderer/components/cluster-settings/cluster-settings-menu-injection-token.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectionToken } from "@ogre-tools/injectable";
+import type { ClusterPreferences } from "../../../common/cluster-types";
+
+export type ChangedClusterPreference = [keyof ClusterPreferences, any];
```
ixrock:
```suggestion
type ClusterPreferenceKeys = keyof ClusterPreferences;
export type ChangedClusterPreference = [ClusterPreferenceKeys, ClusterPreferences[ClusterPreferenceKeys]];
```

That way we will have at least some validation for values.

packages/core/src/renderer/components/cluster-settings/icon-settings.tsx
```diff
@@ -3,106 +3,113 @@
...
+        <MenuActions
+          id={menuId}
+          data-testid={menuId}
+          toolbar={false}
+          autoCloseOnSelect={true}
+          triggerIcon={{ material: "more_horiz" }}
+        >
+          <MenuItem onClick={onUploadClick}>
+            Upload Icon
+          </MenuItem>
```
ixrock, line 85: Shouldn't this also be separated injectable like you did for "Clear"?

packages/core/src/renderer/components/cluster-settings/cluster-settings-menu-clear-item.injectable.ts
```diff
@@ -0,0 +1,31 @@
...
+import type { ChangedClusterPreference } from "./cluster-settings-menu-injection-token";
+import { clusterIconSettingsMenuInjectionToken } from "./cluster-settings-menu-injection-token";
+
+const clusterIconSettingsMenuClearItem = getInjectable({
+  id: "cluster-icon-settings-menu-clear-item",
+
+  instantiate: () => ({
+    id: "clear-icon-menu-item",
+    title: "Clear",
+    disabled: (preferences: ClusterPreferences) => !preferences.icon,
```
ixrock: Why only `ClusterPreferences` are passed as arguments? What if to determine state we need some more from the `Cluster`?
Also if you're feel free to pass `preferences` to `disabled()` why not pass same way to `onClick()` and did the update of preferences directly inside `onClick(evt, preferences)`?

ixrock (dismissed):

packages/core/src/renderer/components/cluster-settings/__tests__/icon-settings.test.tsx
aleksfront: If not wrapped inside `runInAction` we got an error from the `@ogre-tools/injectable`

ixrock: @jansav @Iku-turso ^ wdyt?

packages/core/src/renderer/components/cluster-settings/cluster-settings-menu-injection-token.ts
aleksfront: By design, any menu item injectable can do something with preferences object and return changed ones. As we did for `Clear` menu item. It returns part of the preferences which is saved in `icon-settings.tsx`.

packages/core/src/renderer/components/cluster-settings/icon-settings.tsx
aleksfront: Yes, it should. But not doing this because of following tricky part:

```
const onUploadClick = () => {
    element
      .current
      ?.querySelector<HTMLInputElement>("input[type=file]")
      ?.click();
  };
```

aleksfront: Do not want to pollute menu items from injectable with a bunch of props such as element reference.

packages/core/src/renderer/components/cluster-settings/cluster-settings-menu-clear-item.injectable.ts
aleksfront: > Why only ClusterPreferences are passed as arguments? What if to determine state we need some more from the Cluster?

Still under consideration what should be passes to injectables. This may vary from specific needs of this part of an application.

> and did the update of preferences directly inside onClick(evt, preferences)?

Probably good idea, refactoring...

packages/core/src/renderer/components/cluster-settings/cluster-settings-menu-injection-token.ts
aleksfront: Will check this out.

packages/core/src/renderer/components/cluster-settings/cluster-settings-menu-clear-item.injectable.ts
Nokel81: Maybe passing `clusterId` would be good enough? That way we have the specifics of where this is stored not be part of the contract here?

aleksfront: Ended up passing preferences to `onClick` event and changing them right there inside injectable.

ixrock (approved):

PR #7351, opened 2023-03-14
Fine-tune light theme hotbar colors

aleksfront:
* Less vibrant backgrounds
* More contrast hover shadow

Before 

hotbar before

After

hotbar after


ixrock (approved):

packages/core/src/renderer/themes/lens-light.injectable.ts
```diff
@@ -94,9 +94,10 @@ const lensLightThemeInjectable = getInjectable({
       drawerSubtitleBackground: "#f1f1f1",
       drawerItemNameColor: "#727272",
       drawerItemValueColor: "#555555",
-      clusterMenuBackground: "#d7d8da",
+      clusterMenuBackground: "#e2e2e2",
       clusterMenuBorderColor: "#c9cfd3",
-      clusterMenuCellBackground: "#bbbbbb",
+      clusterMenuCellBackground: "#d2d2d2",
+      clusterMenuCellOutline: "#22222266",
```
ixrock, line 100: Btw, should we declare this new css-variable somewhere? 
Or how otherwise a developer will find out that? Also I notices that we don't have any pre-generated css-file with all available variables which will improve heavily developer experience and reusability of these vars. Wdyt?

PR #7354, opened 2023-03-14
Move Cluster.apiUrl to seperate injectable

Nokel81:
This is part of breaking down the `Cluster` monolith that was started in #6951.

This piece of information is only used on `main` and can be easily factored out. So we should


gabriel-mirantis (approved):

packages/core/src/main/cluster/kube-auth-proxy-server.injectable.ts
```diff
@@ -24,23 +25,23 @@ const thirtySecondsInMs = 30 * 1000;
...
   instantiate: (di, cluster): KubeAuthProxyServer => {
-    const clusterUrl = new URL(cluster.apiUrl.get());
-
-    const createKubeAuthProxy = di.inject(createKubeAuthProxyInjectable);
-    const certificate = di.inject(kubeAuthProxyCertificateInjectable, clusterUrl.hostname);
+    const clusterApiUrl = di.inject(clusterApiUrlInjectable, cluster);
+    const createKubeAuthProxy = di.inject(createKubeAuthProxyInjectable, cluster);
 
     let kubeAuthProxy: KubeAuthProxy | undefined = undefined;
     let apiTarget: ServerOptions | undefined = undefined;
```
jansav, line 32: After this change, we use validated kubeconfig. Previously we didn't. (?)

Nokel81: which change in particular?

jansav: `const clusterApiUrl = di.inject(clusterApiUrlInjectable, cluster);` vs `const clusterUrl = new URL(cluster.apiUrl.get());`

PR #7358, opened 2023-03-14
Release 6.5.0-alpha.2

Nokel81:
## 🧰 Maintenance

- Fix release process in light of changing to independent versioning (**[#7357](https://github.com/lensapp/lens/pull/7357)**) https://github.com/Nokel81

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -73,20 +73,20 @@
       }
     },
     "node_modules/@babel/core": {
-      "version": "7.21.0",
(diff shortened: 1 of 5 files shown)


jim-docker (approved):
There are fewer files changed here than for alpha.1, do the utility feature packages remain at alpha.1?

Nokel81:
Yes, everything that wasn't changed stays at the previous version.

The command `npx lerna publish` which is run in the `release` workflow publishes all packages to the registry (NPM) if they are not present at the version listed in the package.json.

PR #7156, opened 2023-02-14
Document bundled extension format

Nokel81:
- Bundled extensions are always enabled, and are always compatible
- Have bundled extensions be loaded asyncronously to support
  typescript dynamic import (which is typed) as opposed to require

Signed-off-by: Sebastian Malton <sebastian@malton.name>

resolves #6855

PR #7359, opened 2023-03-15
Fix bundled extensions not being loaded

Nokel81:
- Also show that this fixes it by added an example bundled extension to 'open-lens'

resolves #7156 

PROBLEM: bundled extensions are not registered either on `main` or on `renderer`

SOLUTION: As proposed by @jansav, refactor the injection token into its own package and make it a `peerDependency`


packages/legacy-extension-example/.eslintrc.json
Iku-turso, line 1: Would appreciate own commit for a "legacy-extension-example", maybe even a PR :)

packages/legacy-extension-example/webpack.ts
```diff
@@ -0,0 +1,85 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import path from "path";
+
+const webpack = [
```
Iku-turso, line 7: Not ok to show duplicated webpack configuration in an example package :/

packages/open-lens/src/common/example-bundled-extension.injectable.ts
```diff
@@ -0,0 +1,15 @@
+import { bundledExtensionInjectionToken } from "@k8slens/legacy-extensions";
+import { getInjectable } from "@ogre-tools/injectable";
+import exampleBundledExtensionManifest from "@k8slens/legacy-extension-example/package.json";
+
+const exampleBundledExtensionInjectable = getInjectable({
+  id: "example-bundled-extension",
+  instantiate: (di) => ({
+    manifest: exampleBundledExtensionManifest,
+    main: () => require("@k8slens/legacy-extension-example/main").default,
+    renderer: () => require("@k8slens/legacy-extension-example/renderer").default,
```
Iku-turso, line 16: Exotic inline `require` needs at least comment to explain why.

packages/legacy-extension-example/webpack.ts
Nokel81: This can be fixed in the future

packages/technical-features/application/legacy-extensions/package.json
Iku-turso, line 1: Not completely sure, but this new package doesn't house any `Features`, and therefore shouldn't reside under `technical-features`.

Nokel81: Needs to be said that a lot of what you just said hasn't been publicly discussed anywhere. 😞

Nokel81: We also haven't had any discussions about why we are nesting the packages.

Iku-turso (approved):

PR #7349, opened 2023-03-14
Fix cluster view broken scrolling

aleksfront:
If one iframe is on top of other scrolling being blocked for some reason for one of them.

Adding `visibility:hidden` solves the problem.

https://user-images.githubusercontent.com/9607060/224922401-05efd86b-855a-41fe-a4fc-69f099034ab3.mov

Fixes #7322

PR #7365, opened 2023-03-16
Respect cluster avatars entity background color

aleksfront:
Second part of #3846 

First part https://github.com/lensapp/lens/pull/7349

All cluster icons updated if `backgroundColor` changed in cluster preferences.

salmon cluster 1
salmon cluster 2
salmon cluster 3

### DIFF
--- a/packages/core/src/common/cluster-types.ts
+++ b/packages/core/src/common/cluster-types.ts
@@ -105,7 +105,6 @@ export interface ClusterPreferences extends ClusterPrometheusPreferences {
    * cleared by preferences.
    */
   icon?: string | null;
-  iconBackgroundColor?: string;
   httpsProxy?: string;
   hiddenMetrics?: string[];
   nodeShellImage?: string;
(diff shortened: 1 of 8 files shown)


packages/core/src/common/cluster-types.ts
```diff
@@ -105,6 +105,7 @@ export interface ClusterPreferences extends ClusterPrometheusPreferences {
    * cleared by preferences.
    */
   icon?: string | null;
+  iconBackgroundColor?: string;
```
ixrock: Why do we need another field for `icon`? It's possible I guess to save some `string` with color (HEX/rgba) to `icon` field too and distinguish them easily (when it's not empty) by `prefs.icon.startsWith("data:")` to proper show background.

packages/core/src/renderer/components/hotbar/hotbar-entity-icon.tsx
```diff
@@ -40,6 +42,7 @@ interface Dependencies {
   catalogCategoryRegistry: CatalogCategoryRegistry;
   activeEntity: IComputedValue<CatalogEntity | undefined>;
   navigate: Navigate;
+  getClusterById: (id: string) => Cluster | undefined;
```
ixrock: Why this is here? I haven't found usages of `getClusterById` in  other files..

ixrock (changes requested):

Nokel81 (approved):

ixrock (approved):

PR #7367, opened 2023-03-16
Extract messaging and startable stoppable to a NPM package

jansav:
Motivation is to allow Features to install them using NPM.

### DIFF
--- a/packages/technical-features/messaging/agnostic/src/features/actual/computed-channel/computed-channel.injectable.ts
+++ b/packages/technical-features/messaging/agnostic/src/features/actual/computed-channel/computed-channel.injectable.ts
@@ -1,0 +1,211 @@
+import { getInjectable, getInjectionToken } from "@ogre-tools/injectable";
+
+import {
+  _getGlobalState,
+  computed,
+  IComputedValue,
+  observable,
+  onBecomeObserved,
+  onBecomeUnobserved,
+  reaction,
+  runInAction,
+} from "mobx";
+
+import type { MessageChannel } from "../message/message-channel-listener-injection-token";
+import { getMessageChannelListenerInjectable } from "../message/message-channel-listener-injection-token";
+import { sendMessageToChannelInjectionToken } from "../message/message-to-channel-injection-token.no-coverage";
+import { onLoadOfApplicationInjectionToken } from "@k8slens/application";
+import { pipeline } from "@ogre-tools/fp";
+import { filter, groupBy, map, nth, toPairs } from "lodash/fp";
+import { computedInjectManyInjectable } from "@ogre-tools/injectable-extension-for-mobx";
+import type { JsonPrimitive } from "type-fest";
+
+export type JsonifiableObject =
+  | { [Key in string]?: Jsonifiable }
+  | { toJSON: () => Jsonifiable };
+export type JsonifiableArray = readonly Jsonifiable[];
+export type Jsonifiable = JsonPrimitive | JsonifiableObject | JsonifiableArray;
+
+export type ComputedChannelFactory = <T>(
+  channel: MessageChannel<T>,
+  pendingValue: T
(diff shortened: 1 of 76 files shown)


packages/technical-features/messaging/agnostic/src/features/actual/computed-channel/computed-channel.injectable.ts
```diff
@@ -0,0 +1,209 @@
...
+
+export const duplicateChannelObserverGuardInjectable = getInjectable({
+  id: "duplicate-channel-observer-guard",
+
+  instantiate: (di) => {
+    const computedInjectMany = di.inject(computedInjectManyInjectable);
+
+    return {
+      run: () => {
+        reaction(
```
ixrock, line 129: Shouldn't this be returned for disposing at later point?

ixrock (commented):
Looks like a complex thing. Why do we need all that?

packages/technical-features/messaging/agnostic/src/features/actual/computed-channel/computed-channel.injectable.ts
jansav: I don't think it is, since here we know that it is only ever called once in the lifecycle of the application. Disposing would happen when application is quit. Would add it though, just to be sure but at the moment we don't have anything related to application quit/stop extracted yet.

jansav:
> Looks like a complex thing. Why do we need all that?

Not sure in what context you wrote this, so I'll describe the motivation behind stuff here. This is btw. stuff that will have it's own "geek" article soon.

**Messaging** is an abstraction for communication between processes. The idea is simple, there is `channel` and `channel-listener`. You can `sendMessageToChannel` or `requestFromChannel`. The `channel-listener` handles the message. 

Motivation here is to make us not stuck with the IPC (once there's no direct usages of IPC, we can easily try e.g. websockets to see if there's any performance gains), and allow e.g. behavioural unit tests to test both processes as once (`main` and `renderer`). 

**ComputedChannel**

This is higher level tool for the situation where you have `computed` in `main` and would like to access that in `renderer`. Before this has been done by introducing states and trying to keep those states in sync between processes. However, this does it for you automatically, so your code can be fully reactive and therefore mitigate risks of stale states. 

Example of usage:

**common**

```typescript
const someChannel = getChannel<string[]>('some-channel');
```

**main**

```typescript
const someComputedChannelObserverInjectable = getInjectable({
  id: "some-computed-channel-observer",

  instantiate: (di) => ({
    channel: someChannel,
    observer: computed(() => ["some-string", "some-other-string"])
  }),

  injectionToken: computedChannelObserverInjectionToken,
});
```

**renderer**

```typescript
const computedChannel = di.inject(computedChannelInjectionToken);

const someComputed = computedChannel(someChannel, []); // IComputedValue<string[]>
```

With this example, when some one observes the `someComputed` in `renderer`, it gets the value from the `computedChannelObserver` in `main` and whenever the value changes in `main`, the `renderer` recomputes. That being said, it works just like `computed` but between processes (`main` and `renderer`)

ixrock:
> However, this does it for you automatically, so your code can be fully reactive and therefore mitigate risks of stale states.

So, _keeping some state + process data sync + transport abstraction_ this is what handled under the hood by "computed channel", right?

> That being said, it works just like computed but between processes (main and renderer)

Sounds good. Does it mean we can get rid of data-sync model in `base-store.ts` and replace with that?

jansav:
Closing this in favor of https://github.com/lensapp/lens/pull/7368 for merging being too painful

jansav:
> So, keeping some state + process data sync + transport abstraction this is what handled under the hood by "computed channel", right?

Yes. 

> Sounds good. Does it mean we can get rid of data-sync model in base-store.ts and replace with that?

You're right. :)

PR #7372, opened 2023-03-17
Support using Eviction API where applicable

### LINKED ISSUE #5602: Support using Eviction API when deleting Pods and Deployments
**Describe the bug**
A cursory look of the source does not seem to indicate that the [Kubernetes Eviction API](https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/) is used at all by Lens. This leads me to believe that Deployments and Pods are when they are deleted are so without respect to Pod Disruption Budgets. Actions from the UI I would consider to be voluntary and that has implications based on the documentation.

See this section: https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-budgets
> Cluster managers and hosting providers should use tools which respect PodDisruptionBudgets by calling the [Eviction API](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/#eviction-api) instead of directly deleting pods or deployments.

**Expected behavior**
Utilize the Eviction API for DELETE requests

**Screenshots**

All Links:
https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-budgets
https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/#eviction-api
https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/

Reference: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#create-eviction-pod-v1-core

ixrock:
close #5602


packages/core/src/common/k8s-api/kube-object.ts
```diff
@@ -225,30 +247,60 @@ export interface KubeStatusData {
...
+  public readonly message?: string;
+  public readonly reason?: string;
+  public readonly status?: string;
 
   constructor(data: KubeStatusData) {
     this.apiVersion = data.apiVersion;
     this.code = data.code;
     this.message = data.message || "";
     this.reason = data.reason || "";
+    this.status = data.status || "";
```
Nokel81, line 297: won't these `|| ""` break the `??` used below?

Nokel81 (commented):
would be great if there were a couple unit tests to make sure that future changes to this code don't accidently break the findings here.

Nokel81:
Also reminder that enhancements should never go into patch releases

aleksfront (dismissed):

packages/core/src/common/k8s-api/endpoints/pod.api.ts
```diff
@@ -24,6 +35,38 @@ export class PodApi extends KubeApi<Pod> {
     });
   }
 
+  async evict(resource: DeleteResourceDescriptor) {
```
Nokel81, line 38: Need to have some unit tests for this change

Nokel81 (changes requested):

packages/core/src/common/k8s-api/kube-object.ts
ixrock: Hmm.. I think you're right. Gonna update this.

packages/core/src/renderer/components/kube-object-menu/kube-object-menu.tsx
```diff
@@ -31,8 +31,6 @@ import { observer } from "mobx-react";
 
 export interface KubeObjectMenuProps<TKubeObject extends KubeObject> extends MenuActionsProps {
   object: TKubeObject;
-  editable?: boolean;
-  removable?: boolean;
```
jim-docker: This is just unrelated refactoring?

jim-docker (commented):
In light of https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/#troubleshooting-stuck-evictions we might want to support a "force delete" option in the Lens UI, which directly invokes the delete() method instead of evict(). Could be for another PR...

packages/core/src/renderer/components/kube-object-menu/kube-object-menu.tsx
ixrock: Reverted

ixrock:
> In light of https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/#troubleshooting-stuck-evictions we might want to support a "force delete" option in the Lens UI, which directly invokes the delete() method instead of evict(). Could be for another PR...

How exactly you think we should implement it? Keep separated menu items "Delete" and "Evict" for the Pods menu?

packages/core/src/common/k8s-api/endpoints/pod.api.ts
ixrock: Done. PTAL

Nokel81 (approved):
LGTM!

Nokel81:
> How exactly you think we should implement it? Keep separated menu items "Delete" and "Evict" for the Pods menu?

Maybe we could somehow detect that a pod is being evicted and then show a `Force Delete` action in the menu.

jim-docker:
> > How exactly you think we should implement it? Keep separated menu items "Delete" and "Evict" for the Pods menu?
> 
> Maybe we could somehow detect that a pod is being evicted and then show a `Force Delete` action in the menu.

I thought of separate menu items. "Delete" (or "Evict") and "Force Delete" (or "Kill"). Also, there could be a "Force" checkbox on the delete dialog.

PR #7374, opened 2023-03-17
Fix findClosestVersionTagLessThanVersion in release tool

Nokel81:


aleksfront (approved):

packages/release-tool/src/index.ts
```diff
@@ -120,7 +120,7 @@ function findClosestVersionTagLessThanVersion(tags: string[], version: SemVer):
     .filter(isDefined)
     .filter(version => !version.prerelease.includes("cron"))
     .sort(semver.rcompare)
-    .filter(version => semver.lte(version, version));
+    .filter(v => semver.lte(v, version));
```
ixrock, line 123: What's changed here except renamed variable for filtering?

Nokel81: There were two variables in this function called "version" that was the bug

PR #7375, opened 2023-03-17
Fix resolve system proxy error when no windows available

### LINKED ISSUE #6159: HTTP(S) proxy resolving does not work if no windows open
**Describe the bug**
If app does not have any windows it will log these errors:
```
error:   ┏ Error resolving proxy Tried to resolve proxy for "https://api.k8slens.dev/", but no browser window was available +20s
error:   ┃ [ 1] Error: Tried to resolve proxy for "https://api.k8slens.dev/", but no browser window was available
```

**Expected behavior**
Proxy resolver should work even if windows are closed. Maybe Lens could spawn a lightweight hidden window for this purpose (if main window does not exist)?.

jakolehm:
Fixes #6159


packages/core/src/main/utils/resolve-system-proxy/electron-browser-window.injectable.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Copyright (c) OpenLens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+import { getInjectable } from "@ogre-tools/injectable";
+import type { BrowserWindowConstructorOptions } from "electron";
+import { BrowserWindow } from "electron";
+
+const electronBrowserWindowInjectable = getInjectable({
```
Nokel81, line 9: nit: I would call this `createBrowserWindow`

packages/core/src/main/utils/resolve-system-proxy/electron-browser-window.injectable.ts
```diff
@@ -0,0 +1,19 @@
...
+  id: "electron",
```
Nokel81: The `id` should match the injectable name

Nokel81 (commented):
minor nits

Nokel81 (approved):

PR #7389, opened 2023-03-21
Electron 22.3.3

jakolehm:
- https://releases.electronjs.org/release/v20.0.0
- https://releases.electronjs.org/release/v21.0.0
- https://releases.electronjs.org/release/v22.0.0

Note:
> Electron will be ending support for Windows 7/8/8.1 after version 22.x.y [following Chromium's plan to end support](https://support.google.com/chrome/thread/185534985/sunsetting-support-for-windows-7-8-8-1-in-early-2023?hl=en). Older versions of Electron will continue to work, but no further updates will be made for these operating systems.

### DIFF
--- a/packages/core/src/common/app-paths/app-path-names.ts
+++ b/packages/core/src/common/app-paths/app-path-names.ts
@@ -11,7 +11,6 @@ export const pathNames: PathName[] = [
   "home",
   "appData",
   "userData",
-  "cache",
   "temp",
   "exe",
   "module",
--- a/packages/core/src/common/app-paths/app-path-names.ts
+++ b/packages/core/src/common/app-paths/app-path-names.ts
@@ -11,6 +11,7 @@ export const pathNames: PathName[] = [
   "home",
   "appData",
   "userData",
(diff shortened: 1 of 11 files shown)


aleksfront (dismissed):
LGTM 🚀

Nokel81 (dismissed):

packages/core/src/common/app-paths/app-path-names.ts
```diff
@@ -11,7 +11,7 @@ export const pathNames: PathName[] = [
   "home",
   "appData",
   "userData",
-  "cache",
+  "sessionData",
```
Nokel81, line 14: Do we need this?

jakolehm: I believe we get a type error without 🤔

Nokel81 (approved):

packages/core/src/common/app-paths/app-path-names.ts
Nokel81: ok

PR #7390, opened 2023-03-21
Show Message Box when no update is available after manually checking for updates

### LINKED ISSUE #7228: Show "No Updates Available" after manually checking for updates
Show something like the following:

Lens-Update-Dialog

when the user manually checks for updates and no updates are found.

**Why is this needed**:
Current behaviour feels like something is broken because only the tray icon updates and only for a short while so it seems like nothing has been done.

- Kubernetes distribution: all
- Desktop OS: all

gabriel-mirantis:
fixes #7228


aleksfront (dismissed):

Nokel81 (changes requested):
Adding some behavioural tests would be great!

gabriel-mirantis:
> Adding some behavioural tests would be great!

check

Nokel81 (approved):
LGTM!

Will merge once tests are green

PR #7395, opened 2023-03-22
Adding package exporting injection tokens for cluster settings 

aleksfront:
Third part of https://github.com/lensapp/lens/issues/3846. Second part is https://github.com/lensapp/lens/pull/7365.

Adding `cluster-settings` inner package which exports injection tokens needed for cluster settings icon configuration.


packages/cluster-settings/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "types": "./dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "clean": "rimraf dist/",
+    "generate-types": "tsc --d --declarationDir ./dist --declarationMap --emitDeclarationOnly",
+    "build": "npm run generate-types && swc ./src/index.ts -d ./dist"
+  },
+  "devDependencies": {
```
ixrock, line 23: Should we include `typescript` as dev dependency as well?

packages/core/src/renderer/components/cluster-settings/__tests__/icon-settings.test.tsx
```diff
@@ -53,6 +56,29 @@ const newMenuItem = getInjectable({
   injectionToken: clusterIconSettingsMenuInjectionToken,
 });
 
+function customSettingsComponent(props: ClusterIconSettingComponentProps) {
```
ixrock: _nit:_ 

```suggestion
function CustomSettingsComponent(props: ClusterIconSettingComponentProps) {
```

packages/core/src/renderer/components/cluster-settings/__tests__/icon-settings.test.tsx
```diff
@@ -39,6 +40,8 @@ const clusterEntity = new KubernetesCluster({
   },
 });
 
+cluster.preferences.clusterName = "some-cluster-name";
```
ixrock: Looking at `ClusterModel` (first argument of `Cluster`) it seems you can pass `preferences` in the constructor above.

```ts
export interface ClusterModel {
  /** Unique id for a cluster */
  id: ClusterId;

  /** Path to cluster kubeconfig */
  kubeConfigPath: string;

  /** User context in kubeconfig  */
  contextName: string;

  /** Preferences */
  preferences?: ClusterPreferences;

  /** Metadata */
  metadata?: ClusterMetadata;

  /** List of accessible namespaces */

  /**
   * Labels for the catalog entity
   */
  labels?: Partial<Record<string, string>>;
}
```

ixrock (dismissed):

packages/cluster-settings/package.json
aleksfront: Not sure. It works well without specify typescript here. Probably it uses some globally installed typescript somehow.

ixrock: Yes, try to remove global ‘typescript’ package and check again.

ixrock (approved):
LGTM

packages/cluster-settings/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@k8slens/cluster-settings",
+  "version": "6.5.0-alpha.1",
```
Nokel81, line 3: Nit: this should probably be `1.0.0-alpha.1`

aleksfront: Following same versioning as other internal packages. Keeping `6.5.0-alpha.1` for now until clarification. Will change it if needed in follow up PR.

PR #7396, opened 2023-03-22
Fix Check for updates menu item title capitalisation

aleksfront:
Minor: follow same capitalisation rules as for other menu items.

`Check for updates` become `Check for Updates`.

check for updates before

### DIFF
--- a/packages/core/src/features/application-update/child-features/application-update-using-application-menu/main/check-for-updates-menu-item.injectable.ts
+++ b/packages/core/src/features/application-update/child-features/application-update-using-application-menu/main/check-for-updates-menu-item.injectable.ts
@@ -28,7 +28,7 @@ const checkForUpdatesMenuItemInjectable = getInjectable({
       id: "check-for-updates",
       parentId: isMac ? "mac" : "help",
       orderNumber: isMac ? 20 : 50,
-      label: "Check for Updates",
+      label: "Check for Updates...",
       isShown: updatingIsEnabled,
 
       onClick: () => {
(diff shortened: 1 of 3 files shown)


packages/core/src/features/application-update/child-features/application-update-using-application-menu/main/check-for-updates-menu-item.injectable.ts
```diff
@@ -28,7 +28,7 @@ const checkForUpdatesMenuItemInjectable = getInjectable({
       id: "check-for-updates",
       parentId: isMac ? "mac" : "help",
       orderNumber: isMac ? 20 : 50,
-      label: "Check for updates",
+      label: "Check for Updates",
```
gabriel-mirantis: It feels off to me that we have different text from menu to tray.
It seems that if we have a command for `Check for Updates` in one place, then it should be the same without the `...`

gabriel-mirantis (dismissed):

packages/core/src/features/application-update/child-features/application-update-using-application-menu/main/check-for-updates-menu-item.injectable.ts
aleksfront: My guess it `...` shows kind of ongoing progress. That is why it is being used in `Checking for Updates...`. As far I remember at least on Windows three-dots (ellipsis) symbol was added to the menu items which open modal dialogs. So there's no clear rule. But I agree that it should be consistent at least. Adding tree-dots to every associated place...

gabriel-mirantis (approved):

packages/core/src/features/application-update/child-features/application-update-using-application-menu/main/check-for-updates-menu-item.injectable.ts
gabriel-mirantis: You solved it like Slack.  Slack has the ... from the menu - my guess is that if we pick any UI changes up from Slack then they are probably the correct choice, their UI/UX feels pretty modern to me

